PR #13, opened 2013-07-30
[ignore] Extended .gitignore

dizzzz:
Extended file to ignore more logfiles and build generated files


joewiz:
It's good to know that a 10 year old, already-merged PR has no code smells! ;)

PR #127, opened 2014-01-22
Remove URLDecoder call from path parsing

hungerburg:
This brings GET handling in line with PUT handling, to conform to http://tools.ietf.org/html/rfc3986#section-3.3

Most of what the URLDecoder call does is change "+" into "%20" which is clearly an error.

(Actually it also removes double-encodings, which though should not be there from the beginning: eg turns "%252B" into "%2B".)

### DIFF
--- a/build/scripts/junit.xml
+++ b/build/scripts/junit.xml
@@ -162,6 +162,7 @@
 	    	<test fork="yes" name="org.exist.collections.AllCollectionTests" todir="${junit.reports.dat}"/>
 	    	<!-- <test fork="yes" name="org.exist.fulltext.AllFulltextTests" todir="${junit.reports.dat}"/> -->
             <test fork="yes" name="org.exist.security.XMLDBSecurityTest" todir="${junit.reports.dat}"/>
+            <test fork="yes" name="org.exist.security.SecurityManagerRoundtripTest" todir="${junit.reports.dat}"/>
 
             <!-- Execute all other tests except those that have to be called manually.   -->
             <batchtest fork="yes" todir="${junit.reports.dat}">
--- a/test/src/org/exist/security/SecurityManagerRoundtripTest.java
+++ b/test/src/org/exist/security/SecurityManagerRoundtripTest.java
@@ -97,7 +97,7 @@ public class SecurityManagerRoundtripTest {
             root = DatabaseManager.getCollection(baseUri + "/db", "admin", "");
             ums = (UserManagementService)root.getService("UserManagementService", "1.0");
 
-            user = ums.getAccount("testUser");
+            user = ums.getAccount(userName);
             assertNotNull(user);
 
             Group defaultGroup = user.getDefaultGroup();
@@ -150,7 +150,7 @@ public class SecurityManagerRoundtripTest {
             root = DatabaseManager.getCollection(baseUri + "/db", "admin", "");
             ums = (UserManagementService)root.getService("UserManagementService", "1.0");
 
-            user = ums.getAccount("testUser");
+            user = ums.getAccount(userName);
             assertNotNull(user);
 
             Group defaultGroup = user.getDefaultGroup();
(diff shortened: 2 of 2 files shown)


adamretter:
@hungerburg I have done a refactoring and rewrite of the RESTServer https://github.com/eXist-db/exist/pull/549. At the moment I have left the URLDecoder as it is. However I would like to help fix this. What I am really missing is good test cases, would you consider submitting a pull-request which adds several tests around URL encoding to https://github.com/eXist-db/exist/pull/549/files#diff-cae5e7fb8523338720a7e9e8d1205be2

From those test cases and the work you have already started here, I think we should be able to get this solved.

hungerburg:
@adamretter Adam, I am a little at a loss; I guess I should first fork something, so that I can pull; but then I have forked eXist already, and I am 967 commits behind eXist-db:develop -- no idea how to get up to current though, without messing up the pull requests that already are. Rather dump the old fork…

Still willing to join the effort, the most important thing I have learned studying the issue, please consider carefully: passing URIs as Strings causes headaches. The conversion of a string into a URI is not trivial at all. Even the JDK has what might be bugs in that regard.

I see that the tests use java.net.HttpURLConnection, which takes a java.net.URL as its argument, yet the tests always pass Strings. Such a conversion is implicit and prone to errors. Perhaps, in the tests, the passed strings are not ambigous ever. The tests also use XmldbURI, which is another unknown in the process (You might be better off eliminating it, if feasible). Anyways, "String uri" is something, that you do not want to read in your code, ever.

Perhaps It would be best to construct a java.net.URI with the 7 arguments constructor and then URI.toURL() that for the call to HttpURLConnection. That might be the most sane base to verify the workings ot the rest server. I would have to first write some class to learn more about this conversion though, before I could amend the tests, because the offending cases pound on such inconsistencies. So at the moment, from me, only words. Please excuse, the subject always threatens to outrun my developer skills.

adamretter:
@hungerburg Getting your fork up to date is covered here - https://github.com/eXist-db/exist#syncing-a-fork. I would suggest making backup of your working copy first though, in case you have any problems.

Also I was not suggesting that you should use `HttpURLConnection` for your specific test cases. You can use any HTTP Library you want. For example, Apache Http Commons is also included in eXist.

hungerburg:
@adamretter Adam, would you tell me how to run just the RESTServiceTest (not the full suit "./build.sh test") ?

adamretter:
@hungerburg I don't know of a way of doing it from Ant... I normally do it from my IDE.

ljo:
@hungerburg you can do it with bin/run.sh org.exist.http.RESTServiceTest

hungerburg:
@ljo Thank you, that helped: I had to add `"$EXIST_HOME"/lib/test` to `JAVA_ENDORSED_DIRS` of run.sh and then could start the test like this: `bin/run.sh  org.junit.runner.JUnitCore org.exist.http.RESTServiceTest`

hungerburg:
So, now the pull request contains another patch - Unfortunately, its not easy to single this one out; Some knowledge about branching seems required on my side. The patch is at https://github.com/eXist-db/exist/commit/0b56260067072d449e8e9fb20524df8dcf82338a

Running the test locally looks fine: apart from the colon, a resource containing these tricky RFC-3986 pchars ("+" etc.) can be both PUT and GOT by the same name in the unit test case.

Beware though, that my fork has some other patches (see the pull request) -- The results on Your base may be quite different!

Curiously though, when starting the db normally (bin/startup.sh), "http://localhost:8080/exist/rest/db/test/A-Za-z0-9_~!$&'()*+,;=@.xml" gives a 404 in Firefox, while Lynx downloads the thing! (A look in wireshark tells me, that Firefox escapes the single-quote/apostrophe character in its GET request.)

hungerburg:
@adamretter Adam, I did now manage to create a local develop branch, that is the same as upstream/develop; and my changes contained in another branch. I fear though, that (force) pushing this to origin (my fork) will make all the references in this pull request orphans and probably destroy this whole thread here (pull 127). Yet, this looks like the only way to sort the mess, that the history of my fork represents, a little like starting over from a blank slate. Have you got any ideas, how to proceed?

hungerburg:
Cool, the chat is still here! The new pull is #605 and it is much less cluttered and also it is against its own branch. As for the labels, decide as you wish.

PR #158, opened 2014-03-02
Added fn:parse-xml() and fn:parse-xml-fragment() functions.

claudius108:


claudius108:
I will look at it tomorrow.

claudius108:
Well, I check in eXist 4.* to generate a similar document with
```xquery
document {(
    text {"He was "}
    ,
    element i {"so"}
    ,
    text {" kind"}
)}
```
and I got "He was " as output.

I wonder if this is related to the bug you mentioned. Does eXist support the output needed got parse-xml-fragment("He was <i>so</i> kind")?

joewiz:
Thanks for looking into this, Claudius! Are you running your query in eXide? If so, uncheck the "Highlight Index Matches" checkbox above the query results pane, and you should see the expected results. The truncated results you're seeing are the result of a bug with the `util:expand()` function that I reported at https://github.com/eXist-db/exist/issues/1562.

PR #285, opened 2014-08-05
CLI: removed NPEs for users without group

opax:
A quick fix for several NullPointerExceptions when dealing with users without groups. 
I saw there are related tickets (I have not checked whether this patch may help there, too), and this may not be the whole deal, but it is still an improvement.
I just had this patch hanging around for a while, so I thought a pull request was in order.


dizzzz:
@adamretter could you have a look?

adamretter:
@opax I think this issue is better fixed at the source. We certainly should not be adding the UNDEFINED_ID as the group of a user in an ACL. I have fixed your root cause here: https://github.com/eXist-db/exist/pull/587.

I am going to close this pull-request, feel free to open it again if you feel that I have not addressed your problem.

PR #286, opened 2014-08-05
enhancement for in wrapper script in real-world deployments

opax:
Deploying eXist-db on a server usually involves setting up a special user that is the owner of the process. I found that if that user does not have a valid shell (/bin/false or /usr/sbin/nologin is not uncommon for daemon accounts), the wrapper script for starting the database fails (when using RUN_AS_USER).

This is a trivial patch that fixes the problem.


dizzzz:
Closing, I think it is fixed

opax:
Yes, it is addressed in wrapper v3.5.26.

Issue #187, opened 2014-04-01
Untranslated numerical entity &#160; in UTF-8 serialization Question + Patch


pcolot01:
In the default context or encoding set to UTF-8, Every numérical entity is converted as a character represented by his binary value. By example: &#65; became A (binary 65)

All except the non breaking space, follow this rule.
The non breaking space is always converted as &#160;

This is happening at a so low level, it seems impossible to force it. Especially if you want to zip on the fly the result.

What is the reason of this exception ? Is it possible to define charset mapping ?

I identified the source of this behaviour in org.exist.util.serializer.XMLWriter

in the method:

``` java
    private void writeChars(final CharSequence s, final boolean inAttribute) throws IOException {
        final boolean[] specialChars = inAttribute ? attrSpecialChars : textSpecialChars;
        char ch = 0;
        final int len = s.length();
        int pos = 0, i;
        while(pos < len) {
            i = pos;
            while(i < len) {
                ch = s.charAt(i);
                if(ch < 128) {
                    if(specialChars[ch]) {
                        break;
                    } else {
                        i++;
                    }
                } else if(!charSet.inCharacterSet(ch) || ch == 160) {
                                break;
                } else {
                    i++;
                }
            }
            writeCharSeq(s, pos, i);
            // writer.write(s.subSequence(pos, i).toString());

            if (i >= len) {
                    return;
            }

            if(needsEscape(ch)) {
                switch(ch) {
                    case '<':
                        writer.write("&lt;");
                        break;
                    case '>':
                        writer.write("&gt;");
                        break;
                    case '&':
                        writer.write("&amp;");
                        break;
                    case '\r':
                        writer.write("&#xD;");
                        break;
                    case '\n':
                        writer.write("&#xA;");
                        break;
                    case '\t':
                        writer.write("&#x9;");
                        break;
                    case '"':
                        writer.write("&#34;");
                        break;
                    // non-breaking space:
                    case 160:
                        writer.write("&#160;");
                        break;
                    default:
                        writeCharacterReference(ch);
                }
            } else {
                writer.write(ch);
            }

            pos = ++i;
        }
    }
```

I replace it by 

``` java
  private void writeChars(final CharSequence s, final boolean inAttribute) throws IOException {
        final boolean[] specialChars = inAttribute ? attrSpecialChars : textSpecialChars;
        char ch = 0;
        final int len = s.length();
        int pos = 0, i;
        while(pos < len) {
            i = pos;
            while(i < len) {
[...]

PR #408, opened 2014-12-04
Avoid implicit conversion of unicode \x160 as  numerical entity &#160;

pcolot01:
In the serialized result, unicode character \x160 are converted as  numerical entity &amp; # 1 6 0 ;

Even if this is equivalent, on an XML point of view, implicit and unexpected conversions
are source of problems.

See issue (https://github.com/eXist-db/exist/issues/187)

### DIFF
--- a/src/org/exist/util/serializer/XMLWriter.java
+++ b/src/org/exist/util/serializer/XMLWriter.java
@@ -496,7 +496,7 @@ public class XMLWriter {
                     } else {
                         i++;
                     }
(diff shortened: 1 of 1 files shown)


dizzzz:
Any judgements? @ljo @wolfgangmm @adamretter ?

adamretter:
@wolgangmm could we get your comment on this before 3.0 please?

dizzzz:
@wolfgangmm  please could you share some thoughts?

adamretter:
@wolfgangmm what is the status of this please?

PR #411, opened 2014-12-09
Dist script not producing usable dist

gtnewton:
Running build dist produces a distribution with no autodeploy directory, and unsigned jars that netx won't run.
Adjusted dist.xml build script to copy autodeploy dir to dist, and signjar stanza to re-sign the jars that get re-compiled when the dist script is called.

### DIFF
--- a/build/scripts/dist.xml
+++ b/build/scripts/dist.xml
@@ -312,6 +312,17 @@
                 <include name="wrapper"/>
             </fileset>
         </chmod>
+        
+        <!-- Re-sign jar files that get re-built when this script gets invoked -->
+        <signjar alias="exist" storepass="secret" keystore="key.store">
+            <fileset dir="${dist.dir}">
+                <include name="exist*.jar"/>
+                <include name="start.jar"/>
+                <include name="exist-optional.jar"/>
+                <include name="examples.jar"/>
+                <include name="lib/extensions/exist-modules.jar"/>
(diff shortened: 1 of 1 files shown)


build/scripts/dist.xml
```diff
@@ -197,7 +197,14 @@
         <touch file="${dist.dir}/log/.DO_NOT_DELETE"/>
 
         <filter token="database" value="native"/>
-
+        <mkdir dir="${dist.dir}/autodeploy"/>
+        <mkdir dir="${dist.dir}/lib/test"/>
+        <copy todir="${dist.dir}/autodeploy">
```
dizzzz: the autodeploy directory can be best copied with the fileset in the copy task, see line 201

build/scripts/dist.xml
```diff
@@ -305,6 +312,17 @@
                 <include name="wrapper"/>
             </fileset>
         </chmod>
+        
+        <!-- Re-sign jar files that get re-built when this script gets invoked -->
```
dizzzz: I don;t think we can rely on the resign here; when needed at al, better is to call the resign-tasks (which verifies for the availability of the keystore file. just calling resign is not a good idea

dizzzz:
I'd suggest to close this PR ; I'll add the re-sign step in my PR (when the design allows, need to check)

PR #431, opened 2014-12-23
Support for Zip/UnZip and Tar/UnTar using local

pcolot01:
The actual compression library is not compatible with old style Zip and Tar. Processing of historical Zips, including current Microsoft Windows compressed directories, sometimes generates MALFORMED exception.

Zip and Tar specification did not precise the character encoding of file names included. The default local was used. Zips were often using Cp437. Finally, to solve this problem, the file name table has been extended with an additional unicode file name table. A flag has been added  to inform modern tools to use it.

Original Zip files with including filename using Cp437 character like ö, etc are generating MALFORMED error.

Starting with Java 7, the new ZipInputStream API allows to force the current file name encoding (http://docs.oracle.com/javase/7/docs/api/java/util/zip/ZipInputStream.html#ZipInputStream(java.io.InputStream,%20java.nio.charset.Charset) .

I extended the current compression module to support the encoding as an additional parameter. I added and integrated a test suite demonstrating the accurate execution of it.

Best regards


dizzzz:
We should review this PR.... @adamretter @wolfgangmm @ljo @shabanovd @dizzzz ?

adamretter:
Okay, so I will take ownership of this one as it needs quite a lot of cleanup in its commit history which I am happy to do. I will get this sorted in the next few days.

adamretter:
@pcolot01 Sorry this took so long to get resolved. I cleaned up your pull-request and squashed many of the commits down and removed some of the churn. Thanks for your contribution :-)

PR #446, opened 2015-01-25
XQuery 3.1 array and json support

wolfgangmm:
Complete implementation of XQuery 3.1 arrays and json parsing/serialization (as of the December 2014 version of the specs). Fully backwards compatible with eXist 2.2 (no changes to existing code required). Use xquery version "3.1" to enable the new features.

* array constructors
* lookup operator, including wildcard, on maps and arrays
* all array: functions from XQuery 3.1 functions spec, except array:sort (which is not really vital right now and will be added later)
* JSON serialization support as defined by spec. To avoid breaking backwards compatibility with 2.2, the old JSON serializer is selected if a single element node is passed in.
* fn:json-parse and fn:json-doc functions
* fn:apply function
* atomization rules updated to match XQuery 3.1 (atomizining an array may return more than one value).
* updated fn:deep-equal, fn:data and others to properly handle arrays.

See [test cases](https://github.com/wolfgangmm/exist/blob/arrays/test/src/xquery/arrays/arrays.xql) for some examples.


joewiz:
Whoo hoo!  Very exciting!

PR #454, opened 2015-01-28
[refactor] Use ConcurrentHashMap instead of homebrewed approach

adamretter:
So this is a very simple change, and I *think* the concurrency model should be fine.

HOWEVER... Dear merge agent, please please think about this change and confirm my thinking before merging. This is an important change in a delicate area.

For sure the test suite still passes, but is that enough?


dizzzz:
@wolfgangmm ?

shabanovd:
What reasons for such change?

adamretter:
I wanted to use standard java classes and it makes it much simpler to debug hung brokers

dizzzz:
@adamretter that makes sense to me; using std libs and APIs will make the product less complex ...

shabanovd:
@dizzzz, you are optimist -)

adamretter:
@shabanovd Do you see a reason not to make this change?

shabanovd:
there is no problem at this part of code and it do use "std libs and APIs" as much as possible

adamretter:
So hear is my thinking:

1) `org.exist.util.hashtable.MapRWLock` is completely bespoke. Whilst the locking in there looks okay to me at a glance. The class is only used for this single thing in BrokerPool and is redundant if this pull-request is accepted.

2) `org.exist.util.hashtable.MapRWLock` uses a lock and function passing interface (`org.exist.util.hashtable.MapRWLock.LongOperation`) to achieve iterative operations. ConcurrentHashMap does not need to lock for iterative access as its iterators are stable and also it is more efficient as it does not require calling a passed function.

3) Several Map interface methods of MapRWLock are not implemented. This makes interrogating the content of the map in real-time when debugging impossible.

4) Not directly related to locking but, IdentityHashMap is the underlying map used for activeBrokers which makes no sense. If you use a HashMap and your key object does not implement `equals(java.lang.Object)`, then according to the javadoc for `Object.equals` you will fallback to object identity anyway. As our keys are instances of the Thread class, which does not implement `equals`, we can just use HashMap instead of identity HashMap.

5) Don't reinvent the wheel. In this instance it looks to me like there are standard Java classes that offer the functionality so there is no need to write our own. Also it is likely that `ConcurrentHashMap` offers better performance than MapRWLock as it does not necessarily need to lock reads.

It may be that the concurrency model for ConcurrentHashMap is not correct for this situation, certainly it is different to MapRWLock, however after giving it some thought I think it should be okay.

@shabanovd (and everyone else) if you can see that I have missed something and the concurrency model for BrokerPool is violated by switching to ConcurrentHashMap then please do let me know, otherwise I think this change brings the above improvements.

shabanovd:
1-4) MapRWLock was code for use at BrokerPool (only) to get away from synchronized methods and give better concurrent access for get/release methods **without change of behaviour model** (that's why locks there and there was IdentityHashMap before it) 

5) are there any performance issues?

@adamretter, you asking to change behaviour model because if looks cooler? I'm not objecting, but "it **should** be okay" do not make me happy ... waiting for "it **must** be okay"

adamretter:
@shabanovd  Yes I understand that MapRWLock was an improvement over what we had before :-). I am not doing this because it is "cooler" I am doing this because I need to be able to interrogate the map from the debugger (IntelliJ) and you cannot do that with MapRWLock because of its missing methods. Also it reduces *maybe* unnecessary code

shabanovd:
What methods are missing? Maybe simpler to code missing methods?

shabanovd:
to be clear, only one thing concern me: read operation is not snapshot in time with ConcurrentHashMap in heavy loaded systems

ljo:
@shabanovd I agree the reads under heavy load is a concern, but being able to debug it sanely is also of huge importance as argued for by @adamretter.

adamretter:
@shabanovd Exactly. I could have implemented the missing methods or switched to `ConcurrentHashMap`. I chose the benefits of switching and maintaining less custom code.

Actually a read on ConcurrentHashMap is exactly a snapshot at a point in time, as the JavaDoc states: "Retrievals reflect the results of the most recently completed update operations holding upon their onset.". The difference is that updates can happen at the same time, but this does not effect the read as the read is point-in-time. The next read will see the update if it completed.

dizzzz:
@adamretter @shabanovd @wolfgangmm so.... this change is in essence ok?

wolfgangmm:
I basically think the change should be ok, but as usual, I pulled it into my development version to see if any issues pop up while I use it. I will also do some stress testing with one app for which I have jmeter tests to compare with.

shabanovd:
@wolfgangmm any news on this?

wolfgangmm:
@shabanovd ah, yes, I was running the changes against a jmeter stress test, but while I was doing that, I encountered a memory leak and had to fix that first, so I'm not sure what the outcome of the stress test was in the end. I'll repeat it tonight, though I think it should be ok ...

adamretter:
@wolfgangmm Were you able to repeat your JMeter tests?

PR #463, opened 2015-02-06
Regression: messed up namespaces in constructed document

wolfgangmm:
Recent changes to QName class mess up namespaces in in-memory documents. NamePool needs to take prefix and nameType into account when testing for equality.

See http://markmail.org/message/jdub2fv6vj7pm5pd for bug report.


dizzzz:
oops I see a timeout....

https://s3.amazonaws.com/archive.travis-ci.org/jobs/49711973/log.txt ; not related though

adamretter:
As this is a regression, please can we get a test case that demonstrates this. That would stop the same thing from happening again

dizzzz:
dont shoot the messenger :-)

PR #468, opened 2015-02-10
Support for Regex within Spans in Lucene

APivetta:
I recently updated from eXist 2.1 to 2.2 and I came across the problem that Regex queries are no longer supported within Span queries.
Given that the class SpanRegexQuery was deprecated in Lucene 4.0 and is no longer available in exist 2.2, I propose to use the replacement class provided by lucene: SpanMultiTermQueryWrapper<RegexQuery>(new RegexQuery()) to keep supporting the use of Regex within Spans queries.

This has fixed the issue for me, hope you consider ir.
Kind regards.


joewiz:
Wow, this would be great.  Do any of the existing unit tests cover the functionality you're providing here? E.g., https://github.com/eXist-db/exist/blob/develop/extensions/indexes/lucene/test/src/xquery/lucene/queries.xml#L243.  If not, could you add relevant tests?

dizzzz:
A test case would be very welcome....

jensopetersen:
Added a simple test at https://github.com/eXist-db/exist/pull/469.

PR #471, opened 2015-02-14
Only compile aspects with AspectJ compiler

adamretter:
* Can now compile eXist Aspects successfully with Java 8

### DIFF
--- a/build/scripts/build-impl.xml
+++ b/build/scripts/build-impl.xml
@@ -51,8 +51,7 @@
     <!-- additional set of ant tasks -->
     <property name="asocat-exist.jar" location="${tools.ant}/lib/asocat-exist.jar"/>
     
-    <!-- setup class path -->
-    <path id="classpath.core">
+    <path id="classpath.external">
         <fileset dir="${lib.core}">
             <include name="*.jar"/>
         </fileset>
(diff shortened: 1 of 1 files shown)


shabanovd:
will this speed up compilation procedure in general?

adamretter:
Yes this also removes the need to compile all Java classes with `javac` and `iajc`. I see a compilation speed up of ~50% here.

shabanovd:
So, no problems, just speed up?

dizzzz:
And.... we can use fully use features of java 8...

dizzzz:
@adamretter that is right? now we can change build.properties?

dizzzz:
@adamretter we need these changes in the netbeans scripts as well? https://github.com/eXist-db/exist/blob/1ed92e2c5423bd0e533645b0830daf8525cbe194/nbbuild.xml#L81  prepping on aspect there a few more script?  I realise there is some dead wood around https://github.com/eXist-db/exist/search?utf8=✓&q=aspectj

adamretter:
@shabanovd No problems.

@dizzzz Yes you can use 1.8 in build.properties on a JDK 8 now. However, I am still working on moving us to JDK 8 properly, as there are 5 failing tests when you switch 1.7 for 1.8. That work is going on here - https://github.com/adamretter/exist/commits/j8. Regards NetBeans, I don't use it anymore, any change you still do, if so can you update that build script based on my change to `build-imp`l.xml?

dizzzz:
Yes sure !

Issue #474, opened 2015-02-16
some windows batch files have wrong line ends


nicg-corbas:
The Windows batch files in the 2.2 release appear to have been saved with Linux line endings.

dizzzz:
Could you elaborate which of the scripts are affected?

nicg-corbas:
Apologies - should have been much more clear - I'm seeing client.bat, check_jmx_status.bat and get_opts.bat all with (proper operating system) LF only linebreaks

PR #478, opened 2015-02-17
adding Windows line endings (CRLF) to .bat files

jensopetersen:
https://github.com/eXist-db/exist/issues/474


shabanovd:
git may overwrite it

jensopetersen:
One file (don't remember which) had correct line endings, so perhaps it does not ….

shabanovd:
it will each time someone commit to that file. http://stackoverflow.com/questions/1967370/git-replacing-lf-with-crlf

jensopetersen:
Yes, you are right. All existing .bat files have LF only, I was wrong. Please close PR without merging. 

So a Windows user should set git config core.autocrlf?

dizzzz:
The installer can fix it right?

PR #487, opened 2015-02-20
Fixes for names in XQuery 3

adamretter:
* Support for URIBracedLiteral
* Correct use of EQName instead of QName in XQuery 3.0.


dizzzz:
wow, what a work! I fear I can't judge the details here, it looks pretty good. @wolfgangmm ? @ljo ? @shabanovd

wolfgangmm:
I'm checking for compatibility issues with existing apps ... Will merge if ok.

adamretter:
I don't expect any compatibility issues, as whilst references to `qName` have been replaced with `eqName`. The spec itself ensures backwards compatibility by making `eqName` a disjunction of `qName` and `URIQualifiedName`.

wolfgangmm:
Handling qnames in the parser has always been a bit tricky, so I checked if any of the known issues (e.g. errors if one uses a keyword like "for" as element name) would pop up. It seems all ok, though.

Issue #486, opened 2015-02-19
XQuery Parser issue with call to inline function


adamretter:
The following query:

```xquery
function ($x) {
    $x
}("hello")
```

Results an exception from the parser:
```
org.xmldb.api.base.XMLDBException: exerr:ERROR <AST>:0:0: unexpected AST node: DynamicFunction
 [at line 6, column 10]
	at org.exist.xmldb.LocalXPathQueryService.compile(LocalXPathQueryService.java:259)
	at org.exist.client.QueryDialog$QueryThread.run(QueryDialog.java:600)
Caused by: org.exist.xquery.StaticXQueryException: exerr:ERROR <AST>:0:0: unexpected AST node: DynamicFunction
 [at line 6, column 10]
	at org.exist.xquery.XQuery.compile(XQuery.java:165)
	at org.exist.xquery.XQuery.compile(XQuery.java:109)
	at org.exist.xquery.XQuery.compile(XQuery.java:93)
	at org.exist.xquery.XQuery.compile(XQuery.java:85)
	at org.exist.xmldb.LocalXPathQueryService.compileAndCheck(LocalXPathQueryService.java:272)
	at org.exist.xmldb.LocalXPathQueryService.compile(LocalXPathQueryService.java:257)
	... 1 more
Caused by: <AST>:0:0: unexpected AST node: DynamicFunction
	at org.exist.xquery.parser.XQueryTreeParser.expr(XQueryTreeParser.java:3787)
	at org.exist.xquery.parser.XQueryTreeParser.mainModule(XQueryTreeParser.java:4220)
	at org.exist.xquery.parser.XQueryTreeParser.module(XQueryTreeParser.java:4153)
	at org.exist.xquery.parser.XQueryTreeParser.xpath(XQueryTreeParser.java:3807)
	at org.exist.xquery.XQuery.compile(XQuery.java:160)
	... 6 more
Caused by: org.exist.xquery.StaticXQueryException: exerr:ERROR <AST>:0:0: unexpected AST node: DynamicFunction
 [at line 6, column 10]
	at org.exist.xquery.XQuery.compile(XQuery.java:165)
	at org.exist.xquery.XQuery.compile(XQuery.java:109)
	at org.exist.xquery.XQuery.compile(XQuery.java:93)
	at org.exist.xquery.XQuery.compile(XQuery.java:85)
	at org.exist.xmldb.LocalXPathQueryService.compileAndCheck(LocalXPathQueryService.java:272)
[...]

PR #488, opened 2015-02-21
[bugfix] Better distinguish between inline functions

wolfgangmm:
and functions declared in prolog. Closes bug #486


adamretter:
Awesome thanks @wolfgangmm. I think this also cleans up the previous approach of functionDecl for inline functions... Nice :-)

PR #490, opened 2015-02-24
Automatic Resource Management

adamretter:
By using Java 7s `try-with-with-resources` functionality for `Txn` and `DBBroker` we can ensure that a transaction is always automatically closed and that a broker is always automatically returned to the broker pool.

For example, consider a common access pattern in eXist:
```java
DBBroker broker = null;
TransactionManager transact = null;
Txn transaction = null;
try {
  broker = pool.get(pool.getSecurityManager().getSystemSubject());
  transact = pool.getTransactionManager();
  transaction = transact.beginTransaction();

  //various db access here

  transact.commit(transaction);
} catch(Exception e) {
  if(transact != null) {
    transact.abort(transaction);
    throw e;
  }
} finally {
  if(transact != null) {
    transact.abort(transaction);
  }
  if(broker != null) {
    pool.release(broker);
  }
}
```

Instead, through adopting `try-with-resources` we can use the much simpler and safer pattern:
```java
final TransactionManager transact = pool.getTransactionManager();
try(final DBBroker broker = pool.get(pool.getSecurityManager().getSystemSubject());
    final Txn transaction = transact.beginTransaction()) {

  //various db access here

  transact.commit(transaction);
}
```

### DIFF
--- a/extensions/debuggee/test/org/exist/debugger/DebuggerTest.java
+++ b/extensions/debuggee/test/org/exist/debugger/DebuggerTest.java
@@ -489,18 +489,11 @@ public class DebuggerTest implements ResponseListener {
 	}
 	
     private void store(String name,  String data) throws EXistException {
-    	Database pool = BrokerPool.getInstance();;
-    	
-        DBBroker broker = null;
-        TransactionManager transact = null;
-        Txn transaction = null;
-        try {
-            broker = pool.get(pool.getSecurityManager().getSystemSubject());
-            assertNotNull(broker);
-            transact = pool.getTransactionManager();
-            assertNotNull(transact);
-            transaction = transact.beginTransaction();
-            assertNotNull(transaction);
+    	Database pool = BrokerPool.getInstance();
+        final TransactionManager = transact = pool.getTransactionManager();
+
+        try(final DBBroker broker = pool.get(pool.getSecurityManager().getSystemSubject());
+            final Txn transaction = transact.beginTransaction()) {
 
             Collection root = broker.getOrCreateCollection(transaction, TestConstants.TEST_COLLECTION_URI);
     		broker.saveCollection(transaction, root);
@@ -510,12 +503,8 @@ public class DebuggerTest implements ResponseListener {
 
             transact.commit(transaction);
         } catch (Exception e) {
-            if (transact != null)
-                transact.abort(transaction);
             e.printStackTrace();
             fail(e.getMessage());
-        } finally {
-            pool.release(broker);
         }
     }
 
--- a/extensions/indexes/lucene/test/src/org/exist/indexing/lucene/LuceneIndexTest.java
+++ b/extensions/indexes/lucene/test/src/org/exist/indexing/lucene/LuceneIndexTest.java
@@ -27,6 +27,7 @@ import java.io.File;
 import java.io.StringReader;
 import java.util.*;
 
+import org.exist.EXistException;
 import org.exist.Indexer;
 import org.exist.TestUtils;
 import org.exist.collections.Collection;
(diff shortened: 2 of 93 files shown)


dizzzz:
Nice....  but is it backward compatible? e.g.: do I need to update all my extensions to make them build?

adamretter:
There are no API changes, so yes it's backwards compatible.

shabanovd:
more than that it possible to do 
```
try (Txn txn = broker.beginTx()) {
    ...
    txn.success();
}
```

You miss call of 'success' method and redundant call 'commit' because Txn implement AutoCloseable interface

adamretter:
@shabanovd Sure. However, it would now be `txn.commit()` as I renamed it, because `success` implies a completed result, whereas `commit` implies an action.

joewiz:
This sounds like a great feature.  Will it by chance help reduce or eliminate the chance of database corruption?  Just last night I experienced db corruption with an XQuery Update and find it deeply frustrating when this happens; it seems to me we should be able to roll back the db to the last consistent state.

adamretter:
@shabanovd Naming is important so that people can understand the intention. Surely a quick find/replace on your code base could fix this?

adamretter:
@joewiz It is unlikely that this will do much by itself. However I am working on improving transactions in eXist, so hopefully we can get there in the medium-future

wolfgangmm:
@joewiz During crash recovery, eXist tries to restore the state of the database before the crash by redoing committed transactions and rolling back incomplete transactions. Adam's changes make it easier to avoid unclosed/forgotten transactions in the code and may thus help with future code.

However, it is unclear from your description what kind of corruption you experienced. Do you refer to a crash which resulted in a failing recovery? Or did your data get messed up by an update without actual crashing? Those are very different scenarios.

In any case, it would help if you could send me a more detailed report of what happened. And if you can even reproduce the issue, we may already be close to a fix ;-)

shabanovd:
@adamretter why you totally ignore agreement on API refactoring, why??? Normally, people to avoid others to have problems as much as possible do next: add new name, mark old name deprecated and keep it one release cycle as minimum. You do not like that gentleman style, do you?

Second, why you think that your naming much better? Few references: http://neo4j.com/api_docs/2.0.3/org/neo4j/graphdb/Transaction.html or https://www.firebase.com/docs/java-api/javadoc/com/firebase/client/Transaction.html but you may reference https://docs.oracle.com/cd/E17275_01/html/java/com/sleepycat/db/Transaction.html . But here one big difference come up: txn.success() do not perform 'end the transaction' action, it'll happen at 'close'. So, 'success' here because it just flag transaction state.

joewiz:
@adamretter That's great. Looking forward to hearing/discussing more.

@wolfgangmm I just emailed you a full description of what happened. But to summarize it here for the record, I ran an XQuery Update script that caused the db to become inconsistent. Concrete indications of trouble appeared in exist.log as:

    exampleElementImpl.java [getChildNodes]:918) - Internal error while reading child nodes: null javax.xml.stream.XMLStreamException

    EmbeddedXMLStreamReader.java [verifyOriginNodeId]:249) - expected node id 3.3.2.2.13.11.1.3, got 3.3.2.2.13.13.1.6.2.1; resyncing address

I've moved much of my XQuery Update work on individual XML files into oXygen with Saxon's support of the XQuery Update Facility, because I've had good luck with it (and can't corrupt the db).  In this case, though, I was planning a multi-file update, and so moved it back into eXist. 

But I suspect this is a bit off topic for the current thread, so I'll stop here now -- but I look forward to further discussion on improving transactions in eXist and reducing the likelihood that a (presumably) poorly crafted XQuery Update script could cause db corruption.

adamretter:
@shabanovd I searched the eXist code base and there were no uses of `txn.commit()`. `Txn` is part of he internal API, not the public API. Importantly, I have not changed the public API.
As with any project, if people choose to use the internal API then well they can expect breakage. Also as the direct `txn.commit` is a very recent addition, I rather assumed it was one of your unfinished works and that no one was actually using it.

Your references to uses of `success` in other projects don't apply here: In eXist, `success` (now called `commit`) in eXist DOES actually do the commit i.e. it flushes the db and makes the changes to the journal, this is very different to `success` in those other projects where the work is deferred until `close`.

If this really is a *big* problem for you, I would rather spend my energy on better things, so I can rename this method back *just* for you. Let me know...

dizzzz:
oops... if Txn is touched.... more extension will be potentially broken.....

adamretter:
@dizzzz The standard way in eXist to commit a transaction is `transactManager.commit(txn)`. That has not been changed. A very recent addition and seems to me, not yet widely used, was `txn.success()` which has been renamed to `txn.commit` to actually imply what it does and also to maintain consistent naming.

dizzzz:
ok, pfew, sorry for the fuzz :-)

shabanovd:
@adamretter , guess, you misunderstand two different layers here: eXist internal and low level API. Transaction interface is public one and can be used when project connect with eXist by low level API, so it is not eXist internal.

adamretter:
@shabanovd We disagree. But tell me, do you want me to change it back?

shabanovd:
@adamretter I did tell you that I use it, no?

adamretter:
@shabanovd I have reinstated and deprecated `success` and `failure` in org.exist.Transaction and `org.exist.storage.txn.Txn`. Hope that suits you?

shabanovd:
@adamretter yes, it do

adamretter:
So... can someone merge this then plskthnkbai

wolfgangmm:
I assigned it to myself yesterday and will merge it today or tomorrow once I had a chance to read through all the changes.

adamretter:
Cool, thanks. I have another transaction related one waiting after that :-)

PR #493, opened 2015-02-26
Register eXists URL Stream Handler just once

adamretter:
This ensures that the URL Stream Handler is just registered once within the JVM. In particular, this stops a whole bunch of junk log output when running the test suite.

### DIFF
--- a/conf.xml.tmpl
+++ b/conf.xml.tmpl
@@ -164,6 +164,12 @@
         -->
         <startup>
             <triggers>
+
+		<!--
+		    Trigger for registering eXists XML:DB URL handler with Java
+		-->
+		<trigger class="org.exist.protocolhandler.URLStreamHandlerStartupTrigger"/>
+
                 <!-- 
                     EXQuery RESTXQ trigger to load the RESTXQ Registry at startup time 
                 -->
--- a/src/org/exist/protocolhandler/URLStreamHandlerStartupTrigger.java
+++ b/src/org/exist/protocolhandler/URLStreamHandlerStartupTrigger.java
@@ -1,0 +1,75 @@
+/*
+ *  eXist Open Source Native XML Database
+ *  Copyright (C) 2001-2015 The eXist Project
+ *  http://exist-db.org
+ *
+ *  This program is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public License
+ *  as published by the Free Software Foundation; either version 2
+ *  of the License, or (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.protocolhandler;
+
+import org.apache.log4j.Logger;
+import org.exist.storage.DBBroker;
+import org.exist.storage.StartupTrigger;
+
+import java.net.URL;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Startup Trigger to register eXists URL Stream Handler
+ *
+ * @author Adam Retter <adam@exist-db.org>
+ * @author Dannes Wessels
+ */
+public class URLStreamHandlerStartupTrigger implements StartupTrigger {
+
+    private final static Logger LOG = Logger.getLogger(URLStreamHandlerStartupTrigger.class);
+
+    public final static String JAVA_PROTOCOL_HANDLER_PKGS="java.protocol.handler.pkgs";
+    public final static String EXIST_PROTOCOL_HANDLER="org.exist.protocolhandler.protocols";
+
+    @Override
+    public void execute(final DBBroker sysBroker, final Map<String, List<? extends Object>> params) {
+        registerStreamHandlerFactory();
+    }
+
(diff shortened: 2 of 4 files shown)


shabanovd:
will use of static flag be better here? Current refactoring have side effect by adding dependence on conf.xml

dizzzz:
hmmmm when does the URL hadler need to be available? later in the start
process? static could work, startup trigger looks actually elegant, but
maybe it is a bit to large or slow for the little function....

adamretter:
@shabanovd I don't understand your concern. I mean you can't run eXist without a `conf.xml` anyway

adamretter:
@dizzzz The StartupTrigger adds almost no overhead. It runs before the database becomes available so I think it happens at the right time.

shabanovd:
@adamretter I do run it without conf.xml, why you say that I can't?

shabanovd:
@adamretter why you can't hard code it right before StartupTrigger code? why you putting it into conf.xml? Just because you want to use StartupTrigger? what benefits?

adamretter:
@shabanovd You know what, I am trying to make eXist better. I did it this way because I believed it was a good way to do this. If you don't like it, well... then you can fix it.

shabanovd:
@adamretter if your change destroy someone environment you continue think that you making eXist better? All I do is highlighting problems that your change bring in to others. Because of your changers I'm already can't do merge of production code with latest develop branch, so production stay with summertime merge. Do you think that it make eXist better?

Not only you spend time and energy on making eXist better, so please be more patient to other problems and spend time and energy also to make everyone happy with your change, because otherwise deadend.

wolfgangmm:
One question: what happens if someone would remove the startup trigger for the URL stream handler from conf.xml? I ask because I often see people just copying an old conf.xml when updating to a newer version.

adamretter:
@wolfgangmm Well if that was the case then you would not have the startup triggers configured, but then I guess you know that?

adamretter:
@wolfgangmm What would you like me to do here, so we can get this in?

dizzzz:
I had some doubts wether this should be implemented with a startup trigger, but the old implementation was not really good practise. I think this is the reason of hesitation of pulling in this change.

As @shabanovd said, a simple static variable would have worked as well, but the the question is: where in the design does the initialisation fit well? it was originally bound to the validation features, but this is not correct (anymore).

If the unit tests run well, my and @wolfgangmm verdict is..... lets pull it in.

shabanovd:
so, production side is not interesting any more ... that is just another part that do not allow me to put last develop branch to production, because I don't want to spend time on conf.xml manual updates -(

dizzzz:
@shabanovd I feel sorry you feel this way. If it is realy a problem we could retract the PR. But please please please 'talk' with @adamretter and @wolfgangmm about this....... We had long thoughts about it already..

wolfgangmm:
@shabanovd since this pull request had been open for long, I asked @dizzzz (who wrote the stream handler) about the pros and cons. We were concerned this change will break compatibility, but as @dizzz told me, the worst case is that some xslts will fail because they are unable to find some resources. So the consequences should be minor and seem acceptable. If you see a major impact on your app, please send me a note.

In general, it would help if the comments on a pull request like this would be more concrete, e.g. "this cannot be accepted because it breaks feature X or use case Y". Reviewing a pull request costs a lot of time. If you want to reject a pull, please give a clear vote and make a positive suggestion on how the situation should be resolved.

shabanovd:
@wolfgangmm I did describe my problems in comments above:
```Current refactoring have side effect by adding dependence on conf.xml```

also I didn't get answer on my questions:
```why you can't hard code it right before StartupTrigger code? why you putting it into conf.xml? Just because you want to use StartupTrigger? what benefits?```

and very strange answer on your question `what happens if someone would remove the startup trigger for the URL stream handler from conf.xml?`:
```if that was the case then you would not have the startup triggers configured, but then I guess you know that?```

All this show the problem well and core developers ignore it very well too ;-)

Also, because almost everyone here, another problem that also well ignored is renaming methods and move classes without thinking of code outside eXist one. If class/method public that mean it can be and do use by others. You can make yourself feel ok and say that this is internal eXist code and nobody can/should use it, but that is not true. Some parts that do used a lot: database, dbbroker, collection, in memory/stored document (parts of it like element, attribute and etc), transactions.

I'm not against eXist improving process, but it must be some rules that everyone follow.

adamretter:
@shabanovd if you don't want to use a conf.xml then that is a design decision of your app and you are responsible for that and not us.

eXist ships with abd expects a conf.xml. This file changes as exist develops and moved forward. It will continue to do so.

I think you know this, so I don't really understand why you see a non-breaking change to conf.xml as a major issue.

shabanovd:
@adamretter I have no idea why you think that in production I do not use conf.xml, I didn't say that. In production each server have it owen conf.xml, so changers have to be done manually on each (there are 100s of them).

You didn't answer simple question: `what happens if someone would remove the startup trigger for the URL stream handler from conf.xml?` ... as I see it now, eXist will startup without give any signal that this core part of system is not initialize and that is major problem.

Also, as I said, before this change it was possible to start up eXist without conf.xml, after that is not true anymore, so there should exception if conf.xml can't be found as minimum.

dizzzz:
Wolfgang answered your question?

shabanovd:
what question?

dizzzz:
From your last message: "you didn't answer simple question"

shabanovd:
that is not my question, but @wolfgangmm one

adamretter:
@shabanovd I came to that conclusion because above you wrote:

"@adamretter I do run it without conf.xml, why you say that I can't?"

So from that I assumed you were running eXist somehow without a `conf.xml`. If that is not the case, then there is no problem. If you have hundreds of servers then I am sure you have some sort of configuration management and update system (e.g. Puppet/Chef/Ansible/Salt etc), so it's still not an issue IMHO.

shabanovd:
Yes, I do run eXist without conf.xml when do recovery, not in production.

PR #499, opened 2015-03-03
[bugfix] webdav interface broken by a24387d: saving resource failed. 

wolfgangmm:
Temp file needs to be closed last.


dizzzz:
Nice catch..... very subtile

shabanovd:
should it be closed manually at all? BufferedOutputStream will close parent stream.

PR #500, opened 2015-03-03
fix httpclient

gev:
Fix http client. Truncate and close stream

### DIFF
--- a/extensions/modules/src/org/exist/xquery/modules/httpclient/BaseHTTPClientFunction.java
+++ b/extensions/modules/src/org/exist/xquery/modules/httpclient/BaseHTTPClientFunction.java
@@ -380,11 +380,10 @@ public abstract class BaseHTTPClientFunction extends BasicFunction
                         final byte buf[] = new byte[4096];
                         int read = -1;
                         while((read = cfis.read(buf)) > -1) {
-                            baos.write(buf);
+                            baos.write(buf, 0, read);
                         }
-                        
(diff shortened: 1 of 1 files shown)


extensions/modules/src/org/exist/xquery/modules/httpclient/BaseHTTPClientFunction.java
```diff
@@ -380,11 +380,10 @@ public String getCacheClass() {
                         final byte buf[] = new byte[4096];
                         int read = -1;
                         while((read = cfis.read(buf)) > -1) {
-                            baos.write(buf);
+                            baos.write(buf, 0, read);
                         }
-                        
-                        builder.characters(URLEncoder.encode(EncodingUtil.getString(baos.toByteArray(), ((HttpMethodBase)method).getResponseCharSet()), "UTF-8"));
                         baos.close();
+                        builder.characters(URLEncoder.encode(EncodingUtil.getString(baos.toByteArray(), ((HttpMethodBase)method).getResponseCharSet()), "UTF-8"));
```
dizzzz: is this really required?

dizzzz:
Please could you describe what the fix is about?

extensions/modules/src/org/exist/xquery/modules/httpclient/BaseHTTPClientFunction.java
shabanovd: it was there before

shabanovd:
API misuse here, `baos.write(buf);` writing all bytes from array, but it may read just some. That's ehy  `baos.write(buf, 0, read);` must be used.

Issue #504, opened 2015-03-07
lang() function raises type error


jensopetersen:
This code should return the first desc element:

	xquery version "3.0";
	let $doc :=
	<desclist xml:lang="en">
		<desc xml:lang="en-US">
			<line>The first line of the description.</line>
		</desc>
		<desc xml:lang="fr">
			<line>La premi&#232;re ligne de la déscription.</line>
		</desc>
	</desclist>
	return
		$doc//desc[lang("en-US")]

It raises a type error (XPTY0004).

ljo:
Yes, this seems to be the same as a previous issue with fn:lang. I have started to fix it.

PR #506, opened 2015-03-09
Fixing wrong context item for fn:lang #504 by reverting changes by DS in...

ljo:
... commit c307fec3e5c397acefefc92286f221440dad354e.

### DIFF
--- a/src/org/exist/xquery/functions/fn/FunLang.java
+++ b/src/org/exist/xquery/functions/fn/FunLang.java
@@ -1,6 +1,6 @@
 /*
  * eXist Open Source Native XML Database
- * Copyright (C) 2001-2009 The eXist Project
+ * Copyright (C) 2001-2015 The eXist-db Project
  * http://exist-db.org
  *
  * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


shabanovd:
how many tests from XQTS going to fail?

ljo:
From what you wrote, eg working with attribute node in second parameter, it works as expected with this solution, as well as the expected cases like the #504 one. You did not reference any explicit xqts tests.

shabanovd:
it was a fix of ~10s xqts tests, and if you say that you get no degradation there than ok.

dizzzz:
general question: does the xqts tests still run? I guess we are waiting for the new suite, initiated by @adamretter ?

ljo:
Added a commit with 3 unit tests showing how this version passes all, both fn:lang#1, context version, and fn:lang#2, parameter version, with both attribute and element nodes.

dizzzz:
Great, thnx! 

For future tests you might want to consider writing an xquery testsuite ( http://exist-db.org/exist/apps/doc/xqsuite.xml#D1.2.7 ).

ljo:
Yes, we should think about how to proceed with double test systems (junit, xqsuite).

Issue #511, opened 2015-03-11
XQSuite: report syntax errors


dizzzz:
when developing XQuery-based tests, executed from the java testsuite, xquery syntax errors are not reported; there is no feedback about the problem, the .xq file is simply skipped/ignored.

if more .xq files are registered in a suite, there is no hint that some tests files are skipped; this way i missed that I had some serious regressions :-( 

If atleast one .xq file can be executed, the javatest passes.

Workaround: open test in eXide, and refactor.

How to reproduce: simply create a suite+one test xq file, and make a syntax error in the file

dizzzz:
right... in ```org.exist.xquery.functions.inspect.ModuleFunctions.eval(Sequence[], Sequence)`` an exception is converted into an XPathException, which is logged ad DEBUG and is muted.

PR #515, opened 2015-03-12
Module inspection: throw XPathException to detect issues

dizzzz:
When inspection of module fails (e.g. syntax error), throw exception to report issue to user;

https://github.com/eXist-db/exist/issues/511

in old situation issues are swallowed, not visible, as a result development time is lost and failing tests are missed.

### DIFF
--- a/src/org/exist/xquery/functions/inspect/ModuleFunctions.java
+++ b/src/org/exist/xquery/functions/inspect/ModuleFunctions.java
@@ -60,15 +60,20 @@ public class ModuleFunctions extends BasicFunction {
             Module module = null;
 
             try {
-            if (isCalledAs("module-functions-by-uri"))
-                {module = tempContext.importModule(args[0].getStringValue(), null, null);}
-            else
-                {module = tempContext.importModule(null, null, args[0].getStringValue());}
+                if (isCalledAs("module-functions-by-uri")) {
+                    module = tempContext.importModule(args[0].getStringValue(), null, null);
+                } else {
+                    module = tempContext.importModule(null, null, args[0].getStringValue());
+                }
             } catch (final Exception e) {
                 LOG.debug("Failed to import module: " + args[0].getStringValue() + ": " + e.getMessage(), e);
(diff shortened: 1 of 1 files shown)


dizzzz:
@wolfgangmm please could you review? are there consequences?

wolfgangmm:
I think we have to distinguish two cases here:

1. a module is not found
2. module is found but does not compile due to errors

For case 1 the idea was to return the empty sequence, because we need a way to dynamically determine if a certain module is available or not. For 2 an exception could be thrown. Would it be possible to distinguish those two cases, e.g. by looking at the exception?

dizzzz:
I searched the code, the strings are written in Xquery.g ; deep in antlr

Is it ok to verify for the string "Syntax error within user defined function" or should the Xquery.g be updated ?  The http://www.w3.org/TR/xpath-30/#ERRXPST0003 would the correct formal error no?

adamretter:
@wolfgangmm I think the `org.exist.util.function.Either` class in https://github.com/eXist-db/exist/pull/519 would really help you reflect the tri-state nature here. You could create an Either of `java.util.Optional` to help here. e.g.:

```java
Either<XPathException, Optional<Module>>
```

Where you have either:

1) the left is the error if the module does not compile
2) the right is optionally the value of the compiled module if the module is found or empty otherwise

adamretter:
@dizzzz Any progress on this?

dizzzz:
Yes and no ; I made some modifications here, adding a specific ErrorCode to the ANTLR code, and trying to get it out of it, but I do not see this information being propagated in the way I expected it should work.

I think I am sooooo close :-)

adamretter:
@dizzzz Did you consider using `Either`

dizzzz:
Did not consider anything yet because something fundamentally is not working for me....

dizzzz:
Ok I found the problem in the existing eXist classes ; I'll provide a solution tonight...

dizzzz:
Will file a new PR....

PR #518, opened 2015-03-16
Travis should fail the build if the tests fail

adamretter:


dizzzz:
hmmm does this do a quick-fail in stead of running the complete suite?

maybe it is better to do 

```
<junit errorproperty="bla"/>
```

and then to do 

```
<fail if="bla"/>
```
for each of the `junit` blocks after the unit report has been written.....

dizzzz:
```
<junit errorproperty="bla1"/>
<junit errorproperty="bla2"/>
<junit errorproperty="bla3"/>
<junitreport />
<fail if="bla1,bla2,bla3"/>
```

adamretter:
Travis is pretty simple. It does not understand JUnit or Ant. If it gets a non-zero exit code it implies failure and otherwise if it is a zero exit code then success. My change is fail-fast. Anything more complicated would require us to do more scripting to run all tests and then subsequently detect error/failure from JUnit and decide what exit code to use.

@dizzzz I don't think this can be done safely due to the structure of our Ant build system. That is to say that properties are not shared between `build/scripts/extension-common.xml` and `build/scripts/junit.xml` which means that we could not detect an overall success/failure across both the core and extensions.

PR #519, opened 2015-03-16
Cleaned up the XML:DB Local API implementation

adamretter:
This is a refactor of the server-side implementation of the XML:DB Local API. Through the use of Java 8 lambda expressions we the following advantages:

1) A huge reduction in code, including the removal of duplicated code.

2) Correct use of `Broker`/`Txn`/`Lock`. The lambda expressions ensure that when we take a resource we always return it.

3) Correct User security. The lambda expressions ensure that the correct cached XML:DB user credentials are applied to the correct operation.

4) Consistent exception handling. For example, we handle `PermissionDeniedException` in one place and that always leads to a `ErrorCodes.PERMISSION_DENIED` for the `XMLDBException`.

I plan to shortly also apply similar mechanisms to the XML:DB Remote API.

### DIFF
--- a/extensions/debuggee/test/org/exist/debugger/DebuggerTest.java
+++ b/extensions/debuggee/test/org/exist/debugger/DebuggerTest.java
@@ -478,7 +478,7 @@ public class DebuggerTest implements ResponseListener {
 	
     private void store(String name,  String data) throws EXistException {
     	Database pool = BrokerPool.getInstance();
-        final TransactionManager = transact = pool.getTransactionManager();
+        final TransactionManager transact = pool.getTransactionManager();
 
         try(final DBBroker broker = pool.get(pool.getSecurityManager().getSystemSubject());
             final Txn transaction = transact.beginTransaction()) {
--- a/extensions/indexes/lucene/test/src/org/exist/indexing/lucene/LuceneIndexTest.java
+++ b/extensions/indexes/lucene/test/src/org/exist/indexing/lucene/LuceneIndexTest.java
@@ -19,26 +19,28 @@
  */
 package org.exist.indexing.lucene;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.fail;
-
 import java.io.File;
+import java.io.IOException;
 import java.io.StringReader;
 import java.util.*;
+import javax.xml.XMLConstants;
+import javax.xml.parsers.ParserConfigurationException;
 
 import org.exist.EXistException;
 import org.exist.Indexer;
 import org.exist.TestUtils;
 import org.exist.collections.Collection;
 import org.exist.collections.CollectionConfigurationManager;
+import org.exist.collections.CollectionConfigurationException;
 import org.exist.collections.IndexInfo;
+import org.exist.collections.triggers.TriggerException;
+import org.exist.dom.QName;
 import org.exist.dom.persistent.DefaultDocumentSet;
 import org.exist.dom.persistent.DocumentSet;
 import org.exist.dom.persistent.MutableDocumentSet;
-import org.exist.dom.QName;
 import org.exist.indexing.OrderedValuesIndex;
 import org.exist.indexing.QNamedKeysIndex;
+import org.exist.security.PermissionDeniedException;
 import org.exist.security.xacml.AccessContext;
 import org.exist.storage.BrokerPool;
 import org.exist.storage.DBBroker;
@@ -48,6 +50,8 @@ import org.exist.storage.txn.Txn;
 import org.exist.test.TestConstants;
 import org.exist.util.Configuration;
 import org.exist.util.ConfigurationHelper;
+import org.exist.util.DatabaseConfigurationException;
+import org.exist.util.LockException;
 import org.exist.util.MimeTable;
 import org.exist.util.MimeType;
 import org.exist.util.Occurrences;
@@ -55,17 +59,22 @@ import org.exist.xmldb.XmldbURI;
 import org.exist.xquery.XQuery;
 import org.exist.xquery.XQueryContext;
 import org.exist.xquery.CompiledXQuery;
+import org.exist.xquery.XPathException;
 import org.exist.xquery.value.Sequence;
 import org.exist.xupdate.Modification;
 import org.exist.xupdate.XUpdateProcessor;
-import org.junit.After;
+
 import org.junit.AfterClass;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
-import org.xml.sax.InputSource;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 
-import javax.xml.XMLConstants;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
+
 
 public class LuceneIndexTest {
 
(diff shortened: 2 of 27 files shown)


dizzzz:
Is all related to the XMLDB improvements? Seem to be a bit more in it;
for the junit test, please could you add the following? You actually removed some out.println() ?

```java
@Rule
public TestRule watcher = new TestWatcher() {
   protected void starting(Description description) {
      System.out.println("Starting test: " + description.getMethodName());
   }
};
```

adamretter:
I am actually trying to eliminate 'System.out.println' from the tests as the console spew we currently have is pointless abd slows down the build

dizzzz:
it's a huge commit, we'll need some time to review....

adamretter:
The `General test code cleanup` commit (ac6082e) is just cosmetic improvements to tests that I made when I was chasing issues, otherwise the other two commits are directly related to the improvement of the XML:DB Local API implementation. This is well tested as a huge amount of the eXist test-suite uses the XML:DB Local API.

The changes are actually quite simple, as previously there was a huge amount of duplicated code in the Local XML:DB API implementation.

adamretter:
@dizzzz regards the Java lambda syntax: it is something that you get used to with practice. I have the advantage of having been doing this kinda thing in Scala for years before I had to get used to the Java syntax also ;-)

dizzzz:
@ lambda: you are right l I use the IDE to write them, but I choose not to do so for clarity sake

@ junit .... I tend not to agree here. I think printing out the name of the tests does not really flood your screen, it is the stackraces eand printout of results. Af far as I know, the TestWatcher is the way to do it, what you suggest in the Gist .... is bypassing the junit suite startup / cnvention completely?

adamretter:
@dizzzz TestWatcher is not a good solution for two main reasons:

1) You have to remember to copy-and-paste this into every test class. This causes code duplication, and undoubtedly will sometimes be forgotten.

2) You are forcing console output from every test. There is no option to enable/disable this.

Having a JUnit listener solves both of these problems, and it is what a JUnit listener was designed to do. TestWatcher was not designed for the purpose you are using it for because of (1) and (2).

shabanovd:
Agree that printing out test name is redundant coding and evaluation resources spending.

dizzzz:
@adamretter the PR cannot be merged anymore ; please could you revise?

adamretter:
@dizzzz Yup, I just rebase it for you. Should merge fine now.

adamretter:
Any progress on this from anyone?

I have also lined up the XML-RPC refactoring here https://github.com/adamretter/exist/commits/xmlrpc. That is waiting, ready for me to open a pull-request but I need the `xmldb-local` pull-request to land-first as there are some dependencies. Shortly I will also have further pull requests for `xmldb-remote`, the `rest server` and `webdav`.

dizzzz:
A few ppl are studying the code, it requires some time. There are a few concerns though;

- there have no issues been reported about the xml:db classes for a long time, considering them stable; with making (significant) changes there is a chance/risk new issues are introduced.

- lambdas are very modern, but they make code rather difficult to read; because of this we might add extra complexity to maintenance  and debugging activities, probably only fewer ppl who can understand and do it. We should not use lambda expression because we want to use lambda expressions. It needs to have good considerations.

- please don't shoot me, i am just the messenger.

caseydawsonjordan:
+1 for lambdas as they make creating a broker leak or lock leak nearly
impossible if used correctly. Broker and lock leaks are something that have
been a notorious problem with eXist and are very hard to detect without
very detailed code review. The way Adam is using lambda here is a really
smart way to avoid them and about the best example of where you should use
lambdas.

On Tue, Mar 24, 2015 at 2:44 PM, Dannes Wessels <notifications@github.com>
wrote:

> A few ppl are studying the code, it requires some time. There are a few
> concerns though;
>
>    -
>
>    there have no issues been reported about the xml:db classes for a long
>    time, considering them stable; with making (significant) changes there is a
>    chance/risk new issues are introduced.
>    -
>
>    lambdas are very modern, but they make code rather difficult to read;
>    because of this we might add extra complexity to maintenance and debugging
>    activities, probably only fewer ppl who can understand and do it. We should
>    not use lambda expression because we want to use lambda expressions. It
>    needs to have good considerations.
>    -
>
>    please don't shoot me, i am just the messenger.
>
> --
> <https://github.com/eXist-db/exist/pull/519#issuecomment-85638495>.
>



-- 
--
Casey Jordan
easyDITA a product of Jorsek LLC
"CaseyDJordan" on LinkedIn, Twitter & Facebook
(585) 348 7399
easydita.com


This message is intended only for the use of the Addressee(s) and may
contain information that is privileged, confidential, and/or exempt from
disclosure under applicable law.  If you are not the intended recipient,
please be advised that any disclosure  copying, distribution, or use of
the information contained herein is prohibited.  If you have received
this communication in error, please destroy all copies of the message,
whether in electronic or hard copy format, as well as attachments, and
immediately contact the sender by replying to this e-mail or by phone.
Thank you.

dizzzz:
@caseydawsonjordan ok, thnx for the contribution; it helps!

adamretter:
@dizzzz The code in the XMLDB classes was pretty hideous. This is a cleanup of the Local XMLDB part of that.

Apart from reducing the amount of code that needs to be understood/maintained by 30%, because of how I have used the Lambdas, they also bring safety when acquiring/releasing resources as the code to do that only has to be written once and ensured that it is correct once.

Sure, lambdas are a new feature of Java 8, but they are not going away. I don't think they are harder to read, but that is probably because I have a lot of functional programming experience. We are all Java programmers in eXist, and as lambdas are more widely used, for those not familiar with them yet, they will become transparent. Also lambdas are required when working with Java 8's Streams API (which can also assist with basic parallel programming), so I think people will get used to them very quickly ;-)

JoernT:
I'm not familiar with lambdas - however it's not a good argument to not use improved language features cause people are not (yet) familiar with them ("i don't know it so it's no good"). If that argument would hold people would still not use generics. It's our fate as programmers that we have to keep on learning.

Code reduction is a good thing - as it reduces the probability for errors and if it even makes execution more safe then this is a plus.

If there's still uncertainty about possible regressions then probably test coverage needs to be improved.

For a refactoring two factors count: all tests pass and the exposed API does not change.

adamretter:
@JoernT There are many test cases for this code as much of the Test Suite actually uses this API to talk to eXist. Also as far as I am aware I have not changed the API; In most part the API is defined in the xmldb.jar (`org.xmldb`) which is external to eXist, also I would hope the test suite would show a failure if I was to accidentally change the API.

shabanovd:
+1 for lambdas

wolfgangmm:
I was running some tests against large applications to see if there would be any negative effects on performance or memory consumption. I have not experienced any problems. Leif reported some issues on Monday though, so I'm waiting for his feedback. It is much easier to thoroughly test a pull request before it is accepted.

adamretter:
@wolfgangmm thanks that's great.

@ljo please can you share your findings?

ljo:
I will get back to the testing today, yesterday was unfortunately all about firefighting.

adamretter:
@ljo cool, thanks

ljo:
I would say it is a go since the issues I experienced earlier all seems to be due to external factors and/or previous state of the db.

ljo:
Lambdas is the way to go, but there were only some concerns for this part of the codebase.

wolfgangmm:
@dizzzz @ljo please merge if you have time. I couldn't do it yet due to travel and slow internet.

dizzzz:
Happy to merge it :-)

Issue #509, opened 2015-03-10
Can't build an installer from working instance


amclark42:
Briefly: I cannot build an installer from a working installation of eXist without mucking around in the install code.

The long version:

Running eXist 2.2 (stable), I wanted to build a custom installer for local development work. To do so, I ran `./build.sh installer` from my working instance, then secure-copied the resulting JAR to my local Vagrant box. Installing on the VM worked fine, except that all the scripts in `$EXIST_HOME/bin/scripts` used the paths from the first installation - the one on my machine proper:
![screen shot 2015-03-10 at 1 50 48 pm](https://cloud.githubusercontent.com/assets/4063180/6581349/a56f4fe4-c72c-11e4-9570-8b56484dcfee.png)
![screen shot 2015-03-10 at 3 40 04 pm](https://cloud.githubusercontent.com/assets/4063180/6583647/d16247be-c73b-11e4-817d-1081c9932d8e.png)
I confirmed that I could change the paths manually/programmatically to get the new installation to work. This became plan B, since I much preferred to have the custom installer replace the paths and set the admin password.

When you run ./build.sh installer, izPack takes all the `*.sh` scripts in `$EXIST_HOME/bin` and copies them into `$EXIST_HOME/installer/scripts`, overwriting older files as necessary. This system is meant to work for people who want to build an installer from the source. Since my copy of eXistDB came prepackaged, the installer changed all `$EXIST_HOME/bin` scripts' references to `$JAVA_HOME` and `$EXIST_HOME` so they reflected the paths in my personal environment. And then, when I tried to build a new installer, those scripts (with those paths!) were copied over. Not great for sharing or re-use!

The overwriting scripts behavior is dictated by `$EXIST_HOME/build/scripts/installer.xml`, lines 197-209. I propose treating all the shell scripts like build.sh, [copying permissive scripts from `$EXIST_HOME/installer/scripts` into `$EXIST_HOME/bin`](https://github.com/eXist-db/exist/blob/develop/installer/scripts/build.sh), instead of the other way around.

![The main part of the installer fix](https://cloud.githubusercontent.com/assets/4063180/6584333/e1edbd66-c740-11e4-93a0-25ea425dfce5.png)

[...]

PR #520, opened 2015-03-16
Feature/flexible installer

amclark42:
Remove the step of copying the scripts from `bin` to `installer/scripts` in favor of keeping the scripts in the latter. This way, the installer always has the original copy of the scripts. This does not affect those building an installer from source, and allows people to build further, derivative installers that do not contain local filepaths.

To test:
- build an installer from source with `./build.sh installer`;
- install eXist-TEST1 in a separate location;
- build an installer from eXist-TEST1;
- install eXist-TEST2 in a separate location; and
- ensure that eXist-TEST2 starts up with `./bin/startup.sh`.


amclark42:
Resolves #509

joewiz:
When you said "in the latter" (`installer/scripts`), did you mean to say "in the former" (`bin`)?

adamretter:
@amclark42 I am not sure that moving the scripts from `bin` to `installer/scripts` is a good idea as it breaks the ability to use eXist for those who clone directly from Git or obtain a tarball of the source code. The installer is just one mechanism for obtaining/installing eXist.

I think we need to determine another mechanism for fixing your issue. The shell scripts should remain in `installer/scripts`. I will try and work though your issue described here https://github.com/eXist-db/exist/issues/509 so that I can understand what is happening?

Am I right in thinking your issue boils down to creating an installer from a copy of eXist that has been installed using the installer?

amclark42:
@adamretter That makes sense.

Yes, your summary of the issue is correct. I wanted the installer to use fresh copies of the scripts. Currently, if a derivative installer is built, the new installer will contain the JAVA_HOME-ish and EXIST_HOME-ish paths from the original installation.

Of course, as you noted in exist-open, one can work around this by setting those variables oneself. But I still do feel that the installer should use fresh scripts rather than ones altered for working use!

adamretter:
@duncdrum not just yet

adamretter:
Closed by https://github.com/eXist-db/exist/pull/2657

@amclark42 the installer scripts are now generated from templates

Issue #513, opened 2015-03-12
XQuery set user password from digest


pboz:
As an administrator, I need an XQuery function that lets me set a user's password from a digest password. If I am managing users between instances of eXist, I can only access an existing users digest password, I (correctly) do not have access to their original password.

PR #523, opened 2015-03-17
Add XQuery function for setting user passwords from an encodedHash

adamretter:
Closes https://github.com/eXist-db/exist/issues/513


shabanovd:
I can't say that happy the way it go, but going to merge it

PR #531, opened 2015-03-19
[ignore] Remove some outdated xquery stuff

wolfgangmm:
which currently breaks the function docs.


adamretter:
Hmm, is this actually outdated? I mean these samples are useful for people right? Can they not be updated rather than removed?

wolfgangmm:
Nearly all of them use outdated syntax or are duplicates of examples contained in the demo app. One or two might be moved there.

ljo:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

Den 2015-03-19 11:17, Wolfgang Meier skrev:
> Nearly all of them use outdated syntax or are duplicates of examples
> contained in the demo app. One or two might be moved there.
> 
Yes, we have tried to migrate all the samples to demo app or
documentation. So should be quite safe.

-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iD8DBQFVCqKrhcIn5aVXOPIRAiiyAJ47j87vMneE+bdiMeMTgLfWMU8EQwCfa+lt
dBV5Ns8Tp4e3hQDJBuvrO8Y=
=XYM9
-----END PGP SIGNATURE-----

PR #536, opened 2015-03-19
Tests output cleanup

adamretter:
This pull-request addresses the output issues of the test-suite and eXist onto stdout. Working in co-operation with log4j2, all log output (including SLF4j and JUL) is correctly captured to `$EXIST_HOME/webapp/WEB-INF/tests` even when running the test suite. Main features include:

* Reducing calls to `System.out.println`. Replaced with LOG statements where appropriate or otherwise removed.

* Redirecting JUL output from code running under Ant's JUnit task

* Dynamically generating a log4j2 configuration that is suitable for testing

* Refactored the output of the XQuery/XML tests when run from JUnit to improve the visibility of test results.

* Also... allows the test-suite to run correctly in Travis-CI.

As a result the output of the the test suite has been reduced from 22MB to >200KB, instead all significant output from eXist is correctly captured in its log files. There will be a follow-up pull-request to allow access to the eXist log files from Travis-CI when the test suite reports a failure.

### DIFF
--- a/build/scripts/extensions-common.xml
+++ b/build/scripts/extensions-common.xml
@@ -312,6 +312,8 @@ to add: -pre-compile steps
       <junit haltonfailure="${junit.haltonfailure}" haltonerror="${junit.haltonerror}" printsummary="yes" showoutput="${junit.output}"
          maxmemory="${junit.forked.VM.maxmemory}">
 
+        <sysproperty key="java.util.logging.manager" value="org.apache.logging.log4j.jul.LogManager"/>
+
         <sysproperty key="jetty.port" value="${jetty.port}" />
         <sysproperty key="jetty.port.ssl" value="${jetty.port.ssl}" />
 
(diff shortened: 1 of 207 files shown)


dizzzz:
Now I see what you mean; this definately cleans the logging! nice....

q: are nasty traces e.g. NPEs swallowed or are these still visible?

dizzzz:
unfortunately travis does not finish correct yet ; too much logging due to too many stacktraces ; is there a corruption?

adamretter:
@dizzzz failures and exceptions will still show up in the output as those are managed by JUnit and the Ant JUnit task will print these to the console.

Regards travis there does seem to be a test which is causing a database corruption. I need some time to look at that in more detail. It occurs on Travis (which is Ubuntu) but I could not reproduce it on Linux Mint (which is also kinda Ubuntu). I will access the Travis VM again and see what happens...

dizzzz:
ok, thnx...

PR #542, opened 2015-03-27
add .odd as an XML file extension

joewiz:
[ODD](http://en.wikipedia.org/wiki/Text_Encoding_Initiative#ODD) is an XML-based vocabulary for defining schemas for Text Encoding Initiative (TEI) projects.  ODD should be treated as XML syntax but until this pull request was treated as a binary format.


joewiz:
Thanks!

PR #543, opened 2015-03-27
Cleaned up the XML:DB Remote API implementation

adamretter:
General code audit and cleanup. Fixed several small bugs along the way that were the result of previous copy-pasta mistakes or bad code style.

### DIFF
--- a/src/org/exist/util/Occurrences.java
+++ b/src/org/exist/util/Occurrences.java
@@ -32,9 +32,14 @@ public class Occurrences implements Comparable<Occurrences> {
 	private int occurrences = 0;
 	private MutableDocumentSet docs = new DefaultDocumentSet();
     
-	public Occurrences(Comparable name) {
-		term = name;
+	public Occurrences(final Comparable name) {
+		this.term = name;
 	}
+
+    public Occurrences(final Comparable name, final int occurrences) {
+        term = name;
+        this.occurrences = occurrences;
+    }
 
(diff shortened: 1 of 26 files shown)


dizzzz:
The PR looks pretty straightforward to me, ok for me to pull in.

2 questions for adam (not necessarily a request for a change!)
- on a few locations you write `final List params = new ArrayList()` as where `List<Object>` was used first; is that actually a better choice? I remember this solved some checkstyle/pdm/findbugs reports
- one one location an input stream is copied to an output stream (buffer of 64kb?); for readability I tend to use `IOUtils.copy()` for this.  This additional dependancy is probably not too bad.

adamretter:
@dizzzz Ah right, yes, good catch! It should be `List<Object>` and not the parameterised `List`. Let me just fix that.

adamretter:
@dizzzz Okay I have done that and rebased. It is ready to go...

dizzzz:
Thnx ; I'd ike to discuss with @wolfgangmm @shabanovd  @ljo to have a look as wel; e.g. for other PRs @wolfgangmm is testing the PR ; same for this one?

wolfgangmm:
Tested this for a day without finding an issue.

adamretter:
Yay :-)

dizzzz:
woohoo!

ljo:
:)

PR #544, opened 2015-03-27
Cleaned up the XML-RPC API implementation

adamretter:
This is a refactor of the XML-RPC API. Through the use of Java 8 lambda expressions we the following advantages:

1) A huge reduction in code, including the removal of duplicated code.

2) Correct use of Broker/Txn/Lock. The lambda expressions ensure that when we take a resource we always return it.

3) Correct User security.

4) Consistent exception handling.

### DIFF
--- a/src/org/exist/util/VirtualTempFile.java
+++ b/src/org/exist/util/VirtualTempFile.java
@@ -42,7 +42,12 @@ import org.apache.commons.io.output.ByteArrayOutputStream;;
  * 
  * @author jmfernandez
  *
+ * @deprecated Using this class should be avoided as it publishes an API that
+ * makes using it correctly without leaking resources very difficult. It is very
+ * likely that most uses of this class do not correctly cleanup the resources
+ * they obtain. It needs to be rewritten...
  */
+@Deprecated
 public class VirtualTempFile
 	extends OutputStream
 {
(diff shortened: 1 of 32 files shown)


wolfgangmm:
I'll run some tests on my large test db as I did for the local implementation (batch uploads, backup/restore ...)

adamretter:
Cool, thanks @wolfgangmm

src/org/exist/util/VirtualTempFile.java
```diff
@@ -43,6 +43,7 @@
  * @author jmfernandez
  *
  */
+@Deprecated
 public class VirtualTempFile
```
dizzzz: Marked deprecated .... It is used on quite some locations, please could you add a reference to its replacement? (or just mention it here)

dizzzz:
PR looks good to me

src/org/exist/util/VirtualTempFile.java
adamretter: I just added some comments on that in the javadoc and rebased it for you. Hopefully we are all good now.

wolfgangmm:
I tested this one as well, so it could be merged once the conflicts were resolved.

adamretter:
@dizzzz that was quick! I just pushed the rebased a few seconds ago.

dizzzz:
@adamretter sorry :-) I was just hopping by , accidentely. At your service sir :-)

PR #545, opened 2015-03-28
indexer parameter has wrong name in documentation

jensopetersen:
suppress-whitespace-mixed-content => preserve-whitespace-mixed-content

I think preserve-whitespace-mixed-content should be set to "yes" by default, to preserve meaningful whitespace in docbook, tei, etc. Where does whitespace in mixed-content get in the way in other formats?

### DIFF
--- a/conf.xml.tmpl
+++ b/conf.xml.tmpl
@@ -406,8 +406,8 @@
             Changing the parameter will only have an effect on newly loaded 
             files, not old ones.
         
-        - suppress-whitespace-mixed-content:
-            preserve the white space inside a mixed content node
+        - preserve-whitespace-mixed-content:
+            preserve the white space inside a mixed content node: "yes" or "no".
         
         - tokenizer:
             this attribute invokes the Java class used to tokenize a string into


shabanovd:
+1 to set it to `yes` by default. It'll cost some additional space, but "what get in - get out" will be much close after that step.

adamretter:
Will this perhaps surprise some people, by no longer matching their query strings?

shabanovd:
@adamretter do you have example or just think so?

adamretter:
@shabanovd Just a thought!

joewiz:
This seems like an easy accept, no?

wolfgangmm:
There's still a whitespace handling problem introduced by #670 (see there). For certain types of documents I had to partially roll back those changes because eXist was eating non-ignorable whitespace. We have to get this fixed first, then see if we can get rid of the preserve-whitespace-mixed-content setting.

adamretter:
I merged this in manually as it needed rebasing. Thanks @jensopetersen

dizzzz:
sorry :-) I planned to do this but was a  bit too slow :-)

ljo:
OK. Great. Thanks.

PR #546, opened 2015-03-28
updating info on legacy full text index

jensopetersen:
existing info is misleading

### DIFF
--- a/conf.xml.tmpl
+++ b/conf.xml.tmpl
@@ -443,10 +443,7 @@
             <module id="range-index"    class="org.exist.indexing.range.RangeIndex"/>
 
             <!-- 
-                The full text index is always required and should
-                not be disabled. We still have some dependencies on
-                this index in the database core. These will be removed
-                once the redesign has been completed. 
+                The legacy full text index. Replaced by the Lucene index.
             -->
             <!--module id="ft-legacy-index" class="org.exist.fulltext.FTIndex"/-->
 


joewiz:
The previous comment mentioned that there were still dependencies on the legacy full text index, and that the plan was to remove these so the old full text index could be removed. @wolfgangmm: is this already done? If so, can we remove this comment/entry altogether?

adamretter:
@wolfgangmm I would be very happy to delete the old legacy full-text index... just say the word and the pull-request will be on the way.

dizzzz:
Didnt we do this exercise on the 1.4 branch a long time ago? I recall it will be a tough job, deps are on many locations in the code.

Removal.... Is certainly a good idea but I fear for regression. To start with we should remove the &= like expressions in the grammar files? Then users can not use it at all anymore.

adamretter:
@dizzzz I don't mind a tough job! Just let me know if I can go for it...

dizzzz:
@adamretter I would certainly not mind, maybe I can help a bit. But... since there is a risk of instability in the core for some time, we need to be open on this improvement. @ljo @wolfgangmm  @shabanovd  ?

jensopetersen:
Sure, it has to go eventually, but it doesn't do any harm leaving it as it is until after next release – there are quite a number of actual bugs waiting to be addressed.

shabanovd:
I had remove old full text index for long time in my branch and it used in production without any issue around it, so I'm happy to help if/when agree to do so.

shabanovd:
I recall that @wolfgangmm mention some sites that still depend on it.

adamretter:
@jensopetersen Of course, but removing code often leads to less bugs to maintain ;-)

@shabanovd Let's wait and see what @wolfgangmm says then...

Issue #147, opened 2014-02-12
"boundary-space preserve" declaration not observed


jensopetersen:
With a "boundary-space preserve" declaration, white-space between elements is supposed to be maintained. However, when I feed the function below with the exact same contents from an in-memory variable and from a document stored in the database, the result of the in-memory fragment is returned with white-space preserved, whereas the result of the stored document is returned as if the declaration had told the function to strip white-space, if the text node is white-space-only.
# 

```
xquery version "3.0";

declare namespace tei="http://www.tei-c.org/ns/1.0";

declare boundary-space preserve;

declare function local:get-text($input as node()*) as item()* {
   for $node in $input/node()
       return
           typeswitch($node)
               case text() return 
                   $node
                       default return local:get-text($node)
};

let $in-memory := 
<TEI xmlns="http://www.tei-c.org/ns/1.0">
<p>Government of new Territory of Nevada—Governor <name>Nye</name> <lb/>and the practical jokers—<name>Mr. Clemens</name> begins journalistic life <lb/>on <name>Virginia City</name> <name>Enterprise</name>.</p>
</TEI>

let $stored := doc('/db/test/stored.xml')

return
   <result>
       <inline>{string-join(local:get-text($in-memory))}</inline>
       <stored>{string-join(local:get-text($stored))}</stored>
   </result>
```
# 

The result I get is:

```
<result>
    <inline>Government of new Territory of Nevada—Governor Nye and the practical jokers—Mr. Clemens begins journalistic life on Virginia City Enterprise. </inline>
    <stored>Government of new Territory of Nevada—Governor Nyeand the practical jokers—Mr. Clemens begins journalistic life on Virginia CityEnterprise.</stored>
</result>
```

where "Nye" and "and" and "City" and "Enterprise" are run together in the "stored" output, because the space between <name> and <lb> and between <name> and <name> is stripped.

Setting preserve-whitespace-mixed-content="yes" in conf.xml works as a workaround, but is not ideal.

duncdrum:
According to the discussion in the linked PR this is a nofix for now. @jlo how about that documentation any progress since 01/25/2016?

joewiz:
@ljo ^

joewiz:
I performed the original test as specified in the OP using the current `develop-4.x.x`, with the following result:

```xml
<result>
    <inline> Government of new Territory of Nevada—Governor Nye and the practical jokers—Mr. Clemens
        begins journalistic life on Virginia City Enterprise. </inline>
[...]

PR #548, opened 2015-03-30
set preserve-whitespace-mixed-content to "yes" by default

jensopetersen:
preserve-whitespace-mixed-content should be set to "yes" by default, to preserve meaningful whitespace in DocBook, TEI, etc. It does not seem likely that leaving whitespace in mixed-content will interfere with existing queries.

https://github.com/eXist-db/exist/pull/545
https://github.com/eXist-db/exist/issues/147

### DIFF
--- a/conf.xml.tmpl
+++ b/conf.xml.tmpl
@@ -417,7 +417,7 @@
         - track-term-freq:
             not documented.
     -->
-    <indexer caseSensitive="yes" index-depth="5" preserve-whitespace-mixed-content="no"
+    <indexer caseSensitive="yes" index-depth="5" preserve-whitespace-mixed-content="yes"
         stemming="no" suppress-whitespace="none"
         tokenizer="org.exist.storage.analysis.SimpleTokenizer" track-term-freq="yes">
 


ljo:
We need to check everything using JSON serialisation though, since the consumers might not expect all the new whitespace content served. And i certainly know front-end devs are not happy about it.

jensopetersen:
Is the problem that this function doesn't really check whether there is a text node sibling which has non-whitespace contents? I had expected that 

	<p><w>bus</w> <w>conductor</w></p>

was returned without whitespace, but it was, just as

	<p>A <w>bus</w> <w>conductor</w></p>

(which should).

wolfgangmm:
@jensopetersen as you noted above, the real issue seems to be that the configuration property is not working as intended:

```xquery
<p>A <w>bus</w> <w>conductor</w></p>
```

should preserve the space because there's a text node sibling. I'd say this is the bug which should be fixed.

Setting preserve-whitespace-mixed-content="yes" has some undesirable effects if you work with e.g. oXygen (or other xml editors) over webdav. The formatting used to become rather messy when I tried it.

dizzzz:
@ljo is working on a structural solution

adamretter:
@jensopetersen was this fixed by @ljo with his PR https://github.com/eXist-db/exist/pull/975

jensopetersen:
I have to wait a week or so before I can check, being at sea.

dizzzz:
feedback from projects is very positive!

adamretter:
@jensopetersen "being at sea." That sounds very exciting :-)

adamretter:
@jensopetersen Any updates for us?

adamretter:
@ljo what is the status of this please?

adamretter:
@ljo Again, what is the status of this please?

adamretter:
@ljo Please please can you provide an update...

adamretter:
@ljo I could still use an update on this please?

adamretter:
@wolfgangmm I cannot raise @ljo about this. @joewiz suggested you might have some feedback due to your careful use of white-space in TEI... Any thoughts?

adamretter:
After some discussion we feel that changing the default setting here is not a good idea, however we are going to document the whitespace situation - https://github.com/eXist-db/documentation/issues/128

PR #551, opened 2015-03-31
Feature/module import must throw exception

dizzzz:
Finally i found the issue: a missing line in a XPathException constructor; now a specific ErrorCode can be used to detect the situation.

### DIFF
--- a/src/org/exist/xquery/XPathException.java
+++ b/src/org/exist/xquery/XPathException.java
@@ -383,10 +383,10 @@ public class XPathException extends Exception {
 
     public static class FunctionStackElement {
 
-        String function;
-        String file;
-        int line;
-        int column;
+        private final String function;
+        private final String file;
+        private final int line;
+        private final int column;
 
         public FunctionStackElement(UserDefinedFunction func, String file, int line, int column) {
             this.function = func.toString();
--- a/src/org/exist/xquery/XPathException.java
+++ b/src/org/exist/xquery/XPathException.java
@@ -1,14 +1,35 @@
+/*
+ *  eXist Open Source Native XML Database
+ *  Copyright (C) 2001-2015 The eXist Project
+ *  http://exist-db.org
+ *
+ *  This program is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public License
+ *  as published by the Free Software Foundation; either version 2
+ *  of the License, or (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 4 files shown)


src/org/exist/xquery/XPathException.java
```diff
@@ -322,8 +362,39 @@ public String getMessageAsHTML() {
...
+     * 
+     * @return Error value as sequence
+     */
+    public Sequence getErrorVal() {
+        return errorVal;
+    }
+
     public static class FunctionStackElement {
 
+        String function;
```
adamretter: These should be final I think?

src/org/exist/xquery/XPathException.java
```diff
@@ -322,8 +362,39 @@ public String getMessageAsHTML() {
...
+        String file;
+        int line;
+        int column;
+
+        public FunctionStackElement(UserDefinedFunction func, String file, int line, int column) {
```
adamretter: Likewise, I think these should be final...

src/org/exist/xquery/functions/inspect/ModuleFunctions.java
```diff
@@ -60,15 +60,26 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
...
+                if (isCalledAs("module-functions-by-uri")) {
+                    module = tempContext.importModule(args[0].getStringValue(), null, null);
+                } else {
+                    module = tempContext.importModule(null, null, args[0].getStringValue());
+                }
+                
+            } catch (final XPathException e) {
+                LOG.debug("Failed to import module: " + args[0].getStringValue() + ": " + e.getMessage(), e);
+
+                if (e.getErrorCode().equals(ErrorCodes.XPST0003)) {
```
adamretter: I don't think XPST0003 is the correct error message to use here? It is for static error, but this is at runtime so it must by a dynamic error...

adamretter: Also, why would you hide all other XPathExceptions from the user and send them only to the log?

dizzzz: about also: that was the current design; all was swallowed, agreed with wolfgang to do so.....

dizzzz: about XPST0003: we hava a syntactic error in a module, this code looked best matching to me

src/org/exist/xquery/XPathException.java
dizzzz: hmmm yes, my IDE did not show it though :-(

src/org/exist/xquery/XPathException.java
```diff
@@ -322,8 +362,39 @@ public String getMessageAsHTML() {
...
+     * 
+     * @return Error value as sequence
+     */
+    public Sequence getErrorVal() {
+        return errorVal;
+    }
+
     public static class FunctionStackElement {
 
+        private final String function;
```
dizzzz: made final

adamretter:
Okay cool, let me run it through the test suite here and then I am happy.

dizzzz:
:)

adamretter:
All passes here :-)

PR #563, opened 2015-04-03
Decouple XQuery service from DBBroker

adamretter:
This completes some work started and suggested by @shabanovd in 9f1c0c8.

1) Decouples XQuery service from `DBBroker`.

2) Moves `getXQueryService()` from `DBBroker` to `BrokerPool`.

### DIFF
--- a/extensions/debuggee/src/org/exist/debuggee/DebuggeeImpl.java
+++ b/extensions/debuggee/src/org/exist/debuggee/DebuggeeImpl.java
@@ -133,14 +133,14 @@ public class DebuggeeImpl implements Debuggee {
 	
 	        XQuery xquery = broker.getXQueryService();
 			
-	        XQueryContext queryContext = xquery.newContext(AccessContext.REST);
+	        XQueryContext queryContext = new XQueryContext(broker.getBrokerPool(), AccessContext.REST);
 	
 			// Find correct script load path
 			queryContext.setModuleLoadPath(XmldbURI.create(uri).removeLastSegment().toString());
 	
 			CompiledXQuery compiled;
 			try {
-				compiled = xquery.compile(queryContext, source);
+				compiled = xquery.compile(broker, queryContext, source);
 			} catch (IOException e) {
 				e.printStackTrace();
 				return null;
--- a/extensions/debuggee/src/org/exist/debuggee/DebuggeeImpl.java
+++ b/extensions/debuggee/src/org/exist/debuggee/DebuggeeImpl.java
@@ -131,7 +131,7 @@ public class DebuggeeImpl implements Debuggee {
 	
 	        if (source == null) return null;
 	
-	        XQuery xquery = broker.getXQueryService();
+	        XQuery xquery = broker.getBrokerPool().getXQueryService();
 			
 	        XQueryContext queryContext = new XQueryContext(broker.getBrokerPool(), AccessContext.REST);
 	
--- a/extensions/debuggee/src/org/exist/debuggee/DebuggeeJointImpl.java
+++ b/extensions/debuggee/src/org/exist/debuggee/DebuggeeJointImpl.java
@@ -470,9 +470,9 @@ public class DebuggeeJointImpl implements DebuggeeJoint, Status {
 			context.undeclareGlobalVariable(Debuggee.IDEKEY);
 			
 			XQuery service = broker.getXQueryService();
-			CompiledXQuery compiled = service.compile(context, script);
+			CompiledXQuery compiled = service.compile(broker, context, script);
 			
-			Sequence resultSequence = service.execute(compiled, null);
+			Sequence resultSequence = service.execute(broker, compiled, null);
 	
 	        SAXSerializer sax = null;
 	        Serializer serializer = broker.getSerializer();
--- a/extensions/debuggee/src/org/exist/debuggee/DebuggeeJointImpl.java
+++ b/extensions/debuggee/src/org/exist/debuggee/DebuggeeJointImpl.java
@@ -469,7 +469,7 @@ public class DebuggeeJointImpl implements DebuggeeJoint, Status {
 			context.undeclareGlobalVariable(Debuggee.SESSION);
 			context.undeclareGlobalVariable(Debuggee.IDEKEY);
 			
-			XQuery service = broker.getXQueryService();
+			XQuery service = broker.getBrokerPool().getXQueryService();
 			CompiledXQuery compiled = service.compile(broker, context, script);
 			
 			Sequence resultSequence = service.execute(broker, compiled, null);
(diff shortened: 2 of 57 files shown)


dizzzz:
Wondering..... 

1. what are the side effects of moving the `getXQueryService()` thingy from Broker to BrokerPool from a 'execute permissions' perspective? I ask because in the step from brokerpool to brok a 'subject' might be involved

2. is with this change still possible to have multiple database instances defined? (more than just 'exist' ) ?

shabanovd:
I can't understand what problem you are solving? My one solve concurrent issue when broker was shared between thread because it was stored in context object's field. 

I don't like change for change! I prefer change that solve issue.

shabanovd:
what I see:
*  `xquery.newContext(accessContext)` => `new XQueryContext(broker.getBrokerPool(), accessContext)`
* `xquery.compile(context, query)` => `xquery.compile(broker, context, query)`
* `broker.getXQueryService()` => `getPool().getXQueryPool()`

-1 to merge this PR, because it's API change without any problem solving

adamretter:
@dizzzz 
1) Their are no side-effects because the `XQuery` class is stateless.
2) Yes that is still as it was.

adamretter:
@shabanovd Okay, so I guess I have not communicated enough with you so that you understand the issue I am trying to solve. Sorry about that. I discussed it with @wolfgangmm but failed to communicate to the other developers. As suggested in the recent teleconference, I have create a ticket for what I am trying to achieve here - https://github.com/eXist-db/exist/issues/565 and https://github.com/eXist-db/exist/issues/564.

I am trying to bring real transactions into eXist, you can see my work-in-progress here -https://github.com/eXist-db/exist/issues/565

This pull-request was a small foundation piece of work, so I am trying to get it upstreamed. I find it is better to send small atomic pull-requests rather than one huge one. Hopefully now you are aware of the bigger picture, you are assured that there is a purpose to this change.

You actually suggested this particular change yourself in the `TODO` that you added to the code in the commit I referenced, see - https://github.com/eXist-db/exist/commit/9f1c0c886d2fd3111e3f40dcceef4aca88b946b8#diff-2d43c839a493c8b4b4954533b3be5274R58

Regards your specific concerns raised above about API:

1) Yes, there are a few small API changes. These however are *internal* eXist APIs (e.g. undocumented APIs that are not published (i.e. XML-RPC, XML:DB, RESTServer, WebDAV etc.)).

2) The next release of eXist is a 3.0, this allows us a certain latitude in making more significant changes.

3) I have updated the entire eXist code base (the main weight is the test-suite) to take account for the new internal API changes. That only took me about 1.5 hours, and I can tell you there is a lot of copy-and-pasted `xquery.execute ` calls in there. It is unlikely any one else's code base would have that volume of changes, as most developers would have create a single compile/execute method wrapper and reused that.

shabanovd:
Please, stop say 'internal eXist APIs' and that why that change is ok. This is *NOT* true. And it does not matter - do you like it or not. There are different/another RESTServer, WebDAV, indexes and API that do use eXist's internal API (undocumented one too).

I did spend 2 hours to understand your work at https://github.com/adamretter/exist/tree/txn-thread and can say that there are a lot of API renaming that totally unnesesarry. 

How `try(final Txn txn = broker.continueOrBeginTransaction()) {` better then `try(finally Txn tx = broker.beginTx(){`?

I don't see any benefits of current PR in term of work at txn-thread branch. And *PLEASE* minimize number of API renaming just because you do not like the name it have!

PS. My suggestion at https://github.com/eXist-db/exist/commit/9f1c0c886d2fd3111e3f40dcceef4aca88b946b8#diff-2d43c839a493c8b4b4954533b3be5274R58 was to get `XQuery` instance from `Database`, because active broker always can be taken by `db.getActiveBroker()`, but it do not suggest any API change as you made.

adamretter:
@shabanovd I think discussion of the `https://github.com/adamretter/exist/tree/txn-thread ` is better done outside of this issue, otherwise we may confuse the two things.

However to answer your question, if you study the code of `broker.continueOrBeginTransaction` you will see that it does something entirely different to `broker.beginTx`. That is why it has a different name... because it is a different thing.

I am sorry but we disagree here. Internal APIs are internal and are subject to change. NativeBroker et al is part of the internal APIs. I have not changed the indexing API here, so you cannot complain there, what other external APIs (undocumented or otherwise) do you think I may have changed?

dizzzz:
@adamretter , I tend to agree with @shabanovd ; changing these APIs make a lot ppl unhappy, hence we had a strong disagrement earlier. I had quite some demotivating moments; with an large existing code base one should be rather conservative..... (i do not say a 'no' for all changes, I really like to stress the 'be conservative' statement)

wolfgangmm:
I have not looked at the PR in detail yet, so I will not comment on it. But I would like to pick up a suggestion by Joern: we will create a list of classes/packages which are currently considered part of the internal API (in particular those needed for expath package authors). Changes to those classes must be announced and require an increment of the eXist version number, so the package manager can detect incompatible changes. I have implemented the latter, but cannot commit yet since dashboard and other apps need to be updated first. It will fix the "pull and you're dead" issue we experienced recently and allows authors to slowly adopt their packages to newer versions.

adamretter:
@wolfgangmm Yes that is fine. After all there will be an increment to 3.0 soon, hence my comment that:

<blockquote>
The next release of eXist is a 3.0, this allows us a certain latitude in making more significant changes.
</blockquote>

adamretter:
@dizzzz If we cannot change the internal API, then we may as well all stop developing eXist, as it means we will not make any progress. The entire point of having an internal and external API is that we can change the internal one without affecting users. I am also working towards reducing the pain of having to update the each external API in turn when the internal API changes by offering this - https://github.com/eXist-db/exist/issues/564

shabanovd:
@adamretter you are saying strange things for me: method keep external logic (start transaction), but change internal code that's why it have to be renamed. Most of you changers have same logic and now I understand your mania ... Flying Spaghetti Monster bless you and your change, nothing to say more, sorry ;-)

dizzzz:
@adamretter yes I agree with you that changes are required to have progress. But you miss my subtile point. and yes, v3.0 would allow us to do (some  mor or less extended level of) changes of the APIs..

for now, I stop posting in this thread.

dizzzz:
@all we should only discuss the PR here ; if there is some more discussion e.g. with less related subjects , we should have the discussion on the exist-development mailing list, not here on github.

dizzzz:
@adamretter regarding this PR... as far as I can judge it is OK but i do don't oversee the details; anyway  you have answered my questions :-) so it must be ok :-)

adamretter:
@wolfgangmm @dizzzz So we know the next version of eXist will be 3.0, can we get this merged please?

dizzzz:
nice

PR #480, opened 2015-02-17
Updating svn extension library svnkit to version 1.8.8

ljo:
Updating svn extension library svnkit to version 1.8.8 to support jdk7u75+ and jdk8u31+. Please remove your current EXIST_HOME/lib/user/{svnkit,svnkit-cli}.jar to update on build.

PR #567, opened 2015-04-05
Ignore sequence-library.jar from svn extension

joewiz:
This jar was added with #480 to `extensions/svn/extension.xml`


adamretter:
Thanks @joewiz

joewiz:
Cheers!

PR #579, opened 2015-04-08
LuceneIndexWorker.getFieldContent using old API

### LINKED ISSUE #562: LuceneIndexWorker.getFieldContent is broken
* getFieldContent still uses old Lucene API and doesn't work properly
* extend ft:search to optionally return selected field content of hits for sorting and display

wolfgangmm:
* getFieldContent still uses old Lucene API: fails for larger data sets or returns random result
* related: extend ft:search to optionally return selected field content of hits for sorting and display

Closes #562

### DIFF
--- a/extensions/indexes/lucene/src/org/exist/indexing/lucene/LuceneIndexWorker.java
+++ b/extensions/indexes/lucene/src/org/exist/indexing/lucene/LuceneIndexWorker.java
@@ -21,15 +21,7 @@
 package org.exist.indexing.lucene;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


adamretter:
Could you provide regressions tests for the issues with `getFieldContent`?

wolfgangmm:
There are some tests. The actual issue requires a substantial amount of data in the index to produce wrong results (enough to trigger lucene to split the index files), so it's not testable in the test suite. It will show up in the performance test suite I'm working on though.

adamretter:
@wolfgang okay.

PR #585, opened 2015-04-10
Plumb the leaky abstraction in XML:DB Local API

adamretter:
[bugfix] Fixes leaky abstraction in XMLDB:Local API by preventing internal persistent DOM node state from leaking out.

### DIFF
--- a/src/org/exist/xmldb/LocalXMLResource.java
+++ b/src/org/exist/xmldb/LocalXMLResource.java
@@ -19,6 +19,9 @@
  */
 package org.exist.xmldb;
 
+import net.sf.cglib.proxy.Enhancer;
+import net.sf.cglib.proxy.MethodInterceptor;
+import net.sf.cglib.proxy.MethodProxy;
 import org.exist.dom.persistent.NodeProxy;
 import org.exist.dom.persistent.XMLUtil;
 import org.exist.dom.memtree.AttrImpl;
@@ -39,6 +42,7 @@ import org.exist.xquery.value.StringValue;
 import org.exist.xquery.value.Type;
 import org.w3c.dom.DocumentType;
 import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
 import org.xml.sax.*;
 import org.xml.sax.ext.LexicalHandler;
 import org.xmldb.api.base.ErrorCodes;
@@ -49,7 +53,10 @@ import javax.xml.transform.TransformerException;
 import java.io.File;
 import java.io.IOException;
 import java.io.StringWriter;
+import java.lang.reflect.Method;
+import java.util.Optional;
 import java.util.Properties;
+import java.util.stream.Stream;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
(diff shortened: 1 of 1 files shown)


dizzzz:
Is there a jar file missing?

src/org/exist/xmldb/LocalXMLResource.java
```diff
@@ -190,6 +197,42 @@ public Node getContentAsDOM() throws XMLDBException {
...
+    /**
+     * Provides a safe export of an internal persistent DOM
+     * node from eXist via the Local XML:DB API.
+     *
+     * This is done by providing a proxy object that only implements
+     * the appropriate W3C DOM interface. This helps prevent the
+     * XML:DB Local API from leaking implementation through
+     * its abstractions.
+     */
+    private Node exportInternalNode(final Node node) {
```
dizzzz: Potential NPE?

adamretter: Do you mean line 184? if so then no, that is within a lambda and the resulting value is never assigned.

adamretter:
@dizzzz I don't think there is a JAR file missing? Why do you say that?

dizzzz:
Sorry I was on my mobile device, coul dnot get the big picture. i saw some sf.net imports I did not recognize. Hence my question.

src/org/exist/xmldb/LocalXMLResource.java
dizzzz: no because of 192 (202) ; will there be a scenario with a NULL value for `result` ?

adamretter: I don't think so, if you have a `NodeProxy` for a node in a document, then you must always be able to retrieve the real node from the document, otherwise you have bigger problems ;-)

adamretter:
@dizzzz The `net.sf.cglib` imports are for cglib which is in `$EXIST_HOME/lib/core`

dizzzz:
Very Well :)

dizzzz:
Any idea about performance risks? It is only about removing access to 'internal' methods? Are there expensive operations involved? As you see, I am having a hard time reading this code :-(

wolfgangmm:
Accessing the internal DOM directly was never fast, so I don't think the PR will make it slower.

A better solution would be to expand the persistent DOM node into an in-memory DOM. We have efficient streaming methods for this.

adamretter:
@wolfgangmm but then we would still leak an internal model into an external context.

shabanovd:
why do you afraid that so much?

adamretter:
@shabanovd I am not afraid. I am just fixing an engineering bug that can lead to real problems.

shabanovd:
I don't see this as a bug. Java paradigm give us interfaces and classes (implementations) and you afraid that classes gets out?

shabanovd:
also, this wrapping lead to jvm runtime optimizations fail

wolfgangmm:
The persistent DOM has methods to modify the database and I thus agree it should not be exposed. This does not apply to the in-memory DOM though, which is just a standard DOM without side effects.

shabanovd:
and that should be fixed by interface but not by encapsulation, no?

adamretter:
@wolfgangmm As this is an improvement on what we currently have, can I suggest that we merge it please?

Issue #581, opened 2015-04-08
Remove legacy range-index (NativeValueIndex)


adamretter:
I found some detail/discussion that I started about doing this previously: http://exist.markmail.org/thread/uymso4dl6idflmwc#query:+page:1+mid:uymso4dl6idflmwc+state:results

@wolfgangmm It would be good if you could give me any follow-up on that thread, i.e. if you have any strong thoughts on how to deal with fn:matches.

wolfgangmm:
There are still some functions which could not be ported to the new range index yet, including ft:matches, ft:id and ft:idref. Moving the id functions into lucene shouldn't be difficult. fn:matches is more of a problem due to the limited support for regex in lucene.

adamretter:
It looks to me like `org.exist.storage.StorageAddress` encodes the type of the value for a node as indexed by the Legacy Range Index and stores that into the `dom.dbx`. Presumably if the legacy range index was removed we could also remove this now unnecessary storage. @wolfgangmm can you confirm?

adamretter:
Will also lead to the removal of `DBBroker#endElement` and `NativeBroker#endElement`

adamretter:
It seems that we have two options with Lucene and regular expressions:

1. We use the older [RegexQuery](http://lucene.apache.org/core/4_4_0/sandbox/org/apache/lucene/sandbox/queries/regex/RegexQuery.html) class and call `setRegexImplementation(new JavaUtilRegexCapabilities())`.

2. We attempt to compile the XPath regular expression language down into multiple [Regexp](https://lucene.apache.org/core/4_4_0/core/org/apache/lucene/util/automaton/RegExp.html?is-external=true) automata e.g. using the newer [RegexpQuery](https://lucene.apache.org/core/4_0_0/core/org/apache/lucene/search/RegexpQuery.html).

The javadoc for (1) warns that it is not as fast as (2) but it does have full Java regular expression support. A look at the Lucene code shows that it does an index scan calling `matches` on a `java.util.pattern.Matcher`, as this is pretty much what the `org.exist.storage.RegexMatcher` already does in eXist, I am hopeful we can avoid a performance drop.

I will attempt (1) as it seems like the simplest approach.

PR #586, opened 2015-04-11
remove old FT index

### LINKED ISSUE #556: Remove legacy full-text index
The legacy full-text index should be removed for the 3.0 release. We have been advising people not to use it since before eXist 2.0 was released!

shabanovd:
close #556


src/org/exist/storage/TermMatcher.java
```diff
@@ -1,43 +1,31 @@
...
 /**
- * Implementations of this interface can be passed to method
-* {@link org.exist.storage.TextSearchEngine#getNodes(XQueryContext context, DocumentSet docs, NodeSet contextSet, int axis, QName qname,
-	        TermMatcher matcher, CharSequence startTerm)} 
- * to check if an index entry matches a given search term.
+ * Check if an index entry matches a given search term.
  * 
  * @author Wolfgang Meier (wolfgang@exist-db.org)
  */
 public interface TermMatcher {
```
adamretter: Is this interface still needed?

shabanovd: yes, it may be used by value index

adamretter: @shabanovd Okay cool so we can drop after https://github.com/eXist-db/exist/issues/581

PR #620, opened 2015-05-03
Refactoring validation code, and code clean up. More to come.

dizzzz:


dizzzz:
More to come.... But not now :)

PR #630, opened 2015-05-20
Upgrade Tika library to v1.8 

dizzzz:
a lot new eps are pulled in; after stripping down 82megs remain :-/

It is an upgrade from 1.5 -> 1.8:

changelog:
https://tika.apache.org/1.6/index.html
https://tika.apache.org/1.7/index.html
https://tika.apache.org/1.8/index.html

### DIFF
--- a/extensions/contentextraction/ivy.xml
+++ b/extensions/contentextraction/ivy.xml
@@ -2,38 +2,43 @@
 <ivy-module version="2.0">
 	<info organisation="org.exist" module="tika" />
 	<dependencies>
-		<dependency org="org.apache.tika" name="tika-parsers" rev="1.5" conf="*->*,!sources,!javadoc">
-			<exclude module="jdom" />
-			<exclude module="slf4j-api" />
-			<exclude module="commons-codec" />
-			<exclude module="commons-compress" />
-			<exclude module="commons-logging" />
-			<exclude module="commons-httpclient" />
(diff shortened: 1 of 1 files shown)


joewiz:
See change log for Tika 1.8 at https://tika.apache.org/1.8/index.html

dizzzz:
and...

https://tika.apache.org/1.7/index.html
https://tika.apache.org/1.6/index.html

joewiz:
That's right! We'd been at 1.5. 

I like this from **1.6**: 

> PDF: Images in PDF documents can now be extracted as embedded resources
> Extract attachments from RTF files

(I wonder if we can access these via the contentextraction module?)

Intriguing:

> Added module and core Tika interface for translating text between languages and added a default implementation that call's Microsoft's translate service

This will be helpful for hsg policy studies archive cms app:

> Support Outlook Personal Folders File Format *.pst

This from **1.7** is very cool:

> Created a new Tesseract OCR Parser to extract text from images (Requires installation of Tesseract before use)

Again, I wonder if we can trigger this from contentextraction?

And from **1.8**:

> Added basic style support for ODF documents
> Increased the speed of language identification by a factor of two

All in all, neat stuff - but I wonder what's involved in exposing PDF/RTF attachments, OCR text extraction, and even language translation to the contentextraction module?

PR #637, opened 2015-05-31
Make log4j write to $sys:exist.home

hungerburg:
even when starting up from a different directory than $EXIST_HOME

E.g: `cd /tmp; /opt/eXist/bin/startup.sh` fails when property not pased

Should not impact the wrapper, does it? Or the WAR?

### DIFF
--- a/bin/functions.d/eXist-settings.sh
+++ b/bin/functions.d/eXist-settings.sh
@@ -11,7 +11,7 @@ get_exist_home() {
 			p="$PWD/$1"
 		;;
 	esac
-		(cd $(/usr/bin/dirname "$p") ; /bin/pwd)
+		(cd "$(/usr/bin/dirname "$p")" ; /bin/pwd)
 }
 
 check_exist_home() {
@@ -20,6 +20,10 @@ check_exist_home() {
 	EXIST_HOME="$EXIST_HOME_1/..";
     fi
 
+    if [[ "$EXIST_HOME" =~ ' ' ]] ; then
+	echo "Path to eXist-db MUST NOT contain spaces! Exiting now…" > /dev/stderr;
+	exit 1;
+    fi
     if [ ! -f "${EXIST_HOME}/start.jar" ]; then
 	echo "Unable to find start.jar. Please set EXIST_HOME to point to your installation directory." > /dev/stderr;
 	exit 1;
--- a/log4j2.xml
+++ b/log4j2.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Configuration xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
     <Properties>
-        <Property name="logs">webapp/WEB-INF/logs</Property>
+        <Property name="logs">${sys:exist.home}/webapp/WEB-INF/logs</Property>
         <Property name="rollover.max.size">10MB</Property>
         <Property name="rollover.max">14</Property>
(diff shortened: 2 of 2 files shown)


dizzzz:
ah, I was not aware of the issue; this might work very well for the wrapper config too (though the logs are now written to tools/wrapper/logs/... 
for the WAR deployement...... We'd need to figure out. I think there is some work left :-)

dizzzz:
is the $sys:exist.home always reliably set to a value? for e.g. client.sh and startup.sh and ... ?

adamretter:
Nope it's not I am just emailing the log4j guys to find out if there is a better way. What we really want is the log dir to be relative to the log4j2.xml configuration file.

dizzzz:
for tomcat et al we could use the value of rootDir or contextPath I guess; see https://logging.apache.org/log4j/2.0/manual/lookups.html#WebLookup

adamretter:
@dizzzz I am hopeful that we can find a more generic solution yet... but, if not...

hungerburg:
Browsing the various scripts, it looks to me, like EXIST_HOME gets passed to java in any of them: All of `backup.sh  client.sh  server.sh  shutdown.sh  startup.sh` source `functions.d/eXist-settings.sh` and call set_exist_options (OPTIONS="-Dexist.home=$EXIST_HOME") and run the jvm with  ${OPTIONS}. Only `run.sh` does it in itself, which also what the windows `.bat` scripts do.

adamretter:
@hungerburg I think that relies on the user remembering to set EXIST_HOME in their environment though, which we don't currently enforce.

I have now sent a patch to the log4j project that I think will address this https://issues.apache.org/jira/browse/LOG4J2-1050

hungerburg:
@adamretter As I am one of those lazy users, that do not set EXIST_HOME, I can tell you, that the heuristics work: I put in a `set -x` into bin/startup.sh to look at the trace; I set EXIST_HOME in the environment to something false, and startup fails with `Unable to find start.jar.` That tells me, that there wont ever be something for log4j to log, in case EXIST_HOME is wrong. I also ran client.sh, server.sh; but not the wrapper and not the WAR.

hungerburg:
@adamretter Looking at the log4j jira page I tried and found: eXist-db cannot be installed (at least on UNIX like OSes) in a directory, of which the full path contains a space. It is very tricky in bash, to perform string concatenations, as the outsourced functions do, and keep quotes. (One might use arrays for that, but probably not worth the effort…)

adamretter:
Okay my patch to log4j has been merged. It handles spaces in paths now too. When log4j 2.4 is released we should be able to use something like:

```xml
<Property name="logs">${log4j:configParentLocation}/webapp/WEB-INF/logs</Property>
```

dizzzz:
Good busy !!!!

hungerburg:
Looks like log4j 2.4 got released.

adamretter:
@hungerburg Okay I am taking a look now...

adamretter:
hmm seems to be a problem in 2.4 - https://issues.apache.org/jira/browse/LOG4J2-1050?focusedCommentId=14909793&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14909793

adamretter:
I have a patch for this with the Apache Log4j people - https://issues.apache.org/jira/browse/LOG4J2-1050?focusedCommentId=15183496&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15183496

I will open an issue also here so this doesn't end up skipped before 3.0 is released.

PR #645, opened 2015-06-02
Use modification dates to extract only the resources that are newer than their destination files.

ptrckb:
When destination file already exists, it compares the timestamps of resource and destination file and, by default, extracts the resource only if it is newer than the destination file.


adamretter:
@ptrckb Any update on this?

ptrckb:
I thought it was the way the ant copy task was working but it seems I was wrong. I may try again later.

line-o:
I would close this PR for now. It has been open for over 4 years with no further activity.

joewiz:
@ptrckb Please feel free to open a new PR if you return to this.

PR #648, opened 2015-06-04
[bugfix] New range index incorrectly processed != like =

wolfgangmm:
Optimiser did rewrite comparisons using the != or ne operator. However, the index did not support this operator and replaced it by =, leading to wrong query results!


joewiz:
Great! Thanks for getting to the bottom of this head-scratcher, @wolfgangmm!

PR #649, opened 2015-06-06
Better expath compatibility checks

### LINKED ISSUE #559: Prevent loading incompatible expath package jars
Fix the "pull and you're dead" issue by having the classloader check the semver of a package before enabling bundled jars. Allow package authors to ship different versions of a xar targetting different releases of eXist:
 
* enforce checks of semver-min/-max etc. when deploying an expath package
* to prevent a stale installation: make the class loader ignore jars from packages defining a version not compatible with the current semver
* for packages which do not define a processor version, assume they only support version 2.2 and below of eXist: such packages can be installed, but their jars will not be loaded into the class loader.
* extend dashboard to only offer packages which are compatible with the current processor semver.

wolfgangmm:
Avoid installing incompatible Java packages from the package repo. Right now there is no protection against installing a xar package containing incompatible Java libraries.

This PR handles two scenarios:

1. the user tries to install an incompatible xar package: enforce checks of semver-min/-max during deployment. Refuse to deploy a package which indicates to be incompatible with the current eXistdb build.
2. the user updates an eXistdb built from git and has some old xar packages installed: this is the worst case because it will likely mess up the class loader, leading to a broken eXist install. Fix: make the class loader ignore jars from packages defining a version dependency not compatible with the semver of the current eXistdb build.

With this PR, package authors can finally provide different xar package versions for different builds of eXistdb. For example, a xar may define a dependency on the 2.2.x series of eXist by adding the following line to expath-pkg.xml:

```xquery
<dependency processor="http://exist-db.org" semver-max="2.2"/>
```

The next version of the xar may then define:

```xquery
<dependency processor="http://exist-db.org" semver-min="2.3.0"/>
```

so it will be chosen for eXistdb > 2.3.0. Because older versions of eXist do not report a usable version number, the code assumes 2.2.0.

This pull request will be complemented by changes to the public app repository and the dashboard: both should only offer compatible xar packages for deployment.

Closes #559.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -6,7 +6,8 @@
 project.name = eXist-db
 project.version = 2.3dev
 project.version.numeric = 2.3.0
-project.codename = Ruesselsheim
+project.source = develop
+project.codename = Berlin
 
 # build settings
 build.debug = on
--- a/build.properties
+++ b/build.properties
@@ -4,8 +4,8 @@
 #
 # $Id$
 project.name = eXist-db
-project.version = 2.2
-project.version.numeric = 2.2
+project.version = 2.3dev
+project.version.numeric = 2.3.0
 project.codename = Ruesselsheim
 
 # build settings
--- a/build/scripts/build-impl.xml
+++ b/build/scripts/build-impl.xml
@@ -270,6 +270,7 @@
     <target name="jar" depends="git.details,compile,compile-aspectj"
         description="Create eXist-db unsigned jar files">
         <filter token="version" value="${project.version}"/>
+	<filter token="semver" value="${project.version.numeric}"/>
         <filter token="build" value="${DSTAMP}"/>
         <filter token="git.commit" value="${git.commit}"/>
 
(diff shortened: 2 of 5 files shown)


joewiz:
Excellent!  Could you release info for .xar maintainers about what kinds of things need to be recompiled for 2.3dev, or specified as 2.2 only.  All xars with jars, I assume?  Even xquery xars with xquery 3.1 features should probably be re-released as 2.3dev+, right?  Perhaps all xars allowed into eXist-db's public repo should be re-released with semver, just as a baseline?

wolfgangmm:
@joewiz with the next release, all xars with jars should specify the eXist version they support by setting the semver-min and semver-max attributes. Fortunately there are not that many, but since installing an incompatible jar can lead to eXist refusing to start up, it is important to define the version the xar is supposed to work with.

A standard xar without jars is less problematic. In the worst case, it will just not work properly. If you use XQuery 3.1 on eXist 2.2, you'll get an error, but that's it. However, it certainly makes sense to provide XQuery 3.0 and XQuery 3.1 versions of an app, so users get the one matching their install.

There will be an update to the public repo soon. I'm currently testing it on a second server. Because every eXist install will access it, we have to make sure it works properly. If you would like to help testing, just drop me a note.

Issue #377, opened 2014-10-29
Permissions does not copy on copy operation 


shabanovd:
here an old fix https://github.com/shabanovd/exist/commit/8d36f0fa33a65dc235ea7068196de2ea708fd5e6

adamretter:
Can you be specific, are you copying a collection or file?

shabanovd:
To reproduce 
 - create collection
 - add document there
 - set permissions
 - copy that collection

new document must have same permissions as source one have

PR #656, opened 2015-06-09
Fix to ensure mode and ACL on sub-resources and collections is copied when copying from parent

adamretter:
Closes https://github.com/eXist-db/exist/issues/377


dizzzz:
interesting names :-) _1 and _2

adamretter:
@dizzzz Well I just followed the Scala convention here for a Tuple... I think it's fine.

dizzzz:
ok; just learning :-)

PR #670, opened 2015-06-14
Bug fix on whitespace store

shabanovd:
add processText method to make sure that text node processed by same logic all around


wolfgangmm:
Great, I'll test this. One question: why did you change the password in all the tests from null to ""? This does not seem to be related to this PR.

ljo:
Yes, looks very good without those null -> "" for pw in tests.

shabanovd:
@wolfgangmm @ljo That get mixed in. I can make different PR for that. The problem with null that admin password is "", with null authentication exception is raise.

wolfgangmm:
@shabanovd I tracked down an issue reported for xsltforms to this pull request. eXist is now eating up whitespace where it should not. To reproduce, store the following document:

```xml
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
    <xsl:template match="processing-instruction()" mode="xml2string">
        <xsl:text>&lt;?</xsl:text>
        <xsl:value-of select="name()"/>
        <xsl:text> 
            
</xsl:text>
        <xsl:value-of select="."/>
        <xsl:text>?&gt;</xsl:text>
    </xsl:template>
</xsl:stylesheet>
````

and retrieve it via REST:

```
curl http://localhost:8080/exist/rest/db/test.xsl
```

it outputs:

```xml
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
    <xsl:template match="processing-instruction()" mode="xml2string">
        <xsl:text>&lt;?</xsl:text>
        <xsl:value-of select="name()"/>
        <xsl:text/>
        <xsl:value-of select="."/>
        <xsl:text>?&gt;</xsl:text>
    </xsl:template>
</xsl:stylesheet>
```

Please note the <xsl:text/>, which is invalid and breaks the xslt transform. Undoing your commit to Indexer fixes the problem. Unfortunately I don't have much time to dive into the details. Could you have a look?

shabanovd:
@wolfgangmm sure, I'll check over weekend

adamretter:
Should this be reverted perhaps?

wshager:
Please, fix this ASAP

wolfgangmm:
@ljo has some pending changes concerning this issue. It's actually the main showstopper for me right now.

wshager:
@wolfgangmm I should probably write this elsewhere, but there's the mysterious issue with blank documents retrieved through REST (#840) and the FLWOR sort clauses killing performance (#839).

PR #674, opened 2015-06-16
sweeping old FT index staff

shabanovd:


ljo:
I am testing this. I don't want the change in src/org/exist/indexing/IndexController.java though, since that public addIndexWorker() is the only way right now, from what I can see, to add expath indexing libraries. I was going to change the comment to reflect this when your changes came in. So the comment still needs be changed.

adamretter:
@ljo I am not aware of any EXPath indexing modules?

ljo:
@adamretter I am trying to push some out. Firstly a sparql one. But I need to overcome these registration obstacles before it can be released.

adamretter:
@ljo interesting :-)

shabanovd:
@ljo I was think exactly the same, but I was not sure that it do all required preparations.

There is the method 'IndexManager.registerIndex' that fit must better https://github.com/shabanovd/exist/blob/RC-no-move-fix/src/org/exist/indexing/IndexManager.java#L107

I'll make PR for that.

ljo:
OK. Thanks.

adamretter:
@ljo Does that mean you are happy with @shabanovd's change? If so I would be happy to review this and then perhaps merge

ljo:
Merged after some manual resolutions.

PR #678, opened 2015-06-18
JMX interface enhancements

wolfgangmm:
Fix and extend JMX interface to provide more information required to diagnose bottlenecks on a live production system: 

1. use DelayQueue for recent query log, so it shows all queries during the past X minutes which were slower than X ms (configurable)
2. add thread names, so query stacks, lock manager output and thread dumps can be connected
3. optional: log the URI which triggered an xquery where possible

The changes were made during the past months while analyzing runtime issues on real production systems. They fill in some missing bits and pieces which are required to diagnose performance and concurrency issues.

monex 0.8 is able to display the additional information, though any jmx client will work (or just call /exist/status to get an XML dump).

### DIFF
--- a/src/org/exist/management/impl/ProcessReport.java
+++ b/src/org/exist/management/impl/ProcessReport.java
@@ -54,24 +54,32 @@ public class ProcessReport implements ProcessReportMBean {
     private static String[] pIndexNames = { "id" };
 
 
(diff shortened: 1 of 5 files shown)


dizzzz:
Nice......

joewiz:
Happy to see this in the develop branch - this info, as logged in monex, has made a huge difference to us.

PR #693, opened 2015-06-25
Many simple bug fixes that were identified through static analysis

adamretter:
Fixes 363 bugs identified by static analysis with FindBugs. Still some 1000+ issues to check...


shabanovd:
I can't say that I like fix without evidence of problem specially with `synchronized`. It easily can lead to deadlock.

adamretter:
@shabanovd You are welcome to run FindBugs yourself and check through the issues or you can rely that I have done this already.

shabanovd:
@adamretter FindBugs is not that smart after all (and I did it from time to time)

adamretter:
@shabanovd Yes I agree that FindBugs can find false-positives... which is why I looked at each of these changes very carefully before I made them!

shabanovd:
@adamretter to be clean, I'm not against all changers or I do not say that you did not spend time on each commit. All I want to say that `synchronized` at wrong place can bring headache.

adamretter:
@shabanovd I agree, I think it is in the right place though now. Do you think otherwise?

shabanovd:
@adamretter that just scared me ...

adamretter:
@shabanovd So we can merge or you want to investigate?

shabanovd:
if some one else review that then ok

PR #694, opened 2015-06-25
Deadlock when storing binary resource

wolfgangmm:
Storing a binary resource caused a deadlock when the after*Document trigger was called, in particular if that trigger tried to lock another collection (as the restxq trigger inevitably does when compiling imported modules). The deadlock occurs most often when installing xar packages into the db (because many XQuery scripts are updated), but it may also be triggered in other situations.

Fix: Collection.addBinaryResource must release the collection lock before calling any after* triggers. Instead it keeps a read lock on the document to prevent concurrent modification.

Also, the package collection should not remain locked while files are being installed. This is another source of potential deadlocks, but at least limits concurrency.


joewiz:
This is a very exciting commit which, if successful, could make a huge difference for me and all users of xars and restxq apps! Looking forward to testing it!

PR #708, opened 2015-06-29
- patched wrapper from @adamretter. Linux x86_64 binaries

zwobit:
start|status|stop tested gentoo + systemd 64bit


zwobit:
Copied old comment from old PR:

@dizzzz woops.... why is wrapper.jar updated? I'd not do that, is it really patched for linux? Otherwise i'; like to keep the existing one?

PR #716, opened 2015-07-03
allow to store non-wellformed html: if parsing fails during deploymen…

JoernT:
…t and mimetype is html the resource will be stored as binary. This is important for supporting HTML5 libs like Polymer and others.

### DIFF
--- a/src/org/exist/repo/Deployment.java
+++ b/src/org/exist/repo/Deployment.java
@@ -22,6 +22,7 @@
 package org.exist.repo;
 
 import org.exist.SystemProperties;
+import org.exist.collections.triggers.TriggerException;
 import org.exist.dom.memtree.DocumentBuilderReceiver;
 import org.exist.dom.memtree.InMemoryNodeSet;
 import org.exist.dom.memtree.DocumentImpl;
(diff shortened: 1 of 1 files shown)


JoernT:
To this to work properly within a xar application the following snippet needs to be added to the controller.xql
``` 
if (ends-with($exist:resource, ".html")) then (
    util:declare-option("exist:serialize", "method=html5 media-type=text/html"),
    doc($app-root || $exist:resource)
)
```

adamretter:
@JoernT Wouldn't it be better to do this change at a lower level so that all of the APIs benefit from it and not just PKG?

wolfgangmm:
As a developer it would irritate me if my html files were sometimes stored as binaries, sometimes not - in particular because I will not get any feedback about it. For app packages this is ok, but I see it as a temporary solution until we have added a mechanism to repo.xml to describe how particular resources should be handled. So I would accept Joern's PR unless someone has further objections...

JoernT:
replaced this pull request with https://github.com/eXist-db/exist/pull/727 which supports declarative syntax in repo.xml - for the full story eXide needs to be extended to handle HTML files that are stored as binary but storing and serving of 'binary HTML' works.

PR #721, opened 2015-07-05
when XmldbURI created without any exception lastSegment must not raise any too

shabanovd:
when XmldbURI created without any exception lastSegment must not raise any too


dizzzz:
sorry for being late... what exception did you see?

dizzzz:
looks like a simple fix, but I can't judge the consequences; @adamretter @ljo @wolfgangmm  ?

PR #726, opened 2015-07-08
Moving expathRepo init and updateClasspath() earlier

ljo:
in order to have indexing apps available when indexManager is initialized and startup triggers are executed.

### DIFF
--- a/src/org/exist/storage/BrokerPool.java
+++ b/src/org/exist/storage/BrokerPool.java
@@ -619,7 +619,7 @@ public class BrokerPool implements Database {
 
     private ClassLoader classLoader;
 
-    private ExistRepository expathRepo = null;
+    private Optional<ExistRepository> expathRepo = Optional.empty();
 
     /**
      * Creates and configures the database instance.
@@ -873,9 +873,9 @@ public class BrokerPool implements Database {
                         try {
                             // initialize EXPath repository so indexManager and
                             // startup triggers can access it
-                            expathRepo = ExistRepository.getRepository(this.conf);
+                            expathRepo = Optional.ofNullable(ExistRepository.getRepository(this.conf));
                         } catch(final PackageException e) {
-                            LOG.warn("Failed to initialize expath repository: " + e.getMessage() + " - " +
+                            LOG.error("Failed to initialize expath repository: " + e.getMessage() + " - " +
                                      "indexing apps and the package manager may not work.");
                         }
                         ClasspathHelper.updateClasspath(this);
@@ -1268,7 +1268,7 @@ public class BrokerPool implements Database {
         return conf;
     }
 
-    public ExistRepository getExpathRepo() {
+    public Optional<ExistRepository> getExpathRepo() {
         return expathRepo;
     }
 
--- a/src/org/exist/storage/BrokerPool.java
+++ b/src/org/exist/storage/BrokerPool.java
@@ -1,6 +1,6 @@
 /*
  * eXist Open Source Native XML Database
- * Copyright (C) 2003-2014 The eXist-db Project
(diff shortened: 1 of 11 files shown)


shabanovd:
Are you doing this because you need jars from expath repo?

Maybe better to put declaration to package to call something during startup? Or use pluing-ins infrastructure?

ljo:
It is in order to have indexing pipeline modules provided as apps available when initializing IndexManager, this was the simplest change to get it working right now. It also can be part of recovery if reindex is performed. These self contained apps were actually not available for startup triggers either. I will check once more having it working in plugin infra, but think it fits better like this covering both plugins infra and indexing pipeline infra.

src/org/exist/storage/BrokerPool.java
```diff
@@ -871,6 +870,16 @@ protected void initialize() throws EXistException, DatabaseConfigurationExceptio
                         symbols = new SymbolTable(conf);
                         isReadOnly = isReadOnly || !symbols.getFile().canWrite();
 
+                        try {
+                            // initialize EXPath repository so indexManager and
+                            // startup triggers can access it
+                            expathRepo = ExistRepository.getRepository(this.conf);
+                        } catch(final PackageException e) {
+                            LOG.warn("Failed to initialize expath repository: " + e.getMessage() + " - " +
```
adamretter: It seems to me that his should be an error and not a warning. As the package repository will be absolutely unavailable if an exception is thrown here... Not to mention that subsequent attempted access will likely cause an NPE.

ljo:
@adamretter I agree, but it is a warning now, I just moved it to avoid other NPEs which causes indexing apps to not work. I can definitely change it to error if that will make it go in.

src/org/exist/storage/BrokerPool.java
shabanovd: +1 to throw exception, but it also require a way to repair such exception ... thoughts?

wolfgangmm: No, don't do that! To recover your data in case of an emergency you need the broker pool to come up, even if there are inconsistencies.

A damaged expath repo is a minor issue and all your data is likely still intact, so you should be able to start up, run an export and repair afterwards.

adamretter: I would just log an error but use an Optional as the reference type
instead, and then make future access to the Optional conditional to avoid
further NPEs

On 9 July 2015 at 12:51, Dmitriy Shabanov <notifications@github.com> wrote:

> In src/org/exist/storage/BrokerPool.java
> <https://github.com/eXist-db/exist/pull/726#discussion_r34247002>:
>
> > @@ -871,6 +870,16 @@ protected void initialize() throws EXistException, DatabaseConfigurationExceptio
> >                          symbols = new SymbolTable(conf);
> >                          isReadOnly = isReadOnly || !symbols.getFile().canWrite();
> >
> > +                        try {
> > +                            // initialize EXPath repository so indexManager and
> > +                            // startup triggers can access it
> > +                            expathRepo = ExistRepository.getRepository(this.conf);
> > +                        } catch(final PackageException e) {
> > +                            LOG.warn("Failed to initialize expath repository: " + e.getMessage() + " - " +
>
> +1 to throw exception, but it also require a way to repair such exception
> ... thoughts?
>
> —
> <https://github.com/eXist-db/exist/pull/726/files#r34247002>.
>



-- 
Adam Retter

skype: adam.retter
tweet: adamretter
http://www.adamretter.org.uk

ljo: So can we keep it as is in this commit and I will make a new commit with Optional later on?

shabanovd: ok for me

PR #730, opened 2015-07-12
Small minor cleanups

adamretter:


shabanovd:
@adamretter Is it same PR or new one? All comments washout 3 times now.

adamretter:
@shabanovd Same except changes to `[refactor] minor cleanup` as you suggested. Not sure what to do about the `[refactor] Do not config collections twice` one yet, can you comment on the thread on the mailing-list please, and then you comments won't be lost when I rebase.

adamretter:
I moved the double-collection-config fix to a separate branch https://github.com/eXist-db/exist/pull/734 so as not to delay this one further.

shabanovd:
@adamretter this one is also ok, comment was on minor cleanup (move synchronized to method signature), so if everyone happy it can go.

adamretter:
@shabanovd @ljo If happy, please merge...

ljo:
Yes, I am happy.

adamretter:
@ljo please can you merge then

src/org/exist/storage/lock/DeadlockDetection.java
```diff
@@ -21,6 +21,10 @@
  */
 package org.exist.storage.lock;
 
+import org.apache.log4j.LogManager;
```
dizzzz: how can this work? this is log4j 1.2.x right?

dizzzz: answering myself...  because of the log4j compatibility jar in lib/user ....

adamretter: Ah crap! There is an excellent example of why we shouldn't of added it!

PR #734, opened 2015-07-14
[refactor] Do not config collections twice

adamretter:
http://exist-development.markmail.org/thread/g6vgcsmbp3dcnxst

### DIFF
--- a/src/org/exist/collections/CollectionConfigurationManager.java
+++ b/src/org/exist/collections/CollectionConfigurationManager.java
@@ -130,18 +130,6 @@ public class CollectionConfigurationManager {
             // TODO : unlock the collection here ?
             confCol.store(txn, broker, info, config, false);
             // broker.sync(Sync.MAJOR_SYNC);
-
-            latch.writeE(new Callable<Void>() {
-                @Override
-                public Void call() throws Exception {
-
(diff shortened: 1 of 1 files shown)


adamretter:
@ljo this wasn't ready to land yet! @shabanovd had some concerns I think. I am not sure we need to revert, but I might need to send a follow-up.

ljo:
@adamretter OK. I thought your message said split the PR #730 so this is not delayed. But it was the other way around then.

adamretter:
@ljo Yes, sorry for the ambiguous language.

shabanovd:
@adamretter question was what to do with configurations in future. Is it ok to shift to ConcurrentHashMap? (I think yes)

adamretter:
Hmmm... I thought you were worried that the second collection configuration
I removed was thread-safe, whereas the first (which remains) is not?

On 14 July 2015 at 17:02, Dmitriy Shabanov <notifications@github.com> wrote:

> @adamretter <https://github.com/adamretter> question was what to do with
> configurations in future. Is it ok to shift to ConcurrentHashMap? (I think
> yes)
>
> —
> <https://github.com/eXist-db/exist/pull/734#issuecomment-121291270>.
>



-- 
Adam Retter

skype: adam.retter
tweet: adamretter
http://www.adamretter.org.uk

PR #746, opened 2015-07-18
[bugfix] Avoid NoClassDefFoundError

adamretter:


dizzzz:
with our PR... 'ClassNotFoundException's will become fatal?

adamretter:
@dizzzz No they are still caught too

dizzzz:
@adamretter right... sometimes a screen or window can be too small :-)

PR #759, opened 2015-08-04
Feature/bugfix namespace stack

adamretter:
Fixes a bug where the XQuery Context is not cleaned up if a variable reference cannot be resolved. This leads to the first run of the query producing one error, and all subsequent runs producing a different (wrong) error.


dizzzz:
nice catch.....

PR #762, opened 2015-08-12
Wrap non-nodes with <exist:value> element.

### LINKED ISSUE #755: XML RPC, query vs retrieveAll
Looking to `src/org/exist/xmlrpc/RpcConnection.java`, there are mainly two approaches to query eXist:

   1. Using the pair `executeQuery` + `retrieve(All)` methods.
   2. Using the `query` method.

The second has two interfaces, one returning a byte array, and marked as deprecated in the website, and another one returning a string, that I am not sure if it is deprecated or not. 

Nevertheless, using `retrieveAll` or `query` behaves differently. If the query return a set of text() values, the result is the same from both methods: a `<exist:result>` with the strings concatenated. In the other hand, if the query returns a set of URI (for example, mapping `fn:base-uri` to a set of resources), `retrieveAll` returns a `exist:result` with the URI concatenated. If using `query`, the result is a `exist:result` that contain `exist:value` elements.

If this difference is the desired one, then I need to know if `query` returning a string is or not deprecated (and I hope it is not). In the other hand, if the difference is not desired, probably this should be fixed.

Looking to the code, the main different is that `retrieveAll` uses `retrieveAllAsString` that does the following code for each answer:

``````java
                   if (Type.subTypeOf(current.getType(), Type.NODE)) {
                        current.toSAX(broker, handler, null);
                    } else {
                        value = current.toString().toCharArray();
                        handler.characters(value, 0, value.length);
                    }
``````
that is, if the result is a XML fragment, return it, if not, string the string without any outer tag.

In the other hand, `query` uses `printAll` that has a similar iteration, with this code:
`````java
            if (item.getType() == Type.ELEMENT) {
                final NodeValue node = (NodeValue) item;
                writer.write(serializer.serialize(node));
            } else {
                writer.write("<exist:value type=\"");
                writer.write(Type.getTypeName(item.getType()));
                writer.write("\">");
                writer.write(item.getStringValue());
                writer.write("</exist:value>");
            }
``````
I am not sure what the difference between a `Type.ELEMENT` or a `Type.NODE`, but I would say the second option is better and should be used in both situations.

Note that when using the REST API, the structured result is what is obtained. Using the XML-RPC, it depends on what method you call.

(not sure if I made myself clear, also happy for any extra explanation)

### PR DESCRIPTION
**Pull request by ambs**
Just like the REST API, when non-nodes results are returned by the
retrieveAll method, wrap ech one of them with the element `<exist:value>`.
Also, add a type attribute to that element, with the value type.

It fixes #755.

Accordingly with my findings, no code is using retrieveAll. Also, the Java client is retrieving a result at  time (it seems). I will dig that  little deeper, and if it is the case (and if this patch is welcome), I'll change it to using retrieveAll (and hopefully be faster).

### DIFF
--- a/src/org/exist/xmlrpc/RpcConnection.java
+++ b/src/org/exist/xmlrpc/RpcConnection.java
@@ -2170,6 +2170,7 @@ public class RpcConnection implements RpcAPI {
 //			serialize results
             handler.startDocument();
             handler.startPrefixMapping("exist", Namespaces.EXIST_NS);
+            handler.startPrefixMapping("xs", Namespaces.SCHEMA_NS);
             final AttributesImpl attribs = new AttributesImpl();
             attribs.addAttribute("", "hitCount", "hitCount", "CDATA", Integer.toString(qr.result.getItemCount()));
             handler.startElement(Namespaces.EXIST_NS, "result", "exist:result", attribs);
@@ -2178,17 +2179,26 @@ public class RpcConnection implements RpcAPI {
             try {
                 for (final SequenceIterator i = qr.result.iterate(); i.hasNext(); ) {
                     current = i.nextItem();
+
                     if (Type.subTypeOf(current.getType(), Type.NODE)) {
                         current.toSAX(broker, handler, null);
                     } else {
+
+                        final AttributesImpl typeAttr = new AttributesImpl();
+                        typeAttr.addAttribute("", "type", "type", "CDATA", Type.getTypeName(current.getType()));
+                        handler.startElement(Namespaces.EXIST_NS, "value", "exist:value", typeAttr);
+
                         value = current.toString().toCharArray();
                         handler.characters(value, 0, value.length);
+                        
+                        handler.endElement(Namespaces.EXIST_NS, "value", "exist:value");
                     }
                 }
             } catch(final XPathException e) {
                 throw new EXistException(e);
             }
             handler.endElement(Namespaces.EXIST_NS, "result", "exist:result");
+            handler.endPrefixMapping("xs");
             handler.endPrefixMapping("exist");
             handler.endDocument();
             SerializerPool.getInstance().returnObject(handler);

### CODE REVIEW

**Comment by ambs** [comment]
Ah, not sure if your development rules allow that, but it would be great to have this in version 3.0.

**Comment by joewiz** [comment]
I assume part of the motivation is that you're trying to deal with non-element results. Would XQuery 3.1's "adaptive" serialization output method be of use to you? See http://www.w3.org/TR/xslt-xquery-serialization-31/#adaptive-output. It's not supported yet in eXist, but it occurred to me that if this would fill your need, it might be a good motivation for incorporating this. If I'm on the wrong track, apologies.

**Comment by ambs** [comment]
Hmms, probably yes, probably no. I can't understand completely that section of the recommendation.  

A query like `collection("foo")/bar[predicate] ! fn:base-uri(.)` should be considered an array or  sequence of atomic values?

  * If it is considered an array, and therefore, serialized as a JSON array, it might do the trick
  * If, in the other hand, it is a sequence of atomic values, that are serialized as strings, the query would return the strings concatenated, and therefore, raise the same problem.

As another illustrative problem:  `collection(bar)/zbr/text()` should return a concatenated string (at least it is what is done at the moment). And probably with adaptative serialization, it should continue that way. And, how differs this query from the one above?

I would say that in the first one we are returning a sequence of xs:string (or anything like that). In the second, we are returning a sequence of CDATA nodes. This was my interpretation, and the reason I suggested this patch.

(and another reason was to make it coherent with the REST API, so, it might be relevant to include even if adaptative method solved my problem).

**Comment by joewiz** [comment]
I'm going to walk away slowly and suggest that someone with a better understanding of the java internals and these APIs address @ambs's question. (Sorry!)

**Comment by ambs** [comment]
@joewiz, all comments are welcome. I am not aware of the internals details as well. I keep guessing!

**Comment by adamretter** [comment]
The example:
```
collection("foo")/bar[predicate] ! fn:base-uri(.)
```

Would produce a sequence. If you want an array you would need:
```
array { collection("foo")/bar[predicate] ! fn:base-uri(.) }
```

adamretter:
@ambs looks good to me. Thank you very much for your contribution :-)

ambs:
Happy to help :) Thanks

PR #767, opened 2015-08-27
Removes the need of reflection for Loggables

adamretter:
Java 8 provide method references; As the Enumeration of Loggable sub-classes is statically known and their constructors fixed, we can instead use a method reference to the constructor of each Loggable to avoid the expensive use of reflection at runtime.

### DIFF
--- a/src/org/exist/storage/UpdateBinaryLoggable.java
+++ b/src/org/exist/storage/UpdateBinaryLoggable.java
@@ -25,6 +25,7 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -58,11 +59,11 @@ public class UpdateBinaryLoggable extends AbstractLoggable {
     @Override
     public void write(final ByteBuffer out) {
         final String originalPath = original.getAbsolutePath();
-        byte[] data = originalPath.getBytes();
+        byte[] data = originalPath.getBytes(StandardCharsets.UTF_8);
         out.putInt(data.length);
         out.put(data);
         final String backupPath = backup.getAbsolutePath();
-        data = backupPath.getBytes();
+        data = backupPath.getBytes(StandardCharsets.UTF_8);
         out.putInt(data.length);
         out.put(data);
     }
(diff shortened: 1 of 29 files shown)


src/org/exist/storage/UpdateBinaryLoggable.java
```diff
@@ -21,89 +34,78 @@
...
-       }
-    }
-    
-     public String dump() {
-        return super.dump() + " - update "+original+" to "+backup;
+        try(final FileInputStream is = new FileInputStream(backup);
+                final FileOutputStream os = new FileOutputStream(original)) {
+            final byte[] buffer = new byte[4096];
+            int len = -1;
+            while ((len = is.read(buffer)) > -1) {
```
dizzzz: copy of bytes might be simplified with library (if new dependancy is good enough for design)

adamretter: Yes this can be done using `Files.copy`... In-fact this probably is done in https://github.com/eXist-db/exist/pull/768 ;-)

dizzzz:
I could not say why 'LogEntryTypes.addEntryType(LOG_CREATE_PAGE, CreatePageLoggable::new);' is equivalent to 'LogEntryTypes.addEntryType(LOG_CREATE_PAGE, CreatePageLoggable.class);' ; i trust you it is ok :-)

src/org/exist/storage/UpdateBinaryLoggable.java
dizzzz: hehehehe

dizzzz:
for me, both can be pulled in @ljo @wolf

adamretter:
@dizzzz Basically we now have:
```
BiFunction<DBBroker, Long, Loggable> cstr = CreatePageLoggable::new;
Loggable loggable = cstr.apply(broker, transactionId);
```

Instead of the old Java reflection version:
```
addEntryType(TXN_START, TxnStart.class);

public Loggable newInstance(DBBroker broker, long transactId) throws Exception {
  Class<?> constructorArgs[] = { DBBroker.class, long.class };
  final Constructor<? extends Loggable> constructor = clazz.getConstructor(constructorArgs);
  return constructor.newInstance(new Object[] { broker, Long.valueOf(transactId) });
}

newInstance(broker, transactionId);
```

dizzzz:
@adamretter yes I recognize

PR is ok for me

adamretter:
@wolfgangmm Thanks :-)

PR #775, opened 2015-09-01
[bugfix] Small fixes to Backup/Restore process

adamretter:
Under some circumstances after the move to NIO.2 the backup/restore process can fail when it is misused in such a way that it resolves absolute paths against relative paths. There were no tests for such behaviour. Anyway here are some improvements to make it even more robust.

### DIFF
--- a/src/org/exist/backup/Backup.java
+++ b/src/org/exist/backup/Backup.java
@@ -233,7 +233,13 @@ public class Backup
         if(FileUtils.fileName(target).endsWith(".zip")) {
             fWriter = currentName -> new ZipWriter(target, encode(URIUtils.urlDecodeUtf8(currentName)));
         } else {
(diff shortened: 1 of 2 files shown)


ljo:
@adamretter This is also true in NativeBroker.getCollectionBinaryFileFsPath
so it should also check absolute URIs, eg
now
        return fsDir.resolve(uri.getURI().toString());
to
	final String child = uri.getURI().toString();
        return fsDir.resolve(child.charAt(0) == '/' ? child.substring(1) : child);

adamretter:
@ljo Yes I agree that in principle that makes sense. Do you know how I can reproduce the problem though, as we don't seem to have any coverage for that; Ideally we should as it is a NativeBroker issue.

ljo:
@adamretter I used the package manager in the dashboard to deploy packages. Deploy.installAndDeployFromDb

PR #788, opened 2015-09-15
Windows shutdown issues: launch as service

### LINKED ISSUE #713: Windows: Unclean shutdown
On windows 7, 32bit: machine is shut down cleanly, but eXist crashes and reinforces a recovery run upon restart. This used to work in the past, i.e. eXist received the shutdown signal and stopped properly before windows powered off the machine.

Could someone confirm if this happens on other windows versions as well?

wolfgangmm:
With recent Java and Windows versions, the OS is no longer sending a signal to eXist before shutting down, thus causing a crash and a recovery run. The only really reliable method to avoid this is to register eXist as a service. Most new users won't be aware of it though (or do not read the documentation), and get a bad impression.

I have thus extended the desktop launcher to automatically check on startup if eXist has already been installed as a service. If not, the launcher will ask if it should do so. After installation, the service can be controlled from the system tray menu. Advanced users which do not start eXist via the launcher can be expected to know how to install a service.

Tested on win8 and 10. Other operating systems are not affected. Java service wrapper is used to interact with the OS.

Closes #713

### DIFF
--- a/build/scripts/build-impl.xml
+++ b/build/scripts/build-impl.xml
@@ -37,20 +37,20 @@
     <property name="jvmarg" value="-Xmx128000k -Xms32000k -Djetty.home=${jetty.dir} -Dexist.home=${basedir}"/>
     <property name="xmldb.src" value="undefined"/>
     <property name="dist.dir" value="${dist}/${project.name}-${project.version}"/>
-        
+
     <!-- import common targets -->
     <!-- <import file="../../build.xml"/> -->
     <import file="git-support.xml"/>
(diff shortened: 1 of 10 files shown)


dizzzz:
Sorry for the late response

joewiz:
I noticed that the latest version of wrapper mentioned Windows-specific fixes. May be worth investigating too.

joewiz:
Release notes: https://wrapper.tanukisoftware.com/doc/english/release-notes.html#3.5.27

PR #806, opened 2015-10-14
Make sure the IdFunctionTests are stable

adamretter:
Ensures that the test works with all XPath engines including eXist's memtree


dizzzz:
Does the xpath expression potentially return more values? How is that handled?

Looks ok to me afaic judge

adamretter:
@dizzzz no it should always return zero or one values.

PR #813, opened 2015-10-19
Updated to the latest EXPath HTTP Client

adamretter:


joewiz:
You probably want to nix the .DS_Store file from extensions/expath and add to gitignore(s).

extensions/expath/src/org/expath/httpclient/model/exist/EXistResult.java
```diff
@@ -59,11 +59,17 @@
...
+        return new EXistResult(context.copyContext());
+    }
+
     @Override
     public void add(final Reader reader) throws HttpClientException {
-        
-        // START TEMP - replace with a defferred StingReader when eXist has this soon.
+
+        // START TEMP
+        //TODO(AR) - replace with a deferred StingReader when eXist has this soon.
```
joewiz: `Sting` > `String`

adamretter:
@joewiz Thanks Joe, I have fixed those two small issues now.

PR #816, opened 2015-10-20
Close Stream's that use I/O resources

adamretter:
* Solves an issue with eXist leaking file handles, particularly noticeable on `fs.journal` when eXist has been running for a while in production.


dizzzz:
nice....

PR #827, opened 2015-11-07
Fixes for privilege escalation

adamretter:
Fixes an issue with privilege escalation around broker leases, see https://github.com/adamretter/exist/commit/da6b5f38ed7900446ca217658ccaeaadb7809bd3 for details.

The most significant part of the change is https://github.com/adamretter/exist/commit/77a1347f2ea2b7fbda524acf095144fbf2bf0971#diff-a4a35ad924719f7e68d2ecdf372255dd.

The rest of the changes are refactoring to fit the fix.

### DIFF
--- a/extensions/commands/src/org/exist/commands/info/Lock.java
+++ b/extensions/commands/src/org/exist/commands/info/Lock.java
@@ -45,40 +45,33 @@ public class Lock extends AbstractCommand {
 	 */
 	@Override
 	public void process(XmldbURI collectionURI, String[] commandData) throws CommandException {
-
-		Database db = null;
-		DBBroker broker = null;
 		try {
+			final Database db = BrokerPool.getInstance();
 			
-			db = BrokerPool.getInstance();
-			
-			broker = db.getBroker();
-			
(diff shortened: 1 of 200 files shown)


dizzzz:
I need to get a bigger picture of the consequences for extensions that are hosted 'outside' de exist-db repo; whilst nog 3.0 having released we have 'some' freedom still. But compatibility .... remains a point of focus. @ljo please could you check when you have time @wolfgangmm please could you review too, since it is is a significant piece of work; but using the ARMs alone, I like it :-)

adamretter:
@dizzzz I think it highly likely that there will be zero changes for extensions outside of eXist-db that you guys have. If there are any changes... it will at most I think be using `BrokerPool#get` inside a `try-with-resources` and removing the explicit call to `close`/`release`.

dizzzz:
Massive commit, looks good for me (might have missed some details, but in bigger lies it looks OK to me).

Action: I need to re-study the Optional-thing :-)

dizzzz:
Tnx adam.....

PR #828, opened 2015-11-09
Implement XQuery 3.1 arrow operator

wolfgangmm:
Add support for the arrow operator:

``` xquery
('A', 'B', 'C') => count()
```

http://www.w3.org/TR/xquery-31/#id-arrow-operator


joewiz:
Wonderful!

test/src/xquery/xquery3/arrowop.xql
```diff
@@ -0,0 +1,47 @@
...
+declare
+    %test:assertEquals(11)
+function ao:func-in-sequence1() {
+    "Hello" => concat(" world") => string-length()
+};
+
+declare
+    %test:assertEquals("HELLO", "WORLD")
+function ao:func-in-sequence2() {
+    "hello world"=>upper-case()=>normalize-unicode()=>tokenize("\s+")
```
adamretter: The test result does not show that `normalize-unicode` was or was not called. Perhaps a better test case would be:
```xquery
declare
    %test:assertEquals("123", "abc", "dl")
function ao:func-in-sequence2() {
    "123 abc ㎗" => upper-case() => normalize-unicode() => tokenize("\s+")
}
````

**Comment by adamretter** [comment]
The following test-case for the arrow operator fails:

```xquery
function($k, $v) {
	"$k=" || $k
} => map:for-each-entry(
	map {
		"a" : "1",
		"b" : "2",
		"c" : "3"
	},
	?
)
```

adamretter:
Also, I think `map:for-each-entry` was renamed to just `map:for-each` https://www.w3.org/Bugs/Public/show_bug.cgi?id=29269. I will send a separate PR for this later...

joewiz:
Just used the arrow operator for the first time by testing this branch. I can already tell I will be using this a lot.

joewiz:
@adamretter Re: your test case above, where do you see in the spec that `?` can be used to identify which parameter the arrow operator passes in?  As I read it http://www.w3.org/TR/xquery-31/#id-arrow-operator states:

> An arrow operator applies a function to the value of a primary expression, using the value **as the first argument to the function**.

[My emphasis.]

dizzzz:
Oops, still not pulled in? @wolfgangmm

adamretter:
@joewiz Yes that is indeed the case. However, I am using `?` to create a partially applied function, so:

```xquery
  map:for-each(
    map {
      "a" : "1",
      "b" : "2",
      "c" : "3"
    },
    ?
  )
```
Results in an anonymous function, that takes a single parameter, and that is why you can then use the arrow `=>` operator to supply its parameter.

joewiz:
@adamretter Given the status of this issue as described at http://markmail.org/message/3whinktocaznfvr4, do you think this PR can be merged so we can begin using this feature for cases where the value of the primary expression is intended to be used as the first argument to the function. Perhaps your use case could be accounted for once the matter is settled by the WG?

wolfgangmm:
@joe the latest version of the spec indeed changes the operator precedence. we'll have to change the grammar to adopt to this.

joewiz:
Ah, okay! For reference, this topic and a test case is described at https://mailman.uni-konstanz.de/pipermail/basex-talk/2016-January/010050.html.

adamretter:
@joewiz I raised 3 issues, only one of them relates to an unsettled matter, the other 2 are gt genuine errors.

test/src/xquery/xquery3/arrowop.xql
joewiz: Yes, this is definitely better. A small correction: I think "abc" would be upper-cased, so the test case should be

```xquery
declare
    %test:assertEquals("123", "ABC", "dl")
function ao:func-in-sequence2() {
    "123 abc ㎗" => upper-case() => normalize-unicode('NFKC') => tokenize("\s+")
}
```

dizzzz:
a bit quiet here...... for 3.0 or 3.1 ?

joewiz:
I'd love to see this make it into 3.1 if possible? It seems so close!

wolfgangmm:
@joewiz I'm trying to finish it. Test cases pass, but need to align the grammar.

adamretter:
@wolfgangmm cool. I will give it a test after breakfast and all being well I intend to merge it :-)

adamretter:
The first test case (1) I posted:

```xquery
xquery version "3.1";
declare namespace map = "http://www.w3.org/2005/xpath-functions/map";

function($k, $v) {
    "$k=" || $k
} =>
    map:for-each(
        map {
            "a" : "1",
            "b" : "2",
            "c" : "3"
        },
        ?
    )()
```

still fails, although whether it is valid of not is something I should clarify with the XQuery WG. I am happy to report that test cases (2) and (3) that I posted now pass.

I will manually resolve the conflicts and merge...

adamretter:
@wolfgangmm @shabanovd Just a quick note to say, that when you make changes to the XQuery parser, if you can isolate the generated grammar changes to the final commit, then it makes resolving conflicts when people are both working on the parser much much simpler. For example: https://github.com/eXist-db/exist/commit/c9abbef

adamretter:
@wolfgangmm Thank you very much for finishing this, it is a great addition :-)

joewiz:
Agreed! Thank you! This is a great addition to XQuery and eXist!

PR #834, opened 2015-11-18
IP range based authentication

wshager:
Checks for iprange entries under accounts in /db/system/security/IPRange/accounts. If a range is found, the corresponding exist user is authenticated. Account XML format:

``` xml
<account>
    <name>test</name>
    <iprange>
        <start>InetAddress-to-long</start>
        <end>InetAddress-to-long</end>
    </iprange>
 </account>
```

``` java
import java.net.*;

private static long ipToLong(InetAddress ip) {
    byte[] octets = ip.getAddress();
    long result = 0;
    for (byte octet : octets) {
        result <<= 8;
        result |= octet & 0xff;
    }
    return result;
}
```


wshager:
1) Doesn't seem to build against current develop
2) If tests are required, please let me know

joewiz:
Are you going to be able to update the PR yourself, or do you need help to try to get this to build under develop? If the latter, what version does it build against?

wshager:
@joewiz I'm waiting for a reply from @adamretter. It should build against https://github.com/eXist-db/exist/commit/8f7c6898fc65d019718bf144e77a2f27e6408034

dizzzz:
according to travis the build is broken. @adamretter  wouter asked you to step in.....

wshager:
@dizzzz This currently builds and works for my use case. Note that I updated the flag to disable the realm by default.

dizzzz:
So the intention is really that if one is from a ip-range, a specific userid is assigned?

wshager:
Yes, it's a way for known institutions to authenticate.

dizzzz:
ok. I'd recommend atleast not to do this for the admin account or accounts with DBA rights..... quite a risk, more risky than enabling the "XQueryStartupTrigger" in configuring.xml ; 

when 'admin' is required (...) then i'd recommend, at least, for obvious reason to check the document permissions before using it (just like done in the XQueryStartupTrigger). It must be prevented that these  come an attack vector. 

I reworked your code (without the permission check) in https://github.com/dizzzz/exist/tree/wshager-iprange ; I found some potential bugs in your code (making the code probably not work correctly)

wshager:
Indeed, it's not my intention to map to dba accounts, and that may well be prevented. Thanks for your updates, I'll review them tomorrow.

dizzzz:
In my branche, the code has been made compatible with the most recent 3.x (to be released before/around XML Prague)

dizzzz:
My version should be OK, this version is not 3.0 compatible (and has some stability issues)

wshager:
@adamretter @dizzzz sorry, yes it's tested and works for the use case. I don't experience any issues.

wshager:
Indeed, it wasn't tested against 3.0 yet.

wshager:
I'm waiting for #1305 to be solved until I continue.

dizzzz:
the reworked version is here: https://github.com/dizzzz/exist/tree/wshager-iprange I did not test it though...

wshager:
I'm closing this until I can PR @dizzzz version. I already merged it, but my use case awaits 3.x stable...

Issue #835, opened 2015-11-18
NPE when uploading a file to replace an existing file over WebDAV


joewiz:
Steps to reproduce: Using the current develop branch, upload any XML document (e.g., `test.xml`) to the db (e.g., `/db/test.xml`) via WebDAV. This works. Upload it a second time: error 500, with exist.log showing the following error:

```
2015-11-18 10:59:01,434 [eXistThread-41] INFO  (MiltonResource.java [authorise]:351) - PUT /db (write=true) 
2015-11-18 10:59:05,938 [eXistThread-42] INFO  (HttpManager.java [process]:213) - PUT :: http://localhost:8080/exist/webdav/db/test.xml - http://localhost:8080/exist/webdav/db/test.xml 
2015-11-18 10:59:05,939 [eXistThread-42] ERROR (StandardFilter.java [process]:68) - process 
java.lang.NullPointerException
	at java.util.Objects.requireNonNull(Objects.java:203) ~[?:1.8.0_45]
	at java.util.Optional.<init>(Optional.java:96) ~[?:1.8.0_45]
	at java.util.Optional.of(Optional.java:108) ~[?:1.8.0_45]
	at org.exist.webdav.ExistDocument.getCurrentLock(ExistDocument.java:298) ~[exist-webdav.jar:?]
	at org.exist.webdav.MiltonDocument.getCurrentLock(MiltonDocument.java:503) ~[exist-webdav.jar:?]
	at com.bradmcevoy.http.HandlerHelper.isLockedOut(HandlerHelper.java:152) ~[milton-api-1.8.1.3.jar:?]
	at com.bradmcevoy.http.http11.PutHandler.process(PutHandler.java:109) ~[milton-api-1.8.1.3.jar:?]
	at com.bradmcevoy.http.StandardFilter.process(StandardFilter.java:52) [milton-api-1.8.1.3.jar:?]
	at com.bradmcevoy.http.FilterChain.process(FilterChain.java:40) [milton-api-1.8.1.3.jar:?]
	at com.bradmcevoy.http.HttpManager.process(HttpManager.java:228) [milton-api-1.8.1.3.jar:?]
	at com.bradmcevoy.http.MiltonServlet.service(MiltonServlet.java:195) [milton-servlet-1.8.1.3.jar:?]
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:684) [jetty-servlet-8.1.17.v20150415.jar:8.1.17.v20150415]
[...]

PR #841, opened 2015-11-28
[bugfix] Address an NPE issue in WebDAV introduced by bce696e

adamretter:
Closes https://github.com/eXist-db/exist/issues/835


joewiz:
I've tested this branch and confirmed it fixes the issue.

PR #843, opened 2015-11-30
Fix NPE in app deployment; add tests to avoid regressions

wolfgangmm:


joewiz:
Thanks, @wolfgangmm and @adamretter!

PR #867, opened 2015-12-29
[bugfix] Fix the reported error location for functions called dynamic…

wolfgangmm:
…ally. The line/column indicated in the exception were misleading when using dynamic function calls. For example, if the function item was retrieved by calling fn:lookup-function, the error message would point to where lookup-function was evaluated, not the actual dynamic function call.

This makes it hard to work with libraries relying on dynamic function calls.


adamretter:
@wolfgangmm Any chance of a test case to prevent regressions here?

joewiz:
This is so exciting. Can't wait to try this against cases I saw previously, e.g., https://github.com/eXistSolutions/hsg-shell/issues/61#issuecomment-165164880.

PR #870, opened 2016-01-03
Bump exist version to upcoming RC2

dizzzz:
Relabel internal RC2 to 3.0.2 [this is not visible anyway to users] ; this is highly required to get my extensions to work ; there is a long standing conflict with my extensions due to API changes between RC1 and RC2 ;
"3.0.2" helps me/us to continue developing extensions, without over and over again explaining our users that RC2 extension are not perse compatible with RC1. This unclearness + conflict is bad for us!
As discussed in weekly telco. Please pull-in ASAP. And please don't use 'puristic' arguments, this is the best and pragmatic solution.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -4,8 +4,8 @@
 #
 # $Id$
 project.name = eXist-db
-project.version = 3.0.RC1
-project.version.numeric = 3.0.0
+project.version = 3.0.RC2
+project.version.numeric = 3.0.2
 project.source = develop
 project.codename = Berlin
 


adamretter:
@dizzzz As mentioned before. I really think this will cause problems in the near future...

dizzzz:
@adamretter please could you elaborate on your concerns? it is an internal numbering, really

as I said, since some of your API changes, my extensions are useless, I am explaining over and over again which extension is working on which version. People are actually using RC2 already, some are not.

i have been blocked for several MONTHS by the changes mentioned; now I can finally proceed. 

this changes has been discussed with @wolfgangmm and @ljo and is the best and pragmatic solution to continue development.

adamretter:
@dizzzz I think the number is not just used internally. If you ask eXist which version it is, then that is what is reported. There were only two solutions to the problem that I could see, 1) Fix the versioning in EXPath Pkg so it supports something more sensible than just `x.y.z`, i.e. `x.y.z-q`, 2) Do a release of eXist, we really should have done that months and months ago anyway.

wolfgangmm:
1) Changing the expath version numbers would break compatibility with the spec, so I did not consider it; 2) develop was definitely not ready for production until maybe 8 weeks ago, so releasing it was impossible (too many changes after RC1 breaking things). There is still one major showstopper which needs to be resolved (@ljo's whitespace handling fix!!!) before release.

adamretter:
@wolfgangmm, hmm... I think the EXPath spec could have been been easily revised; This problem will almost certainly occur for the next release also, so maybe better to fix the route cause!

Also, I know of more than one serious issue before an RC2..., e.g. https://github.com/eXist-db/exist/pull/637

PR #879, opened 2016-01-10
Various improvements to XPath 3.1 Map functions

adamretter:
1. Corrected the function name `map:for-each-entry` to `map:for-each`
2. Removed `map:new` in favour of `map:merge`
3. Implemented missing function `map:size`
4. Implemented missing function `map:put`

We should now have the correct functions from XPath 3.1 for maps.

### DIFF
--- a/src/org/exist/xquery/functions/map/AbstractMapType.java
+++ b/src/org/exist/xquery/functions/map/AbstractMapType.java
@@ -44,6 +44,8 @@ public abstract class AbstractMapType extends FunctionReference
 
     public abstract Sequence get(AtomicValue key);
 
+    public abstract AbstractMapType put(AtomicValue key, Sequence value) throws XPathException;
+
     public abstract boolean contains(AtomicValue key);
 
     public abstract Sequence keys();
--- a/src/org/exist/xquery/functions/map/MapFunction.java
+++ b/src/org/exist/xquery/functions/map/MapFunction.java
@@ -10,154 +10,182 @@ import java.util.Map;
  * Implements all functions of the map module.
  */
 public class MapFunction extends BasicFunction {
-    
-    public static final FunctionSignature[] signatures = {
-        new FunctionSignature(
-            new QName("new", MapModule.NAMESPACE_URI, MapModule.PREFIX),
-            "Constructs and returns an empty map whose collation is the default collation in the static context.",
-            null,
-            new SequenceType(Type.MAP, Cardinality.EXACTLY_ONE)),
-        new FunctionSignature(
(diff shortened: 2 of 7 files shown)


wolfgangmm:
Merging this will break nearly any app out there, thus further delaying an RC2. Could you re-add deprecated signatures for map:for-each-entry and map:new, which just seem to have changed names?

adamretter:
@wolfgangmm Why not just merge it for RC3 instead?

wolfgangmm:
Well, the problem would be the same: many user-written apps, which would otherwise run fine on 3.0, will break. We have this deprecated flag, so why not add alternatives for map:for-each-entry/map:for-each and map:new/map:merge (with one argument). As far as I can see this would cover most use cases and provides an easy migration path without causing unnecessary stress to users.

adamretter:
@wolfgangmm Well my feeling is that the map functions were not in 2.2, and so if we are going to put them in the next release i.e. 3.0, then we should make sure they are correct. I don't see the logic in adding new functions which are deprecated on inception. If some people have been developing their apps against HEAD, then I believe you and @dizzzz derived a versioning scheme to help them?

If you are worried about the eXist Dashboard, eXide and Monex apps, I am happy to do a find-and-replace and send PRs to you; Although I note there are some PRs against Monex from me that are now almost 1 year old (if you could take a look please).

dizzzz:
I'd like to suggest the 'middle': have the deprecated functions, and remove them after 3.0 asap;

Additionally i'd like to suggest two items:
- let's 'pollute' the log when people use deprecated functions
- do not show deprecated functions i the xquery functions app and eXide ; this will help to force to switch?

dizzzz:
I mean, this is what we really should do because we implement functions ahead of the spec; we should not discourage early adaptors stop early adapting :-)

joewiz:
+1 to @dizzzz's proposal. I think it builds a virtuous cycle and sets a good precedent that could be applicable in other scenarios too: it continues to encourage implementors (like eXist) to be early adopters of draft specs (especially the really valuable parts that users are asking for), encourages all users of draft-era specs to dive in and use/test the implementations despite draft status, and, as the spec reaches final status, feeds pushes all users to get into line.

adamretter:
@dizzzz @joewiz @wolfgangmm My reasoning for not adding deprecated versions was that I didn't think these Map functions were in 2.2. I see that I was mistaken, I just rebased to incorporate the deprecated versions as requested. Hopefully should be all good now?

wolfgangmm:
Thanks @adamretter! I'll test it on some installs tomorrow and merge it afterwards.

dizzzz:
Nice.....

PR #883, opened 2016-01-11
[bugfix] Close memory leak: cleanupTasks in XQueryContext need to be …

wolfgangmm:
…cleared before query is pushed to pool, otherwise the list will keep growing with every call to the query and resources are not freed. Mainly used from http client.


adamretter:
@wolfgangmm Nice catch :-) Just wondering if it should actually go into `XQueryContext#reset` with the other cleanup instead?

wolfgangmm:
The effect would probably be the same if we move the one line into XQueryContext#reset, but since runCleanupTasks is called separately, I found it more appropriate to do the post-cleanup there.

joewiz:
Great! Thanks so much for tracking down this memory leak!

PR #884, opened 2016-01-11
Inline variables

wolfgangmm:
1. inline global variables if they are sequences of literals. Speeds up complex modules using a lot of constants everywhere
2. statically resolve qnames for variable declarations and references at compile time. Saves a lot of time in variable lookups at runtime


dizzzz:
Are there any functional testcases possible?

adamretter:
@wolfgangmm Sounds really nice :-) I will do some testing and merge if all looks good. Just one question (inline) about the new variable bindings...

wolfgangmm:
The err:* variables within try/catch are actually local variables and have to be declared for the static analysis to pass. Previously the analyze just ignored unknown vars.

PR #892, opened 2016-01-14
[ignore] Misc settings for IntelliJ, shouldn't upset anyone

adamretter:


dizzzz:
These fiets reoccur over n over....

PR #893, opened 2016-01-14
Unifying the asocat lib handling. 

ljo:
Changing to correct name for jar file, now only in one place. Makes ant only builds happy.

### DIFF
--- a/build/scripts/build-impl.xml
+++ b/build/scripts/build-impl.xml
@@ -49,7 +49,7 @@
     <available file="key.store" property="key.store.present"/>
 
     <!-- additional set of ant tasks -->
-    <property name="asocat-exist.jar" location="${tools.ant}/lib/asocat-exist.jar"/>
+    <property name="asocat-exist.jar" location="${tools.ant}/lib/asocat-exist-1.0.jar"/>
 
     <path id="classpath.external">
         <fileset dir="${lib.core}">
--- a/build/scripts/installer.xml
+++ b/build/scripts/installer.xml
@@ -84,7 +84,7 @@
         
         <taskdef name="fetch" classname="nl.ow.dilemma.ant.fetch.FetchTask">
             <classpath>
-                <pathelement location="tools/ant/lib/asocat-exist.jar"/>
+                <pathelement location="${asocat-exist.jar}"/>
             </classpath>
         </taskdef>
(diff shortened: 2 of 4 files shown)


adamretter:
@ljo I think this is some tools that @dizzzz wrote to assist with the Ant build, right? As this wasn't in the classpath but the build was still working fine... do we actually still use these tools?

ljo:
@adamretter Yes, we are using it for all extensions, the jarsigner, installer and for the macapp build.

adamretter:
Erm... but I am doing a build with all extensions and the installed and the macapp without your PR and it seems to work?

ljo:
@adamretter yes, that is probably becuase you are not using ant only.

adamretter:
@ljo I am using eXist's `./build.sh` does that make a difference?

ljo:
@adamretter yes, that works flawlessly for me too.

adamretter:
@ljo so you mean invoking an external `Ant` binary?

dizzzz:
afaik the ascot.jar is required to be able to build eXist-db ; note that there are some smart handles that prevent double download of jar files. It is e.g. used to download XAR files for the auto deploy directory

ljo:
@adamretter Yes, when run with any ant binary it actually reads the given paths for the tasks and since the jar name is wrong the antclasspath gets empty.

build/scripts/build-impl.xml
```diff
@@ -49,7 +49,7 @@
     <available file="key.store" property="key.store.present"/>
 
     <!-- additional set of ant tasks -->
-    <property name="asocat-exist.jar" location="${tools.ant}/lib/asocat-exist.jar"/>
+    <property name="asocat-exist.jar" location="${tools.ant}/lib/asocat-exist-1.0.jar"/>
```
dizzzz: so the jar needs to be renamed too? hmmm no it i already called 1.0?

adamretter:
@ljo okay cool. Looks good then.

ljo:
@dizzzz no, the jar file (tools/ant/lib/asocat-exist-1.0.jar) already has the name i changed to in the build-impl.xml since it doesn't match otherwise.

ljo:
@adamretter ok, great.

PR #901, opened 2016-01-24
Upgrade of commons-compress library 

dizzzz:
needed for the XZ compression


adamretter:
PR is fine but as far as I am aware we don't use XZ compression anywhere... do we?

dizzzz:
correct, but I need it for the messaging/replication extension

Issue #255, opened 2014-06-30
move and copy triggers not firing


phattire:
It appears the move document (before and after) and copy (before and after) document are not working in eXist 2.1.  (We have not verified move/copy collection triggers.)  We are not having issues with the other document triggers.  Our collection.xconf trigger definition does not call out which trigger events to fire on.

gspringtech:
The copy- and move-document triggers are documented as having 2 parameters (old-uri,new-uri) - you can see this in the example code above and in the documentation. In fact, these should only have ONE parameter. I have read XQueryTrigger.java (from about 729) and successfully tested the move-document trigger with one parameter. 

So - the 'bug' is in the documentation - not the code.

gspringtech:
Further to the previous comment, the after-move-document is receiving the OLD URI, not the new. This needs to be fixed in the source - either by adding the NEW uri (to match the documentation) or changing the uri from old to new.

gspringtech:
And finally, the after-copy-collection parameters are in the wrong order in the documentation - they should be (old-url, new-url).

PR #904, opened 2016-01-28
Fixed the trigger functionality

ChiSamurai:
- fixes https://github.com/eXist-db/exist/issues/255
- PR for updating the documentation here: https://github.com/eXist-db/documentation/pull/72

### DIFF
--- a/src/org/exist/collections/triggers/XQueryTrigger.java
+++ b/src/org/exist/collections/triggers/XQueryTrigger.java
@@ -646,11 +646,11 @@ public class XQueryTrigger extends SAXTrigger implements DocumentTrigger, Collec
 	}
 
 	@Override
-	public void afterCopyCollection(DBBroker broker, Txn txn, Collection collection, XmldbURI newUri) throws TriggerException {
+	public void afterCopyCollection(DBBroker broker, Txn txn, Collection collection, XmldbURI oldUri) throws TriggerException {
 		if (events.contains(TriggerEvents.EVENTS.COPY_COLLECTION)) {
-			finish(5, broker, txn, collection.getURI(), newUri, true);
+			finish(5, broker, txn, oldUri, collection.getURI(), true);
 		} else {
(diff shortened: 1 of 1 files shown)


shabanovd:
Is it possible to keep order as it was (new-url, old-url)?

ChiSamurai:
Sure, but Joe and Alister proposed to make this consistent for "before" and "after" trigger. Since the relevant triggers did not work at all and all the trigger functionality broke after using them this change imo will not affect existing usage.

adamretter:
@ChiSamurai I am happy for it to change if it means we get a better API

shabanovd:
@adamretter the problem that compatibility will be broken ... again ... -(

shabanovd:
@ChiSamurai the idea was to keep 1st parameter in-storage resource location and 2nd informative (what will be or what was).

The xquery interface also was in-line with java interface https://github.com/eXist-db/exist/blob/develop/src/org/exist/collections/triggers/CollectionTrigger.java#L55 or https://github.com/eXist-db/exist/blob/develop/src/org/exist/collections/triggers/DocumentTrigger.java#L103

ChiSamurai:
hmm... I just tried to change the output order. I don't really understand the complete code but it seems that there are some dependecies on the  (old-url, new-url) for any of these triggers. Every try to change the order broke the trigger functionality completely.

ChiSamurai:
But if compatibility is affected, it would surely be good to change this... maybe somebody more trained in Java can have a look? I got a xquery unit-test which fires each trigger once which I can provide.

shabanovd:
@ChiSamurai can you commit? and I'll review it

ChiSamurai:
@shabanovd The test contains three files: the collection.xconf, the trigger xquery functions itself and the xquery with collection and a xquery script which does the collection and resource operations.  Can I just write you an email? Or maybe skype or jabber?

shabanovd:
@ChiSamurai sure

shabanovd:
fixed by https://github.com/eXist-db/exist/pull/910

PR #910, opened 2016-02-09
Collection trigger fix

shabanovd:


src/org/exist/collections/CollectionCache.java
```diff
@@ -55,6 +55,9 @@ public void add(Collection collection) {
     }
 
     public void add(Collection collection, int initialRefCount) {
+        // don't cache the collection during initialization: SecurityManager is not yet online
+        if(pool.isInitializing()) return;
```
dizzzz: should there be a log/debug line?

src/org/exist/collections/triggers/XQueryTrigger.java
```diff
@@ -731,16 +707,16 @@ public void beforeCopyDocument(DBBroker broker, Txn txn, DocumentImpl document,
...
 			prepare(4, broker, txn, document.getURI(), newUri, false);
 		} else {
-		    execute(true, broker, txn, beforeCopyDocument, document.getURI());
+            execute(true, broker, txn, beforeCopyDocument, document.getURI(), newUri);
 	    }
 	}
 
 	@Override
-	public void afterCopyDocument(DBBroker broker, Txn txn, DocumentImpl document, XmldbURI newUri) throws TriggerException {
+    public void afterCopyDocument(DBBroker broker, Txn txn, DocumentImpl document, XmldbURI oldUri) throws TriggerException {
```
dizzzz: so the name of the parameter has been wrong, always? Or do I need to update my plugins that rely on the interface?

dizzzz:
good to see a test, most of the code looks straight forward;
I cannot judge on the 'actual' changes though. @adamretter @wolfgangmm @ljo ?

src/org/exist/collections/CollectionCache.java
shabanovd: it wasn't before, why now?

src/org/exist/collections/triggers/XQueryTrigger.java
shabanovd: if you implement base on interface then names is same, this implementation had typos

ChiSamurai:
Thanks again for your work. Really nice to finally see the triggers running as intended. This fixes https://github.com/eXist-db/exist/issues/255

PR #914, opened 2016-02-11
java.net.URI picky about what is absolute

hungerburg:
Schema "xmldb:exist:" does not do it, "file:" does -- should be fine, as it seems to be dropped anyways.

### DIFF
--- a/extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
+++ b/extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
@@ -35,7 +35,7 @@ public class ApacheFopProcessorAdapter implements ProcessorAdapter {
                 final EnvironmentProfile environment = EnvironmentalProfileFactory.createDefault(defaultBaseURI, null);
                 builder = new FopFactoryBuilder(environment).setConfiguration(cfg);
             } else {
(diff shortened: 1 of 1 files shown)


hungerburg:
When running the transform that uses hyphenation in FO, an exception will be thrown with current develop. Below top of the trace
> Caused by: java.lang.IllegalArgumentException: URI is not absolute
> 	at java.net.URI.toURL(URI.java:1088)
> 	at org.apache.fop.apps.io.ResourceResolverFactory$NormalResourceResolver.getResource(ResourceResolverFactory.java:224)

Java doc says "An absolute URI specifies a scheme;" "An opaque URI is an absolute URI whose scheme-specific part does not begin with a slash character ('/')." To me that reads, that "xmldb:exist:///db" is an opaque URI; possibly that is why it fails? https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#toURL%28%29 probably means, absolute AND NOT opaque?

Relevant code in FOP http://grepcode.com/file/repo1.maven.org/maven2/org.apache.xmlgraphics/fop/2.0/org/apache/fop/apps/io/ResourceResolverFactory.java#222

adamretter:
@hungerburg, ah ok. Well I don't have a better solution than yours :-)

hungerburg:
I guess, that the resulting URL will not be absolute then; there is no "opaque" URL; the scheme there is called protocol, and the first colon starts the path, which will not be an absolute path for an opaque URI; probably that is the reasoning behind it?

PR #920, opened 2016-02-22
also truncate an already created file

### LINKED ISSUE #906: file:serialize prepends content to file
In current develop file:serialize inserts the serialized content into file. Is this the expected behavior or a bug?

wshager:
This should fix #906


adamretter:
Thanks @wshager

joewiz:
Yes, thanks so much @wshager! I just ran into this problem and am glad there's now a fix. Thanks too for reviewing @adamretter.

PR #927, opened 2016-03-04
Fixes to read-only state control of the database

adamretter:


dizzzz:
in essence the PR looks good to me ; a little concern about not re-checking diskspace when database is readonly due to lack of diskspace.

adamretter:
@dizzzz I was just trying to fix what we have rather than implement a new feature ;-)

adamretter:
@dizzzz Once eXist switches into read-only mode, it can only ever be switched out of read-only mode by restarting it. That has always been the case. You can see that in the existing code base here: https://github.com/eXist-db/exist/blob/develop/src/org/exist/storage/BrokerPool.java#L1323

Basically `isReadOnly` is never set back to `false`.

There is a good reason for this. When the database switches to read-only, we prevent new writes coming in, however we would still flush any pending buffers on the regular database sync. We have no way of knowing if there is enough disk space to flush those buffers, as such this could cause a corruption. If we were to then switch back to read/write mode, we would just compound the problem caused by a corruption if one had occurred. Instead we require the user to restart the db so that the Consistency Check at startup will check for such things.

adamretter:
@dizzzz I have rebased now where I have switched to using NIO FileStore for obtaining the Usable Disk Space... There is not much more we can do than that.

dizzzz:
ok, thnx for the comments!

Issue #845, opened 2015-12-03
Use of "order by" in recursively called function drops NPE


ChiSamurai:
There seems to be an issue with the use of "order by" in an recursively called function, leading to a NPE:

> 2015-12-03 10:59:21,512 [eXistThread-123] WARN  (ServletHandler.java [doHandle]:563) - /exist/apps/tamboti/modules/search/index.html 
> javax.servlet.ServletException: An unknown error occurred: null
>   at org.exist.http.servlets.EXistServlet.doPost(EXistServlet.java:460) ~[exist-optional.jar:?]
>   at javax.servlet.http.HttpServlet.service(HttpServlet.java:755) ~[servlet-api-3.0.jar:?]
>   at javax.servlet.http.HttpServlet.service(HttpServlet.java:848) ~[servlet-api-3.0.jar:?]

[...]

> Caused by: java.lang.NullPointerException
>   at org.exist.xquery.OrderByClause.eval(OrderByClause.java:53) ~[exist.jar:?]
>   at org.exist.xquery.LetExpr.eval(LetExpr.java:111) ~[exist.jar:?]
>   at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71) ~[exist.jar:?]
>   at org.exist.xquery.ForExpr.processItem(ForExpr.java:238) ~[exist.jar:?]
>   at org.exist.xquery.ForExpr.eval(ForExpr.java:178) ~[exist.jar:?]
>   at org.exist.xquery.LetExpr.eval(LetExpr.java:111) ~[exist.jar:?]
>   at org.exist.xquery.UserDefinedFunction.eval(UserDefinedFunction.java:155) ~[exist.jar:?]
>   at org.exist.xquery.FunctionCall.evalFunction(FunctionCall.java:330) ~[exist.jar:?]
>   at org.exist.xquery.FunctionCall.eval(FunctionCall.java:234) ~[exist.jar:?]
>   at org.exist.xquery.LetExpr.eval(LetExpr.java:99) ~[exist.jar:?]

[...]

See full exception quoted here: http://pastebin.com/V9pNFiiL

This code will show the problem:

``` xquery
xquery version "3.0";

declare function local:recursive($collection) {
    let $subcollections := xmldb:get-child-collections($collection)
    for $subcol in $subcollections
        order by $subcol
    return 
        let $fullpath := xs:anyURI($collection || "/" || $subcol)
        return
            <col name="{$fullpath}">
                {local:recursive($fullpath)}
[...]

PR #930, opened 2016-03-07
Fix recursive calls to group by and order by.

wolfgangmm:
Both clauses need to maintain state between calls to eval() because they are collecting tuples. When used inside a recursive function, state may get overwritten (see the included tests). We thus need to keep state information in separate objects and push them to a stack between calls.

Closes issue #845

### DIFF
--- a/src/org/exist/xquery/GroupByClause.java
+++ b/src/org/exist/xquery/GroupByClause.java
@@ -1,6 +1,5 @@
 package org.exist.xquery;
 
-import org.apache.xerces.impl.xpath.XPath;
 import org.exist.dom.QName;
 import org.exist.xquery.util.ExpressionDumper;
 import org.exist.xquery.value.*;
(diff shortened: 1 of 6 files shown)


src/org/exist/xquery/GroupByClause.java
```diff
@@ -1,5 +1,6 @@
 package org.exist.xquery;
 
+import org.apache.xerces.impl.xpath.XPath;
```
dizzzz: Is this the correct import ? From xerces?

wolfgangmm: Oh, not sure where this came from. Removing it...

ljo:
Except for Adam's questions about Stack vs ArrayDeque and his statement about Tuple missing generic parameters I would say go for pull.

wolfgangmm:
Sorry, @ljo, I just managed to overwrite your comment instead of answering to it. Got up too early today it seems.

Well, if someone would remind me how to fix the issue with generic parameters on Tuple, I could quickly change it...

ljo:
Should be just Tuple<QName, Sequence>

adamretter:
@wolfgangmm No problem, I fixed it up for you and merged.

PR #933, opened 2016-03-08
Improve %latest% in start.config

adamretter:
When running `client`/`jetty` from `org.exist.start.Main` eXist did not recognised the new log4j2 jar files as it's definition of `%latest%` in `org/exist/start/start.config` was too narrow.

### DIFF
--- a/src/org/exist/client/NewResourceDialog.java
+++ b/src/org/exist/client/NewResourceDialog.java
@@ -267,10 +267,8 @@ public class NewResourceDialog extends JFrame {
     private void createResource(final ResourceType resourceType, final String filename, final String moduleNamespace, final String moduleNamespacePrefix) {
         
         final StringBuilder resourceContentBuilder = new StringBuilder();
-        Reader reader = null;
-        try {
-            final InputStream is = getClass().getResourceAsStream(resourceType.getTemplatePath());
-            reader = new InputStreamReader(is);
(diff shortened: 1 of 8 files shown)


ljo:
@windauer I have seen a lot of these "in progress" labels being inserted at the same time the request comes in, so my conclusion is this automatically inserted and not that you are doing it? Can I pull this or are you actually looking?

dizzzz:
if the 'in progress' has additional semantics, it is not explained nor discussed on exist-develop or exist-open ...

ljo:
@dizzzz so you mean I should not pull it since the sematics we agreed on,  just to make it clear to others, is "I am doing the review thus it is in progress" or?

dizzzz:
@ljo no, I think you can pull it in, the code looks good....

ljo:
I still think this automatic labeling is not a good change. So if someone can explain and convince me why something automatically can be in progress I would be happier.

windauer:
ehm, I'm a little bit confused.. now it say I removed the "in progress" label a minute ago, but I didn't do anything here. Joern and me tried to find a better view for Github issue or to connect it with Trello. Perhaps something wen't wrong there.. I will have a look.. sorry for the noise if it was me :)

ljo:
@windauer Ok. Yes, this is new behaviour from latest week which should be changed then. ;)

windauer:
I removed 3 of 4 webhooks from the eXist git repo. Should be fine again now

PR #938, opened 2016-03-14
Make sure that URLs inside a FO are correctly resolved

adamretter:
We now support:

* http
* https
* file
* exist

URL schemes from inside FO. You can use `exist://localhost/db` to refer to the root collection of the local database.

Note, Apache FOP does not seem to respect it's `font-base` config parameter, and so you will need to use absolute paths for each font that you wish to embed, see: http://fop-users.markmail.org/search/?q=#query:+page:1+mid:gmsajwfwr2la7qxu+state:results

Should improve on https://github.com/eXist-db/exist/pull/912 and https://github.com/eXist-db/exist/pull/914.

@hungerburg Any comments on this?

### DIFF
--- a/extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
+++ b/extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
@@ -8,12 +8,21 @@ import java.util.Properties;
 import org.apache.avalon.framework.configuration.Configuration;
 import org.apache.avalon.framework.configuration.SAXConfigurationHandler;
 import org.apache.fop.apps.*;
+import org.apache.fop.apps.io.ResourceResolverFactory;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
+import org.apache.xmlgraphics.io.ResourceResolver;
+import org.apache.xmlgraphics.io.URIResolverAdapter;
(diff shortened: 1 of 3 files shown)


joewiz:
This is great. I've struggled with this before. A possibly better link to your message to fop-users: http://markmail.org/message/gmsajwfwr2la7qxu.

One question: is `exist` an established URL scheme? I've only seen `xmldb:exist` - as awkward as that is.

adamretter:
@joewiz Nope, `exist` URI scheme is only used here. The problem is that FOP uses `java.net.URI`, and `xmldb:exist:///` does not work correctly with java.net.URI#resolve, so here I had to devise something else.

dizzzz:
the `exist:` schema is a good alternative, since the double-scheme is a problem over and over again.

dizzzz:
looks good to me

hungerburg:
The schema "exist:" is not a widely known schema, I fear that using it in the FopFactoryBuilder call might spoil the loading of other fop-extensions (especially I think of hyphenation)

hungerburg:
Good, so I looked at #914 ; patched my local eXist to make FopFactoryBuilder use "new URI(exist://localhost/db/)" as its argument  and IT WORKED, i.e. it did get accepted as an absolute URI :) So good to go from me too!

dizzzz:
@wolfgangmm ok to pull in?

PR #939, opened 2016-03-14
Make the Serializer's chain of receivers configurable at runtime

shabanovd:


adamretter:
Hey @shabanovd Can you tell me what this is for as you didn't provide any description above. Looking through the code, I cannot really see any issues... however I cannot see that this code is ever used? i.e. `DBBroker#newSerializer(List<String> chainOfReceivers)` is never called from anywhere?

shabanovd:
@adamretter That change allow to define chain of receivers (filters) runtime. Right now, there are set at conf.xml by custom-filter tags only.

adamretter:
@shabanovd heh yes I understood that much. However there is nothing in eXist that uses this new code you have introduced right?

adamretter:
@shabanovd Also two things just occurred to me:

1) `NativeBroker` keeps a cached serializer accessible from, i.e `NativeBroker#getSerializer()`. DBBrokers reside in an object pool which is reused for different requests from different threads, so I would have thought that dynamically setting this could cause strange behaviour across requests as the serializer is initialised by one request but reused by others.

2) The other I will comment on inline...

shabanovd:
@adamretter I think you miss main point. Different chain can be required for publishing need outside of eXist and controlled by none eXist code. That's why there are no use at eXist code and new `NativeBroker.newSerializer` was added. There is no other option to control processing chain, that why I add that API.

`NativeBroker.getSerializer()` have no relations with `NativeBroker.newSerializer`.

To avoid complex modification I did not cache custom serializer on eXist side assuming that external code will/can handle it if required.

PR #943, opened 2016-03-15
[bugfix] Fix an issue where we were leaking file handles and memory.

adamretter:
BinaryValue's were not closed after an XQuery was executed more than once (i.e. if the XQueryContext was recycled).

I think this one must have been around for quite some time.

### DIFF
--- a/src/org/exist/xquery/XQueryContext.java
+++ b/src/org/exist/xquery/XQueryContext.java
@@ -3537,27 +3537,35 @@ public class XQueryContext implements BinaryValueManager, Context
     @Override
     public void registerBinaryValueInstance(final BinaryValue binaryValue) {
         if(binaryValueInstances == null) {
-             binaryValueInstances = new ArrayList<BinaryValue>();
-             
-             cleanupTasks.add(new CleanupTask() {
-                 
-                 @Override
-                 public void cleanup(final XQueryContext context) {
-                    if(context.binaryValueInstances != null) {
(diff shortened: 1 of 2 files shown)


joewiz:
Hi Adam - I'm just curious, would the issue you identified have resulted in memory leaks any time eXist read in or wrote out binary files to the filesystem, e.g., `file:read()`, `contentextraction:get-metadata-and-content()`, etc.? How about `util:binary-to-string()` and in-memory operations on binary data? 

If the only caveat is that it only happened when an XQuery was executed more than once, this is actually an extremely common scenario for a production server. I imagine this could have been a source of instability. Any info you had on cases you observed in testing or real world scenarios would be interesting to know about.

Thanks for catching this!

adamretter:
@joewiz Anytime an `xs:base64Binary` or `xs:hexBinary` where the data needs to be read more than once and that XQuery was executed more than once.

This issue was introduced between eXist 2.0 and 2.1, I find it incredible that I only had it reported to me just last week.

joewiz:
@adamretter Sounds like it must've been an active issue for binary-heavy applications. As to why it wasn't reported, I wonder if it was because there was no immediate sign that there was a problem. If nothing immediate happened, like the binary value suddenly being inaccessible, the user/developer would have no warning of a problem... the only side effect was an eventual OOM or ulimit issue. Possible?

adamretter:
@joewiz I think so. Still strange no one complained about leaking file handles until recently.

Issue #811, opened 2015-10-16
Modification times for existing resources will change on a failed parse attempt after uploading using XML-RPC


olvidalo:
While using eXist's XML-RPC API I encountered some strange behavior. When parsing an uploaded resource to overwrite an existing one, and parsing fails due to non-wellformed XHTML, the modification time of the old existing resource gets updated while its content does not get changed.

Is this intended behavior? Because it strikes me as a little unlogical and it makes it tricky for example to only upload locally changed files to eXist by checking the mtimes in the database without some workarounds.

This seems to be happening on 2.2 and also 3RC1.

adamretter:
Nope that's not intended. Sounds like a bug

PR #946, opened 2016-03-19
Feature/fix modification time bug

olvidalo:
When trying to overwrite an XML resource with an invalid XML resource, the modification time of the resource would still be updated even if the resource has not been changed. This is due to the method `manageDocumentInformation` in `validatateXMLResource` being called before the document is validated (in Collection.java). 

I have moved the modification time update out of `manageDocumentInformation` into a separate method which will only be called after the document is validated. There's also a test that will fail without the changes in Collection.java and verifies that modification times will still be updated as expected for valid resources.

This fixes the underlying issue for bug  #811.

### DIFF
--- a/src/org/exist/collections/Collection.java
+++ b/src/org/exist/collections/Collection.java
@@ -1779,8 +1779,7 @@ public class Collection extends Observable implements Comparable<Collection>, Ca
      */
     
     private void updateModificationTime(final DocumentImpl document) {
-        DocumentMetadata metadata = new DocumentMetadata();
-        metadata = document.getMetadata();
+        final DocumentMetadata metadata = document.getMetadata();
         metadata.setLastModified(System.currentTimeMillis());
         document.setMetadata(metadata);
     }
--- a/src/org/exist/collections/Collection.java
+++ b/src/org/exist/collections/Collection.java
@@ -1643,6 +1643,7 @@ public class Collection extends Observable implements Comparable<Collection>, Ca
                 if (LOG.isDebugEnabled()) {
                     LOG.debug("removing old document " + oldDoc.getFileURI());
                 }
+                updateModificationTime(document);
                 oldDoc.getUpdateLock().acquire(Lock.WRITE_LOCK);
                 oldDocLocked = true;
                 if (oldDoc.getResourceType() == DocumentImpl.BINARY_FILE) {
@@ -1748,7 +1749,6 @@ public class Collection extends Observable implements Comparable<Collection>, Ca
         if (oldDoc != null) {
             metadata = oldDoc.getMetadata();
             metadata.setCreated(oldDoc.getMetadata().getCreated());
-            metadata.setLastModified(System.currentTimeMillis());
             document.setPermissions(oldDoc.getPermissions());
         } else {
         	//Account user = broker.getCurrentSubject();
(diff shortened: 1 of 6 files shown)


src/org/exist/collections/Collection.java
```diff
@@ -1774,6 +1774,17 @@ private void manageDocumentInformation(final DocumentImpl oldDoc, final Document
         document.setMetadata(metadata);
     }
 
+     /** Update the modification time of a document
+     * @param document
+     */
+    
+    private void updateModificationTime(final DocumentImpl document) {
+        DocumentMetadata metadata = new DocumentMetadata();
```
dizzzz: Why note write like 
`DocumentMetadata metadata = document.getMetadata();`
?

dizzzz:
Change looks good to me, was a bit puzzled (in the diff tool) about the binary document handling

test/src/org/exist/storage/ModificationTimeTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+    protected static String XML_FILENAME = "test.xml";
+    protected static String VALID_XML = "<?xml version=\"1.0\"?>" +
+                                        "<valid/>";
+    protected static String INVALID_XML = "<?xml version=\"1.0\"?>"
+            + "<invalid>";
+
+    
+    @BeforeClass
+    public static void ensureCleanDatabase() throws IOException {
```
adamretter: It seems to me that this would be useful for other tests too.
Could you move the logic of this method into org.exist.TestUtils?

test/src/org/exist/storage/ModificationTimeTest.java
```diff
@@ -0,0 +1,209 @@
...
+            assertNull(info);
+            
+            doc = getDocument(docUri);
+            assertNotNull(doc);
+            
+            long modificationTimeAfter = doc.getMetadata().getLastModified();
+            assertEquals(modificationTimeBefore, modificationTimeAfter);
+    }
+
+   protected BrokerPool startDB() throws DatabaseConfigurationException, EXistException {
```
adamretter: Could this not have an `@Before` annotation and be called once per-test, or better yet if it's possible then `@BeforeClass` so that the database only need be started once for the suite of tests (you would also need to change your `@After` to `@AfterClass`.

src/org/exist/collections/Collection.java
adamretter: Also can you make use of `final` unless you variable need to be mutable for some reason please?

adamretter:
@olvidalo Welcome and thank you for your PR :-) Just a couple of small comments inline, otherwise it looks good to me.

olvidalo:
Hey thanks for the welcome and all your suggestions.

@adamretter I tried having the database start only once in a `@beforeClass` method and stopping it in `@afterClass`but while this works fine when I run the ModificationTimeTest by itself, I ran into problems when running the test at the end of `ant test-local`. Sometimes I would get a random NullPointerException or failed assertions, each time somewhere else. Now what do you think, should I commit these changes so that maybe someone more knowledgeable about eXist's internal workings than me could have a look at it or should I just go with the `@Before` approach, starting the database before each test?

dizzzz:
@adamretter ?

PR #948, opened 2016-03-21
Utilities for trace logging the history of broker leases and broker subject changes

adamretter:
Note - These utilities are only ever enabled when the logging level is set to TRACE

### DIFF
--- a/src/org/exist/storage/BrokerPool.java
+++ b/src/org/exist/storage/BrokerPool.java
@@ -479,6 +479,14 @@ public class BrokerPool implements Database {
      */
     private final Map<Thread, DBBroker> activeBrokers = new ConcurrentHashMap<>();
 
+
+    /**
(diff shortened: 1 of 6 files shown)


src/org/exist/storage/BrokerPool.java
```diff
@@ -1657,7 +1686,7 @@ void release(final DBBroker broker) {
             //Broker is no more used : inactivate it
             for(final DBBroker inactiveBroker : inactiveBrokers) {
                 if(broker == inactiveBroker) {
-                    LOG.error("Broker is already in the inactive list!!!");
+                    LOG.error("Broker " + broker.getId() + " is already in the inactive list!!!");
```
dizzzz: Could have been `LOG.error("Broker {} is already in the inactive list!!!". broker.getId());` ? anyway leave it

PR #949, opened 2016-03-21
[bugfix] New range index returns empty sequence

### LINKED ISSUE #873: optimizer regression for indirect queries on Lucene full-text and range indexes in eXist-2.2 and eXist-3.0.RC1
In eXist-2.2 and eXist-3.3.RC1, the optimizer seems to behave buggy for indirect queries on node sets that have been stored in a variable first. When such indirect queries are run on the Lucene full-text and range indexes, the search is either slow or incorrect, which suggests that the optimizer isn't working optimally.

For comparison, behaviour for both 'direct' and 'indirect' versions of queries will be described below, on both Lucene full-text and range indexes. These tests can be run straight on the eXist demo app.
- Lucene full-text index: indirect queries work, but take _much_ longer than direct queries
  - direct query: 419 results in under 0.020 sec.
    
    let $a := //SPEECH[ft:query(., 'hamlet')]
    return $a
  - indirect query: 419 results in **more than 2 sec.**
    
    let $a := //SPEECH
    return $a[ft:query(., 'hamlet')]
- Lucene range index: indirect queries don't return any results at all
  - direct query: 359 results in 0.018 sec.
    
    let $a := //SPEAKER[range:matches(., 'HAMLET')]
    return $a
  - indirect query: **0 results**
    
    let $a := //SPEAKER
    return $a[range:matches(., 'HAMLET')]

It seems that for indirect queries on the Lucene full-text and range indexes, the optimizer does not query the actual indexed nodes.

wolfgangmm:
when used on filter expression, e.g. $x[range:matches(., "y")]. Closes #873.

### DIFF
--- a/extensions/indexes/range/src/org/exist/indexing/range/RangeIndexWorker.java
+++ b/extensions/indexes/range/src/org/exist/indexing/range/RangeIndexWorker.java
@@ -512,7 +512,7 @@ public class RangeIndexWorker implements OrderedValuesIndex, QNamedKeysIndex {
     public NodeSet query(int contextId, DocumentSet docs, NodeSet contextSet, List<QName> qnames, AtomicValue[] keys, RangeIndex.Operator operator, int axis) throws IOException, XPathException {
         return index.withSearcher(searcher -> {
             List<QName> definedIndexes = getDefinedIndexes(qnames);
-            NodeSet resultSet = NodeSet.EMPTY_SET;
+            NodeSet resultSet = new NewArrayNodeSet();
             for (QName qname : definedIndexes) {
                 Query query;
                 String field = LuceneUtil.encodeQName(qname, index.getBrokerPool().getSymbols());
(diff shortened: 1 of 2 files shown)


extensions/indexes/range/src/org/exist/indexing/range/RangeIndexWorker.java
```diff
@@ -512,7 +512,7 @@ private void write() {
     public NodeSet query(int contextId, DocumentSet docs, NodeSet contextSet, List<QName> qnames, AtomicValue[] keys, RangeIndex.Operator operator, int axis) throws IOException, XPathException {
         return index.withSearcher(searcher -> {
             List<QName> definedIndexes = getDefinedIndexes(qnames);
-            NodeSet resultSet = NodeSet.EMPTY_SET;
+            NodeSet resultSet = new NewArrayNodeSet();
```
dizzzz: should `NodeSet.EMPTY_SET` be returned of none of the `resultSet.addAll(doQuery()))` adds any results?

wolfgangmm: I don't think we need to check for empty here. Adding an empty sequence to an empty sequence should result in an empty sequence ;-)

PR #957, opened 2016-03-27
Open up the ability to index/reindex a single document

adamretter:


dizzzz:
Straightforward change, 2 questions:

- What is the use case ? shouldn't we extend the `xmldb:reindex()` function too to accept a single source, or am I wrong here>
- There are more indexes, e.g. the Spatial. I think @ljo has one under management these days, this means we need to inform the community (exist-develop/exist-open) about this change too ?

adamretter:
@dizzzz The use case for me at the moment is internal to an extension I have for eXist. However I could expose it through `xmldb:reindex()`, do you want me to do that?

@dizzzz @ljo I will update the `SpatialIndex` too but I can't update peoples own indexes as that is outside the scope of `eXist-db`.

dizzzz:
Not sure if the reindex function per doc is required.... Would Ik make sense? For myself I see no need.

Sure, spatial is sufficient

joewiz:
I was trying to think of a use case and came up with one. How's this:

1. I have 100 large, 10 MB XML documents in an un-indexed collection.
2. I want to run a query on one document that requires an index, so I add an index via a new collection.xconf file.
3. Rather than indexing the entire collection in order to query the one document, I use Adam's xmldb:reindex to apply the index to just one document. 
4. I run the query, yay, done.
5. Now in order to prevent the other documents from being indexed upon create/save, I remove the collection.xconf, and again apply Adam's xmldb:reindex to the single document, thus wiping all stored index data on the collection.

So basically, it allows targeted indexing of individual documents.

shabanovd:
@joewiz that is going to be breaking logic. What to do with documents stored between collection. xconf changes?

dizzzz:
Hnmmm not sure if this gonna work, @wolfgangmm can you join us here?
I remember that index-optimized queries do not work (well?) for collections where some documents are indexed and some are not. If an index is available, I think only results of indexed documents are included, the not indexed documents are left out (which makes sense in a way ; it is a design decision)

joewiz:
@shabanovd I didn't consider other docs being added in the scenario above. If the query in step 4 used `collection()`, yes, the logic would break; but if `doc()`, then it would run fine. A problem perhaps better solved by copying to a temp dir.

joewiz:
Okay, a new use case: 

1. I have an app in /db/apps/my-app with lots of indexed data in a subcollection, data. The xconf is applied at the my-app level, as is typical.
2. I want to apply an index to a file stored in my-app's top level collection. I don't want to reindex all data, especially on a production server, just the new file.
3. So I modify the xconf and reindex the one file.

adamretter:
@dizzzz As requested this now covers the SpatialIndex and also implements the `xmldb:reindex` that @joewiz asked for.

dizzzz:
Thnx!

ljo:
Thanks. Last night I also updated my repos of xar indexing apps: geospatial (to replace spatial) and sparql-rdf.

dizzzz:
@ljo great job!

PR #961, opened 2016-03-29
[bugfix] Deprecate the datetime:format-* functions in favour of those from XPath F+O 3.0

adamretter:


joewiz:
Thanks, @adamretter and @dizzzz!

PR #971, opened 2016-04-11
[performance] Remove two calls to System.gc and excessive logging.

wolfgangmm:
The calls to System.gc date back to the early days and are definitely unnecessary today. They dramatically slow down the reindexing process. Stress testing did not show any negative effects after dropping those calls.

### DIFF
--- a/src/org/exist/storage/NativeValueIndex.java
+++ b/src/org/exist/storage/NativeValueIndex.java
@@ -1323,7 +1323,7 @@ public class NativeValueIndex implements ContentLoadingObserver {
                 atomic = new StringValue( value ).convertTo( xpathType );
             }
             catch( final XPathException e ) {
-                LOG.warn( "Node value '" + value + "' cannot be converted to " + Type.getTypeName( xpathType ) );
+                LOG.debug( "Node value '" + value + "' cannot be converted to " + Type.getTypeName( xpathType ) );
                 return( null );
             }
         }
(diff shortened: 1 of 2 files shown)


src/org/exist/storage/NativeValueIndex.java
```diff
@@ -1323,7 +1323,7 @@ private AtomicValue convertToAtomic( int xpathType, String value )
                 atomic = new StringValue( value ).convertTo( xpathType );
             }
             catch( final XPathException e ) {
-                LOG.warn( "Node value '" + value + "' cannot be converted to " + Type.getTypeName( xpathType ) );
+                LOG.debug( "Node value '" + value + "' cannot be converted to " + Type.getTypeName( xpathType ) );
```
adamretter: The change on this line doesn't make sense to me. This I believe is a genuine problem which we are then effectively hiding from the user. This should at least be a WARN IMHO. @wolfgangmm @dizzzz Can you please explain why you think this should be hidden during normal operation?

wolfgangmm: One often gets thousands of those messages during a reindex, which potentially hides more severe issues. It is obvious that e.g. an xs:date index can only be applied to dates and the log message doesn't help much to identify the non-compliant elements or attributes within a large data set. It thus has no added value apart from flooding the logs.

If you want to make sure your data is consistent, checking types via a schema is the way to go.

adamretter: Thanks @wolfgangmm. I understand but I think in future we will need to develop a better error reporting mechanism to our users around indexing

ljo:
@adamretter I agree on the general error reporting mechanism. I also have started an inventory of where we should have error reporting but we do not have it yet.

adamretter:
@ljo :-)

Issue #952, opened 2016-03-23
fop config -> issue resolving fonts via embed-url


windauer:
As described in https://github.com/eXist-db/exist/pull/938 I tried to adjust the fop config for the tei-simple-pm app (https://github.com/wolfgangmm/tei-simple-pm).

I changed the fop config in fo.xql (https://github.com/wolfgangmm/tei-simple-pm/blob/master/modules/fo.xql)  as follows: 

* removed <font-base> base from the config
* changed  <font embed-url="Junicode.ttf" ../> to "<font embed-url="exist://localhost:8080/db/apps/tei-simple/resources/fonts/Junicode.ttf"/>

This results in the following error: javax.servlet.ServletException: An error occurred: URI scheme is not "file". 

Am I doing something wrong? I tried various combinations as embed-url, with and without port 8080, http as url scheme, but nothing worked.

dizzzz:
do you have a longer trace?

adamretter:
@windauer Can you push your changes to `tei-simple-pm` to your fork and let me know where it is so I can test?

windauer:
simply try to render a pdf in https://github.com/wolfgangmm/tei-simple-pm (fop font config in fo.xql breaks pdf rendering)

adamretter:
@windauer Using the latest code from `https://github.com/wolfgangmm/tei-simple-pm` I get a different error (to the one you reported):

```
Problem accessing /exist/apps/tei-simple/modules/fo.xql. Reason:

    Server Error
Caused by:

javax.servlet.ServletException: An error occurred: java.io.IOException: Resource /db/Junicode.ttf not found in database.
	at org.exist.http.servlets.EXistServlet.doGet(EXistServlet.java:263)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:735)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:848)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:684)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:503)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:137)
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:575)
	at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:231)
[...]

Issue #558, opened 2015-04-02
Deadlock in validation (and more?)


adamretter:
I have found a repeatable deadlock in validation:

```java
final TransactionManager transact = brokerPool.getTransactionManager();
try(final DBBroker broker = brokerPool.get(user));
        final Txn transaction = transact.beginTransaction()) {

  Collection coll = broker.getOrCreateCollection("/db/validate-test");
  broker.saveCollection(transaction, coll);

  //do some validatation
  final Properties properties = new Properties();
  final ExistResolver resolver = new ExistResolver(brokerPool);
  ValidateProperty.URI_RESOLVER.put(properties, resolver);
  ValidateProperty.ENTITY_RESOLVER.put(properties, resolver);
  final ValidationDriver driver = new
ValidationDriver(properties.toPropertyMap(), schemaReader);

  // Load schema

  InputSource grammar =
Shared.getInputSource("/db/validate-test/something.nvdl", context);

  driver.loadSchema(grammar); // THIS WILL DEADLOCK PROGRESS

  transaction.commit();
}
```

So the real issue is the interaction between `broker.getOrCreateCollection` and `EmbeddedDownload` used by the validation. Basically what happens is -

1) The `main` thread executing the `broker.getOrCreateCollection` will acquire a `WRITE_LOCK` on the collection `/db/validate-test` and add it to the transaction.

2) When `driver.loadSchema`is executed, it uses the registered `ExistResolver`.

3) The `ExistResolver` will call `EmbeddedInputStream`, and the key thing here is that `EmbeddedInputStream` creates a new thread - `EmbeddedDownloadThread`.

4) `EmbeddedDownloadThread` will attempt to take a `READ_LOCK` on `/db/validate-test`.

5) KABOOM! Basically (1) is waiting on (4) to progress, and (4) is awaiting on the lock held by (1).

This is not the classic deadlock. It happens specifically because `NativeBroker#getOrCreateCollection` causes the transaction to hold onto a `WRITE_LOCK` until it is committed.

Unfortunately, the architecture of this issue is not just limited to validation...

The implication of this, is that within the same transaction, it is impossible to run any code in another thread that may try to access the collection (or a sub-collection or resource of).

That is exactly what the `ExistResolver` does! This limitation will lead to a deadlock anywhere multithreaded code is used within the same transaction (if the same collection (or sub collection/resource) is accessed.

[...]

PR #977, opened 2016-04-16
[feature] Make sure to correctly resolve binary resources from the database as well as XML resources

adamretter:
Closes https://github.com/eXist-db/exist/issues/952


dizzzz:
an error popped up:

```
    [javac] /home/travis/build/eXist-db/exist/src/org/exist/xquery/functions/transform/ExistUriResolver.java:47: error: class EXistURIResolver is public, should be declared in a file named EXistURIResolver.java
```

src/org/exist/xquery/functions/transform/ExistUriResolver.java
```diff
@@ -0,0 +1,176 @@
...
+        this.broker = broker;
+        this.basePath = docPath;
+        LOG.debug("EXistURIResolver base path set to " + basePath);
+    }
+
+    /**
+     * Simplify a path removing any "." and ".." path elements.
+     * Assumes an absolute path is given.
+     */
+    private String normalizePath(final String path) {
```
dizzzz: I 'd hope that such a method would already exist either in our codebase  or in an e.g. apache library :-)
It is probably useful on several locations.....

adamretter: @dizzzz I added a commit to remove some dead code which did nothing previously. This code was previously a sub-class in FnTransform.java, I just moved it out and extended it to be more flexible. There is also the `org.exist.validation.resolver.unstable.ExistResolver` which you added but it is marked unstable and the validation code can lead to deadlocks - https://github.com/eXist-db/exist/issues/558

dizzzz: I marked unstable because it was work in progress, but was not really finished over time :-)

adamretter:
@dizzzz I fixed the file naming issue... I hate the fact that filenames on OSX filesystem are case-insensitive but preserving, I mean who decided that was even remotely a good idea???

dizzzz:
yes, it s*cks ; you can actually switch it off, but that might have serious consequences.

src/org/exist/xquery/functions/transform/ExistUriResolver.java
adamretter: @dizzzz Okay, if it is unused should we remove it? If not perhaps switch to using the one I extended?

dizzzz: yes sounds good. your code looks much better....

src/org/exist/xquery/functions/transform/EXistURIResolver.java
```diff
@@ -0,0 +1,176 @@
...
+            doc = broker.getResource(uri, Permission.READ);
+            if (doc == null) {
+                LOG.error("Document " + path + " not found");
+                throw new TransformerException("Resource " + path + " not found in database.");
+            }
+
+            final Source source;
+            if (doc instanceof BinaryDocument) {
+                final Path p = broker.getBinaryFile((BinaryDocument) doc);
+                source = new StreamSource(p.toFile());
```
dizzzz: he...... do we have here direct access to the Files deep down in WEBINF/data ?

adamretter: Yes we have direct access to binary files in WEB-INF/data/fs for the purpose of reading

Issue #969, opened 2016-04-11
system:export function documentation is incorrect


gspringtech:
(See src/../xquery/functions/system/FnExport.java)
The documentation for this function says "Restore the database or a section of the database" - when in fact this function generates a backup file or directory.

The $dir-or-file parameter should be $target-directory - "The destination for the backup - this directory must be writeable by the database process".
The other parameters are correct.

The associated error message should refer to "export" and not "restore"

adamretter:
@gspringtech Can you send us a PR please with the updated documentation that you would like to see?

PR #978, opened 2016-04-17
Fix for incorrect documentation in system:export function

gspringtech:
See issue #969


adamretter:
@gspringtech Thanks Alister :-)

PR #980, opened 2016-04-19
Content extractor ws handling

shabanovd:
question: should WS handling be configurable? Mean turn it on/off.

### DIFF
--- a/src/org/exist/util/serializer/SAXToReceiver.java
+++ b/src/org/exist/util/serializer/SAXToReceiver.java
@@ -34,9 +34,15 @@ public class SAXToReceiver implements ContentHandler, LexicalHandler {
 
     private Receiver receiver;
     private boolean inCDATASection = false;
+    private boolean sendWStoCharacters = false;
 
     public SAXToReceiver(Receiver receiver) {
+        this(receiver, false);
+    }
+
+    public SAXToReceiver(Receiver receiver, boolean sendWStoCharacters) {
         this.receiver = receiver;
+        this.sendWStoCharacters = sendWStoCharacters;
     }
 
     public void setDocumentLocator(Locator locator) {
@@ -100,6 +106,9 @@ public class SAXToReceiver implements ContentHandler, LexicalHandler {
     }
 
     public void ignorableWhitespace(char[] ch, int start, int length) throws SAXException {
+        if (sendWStoCharacters) {
+            receiver.characters(new XMLString(ch, start, length));
+        }
     }
 
     public void processingInstruction(String target, String data) throws SAXException {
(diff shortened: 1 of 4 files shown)


ljo:
Looks good.

src/org/exist/util/serializer/SAXToReceiver.java
```diff
@@ -34,9 +34,15 @@
 
     private Receiver receiver;
     private boolean inCDATASection = false;
+    private boolean sendWStoCharacters = false;
```
dizzzz: make consistent with "redirectIgnorableWhitespace" ? functionally it is the same...

i'd probably prefer a shorter name, avoiding "Characters". Maybe the 'invertway' notation is a bit more clear, like "ignoreWhitespace" (normally it is not)

shabanovd: @dizzzz on/off switch for ignorableWhitespace redirection to characters ... where is the problem?

dizzzz:
I find the naming confusing/unclear ; a bit too technical

shabanovd:
@dizzzz it have to be technical, no? My view on subject that as long as you can understand that it's ok, or we can spend few week to discuss what naming is better and find no agreement

dizzzz:
I am  not interested into long discussions either, but I find the change not logical regardingnames; 

e.g. why "redirectIgnorableWhitespace " and "sendWStoCharacters" while they are doing exactly the same?  whilst the called method is called "characters" what the code does is actually pass whitespace data to 'something'. so IMO it is better to have the function in the variableName in stead of the name of the method that is codewise called. That is why I referred to the javadoc.

shabanovd:
@dizzzz I'm not interesting to continue discussion. If you need different name you are welcome.

ljo:
@dizzzz Shall I accept then?

dizzzz:
could we at least have consisting naming then? Why is there such a negative response on my feedback?  I feel very bad about this....

adamretter:
@dizzzz @shabanovd I think there are two separate things here and maybe you are talking at cross-purposes, I may be wrong though so please correct me:

1) The naming of the variable `sendWStoCharacters`.

2) Having two variables, which effective control the same outcome, but giving them different names i.e. `redirectIgnorableWhitespace` and `sendWStoCharacters`.

IMHO: Yes (1) could be improved from an *english* perspective, and (2) well if they are truly controlling the same thing (I am not certain), then yes it would be nice to settle on the one name.

I think @shabanovd you bowed out a little quickly, however I understand that you are busy and also agree that if @dizzzz wants to improve on this then he can of course.

@dizzzz I do think it is good to raise concerns here first as you have done. As @shabanovd has now said he is not interested in pursuing this further, it's maybe quicker for you to rebase Dmitriy's PR make changes and merge?

dizzzz:
My proposed design is in https://github.com/eXist-db/exist/pull/981 (on purpose I did not want to put the changes directly into this PR)

PR #982, opened 2016-04-19
XInclude issue

patczar:
There was a problem when in the middle of one serialisation process, through an XInclude tag calling a stored XQuery, and xmldb operations inside it, another serialisation process was started. It all ended in a NPE or other errors.
This change enforces to use a separate DBBroker when executing an XQuery from the context of XInclude. I achieved it by executing the XQuery in a separate thread, which definitely IS a cost, but taking into account that calling XQuery from XInclude during serialisation is an unusual case, maybe it is acceptable...

### DIFF
--- a/src/org/exist/storage/serializers/XIncludeFilter.java
+++ b/src/org/exist/storage/serializers/XIncludeFilter.java
@@ -517,21 +517,26 @@ public class XIncludeFilter implements Receiver {
         return Optional.empty();
     }
 
-	/** Executes the given compiled XQuery in a separate thread, using a separate DBBroker.
-	 * @param compiled the query to execute
-	 * @param contextSeq the context sequence for the query
+	/**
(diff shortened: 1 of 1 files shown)


adamretter:
@patczar Your approach looks fine, however it causes a test failure. Can you take a look and fix it first and then I will merge.

shabanovd:
@patczar did #1038 fix your issue?

JoernT:
one test just fails due to some permission denied exception - the resource not being available for guest user. Maybe the fix is functional already but the test setup is wrong?

shabanovd:
@JoernT #1038 should this XInclude issue. I'm waiting for confirmation to close this PR.

adamretter:
@patczar I believe this was fixed here https://github.com/eXist-db/exist/commit/fea73697c9b93c4b9f3a131d5aa26e168444638a

If not please feel free to send an updated PR.

PR #983, opened 2016-04-23
Indexing enhancements

adamretter:
1. Provides better hints to the Index's StreamListener around the start and end of indexing for a document.

2. Sends start/end document events to the Index's StreamListener for adding and removing binary resources. This provides the facility for a custom index to index binary resources.

### DIFF
--- a/src/org/exist/Indexer.java
+++ b/src/org/exist/Indexer.java
@@ -19,10 +19,7 @@
  */
 package org.exist;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Observable;
-import java.util.Stack;
+import java.util.*;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -68,7 +65,6 @@ import org.xml.sax.ext.LexicalHandler;
  * index-creation.
  * 
  * @author wolf
- * 
  */
 public class Indexer extends Observable implements ContentHandler, LexicalHandler, ErrorHandler {
 
@@ -89,36 +85,36 @@ public class Indexer extends Observable implements ContentHandler, LexicalHandle
     public final static String PROPERTY_SUPPRESS_WHITESPACE = "indexer.suppress-whitespace";
     public static final String PROPERTY_PRESERVE_WS_MIXED_CONTENT = "indexer.preserve-whitespace-mixed-content";
 
-    protected DBBroker broker = null;
-    protected Txn transaction;
+    private final DBBroker broker;
+    private final Txn transaction;
 
-    protected StreamListener indexListener;
+    private StreamListener indexListener;
 
-    protected XMLString charBuf = new XMLString();
-    protected boolean inCDATASection = false;
-    protected int currentLine = 0;
-    protected NodePath currentPath = new NodePath();
+    private XMLString charBuf = new XMLString();
+    private boolean inCDATASection = false;
+    private int currentLine = 0;
+    private final NodePath currentPath = new NodePath();
 
-    protected DocumentImpl document = null;
-    protected IndexSpec indexSpec = null;
+    private DocumentImpl document = null;
+    private IndexSpec indexSpec = null;
 
-    protected boolean insideDTD = false;
-    protected boolean validate = false;
-    protected int level = 0;
-    protected Locator locator = null;
-    protected int normalize = XMLString.SUPPRESS_BOTH;
-    protected Map<String, String> nsMappings = new HashMap<>();
-    protected Element rootNode;
+    private boolean insideDTD = false;
+    private boolean validate = false;
+    private int level = 0;
+    private Locator locator = null;
+    private int normalize = XMLString.SUPPRESS_BOTH;
+    private final Map<String, String> nsMappings = new HashMap<>();
+    private Element rootNode;
 
-    protected Stack<ElementImpl> stack = new Stack<>();
-    protected Stack<XMLString> nodeContentStack = new Stack<>();
+    private final Deque<ElementImpl> stack = new ArrayDeque<>();
+    private final Deque<XMLString> nodeContentStack = new ArrayDeque<>();
 
-    protected StoredNode prevNode = null;
+    private StoredNode prevNode = null;
 
-    protected String ignorePrefix = null;
-    protected ProgressIndicator progress;
+    private String ignorePrefix = null;
+    private ProgressIndicator progress;
 
-    protected boolean suppressWSmixed = false;
+    private boolean suppressWSmixed = false;
 
     protected int docSize = 0;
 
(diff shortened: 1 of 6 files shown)


src/org/exist/Indexer.java
```diff
@@ -305,6 +305,9 @@ public void endDTD() {
     @Override
     public void endDocument() {
         if (!validate) {
+            if(indexListener != null) {
+                indexListener.endDocument(transaction);
```
shabanovd: Better to send also 2nd parameter DocumentImpl and name it startProcessing(Txn, DocumentImpl) & endProcessing(Txn, DocumentImpl) because there are xupdate cases (element and attribute) that process only part of document (one attribute).

adamretter: I don't need to send `DocumentImpl` as `setDocument` will have been called on the Index before `startDocument` is called on the StreamListener. `startDocument` and `endDocument` do not indicate the start or end of a DOM Document Node, rather they indicate that start and end of indexing a document.

shabanovd: `startDocument` & `endDocument` confuse because of sax api methods name http://www.saxproject.org/apidoc/org/xml/sax/ContentHandler.html#startDocument()

adamretter: It's no more confusing than the existing `endElement` method etc which are already there and are not SAX related either ;-)

shabanovd: it's not because similar by behavior and meaning

adamretter: What naming would you like me to change it to?

shabanovd: https://github.com/shabanovd/exist/blob/2.1.0-RC54/src/org/exist/indexing/AbstractStreamListener.java#L86 and https://github.com/shabanovd/exist/blob/2.1.0-RC54/src/org/exist/indexing/AbstractStreamListener.java#L93

shabanovd: and it may looks unrequired DocumentImpl parameter, but it help a lot with debugging to make sure that document is right one (Personally, I do not like that index have to hold state: document and mode. That lead to bugs)

dizzzz:
Looks good to me, @wolfgangmm @ljo ?

adamretter:
@dizzzz I think @shabanovd want's some name changes, see his comments inline...

shabanovd:
@adamretter if you agree I can commit that changes

src/org/exist/Indexer.java
adamretter: @shabanovd Okay so I like `startDocument` (and `endDocument`) rather than `startProcessing` (and `endProcessing`). As `startProcessing` implies to me a complete indexing run (e.g. multiple documents) as opposed to a single document, we might also like to signal reindexing operations by using something like `startProcessing` or `startIndexing`.

I also agree that the Indexer should ideally not keep state. However that was not a change that I was trying to make in this PR. Such a change would require a larger design of the indexing architecture to move the `pending` and `flush` operations into a stream based approach instead. So I won't change the mutable/immutable property of the IndexWorker just yet, maybe in future...

adamretter:
@shabanovd I just responded inline, that I think `startProcessing` has a larger meaning, and that we might want to send a follow-up PR to exploit that.

src/org/exist/Indexer.java
shabanovd: 'keep state' comment was to explain why `startProcessing` have document and mode (in future) parameters. That was not proposal for such redesign.

I'm not sure that understand why batch processing need "start" & "end" methods? I see only requirements to call `flush` method after processing done.

shabanovd: will `startDocument ` & `endDocument` be accepted here https://github.com/eXist-db/exist/blob/503ad3283e7f1cb7049b05eb3c5ed3d5d9a8b891/src/org/exist/dom/persistent/ElementImpl.java#L533 ? (as example)

adamretter: @shabanovd Potentially, at present I have intentionally ignored XQuery Update and XUpdate

shabanovd: by "accepted" I mean will it confuse person that reading code?

adamretter: @shabanovd I am not sure I follow you. The code seems clear to me and I guess to you also.

shabanovd: `listener.startDocument` and `listener.endDocument` will not confuse you when you read that code, right?

adamretter: @shabanovd You mean inside `appendChildren`? If so, yes I could see that might be confusing, but then I might move such calls around a block of XQuery Update to a document.

shabanovd: you can't because xupdate happen on nodes set and `appendChildren` used at different places.

Well, if you like such naming and think that it'll not confuse people ... then I'm 1st confused creature.

adamretter:
@shabanovd Okay, to avoid potential confusion I have renamed:

1. `startDocument` -> `startIndexDocument`
2. `endDocument` -> `endIndexDocument`

shabanovd:
@adamretter thanks, I just was on a way to make similar proposal

src/org/exist/indexing/IndexController.java
```diff
@@ -331,6 +331,18 @@ public void indexNode(final Txn transaction, final IStoredNode node, final NodeP
     }
 
     /**
+     * Helper method: indexing is starting for a document
+     *
+     * @param transaction the current transaction
+     * @param listener the StreamListener which receives the index events
+     */
+    public void startIndexDocument(final Txn transaction, final StreamListener listener) {
```
shabanovd: `StreamListener listener` looks wrong because `IndexController` have `getStreamListener` method https://github.com/adamretter/exist/blob/indexing-enhancements/src/org/exist/indexing/IndexController.java#L277

dizzzz:
oops did ~I pull in too soon?

src/org/exist/indexing/IndexController.java
adamretter: @shabanovd I am following the existing convention which is to pass the StreamListener as the last argument. That is fine.

adamretter:
@dizzzz Nope, your fine. Thank you :-)

shabanovd:
@adamretter @dizzzz yeap, you did!!! =)

src/org/exist/indexing/IndexController.java
shabanovd: @adamretter where is it? any example at current code?

adamretter: @shabanovd sorry, where is what?

If you mean where is the convention? Then look at the other methods like `startElement` they all take a `StreamListener`!

shabanovd: @adamretter I'm totally lost because IndexController hold StreamListener as class field and pass it as last parameter https://github.com/adamretter/exist/blob/0478bc8240a2d26df8a7360cb261f0d6c03ea9b2/src/org/exist/indexing/IndexController.java#L58

adamretter:
@joewiz Only if that was now refactored to be an Indexing Module as opposed to an XQuery Module

joewiz:
@adamretter Okay, I see now, thanks!

src/org/exist/indexing/IndexController.java
shabanovd: @adamretter @wolfgangmm @ljo @dizzzz ok, what way StreamListener should be used as method parameter or class field?

shabanovd:
@joewiz I did add feature to index binary documents at lucene index at my branch. If you are interesting it I can make PR.

PR #994, opened 2016-05-05
Added util:line-number()

adamretter:


shabanovd:
@adamretter what is the purpose of such function? you always know line number where it placed, no?

adamretter:
It's for knowing the line-number at runtime when collecting debug info within an XQuery.

shabanovd:
@adamretter here was discussion but it disappear now ... anyway, I still can't understand how line number of `util:line-number()` will help you ... do I missing something?

adamretter:
@shabanovd Here is how I am using it:

```xquery
declare function assert:equals($expected as item(), $actual as item(), $line-number as xs:integer) {
    if($expected ne $actual) then
        fn:error(xs:QName('assersion'), 'Expected ' || $name || ', but actually ' || $actual || ' at ' || $line-number)
    else()
};
```

Consider the use of the above function for making assertions:

```xquery
...
assert:equals("x", "y", util:line-number())
...
```

dizzzz:
@adamretter the PR has now a merge conflict :-(

adamretter:
@dizzzz I just rebased... should be good now :-)

adamretter:
@dizzzz Mergy mc-merge?

PR #995, opened 2016-05-05
Added $util:error-code for util:catch

adamretter:


dizzzz:
Wouldn't it be better to use the try-catch clauses only for this info? I think we should avoid double code. Did we mark this function as deprecated?

adamretter:
@dizzzz It was already marked as deprecated. I added this for a large XQuery code base which already has lots of `util:catch`.

It was easier to use this in a couple of places than update the whole code base to `try/catch`. Also it makes writing unit tests around `util:catch` trivial.

dizzzz:
Ok :)

I did not see the marker....

PR #997, opened 2016-05-06
[bugfix] Missing collection is a concern and must be logged as an error

adamretter:


dizzzz:
Hmmm are you sure about this?
There is on a other location a nasty stack trace visible when a lock is made, released whilst In between nothing happened. Also a legimate use case with nasty log pollution.

I'll study the details ....

adamretter:
@dizzzz yes absolutely

dizzzz:
Ok, I saw the PR on a mobile device initially. looks arbitrary to me....

PR #1000, opened 2016-05-07
avoid concurrent modification of tabSize for iterator

shabanovd:


dizzzz:
@shabanovd under what conditions does this happen?
I find it difficult to judge, please jump in @wolfgangmm or @adamretter  ?

adamretter:
@shabanovd good catch! :-)

I note there is also an atomicity problem in the constructor of ObjectHashSetStableIterator. But we can fix that next...

PR #1001, opened 2016-05-09
move svn extension to https://github.com/shabanovd/eXist-svn

shabanovd:


dizzzz:
you did not preserve the change history .... is that on purpose?

shabanovd:
yes, because history stay here

dizzzz:
I ask because maintaining history is not that difficult ....

shabanovd:
no needs

ljo:
OK. But then we could say the same about the geospatial index extension I upgraded and moved to an indexing app? That the history stays here that is.

PR #996, opened 2016-05-05
remove XACML code

shabanovd:

PR #1006, opened 2016-05-09
XQuery profiler should track location of function call, not called class

wolfgangmm:
Profiling stats only included the name of the called class for built-in functions. This is useless for profiling: instead you want to see the source location from where the function was called.

Also includes a small memory leak fix for function util:compile.

### DIFF
--- a/src/org/exist/xquery/Profiler.java
+++ b/src/org/exist/xquery/Profiler.java
@@ -181,7 +181,8 @@ public class Profiler {
 
     public final void traceFunctionEnd(Function function, long elapsed) {
         if (stats.isEnabled()) {
-            final String source = String.format("%s [%d:%d]", function.getContext().getXacmlSource().getKey(), function
+            final String source = String.format("%s [%d:%d]", function.getContext().getSource().getKey(),
+                    function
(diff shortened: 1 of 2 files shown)


src/org/exist/xquery/Profiler.java
```diff
@@ -181,13 +181,8 @@ public final void traceFunctionStart(Function function) {
...
     public final void traceFunctionEnd(Function function, long elapsed) {
         if (stats.isEnabled()) {
-            String source;
-            if (function instanceof InternalFunctionCall) {
-                source = ((InternalFunctionCall) function).getFunction().getClass().getName();
-            } else {
-                source = function.getContext().getSource().path();
-            }
-            source = String.format("%s [%d:%d]", source, function.getLine(), function.getColumn());
+            final String source = String.format("%s [%d:%d]", function.getContext().getXacmlSource().getKey(), function
```
shabanovd: `getXacmlSource` was removed by https://github.com/eXist-db/exist/pull/996

wolfgangmm: Yes, see my follow up commit.

PR #1007, opened 2016-05-10
General code cleanup of internal hashtable implementations

adamretter:
1. Reformatted and cleaned up the code
2. Marked whether each implementation is ThreadSafe with JCIP annotations

### DIFF
--- a/src/org/exist/collections/Collection.java
+++ b/src/org/exist/collections/Collection.java
@@ -101,13 +101,13 @@ public class Collection extends Observable implements Comparable<Collection>, Ca
     private int collectionId = UNKNOWN_COLLECTION_ID;
     
     // the documents contained in this collection
-    private Map<String, DocumentImpl> documents = new TreeMap<String, DocumentImpl>();
+    private final Map<String, DocumentImpl> documents = new TreeMap<>();
     
     // the path of this collection
     private XmldbURI path;
     
     // stores child-collections with their storage address
-    private ObjectHashSet<XmldbURI> subCollections = new ObjectHashSet<XmldbURI>(19);
+    private ObjectHashSet<XmldbURI> subCollections = new ObjectHashSet<>(19);
     
     // Storage address of the collection in the BFile
     private long address = BFile.UNKNOWN_ADDRESS;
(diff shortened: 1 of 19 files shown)


src/org/exist/collections/Collection.java
```diff
@@ -804,28 +816,6 @@ public DocumentImpl getDocumentWithLock(final DBBroker broker, final XmldbURI ur
         }
     }
 
-    public DocumentImpl getDocumentNoLock(final DBBroker broker, final String rawPath) throws PermissionDeniedException {
```
dizzzz: is removing a good idea?

adamretter: I could have marked it deprecated... but it is unused in the eXist codebase, so it seemed more sensible to remove it.

shabanovd: it should be `Deprecated` one release as minimum. Let's follow some rules, not just removing methods because they looks wrong.

dizzzz:
The commits make the code more readable, nice!
I guess -since these hash maps are in the core- this need to be tested thoroughly.

On question: I see some exceptions changed (matching the underlying hash* object). can this hurt? might be an incompatible contract change...

src/org/exist/dom/persistent/QNamePool.java
```diff
@@ -113,7 +113,7 @@ private void clear() {
         items = 0;
     }
 
-    private QName insert(final QName value) throws HashtableOverflowException {
+    private QName insert(final QName value) throws HashSetOverflowException {
```
dizzzz: this is one of them

src/org/exist/dom/persistent/QNamePool.java
```diff
@@ -216,6 +216,6 @@ private static boolean equals(final QName qname, final String otherLocalPart, fi
 
     @Override
     public Iterator<QName> iterator() {
-        return null;
+        throw new UnsupportedOperationException();
```
dizzzz: here an other one; exception in stead of a value ; for sure a NPE is avoided :-)

src/org/exist/util/hashtable/Object2LongHashMap.java
```diff
@@ -21,295 +21,316 @@
...
 package org.exist.util.hashtable;
 
+import net.jcip.annotations.NotThreadSafe;
+
 import java.util.Iterator;
 
 /**
  * A hashtable which maps object keys to long values.
- * 
+ * <p>
```
shabanovd: where `</p>`?

src/org/exist/dom/persistent/QNamePool.java
adamretter: I just renamed the class so it is named after what it is

shabanovd: is it for internal use only? or can be used at external code too?

src/org/exist/dom/persistent/DefaultDocumentSet.java
```diff
@@ -38,7 +38,7 @@
 
 /**
  * Manages a set of documents.
- *
+ * <p>
```
shabanovd: where `</p>`?

src/org/exist/dom/persistent/QNamePool.java
adamretter: @shabanovd I would not generally use these, and would actually like to move away from them to more standardised collections

adamretter:
@shabanovd I corrected those `<p>` that you asked for, and I also reinstated the methods I deleted (now marked as deleted).

PR #1012, opened 2016-05-17
Add method Collection.removeResource

shabanovd:
Collection.removeResource(Txn txn, DBBroker broker, DocumentImpl doc)

### DIFF
--- a/src/org/exist/collections/Collection.java
+++ b/src/org/exist/collections/Collection.java
@@ -1096,6 +1096,20 @@ public class Collection extends Observable implements Comparable<Collection>, Ca
         }
     }
 
+    public void removeResource(Txn txn, DBBroker broker, DocumentImpl doc) throws PermissionDeniedException, LockException, IOException, TriggerException {
+        if (doc == null) return;
+
+        if (doc.getCollection() != this) {
+            throw new PermissionDeniedException("illegal: doc '"+doc.getURI()+"' does not belong col '"+getURI()+"'.");
+        }
+
+        if(doc.getResourceType() == DocumentImpl.BINARY_FILE) {
+            removeBinaryResource(txn, broker, doc);
+        } else {
+            removeXMLResource(txn, broker, doc.getFileURI());
+        }
+    }
+
     /**
      * Remove the specified document from the collection.
      *


src/org/exist/collections/Collection.java
```diff
@@ -1096,6 +1096,20 @@ public void removeCollection(final DBBroker broker, final XmldbURI name) throws
         }
     }
 
+    public void removeResource(Txn txn, DBBroker broker, DocumentImpl doc) throws PermissionDeniedException, LockException, IOException, TriggerException {
+        if (doc == null) return;
```
adamretter: Consider `Objects.requireNonNull`

src/org/exist/collections/Collection.java
```diff
@@ -1096,6 +1096,20 @@ public void removeCollection(final DBBroker broker, final XmldbURI name) throws
...
+
+        if (doc.getCollection() != this) {
```
adamretter: Can you remove this check please? It is guarding against bad application logic, this should not be a concern of eXist.

adamretter:
Two small comments, otherwise it looks good to me :-)

src/org/exist/collections/Collection.java
shabanovd: as far as DocumentImpl allowed outside eXist that must stay here because document can be moved concurrently between get document object and call `removeResource`. And why you don't want to help to detect such bad situation?

src/org/exist/collections/Collection.java
shabanovd: `Objects.requireNonNull` throw NullPointerException, that will change API behavior

adamretter: This is new behaviour, how can it change something that didn't exist before?

src/org/exist/collections/Collection.java
adamretter: Sorry I don't understand your response - How is `DocumentImpl` not already outside this method? You have to pass it in to this method!

Inside eXist itself this should never happen, if someone is calling this and has that problem in their app, they have a broken app and should solve that there... no?

src/org/exist/collections/Collection.java
shabanovd: Code that use that method more than a year old

src/org/exist/collections/Collection.java
shabanovd: because `removeResource` method is public and `DocumentImpl` come in from outside. Also, because there is a time interval between getting `DocumentImpl` and passing it to that method when it have to lock. The getting `DocumentImpl` with read lock lead to deadlock because it break locking logic that collection must be locked 1st and document 2nd.

adamretter: @shabanovd I have no idea how what you just said is related to the change I asked for!

src/org/exist/collections/Collection.java
adamretter: @shabanovd What? There can't be any code in eXist that uses that method, because you are adding a new method!

dizzzz:
so?

adamretter:
@dizzzz Personally, I am not keen to merge until @shabanovd has answered my questions

shabanovd:
@adamretter I did, nothing to add

dizzzz:
"Closed with unmerged commits
This pull request is closed, but the api/Collection.removeDocument branch has unmerged commits." ??

adamretter:
@dizzzz I think it is because @shabanovd pushed the `api/Collection.removeDocument` to eXist's GitHub and not his own

PR #1013, opened 2016-05-17
Add method DBBroker.reindexXMLResource(Txn txn, DocumentImpl doc)

shabanovd:


src/org/exist/storage/NativeBroker.java
```diff
@@ -2819,6 +2819,10 @@ public int getNextResourceId(final Txn transaction, final Collection collection)
         return nextDocId;
     }
 
+    public void reindexXMLResource(Txn txn, DocumentImpl doc) {
```
adamretter: Needs `@Override` annotation and please make parameters `final`

shabanovd: done

PR #1016, opened 2016-05-18
Replace jsw with yajsw

wolfgangmm:
Replace Tanuki's java service wrapper (jsw) with yajsw, which

* is actively maintained
* includes win64 support out of the box
* has a clear licensing compatible with eXist (Apache 2 instead of GPL mix)
* provides better integration on all supported systems without having to call shell/bat scripts
* properly handles UAC on windows
* is largely compatible with jsw concerning configuration

On windows, using the wrapper is the only reliable way of running eXist. eXist's desktop launcher will thus recommend to install the service and handles the setup automatically.

Other enhancements:

* the service can be managed via the system tray icon on all supported systems, making it easier for new users to get started (no messing around with shell scripts). It detects if eXist is installed as service
* the system configuration dialog now correctly sets wrapper configuration parameters and jetty ports (as requested by many users)

The wrapper has been extensively tested on OSX 10.11.4, Windows 10 64bit, and Ubuntu 15.10. Please help testing other Linux/Unix variants.


dizzzz:
Nice !!!

Regards

Dannes

--
www.exist-db.org

shabanovd:
Is it possible to have additional configuration file outside installation location? That will simplify maintenance of different environments (upgrade procedure).

src/org/exist/launcher/LauncherWrapper.java
```diff
@@ -59,45 +66,50 @@ public LauncherWrapper(String command) {
...
+
+        final OperatingSystem os = OperatingSystem.instance();
+        final ProcessManager pm = os.processManagerInstance();
+        final Process process = pm.createProcess();
+        final String cmdLine = getJavaCmd() + getJavaOpts(home, vmProperties) + " -jar start.jar " + command;
+        process.setWorkingDir(home);
+        process.setVisible(false);
+        process.setPipeStreams(false, false);
+        process.setCommand(cmdLine);
+        System.out.println(cmdLine);
```
shabanovd: is this for debug needs?

wolfgangmm:
@shabanovd You can change tools/yajsw/bin/setenv.sh to read wrapper.conf from a different location. You could potentially also change the location of conf.xml, jetty.xml etc., but since there are many ways to launch eXist, this is out of scope for this PR.

src/org/exist/launcher/LauncherWrapper.java
wolfgangmm: yes, I could not test every system, so it's good to have the actual command line printed out.

shabanovd:
@wolfgangmm found this [include](http://yajsw.sourceforge.net/YAJSW%20Configuration%20Parameters.html#mozTocId174362)

lgtm

ljo:
Yes, the conditional and default configuration  includes are good. I have also tested this yajsw switch on GNU/Debian and Fedora with systemd and CentOS with supervisord control respectively to find good userland configurations for those.

joewiz:
Sounds like a huge improvement over the old wrapper in so many ways. Thank you @wolfgangmm!

src/org/exist/launcher/LauncherWrapper.java
```diff
@@ -59,45 +66,50 @@ public LauncherWrapper(String command) {
...
 
     public void launch() {
-        launch(true);
+        final String home = System.getProperty("exist.home", ".");
+        final PropertiesConfiguration vmProperties = getVMProperties();
+
+        final OperatingSystem os = OperatingSystem.instance();
+        final ProcessManager pm = os.processManagerInstance();
+        final Process process = pm.createProcess();
+        final String cmdLine = getJavaCmd() + getJavaOpts(home, vmProperties) + " -jar start.jar " + command;
```
joewiz: @wolfgangmm This?

dizzzz: hmmm looks like it

PR #1018, opened 2016-05-20
Fix broken JAAS Login

adamretter:


shabanovd:
I really have hard time to understand that change, mean `getSubject` to `getCurrentSubject`. I going to merge only base current code is broken.

adamretter:
@shabanovd It was renamed because a broker keeps a history of subject changes now rather than just a single subject.

shabanovd:
@adamretter I don't see any other public method with "subject" ... and for me `getSubject` mean current effective subject. That is my opinion.

PR #1027, opened 2016-05-28
Fix xupdate attribute manipulation index processing flow

shabanovd:


src/org/exist/indexing/IndexController.java
```diff
@@ -202,22 +208,33 @@ public void removeCollection(final Collection collection, final DBBroker broker,
...
+            StreamListener stream = getStreamListener(node.getOwnerDocument(), mode);
+            try {
+                stream.startIndexDocument(txn);
+
+                IndexUtils.scanNode(broker, txn, node, listener);
+            } finally {
+                stream.endIndexDocument(txn);
+            }
+
+            //that must be called outside this method to avoid slowdown
```
adamretter: Are you saying that the called to `reindex` is now responsible for calling flush? If so I think you should add this to the javadoc of the method to make it clear.

adamretter:
@shabanovd Just the one small comment inline...

src/org/exist/indexing/IndexController.java
shabanovd: That method to "internal" use, called from `ElementImpl` only. Also flush is happening at sync event on transaction release (and that only place where sync/flush must happen to avoid slowdown). I can't make this method protected because of poor packages design. Do you feel that I still should at comment to fava doc?

adamretter: I don't think that is the case; In your changes, `NativeBroker#flush()` is still called from many of those operations, and that they calls `IndexController#flush()`, so the index is still being flushed on many operations. I think you are trying to avoid that?

Also, if I am not mistaken, the problem with moving `flush` to `TransactionManager#processSystemTasks` within eXist is that this will prohibit RYOW on the index within a transaction. Possibly also deferring all the index writes to large batches that will be single-threaded will slow down transaction concurrency.

shabanovd: 'I think you are trying to avoid that?' - yes

On 2nd part you misunderstand me, I do not recommend to move `flush` there, I'm saying that it's there already.

adamretter: @shabanovd Hmm okay.

But... if you are trying to avoid it, then why are you still doing it (e.g. calling IndexController#flush)?

Also, you did not comment on the issue of batching flushes and the performance impact...

shabanovd: `calling IndexController#flush` where?

`you did not comment on the issue of batching flushes and the performance impact` - what do you mean?

adamretter: 1. `NativeBroker#flush` calls `IndexController#flush` and this still happens from several methods in the index where you haven't removed the calls to `NativeBroker#flush`, i.e. you have only removed some of the calls. 

2. Previously `flush` was called directly from several of the index operations. So now instead of lots of little flushes, we have one big flush when `TransactionManager#processSystemTasks` is eventually called. This has two side-effects:
  1. You cannot RYOW on the index changes until the transaction commits. This could cause data visibility issues.
  2. There is a change in the performance profile, as instead of many small writes we are now doing one big write. I do not know if the change is positive or negative, but I would like to think as you are proposing these changes, you have studied this, and can tell us whether it is positive or negative and what the impact is?

shabanovd: by your arguments I can't understand what are you for -)

`IndexController#flush` is not require concrete here because `NativeBroker#flush` after this method call ([for example](https://github.com/eXist-db/exist/blob/develop/src/org/exist/dom/persistent/ElementImpl.java#L559), so why we start discuss bigger issue for this small change?

src/org/exist/dom/persistent/ElementImpl.java
```diff
@@ -1126,7 +1129,14 @@ private String toString(final boolean top, final Set<String> namespaces) {
                 .append(nodeName.getNamespaceURI())
                 .append("\" ");
         }
-        final NodeList childNodes = getChildNodes();
+
+        //in same cases getChildNodes raise exception (unsaved, for example)
```
adamretter: s/same/some

src/org/exist/dom/persistent/ElementImpl.java
```diff
@@ -1126,7 +1129,14 @@ private String toString(final boolean top, final Set<String> namespaces) {
...
+        NodeList childNodes;
+        try {
+            childNodes = getChildNodes();
+        } catch (Exception ex) {
+            childNodes = new NodeListImpl();
```
adamretter: Please at least log the exception!

src/org/exist/dom/persistent/ElementImpl.java
```diff
@@ -1379,22 +1389,26 @@ public IStoredNode updateChild(final Txn transaction, final Node oldChild, final
         final NodePath oldPath = oldNode.getPath(currentPath);
 
         try(final DBBroker broker = ownerDocument.getBrokerPool().getBroker()) {
-            //May help getReindexRoot() to make some useful things
-            broker.getIndexController().setDocument(ownerDocument);
+            IndexController indexes = broker.getIndexController();
+
```
adamretter: why remove the call to `IndexController#setDocument`?

src/org/exist/dom/persistent/ElementImpl.java
```diff
@@ -1427,16 +1444,16 @@ public Node removeChild(final Txn transaction, final Node oldChild) throws DOMEx
 
         final NodePath oldPath = oldNode.getPath();
         try(final DBBroker broker = ownerDocument.getBrokerPool().getBroker()) {
-            //May help getReindexRoot() to make some useful things
-            broker.getIndexController().setDocument(ownerDocument);
-            final IStoredNode reindexRoot = broker.getIndexController().getReindexRoot(oldNode, oldPath, false);
-            broker.getIndexController().setMode(ReindexMode.REMOVE_SOME_NODES);
+            IndexController indexes = broker.getIndexController();
+
```
adamretter: why remove the call to IndexController#setDocument?

src/org/exist/dom/persistent/ElementImpl.java
```diff
@@ -1446,10 +1463,12 @@ public Node removeChild(final Txn transaction, final Node oldChild) throws DOMEx
...
             setDirty(true);
             broker.updateNode(transaction, this, false);
-            broker.flush();
+
             if(reindexRoot != null && !reindexRoot.getNodeId().equals(oldNode.getNodeId())) {
-                broker.getIndexController().reindex(transaction, reindexRoot, ReindexMode.STORE);
+                indexes.reindex(transaction, reindexRoot, ReindexMode.STORE);
             }
+
+            broker.flush();
```
adamretter: Will be 1 large flush, instead of 1 small one and a subsequent one at sync?

src/org/exist/indexing/IndexManager.java
```diff
@@ -111,11 +109,20 @@ private AbstractIndex initIndex(BrokerPool pool, String id, Element config, Path
...
-        if (LOG.isInfoEnabled())
-            {LOG.info("Registered index " + index.getClass() + " as " + index.getIndexId());}
+        if (LOG.isInfoEnabled()) {
+            LOG.info("Registered index " + index.getClass() + " as " + index.getIndexId());
+        }
         return index;
     }
 
+    public void unregisterIndex(Index index) throws DBException {
+        indexers.remove(index.getIndexId(), index);
```
adamretter: Should we call `Index#close` or should we expect the caller to have already called it?

src/org/exist/dom/persistent/ElementImpl.java
shabanovd: replaced by `listener = indexes.getStreamListener(ownerDocument, ReindexMode.REMOVE_SOME_NODES);`

src/org/exist/dom/persistent/ElementImpl.java
shabanovd: nope, move `broker.flush();` from 3 lines above

src/org/exist/dom/persistent/ElementImpl.java
shabanovd: because it was not required and overwrite by `indexes.reindex`

src/org/exist/dom/persistent/ElementImpl.java
shabanovd: that used for debug only, why log exception?

src/org/exist/dom/persistent/ElementImpl.java
adamretter: But that only happens if the `reindexRoot != null`

src/org/exist/indexing/IndexController.java
```diff
@@ -300,12 +299,12 @@ public StreamListener getStreamListener(final DocumentImpl doc, final ReindexMod
      */
     public StreamListener getStreamListener() {
         if (listener != null) {
-            StreamListener next = listener;
-            while (next != null) {
-                // wolf: setDocument() should have been called before
-                // next.getWorker().setDocument(currentDoc, currentMode);
-                next = next.getNextInChain();
-            }
+//            StreamListener next = listener;
```
adamretter: Why is `next` no longer updated?

src/org/exist/dom/persistent/ElementImpl.java
shabanovd:
```
listener = null;
indexes.reindex(transaction, reindexRoot, ReindexMode.REMOVE_SOME_NODES);
```

`indexes.reindex` have it owen call of `IndexController.setDocument` here
`StreamListener stream = getStreamListener(node.getOwnerDocument(), mode);`

src/org/exist/indexing/IndexController.java
shabanovd: for what? that is local variable. That block did `next.getWorker().setDocument`, without that it do nothing.

adamretter:
@shabanovd I reworked this and merged it. I put everything in apart from the `flush` changes, feel free to send a new PR with tests if you really need those and I will review it.

shabanovd:
@adamretter why should I?

It was fix of critical bug that lead to index corruption. I spend time on it and you just threw it away.

Feel free to fix it yourself now.

@wolfgangmm  @ljo @dizzzz Question: why one person can destroy someone PR and merge it to develop branch without discussing?

I spend two weeks to find out why index was corrupted during xupdate operation. Result of it was this [change](https://github.com/eXist-db/exist/pull/1027/commits/c5cf2573bbdaf2c02856f7ccf3a9af121f923473#diff-a73325b5effe32a1a12c8ceb0017438dR1494) and @adamretter threw it just because he can't understand it. Is it right step forward?

Let discuss PR review procedure. Right now it's a mess!

wolfgangmm:
@shabanovd if it was a misunderstanding, the merge should be undone. It is always possible a merge is not applied correctly or the reviewer oversees something critical. It happened to me as well before.

In general I think we should try harder to explain the purpose of a PR in some detail. Does it contribute to the final goal to get to a stable 3.0 version? If not, is it still important enough to be merged in? It would also help if PRs were limited to the actual bug fix, so reviewers don't need to dig through hundreds of lines to figure out where the issue is. If I can judge from the code that a PR won't have negative effects on stability and performance, I will usually merge it in quickly. If not, reviewing takes longer.

At the moment I simply hold back a lot of stuff, waiting for the main showstoppers to be fixed. Maybe we should start having a separate release branch for 3.0?

shabanovd:
@wolfgangmm I'm agree that PR must have some details. And always trying to give as better description of the change as possible. But here we have totally different problem: @adamretter create commit 1c13579e685c58597a95b289aa1e6f53c569ddac of *his interpretation* of my changes and without any reviewing he did merge it to develop branch. Where button to revert it?

This PR have two type of changes: code reformat and actual bug fix. Bugfix was impossible without code reformat because it improve readability and simplify understanding and reasoning.

I was ready and open to any discussion including skype but I totally shocked by @adamretter actions. And that is happening not for 1st time -(

dizzzz:
@shabanovd sorry to hear that your precious work an research got lost, that is a bad thing; I find it quite difficult though to review some PRs , sometimes it is because the reformatting in a lot of files. I understand it is needed sometimes, I had several occasions as well :-(

@wolfgangmm yes, it might be a good idea to have a (3.0) release branch, but it should be short lived. I guess we should do that, directly after @ljo 's WS works.....

dizzzz:
@wolfgangmm a release branch... that is a typical way of working no? 

image

adamretter:
There are two sides to this story! Regardless I spent almost two days on this trying to help Dmitry and do the best thing for both him and the community. I can explain in detail if really required; Otherwise I think my energy is better put into other things....

dizzzz:
@adamretter Ah sorry dont get me wrong! I do not want to make any judgement! The truth is always in the middle :)

What I do see is that github/mailing list is not the most effective way of communicating.... And then things might not go as good as we would like to!

adamretter:
@dizzzz I completely agree that communication was one of the key problems here

ljo:
@dizzzz Yes, I will have tomorrow off to get the WS intermediate PR into an acceptable state to merge. So a release branch would be good (and communication of course).

shabanovd:
@adamretter @wolfgangmm @dizzzz @ljo so, result of all discussion here that eXist community do not interesting in bugfix of xupdate operation that lead to index corruption, right?

wolfgangmm:
@shabanovd I think everyone's interested to get your bug fix in, despite some misunderstanding. We agreed something went wrong with your pull request. Obviously we cannot leave it as is because you say it misses your main fix. As I understood you, the PR either 1) needs to be rolled back and resubmitted, 2) you send another one to address the deficiencies, or 3) you tell me how to fix it.

shabanovd:
@wolfgangmm I do not see "revert" button here and when I create new PR it say "Can’t automatically merge", so option one is not available. There no big difference between 2 and 3. I'll redo this PR with hope that it was last time when PR request was merge in partly. That just increase time consumption because now I have to review that "partly" merge and manually apply changes.

PR #1038, opened 2016-06-06
create new serializer each time to avoid concurrent use by same thread

shabanovd:


caseydawsonjordan:
This occurs when a SAX handler is being used that at some point during processing invokes another sax handler to process content from another document. For instance, processing something like an xinclude. The way it currently works, the serializer would be shared by both serialization processes leading to errors. This fixes that by ensuring each serialization process uses its own serializer.

caseydawsonjordan:
I should also note that this has been used in our code in production for a year or so.

adamretter:
@shabanovd Looks good, but can we have a Unit test to prevent future regressions please?

shabanovd:
@adamretter I have no idea how test will help here. It's not possible to create test for any possible case because code size of tests will be much bigger base code size ... bigger by huge magnitude.

adamretter:
@shabanovd Can you explain to me exactly how this issue would be triggered? And then perhaps between us we can figure out how we could create a test for it...

shabanovd:
@adamretter deadlock that eXist much critical, so I going to spend time on it. Here, if you like, issue you can reproduce if you create serializer that also use LocalXMLResource to insert some data

adamretter:
@shabanovd I can't understand from what you have told me, how I should reproduce the issue that this bugfix is meant to solve. Can you please tell me numbered steps to reproduce this?

@caseydawsonjordan Any change you could help explain how to reproduce?

shabanovd:
@adamretter @caseydawsonjordan did try to explain it above.

Here simple steps:
 - create `CustomMatchListener` that use `LocalXMLResource` at `startElement`
 - setup to use that `CustomMatchListener` as custom filter
 - serialize resource by `LocalXMLResource.getContentAsSAX`

adamretter:
@shabanovd Thanks I will take a look

adamretter:
@shabanovd The tests were kindly provided by @caseydawsonjordan. Your commit was added unchanged

PR #1039, opened 2016-06-07
no needs to read db for collection url because it's immutable

shabanovd:
it also avoid deadlock condition and speedup

### DIFF
--- a/src/org/exist/xmldb/LocalCollection.java
+++ b/src/org/exist/xmldb/LocalCollection.java
@@ -249,7 +249,7 @@ public class LocalCollection extends AbstractLocal implements CollectionImpl {
 
     @Override
     public String getName() throws XMLDBException {
-        return this.<String>read().apply((collection, broker, transaction) -> collection.getURI().toString());
+        return path.toString();
     }
 
     @Override


src/org/exist/xmldb/LocalCollection.java
```diff
@@ -249,7 +249,7 @@ public int getChildCollectionCount() throws XMLDBException {
 
     @Override
     public String getName() throws XMLDBException {
-        return this.<String>read().apply((collection, broker, transaction) -> collection.getURI().toString());
```
adamretter: This will throw an `XMLDBException(ErrorCodes.INVALID_COLLECTION, ...` if the `Collection` was removed from the server, whereas after your change it is possible to call `getName()` on a Collection which no longer exists.

shabanovd: @adamretter , you are fixing two functionalities into one method: `isExist` and `getUrl`. If collection does not exist it's not possible to get url.

shabanovd: I can't find records after 4 months.

Do you agree that you mixing up two semantically different methods into one? There are no way to prepare "good" error message for LocalCollection that do not reference to non existing collection because no what to get that collection url. That is bad design, IMHO.

adamretter:
Looks like this behaviour has already been merged. Thanks @shabanovd

PR #1043, opened 2016-06-09
[bugfix] fixed configuration of log4j2 for WAR distribution

rvdb:
- download `log4j-web.jar` to a temporary folder; include in WAR; delete temporary folder
  NOTE: since `log4j-web.jar` must have the same release version of the log4j files in the eXist distribution, I've used inline properties for declaring the version number; there are probably better ways
- change path to the log folder in the generated `log4j2.xml` configuration file

### DIFF
--- a/build/scripts/dist-war-log4j.xsl
+++ b/build/scripts/dist-war-log4j.xsl
@@ -4,6 +4,10 @@
 
     <!-- Convert log4j2.xml for use in war-file -->
     <xsl:output method="xml"/>
+  
+    <xsl:template match="Property[@name='logs']">
+        <Property name="logs">${web:rootDir}/WEB-INF/logs</Property>
+    </xsl:template>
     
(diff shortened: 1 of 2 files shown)


dizzzz:
what about downloading from http://central.maven.org/maven2/org/apache/logging/log4j/log4j-web/2.6/log4j-web-2.6.jar ?

rvdb:
Ah, perfect, I wasn't aware the jars are available there individually. Sounds good to me, then the `fetch` Ant task can be replaced with `get`. Let me know if I should change something.

adamretter:
I have a pending pr which build on @rvdb which I will send shortly

dizzzz:
@adamretter so... what is the plan? leave this PR there?

adamretter:
@dizzzz Just manually merged the PR from @rvdb with a small patch

PR #1044, opened 2016-06-09
Further tidy up of the Native Value Index

adamretter:
This code mainly replaces a few sets of `int` constants in `org.exist.xquery.Constants` with `Enum` equivalents. This makes ensuring that the correct values are passed a compile time issue, and also makes debugging in-flight options of the Native Value Index much much easier.

Apart from the `int` -> `Enum` changes, there are no other major code changes.


src/org/exist/xquery/GeneralComparison.java
```diff
@@ -1171,32 +1172,28 @@ protected void switchOperands()
         context.getProfiler().message( this, Profiler.OPTIMIZATIONS, "OPTIMIZATION", "Switching operands" );
 
         //Invert relation
-        switch( relation ) {
+        switch(relation) {
 
-            case Constants.GT: {
-                relation = Constants.LT;
+            case GT:
+                relation = Comparison.LT;
```
shabanovd: remove `Comparison.` here too?

src/org/exist/xquery/parser/XQueryParser.java
```diff
@@ -84,7 +84,7 @@ public boolean foundErrors() {
 	}
 
 	public String getErrorMessage() {
-		StringBuilder buf= new StringBuilder();
+		StringBuffer buf= new StringBuffer();
```
shabanovd: why use `StringBuffer` here? It used by a single thread, no?

src/org/exist/xquery/parser/XQueryTreeParser.java
```diff
@@ -80,7 +81,7 @@ public boolean foundErrors() {
 	}
 
 	public String getErrorMessage() {
-		StringBuilder buf= new StringBuilder();
+		StringBuffer buf= new StringBuffer();
```
shabanovd: why switch to `StringBuffer`?

src/org/exist/xquery/value/DurationValue.java
```diff
@@ -362,7 +364,7 @@ public boolean effectiveBooleanValue() throws XPathException {
     
     public static boolean areReallyEqual(Duration duration1, Duration duration2) {
     	final boolean secondsEqual = zeroIfNull((BigDecimal)duration1.getField(DatatypeConstants.SECONDS)).compareTo(
```
shabanovd: `zeroIfNull` can return `null`

src/org/exist/xquery/Constants.java
```diff
@@ -143,7 +143,6 @@
 	 * Identity operators
 	 */
 	public final static int IS = 14;
-	public final static int ISNOT = 15;
```
shabanovd: is it possible to reuse `15`?

src/org/exist/xquery/GeneralComparison.java
adamretter: No, that won't compile.

src/org/exist/xquery/parser/XQueryParser.java
adamretter: This was in the Antlr `XQuery.g` grammar which generates this Java file, I didn't change this previously. However, I just fixed it in a subsequent commit for you

src/org/exist/xquery/parser/XQueryTreeParser.java
adamretter: This was in the Antlr `XQueryTree.g` grammar which generates this Java file, I didn't change this previously. However, I just fixed it in a subsequent commit for you

src/org/exist/xquery/value/DurationValue.java
adamretter: I haven't changed this in my commit. That is out of scope here, feel free to fix in a future PR.

adamretter:
@shabanovd I hope I have now addressed your concerns?

src/org/exist/xquery/GeneralComparison.java
shabanovd: will static import solve that compile errors? You have mix of patterns: short form, for example  `LT` and long one - `Comparison.LT`. Personally, I prefer short form.

shabanovd: that this should be fixed for all enums

src/org/exist/xquery/value/DurationValue.java
shabanovd: so, you see bug and simply ignore it. Ok.

shabanovd:
I'm not agree with styling of this PR, but going to merge it because enums changes required for other tasks. Hope that comments will be addressed in near future

PR #1047, opened 2016-06-15
[bugfix] Make sure to normalize exist.home

adamretter:
Previously `EXIST_HOME` / `exist.home`, and subsequently `jetty.home` and could look like for example:

```bash
[eXist Home : /Users/aretter/code/exist-git/bin/..] 
[jetty.home : /Users/aretter/code/exist-git/bin/../tools/jetty]
```

these changes ensure the paths are normalized, so the paths now correctly look like:

```bash
[eXist Home : /Users/aretter/code/exist-git] 
[jetty.home : /Users/aretter/code/exist-git/tools/jetty]
```

### DIFF
--- a/bin/functions.d/eXist-settings.sh
+++ b/bin/functions.d/eXist-settings.sh
@@ -14,10 +14,14 @@ get_exist_home() {
 		(cd $(/usr/bin/dirname "$p") ; /bin/pwd)
 }
 
+resolve_dir() {
+    (builtin cd `dirname "${1/#~/$HOME}"`'/'`basename "${1/#~/$HOME}"` 2>/dev/null; if [ $? -eq 0 ]; then pwd; fi)
+}
+
 check_exist_home() {
     if [ -z "${EXIST_HOME}" ]; then
 	EXIST_HOME_1=$(get_exist_home "$1");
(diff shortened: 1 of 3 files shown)


bin/functions.d/eXist-settings.sh
```diff
@@ -14,10 +14,14 @@ get_exist_home() {
 		(cd $(/usr/bin/dirname "$p") ; /bin/pwd)
 }
 
+resolve_dir() {
+    (builtin cd `dirname "${1/#~/$HOME}"`'/'`basename "${1/#~/$HOME}"` 2>/dev/null; if [ $? -eq 0 ]; then pwd; fi)
```
dizzzz: Does this work for OS X and Linux ?

adamretter: That's the idea :-)

ljo:
I didn't manage to test this before merge and I have not tested it yet either. So this is just a question based on what we tested for before in this area. Does this normalisation work well across different filesystems? And sometimes people use symbolic links here and there?

dizzzz:
right..... somehow I trust adam t have this verified :-( :-)

adamretter:
It is claimed to work on Mac and Linux with symbolic links by the author. I
haven't tested with links though

PR #1048, opened 2016-06-16
Reintroducing the 'origin' field of a transaction.

dizzzz:
It is used by the Replication extension only;

For now I'd like to restore it; I'd propose to mark it deprecated and find a better way ASAP. We can't throw it away right now :-(

### DIFF
--- a/src/org/exist/storage/txn/Txn.java
+++ b/src/org/exist/storage/txn/Txn.java
@@ -37,6 +37,7 @@ public class Txn implements Transaction {
     private final TransactionManager tm;
     private final long id;
     private State state;
+    private String originId;
 
     private List<LockInfo> locksHeld = new ArrayList<>();
     private List<TxnListener> listeners = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


ljo:
Sounds good.

shabanovd:
@dizzzz Is it for debugging or just informative note?

dizzzz:
it is not for debugging, for the replication extension the info is used to prevent unwanted behaviour (e.g. re-triggering an event)

adamretter:
@wolfgangmm does `TxnListener ` not give you those facilities? I have used it in Granite to attach metadata and actions to a transaction.

@dizzzz Just wondering what `originId` gives you beyond the `Txn#id`, if you need a custom id you could associate it with the `Txn#id` in your own Map... no? or even use a `TxnListener`

dizzzz:
@adamretter I will follow up soon....

shabanovd:
Handle own map each time you need to ignore same transaction looks too complicated. I like this way and I'll use it too.

PR #1056, opened 2016-06-22
Make last-account-id and last-group-id virtual.

shabanovd:
Fields last-account-id and last-group-id did keep state of last id. That bring in few problems. For example, that state go to backup and restore to none empty database lead to state corruption.

That PR make it virtual and calculate it runtime. Next PR will fix ids conflict.

### DIFF
--- a/src/org/exist/security/AbstractRealm.java
+++ b/src/org/exist/security/AbstractRealm.java
@@ -1,23 +1,21 @@
 /*
  *  eXist Open Source Native XML Database
- *  Copyright (C) 2010-2011 The eXist Project
+ *  Copyright (C) 2001-2016 The eXist Project
  *  http://exist-db.org
- *  
+ *
  *  This program is free software; you can redistribute it and/or
  *  modify it under the terms of the GNU Lesser General Public License
  *  as published by the Free Software Foundation; either version 2
  *  of the License, or (at your option) any later version.
- *  
+ *
  *  This program is distributed in the hope that it will be useful,
  *  but WITHOUT ANY WARRANTY; without even the implied warranty of
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  *  GNU Lesser General Public License for more details.
- *  
- *  You should have received a copy of the GNU Lesser General Public License
- *  along with this program; if not, write to the Free Software
- *  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
- *  
- *  $Id$
+ *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  */
 package org.exist.security;
 
@@ -65,7 +63,7 @@ public abstract class AbstractRealm implements Realm, Configurable {
     protected final PrincipalDbByName<Group> groupsByName = new PrincipalDbByName<>();
     
 
-    private SecurityManager sm;
+    private SecurityManagerImpl sm;
 
     protected Configuration configuration;
 
@@ -75,7 +73,7 @@ public abstract class AbstractRealm implements Realm, Configurable {
     protected Collection collectionRemovedAccounts = null;
     protected Collection collectionRemovedGroups = null;
 	
-    public AbstractRealm(SecurityManager sm, Configuration config) {
+    public AbstractRealm(SecurityManagerImpl sm, Configuration config) {
         this.sm = sm;
         this.configuration = Configurator.configure(this, config);
     }
@@ -86,7 +84,7 @@ public abstract class AbstractRealm implements Realm, Configurable {
     }
 
     @Override
-    public SecurityManager getSecurityManager() {
+    public SecurityManagerImpl getSecurityManager() {
         return sm;
     }
     
@@ -126,7 +124,7 @@ public abstract class AbstractRealm implements Realm, Configurable {
                         //Group group = instantiateGroup(this, conf);
                         final GroupImpl group = new GroupImpl(r, conf);
 
-                        getSecurityManager().addGroup(group.getId(), group);
+                        getSecurityManager().registerGroup(group);
                         principalDb.put(group.getName(), group);
 
                         //set collection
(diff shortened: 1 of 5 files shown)


shabanovd:
@adamretter can you review this?

src/org/exist/security/internal/SecurityManagerImpl.java
```diff
@@ -753,15 +745,41 @@ private Realm findRealmForRealmId(final String realmId) throws ConfigurationExce
...
+     */
+    public void registerAccount(final Account account) {
+        usersById.modify(principalDb -> {
+
+            final int id = account.getId();
+
+            principalDb.put(id, account);
+
+            if (id < MAX_USER_ID) {
+                lastAccountId = Math.max(lastAccountId, id);
```
adamretter: I can't quite put my finger on it, but taking the Math.max here seems to be either unnecessary or a guard against some other problem which should likely be fixed at the root cause instead.

Can you detail a use-case where `id` might be bigger than `lastAccountId`?

src/org/exist/security/internal/SecurityManagerImpl.java
```diff
@@ -96,14 +97,14 @@
...
 
-    @ConfigurationFieldAsAttribute("last-account-id")
-    protected int lastUserId = 0;
+    public final static int INITIAL_LAST_ACCOUNT_ID = 10;
+    public final static int INITIAL_LAST_GROUP_ID = 10;
 
-    @ConfigurationFieldAsAttribute("last-group-id")
-    protected int lastGroupId = 0;
+    //calculated runtime
+    protected int lastAccountId = INITIAL_LAST_ACCOUNT_ID;
```
adamretter: It seems to me that this can only be safely accessed with a lambda to `accountsById#modify` to ensure consistency. The current design breaks the encapsulation of the concurrency concerns, perhaps it would be better to move the `lastId` field into `PrincipalDbById` or to flatten out `PrincipalDbById` into `SecurityManagerImpl`?

src/org/exist/security/internal/SecurityManagerImpl.java
```diff
@@ -96,14 +97,14 @@
...
+    protected int lastGroupId = INITIAL_LAST_GROUP_ID;
```
adamretter: Same concerns as `lastAccountId`

src/org/exist/security/AbstractRealm.java
```diff
@@ -65,7 +63,7 @@
     protected final PrincipalDbByName<Group> groupsByName = new PrincipalDbByName<>();
     
 
-    private SecurityManager sm;
+    private SecurityManagerImpl sm;
```
adamretter: Why concrete class? Could you not just modify the interface for what seems to be function naming & argument changes?

test/src/org/exist/security/internal/BackupRestoreSecurityPrincipalsTest.java
```diff
@@ -232,13 +232,12 @@ public void restoreConflictingUsername() throws PermissionDeniedException, EXist
...
-        assertUser(RealmImpl.INITIAL_LAST_ACCOUNT_ID + 2, "jack", ((XMLResource) result.getResource(3)).getContentAsDOM());
-        assertUser(RealmImpl.INITIAL_LAST_ACCOUNT_ID + 4, "joe", ((XMLResource) result.getResource(4)).getContentAsDOM()); //this is `+ 4` because pre-allocating an id skips one
+        assertUser(SecurityManagerImpl.INITIAL_LAST_ACCOUNT_ID + 1, FRANK_USER, ((XMLResource) result.getResource(2)).getContentAsDOM());
+        assertUser(SecurityManagerImpl.INITIAL_LAST_ACCOUNT_ID + 2, JACK_USER, ((XMLResource) result.getResource(3)).getContentAsDOM());
+        assertUser(SecurityManagerImpl.INITIAL_LAST_ACCOUNT_ID + 3, JOE_USER, ((XMLResource) result.getResource(4)).getContentAsDOM());
 
         //check the last user id after the restore
-        result = xqs.query(lastAccountIdQuery);
-        assertEquals(RealmImpl.INITIAL_LAST_ACCOUNT_ID + 4, Integer.parseInt(result.getResource(0).getContent().toString())); //last account id should be that of 'joe'
+        assertEquals(SecurityManagerImpl.INITIAL_LAST_ACCOUNT_ID + 3, sm.lastAccountId); //last account id should be that of 'joe'
```
adamretter: Such access is breaks encapsulation of concurrency concerns.

test/src/org/exist/security/internal/BackupRestoreSecurityPrincipalsTest.java
```diff
@@ -194,14 +197,11 @@ public void restoreConflictingUsername() throws PermissionDeniedException, EXist
...
-        assertUser(RealmImpl.INITIAL_LAST_ACCOUNT_ID + 2, "jack", ((XMLResource) result.getResource(3)).getContentAsDOM());
+        assertUser(SecurityManagerImpl.INITIAL_LAST_ACCOUNT_ID + 1, "frank", ((XMLResource) result.getResource(2)).getContentAsDOM());
+        assertUser(SecurityManagerImpl.INITIAL_LAST_ACCOUNT_ID + 2, "jack", ((XMLResource) result.getResource(3)).getContentAsDOM());
 
         //check the last user id
-        final String lastAccountIdQuery = "declare namespace c = 'http://exist-db.org/Configuration';\n" +
-            "//c:security-manager/string(@last-account-id)";
-        result = xqs.query(lastAccountIdQuery);
-        assertEquals(RealmImpl.INITIAL_LAST_ACCOUNT_ID + 2, Integer.parseInt(result.getResource(0).getContent().toString())); //last account id should be that of 'jack'
+        assertEquals(SecurityManagerImpl.INITIAL_LAST_ACCOUNT_ID + 2, sm.lastAccountId); //last account id should be that of 'jack'
```
adamretter: Such access is breaks encapsulation of concurrency concerns.

src/org/exist/security/internal/SecurityManagerImpl.java
adamretter: `lastAccountId` must be private, otherwise it is possible to read/write it in a non-thread-safe manner from outside of SecurityManagerImpl

PR #1061, opened 2016-07-10
Small fix to ant scripts for installer

wolfgangmm:
Ant scripts to download xars for installer and autodeploy should pass eXist-db version to get the correct app packages matching this version.

### DIFF
--- a/build/scripts/installer.xml
+++ b/build/scripts/installer.xml
@@ -4,34 +4,34 @@
 <!-- ======================================================================= -->
 
 <project basedir="../.." default="installer" name="Build installer">
-  
+
     <description>Build installer</description>
 
 	<!-- import common targets -->
 	<import file="../../build.xml"/>
 	<import file="../../build/scripts/macosx.xml"/>
     <import file="git-support.xml"/>
-    
+
     <taskdef resource="net/sf/antcontrib/antcontrib.properties">
         <classpath>
             <pathelement location="tools/ant/lib/ant-contrib-1.0b3.jar"/>
         </classpath>
     </taskdef>
-    
+
 	<property name="apps.dir" value="installer/apps"/>
     <property name="installer.scripts" value="installer/scripts"/>
     <property name="installer.mac.icon" value="${installer.scripts}/icon.icns"/>
-        
+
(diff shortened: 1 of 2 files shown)


shabanovd:
@wolfgangmm what will happen if it used for custom eXist build?

joewiz:
@shabanovd What do you mean by "custom"?

shabanovd:
@joewiz @wolfgangmm well, better to rephrase: what is rule(s) for matching correct app packages by eXist version?

dizzzz:
as i see it, e.g. the value "3.0.2" is send to the server. each XAR has a minSemver and maxSemVer value, so the server can pickup the correct XAR....

AFAIK this change makes the selection of the correct (matching) XAR much better....

shabanovd:
@dizzzz I'm not against this PR, just need to know how to use it to get latest xar version. Will last xar return if no `processor` defined?

dizzzz:
I have no idea, actually. I except so....  a recent XAR of wolfgang explicitly defines an upperlimit. That would mean...  when no upperlimit is defined.... it is not used. I hope.

joewiz:
This PR passes `processor=${project.version.numeric}` to [the public-repo's find.xql endpoint](https://github.com/eXist-db/public-xar-repo/blob/master/controller.xql#L77-L80), which according to https://github.com/eXist-db/public-xar-repo/blob/master/modules/find.xql#L14, defaults to '2.2.0' if missing. So rather than the newest .xar you'll get the oldest. I think the key is to ensure your `project.version.numeric` corresponds to the desired version of eXist. 

Note that the installer script here (the one in this PR) doesn't pass some of the more specific version constraints—`semver`, `min`, and `max`— which [app:find-version()](https://github.com/eXist-db/public-xar-repo/blob/master/modules/app.xql#L186-L198) gives higher priority to.

wolfgangmm:
It is as @joewiz explained above. The scripts will assume 2.2.0 if no processor version is given. This is to prevent incompatible jars to be loaded into eXist, which would make the database unusable.

shabanovd:
@wolfgangmm yeap, I see that. Is it possible to get latest app somehow?

PR #1062, opened 2016-07-10
Update to Jetty 9

adamretter:
Configuration in Jetty 9 is modular, the following two files control which configuration is loaded for the normal mode of operation or the standalone mode of operation respectively:

* `$EXIST_HOME/tools/jetty/etc/standard.enabled-jetty-configs`

* `$EXIST_HOME/tools/jetty/etc/standalone.enabled-jetty-configs`


joewiz:
Wow, massive undertaking! Thank you! Any known issues or concerns?

adamretter:
If you have customised your jetty config, then you will need to port those changes to jettys new modular config. Likewise if you have a customised web.xml, you may need to reapply those customisations.

Otherwise, it should be a completely transparent upgrade to the user.

It does "feel" faster to me, but I haven't had the time to benchmark it yet.

HTTP2 is available in Jetty 9, but I haven't enabled it by default as it is very dependent on the exact JDK you are running.

dizzzz:
Great work adam! re @HTTP2 having support is really a step forward! As long we document (or refer to documentation) how to switch this on, that would be sufficient.

adamretter:
@dizzzz At the moment is is rather involved - https://www.eclipse.org/jetty/documentation/9.3.x/http2.html

I hope to send a future PR which tries to ease this for eXist users, although, for those of us that are not `Java Developers` it might be easier to wait until Jetty provides a more JDK version agnostic solution.

dizzzz:
I'll have a look at the PR later today (probably in the eve). @ljo @shabanovd @wolfgangmm  your feedback is welcome :-)

JoernT:
agree with Dannes - http2 support is a big thing - however it does not look too easy to implement now. Nevertheless it could be worthwhile to work it out in some configuration/app as this should improve client perforrmance dramatically and remove the need to optimize client-side stuff which takes away quite some work in any application we build.

adamretter:
To be clear, whilst this PR does enable you to configure eXist and Jetty 9 for HTTP 2 by following the Jetty documentation. Actually doing so, or making this easier for the average user, is outside of the scope of my PR. I agree, that it would be desirable to add it in future.

dizzzz:
adam, I agree !

ljo:
@dizzzz started this morning going through the commits.

JoernT:
@adamretter this was already clear. The average user is not so much my focus for now. Before going to that stage we have to work it out for ourselves - which steps need to be done to use it in a single application. From there we can go for a recipe for the average user maybe.

adamretter:
@dizzzz @JoernT Okay understood :-)

tools/jetty/webapps/portal/index.html
```diff
@@ -1,75 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
dizzzz: I don't understand .... this portal is currently in use? it is a simple re-direct with a 'nice' (ahem) logo to the /app URL?

dizzzz:
Ok, have just one remark.... regarding the removed 'portal' page. I mean I have no objections as long "/" redirects effectively to the dashboard....

tools/jetty/webapps/portal/index.html
dizzzz: hmmm are you sure? I had issues with my browser cache in the past...

adamretter: Yes. But... have you tested it? i.e. build my branch, `run bin/startup.sh` and then visit `http://localhost:8080`.

I think/hope you will find the redirect and everything still works fine.

dizzzz: :-( unfortunately I was right :-(

error_404_-_not_found_and_open_ _exist-db__48601_messages__10_unread_

adamretter: Where did you visit? Was it `http://localhost:8080/`?

dizzzz:
For me, http://localhost:8080/exist/apps/dashboard/index.html does not render into a working page. I see HTML code (view source), the 'gray' background is visible though


dannes1

tools/jetty/webapps/portal/index.html
dizzzz: yes, with browser cache empty

adamretter: Interesting... I will try it again here shortly...

dizzzz: ok... please note my other remark regarding the dashboard

adamretter:
@dizzzz You were correct about the browser cache. I reinstated the `portal` webapp and bought it's config files up to date.

I cannot reproduce the issue you saw with the Dashboard, can you check that again with a clean cache now that the portal webapp is fixed.

tools/jetty/etc/enabled-jetty-config
```diff
@@ -18,6 +18,7 @@ jetty-jaas.xml
 
 
 ## Main Server Config
+jetty-gzip.xml
```
shabanovd: Was it tested? It was disabled because of problems it bring. Can this be disabled by default?

shabanovd:
I see too many changes for single PR.

dizzzz:
@adamretter I confirm that you restored the 'portal'; thnx!

But.... still I have no working 'dashboard' : MSIE/Safari/Chrome

dizzzz:
@adamretter ok; I was able to reproduce it with the attached linked ZIP file on windows7 and my mac ....

dizzzz:
@adamretter probably this helps: 
clipboard01

dizzzz:
And you are right about bf:
clipboard02

adamretter:
@dizzzz Yup exactly, those errors are the same I see on my Slackware VM... Now to work out why I don't get them on my Mac and then make that happen for everyone :-)

dizzzz:
can you try with curl on your dev environment (check the css-es)....

adamretter:
@dizzzz Already did :-p Curl also works fine on my Mac, as does Chrome and Firefox with cache cleared. So I am just finding the difference between my Mac and Slackware VM

dizzzz:
really interesting.... maybe a different set of bf-jars? I see betterform-exist-5.1-SNAPSHOT-20160615.jar and dwr-3.0.RC2.jar ...

adamretter:
@dizzzz Nope it is a difference in the `web.xml`. The one generated on a clean base is not quite right, not sure how mine is okay, but I will send a patch shortly.

adamretter:
@dizzzz Okay sorted. The namespace used in the `web.xml` has changed between Servlet 3.0 which was `http://java.sun.com/xml/ns/javaee` and Servlet 3.1 which is `http://xmlns.jcp.org/xml/ns/javaee`.

The transform `extensions/betterform/main/etc/MergeWebXML.xsl`which adds the betterForm extension into our `webapp/WEB-INF/web.xml` needed to be updated. I have now rebased and pushed that.

dizzzz:
nice catch!!!!!!

adamretter:
@dizzzz Hey, you caught it. Thank you. Yes that is the correct branch still.

dizzzz:
did you push/rebase it? I do not see the changes in the webui https://github.com/adamretter/exist/commits/jetty-9-attempt-2

adamretter:
@dizzzz Yes, as mentioned above I rebased and pushed it. I added it into the commit:
> [bugfix] Update the version of betterForm so that is compatible with Jetty 9.3.9

You can see the change here - https://github.com/adamretter/exist/commit/cd27bfb9ebc08940b1a2492ad0f2c3494a2aee01#diff-82f3c4e0cad7bcf8a587b1ec26519795

joewiz:
The fact that rebasing doesn't leave a timestamp makes it tricky to track down such changes. Elegant, but hard to trace!

adamretter:
@joewiz Actually rebasing will change the `commit timestamp` but not the `author timestamp`. GitHub orders by `author timestamp` so it is tricky to see it GitHub. However, GitLab orders by `commit timestamp` so it is more obvious over there (although that can also be confusing as that is not the order in which the commits will be applied). Or something like that...

joewiz:
@adamretter Ah, I see! They should allow the user to toggle display/sorting!

dizzzz:
Ok, rechecked, and...... all systems are go!

joewiz:
One little change I noticed - the console log is slightly different on startup:

```
----------------------------------------------------- 
Server has started, listening on: 
	http://10.20.4.103:8080/ 
	https://10.20.4.103:8443/ 
Configured contexts: 
	/ 
	/exist 
----------------------------------------------------- 
```

I believe it used to say `localhost` rather than the IP address.  No biggie, just something that screenshots/documentation might need to be updated to reflect.

adamretter:
@joewiz Yes there are a few differences in the logging at startup. So the code that reports the server URLs and contexts was taken from Jetty, or recommended by Joakim from Webtide (the guys behind Jetty).

You will find that it still listens on localhost and also that IP. So I think by default Jetty most likely binds to all IPs on your system.

If you would like to update the documentation that would be lovely ;-)

dizzzz:
having the (a) public IP mentioned helped me out already :-)

Issue #1067, opened 2016-07-15
3.0 RC1: fold-right omits the rightmost sequence element in processing


median1st:
The fn:fold-right function omits the rightmost element from the processed sequence. This xquery snippet demonstrates the bug:
```
let $seq := (1,2,3,4)
return
    fold-right ($seq, (<state i='0'/>) , function($item as xs:integer, $accu as node()*) {
        ($accu, <state i='{$item}'/>)
    })
```
eXist-db 3.0 RC1 returns the following sequence as a result:

```
<state i="0"/>
<state i="3"/>
<state i="2"/>
<state i="1"/>

```
Note that the rightmost number ( 4 ) is omitted. When the same query is run in eXist-db 2.2, the result is correct:

```
<state i="0"/>
<state i="4"/>
<state i="3"/>
<state i="2"/>
<state i="1"/>

```

dizzzz:
confirmed for existdb 3.0 (RC2)

dizzzz:
waiting for a pull-in and review

PR #1076, opened 2016-07-30
Repair fold-right

dizzzz:
As reported in https://github.com/eXist-db/exist/issues/1067

### DIFF
--- a/src/org/exist/xquery/value/ValueSequence.java
+++ b/src/org/exist/xquery/value/ValueSequence.java
@@ -793,7 +793,7 @@ public class ValueSequence extends AbstractSequence implements MemoryNodeSet {
 
     private class ReverseValueSequenceIterator implements SequenceIterator {
 
-        private int pos = size - 1;
+        private int pos = size; // size is not the actual size
 
         public ReverseValueSequenceIterator() {
         }
--- a/test/src/xquery/xquery3/suite.xql
+++ b/test/src/xquery/xquery3/suite.xql
@@ -9,5 +9,6 @@ test:suite((
     inspect:module-functions(xs:anyURI("concat.xql")),
     inspect:module-functions(xs:anyURI("groupby.xql")),
     inspect:module-functions(xs:anyURI("flwor.xql")),
-    inspect:module-functions(xs:anyURI("typeswitch.xql"))
+    inspect:module-functions(xs:anyURI("typeswitch.xql")),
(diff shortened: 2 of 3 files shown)


src/org/exist/xquery/value/ValueSequence.java
```diff
@@ -793,7 +793,7 @@ public Item nextItem() {
 
     private class ReverseValueSequenceIterator implements SequenceIterator {
 
-        private int pos = size - 1;
+        private int pos = size; // size is not the actual size
```
shabanovd: what will happen if `size` change during iteration?

test/src/xquery/xquery3/suite.xql
```diff
@@ -9,5 +9,6 @@ test:suite((
     inspect:module-functions(xs:anyURI("concat.xql")),
     inspect:module-functions(xs:anyURI("groupby.xql")),
     inspect:module-functions(xs:anyURI("flwor.xql")),
-    inspect:module-functions(xs:anyURI("typeswitch.xql"))
+    inspect:module-functions(xs:anyURI("typeswitch.xql")),
+    inspect:module-functions(xs:anyURI("fn.xql"))
```
shabanovd: `fn` is poor naming, no?

dizzzz: My idea is to have a script here For all kind of tests on xquery fn functions 

I already spend quite some time in finding a better name. So the intention is not to have folding tests only.

That said.... Which test target does start the xquery tests?

src/org/exist/xquery/value/ValueSequence.java
dizzzz: Can that happen? By spec sequences are immutable right?

shabanovd: It's not immutable by implementation. There number of methods (like `add`, `addAll`, `clear` and etc) that change state.

dizzzz: The intention of the the fix is to solve something different. You have a point here, but I guess it is a bigger implementation/design issue? How can we solve this? (for an other PR).

for `add` and 'addAll' there are no problems, since it extends the internal array only, the Reverse-class is working from end to start.

For  `clear` you are right for sure.....

joewiz:
Thanks @dizzzz! Great work!

PR #1084, opened 2016-08-05
Add Configurator warnings

adamretter:


dizzzz:
Looks good; i'd guess (wrongly) that the removed line has a side effect that is needed? I recall this from the replication extension....

adamretter:
@dizzzz Nope that removed line has no side-effects, it is just old dead code which should have been removed previously when we switched to push/pop the subjects.

Issue #745, opened 2015-07-16
unmatching try/catch occludes outer try/catch


telic:
A try/catch expression that does not catch the particular error type raised in the try's body changes the error in a way that makes it impossible to catch the error in an outer try/catch.

For example:
```xquery
try {
    try {
        error(xs:QName('test'))
    } catch foo {
        'inner catch'
    }
} catch test {
    'outer catch'
}
```
The expected result is 'outer catch', but instead the query raises an "exerr:ERROR Error." error.  This error cannot be caught with `catch exerr:ERROR` (even if the "exerr" namespace is explicitly bound to "http://www.w3.org/2005/xqt-errors), but can be caught with `catch *`.

This example above obviously contrived, but there are cases where a function might choose to handle some exceptions while passing on others to the caller.  If there several different such errors that might be passed, the caller is left with no means of differentiating them.

ljo:
Contrived or not, thanks for reporting this. The spec has changed a bit since its first appearance for many of the version 3 introduced constructs.

PR #1090, opened 2016-08-08
[bugfix] Fix nested try/catch

adamretter:
The inner try/catch shouldn't mess with an XPathException if it doesn't catch it.
Closes https://github.com/eXist-db/exist/issues/745


joewiz:
@adamretter Nice! Thanks!

PR #1095, opened 2016-08-18
Update of #1082 with additional fixes

wolfgangmm:
Enhance #1082 to fix startup triggers and shutdown issue.

### DIFF
--- a/extensions/debuggee/src/org/exist/debuggee/ScriptRunner.java
+++ b/extensions/debuggee/src/org/exist/debuggee/ScriptRunner.java
@@ -68,8 +68,8 @@ public class ScriptRunner implements Runnable, Observer {
 	public void run() {
 		try {
 			final Database db = BrokerPool.getInstance();
-			
-            BrokerPool.registerStatusObserver(this);
+
+            db.addStatusObserver(this);
 			
 			try(final DBBroker broker = db.getBroker()) {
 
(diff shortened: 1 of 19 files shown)


shabanovd:
BTW, eXist visibility "controlled" by jetty startup time, and jetty started up only after eXist finish all initialization. In case of embedded mode accessibility can be controlled application that use it. So, base on that I prefer to add broker-pool instance to pool as soon as it created and remove when initialization fails.

For now one of uncovered case is recovery procedure [`broker.repair()`](https://github.com/wolfgangmm/exist/blob/5f8a488c1d21ce5f9031063b78ceae6d01ad6cbe/src/org/exist/storage/BrokerPool.java#L644), it require indexes to operate in "normal" mode and that may require support for `BrokerPool.getInstance()` calls from different thread (async tasks).

I think all problems come up because of `BrokerPool.getInstance()` itself: it must be accessible in all cases or removed completely.

dizzzz:
I don't think I can judge.....

adamretter:
@wolfgangmm I will take a look later, but I think there must be a cleaner approach (i.e. better separeartion of concerns) we can take here. I will also consider the comments by @shabanovd and see what I come up with

dizzzz:
adam, will you incorporate this in your PR ?

adamretter:
@dizzzz of course!

PR #1098, opened 2016-08-19
Tests that show some bugs in the new range index and optimizer

adamretter:
The weirdest thing IMHO is that the non-optimized queries do a better (but not perfect) job at finding the results.

i.e. `ot:sr-contains-string-collation-without-diacritics-2` finds one match where `rt:sr-contains-string-collation-without-diacritics-2` finds zero matches. I believe they should both find two matches.

I can also confirm that I get the same results on eXist-3.0.RC1, so this isn't a regression.

### DIFF
--- a/extensions/indexes/range/test/src/xquery/optimizer.xql
+++ b/extensions/indexes/range/test/src/xquery/optimizer.xql
@@ -39,6 +39,25 @@ declare variable $ot:COLLECTION_CONFIG :=
         </index>
     </collection>;
 
+declare variable $ot:SR_COLLECTION_CONFIG :=
+    <collection xmlns="http://exist-db.org/collection-config/1.0">
+        <index xmlns:xs="http://www.w3.org/2001/XMLSchema"
+            xmlns:tei="http://www.tei-c.org/ns/1.0">
+            <range>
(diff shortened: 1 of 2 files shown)


dizzzz:
Difficult to fix; suggestion: document the defect, and release as-is

adamretter:
@dizzzz As long as that doesn't mean "won't fix". @wolfgangmm was looking into this recently for @ttasovac, but I am not sure what his conclusion was?

This issue basically makes the use of full-text indexing on any language that has accents etc (diacriticals) impossible.

dizzzz:
the conclusion is what I understand: it is unfixable due to Lucene issues, upgrade is impossible to to changes/removal of APIs we rely on [you checked on this, right?]

adamretter:
@dizzzz Sorry, I am not sure what you are asking me...

dizzzz:
@wolfgangmm please could you comment? I should not be the messenger for the details :-)

adamretter:
@wolfgangmm I can see that some of the tests from here were rolled into https://github.com/eXist-db/exist/pull/1172. Is there anything left in this PR that we need?

wolfgangmm:
@adamretter No, I moved those tests which still apply into #1172, so we can close this PR.

PR #1099, opened 2016-08-19
fix invalid picture string

joewiz:
as noted at https://github.com/eXist-db/exist/pull/1096#issuecomment-241097001


joewiz:
...sorry about that!

PR #1101, opened 2016-08-19
Fix formatDate timezone presentation

adamretter:


src/org/exist/xquery/value/DurationValue.java
```diff
@@ -211,18 +211,36 @@ protected Duration canonicalZeroDuration() {
...
+			case HOUR:
+				r = duration.getHours();
+				break;
+
+			case MINUTE:
+				r = duration.getMinutes();
+				break;
+
+			case SIGN:
+            	return duration.getSign();
```
shabanovd: wrong spacing here

dizzzz: you mean:

```java
r = duration.getSign();
break;
```

shabanovd: mean `            	return`, it should be `				return` as 3 lines above

dizzzz: sorry I still don't understand.....
reformatting the code shows (hides?) at least there is a kind of interesting flow, which may or may not be intended. At least some comments helps to understand why things are done this way....

adamretter:
@shabanovd @dizzzz I just rebased the formatting into a separate commit. Hopefully that helps you?

shabanovd:
@adamretter yes, it's ok now ... waiting for travis results

dizzzz:
@adamretter thx :-)  but my remark contained an underlying question: is the code really intended to work this way?

clearly it is :-)

joewiz:
Thanks @adamretter for the code and @shabanovd and @dizzzz for your reviews. This will really give eXist users wonderful expressiveness in formatting dates & times.

adamretter:
@dizzzz Ah right sorry. Yes you are right, it was as intended :-)

Issue #1108, opened 2016-09-04
Incorrect results when using maps and wildcard with lookup operator


joewiz:
The XQuery 3.1 spec contains an example of an expression involving a map and lookup operator whose key specifier is a wildcard. eXist's result for this query does not conform to the spec. 

The example is described at https://www.w3.org/TR/xquery-31/#id-lookup:

> If `$m` is bound to the weekdays map described in 3.11.1 Maps, then `$m?*` returns the values `("Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday")`, in implementation-dependent order.

If we construct a query using this [example weekdays map](https://www.w3.org/TR/xquery-31/#id-maps), eXist produces the map's **keys** instead of their values.

**The query:**

```xquery
xquery version "3.1";

let $m := 
    map {
        "Su" : "Sunday",
        "Mo" : "Monday",
        "Tu" : "Tuesday",
        "We" : "Wednesday",
        "Th" : "Thursday",
        "Fr" : "Friday",
        "Sa" : "Saturday"
    }
return
    $m?*
```

**Expected results:**

```xquery
("Sunday","Monday","Tuesday","Wednesday", "Thursday", "Friday","Saturday")
```

**Actual results:** (using eXist develop branch 052c842)

```xquery
("Tu", "Mo", "Su", "Fr", "Sa", "We", "Th")
```

Note that the differing order of the items in the results here is insignificant. The problem I'm reporting is that the spec says the lookup operator whose key specifier is a wildcard, i.e., `?*`, should return **values**, not keys. 

I confirmed the spec's behavior is followed by BaseX 8.5.2 and Saxon 9.7.0.4.

dizzzz:
Closed?

joewiz:
Thanks @dizzzz - it was closed by https://github.com/eXist-db/exist/pull/1118.

dizzzz:
ah ok :-)

PR #1109, opened 2016-09-05
Add test for #1108 - map wildcard lookup

joewiz:
This adds a test for the problem reported in #1108 relating to map wildcard lookups.

### DIFF
--- a/test/src/xquery/maps/maps.xql
+++ b/test/src/xquery/maps/maps.xql
@@ -436,6 +436,7 @@ function mt:lookupWrongType() {
 };
 
 declare 
+    %test:pending
     %test:assertTrue
 function mt:lookupWildcard() {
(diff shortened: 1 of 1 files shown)


joewiz:
It just occurred to me that I should have used `%test:pending` since this test will fail until the problem is fixed. Or not? I'm happy to make this change if needed.

adamretter:
@joewiz Yes it should be marked as `pending` however, you will need to chase then to make sure it isn't forgotten ;-)

ljo:
Yes, I think that is the best. The issue #1108 will still be here to remind us.

joewiz:
@adamretter @ljo Thanks for your guidance. Done, and the CI checks now pass.

joewiz:
@adamretter Thanks for the merge!

PR #1116, opened 2016-09-11
add 'Copy source' ; missing caused NPE for  JMS extension

dizzzz:
Copy source too ; this caused a NPE in combination with the replication extension 

see https://github.com/eXist-db/messaging-replication/issues/84


adamretter:
@dizzzz Is there a test you can add for this?

dizzzz:
@adamretter I am not sure I have an idea how to test ; it is rather an arbitrary change... (yes, these are most dangerous)

PR #1118, opened 2016-09-12
Bugfix lookup wildcard

adamretter:
Fixes an issue where a wildcard lookup on a map would previously return the keys, whereas it should have returned the values.


adamretter:
@joewiz Here you go :-)

joewiz:
@adamretter Thank you for the improved tests—very instructive—and the fix! 

@shabanovd Thank you for the review & merge.

PR #1130, opened 2016-09-20
Serialize attr matches

adamretter:


rvdb:
This is nice! I've tested, and attribute highlighting seems to work elegantly (by inserting into the matching element an `@exist:matches` attribute whose value is the matching attribute name).

Yet, I've noticed how the `"highlight-matches=both"` setting in `util:expand()` still injects the old `|||` match indicators in **all** attributes of an element with a full-text hit. For example, given following input document:

    <p>
      <w xml:id="w1" lemma="this">Nice</w>
      <w xml:id="w2" lemma="commit">commit</w>
    </p>

and appropriate ft index definitions, following query:

    for $a in //w[ft:query(., 'nice')]
    return util:expand($hit, "highlight-matches=both")

...produces: 

    <w xml:id="||||w1||||" lemma="||||nice||||"><exist:match xmlns:exist="http://exist.sourceforge.net/NS/exist">Nice</exist:match></w>

Is this still a leftover that needs to be cleaned up?

joewiz:
@adamretter Agreed with @rvdb that this is nice. I'm also curious - I see in the commit mentions of `@exist:matches-offset` and `@exist:matches-length` attributes. There's a comment that these are suppressed for full text indexes. Does this mean that they'd appear for new range indexes?

PR #1134, opened 2016-09-23
fix XQuery pool deadlock

shabanovd:
deadlock happens because one thread:
- lock XQueryPool
- waiting for CollectionCache lock to validate xquery (document stored at db)
  and 2nd thread:
- lock CollectionCache to create, for example, collection
- on afterCreate trigger event waiting for XQueryPool lock to get xquery script to execute


src/org/exist/storage/XQueryPool.java
```diff
@@ -134,20 +132,17 @@ public void returnCompiledXQuery(final Source source, final CompiledXQuery xquer
...
-                Deque<CompiledXQuery> stack = (Deque<CompiledXQuery>)get(source);
-                if (stack == null) {
-                    stack = new ArrayDeque<>();
-                    source.setCacheTimestamp(System.currentTimeMillis());
-                    put(source, stack);
-                }
+            if (pool.size() < maxPoolSize) {
+                source.setCacheTimestamp(System.currentTimeMillis());
+
+                Deque<CompiledXQuery> stack = pool.computeIfAbsent(source, (k) -> new ArrayDeque<>());
```
adamretter: Could be `final`?

src/org/exist/storage/XQueryPool.java
```diff
@@ -160,78 +155,57 @@ private synchronized void returnObject(final Source source, final CompiledXQuery
...
-            final int idx = getIndex(source);
-            if (idx < 0) {
-                return null;
-            }
-            key = (Source) keys[idx];
-            int validity = key.isValid(broker);
-            if (validity == Source.UNKNOWN) {
-                validity = key.isValid(source);
-            }
+    private Object borrowObject(final Source source) {
```
adamretter: I think the return type of this function could change to `CompiledXQuery` (which would also obsolete some `cast`s) and most likely even be renamed to `borrowCompiledQuery`.

adamretter: Or even merged with `borrowCompiledXQuery` below...

joewiz:
Looks like an important fix. Thank you @shabanovd!

src/org/exist/storage/XQueryPool.java
```diff
@@ -58,6 +56,8 @@
 
     private final static Logger LOG = LogManager.getLogger(XQueryPool.class);
 
+    private ConcurrentMap<Source, Deque<CompiledXQuery>> pool = new ConcurrentHashMap<>();
+
     private long lastTimeOutCheck;
     private long lastTimeOfCleanup;
```
adamretter: Access to `lastTimeOutCheck` and `lastTimeOfCleanup` were previously `synchronized` but now are not. What is the concurrency guarantee of these? I think these are no longer thread safe and most likely need strong concurrency guarantees.

shabanovd: fixed

src/org/exist/storage/XQueryPool.java
shabanovd: fixed

adamretter: @shabanovd Great looks good. Can you squash your fixup commits please into the base commit?

shabanovd: @adamretter why you need this?

adamretter: @shabanovd Because it is good to try and keep as clean a commit history in upstream as possible

shabanovd: @adamretter good for whom? I, as author of this changes, believe that it must stay as it because it clearly define each step of modification.

Also, "keep clean" history of 17k commits sound like paranoia for me (personally).

src/org/exist/storage/XQueryPool.java
adamretter: @shabanovd Okay you now have two `AtomicLong`s if they are used in an invariant together (i.e. in `returnObject`), this can give different results to having a single mutual exclusion around both variables. That might be acceptable, depending on the guarantees that you are offering on consistency. Have you considered this?

adamretter: @shabanovd `returnObject` looks like this, there is an invariant on the `lastTimeOfCleanup` `AtomicLong`:

```java
 private void returnObject(final Source source, final CompiledXQuery xquery) {
        final long ts = source.getCacheTimestamp();
        if (ts == 0 || ts > lastTimeOfCleanup.get()) {
            if (pool.size() >= maxPoolSize) {
                timeoutCheck();
            }
...
```

 Which then calls `timeoutCheck()`, which has an invariant on the`lastTimeOutCheck` `AtomicLong`:

```java
private void timeoutCheck() {
        if (timeoutCheckInterval < 0L) {
            return;
        }

        final long currentTime = System.currentTimeMillis();
        if (currentTime - lastTimeOutCheck.get() < timeoutCheckInterval) {
            return;
```

So you now have two variables, which participate in the same invariant, for which access to each is independently synchronized, where as previously they were both within the same synchronized access block.

As I mentioned, this may be what you intended, but it changes the concurrent consistency guarantees of the class, so I just wanted to make sure, that that was what you had indeed intended?

adamretter: Yes, however in the methods I described, your `time` is effectively `final` with regards to the other invariants so we need not consider it here.

`ConcurrentMap` doesn't have any impact on the `lastTimeOfCleanup` or `timeoutCheckInterval`, that has a different purpose. My point was that two threads where one pre-empts the other could now cause a different outcome on the values of `lastTimeOfCleanup` and `timeoutCheckInterval` because the two things are no longer synchronized together. That was all I was pointing out.

I don't see a problem with Deque, if you want to update it atomically, then just use the `compute...` functions on the `ConcurrentHashMap`.

shabanovd: `ArrayDeque` is not thread-safe, so why you don't see problems? [javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/ArrayDeque.html)

Again, I really can't understand you problem with `lastTimeOfCleanup` and `lastTimeOfCleanup`. It (mostly) always just increasing, and all check in same direction `>`.

shabanovd: `compute` methods of `ConcurrentHashMap` do not guarantee atomic access, so it'll not solve issue

src/org/exist/storage/XQueryPool.java
```diff
@@ -146,23 +147,19 @@ private void returnObject(final Source source, final CompiledXQuery xquery) {
...
-                final Deque<CompiledXQuery> stack = pool.computeIfAbsent(source, (k) -> new ArrayDeque<>());
+                final Set<CompiledXQuery> stack = pool.computeIfAbsent(source, (k) -> new ConcurrentSkipListSet<>());
 
                 if (stack.size() < maxStackSize) {
-                    // check if the query is already in pool before adding,
-                    // may happen for modules, don't add it a second time!
-                    if(!stack.contains(xquery)) {
-                        stack.push(xquery);
-                    }
+                    stack.add(xquery);
```
adamretter: Okay but now you have introduced a different concurrency issue, which is that of duplicated unnecessary work as you have two concurrent structures which are not synchronized as a unit, this is similar to the change in consistency that you introduced by adding two AtomicLong.

Consider this scenario:

1) `Thread A` executes line 150 and creates a new `ConcurrentSkipListSet` in the `pool`.
2) `Thread A` executes line 152 and enters the conditional.
3) `Thread A` is then suspended (unscheduled).
4) `Thread B` is scheduled, and executes line 150, returning the `stack` created by `Thread A`.
5) `Thread B` then executes lines 152 and 153, adding an `xquery` to the stack.
6) `Thread A` is resumed (scheduled) and now also adds the same `xquery` to the stack which `Thread B` previously added

So one issue is that in steps (5) and (6) you are repeating work when there are concurrent threads.

You also need to be aware that calling `ConcurrentSkipListSet#size()` is both an expensive operation and can be inaccurate unless care is taken to synchronize access to it.

I think there is a simpler design possible here which is also safer:

```java
private ConcurrentMap<Source, Deque<CompiledXQuery>> pool = new ConcurrentHashMap<>();
```

```java
pool.compute(source, (k, v) -> {
   if(v == null) {
      v = new ArrayDeque<>();
      v.add(xquery);
   } else if(v.size() < maxStackSize()) {
      v.add(xquery);
   }
   return v;
});  		  
```


final Set<CompiledXQuery> stack = pool.computeIfAbsent(source, (k) -> new ConcurrentSkipListSet<>());
```

Comment (shabanovd) [inline src/org/exist/storage/XQueryPool.java]: @adamretter how `Thread A` will add object to `Set` if that object was added by `Thread B`? How is that possible?

Comment (shabanovd) [inline src/org/exist/storage/XQueryPool.java]: `size` performance is not very critical because number of items is less 1k

Comment (shabanovd) [inline src/org/exist/storage/XQueryPool.java]: 1) how `compute` will return `CompiledXQuery `?
2) fixing `Iterator` generics...

Comment (shabanovd) [inline src/org/exist/storage/XQueryPool.java]: > you have a race condition between checking validity and removing the query, so you may remove a query which is actually valid.

I do not see race condition here. There is only one compiled script peek up at time and `isValid ` check for it. If one is invalid then other will not pass `isValid` checks too, no?

File: src/org/exist/storage/XQueryPool.java [hunk 5]
```diff
@@ -204,8 +204,8 @@ private void timeoutCheck() {
 
         lastTimeOutCheck.set(currentTime);
 
-        for (final Iterator it = pool.entrySet().iterator(); it.hasNext(); ) {
-            final Source next = (Source) it.next();
+        for (final Iterator<Map.Entry<Source, NavigableSet<CompiledXQuery>>> it = pool.entrySet().iterator(); it.hasNext(); ) {
+            final Source next = it.next().getKey();
             if (currentTime - next.getCacheTimestamp() > timeout) {
```
adamretter: You only need the `keySet`!

src/org/exist/storage/XQueryPool.java
adamretter: @shabanovd I fear there is a race between lines 173 and 176:

1) Thread A determines the query is invalid and is then suspended
2) Thread B causes the query to become valid and is then suspended
3) Thread A continues, and removes the valid compiled queries from the stack,

src/org/exist/storage/XQueryPool.java
shabanovd: sure, but why `entrySet()` is bad?

src/org/exist/storage/XQueryPool.java
shabanovd: I do not see this as problem. `timeoutCheck` removing `valid compiled queries` too.

The worse case for the problem you describe that some valid compiled queries will be GCed and it will not be reused. I do not see it as critical issue, more to that I do not see this as issue at all. That is some trade for better concurrency.

There are always chance that compiled query during evaluation become invalid, or modules changed during compilation.

Anyway, all that cases more develop sites issue, then production one, because on production side scripts do not change too often.

adamretter: @shabanovd I guess the problem is that I don't understand what concurrency model you are trying to achieve and at the moment it seems inconsistent to me. Could you document it in the javadoc of the class so that it is clear what the concurrency access model of the class is?

shabanovd: The technical requirements for XQueryPool as I see it:
 - pool must not return compiled script if resource change;
 - pool may lose compiled script;

adamretter:
@shabanovd The Travis build failed... please can you run the test-suite before pushing!

dizzzz:
question: with this change, no compiled xquery will be cached at all? I see that the 'time of cache' has been removed, so there is no timeout anymore?

shabanovd:
@dizzzz what do you mean?

dizzzz:
is de caching of compiled queries not available anymore? can they still expire?

shabanovd:
@dizzzz yes, compiled queries still can expire. Now pool itself manage that value. Before there was design bug, compiled query hold pool variable and that break SOLID principles.

src/org/exist/storage/XQueryPool.java
joewiz: Instead of needing to rebase, etc., how about if we "squash and merge"? See https://github.com/blog/2141-squash-your-commits. Looks like there are conflicts now though, so once the conflict is cleared up, this could be merged in as a single commit.

shabanovd: @joewiz I agree to anything that allow changes to go in and if it do not require any additional time spending -)

src/org/exist/storage/XQueryPool.java
alebastrov: yes

alebastrov:
Will it be merged finally?

dizzzz:
@adamretter this PR has stalled, could you have a look again? @joewiz one for tonight.... @wolfgangmm  please could you step in?

adamretter:
@shabanovd I think this likely superseded by https://github.com/eXist-db/exist/pull/1590. Is there anything in this PR that you think is still needed?

shabanovd:
@adamretter I have no idea ... but I'm going to close it

shabanovd:
@adamretter did you fix [this](https://github.com/eXist-db/exist/pull/1134/files#diff-78c7f72bc12b13423208dcdddaac4422R520)?

XQuery check only modules and did not check itself.

adamretter:
@shabanovd No currently the validity of an XQuery in the XQueryCache is based on the fact that it is either static, or that it has a last-modified time which can be checked against the DB or filesystem.

I think you are correct in that modules may also need to be considered. It is a shame, ideally we should really be able to consider these things independently. Unfortunately just recursing through the modules is not enough and could lead to cyclic checking. We will need to build some sort of acyclic check, I have some facilities like that with a custom XQuery cache in the RESTXQ Server, I will give this some further thought...

adamretter:
@shabanovd So I spoke too soon.

For checking validity, after checking the last-modified date of the source, we call `CompiledXQuery#isValid()`, which calls `XQueryContext#checkModulesValid()`, which in turn calls `ExternalModule#moduleIsValid()` on each external module used by the XQuery. `moduleIsValid()` checks that the source of the module is still valid in the same manner as we do for the root XQuery (i.e.  either static, or that it has a last-modified time which can be checked against the DB or filesystem.).

So, we are definitely now checking both the XQuery itself and its modules. Does that mean that your patch is no longer needed?

caseydawsonjordan:
@shabanovd are we good with this for now or is there any immediate action we need to take?

adamretter:
@shabanovd @caseydawsonjordan Okay I am going to close this because I believe it is fixed. If you find a problem please open an issue.

PR #1136, opened 2016-09-27
Correct mime-type names for JavaScript & Markdown

joewiz:
For JavaScript, see https://tools.ietf.org/html/rfc4329 (via http://stackoverflow.com/questions/9664282/difference-between-application-x-javascript-and-text-javascript-content-types)
For Markdown, see https://tools.ietf.org/html/rfc7763 (via http://stackoverflow.com/questions/10701983/what-is-the-mime-type-for-markdown)


joewiz:
@adamretter Thanks!

adamretter:
@joewiz All the thanks goes to you!

PR #1139, opened 2016-10-03
Kwic

adamretter:
Previously serializing introduced erroneous whitespace text nodes, i.e.

```xquery
text { ("a", "b") }
```
results in: "`a b`". Whereas we should have had:

```xquery
 text { "a" }, text { "b" }
```
which results in: "`ab`".

I have also introduced:

```xml
<config whitespace-text-nodes="collapse|drop|leave"/>
```

This allows the end-user to have greater control over the processing of spans of whitespace.

### DIFF
--- a/src/org/exist/xquery/lib/kwic.xql
+++ b/src/org/exist/xquery/lib/kwic.xql
@@ -23,7 +23,8 @@
     <config
         width="character width"
         table="yes|no"
-        link="URL to which the match is linked"/>
+        link="URL to which the match is linked"
+        whitespace-text-nodes="collapse|drop|leave"/>
 :)
 xquery version "3.0";
 
@@ -156,6 +157,40 @@ function kwic:string-length($nodes as item()*) as xs:integer {
 		0
 };
 
+declare
+    %private
+function kwic:collapse-whitespace-fn($callback as (function(text(), xs:string) as text()?)?) as (function(text(), xs:string) as text()) {
+    function($text as text(), $mode as xs:string) as text()? {
+        let $text :=
+            if(matches($text, "\s+")) then
+                text { " " }
+            else
+                $text
+        return
+            if(exists($callback))then
+                $callback($text, $mode)
+            else
+                $text
+    }
+};
+
+declare
+    %private
+function kwic:drop-whitespace-fn($callback as (function(text(), xs:string) as text()?)?) as (function(text(), xs:string) as text()?) {
+    function($text as text(), $mode as xs:string) as text()? {
+        let $text :=
+            if(matches($text, "\s+")) then
+                ()
+            else
+                $text
+        return
+            if(exists($text) and exists($callback))then
+                $callback($text, $mode)
+            else
+                $text
+    }
+};
+
 declare function kwic:get-summary($root as node(), $node as element(), 
 	$config as element(config)?) as element() {
 	kwic:get-summary($root, $node, $config, ())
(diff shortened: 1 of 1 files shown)


src/org/exist/xquery/lib/kwic.xql
```diff
@@ -142,16 +172,16 @@ declare function kwic:get-summary($root as node(), $node as element(),
 	@param $callback (optional) reference to a callback function which will be called
 	once for every text node before it is appended to the displayed text. The function
 	should accept 2 parameters: 1) a single text node, 2) a string indicating the
-	current direction in which text is appended, i.e. "before" or "after". The function
-	may return the empty sequence if the current node should be ignore (e.g. if it belongs
-	to a "footnote" which should not be displayed). Otherwise it should return a single
-	string.
+	current direction in which text is appended, i.e. $kwic:MODE_BEFORE or $kwic:MODE_AFTER.
```
joewiz: Are you sure you want to reference the variable name here and not the literal values that users will be entering? I don't think xqdoc would resolve these.

src/org/exist/xquery/lib/kwic.xql
```diff
@@ -17,34 +17,57 @@
...
 
 (:~
 	Like fn:substring, but takes a node argument. If the node is an element,
 	a new element is created with the same node-name as the old one and the
 	shortened text content.
 :)
-declare function kwic:substring($node as item(), $start as xs:int, $count as xs:int) as item()? {
+declare
+    %private
+function kwic:substring($node as item(), $start as xs:integer, $count as xs:integer) as item()? {
```
joewiz: would `node()` be more specific, since we're now returning either an `element()` or `text()`?

src/org/exist/xquery/lib/kwic.xql
```diff
@@ -156,6 +157,40 @@ function kwic:string-length($nodes as item()*) as xs:integer {
...
+    %private
+function kwic:drop-whitespace-fn($callback as (function(text(), xs:string) as text()?)?) as (function(text(), xs:string) as text()?) {
+    function($text as text(), $mode as xs:string) as text()? {
+        let $text :=
+            if(matches($text, "\s+")) then
+                ()
+            else
+                $text
+        return
+            if(exists($text) and exists($callback))then
```
joewiz: Whitespace is preferred around the condition of the `if` clause. Not that this is a bible, but http://xqdoc.org/xquery-style.pdf section 6.1 is good guidance.

src/org/exist/xquery/lib/kwic.xql
adamretter: @joewiz Yes, because the variables are public so the developer can use them from their own modules without caring what the values are.

src/org/exist/xquery/lib/kwic.xql
adamretter: Good catch. I will adjust...

src/org/exist/xquery/lib/kwic.xql
adamretter: As you asked so nicely, I will adjust...

adamretter:
@joewiz I made those changes as requested.

src/org/exist/xquery/lib/kwic.xql
joewiz: @adamretter Ah, that hadn't occurred to me.

joewiz:
Thank you, Adam! Very excited to try this and the other PR from today. The extra whitespace in KWIC results have always flummoxed me.

PR #1143, opened 2016-10-13
Update to the latest version of Lucene 4

adamretter:
As can be seen by the commit history of this PR, I was only able to initially upgrade from Lucene `4.4.0` to `4.6.1`, however...

I spent perhaps two days trying to understand how the Faceting code should work, my conclusion is that it likely never worked correctly, and the experimental nature of the Faceting API in Lucene (which changed frequently) makes it much harder to upgrade Lucene (although I did make some attempts with Faceting, but it didn't work out).

As discussed with @wolfgangmm previously: after I dropped the existing Faceting code (which I think is unused and certainly undocumented), upgrading Lucene was relatively painless. I have managed to upgrade us to the latest 4.x version which is **`4.10.4`** from March 2015.

I am hopeful that we could add Faceting properly in future, once we reach a sufficiently modern version of Lucene.

Also, I have almost completed the work to move us to Lucene 5, https://github.com/adamretter/exist/commit/726a1badfeda6cfd978a8f15fa1f091e38f5dfa5
However Lucene 5+ does not allow the per document `Analyzers` which we currently use with Lucene 4; See https://issues.apache.org/jira/browse/LUCENE-6212 for the reasons for the change in Lucene. So we need to rethink how we do that kind of thing, and then we can take Lucene 5. My goal to keep updating us until we get to to the latest version 6 of Lucene.


shabanovd:
@adamretter or you can as for a help to update facets code instead of throw out it -(

adamretter:
@shabanovd Sorry, I for some reason thought you were aware of the difficulties.

The main reason I threw it out, was that I couldn't seem to get it to work in  a sensible way, for example the existing eXist code required that each document only have a single facet value per-facet key.

So, would you like to update the Facet code? If so I see two options:

1) If you could do so this in the next 1-2 days: then create your own branch of this branch, and rebase it to remove commit 682f6e5, you then need to update each subsequent commit so that the faceting works with each new version of Lucene.

2) If this is either something that you are not interested in, or not something that you can do in the near future: then I would suggest you/we wait until we hit a recent version of Lucene (i.e. Lucene 6), and then do a complete implementation of Faceting (including XQuery functions, tests and docs etc) so that it works correctly.

Personally I think (2) will be the easier option here.

Just let me know @shabanovd which you would like to do...

adamretter:
@dizzzz Thanks, I think it was a feature that was never fully developed though, as there was no facility for a user to directly configure facet indexing or create searches from XQuery (?e.g. `ft:facet()`) based on Facets. I may be mistaken but I also think the indexing of the Facets that was there has some problems.

joewiz:
For what it's worth, I'm on record as supporting efforts (1) full text and (2) faceted search abilities in eXist - in that order of priority. As long as eXist relies on Lucene for one or both of these abilities, I think it's wise to make whatever adjustments are necessary to keep up with Lucene's development. Looking at their release history, it appears Lucene 4 is not actively maintained, and 5 and 6 are where the main lines of development are happening, so preparing to make the leap to 5 and/or 6 seems wise. Thanks @adamretter @shabanovd @dizzzz for your efforts in this area!

joewiz:
Also, on the topic of facets, @wsalesky has created a partial, XQuery-based implementation of the EXPath facet spec: https://github.com/srophe/srophe-eXist-app/blob/ec349742ed49545e82c4040d095a045ab4862d6e/srophe-app/modules/lib/facet.xqm.

shabanovd:
@adamretter I will not have time next 2 days, so go with option 2.

Information for you, facets code was not updated because you reject metadata code changes that required for facets. And I have facets that working with all 4.x and 5.x versions of lucene, but as I see you accept only your own code, that's why you are welcome to code everything yourself -(

@dizzzz Facets in production for long time, but that code can't get to develop branch because of workflow and discussion process in used right now

adamretter:
@shabanovd I have spent a great deal of time and energy reviewing your code. As you well know I just recently merged some of your PRs and reviewed some more in detail.

I would have accepted your code for the metadata stuff, had it been commented, included test cases, did not mix concerns and you made the changes that were requested in the PR from the team. I was not the only person that had concerns, although I may have been the most vocal because I try and review so many of the PRs.

If you check the history you will see that I have accepted many PRs from many authors, but never my own, that is prohibited; No one on this project is allowed to accept their own PRs.

I feel like I often review your code and request minor changes, but your response is often that you won't make the requested changes, which leaves things in limbo. Ideally someone else would arbitrate, but that doesn't seem to happen. The point of PRs and code review is you have to follow the process; If you abandon it then everyone looses.

shabanovd:
@adamretter I didn't mean you personally, but project in general. Concerns that do not say aloud can't be discussed!

Same time your comment "I for some reason thought you were aware of the difficulties." I see as communication issue and I have no idea how to fix it -(

PR #1144, opened 2016-10-15
Remove last traces of java.io.file in favour of NIO.2

adamretter:
During the audit I found several issues which were also repaired as a result of these changes:

1. Input and Output streams not being closed
2. File handles not being closed (some related to above)
3. Inefficient copying to/from files using a loop and a fixed sized buffer, now using `Files#copy`
4. Inefficient reading/writing of a `byte[]` from/to a file using a loop and fixes sized buffer, now using `Files#readAll..` etc.


adamretter:
Note that only the final commit `[c0b66ce](https://github.com/eXist-db/exist/pull/1144/commits/c0b66cecad66cd6f00c91d12301914f65e42cf39)` needs to be reviewed (if at all).

dizzzz:
Looks good :)

PR #1153, opened 2016-10-24
Fix a small issue with EXPath PKG deployment at startup

adamretter:


adamretter:
Only the first of the two commits really needs to be reviewed.

dizzzz:
great, this fixes the issue. thnx!

PR #1157, opened 2016-10-31
Make build fail when JAVA_HOME and EXIST_HOME are not set

ljo:
Failing the build if environment variables JAVA_HOME and EXIST_HOME are not set for the yajsw wrapper config preparation. Addresses #1147.

PR #1159, opened 2016-10-31
Reworked LJO's PR #1157 

dizzzz:
let's check if travis finds it ok. rework of https://github.com/eXist-db/exist/pull/1157

### DIFF
--- a/tools/yajsw/build.xml
+++ b/tools/yajsw/build.xml
@@ -29,21 +29,16 @@
     	<pathelement path="${server.dir}/start.jar"/>
     	<pathelement path="${server.dir}/exist.jar"/>
     </path>
-    
+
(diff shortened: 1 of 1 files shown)


dizzzz:
@ljo @wolfgangmm  @shabanovd ? it is a prerequisite for my other PR

ljo:
I would prefer an absolute path since yajsw will to my findings act strangly otherwise.

dizzzz:
@ljo please can you be more specific ? Inthin if you set EXIST_HOME explicitly for the build,  this value will be used,

dizzzz:
@ljo that will only be possible if we set a variable in a higher ant build file. The challenge then is.... how to be sure this value is passed?

dizzzz:
@ljo I have found a solution....

Issue #1163, opened 2016-11-03
Set and check EXIST_HOME in JAJSW scripts


dizzzz:
for example, in the script `./tools/yajsw/bin/installDaemon.sh` the variable EXIST_HOME is used, but it nowhere set (neither in `setenv.sh` which is sourced in), nor the install script checks for a value.

As a result the `installDaemon.sh` does not work correctly for macOS, probably this is the reason that things do not work for systemd as reported by @tasmo

so please make sure your dev shell does not have 'state' .... the EXIST_HOME variable

dizzzz:
@ljo ?

duncdrum:
@ljo ?

duncdrum:
YAJSW is no more, exist 3 is ancient history

Issue #1154, opened 2016-10-26
yajsw wrapper does not work on Ubuntu


tasmo:
When I tried to start exist with yajsw I got a bunch of errors and exist does not run although the wrapper script seems to start.

The directory `/tmp/vfs_cache` is needed by the wrapper. So I solved the Java error by creating it.

The errors caused by the missing libs in `tools/yajsw/lib/extended/` are not important but I installed the missing libs from the yajsw sources.

The next problem is (and still is) the `${env.EXIST_HOME}` environment variable. The system itself echoes the right path with `echo $EXIST_HOME` in any shell. Since I changed that with success the wrapper now does not try to make the folder with the name `${env.EXIST_HOME}` anymore but one with the name `?unresolved?` in the root of `$EXIST_HOME`.

exist is still not running after a successful installation with `./tools/yajsw/bin/installDaemon.sh`; `sudo systemctl daemon-reload` and `sudo systemctl start eXist-db`. systemctl starts the script and echoes success. Nor a direct call with `/etc/init.d/eXist-db start` leaves a running exist.

Starting `bin/startup.sh` in a screen session runs exist without any errors.

dizzzz:
duplicate of https://github.com/eXist-db/exist/issues/1147

dizzzz:
thnx for your useful details....

tasmo:
Thought because the issue #1147 is about macOS. :)

dizzzz:
Right you are, But it is the same issue :)

dizzzz:
issue is still valid, and different that macOS

joewiz:
Is this issue still present now that #1165 has been merged?

adamretter:
@tasmo Is this now fixed for you with HEAD?

duncdrum:
YAJSW is gone, no response since 2016. Closing

Issue #1147, opened 2016-10-17
Wrapper config can be generated with unset variables eg replaced with property name on a direct repo build


dizzzz:
The work from @ljo 's PR https://github.com/eXist-db/exist/pull/1141 does not yet for macOS:
(added: same for ubuntu)

During startup the directory `tools/yajsw/${env.EXIST_HOME}/` ; the variable does not seem to be substituted?  

start with

``` bash
./tools/yajsw/bin/runConsole.sh
```

In the attached logfile the following line is relevant:

```
java.io.FileNotFoundException: ${env.EXIST_HOME}/tools/yajsw/tmp/in_-5209836399611897029$1476732036731 (Permission denied)
    at java.io.RandomAccessFile.open0(Native Method)
```

[error.txt](https://github.com/eXist-db/exist/files/534495/error.txt)

setting EXIST_HOME manually does not fix the issue.

ljo:
Thanks, I will check how this can be the case.

dizzzz:
Just before executing the command mentioned above

dizzzz:
any idea yet about this issue @ljo ? Or maybe a work around?

dizzzz:
@ljo https://github.com/eXist-db/exist/issues/1154 contains valuable details....

tasmo:
I tried to install the daemon at my local Arch Linux running systemd. I set up successful `$JAVA_HOME` to `/usr/lib/jvm/default` which seems to be right and `$EXIST_HOME` to a symlinked `/opt/exist`.

When I try to start `systemctl --user start eXist-db` I get an error that "the control process exited with error code." Starting the wrapper manually with `/usr/lib/jvm/default/bin/java -Dwrapper.pidfile=/opt/exist/tools/yajsw/work/wrapper.eXist-db.pid -Dwrapper.service=true -Dwrapper.visible=false -Djna_tmpdir=/opt/exist/tools/yajsw/tmp -jar /opt/exist/tools/yajsw/wrapper.jar -tx /opt/exist/tools/yajsw/conf/wrapper.conf` gives me the following error:

```shell
INFO: lib not found: /home/me/exist/tools/yajsw/lib/extended/groovy/groovy-patch.jar
INFO: lib not found: /home/me/exist/tools/yajsw/lib/extended/groovy/groovy-all-2.3.11.jar
INFO: lib not found: /home/me/exist/tools/yajsw/lib/extended/commons/commons-httpclient-3.0.1.jar
[...]

PR #1165, opened 2016-11-04
Yajsw systemd fixes

adamretter:
This is the first round. It solves the issue with PID, but there is a bigger problem at play which I need to discuss with @ljo.


dizzzz:
related to #1154 #1163 #1147

PR #1170, opened 2016-11-05
Improve log function

### LINKED ISSUE #1169: log-system-out writes "raw" query to std out too
when `util:log-system-out("hahaha")`  is called from a not-stored-query, also the query itself is printed to standard out, giving a lot of noise.

e.g. the above yields into 

```
(Line: 1 util:log-system-out("hahaha")) hahaha
```

whilst I'd expect

```
(Line: 1 String) hahaha
```

the cause is in org.exist.xquery.AbstractExpression#getSource line 121:

```java
if(getSource() != null && getSource().getKey() != null) {
  buf.append(" ");
  buf.append(getSource().getKey());
}
```

as far as I see `getSource().getKey()` returns the whole query string, which does not make sense to me?

dizzzz:
Fixes #1169

### DIFF
--- a/src/org/exist/xquery/functions/util/LogFunction.java
+++ b/src/org/exist/xquery/functions/util/LogFunction.java
@@ -3,43 +3,32 @@
  *  Copyright (C) 2001-2009 Wolfgang M. Meier
  *  wolfgang@exist-db.org
  *  http://exist.sourceforge.net
- *  
+ *
  *  This program is free software; you can redistribute it and/or
  *  modify it under the terms of the GNU Lesser General Public License
  *  as published by the Free Software Foundation; either version 2
  *  of the License, or (at your option) any later version.
- *  
+ *
  *  This program is distributed in the hope that it will be useful,
  *  but WITHOUT ANY WARRANTY; without even the implied warranty of
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  *  GNU Lesser General Public License for more details.
- *  
+ *
  *  You should have received a copy of the GNU Lesser General Public License
  *  along with this program; if not, write to the Free Software
  *  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
- *  
+ *
  *  $Id$
  */
 package org.exist.xquery.functions.util;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-
 import org.exist.dom.QName;
 import org.exist.source.StringSource;
 import org.exist.storage.serializers.Serializer;
-import org.exist.xquery.BasicFunction;
-import org.exist.xquery.Cardinality;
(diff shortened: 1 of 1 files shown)


src/org/exist/xquery/functions/util/LogFunction.java
```diff
@@ -173,10 +174,7 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
                 final String logname = args[1].getStringValue();
 
                 // Use specific logger when provided
-                Logger logger = LOG;
-                if (logname != null && logname.length() > 0) {
-                    logger = LogManager.getLogger(logname);
-                }
+                Logger logger = StringUtils.isBlank(logname) ? LOG : LogManager.getLogger(logname);
```
adamretter: Do we really need `StringUtils.isBlank(logname)`?  I would rather `logname == null || logname.isEmpty()`

src/org/exist/xquery/functions/util/LogFunction.java
```diff
@@ -3,191 +3,213 @@
...
+    };
+
+    public LogFunction(XQueryContext context, FunctionSignature signature) {
+        super(context, signature);
+    }
+
+    /* (non-Javadoc)
+     * @see org.exist.xquery.BasicFunction#eval(org.exist.xquery.value.Sequence[], org.exist.xquery.value.Sequence)
+     */
+    public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathException {
```
adamretter: The junk Eclipse javadoc comment should be replaced with `@Override` whilst you are cleaning up.

src/org/exist/xquery/functions/util/LogFunction.java
```diff
@@ -3,191 +3,213 @@
...
-}
+    /**
+     *  Write log message to logger with a priority.
+     *
+     * @param buffer    The log text
+     * @param loglevel  The priority of the log message
+     * @param logger    The actual logger
+     */
+    private void writeLog(StringBuilder buffer, String loglevel, Logger logger) {
+        switch (loglevel) {
```
adamretter: These variables could be `final`

src/org/exist/xquery/functions/util/LogFunction.java
dizzzz: it is essentially the same right? I use this class basically always..... for readability.
The isBlank() also checks for whitespace characters....

dizzzz: ok, done

src/org/exist/xquery/functions/util/LogFunction.java
dizzzz: ok, done

src/org/exist/xquery/functions/util/LogFunction.java
dizzzz: done

PR #1172, opened 2016-11-06
Range index bug fixes and enhancements

### LINKED ISSUE #1055: Bug in Lucene full text query with diacritics="no" filter
We are using the diacritics="no" filter on our Lucene index.  When searching for a name with diacritics and a wildcard I get no results, while searching for the same name without diacritics and a wildcard will return the expected results. 

Here are a few links demonstrating the various results:

Full name without diacritics and a wildcard: 
http://syriaca.org/geo/search.html?q=abajal* 

Diacritics and wildcard:
http://syriaca.org/geo/search.html?q=%C4%80b%C4%81j%C4%81l*

Here is our index configuration file:
https://github.com/srophe/srophe-app-data/blob/master/collection.xconf

Here is a sample record: 
https://github.com/srophe/srophe-app-data/blob/master/data/places/tei/1949.xml

And a simplified version of the query:

```
xquery version "3.0";
declare namespace tei = "http://www.tei-c.org/ns/1.0";

for $r in collection('/db/apps/srophe-data/data/places/tei')//tei:placeName[ft:query(.,"Ābājāl*")]
return $r
```

wolfgangmm:
1. fix issue with byte reference becoming invalid; 
2. throw error when using fn:contains, fn:starts-with or fn:ends-with on a collated index. A collated index is not suitable for substring matching.
3. generalize analyzer configuration so arbitrary lucene token filters can be plugged into it via collection.xconf (see test cases). It would also be possible to extend this to pass in optional parameters, but looking at the token filters which might apply to a range index, there are hardly any useful ones for the time being, so I skipped this
4. fix wrong gt, ge, lt, le comparison for analyzed strings
5. extended test cases

Closes #1055, #1098.

### DIFF
--- a/extensions/indexes/range/src/org/exist/indexing/range/ComplexRangeIndexConfigElement.java
+++ b/extensions/indexes/range/src/org/exist/indexing/range/ComplexRangeIndexConfigElement.java
@@ -50,8 +50,10 @@ public class ComplexRangeIndexConfigElement extends RangeIndexConfigElement {
             Node child = children.item(i);
             if (child.getNodeType() == Node.ELEMENT_NODE) {
                 if (FIELD_ELEMENT.equals(child.getLocalName())) {
-                    RangeIndexConfigField field = new RangeIndexConfigField(path, (Element)child, namespaces);
+                    RangeIndexConfigField field = new RangeIndexConfigField(path, (Element) child, namespaces);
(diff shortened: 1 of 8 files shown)


joewiz:
The ability to plug in arbitrary lucene token filters will be really helpful! Thanks @wolfgangmm!

PR #1180, opened 2016-11-14
Do not remove the .DO_NOT_DELETE file from jetty/tmp during cleanup.

dizzzz:
@wolfgangmm  this might be the fix of your issue


joewiz:
Thank you @dizzzz! This one was getting me too.

PR #1188, opened 2016-11-28
fix 'x' flag parsing

telic:
as seen in FunMatches.java#L528, an 'x' flag in regular expressions should be interpreted as Pattern.COMMENTS

### DIFF
--- a/src/org/exist/xquery/functions/fn/FunAnalyzeString.java
+++ b/src/org/exist/xquery/functions/fn/FunAnalyzeString.java
@@ -195,7 +195,7 @@ public class FunAnalyzeString extends BasicFunction {
                 break;
                 
             case 'x' :
-                iFlags |= Pattern.CANON_EQ;
+                iFlags |= Pattern.COMMENT;
                 break;
                 
             case 'q' :
--- a/src/org/exist/xquery/functions/fn/FunAnalyzeString.java
+++ b/src/org/exist/xquery/functions/fn/FunAnalyzeString.java
@@ -195,7 +195,7 @@ public class FunAnalyzeString extends BasicFunction {
                 break;
                 
             case 'x' :
-                iFlags |= Pattern.COMMENT;
+                iFlags |= Pattern.COMMENTS;
                 break;
                 
             case 'q' :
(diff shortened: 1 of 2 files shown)


dizzzz:
could you provide a small test case?

dizzzz:
The change does not build.....

```
compile:
     [echo] Compiling with Java 1.8 from 1.8 source to 1.8 target, debug on, optimize on, deprecation off
    [javac] Compiling 1486 source files to /home/travis/build/eXist-db/exist/build/classes
    [javac] /home/travis/build/eXist-db/exist/src/org/exist/xquery/functions/fn/FunAnalyzeString.java:198: error: cannot find symbol
    [javac]                 iFlags |= Pattern.COMMENT;
    [javac]                                  ^
    [javac]   symbol:   variable COMMENT
    [javac]   location: class Pattern
    [javac] Note: Some input files use or override a deprecated API.
    [javac] Note: Recompile with -Xlint:deprecation for details.
    [javac] Note: Some input files use unchecked or unsafe operations.
    [javac] Note: Recompile with -Xlint:unchecked for details.
    [javac] 1 error
```

telic:
Sorry, that's what I get for trying to do a quick fix directly in Github :)

Did you want a Java test case, or just an example of current failure and expected behaviour in XQuery?

dizzzz:
Xquery is good enough; we have a xqtest suite for this :)

telic:
Example failure: `fn:analyze-string('test', 't e s t', 'x')` returns
```
<fn:analyze-string-result xmlns:fn="http://www.w3.org/2005/xpath-functions">
  <fn:non-match>test</fn:non-match>
</fn:analyze-string-result>
```

The expected result is
```
<fn:analyze-string-result xmlns:fn="http://www.w3.org/2005/xpath-functions">
  <fn:match>test</fn:match>
</fn:analyze-string-result>
```
since `fn:matches('test', 't e s t', 'x')` returns `true()`.

telic:
I'm not clear how/where I would add a test for this in the code... Should I just add an XQuery with test annotations in test/src/xquery/?

adamretter:
@telic It would be great if you could add some tests to https://github.com/eXist-db/exist/blob/develop/test/src/xquery/xquery3/fn.xql for `fn:analyze-string`

adamretter:
@telic Thank you for your contribution :-)

PR #1200, opened 2016-12-14
Fix BC dates in the range index DateConverter

olvidalo:
In our project we have to deal with dates BC, like "-300" or "-0004-03-29" that, when indexed, have to be normalized using the DateConverter (which should definitely be documented somewhere... utter happiness ensued when I stumbled upon this while digging through the sources 😄 )

These changes make the converter work correctly with negative (BC) dates. Would be excellent if this could go into 3.0.

### DIFF
--- a/extensions/indexes/range/src/org/exist/indexing/range/conversion/DateConverter.java
+++ b/extensions/indexes/range/src/org/exist/indexing/range/conversion/DateConverter.java
@@ -42,13 +42,13 @@ public class DateConverter implements TypeConverter {
 
     protected static final Logger LOG = LogManager.getLogger(DateConverter.class);
 
-    private final static Pattern DATE_REGEX = Pattern.compile("(\\d+)-(\\d+)-(\\d+)");
+    private final static Pattern DATE_REGEX = Pattern.compile("(-?\\d+)-(\\d+)-(\\d+)");
 
     @Override
     public Field toField(String fieldName, String content) {
         try {
             DateValue dv;
-            if (content.indexOf('-') < 0) {
+            if (content.indexOf('-') < 1) {
                 // just year
                 int year = Integer.parseInt(content);
                 XMLGregorianCalendar calendar = TimeUtils.getInstance().newXMLGregorianCalendar();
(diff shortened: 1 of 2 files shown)


extensions/indexes/range/src/org/exist/indexing/range/conversion/DateConverter.java
```diff
@@ -42,13 +42,13 @@
...
 
-    private final static Pattern DATE_REGEX = Pattern.compile("(\\d+)-(\\d+)-(\\d+)");
+    private final static Pattern DATE_REGEX = Pattern.compile("(-?\\d+)-(\\d+)-(\\d+)");
 
     @Override
     public Field toField(String fieldName, String content) {
         try {
             DateValue dv;
-            if (content.indexOf('-') < 0) {
+            if (content.indexOf('-') < 1) {
```
dizzzz: Please use the String.contains()

dizzzz:
Nice :) any idea why Travis fails?

extensions/indexes/range/src/org/exist/indexing/range/conversion/DateConverter.java
olvidalo: You mean like `!(content.substring(1).contains('-'))`? As this should only evaluate false if the string contains '-' in any other position than the first position, accounting for negative "only-year" values.

Seems more complicated to me, but if you insist... 😉

olvidalo:
Not sure! It seems to fail due to dependency issues regarding betterform, it cannot find jgroup.

Maybe just a glitch? Can't imagine that this is caused by these changes somehow...

joewiz:
Nice addition! Thanks!

extensions/indexes/range/src/org/exist/indexing/range/conversion/DateConverter.java
adamretter: @olvidalo I think it is fine as it is

adamretter:
@olvidalo Thanks for your contribution :-)

olvidalo:
My pleasure. Thanks for merging 🎉

duncdrum:
just out of curiosity @olvidalo will this normalise '0' to '-0001'?

olvidalo:
No, it would "normalise" it to "0000-00-00" which would then throw an error somewhere down the line because this is an invalid value for xs:date.

dizzzz:
Thnx!

dizzzz:
Thnx!

PR #1203, opened 2016-12-19
Refactor Collection and Lock

adamretter:
This PR extracts an interface for Collection and adds all the missing javadoc for its public methods.

It also changes the Locking Mode to be an enumeration from an int, which revealed at least one bug from a mistaken use of the wrong int in `org.exist.util.io.Resource`.

This is laying the ground work for some serious improvements to the Collection Cache and its synchronization with Collections, which will follow in a separate PR shortly.


adamretter:
@wolfgangmm thanks :-)

PR #1205, opened 2016-12-23
Small bugfix to Locks

adamretter:
One commit fixes an issue that was recently introduced as a regression. The other is a guard against possible misuse.


dizzzz:
Looks good ...... I think

ljo:
Yes.

PR #1206, opened 2016-12-30
Further junit rules

adamretter:
Further cleanup of the JUnit tests which directly use the Internal API of eXist. Should result in less code and some tests now only start the database once rather than multiple times which should speed up the test suite.


dizzzz:
Looks good to me.... (from small device)

PR #1214, opened 2017-01-09
Increase max memory default and cache size, small fixes

wolfgangmm:
Increase the default used by the installer for Java max memory from 1024m to 2048m. Most computers today should have >4gb main memory. Giving just 1gb to eXist easily results in out of memory issues and crashes.

Also fix a ClassCastException popping up during startup/shutdown.

### DIFF
--- a/bin/functions.d/eXist-settings.sh
+++ b/bin/functions.d/eXist-settings.sh
@@ -76,7 +76,7 @@ restore_library_path() {
 
 set_client_java_options() {
     if [ -z "${CLIENT_JAVA_OPTIONS}" ]; then
-	CLIENT_JAVA_OPTIONS="-Xms128m -Xmx1024m -Dfile.encoding=UTF-8";
+	CLIENT_JAVA_OPTIONS="-Xms128m -Xmx2048m -Dfile.encoding=UTF-8";
     fi
 
     OS=`uname`
@@ -90,7 +90,7 @@ set_client_java_options() {
 
 set_java_options() {
     if [ -z "${JAVA_OPTIONS}" ]; then
-	JAVA_OPTIONS="-Xms128m -Xmx1024m -Dfile.encoding=UTF-8";
+	JAVA_OPTIONS="-Xms128m -Xmx2048m -Dfile.encoding=UTF-8";
     fi
     JAVA_OPTIONS="${JAVA_OPTIONS} -Djava.endorsed.dirs=${JAVA_ENDORSED_DIRS}";
 }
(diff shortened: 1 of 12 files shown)


tools/yajsw/src/org/exist/yajsw/Main.java
```diff
@@ -11,7 +11,7 @@
 /**
  *
  */
-public class Main implements Observer {
+public class Main implements Observer, Comparable {
```
dizzzz: I have seen this issue once, or twice, not reproducible....

bin/functions.d/eXist-settings.sh
```diff
@@ -76,7 +76,7 @@ restore_library_path() {
 
 set_client_java_options() {
     if [ -z "${CLIENT_JAVA_OPTIONS}" ]; then
-	CLIENT_JAVA_OPTIONS="-Xms128m -Xmx1024m -Dfile.encoding=UTF-8";
+	CLIENT_JAVA_OPTIONS="-Xms128m -Xmx2048m -Dfile.encoding=UTF-8";
```
adamretter: I don't think the Java Client needs very much at all (unless running in embedded mode), maybe just change the server setting?

installer/userInput.xml
```diff
@@ -28,11 +28,11 @@
         <field type="space"/>
         <field type="text" variable="MAX_MEMORY">
           <description align="left" txt="Configure the maximum amount of memory to be used by eXist-db. The recommended minimum memory setting is 1024m. If your machine has enough memory, using 2048m should be sufficient for most data sets. Note: 1200m is usually the limit on 32-bit systems."/>
-          <spec txt="Maximum memory in mb:" size="5" set="1024"/>
+          <spec txt="Maximum memory in mb:" size="5" set="2048"/>
```
adamretter: "mb" should be "MB"

installer/userInput.xml
```diff
@@ -28,11 +28,11 @@
...
         </field>
         <field type="text" variable="cacheSize">
             <description align="left" txt="Configure the amount of memory to be reserved for internal caches. If the maximum memory is set above to 1024m, 128m is a good value; for 2048m, use 256m; for 512m, not more than 64mb. Too small settings may lead to bad upload/indexing performance."/>
-          <spec txt="Cache memory in mb:" size="5" set="128"/>
+          <spec txt="Cache memory in mb:" size="5" set="256"/>
```
adamretter: "mb" should be "MB"

bin/functions.d/eXist-settings.sh
wolfgangmm: If you run backup/restore via the client having a larger max memory is sometimes required. For normal operations, the memory won't be used, so setting it a bit higher won't do harm I think?

Issue #1215, opened 2017-01-10
required jars for contentextraction module are missing in WAR build


rvdb:
When building a WAR file via the dist-war build task, the Tika-related libs needed for the contentextraction module aren't included, resulting in error messages such as 

    java.lang.NoClassDefFoundError: org/apache/tika/parser/Parser 
    
Manually copying all jars from the `${exist.home}/extensions/contentextraction/lib` folder to the resulting `WEB-INF/lib` folder in the extracted WAR tree solves the issue. This had been reported on the mailing list earlier: http://markmail.org/message/nlyeflol4cy4pvey.

I'll try and provide a fix for the build script in a PR.

dizzzz:
which exist-db version?

rvdb:
Sorry, from eXist-3.0RC1 onwards.

PR #1216, opened 2017-01-10
include required jars for contentextraction module in WAR build

rvdb:
This should fix https://github.com/eXist-db/exist/issues/1215

### DIFF
--- a/build/scripts/dist.xml
+++ b/build/scripts/dist.xml
@@ -127,6 +127,11 @@
               <include name="*.jar"/>
             </lib>
 
+            <!-- Include content extraction jars -->
+            <lib dir="extensions/contentextraction/lib">
+              <include name="*.jar"/>
+            </lib>
+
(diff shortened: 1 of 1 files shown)


rvdb:
@dizzzz Thanks. Sorry if I'm dense, but I'm not sure if I'm fully understanding your XAR remark. Since the Tika-related jars required for the contentextraction module were missing in the generated WAR file, I've just added an ANT instruction in my patch to include them when building the WAR. I've based this on the other &lt;lib> instructions in the build file that select jars for other extension modules to be included in the WAR file that's being built. 

Or do you mean the contentextraction module should be refactored to an independent XAR module that can be imported into a running eXist instance?

dizzzz:
Sorry I was not so clear :-) I have been thinking of moving the content-extractor into a separate XAR file, for easy deployment. But it is not a valid remark in the context of the WAR fix, sorry.

PR #1219, opened 2017-01-11
[feature] A facility for optimising indexes which support differential updates when replacing a document

adamretter:
Such a facility is used by the Algolia Index plugin - https://gitlab.com/adamretter/exist-algolia-index/

### DIFF
--- a/src/org/exist/collections/MutableCollection.java
+++ b/src/org/exist/collections/MutableCollection.java
@@ -1258,7 +1258,7 @@ public class MutableCollection implements Collection {
         if(info.isCreating()) {
             info.getTriggers().afterCreateDocument(broker, transaction, document);
         } else {
-            final StreamListener listener = broker.getIndexController().getStreamListener(document, StreamListener.ReindexMode.REPLACE_DOCUMENT);
+            final StreamListener listener = broker.getIndexController().getStreamListener();
             listener.endReplaceDocument(transaction);
 
             info.getTriggers().afterUpdateDocument(broker, transaction, document);
--- a/src/org/exist/collections/MutableCollection.java
+++ b/src/org/exist/collections/MutableCollection.java
@@ -1258,6 +1258,9 @@ public class MutableCollection implements Collection {
         if(info.isCreating()) {
             info.getTriggers().afterCreateDocument(broker, transaction, document);
         } else {
+            final StreamListener listener = broker.getIndexController().getStreamListener(document, StreamListener.ReindexMode.REPLACE_DOCUMENT);
+            listener.endReplaceDocument(transaction);
+
             info.getTriggers().afterUpdateDocument(broker, transaction, document);
         }
         
@@ -1441,6 +1444,14 @@ public class MutableCollection implements Collection {
                 updateModificationTime(document);
                 oldDoc.getUpdateLock().acquire(LockMode.WRITE_LOCK);
                 oldDocLocked = true;
+
+                /**
+                 * Matching {@link StreamListener#endReplaceDocument(Txn)} call is in
+                 * {@link #storeXMLInternal(Txn, DBBroker, IndexInfo, Consumer2E)}
+                 */
+                final StreamListener listener = broker.getIndexController().getStreamListener(document, StreamListener.ReindexMode.REPLACE_DOCUMENT);
+                listener.startReplaceDocument(transaction);
+
                 if (oldDoc.getResourceType() == DocumentImpl.BINARY_FILE) {
                     //TODO : use a more elaborated method ? No triggers...
                     broker.removeBinaryResource(transaction, (BinaryDocument) oldDoc);
--- a/src/org/exist/indexing/StreamListener.java
+++ b/src/org/exist/indexing/StreamListener.java
@@ -43,6 +43,13 @@ public interface StreamListener {
         STORE,
 
         /**
+         * Mode for replacing the nodes of a document with another document
+         * this is really a group mode, which is later followed by {@link #REMOVE_ALL_NODES}
+         * and then {@link #STORE}
+         */
+        REPLACE_DOCUMENT,
+
+        /**
          * Mode for removing all the nodes of a document
          */
         REMOVE_ALL_NODES,
@@ -80,6 +87,30 @@ public interface StreamListener {
      * @return the next listener in the chain.
      */
     StreamListener getNextInChain();
+
+    /**
+     * Starting to replace a document
+     *
+     * After which the sequence of {@link #startIndexDocument(Txn)} / events / {@link #endIndexDocument(Txn)}
+     * will be called twice, first where the index mode will be {@link ReindexMode#REMOVE_ALL_NODES}
+     * and second where the index mode will be {@link ReindexMode#STORE}
+     * this is then finished by {@link #endReplaceDocument(Txn)}
+     *
+     * This can be used in conjunction with {@link #endReplaceDocument(Txn)} in indexes
+     * which support differential updates
+     *
+     * @param transaction The current executing transaction
+     */
+    void startReplaceDocument(Txn transaction);
+
+    /**
+     * Finished replacing a document
+     *
+     * See {@link #startReplaceDocument(Txn)} for details
+     *
+     * @param transaction The current executing transaction
+     */
+    void endReplaceDocument(Txn transaction);
 
     /**
      * Starting to index a document
--- a/src/org/exist/indexing/AbstractStreamListener.java
+++ b/src/org/exist/indexing/AbstractStreamListener.java
@@ -49,6 +49,20 @@ public abstract class AbstractStreamListener implements StreamListener {
     }
 
     @Override
+    public void startReplaceDocument(final Txn transaction) {
+        if (next != null) {
+            next.startReplaceDocument(transaction);
+        }
+    }
+
+    @Override
+    public void endReplaceDocument(Txn transaction) {
+        if (next != null) {
+            next.endReplaceDocument(transaction);
+        }
+    }
+
+    @Override
     public void startIndexDocument(final Txn transaction) {
         if (next != null) {
             next.startIndexDocument(transaction);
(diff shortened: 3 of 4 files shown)


joewiz:
@adamretter Sounds interesting! The gitlab link you included in the description isn't public. I assume this is intentional but wanted to mention it just in case.

adamretter:
@joewiz Ah okay, well that is probably for the best for now. I need @ttasovac to sign it off as Open Source and then we can move it to a more appropriate repo. I guess I just wanted to covey that this feature is being used in real life!

src/org/exist/collections/MutableCollection.java
```diff
@@ -1258,6 +1258,9 @@ private void storeXMLInternal(final Txn transaction, final DBBroker broker, fina
         if(info.isCreating()) {
             info.getTriggers().afterCreateDocument(broker, transaction, document);
         } else {
+            final StreamListener listener = broker.getIndexController().getStreamListener(document, StreamListener.ReindexMode.REPLACE_DOCUMENT);
```
shabanovd: it going to be new `StreamListener`, not same as at `startReplaceDocument` ... lock wrong to me

src/org/exist/indexing/StreamListener.java
```diff
@@ -43,6 +43,13 @@
         STORE,
 
         /**
+         * Mode for replacing the nodes of a document with another document
+         * this is really a group mode, which is later followed by {@link #REMOVE_ALL_NODES}
+         * and then {@link #STORE}
+         */
+        REPLACE_DOCUMENT,
```
shabanovd: it's very confusing, there should be `REPLACE_DOCUMENT` mode and `startIndexDocument` + `endIndexDocument` or just methods `startReplaceDocument` + `endReplaceDocument`. But I can't understand why you adding anything of this if there are `REMOVE_ALL_NODES` or `REMOVE_SOME_NODES` + `STORE` and methods `startIndexDocument` + `endIndexDocument`.

src/org/exist/collections/MutableCollection.java
adamretter: @shabanovd You are mistaken.
Each `DBBroker` has an `IndexController`. Each `IndexController` has one `IndexWorker` per-index; therefore each DBBroker has a single `IndexWorker` per Index.

So far in all the index implementations, my own included, the `IndexWorker` has a single `StreamListener`. So `startReplaceDocument` will be called on the same `StreamListener` as the other events for the same document being processed by the same broker.

src/org/exist/indexing/StreamListener.java
adamretter: Previously when replacing a document the index calls would be:

```
setMode(REMOVE_ALL_NODES)
startIndexDocument()
...
endIndexDocument()

setMode(STORE)
startIndexDocument()
...
endIndexDocument()
```

The index had no way of knowing if the above sequence of events was the action of a document being replaced, or if it was two separate actions i.e. removing a document and then later storing a new document.

I needed to be able to differentiate between the two to perform the most efficient differential updates to my index.

With this PR, the the index calls for replacing a document are now:

```
setMode(REPLACE_DOCUMENT)
startReplaceDocument()

setMode(REMOVE_ALL_NODES)
startIndexDocument()
...
endIndexDocument()

setMode(STORE)
startIndexDocument()
...
endIndexDocument()

endReplaceDocument()
```

joewiz:
Would this new method open up new ways for update operations on indexed data to be made more efficient? I'm thinking of cases where we use XQuery Update to replace or update the value of a node in a large document. I forget if the whole document needs to be reindexed now, or if it already is capable of only reindexing the updated portion of the document?

wolfgangmm:
@joewiz updates should only require a reindex of the updated node and sometimes the parent node. Older version of eXist indeed started to reindex the entire document, but this was a bug if I remember well.

adamretter:
@joewiz XQuery Update just modifies the nodes that it needs to.

The problem I am solving is when there is a document in the database, you open it (say in Oxygen) and just make some changes to it, and save it. eXist currently sees that as an entirely new document and basically completely deletes the old document and replaces it with a new document. This change allows indexes to be aware that this is a replace operation.

My Algolia index takes advantage of this "replace" mode by diffing the previous document nodes and the new document nodes, and then only sending additions, updates and deletions to the remote Algolia server. This means that it doesn't have to remove all entries for the document from the remote Algolia Index and reindexing the whole document again. Algolia charges for access to their API based on the amount of updates you make to your indexes, so doing differential updates can drastically reduce the cost.

src/org/exist/collections/MutableCollection.java
shabanovd: `IndexController.getStreamListener()` return same listener until it reset by set it to `null`. `IndexController.getStreamListener(DocumentImpl, ReindexMode)` call `setMode` and it set `listener` to `null` when mode change.

adamretter: @shabanovd Yes but that happens before, so it doesn't affect the sequence of events as I described. I know this to be the case as I have tested it.

shabanovd: the difference that you do
```
setMode(REPLACE_DOCUMENT)
startReplaceDocument()

setMode(REMOVE_ALL_NODES)
startIndexDocument()
...
endIndexDocument()

setMode(STORE)
startIndexDocument()
...
endIndexDocument()

**setMode(REPLACE_DOCUMENT)**
endReplaceDocument()```

That mean `startIndexDocument` + `endIndexDocument` happen for same listener on `REMOVE_ALL_NODES` & `STORE`, but your `startReplaceDocument` + `endReplaceDocument` happen on different listeners. That is design bug, IMHO.

File: src/org/exist/indexing/StreamListener.java (hunk 2 above)
Comment (shabanovd) [inline src/org/exist/indexing/StreamListener.java]: You miss last `setMode(REPLACE_DOCUMENT)` before `endReplaceDocument()`

File: src/org/exist/collections/MutableCollection.java (hunk 1 above)
Comment (adamretter) [inline src/org/exist/collections/MutableCollection.java]: You are still mistaken, running this through the debugger shows that §startReplaceDocument` and `endReplaceDocument` happen on the same listener as `REMOVE_ALL_NODES` and `STORE`.

Just because the `listener` is set to `null` in `setMode` and `setDocument`, does not mean that you will get a new `listener`, it is entirely at the discretion of the `IndexWorker`;  None of the existing implementations of `IndexWorker` create a new listener when `getListener` is called.

Please check the code thoroughly first!

File: src/org/exist/indexing/StreamListener.java (hunk 2 above)
Comment (adamretter) [inline src/org/exist/indexing/StreamListener.java]: @shabanovd no I don't the current approach is intentional

File: src/org/exist/collections/MutableCollection.java (hunk 1 above)
Comment (shabanovd) [inline src/org/exist/collections/MutableCollection.java]: Why do you think that I did not check code before?

Do you mean that it's same `listener` of your index or same `listener` from IndexController?

Comment (adamretter) [inline src/org/exist/collections/MutableCollection.java]: @shabanovd Well I guessed you hadn't checked thoroughly otherwise I wouldn't be explaining so much.

The `listener` in the index controller is taken from the index worker, so they are effectively the same thing.

Comment (shabanovd) [inline src/org/exist/collections/MutableCollection.java]: Nope, they are different. `IndexController` create chain by calling [this](https://github.com/adamretter/exist/blob/c841f6b7d17308be143b1e218a40750e0697264d/src/org/exist/indexing/IndexController.java#L315), but there was no restrictions to avoid `listener` creation by `IndexWorker`.

By current PR you force `IndexWorker` to return same `listener`. That is contract change, no?

Comment (adamretter) [inline src/org/exist/collections/MutableCollection.java]: @shabanovd I haven't forced anything, `IndexWorker` has the same event flow as before. The chaining does not effect the behaviour I expect.

File: src/org/exist/indexing/StreamListener.java (hunk 2 above)
Comment (shabanovd) [inline src/org/exist/indexing/StreamListener.java]: that call is [here](https://github.com/adamretter/exist/blob/c841f6b7d17308be143b1e218a40750e0697264d/src/org/exist/collections/MutableCollection.java#L1261)

**Comment by joewiz** [comment]
Thanks for the explanations @wolfgangmm and @adamretter. I can see why this makes sense as a change to indexing - namely, you're updating an index with the new content - but is there a reason you didn't approach this using triggers? (The `before-update-document` and `after-update-document` come to mind.) I'm just curious - this seems like a useful change as is.

File: src/org/exist/collections/MutableCollection.java (hunk 1 above)
Comment (shabanovd) [inline src/org/exist/collections/MutableCollection.java]: @adamretter if you don't want to discuss and 110% sure that you are right, how I can't show you where you are wrong?

I already say everything I wish to, no more comments.

Comment (adamretter) [inline src/org/exist/collections/MutableCollection.java]: I am happy to discuss, I just cannot see the problem that you say is there. I have tested my indexing extension with a real index that communicates with Algolia and it appears to work well

File: src/org/exist/indexing/StreamListener.java (hunk 2 above)
Comment (adamretter) [inline src/org/exist/indexing/StreamListener.java]: Ah I see, sorry I thought you were saying the opposite. Okay I need to make a small change so that is:
```java
final StreamListener listener = broker.getIndexController().getStreamListener();
listener.endReplaceDocument(transaction);
```

dizzzz:
@wolfgangmm are you OK with the PR?

src/org/exist/indexing/StreamListener.java
shabanovd: And last question - how critical to have `REPLACE_DOCUMENT`? Will you do same by having only `startReplaceDocument` and `endReplaceDocument` methods?

adamretter: Yes that should stay

shabanovd: I do not see reason for it
```
setMode(REPLACE_DOCUMENT)
startReplaceDocument()

setMode(REMOVE_ALL_NODES)
startIndexDocument()
...
endIndexDocument()

setMode(STORE)
startIndexDocument()
...
endIndexDocument()

endReplaceDocument()```

You call `startReplaceDocument` at mode `REPLACE_DOCUMENT`, and `endReplaceDocument` at mode `STORE`.  That looks a bit strange as minimum.

Why you can't call `startReplaceDocument` at mode `UNKNOWN` (for example)?

Issue #1041, opened 2016-06-08
doc() no longer accepts file: protocol


rvdb:
In the current development code (exist-3.0.RC2-develop, rev. a88848d), the doc() function no longer accepts the `file:` protocol.

For example, suppose I want to fetch a document that lives at F:/eXist/eXist-git/log4j2.xml, following function call:

```
doc('file:///F:/eXist/eXist-git/log4j2.xml')
```

throws following error: 

```
org.xmldb.api.base.XMLDBException: err:FODC0005 Illegal char <:> at index 4: file:///F:/eXist/eXist-git/log4j2.xml [at line 1, column 1, source: doc('file:///F:/eXist/eXist-git/log4j2.xml')]
    at org.exist.xmldb.RemoteXPathQueryService.throwException(RemoteXPathQueryService.java:181)
    at org.exist.xmldb.RemoteXPathQueryService.query(RemoteXPathQueryService.java:106)
    at org.exist.xmldb.RemoteXPathQueryService.query(RemoteXPathQueryService.java:73)
    at org.exist.xmldb.RemoteXPathQueryService.execute(RemoteXPathQueryService.java:441)
    at org.exist.client.QueryDialog$QueryThread.run(QueryDialog.java:612)
Caused by: org.exist.xquery.XPathException: exerr:ERROR err:FODC0005 Illegal char <:> at index 4: file:///F:/eXist/eXist-git/log4j2.xml [at line 1, column 1, source: doc('file:///F:/eXist/eXist-git/log4j2.xml')] [at line 1, column 1]
    at org.exist.xmldb.RemoteXPathQueryService.throwException(RemoteXPathQueryService.java:180)
    ... 4 more
Caused by: org.exist.xquery.XPathException: exerr:ERROR err:FODC0005 Illegal char <:> at index 4: file:///F:/eXist/eXist-git/log4j2.xml [at line 1, column 1, source: doc('file:///F:/eXist/eXist-git/log4j2.xml')] [at line 1, column 1]
    at org.exist.xmldb.RemoteXPathQueryService.throwException(RemoteXPathQueryService.java:180)
    at org.exist.xmldb.RemoteXPathQueryService.query(RemoteXPathQueryService.java:106)
    at org.exist.xmldb.RemoteXPathQueryService.query(RemoteXPathQueryService.java:73)
    at org.exist.xmldb.RemoteXPathQueryService.execute(RemoteXPathQueryService.java:441)
[...]

PR #1221, opened 2017-01-12
Resolving XQuery modules or loading documents from disk fails

### LINKED ISSUE #1217: XQuery import statement fails in current development code, when run from the file system
In eXist-3.0RC2 and the latest nightly build (d555767), XQuery import statements fail when executed from the file system. 

Given following main XQuery file `main.xq`:

    xquery version "3.0";
    import module namespace x="x" at "imported.xq";
    x:echo("main query")

...and following imported XQuery file `imported.xq`:

    xquery version "3.0";
    module namespace x="x";
    declare function x:echo($string) {
      let $output := "imported query"
      return string-join(($string, $output), ' -- ')
    };

When both files are stored in `${exist.home}/webapps/test-import`, execution of the main XQuery file at http://localhost:8080/exist/test-import/main.xq fails with following error:

    Illegal char <:> at index 4: file:///F:/devtools/eXist/eXist-nightly/webapp/test-import/imported.xq

This worked as expected in eXist-3.0RC1 and earlier. Yet, when the files are stored in and run from the db, the XQuery works as expected without errors.

I'll include the files in this file: [test-import.zip](https://github.com/eXist-db/exist/files/696275/test-import.zip)

wolfgangmm:
InvalidPathException is thrown while probing locations in SourceFactory. Issue only pops up when running tests on windows.

Closes #1217 and #1041

### DIFF
--- a/src/org/exist/source/SourceFactory.java
+++ b/src/org/exist/source/SourceFactory.java
@@ -2,21 +2,21 @@
  *  eXist Open Source Native XML Database
  *  Copyright (C) 2001-04 The eXist Project
  *  http://exist-db.org
- *  
+ *
  *  This program is free software; you can redistribute it and/or
  *  modify it under the terms of the GNU Lesser General Public License
  *  as published by the Free Software Foundation; either version 2
  *  of the License, or (at your option) any later version.
- *  
+ *
  *  This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


joewiz:
Wow. Thank you @wolfgangmm!

joewiz:
If only CI systems could be configured to run the test suite on multiple OSes to catch regressions like this...

PR #1222, opened 2017-01-13
Fix recovery issues under windows

wolfgangmm:
* Properly close lucene-based indexes before removing and use nio for deleting the files
* Fix test suite so it runs clean again

### DIFF
--- a/test/src/org/exist/storage/MoveResourceTest.java
+++ b/test/src/org/exist/storage/MoveResourceTest.java
@@ -45,6 +45,8 @@ import org.exist.xmldb.CollectionManagementServiceImpl;
 import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


test/src/org/exist/storage/MoveResourceTest.java
```diff
@@ -54,10 +56,11 @@
 import org.xmldb.api.base.Resource;
 import org.xmldb.api.base.XMLDBException;
 
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 public class MoveResourceTest {
 
     @Test
-    public void store() throws LockException, SAXException, PermissionDeniedException, EXistException, IOException {
+    public void store1() throws LockException, SAXException, PermissionDeniedException, EXistException, IOException {
```
shabanovd: did you rename methods to guarantee order of execution?

PR #1223, opened 2017-01-14
Fix variable declaration

### LINKED ISSUE #1114: ft:query-field stops working when using declare variable
I have indexed a bunch of TEI files into a Lucene field using a collection.xconf like:

``` xml
    <index xmlns:tei="http://www.tei-c.org/ns/1.0" xmlns:xs="http://www.w3.org/2001/XMLSchema">
        <fulltext default="none" attributes="false"/>
        <lucene>
            <text field="text" qname="tei:l"/>
            <text field="text" qname="tei:p"/>
            <text field="text" qname="tei:head"/>
            <text field="text" qname="tei:stage"/>
        </lucene>
    </index>
</collection>
```

Now, if I run the following XQuery, I get a few results as expected:

``` xquery
xquery version "3.0";

<result>{ft:query-field('text', 'wiesenheu')}</result>
```

However, if I introduce a variable declaration such that the query is changed to the following, ft:query-field does no longer yields results — I only get the empty `<result/>` element:

``` xquery
xquery version "3.0";
declare variable $foo := 'bar';

<result>{ft:query-field('text', 'wiesenheu')}</result>
```

Tested using eXist 3.0 RC1, but I am quite sure the same problem already exists in eXist 2.2.

wolfgangmm:
A variable declaration should not return a context sequence, not even an empty one or followup expressions may start from a wrong context.

Closes #1114.

### DIFF
--- a/src/org/exist/storage/ProcessMonitor.java
+++ b/src/org/exist/storage/ProcessMonitor.java
@@ -26,6 +26,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.exist.http.servlets.RequestWrapper;
 import org.exist.http.urlrewrite.XQueryURLRewrite;
+import org.exist.source.Source;
 import org.exist.util.Configuration;
 import org.exist.xquery.Variable;
 import org.exist.xquery.XPathException;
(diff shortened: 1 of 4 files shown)


src/org/exist/storage/ProcessMonitor.java
```diff
@@ -162,7 +163,8 @@ public void queryCompleted(XQueryWatchDog watchdog) {
         final long elapsed = System.currentTimeMillis() - watchdog.getStartTime();
         if (found && elapsed > minTime) {
             synchronized (history) {
-                final String sourceKey = watchdog.getContext().getSource().path();
+                final Source source = watchdog.getContext().getSource();
```
shabanovd: smell like a bug? should it be logged with the trace?

dizzzz: @shabanovd I missed your comment.... was I too quick with pulling-in?

shabanovd: well, it can go into different PR if agreed

Issue #1226, opened 2017-01-15
range:index-keys-for-field with dynamic context only works correctly when wrapped inside dynamic function call


olvidalo:
```xq
range:index-keys-for-field("some-field", function($key, $nums) { $key }, 10)`
```
✅  returns all keys for `some-field` for the statically known documents.  
  <br/>
```xq
collection("/db/some-collection/")/range:index-keys-for-field("some-field", function($key, $nums) { $key }, 10)`
```
❌ will evaluate the function for each document in the collection instead of for the context document set at once, so it returns the keys multiple times for each document instead of once for the collection 
  <br/>
```xq
let $func := function-lookup(xs:QName("range:index-keys-for-field"), 3) 
return  collection("/db/some-collection/")/$func("some-field", function($key, $nums) { $key }, 10)
```
✅ however, wrapping it inside a dynamic function will make it work
  <br/>
```xq
let $func := function($a, $b, $c) {
    range:index-keys-for-field($a, $b, $c)
}

return  collection("/db/some-collection/")/$func("some-field", function($key, $nums) { $key }, 10)
```
✅ also works
  <br/>
```xq
declare function local:func($a, $b, $c) {
  range:index-keys-for-field($a, $b, $c )  
};

return  collection("/db/some-collection/")/local:func("some-field", function($key, $nums) { $key }, 10)
```
❌ does not work
  <br/>

It might have something to do with the expression dependencies being incorrectly set at one point down the stack (the evaluation path leading to the incorrect results seems to start in `PathExpr.eval()` when the expression's dependencies are determined)  but as I have no idea what I'm actually talking about I'll leave it at that for the time being...

There'll be a pull request with a test coming in shortly

dizzzz:
did you test this with eXist-db 3.0RC2/develop ?

olvidalo:
Tested it with the current develop commit

duncdrum:
if the test is passing, this seems to be fixed?

dizzzz:
is it passing? maybe we could add a simple test to the suite for this>

PR #1177, opened 2016-11-11
Add test showing bug in typeswitch - eXist allows invalid "empty()" check

joewiz:
This PR adds a failing test (with `%test:pending` annotation) demonstrating the problem documented at https://github.com/eXist-db/exist/issues/1155.

PR #1227, opened 2017-01-15
Add test showing problem with dynamic context in range:index-keys-for-field

olvidalo:
demonstrates issue in #1226

### DIFF
--- a/extensions/indexes/range/test/src/xquery/index-keys.xql
+++ b/extensions/indexes/range/test/src/xquery/index-keys.xql
@@ -1,0 +1,102 @@
+xquery version "3.0";
+
+module namespace rtik="http://exist-db.org/xquery/range/test/index-keys";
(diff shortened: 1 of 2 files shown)


shabanovd:
hmm,  tests pass

dizzzz:
Probably the invocation of the script need to be configured. I remember a file for this....

olvidalo:
I just added the pending annotation because @joewiz did it in #1177 and I thought this must be common practise. Apparently it isn't :) Pending tests don't get executed at all but show up in the report as such. So this could be merged even before fixing the bug without breaking CI. Might be useful as this contains some passing tests for range:index-keys-for-field which hadn't been tested before.

But I can remove the %pending of course. Let me know..

dizzzz:
I missed the %pending part :-)

PR #1239, opened 2017-01-30
Package repo stability

wolfgangmm:
The expath package repository has been a source of instability, in particular for windows users. A damaged repo would result in eXist failing to start, even though the database was completely ok. This PR thus tries to make the repo more reliable without rewriting all of it (which needs to be done at some point):

* make sure eXist will come up even if expathrepo contains errors, so users have a chance to repair
* [forked](https://github.com/wolfgangmm/expath-pkg-java) [expath-pkg-java](https://github.com/fgeorges/expath-pkg-java) to partially move to NIO for important methods, like moving directories
* fix error reporting by eXist in case anything goes wrong during a package install


src/org/exist/start/start.config
```diff
@@ -51,7 +51,7 @@ lib/core/log4j-core-%latest%.jar                   always
 lib/core/log4j-jul-%latest%.jar                    always
 lib/core/log4j-slf4j-impl-%latest%.jar             always
 lib/core/slf4j-api-%latest%.jar                    always
-lib/core/pkg-repo.jar                              always
+lib/core/pkg-java-fork.jar                         always
```
shabanovd: did you send PR to expath-pkg-java?

wolfgangmm: Nope, my switch to Java NIO is not complete yet. More work needs to be done, but for now, the goal is just to get rid of those issues plaguing windows users in particular.

PR #1240, opened 2017-01-30
[bugfix] Range index query optimizer fix

### LINKED ISSUE #1238: Applying range index changes query results
I noticed some strange results tonight using current develop (9e8a2cf), which I reduced to the following test case. 

To summarize, applying a (new) range index changes the result of a query (1) traversing the parent or ancestor axis and (2) making use of an inner predicate. 

The test document:

```xml
<p class="foo">
    <a class="bar"/>
</p>
```

The test query:

```xquery
//a[./parent::p[@class = 'foo']]
```

Without a range index, the query returns the expected result: `<a class="bar"/>`. However, when a range index is applied to `@class`, the query returns an empty sequence instead.

In the complete test below, the problem is revealed in test 1b - which returns `false`, whereas it should match 1a's result, `true`. 

Test sets 2 and 3 illustrate variants that do not trigger the problem. Test set 2 drops the inner predicate, and test set 3 switches from parent to child axis. 

```xquery
xquery version "3.1";

(: setup :)

let $login := xmldb:login('/db', 'admin', '')
let $test1-col := xmldb:create-collection('/db', 'test1')
let $test1-config-col := xmldb:create-collection('/db/system/config/db', 'test1')
let $test2-col := xmldb:create-collection('/db', 'test2')
let $test2-config-col := xmldb:create-collection('/db/system/config/db', 'test2')
let $test1-xconf := 
            <fulltext default="none" attributes="false"/>
        </index>
    </collection>
let $test2-xconf := 
            <fulltext default="none" attributes="false"/>
            <range>
                <create qname="@class" type="xs:string"/>
            </range>
        </index>
    </collection>
let $store-xconfs := 
    (
        xmldb:store($test1-config-col, 'collection.xconf', $test1-xconf),
        xmldb:store($test2-config-col, 'collection.xconf', $test2-xconf)
    )
let $doc := 
    <p class="foo">
        <a class="bar"/>
    </p>
let $doc1 := xmldb:store($test1-col, 'doc.xml', $doc)
let $doc2 := xmldb:store($test2-col, 'doc.xml', $doc)

(: test :)

return
    element tests {
        element test { 
            attribute n { "1a" },
            element query { "exists(doc($doc1)//a[./parent::p[@class = 'foo']])" },
            element result { exists(doc($doc1)//a[./parent::p[@class = 'foo']]) }
        },
        element test { 
            attribute n { "1b" },
            element query { "exists(doc($doc2)//a[./parent::p[@class = 'foo']])" },
            element result { exists(doc($doc2)//a[./parent::p[@class = 'foo']]) }
        },
        element test { 
            attribute n { "2a" },
            element query { "exists(doc($doc1)//a[./parent::p/@class = 'foo'])" },
            element result { exists(doc($doc1)//a[./parent::p/@class = 'foo']) }
        },
        element test { 
            attribute n { "2b" },
            element query { "exists(doc($doc2)//a[./parent::p/@class = 'foo'])" },
            element result { exists(doc($doc2)//a[./parent::p/@class = 'foo']) }
        },
        element test { 
            attribute n { "3a" },
            element query { "exists(doc($doc1)/p[./child::a[@class = 'bar']])" },
            element result { exists(doc($doc1)/p[./child::a[@class = 'bar']]) }
        },
        element test { 
            attribute n { "3b" },
            element query { "exists(doc($doc2)/p[./child::a[@class = 'bar']])" },
            element result { exists(doc($doc2)/p[./child::a[@class = 'bar']]) }
        }
    }
```

Actual results:

```xml
<tests>
    <test n="1a">
        <query>exists(doc($doc1)//a[./parent::p[@class = 'foo']])</query>
        <result>true</result>
    </test>
    <test n="1b">
        <query>exists(doc($doc2)//a[./parent::p[@class = 'foo']])</query>
        <result>false</result>
    </test>
    <test n="2a">
        <query>exists(doc($doc1)//a[./parent::p/@class = 'foo'])</query>
        <result>true</result>
    </test>
    <test n="2b">
        <query>exists(doc($doc2)//a[./parent::p/@class = 'foo'])</query>
        <result>true</result>
    </test>
    <test n="3a">
        <query>exists(doc($doc1)/p[./child::a[@class = 'bar']])</query>
        <result>true</result>
    </test>
    <test n="3b">
        <query>exists(doc($doc2)/p[./child::a[@class = 'bar']])</query>
        <result>true</result>
    </test>
</tests>
```

wolfgangmm:
Fix query optimizer to not rewrite query using new range index if context step is on a reverse axis. Results will be wrong. Closes #1238.

Add test cases.

### DIFF
--- a/extensions/indexes/range/src/org/exist/xquery/modules/range/RangeQueryRewriter.java
+++ b/extensions/indexes/range/src/org/exist/xquery/modules/range/RangeQueryRewriter.java
@@ -41,6 +41,12 @@ public class RangeQueryRewriter extends QueryRewriter {
 
     @Override
     public Pragma rewriteLocationStep(LocationStep locationStep) throws XPathException {
+        int axis = locationStep.getAxis();
+        if (!(axis == Constants.CHILD_AXIS || axis == Constants.DESCENDANT_AXIS ||
+                axis == Constants.DESCENDANT_SELF_AXIS || axis == Constants.ATTRIBUTE_AXIS ||
+                axis == Constants.DESCENDANT_ATTRIBUTE_AXIS || axis == Constants.SELF_AXIS)) {
+            return null;
+        }
         if (locationStep.hasPredicates()) {
             Expression parentExpr = locationStep.getParentExpression();
             if ((parentExpr instanceof RewritableExpression)) {
@@ -69,7 +75,6 @@ public class RangeQueryRewriter extends QueryRewriter {
                     }
 
                     // check if inner steps are on an axis we can optimize
-                    final int axis;
                     if (innerExpr instanceof InternalFunctionCall) {
                         InternalFunctionCall fcall = (InternalFunctionCall) innerExpr;
                         axis = ((Optimizable) fcall.getFunction()).getOptimizeAxis();
--- a/extensions/indexes/range/test/src/xquery/optimizer.xql
+++ b/extensions/indexes/range/test/src/xquery/optimizer.xql
@@ -536,7 +536,7 @@ declare
     %test:args("main", "official", "Dorfprozelten")
     %test:assertEquals("Dorfprozelten")
 function ot:equality-field-nested($type as xs:string, $subtype as xs:string, $name as xs:string) {
-    //tei:placeName[@type = $type][@subtype = $subtype][. = $name]/text()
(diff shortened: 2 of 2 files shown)


joewiz:
@wolfgangmm Thank you!

PR #1250, opened 2017-02-05
define void-elements for HTML5 writer

### LINKED ISSUE #46: Empty tag "source" should be self-closing in HTML5
When serializing output as HTML5, the `<source>` tag is output with a separate closing tag, instead of being a self-closing tag -- it's an empty tag according to the specification:

http://www.w3.org/TR/html51/embedded-content-0.html#the-source-element

I think this is because the HTML5 serializer is a subclass of the XHTML serializer, and this uses a list of element names (emptyTags) to determine which tags should be output as self-closing tags, and "source" isn't on the list:

https://github.com/eXist-db/exist/blob/develop/src/org/exist/util/serializer/XHTMLWriter.java

Outputting `<source>` with a separate closing tag results in validation errors from the HTML5 validator.

shabanovd:
close #46

### DIFF
--- a/src/org/exist/util/serializer/XHTML5Writer.java
+++ b/src/org/exist/util/serializer/XHTML5Writer.java
@@ -9,6 +9,26 @@ import org.exist.util.hashtable.ObjectHashSet;
  */
 public class XHTML5Writer extends XHTMLWriter {
 
+    protected final static ObjectHashSet<String> EMPTY_TAGS = new ObjectHashSet<>(31);
+    static {
+        //https://www.w3.org/TR/html5/syntax.html#void-elements
+        EMPTY_TAGS.add("area");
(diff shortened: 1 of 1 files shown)


dizzzz:
I think the PR is not complete?

shabanovd:
why do you think so?

dizzzz:
so.... just alone that I ask the question should trigger you :-)

so subclasses define a new `protected final static ObjectHashSet` which was already defined in the superclass. I am not sure if this is 'clean, clear and required' ....

shabanovd:
the set are different, that's why new set created. I do not see how it can be related to your question of incompleteness ;-)

it's static, as I understand it, to make that set to be singleton

dizzzz:
yes you are write ; it is the drawback of reviewing code in a mobile app

PR #1251, opened 2017-02-06
General cleanup

adamretter:
A collection of small improvements and bug fixes.


src/org/exist/protocolhandler/embedded/EmbeddedDownload.java
```diff
@@ -130,32 +131,32 @@ public void stream(XmldbURL xmldbURL, OutputStream os, Subject user) throws IOEx
...
+                                serializer.serialize(resource, w);
+                            }
 
                         } else {
                             broker.readBinaryResource((BinaryDocument) resource, os);
                         }
                     }
                 } finally {
-                    if(resource != null){
+                    if (collection != null) {
```
shabanovd: why do you check unlock order?

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
```diff
@@ -70,41 +74,50 @@ public EmbeddedDownloadThread(XmldbURL url, OutputStream bos) {
...
+            if (brokerPool == null) {
+                this.brokerPool = BrokerPool.getInstance(url.getInstanceName());
+            } else {
+                this.brokerPool = brokerPool;
+            }
             
-            subject = brokerPool.getActiveBroker().getCurrentSubject();
-        } catch (Throwable e) {
-            //e.printStackTrace();
+            this.subject = this.brokerPool.getActiveBroker().getCurrentSubject();
```
shabanovd: why `this.` required?

src/org/exist/protocolhandler/embedded/EmbeddedDownload.java
adamretter: So that it is symmetrical which reduces opportunities for deadlocks.

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
adamretter: Otherwise if the Broker pool parameter is null  you will get an NPE

dizzzz: ??

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
```diff
@@ -38,29 +39,32 @@
...
+        this.xmldbURL = url;
         this.bos = bos;
         
         try {
-            BrokerPool pool = BrokerPool.getInstance(url.getInstanceName());
-            subject = pool.getActiveBroker().getCurrentSubject();
-        } catch (Throwable e) {
-            //e.printStackTrace();
+            final BrokerPool pool = BrokerPool.getInstance(url.getInstanceName());
+            this.subject = pool.getActiveBroker().getCurrentSubject();
```
dizzzz: I see that you do this on more locations..... not needed, but is it a good practice I don;t know of?

src/org/exist/protocolhandler/embedded/EmbeddedInputStream.java
```diff
@@ -33,66 +33,66 @@
...
+    public EmbeddedInputStream(final BrokerPool brokerPool, final XmldbURL xmldbURL) throws IOException {
 
-        logger.debug("Initializing EmbeddedInputStream");
+        if(LOG.isDebugEnabled()) {
+            LOG.debug("Initializing EmbeddedInputStream");
+        }
 
-        bis = new PipedInputStream(2048);
-        bos = new PipedOutputStream(bis);
+        this.bis = new PipedInputStream(4096);
```
dizzzz: here the `this.` as well...  I don't understand (yet)

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
shabanovd: @adamretter 1st, it's very confusing to have equal names for parameter and class field, but it's getting more confusing to have parameter that "modifiable".

2nd, my question was on `this.subject`, not `this.brokerPool`

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
adamretter: `this` makes it immediately obvious that it's a class member as opposed to local

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
adamretter: Java standard is to have equal name for class member and setter parameter.

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
dizzzz: but it is not really needed.... is it?

src/org/exist/storage/BrokerWatchdog.java
```diff
@@ -8,10 +27,19 @@
 
 import org.exist.EXistException;
 
+/**
+ * Traces the lease of a Broker
+ *
+ * Note that tracing the stack is expensive
+ * this should only be done in debug mode by
+ * enabling the System Property {@link #TRACE_BROKERS_PROPERTY_NAME}
```
dizzzz: We need to document this, 'somewhere'....

test/src/org/exist/test/EmbeddedExistTester.java
```diff
@@ -1,140 +0,0 @@
-/*
```
dizzzz: is it an option to have this 'deprecated' for a little time? removing it right away might be a bit harsh?

adamretter: @dizzzz After the refactoring it is unused code, so it seems to me there is no point to keep it.

src/org/exist/storage/BrokerWatchdog.java
adamretter: That was me documenting existing behaviour that was undocumented ;-)

I think perhaps you mean we should document the System property? If so, I totally agree.

There are a whole load of System properties that are embedded in eXist code which are undocumented. I discover new ones all the time. So I guess a page in the documentation about System properties would be good - obviously some of them are only relevant to developers.

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
adamretter: No, it will compile without. However personally I prefer clear intentions when reading code (even if it was written by me ;-))

src/org/exist/protocolhandler/embedded/EmbeddedInputStream.java
adamretter: Its an intention thing. When reading code, from only looking at the line:

```java
this.bis = new PipedInputStream(4096);
```

I can tell immediately that this line mutates shared state, i.e. it modifies a class member and not a local variable. Because the intention is clear, I don't have to search around to find out if `bis` is mutable shared state or not.

src/org/exist/storage/BrokerWatchdog.java
dizzzz: Are there a few or many? I guess spending some time in documenting these is a good thing; I'll dive into this later today...

test/src/org/exist/test/EmbeddedExistTester.java
dizzzz: except for @shabanovd  PR :-(

adamretter:
@dizzzz This PR introduces no API changes. So there is no risk of incompatibility for extensions or anything else.

src/org/exist/protocolhandler/embedded/EmbeddedInputStream.java
dizzzz: Ah, ok. That makes sense to me, I'll keep this in mind when refactoring / creating new code. Thnx....

test/src/org/exist/test/EmbeddedExistTester.java
adamretter: @dizzzz Sure, but he only used that because this PR has been languishing and not merged. As he is unwilling to do so, I will fix his PR.

src/org/exist/storage/BrokerWatchdog.java
adamretter: @dizzzz Many! If you look at org.exist.util.Configuration - you will also find that many (but not all) options from conf.xml are overridable by setting System properties. Also much of the eXist jetty config is overridable with System properties (and even was before I moved us to Jetty 9). Also there is a good spattering of `System.getProperty` in other places of the eXist code base.

If you start a documentation project, I am happy to contribute some too :-)

src/org/exist/protocolhandler/embedded/EmbeddedInputStream.java
adamretter: Cool. So to add further explanation to why that intention is important to me. As soon as I see mutation of shared state, I ask myself two questions:

1) Does this really need to be shared state? If not, I refactor to avoid it. Eliminating shared state allows me to ignore any worries of concurrent modification. Even if the code is currently intended to be single-threaded, the removal of shared state gives me the option to use this code concurrently in future.

2) If this does need to be shared state, what is the concurrency model? If it is anything other than single-threaded, what should I do to ensure the correct semantics when reading/writing this state.

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
shabanovd: no problems, until you force others to follow your coding style ;-)

src/org/exist/protocolhandler/embedded/EmbeddedDownloadThread.java
shabanovd: Until you code for android or at eclipse community ;-)

adamretter: Eeek! I hope I never have to code for the Eclipse community

shabanovd: I think you didn't get the point - there are a lot of different code styles around

adamretter: @shabanovd yes I got that ;-)

PR #1252, opened 2017-02-08
[feature] extend range index conditions

olvidalo:
This adds a few extensions to the attribute range index condition introduced in #1233:

- ability to specify an operator for the condition, implemented operators are`eq, ne, lt, gt, le, ge, starts-with, ends-with, contains, matches`. Can be specified using an `operator` attribute on the `condition` element. If the attribute is missing, `eq` is assumed as default.

- case insensitive comparison by specifing `case="no"` on the `condition` element

- numeric comparison by specifing `numeric="yes"`. When enabled, `01.0` will equal `1` and `2` will be less than `110`, for example. By default, string matching is assumed. The rewriter will respect the type of the value (string, numeric) when matching a condition to a predicate.

- rewriting improvements:
    - the order of attribute name and value can be reversed in the predicate, respecting semantics of the operator (`[@a = "x"]` and `["x" = @a]` will both match a condition on `a`, for example)
  - variables and function calls in predicates will be resolved to a value

- some refactoring:
  - `RangeIndexConfigCondition` is now an abstract base class with `RangeIndexConfigAttributeCondition` being a subclass. This should make it easier to implement other types of conditions in the future
  - moved most of the rewriting code from `OptimizeFieldPragma` to the condition itself as it will differ for different kinds of conditions

I'll be happy to add some documentation on this to the "New Range Index" page later on...

### DIFF
--- a/extensions/indexes/range/src/org/exist/indexing/range/ComplexRangeIndexConfigElement.java
+++ b/extensions/indexes/range/src/org/exist/indexing/range/ComplexRangeIndexConfigElement.java
@@ -27,6 +27,7 @@ import org.apache.lucene.analysis.Analyzer;
 import org.exist.dom.QName;
 import org.exist.storage.NodePath;
 import org.exist.util.DatabaseConfigurationException;
+import org.exist.xquery.Predicate;
 import org.exist.xquery.value.Type;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
@@ -66,7 +67,7 @@ public class ComplexRangeIndexConfigElement extends RangeIndexConfigElement {
                     RangeIndexConfigField field = new RangeIndexConfigField(path, (Element) child, namespaces);
                     fields.put(field.getName(), field);
                 } else if (CONDITION_ELEMENT.equals(child.getLocalName())){
-                    conditions.add(new RangeIndexConfigCondition((Element) child, path));
+                    conditions.add(new RangeIndexConfigAttributeCondition((Element) child, path));
                 } else if (FILTER_ELEMENT.equals(child.getLocalName())) {
                     analyzer.addFilter((Element) child);
                 } else {
@@ -162,9 +163,9 @@ public class ComplexRangeIndexConfigElement extends RangeIndexConfigElement {
         return true;
     }
 
-    public boolean findCondition(QName lhe, String rhe, RangeIndex.Operator operator) {
+    public boolean findCondition(Predicate predicate) {
         for (RangeIndexConfigCondition condition : conditions) {
-            if (condition.find(lhe, rhe, operator))
+            if (condition.find(predicate))
                 return true;
         }
 
--- a/extensions/indexes/range/src/org/exist/indexing/range/RangeIndex.java
+++ b/extensions/indexes/range/src/org/exist/indexing/range/RangeIndex.java
@@ -29,6 +29,9 @@ import org.exist.indexing.IndexWorker;
 import org.exist.indexing.lucene.LuceneIndex;
 import org.exist.storage.DBBroker;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Main implementation class for the new range index. This extends the existing LuceneIndex.
  *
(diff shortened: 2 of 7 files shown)


joewiz:
Looks great! Thanks again for contributing these really significant enhancements!

extensions/indexes/range/src/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
```diff
@@ -26,76 +26,380 @@
...
+
+        if (expr instanceof AtomicValue) {
+            return (AtomicValue) expr;
+        }
+        else if (expr instanceof LiteralValue) {
+            return ((LiteralValue) expr).getValue();
+
+        } else if (expr instanceof VariableReference || expr instanceof Function) {
+            try {
+                final Sequence result = expr.eval(null);
```
olvidalo: Now that this is getting some attention... :)

I'm wondering if this could be a problem if someone would, for whatever reason, put an updating function in the predicate or something else with side effects. Maybe the expression could be evaluated in some kind of read only copy of the xquery context? Or wouldn't that be necessary? 

I also thought about possible security risks of evaluating an expression during rewriting, but as it is evaluated with the same rights as the script itself, that shouldn't be a problem per se, right?

adamretter: You don't need to worry about the security, the same user as executed the query will be used.

extensions/indexes/range/src/org/exist/indexing/range/ComplexRangeIndexConfigElement.java
```diff
@@ -162,9 +163,9 @@ public boolean matchConditions(Node node) {
         return true;
     }
 
-    public boolean findCondition(QName lhe, String rhe, RangeIndex.Operator operator) {
+    public boolean findCondition(Predicate predicate) {
```
adamretter: In future please consider using `final` on all variables that are not reassigned.

extensions/indexes/range/src/org/exist/indexing/range/RangeIndex.java
```diff
@@ -58,6 +61,15 @@
         private final String name;
         private final boolean supportsCollation;
 
+        private static final Map<String, Operator> LOOKUP_MAP;
+
+        static {
+            LOOKUP_MAP = new HashMap<String, Operator>();
```
adamretter: Could you use Java 7 Diamond Syntax here?

extensions/indexes/range/src/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
```diff
@@ -0,0 +1,405 @@
+/*
+ *  eXist Open Source Native XML Database
+ *  Copyright (C) 2013 The eXist Project
```
adamretter: Should be 2017

extensions/indexes/range/src/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
```diff
@@ -0,0 +1,405 @@
...
+    private Pattern pattern = null;
+
+    public RangeIndexConfigAttributeCondition(Element elem, NodePath parentPath) throws DatabaseConfigurationException {
+
+        if (parentPath.getLastComponent().getNameType() == ElementValue.ATTRIBUTE) {
+            throw new DatabaseConfigurationException("Range index module: Attribute condition cannot be defined for an attribute:" + parentPath.toString());
+        }
+
+        this.attributeName = elem.getAttribute("attribute");
+        if (this.attributeName == null || this.attributeName.length() == 0) {
```
adamretter: Use `isEmpty()` and `not length() == 0`

extensions/indexes/range/src/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
```diff
@@ -0,0 +1,405 @@
...
+        this.caseSensitive = (caseString != null && !caseString.equalsIgnoreCase("no"));
+        this.numericComparison = (numericString != null && numericString.equalsIgnoreCase("yes"));
+
+        // try to create a pattern matcher for a 'matches' condition
+        if (this.operator == Operator.MATCH) {
+            final int flags = this.caseSensitive ? 0 : CASE_INSENSITIVE;
+            try {
+                this.pattern = Pattern.compile(this.value, flags);
+            } catch (PatternSyntaxException e) {
+                RangeIndex.LOG.error(e);
```
adamretter: Probably better to have `RangeIndex.LOG` as private and declare your own logger for this class.

adamretter:
@olvidalo Thanks for your contributions. I had a couple small comments for future reference, but otherwise this looks nice and clean. Would love to see some documentation added to https://github.com/eXist-db/documentation/ ;-)

PR #1254, opened 2017-02-09
[ignore] Increment build number for final 3.0

wolfgangmm:


adamretter:
We *must* have a public RC2 before any final release. Please revert this

wolfgangmm:
Strictly speaking RC2 was available since Aug 03, 2016. People had 2 years to test since RC1. Current code is running stable according to all who had it in production. We just cannot delay this any further or we'll wait another 6 months. The time window is closing.

adamretter:
Where was RC2 to download?

joewiz:
https://bintray.com/existdb/releases/exist/3.0.RC2

joewiz:
There was some related discussion on exist-open: http://markmail.org/message/jhdfsix2hr6sf3n7.

adamretter:
This is very disappointing! I know there is a 3.0.RC2 on BinTray but we never told anyone about it because we all know it was a broken build, which is exactly why the exist-db.org homepage lists 3.0.RC1 as the latest and the Maven repo only contains 3.0.RC1.

We previously (many months agreed) to an email that I sent around saying that we would do a release, and that would be a proper RC2 with public involvement, during which there would be a period of two weeks for critical bug fixes (if any new ones were reported) after which we would do a final release. This did not transpire months ago, because I was told that @wolfgangmm had bugs that he had to fix for customers, which I accepted in good faith.

I am however upset that we have not done a public RC2 as we all agreed and then a release; I expect to receive a lot more issues and have to do a 3.0.1 imminently now...

wolfgangmm:
@Adam I'm sincerely sorry you're upset. This was not our intention. I just felt a massive demand by users to finally have something to safely download and use, and past weeks of testing showed we reached a stable state, at least according to all people we talked to.

StephanMa:
As i can say... in our project we were not able to use eXist 3.0 until some of the very important features are working (see #967 )

adamretter:
@wolfgangmm I would probably be less upset if *anyone* had communicated with me about what was happening

PR #1256, opened 2017-02-10
Variable names that have no namespace prefix are in no namespace.

shabanovd:
See http://markmail.org/thread/mke5b5p7yflib5cq

### DIFF
--- a/src/org/exist/xquery/parser/XQueryTree.g
+++ b/src/org/exist/xquery/parser/XQueryTree.g
@@ -470,7 +470,7 @@ throws PermissionDeniedException, EXistException, XPathException
 			{
 				PathExpr enclosed= new PathExpr(context);
 				SequenceType type= null;
-				QName qn = QName.parse(staticContext, qname.getText());
+				QName qn = QName.parse(staticContext, qname.getText(), null);
 				if (declaredGlobalVars.contains(qn))
 					throw new XPathException(qname, "err:XQST0049: It is a " +
 						"static error if more than one variable declared or " +
--- a/src/org/exist/xquery/parser/XQueryTreeParser.java
+++ b/src/org/exist/xquery/parser/XQueryTreeParser.java
@@ -4360,7 +4360,7 @@ public XQueryTreeParser() {
 				
 								PathExpr enclosed= new PathExpr(context);
 								SequenceType type= null;
-								QName qn = QName.parse(staticContext, qname.getText());
+								QName qn = QName.parse(staticContext, qname.getText(), null);
 								if (declaredGlobalVars.contains(qn))
 									throw new XPathException(qname, "err:XQST0049: It is a " +
 										"static error if more than one variable declared or " +
--- a/test/src/org/exist/xquery/functions/xquery3/DeclareVariableTest.java
+++ b/test/src/org/exist/xquery/functions/xquery3/DeclareVariableTest.java
@@ -1,0 +1,47 @@
+/*
+ *  eXist Open Source Native XML Database
+ *  Copyright (C) 2001-2017 The eXist Project
+ *  http://exist-db.org
+ *
+ *  This program is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public License
+ *  as published by the Free Software Foundation; either version 2
+ *  of the License, or (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.functions.xquery3;
+
+import static org.junit.Assert.assertEquals;
+
+import org.exist.test.EmbeddedExistTester;
+import org.junit.Test;
+import org.xmldb.api.base.ResourceSet;
+import org.xmldb.api.base.XMLDBException;
+
+public class DeclareVariableTest extends EmbeddedExistTester {
+
+    @Test
+    public void defaultNamespaceTest() throws XMLDBException {
+        final String query = "xquery version \"3.1\";\n"
+            + "\n"
+            + "    declare default element namespace \"http://www.w3.org/1999/xhtml\";\n"
+            + "\n"
+            + "    declare variable $docName := 'test.xml';\n"
+            + "    declare variable $docPath := concat('/db/', $docName);\n"
+            + "\n"
+            + "    $docPath";
+
+        final ResourceSet results = executeQuery(query);
+        final String r = (String) results.getResource(0).getContent();
+
+        assertEquals("/db/test.xml", r);
+    }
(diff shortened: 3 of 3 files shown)


joewiz:
Is there a way to test for the behavior this fixes so we can prevent this regression from recurring?

dizzzz:
Good point; we need to have a test (from exist-open?)

dizzzz:
@shabanovd build fails :?

martindholmes:
This simple file catches it:

~~~~~
    xquery version "3.1";

    declare default element namespace "http://www.w3.org/1999/xhtml";

    declare variable $docName := 'ABCH1.htm';
    declare variable $docPath := concat('/db/apps/moeml/data/', $docName);

    $docPath (: should return '/db/apps/moeml/data/ABCH1.htm' :)
~~~~~

dizzzz:
```
    [junit] Running org.exist.xquery.functions.util.ExpandTest
    [junit] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 3.621 sec
BUILD FAILED
/home/travis/build/eXist-db/exist/build/scripts/junit.xml:124: Test org.exist.xquery.functions.util.ExpandTest failed
```

shabanovd:
now that test is failing:
```let $doc-path := xmldb:store('/db', 'test.xml', <concept/>)

let $doc := doc($doc-path)
return
<ok xmlns='some'>
{util:expand($doc)}
</ok>```

because `$doc` have default NS `some`

Is it correct or `{` must reset default NS to "empty" one?

**Comment by martindholmes** [comment]
As far as I can see, the namespace on a container XML element should not affect the namespace of a variable that's reference in a block of XQuery inside it.

**Comment by dizzzz** [comment]
what has changed that the test runs OK now?

**Comment by martindholmes** [comment]
I tested a build with this fix and a) the fix works and the original problem is gone, and b) no new issues are apparent to me testing a couple of my regular apps.

**Comment by adamretter** [comment]
@dizzzz Thanks for adding me as a reviewer, but I am still waiting for a decision on 3.1 before any merges

**Comment by dizzzz** [comment]
@adamretter we have been talking about a 3.0.x release, not a 3.1.x release;  as @joewiz said, a quick 3.0.x release looks good, a quick 3.1.x release ......looks not that good....

**Comment by adamretter** [comment]
@dizzzz Forget the version number, I obviously still need to work on explaining why a 3.0.x won't work. Regardless, I still haven't reviewed or merge because the issue of the next release is undecided.

**Comment by dizzzz** [comment]
@adamretter you don't need to re-explain ; your message is clear ; but without having things escalated, we seem not to want that **for now**. Not for the quick re-release we like to have. for now, 3.1.x is confusing from an enduser perspective.

**Comment by adamretter** [comment]
@dizzzz who is "we"? I have only heard from "you" ;-)

**Comment by dizzzz** [comment]
this thread is not the good location to discuss this subject

**Comment by shabanovd** [comment]
@adamretter can you use "right" thread for discussion? I'm going to delete your comments here.

**Comment by dizzzz** [comment]
we have code, we have a test.....

File: src/org/exist/xquery/parser/XQueryTree.g [hunk 1]
```diff
@@ -470,7 +470,7 @@ throws PermissionDeniedException, EXistException, XPathException
 			{
 				PathExpr enclosed= new PathExpr(context);
 				SequenceType type= null;
-				QName qn = QName.parse(staticContext, qname.getText());
+				QName qn = QName.parse(staticContext, qname.getText(), null);
```
adamretter: Instead of `null` probably better to show your intention and use `XMLConstants.NULL_NS_URI`

src/org/exist/xquery/parser/XQueryTreeParser.java
```diff
@@ -4360,7 +4360,7 @@ public final void prolog(AST _t,
 				
 								PathExpr enclosed= new PathExpr(context);
 								SequenceType type= null;
-								QName qn = QName.parse(staticContext, qname.getText());
+								QName qn = QName.parse(staticContext, qname.getText(), null);
```
adamretter: Same as previous comment

test/src/org/exist/xquery/functions/xquery3/DeclareVariableTest.java
```diff
@@ -0,0 +1,47 @@
...
+package org.exist.xquery.functions.xquery3;
+
+import static org.junit.Assert.assertEquals;
+
+import org.exist.test.EmbeddedExistTester;
+import org.junit.Test;
+import org.xmldb.api.base.ResourceSet;
+import org.xmldb.api.base.XMLDBException;
+
+public class DeclareVariableTest extends EmbeddedExistTester {
```
adamretter: Please don't use `EmbeddedExistTester` it will be removed here https://github.com/eXist-db/exist/pull/1251, please instead use the appropriate `ClassRule`

Also, I am wondering does this need to be added to `junit.xml` so it is executed?

src/org/exist/xquery/parser/XQueryTree.g
shabanovd: it's same as [other](https://github.com/shabanovd/exist/blob/5e93f51c0d0a0d699179ddac22e0712500b56517/src/org/exist/xquery/parser/XQueryTree.g#L1476) [places](https://github.com/shabanovd/exist/blob/5e93f51c0d0a0d699179ddac22e0712500b56517/src/org/exist/xquery/parser/XQueryTree.g#L1967) and [expected](https://github.com/eXist-db/exist/blob/005045fda2ee0df0c3871b12d1b90a79f086ffbc/src/org/exist/dom/QName.java#L330)

test/src/org/exist/xquery/functions/xquery3/DeclareVariableTest.java
shabanovd: That PR was not get in yet, so let's keep this test as it and you are welcome to update later if you need

adamretter: @shabanovd That one is in the queue before this one and also contains critical bug fixes

shabanovd: I don't have more time for this PR. It have fix and test. Rest is your wishes.

dizzzz:
:-( an unfortunate side effect of the recent discussions :-( 

re: `ClassRule` are there simple examples of this in the codebase?

adamretter:
@dizzzz Yes, all eXist Unit Tests that need to start or stop the database and/or a Jetty server now use ClassRules. If you search in `$EXIST_HOME/test/src/org/exist` for `@ClassRule` or `@Rule` you will find many many examples.

`@ClassRule` is for where you want one instance for the whole class of tests, similar to `@BeforeClass`.
`@Rule` is for where you want on instance for each test in the class, similar to `@Before`.

You can still use `@BeforeClass` and `@Before` if you need test class / test specific initialisation, and you will find plenty of examples of that.

The difference now is that we only have one or two functions in the test suite for starting and stopping the database and/or jetty server. This means we have removed a massive amount of duplicated code that was sometimes inconsistent. There is now one correct way to start/stop eXist and/or Jetty for your tests.

adamretter:
Just to be clear. These ClassRules have been in the codebase for a few months now and I have been migrating the tests over part by part. This PR finally moves the last of the tests over to it to complete that work.

dizzzz:
thnx!

joewiz:
Thanks @shabanovd and @adamretter, for finding a way to make this work and get it integrated with the other related changes slated for the 3.1 release.

PR #1262, opened 2017-02-16
[fix?] Lucene index – make query-field only depend on the context sequence

olvidalo:
I've noticed that ft:query-field is not evaluated only once for the context sequence, but once for each item in the context sequence. For example, given an index named "testField" on the qname `test` with a collection with test1.xml and test2.xml both containing the following data:
 
```xml
<test>
  <p>Rüsselsheim</p>
  <p>Russelsheim</p>
  <p>Māori</p>
  <p>Maori</p>
</test>
````

the query `/test[ft:query-field("testField", "Rüsselsheim", <options/>)]` will execute ft:query-field two times instead of only once. While this still leads to the correct results, the performance is not as good as it could be for larger context sequences. I've included an xquery test demonstrating this.

This code fixes the issue by making the function only depend on the context sequence. Please be aware that I don't think that I have completely understood the dependency stuff yet, but I am pretty certain that query-field will only ever depend on the context sequence, never on the context item. If this assumption is correct, the changes should be safe to merge. At least all lucene index tests still pass (at least those that are passing on the current develop – there are some failing xquery tests in there, maybe someone should have a look at those, too..).

### DIFF
--- a/extensions/indexes/lucene/src/org/exist/xquery/modules/lucene/QueryField.java
+++ b/extensions/indexes/lucene/src/org/exist/xquery/modules/lucene/QueryField.java
@@ -89,6 +89,8 @@ public class QueryField extends Query implements Optimizable {
     /* (non-Javadoc)
     * @see org.exist.xquery.PathExpr#analyze(org.exist.xquery.Expression)
     */
+
+    @Override
     public void analyze(AnalyzeContextInfo contextInfo) throws XPathException {
         super.analyze(new AnalyzeContextInfo(contextInfo));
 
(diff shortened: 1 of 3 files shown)


adamretter:
@wolfgangmm I think this looks good, if you could comment then we could get it merged before tomorrow maybe

adamretter:
@wolgangmm thank you :-)

olvidalo:
Thanks! :)

joewiz:
Looks like a potentially big performance improvement. Thanks for spotting this @olvidalo and for the fix & tests.

PR #1264, opened 2017-02-16
make sure to unlock xml resource

shabanovd:
if url is xml resource then it get read-locked


shabanovd:
@adamretter how your commit differ from my PR? and it wasn't review!!!

Issue #1199, opened 2016-12-13
sql:execute escapes some characters twice


opax:
If there is a text in an SQL database that contains `&`, `<` or `>`, the result of a `SELECT` returned by `sql:execute` incorrectly escapes those characters twice:

```xml
<sql:result count="1">
<sql:row index="1">
<sql:field name="EXAMPLE" sql:type="VARCHAR2" xs:type="xs:string"
>&amp;lt; &amp;gt; &amp;amp; </sql:field>
</sql:row>
</sql:result>
```

The problem is that the character data is escaped before feeding it to the builder that creates the XML response. The problem is here:

https://github.com/eXist-db/exist/blob/4332216c5fcb4178af926aa44a7ef650f711f2c3/extensions/modules/src/org/exist/xquery/modules/sql/ExecuteFunction.java#L287

The builder already takes care of escaping, so the call to `SQLUtils.escapeXmlText` should be removed. The same problem appears in lines 363 and 382 .

PR #1265, opened 2017-02-17
Unit test and fix for #1199

opax:
### Description:

The results created by sql:execute contain '&amp;lt;' instead of '<', the same problem exists for '>' and '&'.

### Reference:

Ticket #1199 

### Type of tests:

Java Unit Test of ExecuteFunction, the database is mocked, no further dependencies introduced

### DIFF
--- a/test/src/org/exist/xquery/modules/sql/ExecuteFunctionTest.java
+++ b/test/src/org/exist/xquery/modules/sql/ExecuteFunctionTest.java
@@ -1,0 +1,199 @@
+package org.exist.xquery.modules.sql;
+
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.mock;
+import static org.easymock.EasyMock.replay;
+import static org.junit.Assert.assertEquals;
+
+import org.exist.EXistException;
+import org.exist.dom.QName;
+import org.exist.dom.memtree.MemTreeBuilder;
+import org.exist.xquery.FunctionSignature;
+import org.exist.xquery.XPathException;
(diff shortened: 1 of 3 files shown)


test/src/org/exist/xquery/modules/sql/ExecuteFunctionTest.java
```diff
@@ -0,0 +1,199 @@
...
+import org.junit.Test;
+import org.w3c.dom.Node;
+import xquery.TestRunner;
+
+import java.sql.*;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * TODO document SQL
```
shabanovd: remove `TODO`

test/src/xquery/modules/sql/suite.xql
```diff
@@ -0,0 +1,3 @@
+xquery version "3.0";
+
+<empty/>
```
shabanovd: delete this file?

opax: That was needed because I used the TestRunner base class - I originally planned to write the test using XQUnit test. I will change it.

opax:
Thanks! By the way - this patch should probably be mentioned in the release notes whenever it will be rolled out, as it changes the behavior of sql:execute. If anyone has implemented a workaround for the buggy behavior by post-processing the results of sql:execute, that workaround must be removed.

dizzzz:
@adamretter you write this extension a long time ago, If I remember correctly

PR #1266, opened 2017-02-18
make sure to unlock xml resource

shabanovd:
if url is xml resource then it get read-locked
(similar to #1264 )

### DIFF
--- a/extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
+++ b/extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
@@ -1,22 +1,21 @@
 /*
- * eXist Open Source Native XML Database
- * Copyright (C) 2010-2015 The eXist-db Project
- * http://exist-db.org
+ *  eXist Open Source Native XML Database
+ *  Copyright (C) 2001-2017 The eXist Project
+ *  http://exist-db.org
  *
- * This program is free software; you can redistribute it and/or
- * modify it under the terms of the GNU Lesser General Public License
- * as published by the Free Software Foundation; either version 2
- * of the License, or (at your option) any later version.
+ *  This program is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public License
+ *  as published by the Free Software Foundation; either version 2
+ *  of the License, or (at your option) any later version.
  *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU Lesser General Public License for more details.
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Lesser General Public License for more details.
  *
- * You should have received a copy of the GNU Lesser General Public License
- * along with this program; if not, write to the Free Software Foundation
- * Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
- *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  */
 package org.exist.xquery.modules.expathrepo;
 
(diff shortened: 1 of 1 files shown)


extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
```diff
@@ -150,15 +149,23 @@ private URI _getURI(String s) throws XPathException
...
+        notValidXar(path);
+      }
+      return (BinaryDocument) doc;
+    } catch (PermissionDeniedException e) {
+      throw new XPathException(this, EXPathErrorCode.EXPDY003, e.getMessage(), new StringValue(path), e);
     }
+  }
+
+  private void notValidXar(String path) throws XPathException {
+    throw new XPathException(this, EXPathErrorCode.EXPDY001, path + " is not a valid .xar", new StringValue(path));
```
dizzzz: I'd suggest to have two simple throw exceptions on line 160 and 164 ; with this you could probably change the text into "Resource not found" and "Resource is not binary" .... which helps in finding a root cause.

shabanovd: done

extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
```diff
@@ -150,15 +149,25 @@ private URI _getURI(String s) throws XPathException
...
-		} catch (PermissionDeniedException e) {
-			throw new XPathException(this, EXPathErrorCode.EXPDY003, e.getMessage(), new StringValue(path), e);
-		}
+  private BinaryDocument _getDocument(String path) throws XPathException {
+    try {
+      XmldbURI uri = XmldbURI.createInternal(path);
+      DocumentImpl doc = context.getBroker().getXMLResource(uri, LockMode.READ_LOCK);
+      if (doc == null) {
+        throw new XPathException(this, EXPathErrorCode.EXPDY001,
+            path + " is no .xar resource",
```
adamretter: "is no .xar resource" -> " does not exist"

extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
```diff
@@ -150,15 +149,25 @@ private URI _getURI(String s) throws XPathException
...
+            new StringValue(path)
+        );
+      } else if (doc.getResourceType() != DocumentImpl.BINARY_FILE) {
+        doc.getUpdateLock().release(LockMode.READ_LOCK);
+        throw new XPathException(this, EXPathErrorCode.EXPDY001,
+            path + " is not a valid .xar, it's not a binary resource",
```
adamretter: " is not a valid .xar, it's not a binary resource" -> " is not a valid .xar EXPath Package"

adamretter:
@shabanovd Could we have  a test please?

extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
dizzzz: the test is actually only if it is a binary resource.....

adamretter: @dizzzz Yes I realise that

extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
shabanovd: disagree, because your message unclear that problem around `.xar`

extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
shabanovd: so, message must include information that it fail because resource is not `binary`

shabanovd:
It's unclear for me how to write test for it. @adamretter, please do, if you know how to.

dizzzz:
I have no idea either

adamretter:
@shabanovd @dizzzz A test just needs to call InstallFunction#eval and afterwards check whether the lock is still held or not. Mocks will help!

dizzzz:
speaking for myself, I'd still have no idea :-(

joewiz:
@adamretter @wolfgangmm Any outstanding concerns here?

adamretter:
@joewiz Yeah it is awaiting time for me to add a test as @shabanovd didn't want to do it.

extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
adamretter: If you insist, can you fix the language so that it reads "is not a .xar resource"?

extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
adamretter: The spec for EXPath packages says that they must be valid Zip files. Therefore a valid Xar can only be a binary file. As such I think that means the error can be simplified and improved here

shabanovd: The xar file can be invalid for many reasons: corrupted zip, missing file at zip, wrong folders structure. This error define clearly that eXist do not see that resource as binary one.

extensions/modules/src/org/exist/xquery/modules/expathrepo/InstallFunction.java
shabanovd: done

PR #1267, opened 2017-02-18
make sure to unlock resource

shabanovd:
resource was not unlocked at all

### DIFF
--- a/src/org/exist/http/AuditTrailSessionListener.java
+++ b/src/org/exist/http/AuditTrailSessionListener.java
@@ -1,6 +1,6 @@
 /*
  *  eXist Open Source Native XML Database
- *  Copyright (C) 2010 The eXist Project
+ *  Copyright (C) 2001-2017 The eXist Project
  *  http://exist-db.org
  *
  *  This program is free software; you can redistribute it and/or
@@ -13,11 +13,9 @@
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  *  GNU Lesser General Public License for more details.
  *
- *  You should have received a copy of the GNU Lesser General Public License
- *  along with this program; if not, write to the Free Software
- *  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
- *
- * $Id$
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  */
 package org.exist.http;
 
@@ -160,6 +158,10 @@ public class AuditTrailSessionListener implements HttpSessionListener {
                         LOG.info("XQuery execution results: " + result.toString() + " in " + queryTime + "ms.");
                     } finally {
                         xqpool.returnCompiledXQuery(source, compiled);
+                    }
+                } finally {
+                    if (resource != null) {
+                        resource.getUpdateLock().release(LockMode.READ_LOCK);
                     }
                 }
 


adamretter:
@shabanovd Please can you add a test case

shabanovd:
It's unclear for me how to write test for it. @adamretter, please do, if you know how to.

dizzzz:
I'd have no idea....

adamretter:
@shabanovd @dizzzz This one is nice and easy, just call either `AuditTrailSessionListener#sessionCreated` or `AuditTrailSessionListener#sessionDestroyed` and afterwards check the lock state.

dizzzz:
ok thnx for providing the hints. that helps.

shabanovd:
To be clear my opinion that the owner of the feature must carry out tests, not me - contributor of bugfix for that feature. It related to other (my) PRs.

adamretter:
@shabanovd It has always been the responsibility of those submitting code to also submit tests. It is even in our docs and have been for a long time, see https://github.com/eXist-db/exist/blob/develop/CONTRIBUTING.md

> * **Test**. If you fix something prove it, write a test that illustrates the issue before you fix the issue and validate the test. If you add a new feature it needs tests, so that we can understand its intent and try to avoid regressions in future as much as possible.

dizzzz:
@adamretter you are basically right, but such an approach might and will withhold contributors to file any fixes for obvious but hard to test bugs ; we need to try to take it a bit pragmatic side...

adamretter:
@dizzzz Sure, but this is easy to test.

Not wanting to write tests or learn how to write tests, is not an excuse for not doing so. I am always happy to teach how to write tests if someone needs help.

I would probably be more relaxed about the odd PR without tests, but it is the case that I frequently have to ask for tests in PRs. If there are not tests for code that adds risk and unmaintability issues; As it is we have accumulated far too much of this technical debt in the past. Time to draw a line.

If we are not in agreement, we should also discuss this in the upcoming community telco?

dizzzz:
we are in agreement for sure! But sometimes a combi of code-inspection and the current test suite might be enough.

adamretter:
@dizzzz Agreed. In this case the current test suite doesn't test this. Hence my call for tests :-)

shabanovd:
@adamretter if you want test - write one. my contribution ends on point to bug place. you should not expect more from me here.

adamretter:
@shabanovd So we should merge your code without tests, because you are special?
This is not the first time you have abandoned PRs because you refuse to finish them by writing tests. Personally I don't think your behaviour is really in the spirit of community collaboration (which is a documented process). That is just my opinion though. AFAIK no one else sends PRs and then expects *you* to write the tests for them.

More than likely, I will write the tests again, because you won't *sighhh*. It's not like I don't have other PRs to work on!!!

dizzzz:
Gents......

adamretter:
@dizzzz @shabanovd For reference here is how you test such code - https://github.com/eXist-db/exist/blob/develop/test/src/org/exist/http/AuditTrailSessionListenerTest.java

I have now merged this along with its test. The test only took me 10 minutes to write :-)

shabanovd:
@adamretter can you do not send commits to develop branch without reviews?

I don't think that your test cover current PR anyhow.

adamretter:
@dizzzz Why do you think the test does not cover your changes. If you undo your change https://github.com/eXist-db/exist/pull/1267/files#diff-6fe9b581946cde755de400a26ab9400eR164 then the test will fail.

dizzzz:
@adamretter I did not write anything ?

adamretter:
@dizzzz Sorry. I meant Dmitriy (edited above)

PR #1271, opened 2017-02-20
make sure that functionDef do not shared between DeferredFunctionCallImpl

### LINKED ISSUE #758: Error in nested fn:fold-left call
The following code returns the wrong result:

``` xquery
xquery version "3.1";

declare function local:join($value){
    let $args := 
        array:for-each($value("args"),function($_){
            if($_ instance of map(xs:string, item()?)) then
                local:join($_)
            else
                $_
        })
    let $args2 :=
        array:fold-left($args,[],function($pre,$cur){
            array:append($pre,$cur)
        })
    return
        string-join(array:flatten($args2),",")
};

let $map := map { 
    "args" := [
        "'a'",
        map {
            "args":=["'x'","'y'"]
        },
        "'b'"]
}
return local:join($map)
```

The result should read `'a','x','y','b'` but instead it's `'a','x','x','b'`

shabanovd:
close #758

### DIFF
--- a/src/org/exist/xquery/FunctionCall.java
+++ b/src/org/exist/xquery/FunctionCall.java
@@ -280,7 +280,7 @@ public class FunctionCall extends Function {
         
         if(isRecursive()) {
             //LOG.warn("Tail recursive function: " + functionDef.getSignature().toString());
-            return new DeferredFunctionCallImpl(functionDef.getSignature(), contextSequence, contextItem, seq, contextDocs);
+            return new DeferredFunctionCallImpl(this, contextSequence, contextItem, seq, contextDocs);
         } else {
             
             //XXX: should we have it? org.exist.xquery.UserDefinedFunction do a call -shabanovd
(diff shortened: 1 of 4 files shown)


src/org/exist/xquery/FunctionCall.java
```diff
@@ -362,15 +362,20 @@ public void accept(ExpressionVisitor visitor) {
         visitor.visitFunctionCall(this);
     }
 
-    private class DeferredFunctionCallImpl extends DeferredFunctionCall {
+    private static class DeferredFunctionCallImpl extends DeferredFunctionCall {
```
dizzzz: are you sure this inner class should be static?

shabanovd: yeap, it was design bug to not make it static

adamretter: @dizzzz Inner classes should always be `static` unless they explicitly and directly and need access to the outer classes variable members.

However, that said, the approach @shabanovd is taking looks wrong here to me. He explicitly passed the outer class because he needs access to the members. I think it is better to either:

1) Not make the inner class static.
2) Make the inner class static, but do not access the member variables directly, instead call the accessor functions, then these could be mocked for testing. i.e. instead of `call.mySignature` use `call.getSignature()`

adamretter: My preference would like be for (1) as that means less change and that was working before.

shabanovd: @adamretter you did not understand the problem and the solution I provide. Please, spend a bit of time to thinking before write proposal. [Point by finger](https://github.com/eXist-db/exist/pull/1271/files/ee056ce36d5885ba5effde23a9bccd1cc5b67854#diff-f2b23bccdd5042ec21691c3e7464d672R443)

That is `private static`, so I do not any issue to access `parent` class variable directly.

adamretter:
@shabanovd @wolfgangmm Okay, sounds good, I would suggest that we move the inner class out. So that it is a standalone class in its own right.

src/org/exist/xquery/FunctionCall.java
adamretter: I see, still looks weird to me, see my comment below. I think it best to make this a non-inner class now.

shabanovd:
@adamretter that is only place where this class is used, so I prefer to keep code together.

adamretter:
Travis is reporting test failures though - could you take a look please?

shabanovd:
@adamretter how to rerun travis?

adamretter:
@shabanovd If you visit the Travis build - https://travis-ci.org/eXist-db/exist/builds/203607846
You should see a button near the top left which says "Restart Build"

shabanovd:
@adamretter [this](https://github.com/eXist-db/exist/blob/005045fda2ee0df0c3871b12d1b90a79f086ffbc/test/src/org/exist/xquery/DeferredFunctionCallTest.java#L110) is failing. It's much harder to read than actual code and I have no idea how to fix it.

shabanovd:
@adamretter fixed

PR #1273, opened 2017-02-21
where expresion must not be evaluated for empty sequence

### LINKED ISSUE #1270: Dynamic function call in a 'where' filter of 'for' on an empty sequence is evaluated
In the following example, the filter function gets evaluated even though the sequence of the 'for' is empty (`empty($words) => true()`)

``` xquery
xquery version "3.1";

declare function local:test($x as xs:string) as xs:boolean {
    true()
};
let $f := local:test#1
let $words := analyze-string('', '\w+')/fn:match/text()
return for $w in $words where $f($w) return $w
```

resulting in the error
```
exerr:ERROR XPTY0004: The actual cardinality for parameter 1 does not match the cardinality declared in the function's signature: local:text($x as xs:string) xs:boolean. Expected cardinality: exactly one, got 0.
```

If the 'where' clause is changed to avoid the dynamic function call (`where local:test($w)`), the filter function is not called and no error is reported.

Tested on eXistdb v3.0.

shabanovd:
close #1270

### DIFF
--- a/test/src/xquery/xquery3/forwhere.xml
+++ b/test/src/xquery/xquery3/forwhere.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<TestSet>
+    <testName>xquery3 for-where</testName>
+    <description>
+        <p>For-where in XQuery 3</p>
+        <author>Dmitiry Shabanov</author>
+        <author>https://github.com/ptrckb</author>
+    </description>
+    <setup>
+        <create-collection parent="/db" name="xq3-test"/>
+    </setup>
+    <tearDown>
+        <remove-collection collection="/db/xq3-test"/>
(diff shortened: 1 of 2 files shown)


test/src/xquery/xquery3/forwhere.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<TestSet>
+    <testName>xquery3 for-where</testName>
+    <description>
+        <p>For-where in XQuery 3</p>
+        <author>Dmitiry Shabanov</author>
+        <author>https://github.com/ptrckb</author>
```
adamretter: Interesting. Is that link correct?

adamretter:
Looks good :-)

PR #1276, opened 2017-02-22
up version to 3.1.0

shabanovd:
and remove `project.version.numeric` property

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -2,10 +2,8 @@
 # Don't directly modify this file. Instead, copy it to local.build.properties and
 # edit that.
 #
-# $Id$
 project.name = eXist-db
-project.version = 3.0
-project.version.numeric = 3.0.4
+project.version = 3.1.0
 project.source = develop
 project.codename = Berlin
 
--- a/build.properties
+++ b/build.properties
@@ -4,7 +4,7 @@
 #
 project.name = eXist-db
 project.version.numeric = 3.1.0
-project.version.desc = RC1
+project.version.stage = RC1
 project.source = develop
 project.codename = Berlin
 
--- a/build.properties
+++ b/build.properties
@@ -3,7 +3,8 @@
 # edit that.
 #
 project.name = eXist-db
-project.version = 3.1.0
+project.version.numeric = 3.1.0
+project.version.desc = RC1
 project.source = develop
 project.codename = Berlin
 
--- a/build/scripts/build-impl.xml
+++ b/build/scripts/build-impl.xml
@@ -276,15 +276,15 @@
     <target name="jar" depends="git.details,compile,compile-aspectj"
         description="Create eXist-db unsigned jar files">
         <filter token="version" value="${project.version}"/>
-	<filter token="semver" value="${project.version.numeric}"/>
-	<filter token="source" value="${project.source}"/>
+        <filter token="semver" value="${project.version}"/>
+        <filter token="source" value="${project.source}"/>
         <filter token="build" value="${DSTAMP}"/>
         <filter token="git.commit" value="${git.commit}"/>
 
         <!--copy file="${src}/CatalogManager.properties" todir="${build.classes}"/-->
         <copy todir="${build.classes}/org/exist/client">
             <fileset dir="${src}/org/exist/client">
-		<include name="*.tmpl"/>
+            <include name="*.tmpl"/>
                 <include name="icons/**.png"/>
                 <include name="icons/**.gif"/>
                 <include name="*.xsl"/>
--- a/build/scripts/build-impl.xml
+++ b/build/scripts/build-impl.xml
@@ -36,7 +36,7 @@
     <property name="build.deprecation" value="on"/>
     <property name="jvmarg" value="-Xmx128000k -Xms32000k -Djetty.home=${jetty.dir} -Dexist.home=${basedir}"/>
     <property name="xmldb.src" value="undefined"/>
-    <property name="dist.dir" value="${dist}/${project.name}-${project.version}"/>
+    <property name="dist.dir" value="${dist}/${project.name}-${project.version.numeric}-${project.version.desc}"/>
     
    
 
@@ -117,7 +117,7 @@
         <property file="VERSION.txt"/>
 
         <echo message="-------------------------"/>
-        <echo message="eXist ${project.version} build"/>
+        <echo message="eXist ${project.version.numeric}-${project.version.desc} build"/>
         <echo message="-------------------------"/>
         <echo message="Git commit ID: ${git.commit}"/>
         <echo message=""/>
@@ -130,7 +130,7 @@
         <mkdir dir="${jetty.dir}/logs"/>
         <mkdir dir="${jetty.dir}/work"/>
 
-        <filter token="version" value="${project.version}-:${git.commit}-${DSTAMP}"/>
+        <filter token="version" value="${project.version.numeric}-:${git.commit}-${DSTAMP}"/>
         <filter token="database" value="native"/>
 
         <copy file="${src.webapp}/xqts/config.xml.tmpl" tofile="${src.webapp}/xqts/config.xml"
@@ -275,8 +275,8 @@
     <!-- ================================================================== -->
     <target name="jar" depends="git.details,compile,compile-aspectj"
         description="Create eXist-db unsigned jar files">
-        <filter token="version" value="${project.version}"/>
-        <filter token="semver" value="${project.version}"/>
+        <filter token="version" value="${project.version.numeric}-${project.version.desc}"/>
+        <filter token="semver" value="${project.version.numeric}"/>
         <filter token="source" value="${project.source}"/>
         <filter token="build" value="${DSTAMP}"/>
         <filter token="git.commit" value="${git.commit}"/>
@@ -345,7 +345,7 @@
         <jar basedir="${build.classes}" compress="true" jarfile="exist.jar">
             <manifest>
                 <attribute name="Project-Name" value="${project.name}"/>
-                <attribute name="Project-Version" value="${project.version}"/>
+                <attribute name="Project-Version" value="${project.version.numeric}"/>
                 <attribute name="Project-Build" value="${DSTAMP}"/>
                 <attribute name="Project-CodeName" value="${project.codename}"/>
                 <attribute name="Git-commit" value="${git.commit}"/>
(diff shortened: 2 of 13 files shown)


dizzzz:
this will not work ! we need to have this x.y.x schema, expath will fail all over the place.....

build.properties
```diff
@@ -2,10 +2,8 @@
 # Don't directly modify this file. Instead, copy it to local.build.properties and
 # edit that.
 #
-# $Id$
 project.name = eXist-db
-project.version = 3.0
-project.version.numeric = 3.0.4
+project.version = 3.1.0
 project.source = develop
 project.codename = Berlin
```
dizzzz: for @ljo let's make this "Göteborg"  ?

build.properties
```diff
@@ -2,10 +2,8 @@
 # Don't directly modify this file. Instead, copy it to local.build.properties and
 # edit that.
 #
-# $Id$
 project.name = eXist-db
-project.version = 3.0
```
dizzzz: I think this short number is used in the installer etc?

dizzzz:
@ljo's remark about RCx / alpha is quite valid....

build.properties
ljo: Yes, there is already semantics in these keys also by ant.

shabanovd: I remove `project.version.numeric` because I did not found any references to it. If external application depend on that property then I can change it.

shabanovd:
@ljo @dizzzz we agree yesterday to follow http://semver.org/ rules, so no `RCx / alpha` at version

build.properties
ljo: @shabanovd Yes, thanks, like I wrote this is used for all package related  sem-ver checks both in package manager (dashboard) and public repo apps.

wolfgangmm: For compatibility checks, the package manager relies on the product-semver property from system.properties, which is set to project.version.numeric by the build scripts. So either keep project.version.numeric or change the build scripts to use project.version (which then must be 3 digit).

ljo:
@shabanovd well, since semver allows prerelease labels and build metadata I don't think it changes any of my concerns.

dizzzz:
I think now things get mixed up.....

project.version is actually a free-format text,  project.version.numeric is used on many places as semver....

the latter can really not be removed... and we need to be able to indicate RCs right? we only agreed that for a RC not a new sem.ver would be assigned.

build.properties
shabanovd: `build` folder was `excluded` and that's why my search for `project.version.numeric` show nothing. I did commit updates to build instructions.

ljo:
@shabanovd I would have preferred you keep project.version.numeric since it is semantically unambigous. Numeric only means digits.

shabanovd:
@dizzzz we agree to have single version. Also, we agree to make RC abstract, so it just note at release candidate text.

```
project.version = 3.1.0
project.version.numeric = 3.1.0
```
That will confuse for sure.

build.properties
wolfgangmm: well, as @dizzzz wrote, if we use project.version instead of project.version.numeric, we won't have a way to indicate something like "3.2.0RC" because the semver check does a > or < comparison. So this would mean to either introduce a new property or leave the old two properties, sorry :-(

wolfgangmm:
Ok, if "RC" is just a note somewhere outside the properties, I'm fine. It should just be clear that project.version *needs to be numeric*.

ljo:
@shabanovd But why remove the only one key containing only digits. If that is what you want too and keep the one which has all possibilities to confuse by its previous use with allowing alphanumerics?

shabanovd:
@ljo `project.version.numeric` vs `project.version`: Define versioning policy by `.numeric` a bit strange for me. It must be declared in the documentation.

Also,  `project.version` commonly used naming: maven, gradle and so on.

ljo:
@shabanovd  Ant there you for sure have prerelease labels **3.1-SNAPSHOT** etcetera.

build/scripts/build-impl.xml
```diff
@@ -276,15 +276,15 @@
     <target name="jar" depends="git.details,compile,compile-aspectj"
         description="Create eXist-db unsigned jar files">
         <filter token="version" value="${project.version}"/>
```
shabanovd: should we keep both `version` and `semver`?

ljo: Not unless RC, SNAPSHOT etc should be part of installer filename and so on like now

shabanovd:
@ljo @dizzzz @wolfgangmm so, we end up with two versions or I miss something?

ljo:
@shabanovd Why couldn't you just change project.version.numeric= 3.0.4 to 3.1.0. Now we risk all sorts of breakage.

shabanovd:
@ljo the question was to get to single version (and property)

shabanovd:
Ok, I think we should change name of `project.version` to avoid confusion. How about `project.version.desc=3.1.0_RC1`?

ljo:
@shabanovd We decided to have a single version. One Property is just invented today. I only say you removed the wrong property. I.e. the one which is unambiguous and is actually used for the semver calculations already today. The other one is only for show. I can live with project.version.desc if you keep the .numeric.

dizzzz:
or project.version.desc=RC1 and have the semver put in front of it ;
we need to check how things end up in the JAR files btw... (and the installer code)

shabanovd:
@ljo `well, since semver allows prerelease labels and build metadata I don't think it changes any of my concerns.` - you are no right here: `A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes. X is the major version, Y is the minor version, and Z is the patch version. Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.`

ljo:
@shabanovd No, you read too little of the definition.

shabanovd:
@ljo can you explain?

shabanovd:
@ljo 
`
A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92.`
do you mean this?

ljo:
Yes, items 9 and 10 of the definition. It makes it contain alphanumerics not only digits.

build.properties
adamretter: I think we could just delete the `codename` as it is unused

build.properties
adamretter: @wolfgangmm No! We discussed this yesterday in the teleconf.
We agreed to one version number (e.g. `project.version`) number which is semver. Build scripts can be updated to use that.

Release candidates for 3.1.0 will report their version as 3.1.0. Extensions don't need to know the difference because of the short life-span of a RC. i.e. Extensions are never released against an RC they are only tested against RCs and releases of Extensions are against final eXist versions.

adamretter:
@wolfgangmm @shabanovd @dizzzz @ljo @joewiz Everyone stop!

We made agreements yesterday (in the teleconf) and yet the above seems to show a whole load of confusion and different opinions again. @shabanovd was initially correct above in what he was saying, his comments were inline with what was agreed yesterday.

As there is still so much confusion. I would suggest that myself and @shabanovd should write up the `Versioning and Release` document that was suggested by @joewiz and that should be circulated first. Then everyone knows what is happening. We can then revisit bumping version numbers...

ljo:
@adamretter No, you misread something -- there are no different opinions. We made an agreement.

shabanovd:
@adamretter don't panic! -)

@ljo @dizzzz @wolfgangmm I return back `project.version.numeric` and add `project.version.desc` by 204af22

ljo:
@shabanovd Thanks :).

shabanovd:
summary:
 - `${project.version.numeric}` is pure eXist version;
 - distribute file name: `${project.name}-*-${project.version.numeric}-${project.version.desc}-${git.commit}.*`

comments?

ljo:
@shabanovd I am happy with your summary.

adamretter:
@ljo @shabanovd This is not what we agreed yesterday. Where has this RC descriptor come from? I think we have to write up the document first still

shabanovd:
my own perfectionist rebels: use `${project.version.stage}` instead of `${project.version.desc}` because it going to be only `RC1`, `RC2`, ..., `FINAL`

shabanovd:
@adamretter do you not agree that distribution files must have development stage in it name?

adamretter:
I would ask people to cast their minds back to the telco yesterday where I tried to explain:

1.  That an RC is just a snapshot (git tag or whatever) (of what will hopefully become the final version).

2. Filenames can be named RC1 or whatever for marketing, but internally it should be identical to what we believe the final version will be.

Let's get a Versioning document written so we can agree...

ljo:
@adamretter Yes, please do that document. But it is for sure a single version number with three numbers and two dots inbetween which is exactly what we agreed on and it allows prerelease labels.

build.properties
shabanovd: @adamretter we agree to have single version, but we did not have any agreement that it going to be `project.version`.

You are right that extensions should know what development stage, but human that use eXist must know that. And that's why distribution must clear declare development stage information.

adamretter: @shabanovd When you say:

> You are right that extensions should know what development stage, but human that use eXist must know that.

Did you actually mean:
> You are right that extensions should *not* know what development stage, but human that use eXist must know that.

shabanovd:
@adamretter What of your items this PR break?

build.properties
shabanovd: Project must have traditions, don't kill it -)

build.properties
shabanovd: @adamretter correct, fixed

adamretter:
I would encourage everyone to expend effort on getting the list of outstanding PRs reviewed, fixed and merged. I think we can revisit this PR once the Versioning and Release doc is agreed. I will try and draft a version tonight and share with @shabanovd to edit first before I circulate

joewiz:
The notes from yesterday's call show that there was agreement about the next version number (3.1), adopting semantic versioning, and the release timeline. But as to implementing the versioning changes, the very first action item ("Next Steps") was drafting a proposal to be circulated. 

> AR+DS to Create Versioning Document

I'd encourage @adamretter and @shabanovd to please be specific about the rationale for the proposed practices, and provide a sketch of what the next few releases would look like (covering release candidates, patch/bugfix, minor, and major versions). Ideally, we'll end up with a policy that leaves no ambiguity - and we'll just be able to push forward.

ljo:
@adamretter OK. Great.

dizzzz:
I am pretty sure this property only influences the name of the installer/dmg/...

adamretter:
@dizzzz You need to see the Versioning Scheme document first (still in draft). This does not match that. So please hold fire

shabanovd:
@adamretter that do not much only your view ;-)

PR #1283, opened 2017-02-24
[feature] The Lookup Operator ("?") for Maps

shabanovd:
close #1284


wolfgangmm:
Just for clarification: this only fixes the issue of using ? as a postfix expression after a map constructor, right?

test/src/xquery/maps/mapsLookup.xql
```diff
@@ -0,0 +1,23 @@
+xquery version "1.0";
```
joewiz: version "3.0" or "3.1" is required for maps

joewiz: ... and function annotations

shabanovd:
@wolfgangmm I'm inline [map constructor](https://github.com/eXist-db/exist/pull/1283/files#diff-067a678c32cd2039acc08c3d652a36e1R1267) with array constructor code. And add [this](https://github.com/eXist-db/exist/pull/1283/files#diff-6ca351c43af18a1bf85f0d17ce75a870R1932).

shabanovd:
@wolfgangmm also, massive change [here](https://github.com/eXist-db/exist/pull/1283/files#diff-6ca351c43af18a1bf85f0d17ce75a870R3171)

test/src/xquery/maps/mapsLookup.xql
shabanovd: @joewiz I notice it too. It's same way as other [test](https://github.com/eXist-db/exist/blob/develop/test/src/xquery/maps/maps.xql#L1)s too.

joewiz: I've submitted a PR to fix that one. Isn't it odd that there's no error raised when using 3.x features when declaring the query as 1.0? 

```xquery
xquery version "1.0";

"a" || "b"
```

raises this error:

> Cannot compile xquery: exerr:EXXQDY0003 string concatenation operator is not available before XQuery 3.0 [at line 3, column 1, source: xquery version "1.0"; "a" || "b"]

but this doesn't:

```xquery
xquery version "1.0";

map { "a": "b" }
```

shabanovd: @joewiz, you are right. It's simple to fix by that [code](https://github.com/eXist-db/exist/blob/005045fda2ee0df0c3871b12d1b90a79f086ffbc/src/org/exist/xquery/ConcatExpr.java#L14-L17). That going to be staff for next PR.

dizzzz:
jikes......

shabanovd:
@dizzzz that are autogenerated staff

test/src/xquery/maps/mapsLookup.xql
joewiz: Just confirmed that with your now-merged PR, my test query produces the proper warning. Nice job @shabanovd!

```xquery
xquery version "1.0";

map { "a": "b" }
```

> Cannot compile xquery: exerr:EXXQDY0003 Map is not available before XQuery 3.0 [source: xquery version "1.0"; map { "a": "b" }]

PR #1285, opened 2017-02-24
module-functions must throw exception if can't compile (import) module

shabanovd:
Fix XQSuite issue that tests do not show up as passed or failed one (silently ignored)

### DIFF
--- a/src/org/exist/xquery/functions/inspect/ModuleFunctions.java
+++ b/src/org/exist/xquery/functions/inspect/ModuleFunctions.java
@@ -56,27 +56,18 @@ public class ModuleFunctions extends BasicFunction {
             final XQueryContext tempContext = new XQueryContext(context.getBroker().getBrokerPool());
             tempContext.setModuleLoadPath(context.getModuleLoadPath());
 
-            Module module = null;
+            Module module;
 
-            try {
-                if (isCalledAs("module-functions-by-uri")) {
-                    module = tempContext.importModule(args[0].getStringValue(), null, null);
-                } else {
-                    module = tempContext.importModule(null, null, args[0].getStringValue());
-                }
-                
-            } catch (final XPathException e) {
-                LOG.debug("Failed to import module: " + args[0].getStringValue() + ": " + e.getMessage(), e);
-
-                throw new XPathException(this, e.getMessage(), e);
-
-            } catch (final Exception e) {
-                LOG.debug("Failed to import module: " + args[0].getStringValue() + ": " + e.getMessage(), e);
+            if (isCalledAs("module-functions-by-uri")) {
+                module = tempContext.importModule(args[0].getStringValue(), null, null);
+            } else {
+                module = tempContext.importModule(null, null, args[0].getStringValue());
             }
-            
+                
             if (module == null) {
                 return Sequence.EMPTY_SEQUENCE;
             }
+
             addFunctionRefsFromModule(tempContext, list, module);
         } else {
             addFunctionRefsFromContext(list);
--- a/src/org/exist/xquery/functions/inspect/ModuleFunctions.java
+++ b/src/org/exist/xquery/functions/inspect/ModuleFunctions.java
@@ -68,9 +68,7 @@ public class ModuleFunctions extends BasicFunction {
             } catch (final XPathException e) {
                 LOG.debug("Failed to import module: " + args[0].getStringValue() + ": " + e.getMessage(), e);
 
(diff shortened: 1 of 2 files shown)


src/org/exist/xquery/functions/inspect/ModuleFunctions.java
```diff
@@ -68,9 +68,7 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
             } catch (final XPathException e) {
                 LOG.debug("Failed to import module: " + args[0].getStringValue() + ": " + e.getMessage(), e);
 
-                if (e.getErrorCode().equals(ErrorCodes.XPST0003)) {
-                    throw new XPathException(this, e.getMessage());
-                }
+                throw new XPathException(this, e.getMessage(), e);
```
shabanovd: I don't know how correct is this change

test/src/xquery/xquery3/fn.xql
```diff
@@ -34,23 +34,28 @@ function fnt:has-children-contextItem-noChildren() {
...
 };
+:)
 
 declare
     %test:assertError("XPDY0002")
 function fnt:has-children-contextItem-absent() {
     has-children()
 };
 
+(: fail with XPTY0019
```
dizzzz: so this test fails right now?

joewiz:
@dizzzz Sure, I'll add this PR to the agenda for discussion.

test/src/xquery/xquery3/fn.xql
shabanovd: the problem that `fn.xql` silently ignored right now because of [this](https://github.com/eXist-db/exist/pull/1285/commits/726ca3407f94cf2c7bfd44959223a066bc744607#diff-07e7d69906fa424961c02b9044da8db4R96) and [that](https://github.com/eXist-db/exist/pull/1285/commits/726ca3407f94cf2c7bfd44959223a066bc744607#diff-07e7d69906fa424961c02b9044da8db4R122).

And after it run this test fail because actual error code is different.

To fix root issue I commit [this](https://github.com/eXist-db/exist/pull/1285/commits/726ca3407f94cf2c7bfd44959223a066bc744607#diff-8306b9c569c683640a7bdb63c01cfae1R71) change.

joewiz: Just curious, is there a reason you comment out the failing tests, versus using `%test:pending`? Are they static errors that prevent `fn.xql` working at all?

shabanovd: @joewiz because I had no idea that I can use  `%test:pending` -)

joewiz: Ha! Here's an example - https://github.com/eXist-db/exist/pull/1177 - a PR awaiting review in fact.

joewiz:
Cool!

joewiz:
Looks like this is good to merge and label with the 3.1.0 milestone.

shabanovd:
@dizzzz good to merge?

adamretter:
@shabanovd The two tests you have marked as `%pending` what is the plan for those?

shabanovd:
@adamretter that tests fail from your commit. Guess, you know how to fix it better then I'm.

adamretter:
@shabanovd I have no idea as the tests were working before your changes. I think we need to address this before we can merge this PR

shabanovd:
@adamretter how tests can pass if there was compilation error [here](https://github.com/eXist-db/exist/pull/1285/files#diff-07e7d69906fa424961c02b9044da8db4R123)?

adamretter:
@shabanovd I will investigate...

joewiz:
@adamretter Any updates on this? It seems this PR would be a nice pairing with my PR for eXide focused on running test suites, https://github.com/wolfgangmm/eXide/issues/156.

shabanovd:
@adamretter can that tests be fixed by different PR?

duncdrum:
@shabnovd could this be the same issue regarding silently ignored tests? https://github.com/eXist-db/exist/issues/1389

dizzzz:
@duncdrum I think it is is a different use case ..... ignored test should be ignored no? (nr of 'pending' tests ais reported)

duncdrum:
@dizzzz agreed. There should be either 0 tests and 1 pending. Or 2 tests 1 pending. But we have 1 test, 1 pending and 1 silently ignored. The last bit seemed relevant. I still don t get it, but if you all think it's ok, then that's fine.

PR #1296, opened 2017-02-26
[refactor] Moving lock ownership is dangerous

adamretter:
This simple change greatly simplifies lock release code when locks need to be associated with the transaction lifecycle.

Previously moving lock ownership to a transaction was error prone and likely to lead to complex and possibly incorrect release semantics. This small refactoring opens the door to using ARM (Automatic Resource Management) for lock leases.

### DIFF
--- a/src/org/exist/security/PermissionFactory.java
+++ b/src/org/exist/security/PermissionFactory.java
@@ -34,6 +34,7 @@ import org.exist.storage.txn.TransactionException;
 import org.exist.storage.txn.TransactionManager;
 import org.exist.storage.txn.Txn;
 import com.evolvedbinary.j8fu.function.ConsumerE;
+import org.exist.util.LockException;
 import org.exist.xmldb.XmldbURI;
 import org.exist.xquery.XPathException;
 
@@ -110,40 +111,52 @@ public class PermissionFactory {
     }
 
     public static void updatePermissions(final DBBroker broker, final XmldbURI pathUri, final ConsumerE<Permission, PermissionDeniedException> permissionModifier) throws PermissionDeniedException {
-        DocumentImpl doc = null;
         final TransactionManager transact = broker.getBrokerPool().getTransactionManager();
         try(final Txn transaction = transact.beginTransaction()) {
-            final Collection collection = broker.openCollection(pathUri, LockMode.WRITE_LOCK);
-            if (collection == null) {
-                doc = broker.getXMLResource(pathUri, LockMode.WRITE_LOCK);
-                if(doc == null) {
-                    transact.abort(transaction);
-                    throw new XPathException("Resource or collection '" + pathUri.toString() + "' does not exist.");
+            Collection collection = null;
+            try {
+                collection = broker.openCollection(pathUri, LockMode.WRITE_LOCK);
+                if (collection == null) {
+                    DocumentImpl doc = null;
+                    try {
+                        doc = broker.getXMLResource(pathUri, LockMode.WRITE_LOCK);
+                        if (doc == null) {
+                            transact.abort(transaction);
+                            throw new XPathException("Resource or collection '" + pathUri.toString() + "' does not exist.");
+                        }
+
+                        // keep a write lock in the transaction
+                        transaction.acquireLock(doc.getUpdateLock(), LockMode.WRITE_LOCK);
+
+                        final Permission permissions = doc.getPermissions();
+                        permissionModifier.accept(permissions);
+
+                        broker.storeXMLResource(transaction, doc);
+                    } finally {
+                        if(doc != null) {
+                            doc.getUpdateLock().release(LockMode.WRITE_LOCK);
+                        }
+                    }
+                    transact.commit(transaction);
+                    broker.flush();
+                } else {
+                    // keep a write lock in the transaction
+                    transaction.acquireLock(collection.getLock(), LockMode.WRITE_LOCK);
+
+                    final Permission permissions = collection.getPermissionsNoLock();
+                    permissionModifier.accept(permissions);
+
+                    broker.saveCollection(transaction, collection);
+                    transact.commit(transaction);
+                    broker.flush();
                 }
-
-                final Permission permissions = doc.getPermissions();
-                permissionModifier.accept(permissions);
-
-                broker.storeXMLResource(transaction, doc);
-                transact.commit(transaction);
-                broker.flush();
-            } else {
-                // keep the write lock in the transaction
-                transaction.registerLock(collection.getLock(), LockMode.WRITE_LOCK);
-
-                final Permission permissions = collection.getPermissionsNoLock();
-                permissionModifier.accept(permissions);
-
-                broker.saveCollection(transaction, collection);
-                transact.commit(transaction);
-                broker.flush();
+            } finally {
+                if(collection != null) {
+                    collection.release(LockMode.WRITE_LOCK);
+                }
             }
-        } catch(final XPathException | PermissionDeniedException | IOException | TriggerException | TransactionException e) {
+        } catch(final XPathException | PermissionDeniedException | IOException | TriggerException | TransactionException | LockException e) {
             throw new PermissionDeniedException("Permission to modify permissions is denied for user '" + broker.getCurrentSubject().getName() + "' on '" + pathUri.toString() + "': " + e.getMessage(), e);
-        } finally {
-            if(doc != null) {
-                doc.getUpdateLock().release(LockMode.WRITE_LOCK);
-            }
         }
     }
 }
--- a/src/org/exist/storage/NativeBroker.java
+++ b/src/org/exist/storage/NativeBroker.java
@@ -1536,25 +1536,25 @@ public class NativeBroker extends DBBroker {
                 if(!isRoot) {
                     // remove from parent collection
                     //TODO : resolve URIs ! (uri.resolve(".."))
-                    final Collection parentCollection = openCollection(collection.getParentURI(), LockMode.WRITE_LOCK);
-                    if(parentCollection != null) {
-                        // keep the lock for the transaction
-                        if(transaction != null) {
-                            transaction.registerLock(parentCollection.getLock(), LockMode.WRITE_LOCK);
-                        }
+                    Collection parentCollection = null;
+                    try {
+                        parentCollection = openCollection(collection.getParentURI(), LockMode.WRITE_LOCK);
+                        if(parentCollection != null) {
+                            // keep a lock for the transaction
+                            if(transaction != null) {
+                                transaction.acquireLock(parentCollection.getLock(), LockMode.WRITE_LOCK);
+                            }
 
-                        try {
                             LOG.debug("Removing collection '" + collName + "' from its parent...");
                             //TODO : resolve from collection's base URI
                             parentCollection.removeCollection(this, uri.lastSegment());
                             saveCollection(transaction, parentCollection);
-
-                        } catch(final LockException e) {
-                            LOG.warn("LockException while removing collection '" + collName + "'");
-                        } finally {
-                            if(transaction == null) {
-                                parentCollection.getLock().release(LockMode.WRITE_LOCK);
-                            }
+                        }
+                    } catch(final LockException e) {
+                        LOG.warn("LockException while removing collection '" + collName + "'");
+                    } finally {
+                        if(parentCollection != null) {
+                            parentCollection.getLock().release(LockMode.WRITE_LOCK);
                         }
                     }
                 }
--- a/src/org/exist/storage/txn/Txn.java
+++ b/src/org/exist/storage/txn/Txn.java
@@ -60,11 +60,16 @@ public class Txn implements Transaction {
     public long getId() {
         return id;
     }
-    
+
+    /**
+     * @deprecated Moving a Lock to a Txn is error prone and should not be done
+     *   instead use {@link #acquireLock(Lock, LockMode)} to take a second lock
+     */
+    @Deprecated
     public void registerLock(final Lock lock, final LockMode lockMode) {
         locksHeld.add(new LockInfo(lock, lockMode));
     }
-    
+
     public void acquireLock(final Lock lock, final LockMode lockMode) throws LockException {
         lock.acquire(lockMode);
         locksHeld.add(new LockInfo(lock, lockMode));
--- a/src/org/exist/xmlrpc/RpcConnection.java
+++ b/src/org/exist/xmlrpc/RpcConnection.java
@@ -1486,8 +1486,8 @@ public class RpcConnection implements RpcAPI {
     private boolean storeBinary(final byte[] data, final XmldbURI docUri, final String mimeType,
                                 final int overwrite, final Date created, final Date modified) throws EXistException, PermissionDeniedException {
         return this.<Boolean>writeCollection(docUri.removeLastSegment()).apply((collection, broker, transaction) -> {
-            // keep the write lock in the transaction
-            transaction.registerLock(collection.getLock(), LockMode.WRITE_LOCK);
+            // keep a write lock in the transaction
+            transaction.acquireLock(collection.getLock(), LockMode.WRITE_LOCK);
             if (overwrite == 0) {
                 final DocumentImpl old = collection.getDocument(broker, docUri.lastSegment());
                 if (old != null) {
@@ -1880,8 +1880,8 @@ public class RpcConnection implements RpcAPI {
 
     private boolean remove(final XmldbURI docUri) throws EXistException, PermissionDeniedException {
         return this.<Boolean>writeCollection(docUri.removeLastSegment()).apply((collection, broker, transaction) -> {
-            // keep the write lock in the transaction
-            transaction.registerLock(collection.getLock(), LockMode.WRITE_LOCK);
+            // keep a write lock in the transaction
+            transaction.acquireLock(collection.getLock(), LockMode.WRITE_LOCK);
 
             final DocumentImpl doc = collection.getDocument(broker, docUri.lastSegment());
             if (doc == null) {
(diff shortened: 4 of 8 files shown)


src/org/exist/security/PermissionFactory.java
```diff
@@ -110,40 +111,49 @@ public static Permission getPermission(final SecurityManager sm, final String us
...
-            } else {
-                // keep the write lock in the transaction
-                transaction.registerLock(collection.getLock(), LockMode.WRITE_LOCK);
+                        broker.storeXMLResource(transaction, doc);
+                    } finally {
+                        if(doc != null) {
+                            doc.getUpdateLock().release(LockMode.WRITE_LOCK);
+                        }
+                    }
+                    transact.commit(transaction);
```
shabanovd: I don't think that is right to unlock resource before finish transaction. `doc.getUpdateLock().release(LockMode.WRITE_LOCK);` must happen after `transact.commit(transaction);`.

src/org/exist/storage/txn/Txn.java
```diff
@@ -61,10 +61,6 @@ public long getId() {
         return id;
     }
     
-    public void registerLock(final Lock lock, final LockMode lockMode) {
```
adamretter: Yes, however I really did not think anyone outside of the eXist codebase would be using it. Do you want me to re-add it and mark it as deprecated?

src/org/exist/security/PermissionFactory.java
adamretter: Considering that Transactions in eXist are not proper transactions; Can you explain why?

src/org/exist/storage/txn/Txn.java
shabanovd: Rules are rules. Yes, please, add it back and mark as deprecated.

src/org/exist/security/PermissionFactory.java
shabanovd: because it also flush modification(s) to disk

adamretter:
@shabanovd Okay I addressed your two requests and have rebased.

src/org/exist/storage/txn/Txn.java
adamretter: Done.

src/org/exist/security/PermissionFactory.java
adamretter: Done.

src/org/exist/security/PermissionFactory.java
```diff
@@ -110,40 +111,52 @@ public static Permission getPermission(final SecurityManager sm, final String us
...
+                    DocumentImpl doc = null;
+                    try {
+                        doc = broker.getXMLResource(pathUri, LockMode.WRITE_LOCK);
+                        if (doc == null) {
+                            transact.abort(transaction);
+                            throw new XPathException("Resource or collection '" + pathUri.toString() + "' does not exist.");
+                        }
+
+                        // keep a write lock in the transaction
+                        transaction.acquireLock(doc.getUpdateLock(), LockMode.WRITE_LOCK);
```
shabanovd: why do you put write lock two times?

src/org/exist/storage/NativeBroker.java
```diff
@@ -1524,25 +1526,28 @@ public boolean removeCollection(final Txn transaction, final Collection collecti
...
-                    if(parentCollection != null) {
-                        try {
+                    Collection parentCollection = null;
+                    try {
+                        parentCollection = openCollection(collection.getParentURI(), LockMode.WRITE_LOCK);
+                        if(parentCollection != null) {
+
+                            // keep a lock for the transaction
+                            if(transaction != null) {
+                                transaction.acquireLock(parentCollection.getLock(), LockMode.WRITE_LOCK);
```
shabanovd: why do you get write lock 2 times?

src/org/exist/storage/NativeBroker.java
```diff
@@ -1934,12 +1943,11 @@ public DocumentImpl storeTempResource(final org.exist.dom.memtree.DocumentImpl d
...
                 //store the temporary document
                 temp.addDocument(transaction, this, targetDoc); //NULL transaction, so temporary fragment is not journalled - AR
-                // unlock the temp collection
-                if (transaction == null) {
-                    temp.getLock().release(LockMode.WRITE_LOCK);
-                } else if (!created) {
-                    transaction.registerLock(temp.getLock(), LockMode.WRITE_LOCK);
+
+                if(transaction != null) {
+                    transaction.acquireLock(temp.getLock(), LockMode.WRITE_LOCK);
```
shabanovd: why do you allocate write lock 2nd time?

src/org/exist/xmlrpc/RpcConnection.java
```diff
@@ -1470,8 +1470,8 @@ public boolean storeBinary(final byte[] data, final String documentPath, final S
     private boolean storeBinary(final byte[] data, final XmldbURI docUri, final String mimeType,
                                 final int overwrite, final Date created, final Date modified) throws EXistException, PermissionDeniedException {
         return this.<Boolean>writeCollection(docUri.removeLastSegment()).apply((collection, broker, transaction) -> {
-            // keep the write lock in the transaction
-            transaction.registerLock(collection.getLock(), LockMode.WRITE_LOCK);
+            // keep a write lock in the transaction
+            transaction.acquireLock(collection.getLock(), LockMode.WRITE_LOCK);
```
shabanovd: why you allocate write lock 2nd time?

src/org/exist/xmlrpc/RpcConnection.java
```diff
@@ -1848,8 +1848,8 @@ public boolean remove(final String documentPath) throws URISyntaxException, EXis
 
     private boolean remove(final XmldbURI docUri) throws EXistException, PermissionDeniedException {
         return this.<Boolean>writeCollection(docUri.removeLastSegment()).apply((collection, broker, transaction) -> {
-            // keep the write lock in the transaction
-            transaction.registerLock(collection.getLock(), LockMode.WRITE_LOCK);
+            // keep a write lock in the transaction
+            transaction.acquireLock(collection.getLock(), LockMode.WRITE_LOCK);
```
shabanovd: why you allocate write lock 2nd time?

src/org/exist/xmlrpc/RpcConnection.java
```diff
@@ -1873,8 +1873,8 @@ public boolean removeCollection(final String collectionName) throws URISyntaxExc
     private boolean removeCollection(final XmldbURI collURI) throws EXistException, PermissionDeniedException {
         try {
             return this.<Boolean>writeCollection(collURI).apply((collection, broker, transaction) -> {
-                // keep the write lock in the transaction
-                transaction.registerLock(collection.getLock(), LockMode.WRITE_LOCK);
+                // keep a write lock in the transaction
+                transaction.acquireLock(collection.getLock(), LockMode.WRITE_LOCK);
```
shabanovd: why you allocate write lock 2nd time?

adamretter:
@shabanovd So the whole purpose of the commit is avoid moving the ownership of a lock, as that can introduce dangers as such semantics make writing clear release code very difficult, it also prohibits any use of ARM to improve lock lease management. This is not dissimilar in its improvements to the modern approach of std::copy assignment of std::shared_ptr in C++.

It is much better for both: 1) code simplicity (around lock release), and 2) the ability of a developer to reason about lock lifecycles, to simply take a second lock for the same thread in the transaction.

shabanovd:
@adamretter your comment do not explain why `acquireLock ` better then `registerLock`?

I see that both doing same on `transaction.close`, but `acquireLock ` allocate 2nd time lock and that's totally wrong for me.

adamretter:
@shabanovd I thought that the PR description and my comment above explains this quite well. I will try and think of another formulation for you and post something shortly...

adamretter:
@dizzzz There are no API changes just a deprecated method

src/org/exist/security/PermissionFactory.java
adamretter: To avoid moving the ownership of the lock, which is dangerous and the thing this PR tries to solve. I will explain in detail below

dizzzz:
@shabanovd are you OK with the answer?

adamretter:
So I will try and set out here why I believe that moving the ownership of lock semantics (i.e. `Txn#registerLock`) is dangerous, and instead we should prefer incrementing the lock hold count (i.e. `Txn#acquireLock` (when we already hold the lock)).


1. A major problem with moving the lock ownership is that it leads to a asymmetrical API which confuses developers about whether they should or should not release locks:

    1.1 For example, `NativeBroker#openCollection`, `NativeBroker#getCollection`, `NativeBroker#createCollection` and `NativeBroker#getOrCreateCollection` all have different lock semantics, some of those functions require the caller to manually release lock(s) and others do not.

    1.2 The confusion this causes to developers, is both significant and real, one such reported incidence is here: https://github.com/eXist-db/messaging-replication/pull/121#pullrequestreview-24330301


2. Moving lock ownership (with `Txn#registerLock`) greatly complicates the code for lock release and is often incorrect:

	2.1 `NativeBroker#removeCollection` hides a potential NPE https://github.com/eXist-db/exist/blob/develop/src/org/exist/storage/NativeBroker.java#L1533 and also has uncertain release semantics https://github.com/eXist-db/exist/blob/develop/src/org/exist/storage/NativeBroker.java#L1547

	2.2 `NativerBroker#storeTempResource` hides a potential NPE https://github.com/eXist-db/exist/blob/develop/src/org/exist/storage/NativeBroker.java#L1921 and has complicated release semantics https://github.com/eXist-db/exist/blob/develop/src/org/exist/storage/NativeBroker.java#L1941 which are impossible to predict without running the code.

	2.3 `RpcConnection#remove`, `RpcConnection#removeCollection` and `RpcConnection#storeBinary` all release their locks too many times.

	2.4 There are plenty of other examples in non-core code too...

3. Moving lock ownership prevents us from using ARM to ensure that Locks are easily and correctly closed when they should be by using try-with-resources (or similar). For example, a future improvement could be to have something like the following:

```java
try(final Collection col = broker.openCollection(uri, LockMode.WRITE_LOCK)) {
	//do some write operations on the collection...
}
```

If we tried to use `Txn#registerLock` in there:
```java
try(final Collection col = broker.openCollection(uri, LockMode.WRITE_LOCK)) {
	//do some write operations on the collection...

	txn.regsisterLock(col.getLock(), LockMode.WRITE_LOCK);

	...
}
```

Then the lock would be incorrectly released twice, and you could not tell from the above code whether it would be finally released by `Txn#commit` or `Collection#close`.

However, if we switch to `Txn#acquireLock`, then all is fine:

```
If we tried to use `Txn#registerLock` in there:
```java
try(final Collection col = broker.openCollection(uri, LockMode.WRITE_LOCK)) {
	//do some write operations on the collection...

	txn.acquireLock(col.getLock(), LockMode.WRITE_LOCK);

	...
}
```

In this case there the lock hold count for the Collection is `2`. One will be released by `Txn#close` and the other by `Collection#close`. There are no problems with release lock ordering or releasing locks too many times.

shabanovd:
@adamretter I still can't understand why `acquireLock` better then `registerLock`. It still require to unlock at finally-block. And logic become very confusion: you call method that allocate lock that's why you must unlock at finally-block, but also you must allocate lock 2nd time by calling `acquireLock`.

Making `Collection` or `Document` closable is design bug because it's singleton and you can't handle state without loosing thread safety (simple way).

There is only one object that design to hold state for transaction time - transaction itself. Then code should be next:
```
try (Txn txn = ...) {
	final Collection col = broker.openCollection(uri, LockMode.WRITE_LOCK)
	//do some write operations on the collection...
	...
	txn.commit();
}
```

To make this happen `openCollection` (and others) method must allocate lock and register it with transaction.

I don't see how it possible to do without API changes (change in behavior) and that's why it can be done for 4.x only.

adamretter:
@shabanovd A few answers to your concerns/confusion:

1. There is **no** second lock that is allocated, once you have a lock, only the lock hold count is incremented when you call acquire more than once.

2. `Collection` is not a singleton. Collection is only an interface. `MutableCollection` is not a singleton, but the way that it is served from the `CollectionCache` ensures that there *should* only be one instance of a `MutableCollection` class for each collection in the database.

    2.1 Perhaps consider that in future there could be other implementations of `Collection` which are `AutoCloseable`, such as a [`LockedCollection`](https://github.com/adamretter/exist/blob/cache-improvements/src/org/exist/collections/LockedCollection.java).

3. **This PR** does not make any API changes, it simply improves the safety of locking and simplifies the code around lock release semantics to make it more easily understandable by developers. If we need to change an API in *future* then yes we would do a 4.x or whichever `MAJOR` release it happens to be.

shabanovd:
@adamretter you answer your own questions -(

> There is no second lock that is allocated, once you have a lock, only the lock hold count is incremented when you call acquire more than once.

Programmer see it as 2nd lock and it doesn't matter how it handled internally.

> Collection is not a singleton. ... the CollectionCache ensures that there should only be one instance

That make it singleton!

>  `LockedCollection`

Collection object going to do transaction object job - bad design.

> `This PR does not make any API changes`

Where did I say that this PR make any API changes?

adamretter:
> Programmer see it as 2nd lock and it doesn't matter how it handled internally.
Good I want them to see it as a lock; and it kinda is as when the lock hold count is greater than `0` it prevents use exclusively by other threads; Internally it will have much better performance.
So it's a win-win.

> That make it singleton!
In Java a `Singleton` is something where there can `only` be one instance of an object identity in memory at once; I think you already know that ;-)

> Collection object going to do transaction object job - bad design.
Not at all they have different lifetimes and responsibilities with regards to locking. A Transaction *may* have acquired the locks of many Collections.

> You can say it as much times as you like, but that will not make me to see it. I don't see any improves!

That's sad. I will find others to do the review and see what their conclusions are.

shabanovd:
> Good I want them to see it as a lock; and it kinda is as when the lock hold count is greater than 0 it prevents use exclusively by other threads; Internally it will have much better performance.
So it's a win-win.

1st, performance can be proved only by test, not by words.
2nd, see that resource require 2 times locking confuse a lot.

> In Java a Singleton is something where there can only be one instance of an object identity in memory at once; I think you already know that ;-)

Did I say `Java a Singleton` anywhere? or always reference to `design singleton`? Collections are singleton by design because of `CollectionCache`?

> Not at all they have different lifetimes and responsibilities with regards to locking. A Transaction may have acquired the locks of many Collections.

That is your way to see things. I see it differently and I'm in panic if design do not clearly separate meaning, behavior and responsibility of objects (classes).

shabanovd:
@adamretter put this PR to call agenda, no more comments or discussions after `I don't believe that Dmitriy is working in the spirit of the community`

caseydawsonjordan:
So I think the issue here is largely semantic and I have a suggestion which I think would solve all the confusion. 

[First let me explain my understanding of the issue]

Put simply, aquireLock() or registerLock() is telling the transaction to persist that lock until the transaction closes. However, because we are also acquiring a lock in the try-with-resources this becomes confusing to the developer, as it appears that the lock is being acquired twice (@shabanovd 's point), but in reality it is simply prolonging the lifetime of the lock (Which is needed in some cases). 

I propose a simple method change that I think would make this much clearer to any developer reading it. Change `txn.aquireLock(col.getLock(), LockMode.WRITE_LOCK)` to `txn.persistLockUntilTxnClose(col.getLock(), LockMode.WRITE_LOCK)`.

While this change makes a very verbose method, it also makes it very clear to any developer what is happening when that method is called.

Granted, I think that in general a larger API change down the line is needed to really bring all of this together, but that is not going to be an easy change.

What @adamretter has here I think is a great improvement and if the tests show that it solves some of the major locking issues we should agree on this change and get this into the codebase. 

Cheers,

Casey

shabanovd:
@caseydawsonjordan I agree that @adamretter 's bugfixes are important but problem that he did mixed up into simple PR bugfixes and API changes. It must be 2 different PRs to simplify everyone life.

adamretter:
@shabanovd where do you see API changes?

adamretter:
@wolfgangmm @dizzzz I want to be clear that there are no API changes introduced by this PR. 

@caseydawsonjordan has suggested that I change: `txn.aquireLock(col.getLock(), LockMode.WRITE_LOCK)` to `txn.persistLockUntilTxnClose(col.getLock(), LockMode.WRITE_LOCK)`. That has not yet been done, however if I do that I would add a method and deprecate the old one so that there are still no incompatible API changes.

shabanovd:
@adamretter [this](https://github.com/eXist-db/exist/pull/1296/files#diff-ad5cd86a3ad8a3efe443b57d78446cd4R68) is API change!

adamretter:
@shabanovd changing a private method is not an API change.

shabanovd:
@adamretter `registerLock` is not a private method, it have clear `public` declaration. Why do you call it private?

adamretter:
@shabanovd The link you posted does not point to `Txn#registerLock`.

The only change I made to `registerLock` is to add a `@Deprecated` label. That is not a breaking API change, to quote the rules that you like to reference:

> `MINOR` versions add functionality or deprecate API functions, without breaking backward compatibility; and

shabanovd:
@adamretter why link above didn't point to right place question to github, sometime it fail -(

`@Deprecated` have clear meaning - that method (class) will be removed, and that's why it's "future" API change.

adamretter:
@shabanovd I have no idea what argument you are trying to make any more. The versioning doc says that marking a API method as deprecated is a `MINOR` revision. I submitted this PR in good faith, so that it might be included in eXist 3.1.0. 3.1.0 is a `MINOR` revision from 3.0.0, so I do not see any problem. Please, tell me what problem you see?

wolfgangmm:
I'm not sure if this is a coincidence caused by my extensive testing before, but after applying this PR locally, I was running into a deadlock. See attached thread dump.

[lock.txt](https://github.com/eXist-db/exist/files/819342/lock.txt)

adamretter:
@wolfgangmm As this PR does not change the locking order or semantics, I think it is likely a coincidence and the deadlock was present before this PR. To be sure though, can you tell me how to reproduce what you are seeing?

wolfgangmm:
@adamretter as with most deadlocks, exact timing is key, so this is not easily reproducible and I fear all we have is the thread dump I managed to save. I have not experienced any real deadlocks since months (or years) though. I'll see if I can run into the issue again.

dizzzz:
@wolfgangmm no I did not test it with the extension .... yet...

dizzzz:
Pooh, this PR is hanging for a long time already, a pity.

adamretter:
@dizzzz I fixed the merge conflicts and rebased.

PR #1297, opened 2017-02-26
[feature] Add support for string constructors

wolfgangmm:
A String Constructor creates a string from literal text and interpolated expressions. See
 [XQuery 3.1 spec](https://www.w3.org/TR/xquery-31/#id-string-constructors)


joewiz:
Slick - another XQuery 3.1 feature I cover in my book that eXist will now support. Thanks @wolfgangmm!

PR #1299, opened 2017-02-26
[feature] Implement fn:load-xquery-module

wolfgangmm:
This reuses some code from the inspection module, which is well-tested. A temporary context is created for the imported module so the current context does not get polluted.


shabanovd:
@wolfgangmm what is purpose of it? (usecase?)

wolfgangmm:
@shabanovd see [the spec](https://www.w3.org/TR/xpath-functions-31/#func-load-xquery-module). You need it to dynamically load a module and call its functions, which is done by many of our libraries (templating, xqsuite...). For example, it could be used instead of the non-standard inspect:module-functions in suite.xql.

shabanovd:
@wolfgangmm I see. Should we deprecate `inspect:module-functions`?

wolfgangmm:
@shabanovd I thought about it, but unfortunately we still need the inspect module in general because it's the only way to get additional information about a function (annotations!). Also inspect:module-functions without argument can be used to get all functions of the current module. So I'm not sure I want to deprecate it for the time being.

I also tried to rewrite suite.xql to use load-xquery-module instead, but it becomes a bit verbose:

```xquery
declare function local:funcs($uri, $location) {
    load-xquery-module($uri, map {
        "location-hints": $location
    })?functions?*?*
};

test:suite((
    local:funcs("http://exist-db.org/xquery/test/load-xquery-module", "load-xquery-module.xql")
))
```

joewiz:
@wolfgangmm Thanks for the template for adapting to this - and for the next function in support of XQuery 3.1!

shabanovd:
@adamretter is it ok to go?

adamretter:
@shabanovd I am still working through the other older PRs first. I will get to this soon.

src/org/exist/xquery/ErrorCodes.java
```diff
@@ -213,6 +213,7 @@
     public static final ErrorCode FOJS0003 = new W3CErrorCode("FOJS0003", "JSON duplicate keys.");
 
     public static final ErrorCode FOUT1170 = new W3CErrorCode("FOUT1170", "Invalid $href argument to fn:unparsed-text() (etc.)");
+    public static final ErrorCode FOQM0002 = new W3CErrorCode("FOQM0002", "Module URI not found.");
```
adamretter: I think we are missing `FOQM0001`, `err:FOQM0003`, `err:FOQM0005` and `err:FOQM0006`.
Although `err:FOQM0006` is unlikely in eXist-db.

src/org/exist/xquery/functions/fn/LoadXQueryModule.java
```diff
@@ -0,0 +1,214 @@
...
+                            "the same name (but different arity), they share the same entry. The key of the entry is the name of the " +
+                            "function(s), as an xs:QName value; the associated value is a map A. This map (A) contains one entry for each " +
+                            "function with the given name; its key is the arity of the function, as an xs:integer value, and its associated " +
+                            "value is the function itself, as a function item. The function can be invoked using the rules for dynamic " +
+                            "function invocation.")
+    );
+
+    public final static StringValue OPTIONS_LOCATION_HINTS = new StringValue("location-hints");
+    public final static StringValue OPTIONS_XQUERY_VERSION = new StringValue("xquery-version");
+
```
adamretter: We are missing the options:

1. `context-item`
2. `variables`
3. `vendor-options`

I think that (1) and (2) should be easy to support. We don't have to support any actual options for `3` yet.

src/org/exist/xquery/functions/fn/LoadXQueryModule.java
```diff
@@ -0,0 +1,214 @@
...
+                        xqVersion.equals(getXQueryVersion(tempContext.getXQueryVersion()))) {
+                    module = importedModule;
+                    break;
+                }
+            }
+        }
+
+        // not found, raise error
+        if (module == null) {
+            throw new XPathException(this, ErrorCodes.FOQM0002, "Module with URI " + targetNamespace + " not found");
```
adamretter: I don't think this is the right error if `xqVersion` doesn't match. However I am not certain from the spec what the error should be.

src/org/exist/xquery/functions/fn/LoadXQueryModule.java
```diff
@@ -0,0 +1,214 @@
...
+        final MapType result = new MapType(context);
+
+        final ValueSequence functionSeq = new ValueSequence();
+        addFunctionRefsFromModule(this, tempContext, functionSeq, module);
+        final MapType functions = new MapType(context);
+        for (final SequenceIterator i = functionSeq.iterate(); i.hasNext(); ) {
+            final FunctionReference ref = (FunctionReference) i.nextItem();
+            final FunctionSignature signature = ref.getSignature();
+            final QNameValue qn = new QNameValue(context, signature.getName());
+            MapType entry;
```
adamretter: This could be `final`

src/org/exist/xquery/functions/fn/LoadXQueryModule.java
```diff
@@ -0,0 +1,214 @@
...
+    public static void addFunctionRefsFromModule(Expression parent, XQueryContext tempContext, ValueSequence resultSeq,
+                                                 Module module)
+            throws XPathException {
+        final FunctionSignature signatures[] = module.listFunctions();
+        for (final FunctionSignature signature : signatures) {
+            if (!signature.isPrivate()) {
+                if (module.isInternalModule()) {
+                    int arity;
+                    if (signature.isOverloaded())
+                    {arity = signature.getArgumentTypes().length;}
```
adamretter: Not critical, but, how would you feel about using the IntelliJ code formatter?

src/org/exist/xquery/functions/fn/LoadXQueryModule.java
wolfgangmm: (2) is difficult because eXist does neither support external variables in library modules nor default values. I tried to add both, so there will be a new commit once the test suite passes.

adamretter: Okay sounds good.

wolfgangmm:
I'm not entirely happy with my second commit, which - besides addressing @adamretter's requests - adds support for specifying a default value on a variable declared as external. All tests pass and everything works fine, but due to the different ways we use to set external variable values throughout the test suite and interfaces, two or three tweaks are required to handle all cases. Most of those tweaks have been there a long time and are not related to this PR, yet I'm unhappy about them. A thorough review of all places where external variables are set from Java code would be necessary, but it would cost several more days.

@adamretter @shabanovd @dizzzz @ljo @joewiz if you think the benefit of supporting missing XQuery features is worth it (notably external vars with defaults in library modules), you may want to merge the PR and I'll continue the cleanup later. Otherwise please leave this PR for a later version.

shabanovd:
@wolfgangmm I'm happy to have feature partly implemented because it small step forward. You or someone else will be able to put in missing parts in future.

src/org/exist/xquery/XQueryContext.java
```diff
@@ -256,6 +256,9 @@
     /** Should empty order greatest or least? */
     private boolean                                    orderEmptyGreatest            = true;
 
+    /** The context item set in the query prolog or externally */
+    private Sequence                                   contextItem                   = Sequence.EMPTY_SEQUENCE;
```
shabanovd: this require reset?

src/org/exist/xquery/VariableDeclaration.java
```diff
@@ -37,9 +40,9 @@
  */
 public class VariableDeclaration extends AbstractExpression implements RewritableExpression {
 
-	QName qname;
-	SequenceType sequenceType = null;
-	Expression expression;
+	final QName qname;
```
dizzzz: indenting not OK ?

adamretter:
@wolfgangmm As this sounds like a positive step in the right direction and @dizzzz and @shabanovd indicated they were happy. I have manually merged this in (as I needed to rebase it because of a conflict in the generated Antlr parser).

PR #1304, opened 2017-02-28
Upgrade of log4j2 framework 

dizzzz:
### Description:

upgrade of log4j framework

### Reference:

https://logging.apache.org/log4j/2.x/changes-report.html#a2.8

### Type of tests:

- run unit testsuite
- start database


adamretter:
@dizzzz Looks good. Just one thing -  I think  you also need to update the `.classpath` file

dizzzz:
and the netbeans files........ a switch to maven would be a help here :-)
actually I think this file (c|sh)ould be generated

shabanovd:
@dizzzz really - that files do not autogenerated and IDE will raise error.

dizzzz:
Done

dizzzz:
@shabanovd  I only wanted to state that this file is soo simple and could be derived from other sources. Just like the netbeans project files, the fileset is not complete. I'd rather see these removed, because clearly no-one used them as they are outdated... all the time. they just do not work :-)

shabanovd:
@dizzzz can you add note to next call agenda?

dizzzz:
re: removing the project files?

shabanovd:
@dizzzz yes, `netbeans and eclipse project files`

adamretter:
@dizzzz So NetBeans and IntelliJ can work directly with Maven projects without the need for generating extra project files. Maven also have a plugin for generating Eclipse project files

dizzzz:
@adamretter I know, I know

PR #1307, opened 2017-03-01
Reworked IP range

dizzzz:
reworked version of https://github.com/eXist-db/exist/pull/834

### DIFF
--- a/extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
+++ b/extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
@@ -104,6 +104,7 @@ public class IPRangeRealm extends AbstractRealm {
 
             if (xquery == null) {
                 LOG.error("IPRange broker unable to retrieve XQueryService");
+                return null;
             }
 
             String query = "collection('/db/system/security/IPRange/accounts')/account/iprange[" + ipToTest + " ge number(start) and " + ipToTest + " le number(end)]/../name";
--- a/extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
+++ b/extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
@@ -1,6 +1,6 @@
 /*
  *  eXist Open Source Native XML Database
- *  Copyright (C) 2011 The eXist Project
+ *  Copyright (C) 2017 The eXist Project
  *  http://exist-db.org
  *
  *  This program is free software; you can redistribute it and/or
@@ -16,8 +16,6 @@
  *  You should have received a copy of the GNU Lesser General Public
  *  License along with this library; if not, write to the Free Software
  *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
- *
- *  $Id$
  */
 package org.exist.security.realm.iprange;
 
@@ -57,12 +55,16 @@ public class IPRangeRealm extends AbstractRealm {
     @ConfigurationFieldAsAttribute("version")
     public final static String version = "1.0";
 
-    protected final static Logger LOG = LogManager.getLogger(IPRangeRealm.class);
-    protected static IPRangeRealm instance = null;
+    private final static Logger LOG = LogManager.getLogger(IPRangeRealm.class);
+    private static IPRangeRealm instance = null;
 
     public IPRangeRealm(final SecurityManagerImpl sm, final Configuration config) throws ConfigurationException {
         super(sm, config);
         instance = this;
+    }
+
+    static IPRangeRealm getInstance(){
+        return instance;
     }
 
     private static long ipToLong(final InetAddress ip) {
@@ -101,7 +103,7 @@ public class IPRangeRealm extends AbstractRealm {
     public Subject authenticate(final String ipAddress, final Object credentials) throws AuthenticationException {
 
         // Elevaste to system privileges
-        try (final DBBroker broker = BrokerPool.getInstance().get(Optional.of(getSecurityManager().getSystemSubject()))) {
+        try (final DBBroker broker = getSecurityManager().database().get(Optional.of(getSecurityManager().getSystemSubject()))) {
 
             // Convert IP address
             final long ipToTest = ipToLong(InetAddress.getByName(ipAddress));
@@ -114,7 +116,7 @@ public class IPRangeRealm extends AbstractRealm {
             }
 
             // Construct XQuery
-            final String query = "collection('/db/system/security/IPRange/accounts')/account/" +
+            final String query = "collection('/db/system/security/iprange/accounts')/account/" +
                     "iprange[" + ipToTest + " ge number(start) and " + ipToTest + " le number(end)]/../name";
             final XQueryContext context = new XQueryContext(broker.getBrokerPool());
 
@@ -133,7 +135,7 @@ public class IPRangeRealm extends AbstractRealm {
                 LOG.warn("IP address " + ipAddress + " matched multiple ipranges. Using first result only.");
             }
 
-            if (!"".equals(username)) {
+            if (!username.isEmpty()) {
                 final Account account = getSecurityManager().getAccount(username);
                 if (account != null) {
                     LOG.info("IPRangeRealm trying " + account.getName());
@@ -147,7 +149,7 @@ public class IPRangeRealm extends AbstractRealm {
             }
             return null;
 
-        } catch (EXistException | UnknownHostException | XPathException | PermissionDeniedException e) {
+        } catch (final EXistException | UnknownHostException | XPathException | PermissionDeniedException e) {
             throw new AuthenticationException(AuthenticationException.UNNOWN_EXCEPTION, e.getMessage());
         }
     }
--- a/extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
+++ b/extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
@@ -21,128 +21,120 @@
  */
 package org.exist.security.realm.iprange;
 
-import java.util.Properties;
-import java.util.Optional;
-import java.net.*;
-
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.exist.EXistException;
 import org.exist.config.Configuration;
 import org.exist.config.ConfigurationException;
-import org.exist.config.annotation.*;
-import org.exist.security.AbstractAccount;
-import org.exist.security.AbstractRealm;
-import org.exist.security.Account;
-import org.exist.security.AuthenticationException;
-import org.exist.security.Group;
-import org.exist.security.PermissionDeniedException;
-import org.exist.security.Subject;
+import org.exist.config.annotation.ConfigurationClass;
+import org.exist.config.annotation.ConfigurationFieldAsAttribute;
+import org.exist.security.*;
 import org.exist.security.internal.SecurityManagerImpl;
 import org.exist.security.internal.SubjectAccreditedImpl;
 import org.exist.storage.BrokerPool;
 import org.exist.storage.DBBroker;
 import org.exist.xquery.CompiledXQuery;
+import org.exist.xquery.XPathException;
 import org.exist.xquery.XQuery;
 import org.exist.xquery.XQueryContext;
 import org.exist.xquery.value.Sequence;
 import org.exist.xquery.value.SequenceIterator;
-import org.exist.xquery.XPathException;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.Optional;
+import java.util.Properties;
 
 /**
  * @author <a href="mailto:wshager@gmail.com">Wouter Hager</a>
- *
  */
 @ConfigurationClass("realm") //TODO: id = IPRange
 public class IPRangeRealm extends AbstractRealm {
-	protected final static Logger LOG = LogManager.getLogger(IPRangeRealm.class);
-	
-	protected static IPRangeRealm instance = null;
-	
-	@ConfigurationFieldAsAttribute("id")
-	public final static String ID = "IPRange";
-	
-	@ConfigurationFieldAsAttribute("version")
-	public final static String version = "1.0";
-	
-	public IPRangeRealm(final SecurityManagerImpl sm, Configuration config) throws ConfigurationException {
-		super(sm, config);
-		instance = this;
-	}
+    @ConfigurationFieldAsAttribute("id")
+    public final static String ID = "IPRange";
+    @ConfigurationFieldAsAttribute("version")
+    public final static String version = "1.0";
+    protected final static Logger LOG = LogManager.getLogger(IPRangeRealm.class);
+    protected static IPRangeRealm instance = null;
 
-	@Override
-	public String getId() {
-		return ID;
-	}
-	
+    public IPRangeRealm(final SecurityManagerImpl sm, Configuration config) throws ConfigurationException {
+        super(sm, config);
+        instance = this;
+    }
+
+    private static long ipToLong(InetAddress ip) {
+        byte[] octets = ip.getAddress();
+        long result = 0;
+        for (byte octet : octets) {
+            result <<= 8;
+            result |= octet & 0xff;
+        }
+        return result;
+    }
+
 	/*@Override
-	public void start(DBBroker broker) throws EXistException {
+    public void start(DBBroker broker) throws EXistException {
 		super.start(broker);
 	}*/
-	
-	@Override
-	public boolean deleteAccount(Account account) throws PermissionDeniedException, EXistException, ConfigurationException {
-		// Auto-generated method stub
-		return false;
-	}
 
-	@Override
-	public boolean deleteGroup(Group group) throws PermissionDeniedException, EXistException, ConfigurationException {
-		// Auto-generated method stub
-		return false;
-	}
-	
-	@Override
-	public Subject authenticate(final String ip, Object credentials) throws AuthenticationException {
-		//elevate to system privs
-		try(final DBBroker broker = BrokerPool.getInstance().get(Optional.of(getSecurityManager().getSystemSubject()))) {
-			long ipToTest = ipToLong(InetAddress.getByName(ip));
-			
-			XQuery xquery = broker.getBrokerPool().getXQueryService();
+    @Override
+    public String getId() {
+        return ID;
+    }
 
-			if (xquery == null) {
-				LOG.error("IPRange broker unable to retrieve XQueryService");
-			}
-			
-			String query = "collection('/db/system/security/IPRange/accounts')/account/iprange[" + ipToTest + " ge number(start) and " + ipToTest + " le number(end)]/../name";
-			
-			XQueryContext context = new XQueryContext(broker.getBrokerPool());
+    @Override
+    public boolean deleteAccount(Account account) throws PermissionDeniedException, EXistException, ConfigurationException {
+        // Auto-generated method stub
+        return false;
+    }
 
-			CompiledXQuery compiled = xquery.compile(broker, context, query);
+    @Override
+    public boolean deleteGroup(Group group) throws PermissionDeniedException, EXistException, ConfigurationException {
+        // Auto-generated method stub
+        return false;
+    }
 
-			Properties outputProperties = new Properties();
+    @Override
+    public Subject authenticate(final String ip, Object credentials) throws AuthenticationException {
+        //elevate to system privs
+        try (final DBBroker broker = BrokerPool.getInstance().get(Optional.of(getSecurityManager().getSystemSubject()))) {
+            long ipToTest = ipToLong(InetAddress.getByName(ip));
 
-			Sequence result = xquery.execute(broker, compiled, null, outputProperties);
-			SequenceIterator i = result.iterate();
-			String username = "";
-			if(i.hasNext()) {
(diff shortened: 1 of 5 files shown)


joewiz:
@dizzzz Assuming all is well, do you prefer a straight merge, or a squash and merge?

dizzzz:
Squash n merge seem to be disabled :-(

dizzzz:
I'd like to have a confirmation of @wshager that it works.....

adamretter:
@dizzzz I disabled the newer PR merge options, because it makes undoing merges very hard. In the teleconf, @wolfgangmm voiced that he wanted the ability to be able to un-merge PRs that were proving troublesome before committing to a final release.

adamretter:
Nice @dizzzz. I am happy to review and squash this once @wshager confirms success :-)

dizzzz:
I have  an other idea in this context, making things a bit more pluggable; future PR

wshager:
It would be nice if security realms could be some kind of plugin, and not rely in changes to core classes. I'm aware of FIXME's in the code suggesting this direction. I myself don't think something this specific should be merged.

wshager:
It just makes things more complex for my clients. They should just be able to run exist ootb.

shabanovd:
@wshager I think that it going to be resolved by new authentication architecture. It's good time to do it because there are several different realms with similar needs to attach to servlet container (jetty) + better integration with jetty authentication.

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
```diff
@@ -0,0 +1,154 @@
...
+    public boolean deleteGroup(final Group group) throws PermissionDeniedException, EXistException, ConfigurationException {
+        // Auto-generated method stub
+        return false;
+    }
+
+    @Override
+    public Subject authenticate(final String ipAddress, final Object credentials) throws AuthenticationException {
+
+        // Elevaste to system privileges
+        try (final DBBroker broker = BrokerPool.getInstance().get(Optional.of(getSecurityManager().getSystemSubject()))) {
```
shabanovd: replace `BrokerPool.getInstance()` by `sm.database()`

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
```diff
@@ -0,0 +1,121 @@
...
+                    return;
+                }
+
+
+                final HttpSession session = request.getSession();
+                // store the user in the session
+                if (session != null) {
+                    jsonResponse = "{\"user\":\"" + user.getUsername() + "\",\"isAdmin\":\"" + user.hasDbaRole() + "\"}";
+                    LOG.info("IPRangeServlet setting session attr " + XQueryContext.HTTP_SESSIONVAR_XMLDB_USER);
+                    session.setAttribute(XQueryContext.HTTP_SESSIONVAR_XMLDB_USER, user);
```
shabanovd: this looks a bit wrong. Do you want to authenticate to any operations on server?

If yes then here better [code](https://github.com/eXist-db/exist/blob/develop/extensions/security/oauth/src/org/exist/security/realm/oauth/ServiceFacebook.java#L118-L131) for it.

dizzzz: it is really an IP range authentication extension

shabanovd: what does that mean? only authentication and no authorisation?

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
```diff
@@ -0,0 +1,154 @@
+/*
+ *  eXist Open Source Native XML Database
+ *  Copyright (C) 2011 The eXist Project
```
adamretter: Now I feel young! `2011`?

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
```diff
@@ -0,0 +1,154 @@
...
+
+            // Get xquery service
+            final XQuery queryService = broker.getBrokerPool().getXQueryService();
+            if (queryService == null) {
+                LOG.error("IPRange broker unable to retrieve XQueryService");
+                return null;
+            }
+
+            // Construct XQuery
+            final String query = "collection('/db/system/security/IPRange/accounts')/account/" +
```
adamretter: I would personally prefer `/db/system/security/iprange/accounts`

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
```diff
@@ -0,0 +1,154 @@
...
+                    "iprange[" + ipToTest + " ge number(start) and " + ipToTest + " le number(end)]/../name";
+            final XQueryContext context = new XQueryContext(broker.getBrokerPool());
+
+            final CompiledXQuery compiled = queryService.compile(broker, context, query);
+
+            final Properties outputProperties = new Properties();
+
+            // Execute xQuery
+            final Sequence result = queryService.execute(broker, compiled, null, outputProperties);
```
adamretter: This is maybe a comment for a future improvement:

At the moment you execute an XQuery upon every authenticate request. I should imagine that is not going to be very performant. It would be better to load the configuration from the `/db/system/security/IPRange/accounts` at startup or on modification.

A quick improvement would be to cache the compiled query, at the moment the query is compiled on every call.

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
```diff
@@ -0,0 +1,154 @@
...
+            final SequenceIterator i = result.iterate();
+
+            // Get FIRST username when present
+            final String username = i.hasNext() ? i.nextItem().getStringValue() : "";
+
+            if (i.hasNext()) {
+                LOG.warn("IP address " + ipAddress + " matched multiple ipranges. Using first result only.");
+            }
+
+            if (!"".equals(username)) {
```
adamretter: `if(!username.isEmpty())`

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
```diff
@@ -0,0 +1,154 @@
...
+                final Account account = getSecurityManager().getAccount(username);
+                if (account != null) {
+                    LOG.info("IPRangeRealm trying " + account.getName());
```
adamretter: I think this (and a few of the below statements) could be `LOG.trace` otherwise there is going to be a lot of logging.

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
```diff
@@ -0,0 +1,154 @@
...
+                } else {
+                    LOG.info("IPRangeRealm couldn't resolve account for " + username);
+                }
+
+            } else {
+                LOG.info("IPRangeRealm xquery found no matches");
+            }
+            return null;
+
+        } catch (EXistException | UnknownHostException | XPathException | PermissionDeniedException e) {
```
adamretter: `catch (final`

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
```diff
@@ -0,0 +1,121 @@
+/*
+ *  eXist Open Source Native XML Database
+ *  Copyright (C) 2009-2011 The eXist Project
```
adamretter: 2011 again? ;-)

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @author <a href="mailto:wshager@gmail.com">Wouter Hager</a>
+ */
+public class IPRangeServlet extends HttpServlet {
+
+    protected final static Logger LOG = LogManager.getLogger(IPRangeServlet.class);
+    private static final long serialVersionUID = -568037449837549034L;
+    public static AbstractRealm realm = null;
+
+    public IPRangeServlet() throws ServletException {
```
adamretter: Empty default constructor is redundant code.

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
```diff
@@ -0,0 +1,121 @@
...
+            ipAddress = request.getRemoteAddr();
+        }
+
+
+        LOG.info("Detected IPaddress " + ipAddress);
+
+        String jsonResponse = "{\"fail\":\"IP range not authenticated\"}";
+
+        try {
+            final SecurityManager securityManager = IPRangeRealm.instance.getSecurityManager();
```
adamretter: This is not nice. At least consider `IPRangeRealm.getInstance()`. Otherwise go through the `BrokerPool` to get the `SecurityManager`

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
```diff
@@ -0,0 +1,154 @@
...
+public class IPRangeRealm extends AbstractRealm {
+
+    @ConfigurationFieldAsAttribute("id")
+    public final static String ID = "IPRange";
+
+    @ConfigurationFieldAsAttribute("version")
+    public final static String version = "1.0";
+
+    protected final static Logger LOG = LogManager.getLogger(IPRangeRealm.class);
+    protected static IPRangeRealm instance = null;
```
adamretter: This should be `private`

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
```diff
@@ -0,0 +1,154 @@
...
+@ConfigurationClass("realm") //TODO: id = IPRange
+public class IPRangeRealm extends AbstractRealm {
+
+    @ConfigurationFieldAsAttribute("id")
+    public final static String ID = "IPRange";
+
+    @ConfigurationFieldAsAttribute("version")
+    public final static String version = "1.0";
+
+    protected final static Logger LOG = LogManager.getLogger(IPRangeRealm.class);
```
adamretter: This should be `private`

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * IPRange authenticator servlet.
+ *
+ * @author <a href="mailto:wshager@gmail.com">Wouter Hager</a>
+ */
+public class IPRangeServlet extends HttpServlet {
+
+    protected final static Logger LOG = LogManager.getLogger(IPRangeServlet.class);
+    private static final long serialVersionUID = -568037449837549034L;
+    public static AbstractRealm realm = null;
```
adamretter: This should be `private`

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.PrintWriter;
+
+/**
+ * IPRange authenticator servlet.
+ *
+ * @author <a href="mailto:wshager@gmail.com">Wouter Hager</a>
+ */
+public class IPRangeServlet extends HttpServlet {
+
+    protected final static Logger LOG = LogManager.getLogger(IPRangeServlet.class);
```
adamretter: This should be `private`

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+                final HttpSession session = request.getSession();
+                // store the user in the session
+                if (session != null) {
+                    jsonResponse = "{\"user\":\"" + user.getUsername() + "\",\"isAdmin\":\"" + user.hasDbaRole() + "\"}";
+                    LOG.info("IPRangeServlet setting session attr " + XQueryContext.HTTP_SESSIONVAR_XMLDB_USER);
+                    session.setAttribute(XQueryContext.HTTP_SESSIONVAR_XMLDB_USER, user);
+                } else {
+                    LOG.info("IPRangeServlet session is null");
```
adamretter: Should this be `info` level?

wshager:
@adamretter Wow, that's a lot to improve. Clearly I'm no Java expert.

The logging is a bit arbitrary in some places. Perhaps I should add a debug flag? 

What's the severity in case session is null?

adamretter:
@wshager They are only suggestions not requirements ;-)

Regards the log level - you need to ask yourself - If there is no session, is it a critical failure for your IP Range Auth or not? If so, then it is `ERROR`, if not it might be `WARN` or `TRACE.

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
dizzzz: the idea is: certain IP ranges get a fixed user assigned. so the IPaddress is the authentication. I would never do this, but it is wouter's use-case.

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
dizzzz: I forgot to make a similar remark. Probably it is better to have xquery functions to set/unset the data used for authentication.

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
dizzzz: actually this is copied from OAuthServlet

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
dizzzz: done

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
dizzzz: done

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
dizzzz: done

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
dizzzz: I changed this, it is more consistent. please be aware @wshager

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
dizzzz: how could I miss this....

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeRealm.java
dizzzz: done

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
dizzzz: done

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
dizzzz: removed, was not used :-/

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
dizzzz: done

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
dizzzz: done

dizzzz:
reworked

adamretter:
@dizzzz Looks good to me

wshager:
The use case should eventually be expanded / changed into authentication services such as OpenAthens or Shibboleth. This code could serve as a basis.

wshager:
To be more generic: SAML auth. Any thoughts on this?

dizzzz:
These are all about exchanging security data, no? not about actual authentication/authorization mechanisms?

wshager:
@dizzzz right. I got that a bit mixed up in the IP range check already, but that's not a problem imo. We just need to make sure this doesn't compromise db security, but since no passwords are passed and dba is disallowed it should be OK. Don't try this at home folks...

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
shabanovd: then code above must be used for proper integration with jetty authentication mechanism

shabanovd:
@wshager SAML realm is [here](https://github.com/shabanovd/exist/tree/2.1.0-RC90/extensions/security/saml)

wshager:
@shabanovd I'd give SAML priority over this, as it became clear to me IP based authentication should indeed be authorization. The reason I chose this approach is because there's no model for authorization in eXist AFAIK. I especially need to authorize multiple levels in document/collection hierarchy and find a check for this on every request a bad idea.

shabanovd:
@wshager do you want to close this PR?

wshager:
Not until there's a better solution. As this works and doesn't require a custom session based script, I still have a use case.

dizzzz:
So let's pull it in?

wshager:
Please

wshager:
Ignore that. Why would you? Design is wrong. Feature bloat. Bad priority.

shabanovd:
@dizzzz [this](https://github.com/eXist-db/exist/pull/1307#discussion-diff-103746084R101) wasn't address yet

dizzzz:
@shabanovd  it is not clear what you refer to, sorry
@wshager  you actually say... forget about it? So i spent many hours on... nothing?

wshager:
@dizzzz so have I, and learned from the experience. This needs another model, as you yourself already mentioned. Right now a check is performed on every request, because the relation between authenticated user and IP range is trivial. This seems to me bad design. What you want is to authorize a party to view (a group of) documents, not necessarily log them into the db. A session should suffice, but the problem is: how to handle read permissions on many documents? This realm abuses the authentication model for that purpose, and so IMO is bad design. In retrospect I should not have created it. At first I thought to appease you and others who invested time in this, but I honestly don't want to be responsible for making a bad decision.

shabanovd:
@wshager I'm a bit confused by your comments, but understand your worries. For me this is only useful in case when db is not public (with public ip) and you need only one subnet to have `guest` access. Single IP authentication is bad design because it's easy to violate it (just change ip on computer manually).

@dizzzz proper integration with jetty authentication infrastructure is this [code](https://github.com/eXist-db/exist/blob/develop/extensions/security/oauth/src/org/exist/security/realm/oauth/ServiceFacebook.java#L118-L131)
```
Account principal = new SubjectAccreditedImpl((AbstractAccount) found, accessToken);
		
HttpSession session = request.getSession(true);

Subject subject = new Subject();

//*******************************************************
DefaultIdentityService _identityService = new DefaultIdentityService();
UserIdentity user = _identityService.newUserIdentity(subject, principal, new String[0]);
        
Authentication cached=new HttpSessionAuthentication(session, user);
session.setAttribute(HttpSessionAuthentication.__J_AUTHENTICATED, cached);
//*******************************************************
```
OAuth, OpenID and SAML realms use code above.

This code
```
session.setAttribute(XQueryContext.HTTP_SESSIONVAR_XMLDB_USER, user);
```
is hack that eXist have because of no integration with jetty.

dizzzz:
@shabanovd i'll try, check.
Since the extension is switched off by default 9safe!), but provides a solution for a usecase and is/was used by @wshager  I thought the project would benefit from it.

wshager:
@dizzzz My client will benefit (as long as no IP is spoofed, but that doesn't seem to be a major concern). I believe you're on the payroll, so no harm done. The code will be used as long as there's no better alternative, but I don't think it should be merged. We'll keep it in a separate branch. My apologies. When there's a similar use case with the same kind of laissez fair attitude you could reconsider.

adamretter:
@dizzzz It is not clear whether this should be merged or closed?

dizzzz:
@adamretter  my answer is 2 fold.....
- i need to apply changes as requested by @shabanovd 
- @wshager actually uses this code, but is happy with his separate branche. personally I think we could/should pull it in, it has added value and... it does not harm since it is switched off by default. And it will be a trigger to restructure the code a bit, so we can remove the 'start-the-servlet-code' from JettyStart.java.

adamretter:
@dizzzz Okay, let me know when you are ready...

wshager:
Thanks @dizzzz for pushing this.

adamretter:
@dizzzz @wshager @shabanovd Will be merged after version 3.1.1 is released.

dizzzz:
Can one pull-in this PR as agreed?

dizzzz:
@wshager

PR #1308, opened 2017-03-02
Write operations without lock and possible NPE fix

shabanovd:
### Description:
Address 2 bugs:
 - storeTempResource must hold write lock on collection
 - avoid NPE removeCollection (possible only when storage corrupted)

### Reference:
address bugfixes of #1296


dizzzz:
interesting.....

PR #1316, opened 2017-03-03
XSL stylesheet caching and compiling bugfixes

### LINKED ISSUE #1312: calls to doc() in an XSLTSerlvet context executed within the database throw URIResolver errors
### What is the problem
When called in an XSLT stylesheet, which is executed via XSLTServlet within the database, the `doc()` function fails with following error:

    SAX exception while transforming node: Exception thrown by URIResolver

Yet, the URI does resolve, and the same XSLT stylesheet is working without errors when called with the `transform:transform()` function, and when executed via XSLTServlet on the file system.

Moreover, **execution order** seems to matter (see below).

### What did you expect

In the sample app (see below), I'd expect the call to `doc('http://localhost:8080/exist/apps/doc-test/test.xml')` to be resolved properly, and to get following output from the simple XSLT stylesheet, irrespective of the execution context / order: 

    <xslt-output><h1>doc() has succeeded!</h1></xslt-output>

The zip file attached [doc-test.zip](https://github.com/eXist-db/exist/files/814514/doc-test.zip) contains a sample web application demonstrating the problem, which can be reproduced by following steps:

1. install the `doc-test-0.1.xar` file in the database
2. navigate to http://localhost:8080/exist/apps/doc-test/test-XSLTServlet/

This will cause an HTTP ERROR 500, stating that 

> An error occurred: SAX exception while transforming node: Exception thrown by URIResolver

Additional remarks: 

#### 1. execution context matters
The error only occurs when the XSLT transformation is executed via XSLTServlet within the database.

- If the same XSLT transformation is executed via the `transform:transform()`  function in an XQuery script stored in the database, it succeeds without errors. See http://localhost:8080/exist/apps/doc-test/test-transform/ for an example.
- If the same XSLT transformation is executed via XSLTServlet on the file system, it succeeds without errors. See http://localhost:8080/exist/doc-test/test-XSLTServlet/ for an example (this requires that the zip file is extracted in the `${exist.home}/webapps/` directory.

#### 2. execution order matters
The error occurs when the XSLT transformation is executed via XSLTServlet within the database, but once the same transformation is executed via the `transform:transform()`  function in an XQuery script stored in the database, subsequent executions via XSLTServlet succeed without errors. See following steps:

1. Execute via XSLTServlet: http://localhost:8080/exist/doc-test/test-XSLTServlet/ --> error
2. Execute via `transform:transform()`: http://localhost:8080/exist/apps/doc-test/test-transform/ --> success
3. Execute via XSLTServlet: http://localhost:8080/exist/doc-test/test-XSLTServlet/ --> **success**

Since this error doesn't occur in eXist-3.0.RC1 (nor in eXist-3.0.RC2, although that version is giving me troubles with local paths to `doc()` that I couldn't reproduce, unfortunately), I suspect that it might be related to the issues that have been addressed recently in https://github.com/eXist-db/exist/pull/1221.

### Context information
- eXist-db version: 3.0 /  acd0c14
- Java: 1.8.0_112
- Operating system: Windows 7
- 64 bit
- No custom changes

shabanovd:
### Description:
inline xsl stylesheets caching and compiling between XSLTServlet and xquery `transform` function;
[bugfix] avoid concurrent use of different variables: DBBroker, ErrorListener, SAXTransformerFactory (and it attributes).

### Reference:
close #1312

### DIFF
--- a/extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
+++ b/extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
@@ -14,7 +14,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.xmlgraphics.io.ResourceResolver;
 import org.apache.xmlgraphics.io.URIResolverAdapter;
 import org.exist.storage.DBBroker;
-import org.exist.xquery.functions.transform.EXistURIResolver;
+import org.exist.xslt.EXistURIResolver;
 import org.exist.xquery.value.NodeValue;
 import org.xml.sax.ContentHandler;
 import org.xml.sax.SAXException;
@@ -126,7 +126,9 @@ public class ApacheFopProcessorAdapter implements ProcessorAdapter {
      */
     private ResourceResolver getResourceResolver(final DBBroker broker, final String baseUri) {
         final ResourceResolverFactory.SchemeAwareResourceResolverBuilder builder = ResourceResolverFactory.createSchemeAwareResourceResolverBuilder(ResourceResolverFactory.createDefaultResourceResolver());
-        final URIResolverAdapter uriResolver = new URIResolverAdapter(new ExistSchemeRewriter(new EXistURIResolver(broker, baseUri)));
+        final URIResolverAdapter uriResolver = new URIResolverAdapter(
+            new ExistSchemeRewriter(new EXistURIResolver(broker.getBrokerPool(), baseUri))
+        );
         builder.registerResourceResolverForScheme("exist", uriResolver);
         builder.registerResourceResolverForScheme("http", uriResolver);
         builder.registerResourceResolverForScheme("https", uriResolver);
--- a/src/org/exist/http/servlets/XSLTServlet.java
+++ b/src/org/exist/http/servlets/XSLTServlet.java
@@ -1,23 +1,21 @@
 /*
- *  eXist Open Source Native XML Database
- *  Copyright (C) 2010 The eXist Project
- *  http://exist-db.org
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2017 The eXist Project
+ * http://exist-db.org
  *
- *  This program is free software; you can redistribute it and/or
- *  modify it under the terms of the GNU Lesser General Public License
- *  as published by the Free Software Foundation; either version 2
- *  of the License, or (at your option) any later version.
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
  *
- *  This program is distributed in the hope that it will be useful,
- *  but WITHOUT ANY WARRANTY; without even the implied warranty of
- *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- *  GNU Lesser General Public License for more details.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
  *
- *  You should have received a copy of the GNU Lesser General Public
- *  License along with this library; if not, write to the Free Software
- *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
- *
- * $Id$
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  */
 package org.exist.http.servlets;
 
@@ -25,16 +23,11 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import org.exist.EXistException;
-import org.exist.collections.Collection;
-import org.exist.dom.persistent.DocumentImpl;
 import org.exist.security.AuthenticationException;
-import org.exist.security.Permission;
-import org.exist.security.PermissionDeniedException;
 import org.exist.security.Subject;
 import org.exist.security.internal.web.HttpAccount;
 import org.exist.storage.BrokerPool;
 import org.exist.storage.DBBroker;
-import org.exist.storage.lock.Lock.LockMode;
 import org.exist.storage.serializers.Serializer;
 import org.exist.storage.serializers.XIncludeFilter;
 import org.exist.util.serializer.Receiver;
@@ -49,8 +42,11 @@ import org.exist.xquery.value.Item;
 import org.exist.xquery.value.NodeValue;
 import org.exist.xquery.value.Type;
 import org.exist.xquery.value.ValueSequence;
+import org.exist.xslt.Stylesheet;
+import org.exist.xslt.TemplatesFactory;
 import org.exist.xslt.TransformerFactoryAllocator;
 
+import org.exist.xslt.XSLTErrorsListener;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 import org.xml.sax.SAXParseException;
@@ -61,25 +57,16 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-import javax.xml.transform.Source;
-import javax.xml.transform.Templates;
 import javax.xml.transform.Transformer;
-import javax.xml.transform.TransformerConfigurationException;
 import javax.xml.transform.TransformerException;
-import javax.xml.transform.URIResolver;
-import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.sax.SAXResult;
-import javax.xml.transform.sax.SAXTransformerFactory;
-import javax.xml.transform.sax.TemplatesHandler;
 import javax.xml.transform.sax.TransformerHandler;
-import javax.xml.transform.stream.StreamSource;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Writer;
-import java.net.*;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -104,9 +91,17 @@ public class XSLTServlet extends HttpServlet {
 
     private final static Logger LOG = LogManager.getLogger(XSLTServlet.class);
 
+    private final static XSLTErrorsListener<ServletException> errorListener =
+        new XSLTErrorsListener<ServletException>(true, false) {
+
+            @Override
+            protected void raiseError(String error, Exception ex) throws ServletException {
+                throw new ServletException(error, ex);
+            }
+        };
+
     private BrokerPool pool;
 
-    private final Map<String, CachedStylesheet> cache = new HashMap<String, CachedStylesheet>();
     private Boolean caching = null;
 
     /**
@@ -124,12 +119,11 @@ public class XSLTServlet extends HttpServlet {
         return caching;
     }
 
-
     @Override
     protected void doPost(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
 
-        final String stylesheet = (String) request.getAttribute(REQ_ATTRIBUTE_STYLESHEET);
-        if (stylesheet == null) {
+        final String uri = (String) request.getAttribute(REQ_ATTRIBUTE_STYLESHEET);
+        if (uri == null) {
             throw new ServletException("No stylesheet source specified!");
         }
 
@@ -155,7 +149,9 @@ public class XSLTServlet extends HttpServlet {
                                 sourceAttrib);
                     }
 
-                    LOG.debug("Taking XSLT input from request attribute " + sourceAttrib);
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Taking XSLT input from request attribute " + sourceAttrib);
+                    }
 
                 } else {
                     throw new ServletException("Input for XSLT servlet is not a node. Read from attribute " +
@@ -190,16 +186,15 @@ public class XSLTServlet extends HttpServlet {
             }
         }
 
-        final SAXTransformerFactory factory = TransformerFactoryAllocator.getTransformerFactory(pool);
-        final Templates templates = getSource(user, request, response, factory, stylesheet);
-        if (templates == null) {
+        final Stylesheet stylesheet = stylesheet(uri, request, response);
+        if (stylesheet == null) {
             return;
         }
 
         //do the transformation
         try (final DBBroker broker = pool.get(Optional.of(user))) {
 
-            final TransformerHandler handler = factory.newTransformerHandler(templates);
+            final TransformerHandler handler = stylesheet.newTransformerHandler(broker, errorListener);
             setTransformerParameters(request, handler.getTransformer());
 
             final Properties properties = handler.getTransformer().getOutputProperties();
@@ -213,12 +208,8 @@ public class XSLTServlet extends HttpServlet {
 
             final String mediaType = properties.getProperty("media-type");
             if (mediaType != null) {
-                if (encoding == null) {
-                    response.setContentType(mediaType);
-                }
-
                 //check, do mediaType have "charset"
-                else if (mediaType.indexOf("charset") == -1) {
+                if (!mediaType.contains("charset")) {
                     response.setContentType(mediaType + "; charset=" + encoding);
                 } else {
                     response.setContentType(mediaType);
@@ -239,18 +230,17 @@ public class XSLTServlet extends HttpServlet {
             try {
                 XIncludeFilter xinclude = new XIncludeFilter(serializer, receiver);
                 receiver = xinclude;
-                String moduleLoadPath;
 
+                String baseUri;
                 final String base = (String) request.getAttribute(REQ_ATTRIBUTE_BASE);
                 if (base != null) {
-                    moduleLoadPath = getServletContext().getRealPath(base);
-                } else if (stylesheet.startsWith("xmldb:exist://")) {
-                    moduleLoadPath = XmldbURI.xmldbUriFor(stylesheet).getCollectionPath();
+                    baseUri = getServletContext().getRealPath(base);
+                } else if (uri.startsWith("xmldb:exist://")) {
+                    baseUri = XmldbURI.xmldbUriFor(uri).getCollectionPath();
                 } else {
-                    moduleLoadPath = getCurrentDir(request).toAbsolutePath().toString();
+                    baseUri = getCurrentDir(request).toAbsolutePath().toString();
                 }
-
-                xinclude.setModuleLoadPath(moduleLoadPath);
+                xinclude.setModuleLoadPath(baseUri);
 
                 serializer.setReceiver(receiver);
                 if (inputNode != null) {
@@ -306,9 +296,8 @@ public class XSLTServlet extends HttpServlet {
     /*
      * Please add comments to this method. make assumption clear. These might not be valid.
      */
-    private Templates getSource(Subject user, HttpServletRequest request, HttpServletResponse response,
-                                SAXTransformerFactory factory, String stylesheet)
-            throws ServletException, IOException {
+    private Stylesheet stylesheet(String stylesheet, HttpServletRequest request, HttpServletResponse response)
+            throws IOException {
 
         // Check if stylesheet contains an URI. If not, try to resolve from file system
         if (stylesheet.indexOf(':') == Constants.STRING_NOT_FOUND) {
@@ -348,26 +337,7 @@ public class XSLTServlet extends HttpServlet {
             }
         }
 
-        // Try to figure out the base directory of the stylesheet file.
-        // This is required to locate resources imported within the stylesheet.
-        String base;
-        final int p = stylesheet.lastIndexOf("/");
-        if (p != Constants.STRING_NOT_FOUND) {
-            base = stylesheet.substring(0, p);
-        } else {
-            base = stylesheet;
-        }
-
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("Loading stylesheet from " + stylesheet);
-        }
-
-        CachedStylesheet cached = cache.get(stylesheet);
-        if (cached == null) {
-            cached = new CachedStylesheet(factory, user, stylesheet, base);
-            cache.put(stylesheet, cached);
-        }
-        return cached.getTemplates(user);
+        return TemplatesFactory.stylesheet(stylesheet, "", isCaching());
     }
 
     /*
@@ -429,187 +399,6 @@ public class XSLTServlet extends HttpServlet {
                     properties.setProperty(name.substring(REQ_ATTRIBUTE_OUTPUT.length()), value.toString());
                 }
             }
-        }
-    }
-
-    private class CachedStylesheet {
-
-        SAXTransformerFactory factory;
-        long lastModified = -1;
-        Templates templates = null;
-        String uri;
-
-        public CachedStylesheet(SAXTransformerFactory factory, Subject user, String uri, String baseURI) throws ServletException {
-            this.factory = factory;
-            this.uri = uri;
-            if (!baseURI.startsWith("xmldb:exist://")) {
-                factory.setURIResolver(new ExternalResolver(baseURI));
-            }
-            getTemplates(user);
-        }
-
-        public Templates getTemplates(Subject user) throws ServletException {
-            if (uri.startsWith("xmldb:exist://")) {
-                final String docPath = uri.substring("xmldb:exist://".length());
-
-
-                try (final DBBroker broker = pool.get(Optional.of(user))) {
-                    DocumentImpl doc = null;
-                    try {
-                        doc = broker.getXMLResource(XmldbURI.create(docPath), LockMode.READ_LOCK);
-                        if (doc == null) {
-                            throw new ServletException("Stylesheet not found: " + docPath);
-                        }
-
-                        if (!isCaching() || (doc != null && (templates == null
-                                || doc.getMetadata().getLastModified() > lastModified))) {
-                            templates = getSource(broker, doc);
-                        }
-
-                        lastModified = doc.getMetadata().getLastModified();
-                    } finally {
-                        if (doc != null) {
-                            doc.getUpdateLock().release(LockMode.READ_LOCK);
-                        }
-                    }
-                } catch (final PermissionDeniedException e) {
-                    throw new ServletException("Permission denied to read stylesheet: " + uri, e);
-
-                } catch (final EXistException e) {
-                    throw new ServletException("Error while reading stylesheet source from db: " + e.getMessage(), e);
-
-                }
-            } else {
-                try {
-                    final URL url = new URL(uri);
-                    final URLConnection connection = url.openConnection();
-                    long modified = connection.getLastModified();
-
-                    if (!isCaching() || (templates == null || modified > lastModified || modified == 0)) {
-                        LOG.debug("compiling stylesheet " + url.toString());
-                        templates = factory.newTemplates(new StreamSource(connection.getInputStream()));
-                    }
-                    lastModified = modified;
-
-                } catch (final IOException e) {
-                    throw new ServletException("Error while reading stylesheet source from uri: " + uri +
-                            ": " + e.getMessage(), e);
-
-                } catch (final TransformerConfigurationException e) {
-                    throw new ServletException("Error while reading stylesheet source from uri: " + uri +
-                            ": " + e.getMessage(), e);
-                }
-            }
-            return templates;
-        }
-
-        private Templates getSource(DBBroker broker, DocumentImpl stylesheet)
-                throws ServletException {
-            factory.setURIResolver(new DatabaseResolver(broker, stylesheet));
-            try {
-                final TemplatesHandler handler = factory.newTemplatesHandler();
-                handler.startDocument();
-
-                final Serializer serializer = broker.getSerializer();
-                serializer.reset();
-                serializer.setSAXHandlers(handler, null);
-                serializer.toSAX(stylesheet);
-
-                handler.endDocument();
-                return handler.getTemplates();
-
-            } catch (final SAXException e) {
-                throw new ServletException("A SAX exception occurred while compiling the stylesheet: "
-                        + e.getMessage(), e);
-
-            } catch (final TransformerConfigurationException e) {
-                throw new ServletException("A configuration exception occurred while " +
-                        "compiling the stylesheet: " + e.getMessage(), e);
-            }
-        }
-    }
-
-    /*
-     * TODO: create generic resolver for whole database
-     */
-    private static class ExternalResolver implements URIResolver {
-
-        private String baseURI;
-
-        public ExternalResolver(String base) {
-            this.baseURI = base;
-        }
-
-        /* (non-Javadoc)
-         * @see javax.xml.transform.URIResolver#resolve(java.lang.String, java.lang.String)
-         */
-        @Override
-        public Source resolve(String href, String base)
-                throws TransformerException {
-            URL url;
-            try {
-                //TODO : use dedicated function in XmldbURI
-                url = new URL(baseURI + "/" + href);
-                final URLConnection connection = url.openConnection();
-                return new StreamSource(connection.getInputStream());
-
-            } catch (final MalformedURLException e) {
-                return null;
-
-            } catch (final IOException e) {
-                return null;
-            }
-        }
-    }
-
-    /*
-     * TODO: create generic resolver for whole database
-     */
-    private static class DatabaseResolver implements URIResolver {
-
-        DocumentImpl doc;
-        DBBroker broker;
-
-        public DatabaseResolver(DBBroker broker, DocumentImpl myDoc) {
-            this.broker = broker;
-            this.doc = myDoc;
-        }
-
-
-        /* (non-Javadoc)
-         * @see javax.xml.transform.URIResolver#resolve(java.lang.String, java.lang.String)
-         */
-        @Override
-        public Source resolve(String href, String base)
-                throws TransformerException {
-            final Collection collection = doc.getCollection();
-            String path;
-
-            //TODO : use dedicated function in XmldbURI
-            if (href.startsWith("/")) {
-                path = href;
-            } else {
-                path = collection.getURI() + "/" + href;
-            }
-
-            DocumentImpl xslDoc;
-            try {
-                xslDoc = (DocumentImpl) broker.getXMLResource(XmldbURI.create(path));
-            } catch (final PermissionDeniedException e) {
-                throw new TransformerException(e.getMessage(), e);
-            }
-
-            if (xslDoc == null) {
-                LOG.debug("Document " + href + " not found in collection " + collection.getURI());
-                return null;
-            }
-
-            if (!xslDoc.getPermissions().validate(broker.getCurrentSubject(), Permission.READ)) {
-                throw new TransformerException("Insufficient privileges to read resource " + path);
-            }
-
-            final DOMSource source = new DOMSource(xslDoc);
-            return source;
         }
     }
 }
--- a/src/org/exist/xquery/functions/transform/EXistURIResolver.java
+++ b/src/org/exist/xquery/functions/transform/EXistURIResolver.java
@@ -43,6 +43,8 @@ import java.nio.file.Path;
 /**
  * Implementation of URIResolver which
  * will resolve paths from the eXist database
+ *
+ * @Deprecated use org.exist.xslt.EXistURIResolver
  */
 public class EXistURIResolver implements URIResolver {
     private static final Logger LOG = LogManager.getLogger(EXistURIResolver.class);
--- a/src/org/exist/xquery/functions/transform/EXistURIResolver.java
+++ b/src/org/exist/xquery/functions/transform/EXistURIResolver.java
@@ -44,7 +44,7 @@ import java.nio.file.Path;
  * Implementation of URIResolver which
  * will resolve paths from the eXist database
  *
- * @Deprecated use org.exist.xslt.EXistURIResolver
+ * @Deprecated use {@link org.exist.xslt.EXistURIResolver}
  */
 public class EXistURIResolver implements URIResolver {
     private static final Logger LOG = LogManager.getLogger(EXistURIResolver.class);
--- a/src/org/exist/xquery/functions/transform/Transform.java
+++ b/src/org/exist/xquery/functions/transform/Transform.java
@@ -1,23 +1,21 @@
 /*
- *  eXist Open Source Native XML Database
- *  Copyright (C) 2001-2011 The eXist Project
- *  http://exist-db.org
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2017 The eXist Project
+ * http://exist-db.org
  *
- *  This program is free software; you can redistribute it and/or
- *  modify it under the terms of the GNU Lesser General Public License
- *  as published by the Free Software Foundation; either version 2
- *  of the License, or (at your option) any later version.
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
  *
- *  This program is distributed in the hope that it will be useful,
- *  but WITHOUT ANY WARRANTY; without even the implied warranty of
- *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- *  GNU Lesser General Public License for more details.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
  *
- *  You should have received a copy of the GNU Lesser General Public
- *  License along with this library; if not, write to the Free Software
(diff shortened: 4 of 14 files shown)


dizzzz:
wow..... that is a fast fix!

dizzzz:
```
BUILD FAILED
/home/travis/build/eXist-db/exist/build/scripts/junit.xml:124: Test org.exist.xslt.TransfomerFactoryAllocatorTest failed
```

shabanovd:
@dizzzz that is huge fix for sure. There was so many concurrent bugs and code copies. That's why I decided to refactor it ;-)

shabanovd:
@dizzzz fixed

src/org/exist/xquery/functions/transform/EXistURIResolver.java
```diff
@@ -43,134 +43,136 @@
...
-        }
+    final int numPathComponents = Array.getLength(pathComponents);
+    final String[] simplifiedComponents = new String[numPathComponents];
+    int numSimplifiedComponents = 0;
 
-        if (path.endsWith("/")) {
-            b.append("/");
-        }
+    for(final String s : pathComponents) {
+      if (s.length() == 0) {continue;}		// Remove empty elements ("//")
```
dizzzz: s.isEmpty()

src/org/exist/xquery/functions/transform/EXistURIResolver.java
```diff
@@ -43,134 +43,136 @@
...
+      if (".".equals(s)) {continue;}		// Remove identity elements ("/./")
+      if ("..".equals(s)) {				// Remove parent elements ("/../") unless at the root
+        if (numSimplifiedComponents > 0) {numSimplifiedComponents--;}
```
dizzzz: reformat

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        final String[] pathComponents = path.substring(1).split("/");
+
+        final int numPathComponents = Array.getLength(pathComponents);
+        final String[] simplifiedComponents = new String[numPathComponents];
+        int numSimplifiedComponents = 0;
+
+        for(final String s : pathComponents) {
+            if (s.length() == 0) {continue;}		// Remove empty elements ("//")
```
dizzzz: double code; s.isEMpty()

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,192 @@
...
+
+        if (href.isEmpty()) {
+            path = base;
+        } else {
+            URI hrefURI = null;
+            try {
+                hrefURI = new URI(href);
+            } catch (final URISyntaxException e) {
+            }
+            if (hrefURI != null && hrefURI.isAbsolute()) {
```
dizzzz: double code?

src/org/exist/xslt/StylesheetUri.java
```diff
@@ -0,0 +1,189 @@
...
+  String uri;
+  String base;
+
+  Properties properties;
+
+  public StylesheetUri(String uri) {
+    this.uri = uri;
+
+    final int p = uri.lastIndexOf("/");
+    if (p != Constants.STRING_NOT_FOUND) {
```
dizzzz: you can use Java6' String.contains()

src/org/exist/xslt/StylesheetUri.java
```diff
@@ -0,0 +1,189 @@
...
+
+    return handler;
+  }
+
+  private <E extends Exception> Templates compileTemplates(
+      DBBroker broker,
+      DocumentImpl stylesheet,
+      XSLTErrorsListener<E> errorListener)
+      throws E, TransformerConfigurationException, SAXException
+  {
```
dizzzz: please reformat

src/org/exist/xslt/StylesheetUri.java
```diff
@@ -0,0 +1,189 @@
...
+    return t;
+  }
+
+  private SAXTransformerFactory cachedFactory(BrokerPool db) {
+    if (factory == null) {
+      factory = TransformerFactoryAllocator.getTransformerFactory(db);
+
+      if (properties != null) {
+        //set any attributes
+        for (final Map.Entry<Object, Object> attribute : properties.entrySet()) {
```
dizzzz: candidate for lambda expression

src/org/exist/xslt/TemplatesFactory.java
```diff
@@ -0,0 +1,151 @@
...
+          uri = f.toUri().toASCIIString();
+        }
+      }
+    }
+    return uri;
+  }
+
+  public static <E extends Exception> Stylesheet stylesheet(DBBroker broker, NodeValue node, String baseUri)
+      throws E, TransformerConfigurationException
+  {
```
dizzzz: reformat

src/org/exist/xslt/XSLTErrorsListener.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.logging.log4j.Logger;
+
+/**
+ * @author <a href="mailto:shabanovd@gmail.com">Dmitriy Shabanov</a>
+ */
+public abstract class XSLTErrorsListener<E extends Exception> implements ErrorListener {
+
+  private final static Logger LOG = LogManager.getLogger(XSLTErrorsListener.class);
+
+  private final static int NO_ERROR = 0;
```
dizzzz: would it make sense to convert this to an enum?

src/org/exist/xquery/functions/transform/EXistURIResolver.java
```diff
@@ -43,134 +43,136 @@
...
 
-        return b.toString();
+    if (numSimplifiedComponents == 0) {
+      return "/";
     }
 
-    @Override
-    public Source resolve(final String href, String base) throws TransformerException {
-        String path;
+    final StringBuilder b = new StringBuilder(path.length());
```
dizzzz: b .... maybe a descriptive name of what will be in the buffer?

src/org/exist/xquery/functions/transform/EXistURIResolver.java
shabanovd: EXistURIResolver is old code

src/org/exist/xslt/EXistURIResolver.java
shabanovd: EXistURIResolver is old code

src/org/exist/xslt/EXistURIResolver.java
shabanovd: EXistURIResolver is old code

src/org/exist/xquery/functions/transform/EXistURIResolver.java
shabanovd: that is old code that I put back as `deprecated`

src/org/exist/xquery/functions/transform/EXistURIResolver.java
shabanovd: that is old code that I put back as `deprecated`

src/org/exist/xslt/XSLTErrorsListener.java
shabanovd: that is for internal use only

src/org/exist/xquery/functions/transform/EXistURIResolver.java
shabanovd: that is old code, rename to`sb` if you like?

src/org/exist/xslt/TemplatesFactory.java
shabanovd: I did, that is formatted code. What do you think wrong?

shabanovd:
@dizzzz that is not duplicate code (pay attention to commits description -)
The duplicate code required to make it to 3.1.0, without that `deprecated` class it must go to 4.0.0 only.

src/org/exist/xslt/StylesheetUri.java
shabanovd: That is not possible because `p` required for `uri.substring(0, p);`

src/org/exist/xslt/StylesheetUri.java
shabanovd: why?

shabanovd:
@dizzzz EXistURIResolver reformatted

src/org/exist/xslt/StylesheetUri.java
dizzzz: no specific reason :-)

src/org/exist/xslt/TemplatesFactory.java
dizzzz: the { is on the wrong line?

extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
```diff
@@ -126,7 +126,9 @@ private FOUserAgent setupFOUserAgent(final FOUserAgent foUserAgent, final Proper
      */
     private ResourceResolver getResourceResolver(final DBBroker broker, final String baseUri) {
         final ResourceResolverFactory.SchemeAwareResourceResolverBuilder builder = ResourceResolverFactory.createSchemeAwareResourceResolverBuilder(ResourceResolverFactory.createDefaultResourceResolver());
-        final URIResolverAdapter uriResolver = new URIResolverAdapter(new ExistSchemeRewriter(new EXistURIResolver(broker, baseUri)));
+        final URIResolverAdapter uriResolver = new URIResolverAdapter(
+            new ExistSchemeRewriter(new EXistURIResolver(broker.getBrokerPool(), baseUri))
+        );
         builder.registerResourceResolverForScheme("exist", uriResolver);
```
dizzzz: xmldb: ??

shabanovd: that is not my code or code this PR. If you want to address this issue open new PR for it

shabanovd:
@dizzzz there is already [test](https://github.com/eXist-db/exist/pull/1316/files#diff-ede73efa2260d94d3434c2b656b4ab5fR36) for it

extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
dizzzz: sorry my remark is a bit lower: 'exist' is registered, should this be 'xmldb' ? or both?

shabanovd: @dizzzz better to ask @adamretter why it here at all. He did point [here](https://github.com/eXist-db/exist/pull/1340) that we did/should not use `exist:///`, so this must cleaned up.

If you want it to be fixed open issue for it (to remember).

src/org/exist/xslt/TemplatesFactory.java
```diff
@@ -0,0 +1,151 @@
...
+
+    return new StylesheetUri(uri, properties);
+  }
+
+  private static Stylesheet stylesheet(String stylesheet, String baseUri) {
+
+    String uri = uri(stylesheet, baseUri);
+
+    cache.computeIfAbsent(uri, StylesheetUri::new);
+    return cache.get(uri);
```
adamretter: Just `return cache.computeIfAbsent(uri, StylesheetUri::new);`

src/org/exist/xslt/StylesheetUri.java
```diff
@@ -0,0 +1,189 @@
...
+
+    final Templates t = handler.getTemplates();
+
+    //check for errors
+    errorListener.checkForErrors();
+
+    return t;
+  }
+
+  private SAXTransformerFactory cachedFactory(BrokerPool db) {
```
adamretter: It seems to me this is used concurrently from `TemplatesFactory` so you will need synchronization on the cached `factory` and the mutable class members.

src/org/exist/xslt/TemplatesFactory.java
```diff
@@ -0,0 +1,151 @@
...
+import org.exist.security.PermissionDeniedException;
+import org.exist.storage.DBBroker;
+import org.exist.xquery.Constants;
+import org.exist.xquery.value.NodeValue;
+import org.w3c.dom.Document;
+import org.xml.sax.SAXException;
+
+/**
+ * @author <a href="mailto:shabanovd@gmail.com">Dmitriy Shabanov</a>
+ */
```
adamretter: Please add the `@ThreadSafe` JCIP annotation if your intention is that this class is thread safe

src/org/exist/xslt/TemplatesFactory.java
```diff
@@ -0,0 +1,151 @@
...
+      if (uri != null) {
+        base = uri.substring(0, uri.lastIndexOf('/'));
+      }
+    }
+
+    if (base != null) {
+      factory.setURIResolver(new EXistURIResolver(broker.getBrokerPool(), base));
+    }
+
+    return new Stylesheet() {
```
adamretter: I don't like this overriding the class inline. I think you should create a proper sub-class

src/org/exist/xslt/TransformerFactoryAllocator.java
```diff
@@ -75,21 +73,6 @@ private TransformerFactoryAllocator() {
      * TransformerFactoryAllocator.getTransformerFactory(broker).newInstance()
      */
     public static SAXTransformerFactory getTransformerFactory(final BrokerPool pool) {
```
adamretter: Why can't we reuse a single TransformerFactory instance?

src/org/exist/xslt/XSLTErrorsListener.java
adamretter: Even for private use an Enum has advantages.

src/org/exist/xslt/XSLTErrorsListener.java
```diff
@@ -0,0 +1,93 @@
...
+  private final static int WARNING = 1;
+  private final static int ERROR = 2;
+  private final static int FATAL = 3;
+
+  boolean stopOnError;
+  boolean stopOnWarn;
```
adamretter: These should not be `package private`. Also they should be `final`.

src/org/exist/xslt/XSLTErrorsListener.java
```diff
@@ -0,0 +1,93 @@
...
+
+  private int errorCode = NO_ERROR;
+  private Exception exception;
+
+  public XSLTErrorsListener(boolean stopOnError, boolean stopOnWarn) {
```
adamretter: `final` arguments please

src/org/exist/xslt/XSLTErrorsListener.java
```diff
@@ -0,0 +1,93 @@
...
+    LOG.warn("XSL transform reports recoverable error: " + except.getMessage(), except);
+    errorCode = ERROR;
+    exception = except;
+    if (stopOnError) {
+      throw except;
+    }
+  }
+
+  public void fatalError(TransformerException except) throws TransformerException {
+    LOG.warn("XSL transform reports fatal error: " + except.getMessage(), except);
```
adamretter: Should be `LOG.error`

src/org/exist/xslt/XSLTErrorsListener.java
```diff
@@ -0,0 +1,93 @@
...
+    LOG.warn("XSL transform reports warning: " + except.getMessage(), except);
+    errorCode = WARNING;
+    exception = except;
+    if (stopOnWarn) {
+      throw except;
+    }
+  }
+
+  public void error(TransformerException except) throws TransformerException {
+    LOG.warn("XSL transform reports recoverable error: " + except.getMessage(), except);
```
adamretter: Should be `LOG.error`

src/org/exist/http/servlets/XSLTServlet.java
```diff
@@ -104,9 +91,17 @@
 
     private final static Logger LOG = LogManager.getLogger(XSLTServlet.class);
 
+    private final static XSLTErrorsListener<ServletException> errorListener =
+        new XSLTErrorsListener<ServletException>(true, false) {
+
+            @Override
+            protected void raiseError(String error, Exception ex) throws ServletException {
```
adamretter: `final` arguments please

src/org/exist/http/servlets/XSLTServlet.java
```diff
@@ -104,9 +91,17 @@
 
     private final static Logger LOG = LogManager.getLogger(XSLTServlet.class);
 
+    private final static XSLTErrorsListener<ServletException> errorListener =
+        new XSLTErrorsListener<ServletException>(true, false) {
```
adamretter: Diamond syntax please

src/org/exist/http/servlets/XSLTServlet.java
```diff
@@ -306,9 +296,8 @@ else if (mediaType.indexOf("charset") == -1) {
     /*
      * Please add comments to this method. make assumption clear. These might not be valid.
      */
-    private Templates getSource(Subject user, HttpServletRequest request, HttpServletResponse response,
-                                SAXTransformerFactory factory, String stylesheet)
-            throws ServletException, IOException {
+    private Stylesheet stylesheet(String stylesheet, HttpServletRequest request, HttpServletResponse response)
```
adamretter: `final` arguments please

src/org/exist/http/servlets/XSLTServlet.java
```diff
@@ -348,26 +337,7 @@ private Templates getSource(Subject user, HttpServletRequest request, HttpServle
...
-            LOG.debug("Loading stylesheet from " + stylesheet);
-        }
-
-        CachedStylesheet cached = cache.get(stylesheet);
-        if (cached == null) {
-            cached = new CachedStylesheet(factory, user, stylesheet, base);
-            cache.put(stylesheet, cached);
-        }
-        return cached.getTemplates(user);
+        return TemplatesFactory.stylesheet(stylesheet, "", isCaching());
```
adamretter: I think you should use `null` and not an empty string and then add `@Nullable` to the method argument.

src/org/exist/xquery/functions/transform/Transform.java
```diff
@@ -208,10 +190,16 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
                 "yes".equals(serializationProps.getProperty(EXistOutputKeys.EXPAND_XINCLUDES, "yes"));
 
 
+        final XSLTErrorsListener<XPathException> errorListener =
+            new XSLTErrorsListener<XPathException>(stopOnError, stopOnWarn) {
+                @Override
+                protected void raiseError(String error, Exception ex) throws XPathException {
```
adamretter: `final` arguments please

src/org/exist/xquery/functions/transform/Transform.java
```diff
@@ -360,71 +348,75 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
      * @throws TransformerFactoryConfigurationError
      * @throws XPathException
      */
-    private TransformerHandler createHandler(Item stylesheetItem, Properties options, Properties attributes) throws TransformerFactoryConfigurationError, XPathException {
-        final SAXTransformerFactory factory = TransformerFactoryAllocator.getTransformerFactory(context.getBroker().getBrokerPool());
-
-        //set any attributes
-        for (final Map.Entry<Object, Object> attribute : attributes.entrySet()) {
-            factory.setAttribute((String) attribute.getKey(), attribute.getValue());
+    private TransformerHandler createHandler(
```
adamretter: Fix the formatting here please

adamretter: Also `final` arguments please

src/org/exist/xquery/functions/transform/Transform.java
```diff
@@ -360,71 +348,75 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
...
+        XSLTErrorsListener<XPathException> errorListener
+    )
+        throws TransformerFactoryConfigurationError, XPathException
+    {
+
+        boolean useCache = true;
+        final Object property = context.getBroker().getConfiguration().getProperty(TransformerFactoryAllocator.PROPERTY_CACHING_ATTRIBUTE);
+        if (property != null) {
+            useCache = (Boolean) property;
         }
```
adamretter:
```
final Object property = context.getBroker().getConfiguration().getProperty(TransformerFactoryAllocator.PROPERTY_CACHING_ATTRIBUTE);
final boolean useCache = property != null && (Boolean)property;
```

src/org/exist/xquery/functions/transform/Transform.java
```diff
@@ -360,71 +348,75 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
...
-            handler = factory.newTransformerHandler(templates);
+
+            handler = stylesheet.newTransformerHandler(getContext().getBroker(), errorListener);
 
             if (options != null) {
                 setParameters(options, handler.getTransformer());
             }
 
-        } catch (final TransformerConfigurationException e) {
+        } catch (final Exception e) {
```
adamretter: I would prefer not to blanket catch `Exception`

extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
shabanovd: @adamretter @dizzzz that is exactly reason why I'm pushing `xmldb:/` (as developer independent protocol for in db uri/url reference)

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,201 @@
...
+
+  public EXistURIResolver(final BrokerPool db, final String docPath) {
+    this.db = db;
+    this.basePath = normalize(docPath);
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("EXistURIResolver base path set to " + basePath);
+    }
+  }
+
+  private String normalize(String uri) {
```
shabanovd: @adamretter I'm going to ignore your requests to use `final` because that is your coding style, not mine

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,201 @@
...
+   * Simplify a path removing any "." and ".." path elements.
+   * Assumes an absolute path is given.
+   */
+  private String normalizePath(final String path) {
+    if (!path.startsWith("/")) {
+      throw new IllegalArgumentException("normalizePath may only be applied to an absolute path; " +
+          "argument was: " + path + "; base: " + basePath);
+    }
+
+    final String[] pathComponents = path.substring(1).split("/");
```
shabanovd: then what `if (!path.startsWith("/")) {` do?

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,201 @@
...
+
+    final int numPathComponents = Array.getLength(pathComponents);
```
shabanovd: that is question to you because this code copy of org.exist.xquery.functions.transform.EXistURIResolver

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,201 @@
...
+    final String[] simplifiedComponents = new String[numPathComponents];
+    int numSimplifiedComponents = 0;
+
+    for (final String s : pathComponents) {
```
shabanovd: I don't think that `Paths` way is right

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,201 @@
...
+      }
+      simplifiedComponents[numSimplifiedComponents++] = s;
+    }
+
+    if (numSimplifiedComponents == 0) {
+      return "/";
+    }
+
+    final StringBuilder sb = new StringBuilder(path.length());
+    for (int x = 0; x < numSimplifiedComponents; x++) {
```
shabanovd: that is question to you because this code copy of org.exist.xquery.functions.transform.EXistURIResolver

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,201 @@
...
+      sb.append("/").append(simplifiedComponents[x]);
+    }
+
+    if (path.endsWith("/")) {
+      sb.append("/");
```
shabanovd: I don't think that is right

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,201 @@
...
+    String path;
+
+    if (href.isEmpty()) {
+      path = base;
+    } else {
+      URI hrefURI = null;
+      try {
+        hrefURI = new URI(href);
+      } catch (final URISyntaxException e) {
+      }
```
shabanovd: I have no idea why it's empty, code copy from org.exist.xquery.functions.transform.EXistURIResolver

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,201 @@
...
+      return new StreamSource(url.openStream());
+    } catch (final IOException e) {
+      throw new TransformerException(e.getMessage(), e);
+    }
+  }
+
+  private Source databaseSource(final String path) throws TransformerException {
+    final XmldbURI uri = XmldbURI.create(path);
+
+    final DBBroker broker = db.getActiveBroker();
```
shabanovd: why not? No needs to allocate broker because it must be already allocated for processing.

src/org/exist/xslt/EXistURIResolver.java
```diff
@@ -0,0 +1,201 @@
...
+    final DocumentImpl doc;
+    try {
+      doc = broker.getResource(uri, Permission.READ);
+      if (doc == null) {
+        LOG.error("Document " + path + " not found");
+        throw new TransformerException("Resource " + path + " not found in database.");
+      }
+
+      final Source source;
+      if (doc instanceof BinaryDocument) {
```
shabanovd: right, copy of org.exist.xquery.functions.transform.EXistURIResolver

src/org/exist/xslt/StylesheetUri.java
```diff
@@ -0,0 +1,189 @@
...
+import org.exist.xmldb.XmldbURI;
+import org.exist.xquery.Constants;
+import org.xml.sax.SAXException;
+
+/**
+ * @author <a href="mailto:shabanovd@gmail.com">Dmitriy Shabanov</a>
+ */
+public class StylesheetUri implements Stylesheet {
+
+  protected final static Logger LOG = LogManager.getLogger(StylesheetUri.class);
```
shabanovd: why not?

src/org/exist/xslt/StylesheetUri.java
```diff
@@ -0,0 +1,189 @@
...
+
+  SAXTransformerFactory factory;
```
shabanovd: why is it?

src/org/exist/xslt/StylesheetUri.java
```diff
@@ -0,0 +1,189 @@
...
+import org.exist.security.PermissionDeniedException;
+import org.exist.storage.BrokerPool;
+import org.exist.storage.DBBroker;
+import org.exist.storage.lock.Lock.LockMode;
+import org.exist.storage.serializers.Serializer;
+import org.exist.xmldb.XmldbURI;
+import org.exist.xquery.Constants;
+import org.xml.sax.SAXException;
+
+/**
```
shabanovd: done

src/org/exist/xslt/StylesheetUri.java
```diff
@@ -0,0 +1,189 @@
...
+    } else {
+      final URL url = new URL(uri);
+      final URLConnection connection = url.openConnection();
+      long modified = connection.getLastModified();
+      if (templates == null || modified > lastModified || modified == 0) {
+        if (LOG.isDebugEnabled()) {
+          LOG.debug("compiling stylesheet " + url);
+        }
+        try (final InputStream is = connection.getInputStream()) {
+          templates = factory(broker.getBrokerPool(), errorListener).newTemplates(new StreamSource(is));
```
shabanovd: because of `errorListener`

src/org/exist/xslt/StylesheetUri.java
```diff
@@ -0,0 +1,189 @@
...
+    return handler;
+  }
+
+  private <E extends Exception> Templates compileTemplates(
+      DBBroker broker,
+      DocumentImpl stylesheet,
+      XSLTErrorsListener<E> errorListener)
+      throws E, TransformerConfigurationException, SAXException
+  {
+    //factory.setURIResolver(new EXistURIResolver(broker, stylesheet.getCollection().getURI().toString()));
```
shabanovd: that is note

src/org/exist/xslt/TemplatesFactory.java
shabanovd: nope, because `computeIfAbsent ` return `the previous value associated with the specified key`

src/org/exist/xslt/StylesheetUri.java
shabanovd: @adamretter I think that you didn't understand where concurrent issue was and how it was fixed. The problem was at `SAXTransformerFactory` use. It was shared and methods `setURIResolver` & `setErrorListener` was used. That create number of concurrent issues including `execution order bug`.

`cachedFactory` used only at places where `setErrorListener` is not required and `factory` at places where `setErrorListener` is used (templates compiling).

shabanovd: I do not see any place where synchronization can be required. Can you point?

src/org/exist/xslt/TemplatesFactory.java
```diff
@@ -0,0 +1,151 @@
...
+import javax.xml.transform.sax.TemplatesHandler;
+import javax.xml.transform.sax.TransformerHandler;
+import org.exist.security.PermissionDeniedException;
+import org.exist.storage.DBBroker;
+import org.exist.xquery.Constants;
+import org.exist.xquery.value.NodeValue;
+import org.w3c.dom.Document;
+import org.xml.sax.SAXException;
+
+/**
```
shabanovd: done

src/org/exist/xslt/TemplatesFactory.java
shabanovd: done

src/org/exist/xslt/TemplatesFactory.java
```diff
@@ -0,0 +1,151 @@
...
+
+  private static String uri(String stylesheet, String baseUri) {
+    String uri = stylesheet;
+    if (stylesheet.indexOf(':') == Constants.STRING_NOT_FOUND) {
+      Path f = Paths.get(stylesheet).normalize();
+      if (Files.isReadable(f)) {
+        uri = f.toUri().toASCIIString();
+      } else {
+        stylesheet = baseUri + File.separatorChar + stylesheet;
+        f = Paths.get(stylesheet).normalize();
```
shabanovd: done

src/org/exist/xslt/TemplatesFactory.java
shabanovd: java support closure, why it can't be used?

src/org/exist/xslt/TransformerFactoryAllocator.java
shabanovd: see comment above

src/org/exist/http/servlets/XSLTServlet.java
shabanovd: can't because of `cannot use "<>" with anonymous inner classes`

src/org/exist/http/servlets/XSLTServlet.java
shabanovd: Nope, because base-uri must be "" (empty string)

src/org/exist/xquery/functions/transform/Transform.java
shabanovd: formatting is ok here

src/org/exist/xquery/functions/transform/Transform.java
shabanovd: I prefer 1st version of code because it clear that default is true

src/org/exist/xquery/functions/transform/Transform.java
shabanovd: why not?

extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
adamretter: Okay but I think this is a bigger issue that we should all agree. We should at one of our teleconferences have a read and build a proposal for a unified URL scheme that is compatible with java.net.URI and also how we can support backwards compatibility with `xmldb:exist:/` for a period of time.

src/org/exist/xslt/EXistURIResolver.java
adamretter: It appears that other people like @wolfgangmm have also adopted it. What do you see as the downside of using `final`?

src/org/exist/xslt/EXistURIResolver.java
adamretter: This will cause an NPE:
```java
final String path = "/";

...

if (!path.startsWith("/")) {
  final String[] pathComponents = path.substring(1).split("/");
}
```

src/org/exist/xslt/EXistURIResolver.java
adamretter: Okay, then it is a mistake. Can you please correct it?

src/org/exist/xslt/EXistURIResolver.java
adamretter: Why? `Paths` is tried and tested code that we don't have to maintain, and we know that it's normalize function will likely work for all cases.

src/org/exist/xslt/EXistURIResolver.java
adamretter: Okay, then it is a mistake. Can you please correct it?

src/org/exist/xslt/EXistURIResolver.java
adamretter: Why don't you think that is right?

src/org/exist/xslt/EXistURIResolver.java
adamretter: Okay, then it is a mistake. Can you please correct it?

src/org/exist/xslt/EXistURIResolver.java
adamretter: Sure, we should think about this in the future - https://github.com/eXist-db/exist/issues/1377

src/org/exist/xslt/TemplatesFactory.java
adamretter: It makes understanding the code harder. It is much nicer to have a clean separate subclass which has a javadoc description as to its purpose.

src/org/exist/xslt/TransformerFactoryAllocator.java
adamretter: Sorry I don't see any comments. What/where is the comment?

src/org/exist/http/servlets/XSLTServlet.java
adamretter: Ok

src/org/exist/http/servlets/XSLTServlet.java
adamretter: Just put a check in the top of `TemplatesFactory#stylesheet` which checks if null was supplied and uses an empty string. Passing an empty string seems like it will lead to mistakes in future.

src/org/exist/xquery/functions/transform/Transform.java
adamretter: Because it does not show what exceptions are under consideration. Also it will catch RuntimeException's which it should not.

src/org/exist/xslt/EXistURIResolver.java
shabanovd: It detracts from the understanding of the code.

src/org/exist/xslt/EXistURIResolver.java
shabanovd: your code is different from code at `EXistURIResolver`

src/org/exist/xslt/EXistURIResolver.java
shabanovd: Can it be at different PR?

src/org/exist/xslt/EXistURIResolver.java
shabanovd: because `Paths` will not play well with urls

src/org/exist/xslt/EXistURIResolver.java
shabanovd: why is that? It is ok to use any variable name, no?

src/org/exist/xslt/EXistURIResolver.java
shabanovd: because `sb.append('/');` will add `/` where it was not

src/org/exist/xslt/EXistURIResolver.java
shabanovd: that is not purpose of this PR

src/org/exist/xslt/TemplatesFactory.java
shabanovd: not for me, this way it's much clear for me

src/org/exist/http/servlets/XSLTServlet.java
shabanovd: nope, that is normal practice with base-uri, similar strategy have xmlcalabash for example

extensions/modules/src/org/exist/xquery/modules/xslfo/ApacheFopProcessorAdapter.java
shabanovd: I did not say that we must not support `xmldb:exist:/` but that we must start to support `xmldb:/`

src/org/exist/xquery/functions/transform/Transform.java
shabanovd: What is bad at catching NPE or ClassCastException and wrapped it into XPathException?

src/org/exist/xslt/TransformerFactoryAllocator.java
shabanovd: "Show outdated" your friend. Here the copy of it:

> @adamretter I think that you didn't understand where concurrent issue was and how it was fixed. The problem was at SAXTransformerFactory use. It was shared and methods setURIResolver & setErrorListener was used. That create number of concurrent issues including execution order bug.
> 
> cachedFactory used only at places where setErrorListener is not required and factory at places where setErrorListener is used (templates compiling).

src/org/exist/xquery/functions/transform/Transform.java
dizzzz: Catch exception/throwable .... IMO a good enough option if you don't trust a 3rd party library.  Saxon sometimes throws unpredictable exceptions.... due to our integration.

src/org/exist/xslt/TemplatesFactory.java
```diff
@@ -66,8 +66,7 @@ private static Stylesheet stylesheet(String stylesheet, String baseUri) {
 
     String uri = uri(stylesheet, baseUri);
 
-    cache.computeIfAbsent(uri, StylesheetResolverAndCompiler::new);
-    return cache.get(uri);
+    return cache.computeIfAbsent(uri, StylesheetResolverAndCompiler::new);
```
adamretter: :-)

src/org/exist/xslt/XSLTErrorsListener.java
alebastrov: I would propose to use enums everywhere instead of magic numbers

src/org/exist/xslt/EXistURIResolver.java
alebastrov: Use final only when it necessary or when there is a similar name in block of code

src/org/exist/xslt/EXistURIResolver.java
alebastrov: I would say it;s completely wrong
~~~
if (!path.startsWith("/")) {
  final String[] pathComponents = path.substring(1).split("/");
}
~~~

Maybe you meant inverted expression like
~~~
if (path.startsWith("/")) {
  final String[] pathComponents = path.substring(1).split("/");
}
~~~

shabanovd:
@wolfgangmm ping

shabanovd:
@adamretter ping

adamretter:
@shabanovd The decision for this was left with @wolfgangmm and @dizzzz from what I recall

shabanovd:
@wolfgangmm @dizzzz should I close this PR?

dizzzz:
@shabanovd  no, I don't think so; I remember in our Monday-telco that we agreed to have additional testcases here?

shabanovd:
@dizzzz no, we agree to merge it next day after call ... now it's 11 days and counting

adamretter:
@shabanovd My recollection of the call is different; I don't remember any actual agreement to merge.

I remember, that I personally didn't want to merge as the single-threaded test cases described by @rvdb https://github.com/eXist-db/exist/issues/1312#issue-211458721 are missing.

I remember @wolfgangmm saying something like "tomorrow, we will *think* about merging it".

My offer to merge once the test cases have been added, still stands; if you want it addressed more quickly than perhaps @wolfgangmm or @dizzzz can get to it.

shabanovd:
@adamretter I think that was very clear that I don't have time to spend on it and I think that it'll be lost because nobody from "core" developers pay attention to it.

To give clear message to end-user I'll close this PR in next week if it do not get merged. I don't want to confuse anyone.

dizzzz:
i'd need to recheck my notes but I remember there were 'threading' concerns, making us consider to atleast have some a few more tests. @shabanovd indicated not to have time for this anymore, can't recall the next step.

from my side:
- i see the added value
- this part of the code I can only perform generic code reviews - it is not my domain
- on threads/locks .... I am not not very accountable for this subject, so I cannot step in here too

so yes i'd like to see the code integrated, but  with some concerns addressed.

wolfgangmm:
I will merge this after discussing again with @joewiz  and @ljo because it obviously indeed fixes some real issues for several users. Still tests should be added in the future. @ljo expressed interest to add some.

PR #1323, opened 2017-03-05
Minor fixes

adamretter:
Minor unrelated fixes.

### DIFF
--- a/extensions/webdav/src/org/exist/webdav/ExistCollection.java
+++ b/extensions/webdav/src/org/exist/webdav/ExistCollection.java
@@ -443,16 +443,17 @@ public class ExistCollection extends ExistResource {
         }
 
         Collection srcCollection = null;
-        final LockMode srcCollectionLockMode = (mode == Mode.COPY ? LockMode.READ_LOCK : LockMode.WRITE_LOCK);
+        // This class contains already the URI of the resource that shall be moved/copied
+        final XmldbURI srcCollectionUri = xmldbUri;
+        // use WRITE_LOCK if moving or if src and dest collection are the same
+        final LockMode srcCollectionLockMode = mode == Mode.MOVE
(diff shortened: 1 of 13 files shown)


adamretter:
Please don't merge this until the current code-freeze is lifted ;-)

extensions/webdav/src/org/exist/webdav/ExistCollection.java
```diff
@@ -443,17 +443,18 @@ void resourceCopyMove(XmldbURI destCollectionUri, String newName, Mode mode) thr
         }
 
         Collection srcCollection = null;
-        final LockMode srcCollectionLockMode = (mode == Mode.COPY ? LockMode.READ_LOCK : LockMode.WRITE_LOCK);
+        // This class contains already the URI of the resource that shall be moved/copied
+        final XmldbURI srcCollectionUri = xmldbUri;
+        // use WRITE_LOCK if moving or if src and dest collection are the same
```
dizzzz: though not needed, you might want for readability write this as

`LockMode srcCollectionLockMode = (mode == Mode.MOVE
 +                || destCollectionUri.equals(srcCollectionUri)) ? LockMode.WRITE_LOCK : LockMode.READ_LOCK;`

joewiz:
Thanks @adamretter, for the improvements, and @dizzzz and @shabanovd for the review.

@adamretter For the new feature in this PR, "[feature] Allow temporary storage to be used for the database during test" - can you give an example of a test now possible that wasn't before?

adamretter:
@joewiz Two main things:

1) I needed it for a situation where I needed to be able to start the database and know absolutely that the database directory would be clean.
2) It opens a path to running tests in parallel, as each database could operate on a separate data directory; However if we want to run the Junit tests in non-forked mode, there are some singletons in the eXist code base that also need to be cleaned up first.

joewiz:
@adamretter Cool, thanks for the info! The possibility of running tests in parallel could be incredible for reducing the time to test a build, not to mention the extended test suite that we've been discussing.

PR #1324, opened 2017-03-06
Switch to nonblocking free ids processing algorithms

shabanovd:
### Description:
Free ids stack use locking that lead to deadlock. Switch to nonblocking stacks and remove locks.

### DIFF
--- a/src/org/exist/storage/index/CollectionStore.java
+++ b/src/org/exist/storage/index/CollectionStore.java
@@ -1,7 +1,21 @@
 /*
- * CollectionStore.java - Jun 19, 2003
- * 
- * @author wolf
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2017 The eXist Project
+ * http://exist-db.org
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


shabanovd:
@wolfgangmm this address your deadlock situation

adamretter:
Removing the locking for those items where it is not needed is a nice idea.

However if this is the cause of the deadlock seen by @wolfgangmm if we examine the code for the call-sites to `CollectionStore#getFreeCollectionId` and `CollectionStore#getFreeResourceId`  we can see this is not the root cause, and that this change would just move such a deadlock to `NativeBroker#getNextCollectionId` and `NativeBroker#getNextResourceId`.

Also looking at the existing code, it seems to me that there may be an existing race-condition concern around the inconsistent locking of the in-memory list of free ids in `CollectionStore` and the on-disk list in the `collections.dbx`.

shabanovd:
@adamretter the problem much bigger than your description above. This is one small step to make eXist-db lock free, because any 2 locks will have deadlock and only no lock design avoid it.

shabanovd:
@adamretter next problem that collection lock (write or read) required only for `documents`, `subCollections` or metadata modification or read. Good example of it [here](https://github.com/eXist-db/exist/blob/552be315ac6e7fb61fca4e811af5f4fc49fe5682/src/org/exist/collections/MutableCollection.java#L253-L262). Any other code that have two (or more) locks during processing will result deadlock.

adamretter:
@shabanovd It will not be possible to make eXist-db lock free without replacing all storage backends, and even then you will loose some of the consistency that it currently has.

Regards to locks deadlocking, yes when you have two lock a deadlock is a possibility, but there is decades of research into lock scheduling and lock protocols to help us avoid such situations.

Let's just focus on this PR rather than talking theory. This PR IMHO does not solve any problem, it just moves the problem up one stack-frame.

shabanovd:
@adamretter So, your statement that this PR do not solve any issue, right?

adamretter:
@shabanovd My statement is, that as far as I can see, that this PR just moves the issue up one stack-frame

shabanovd:
@adamretter then let's say this "@wolfgangmm this PR do not address your deadlock situation, but improve concurrency for free ids". Good to go?

PR #1326, opened 2017-03-06
[bugfix] Correct the args definitions for shutdown command line tools

adamretter:
There was a small mistake in the arguments definitions for the shutdown command line tools. This corrects that.


adamretter:
@dizzzz I have no idea, but that is what it was in the past.

ljo:
I always used the long form --uri and showed everyone how to do it like that so that is why I managed to miss it.

PR #1332, opened 2017-03-07
Fix creation of the systemd unit configuration file.

ptrckb:
When executed by a regular user, the installation of the systemd unit configuration file fails trying to write to a protected space (`/etc/systemd/system`). This patch ensures that the file is created as the superuser.


dizzzz:
Thank you for your PR!

dizzzz:
@ljo please could you review?

Issue #1331, opened 2017-03-07
Missing jar for install as service on Linux (systemV)


ptrckb:
When installing as a systemV-init service, the [wrapper configuration file](https://github.com/eXist-db/exist/blob/develop/tools/yajsw/conf/wrapper.conf.in#L218) uses groovy scripting but the interpreter is not available:

```
XXX $ tools/yajsw/bin/installDaemon.sh
...
************* INSTALLING eXist-db ***********************

Mar 06, 2017 4:03:56 PM org.rzo.yajsw.os.posix.PosixService getPid
INFO: wrapper pid file: /run/wrapper.eXist-db.pid
Mar 06, 2017 4:03:56 PM org.rzo.yajsw.os.posix.PosixService stop
INFO: Stopping eXist-db ...
INFO: lib not found: XXX/tools/yajsw/lib/extended/groovy/groovy-patch.jar
INFO: lib not found: XXX/tools/yajsw/lib/extended/groovy/groovy-all-2.3.11.jar
INFO: lib not found: XXX/tools/yajsw/lib/extended/commons/commons-httpclient-3.0.1.jar
...
Mar 06, 2017 4:03:58 PM org.rzo.yajsw.os.posix.PosixService install
INFO: error on creating script link ${if (new File('/etc/rc0.d').exists()) return '/etc/rc4.d' else return '/etc/init.d/rcX.d'}/S20eXist-db
Mar 06, 2017 4:03:58 PM org.rzo.yajsw.os.posix.PosixService install
INFO: error on creating script link ${if (new File('/etc/rc0.d').exists()) return '/etc/rc2.d' else return '/etc/init.d/rcX.d'}/S20eXist-db
Mar 06, 2017 4:03:58 PM org.rzo.yajsw.os.posix.PosixService install
...
```

Adding the `groovy-all-2.3.11.jar` and `groovy-patch.jar` from the [YAJSW](http://yajsw.sourceforge.net) distribution version 12.08  to `tools/yajsw/lib/extended/groovy/` solved the problem.

eXist-db 3.0.0 on Ubuntu 16.04.2 LTS

dizzzz:
@ptrckb so it is not about the 'lib not found' message but really that things installing as a service does not work?

dizzzz:
wow..... this pulls in many megabytes for one simple check :-/

ptrckb:
@dizzzz correct, the 'lib not found' is only the cause of the issue.

ljo:
[...]

PR #1333, opened 2017-03-07
Add groovy jars for system-v

dizzzz:
Should fix  https://github.com/eXist-db/exist/issues/1331 (need to be tested!)


shabanovd:
@dizzzz was missing commons-httpclient-3.0.1.jar address too?

dizzzz:
@shabanovd there are many more missing files. as @ljo said, the affective code is soo little, that we could leave groovy out and have this fixed in a simpler=better way.

PR #1340, opened 2017-03-09
fix XmldbURI to handle 'xmldb:/' case correctly

shabanovd:
Fix the bug prevent use of short 'xmldb:/' url prefix

### DIFF
--- a/test/src/org/exist/xmldb/XmldbURITest.java
+++ b/test/src/org/exist/xmldb/XmldbURITest.java
@@ -42,6 +42,9 @@ public class XmldbURITest {
 
         XmldbURI.xmldbUriFor("xmldb:///db/");
         XmldbURI.xmldbUriFor("xmldb:///db/test");
+
+        XmldbURI.xmldbUriFor("xmldb:/db/");
+        XmldbURI.xmldbUriFor("xmldb:/db/test");
     }
 
     @Test
--- a/test/src/org/exist/xmldb/XmldbURITest.java
+++ b/test/src/org/exist/xmldb/XmldbURITest.java
@@ -41,6 +41,9 @@ public class XmldbURITest {
         XmldbURI.xmldbUriFor("//localhost:" + jettyPort + "/db/");
         //XXX: this MUST work or no MS OS support at all
         //XmldbURI.xmldbUriFor("D:\\workspace\\");
+
+        XmldbURI.xmldbUriFor("xmldb:///db/");
+        XmldbURI.xmldbUriFor("xmldb:///db/test");
     }
 
     @Test
@@ -320,6 +323,17 @@ public class XmldbURITest {
         assertEquals(XmldbURI.API_LOCAL, xmldbURI.getApiName());
         assertEquals("123",xmldbURI.getFragment());
         assertEquals("param=value",xmldbURI.getQuery());
+    }
+
+    @Test
+    public void xmldbURIConstructor16() throws URISyntaxException {
+        XmldbURI xmldbURI = XmldbURI.xmldbUriFor("xmldb:///db/aa/bb/ccc");
+        assertEquals("xmldb", xmldbURI.getInstanceName());
+        assertNull(xmldbURI.getHost());
+        assertEquals(-1, xmldbURI.getPort());
+        assertNull(xmldbURI.getContext());
+        assertEquals("/db/aa/bb/ccc", xmldbURI.getCollectionPath());
+        assertEquals(XmldbURI.API_LOCAL, xmldbURI.getApiName());
     }
 
     @Test
--- a/test/src/org/exist/xmldb/XmldbURITest.java
+++ b/test/src/org/exist/xmldb/XmldbURITest.java
@@ -13,7 +13,7 @@ import static org.junit.Assert.fail;
 
 public class XmldbURITest {
 
-    private final static String jettyPort = System.getProperty("jetty.port");
+    private final static String jettyPort = System.getProperty("jetty.port", "8080");
 
     @Test
     public void xmldbURIConstructors() throws URISyntaxException {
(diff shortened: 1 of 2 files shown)


dizzzz:
```
/home/travis/build/eXist-db/exist/build/scripts/junit.xml:124: Test org.exist.xmldb.XmldbLocalTests failed
```

@shabanovd how come that you don't see this when running the tests locally?

shabanovd:
@dizzzz strange that it did not fail before. There is no system property "jetty.port" anymore -\

test/src/org/exist/xmldb/XmldbURITest.java
```diff
@@ -13,7 +13,7 @@
 
 public class XmldbURITest {
 
-    private final static String jettyPort = System.getProperty("jetty.port");
+    private final static String jettyPort = System.getProperty("jetty.port", "8080");
```
adamretter: This doesn't need to be a `System.getProperty`. This test has no real reliance on Jetty.  Instead I would suggest:

```java
private final static int PORT = 8080;
```

Although it is arguable that you even need it as a global constant.

test/src/org/exist/xmldb/XmldbURITest.java
```diff
@@ -323,39 +326,30 @@ public void xmldbURIConstructor15() throws URISyntaxException {
...
     }
 
     @Test
-    public void xmldbURIFaultyConstructor2() {
+    public void xmldbURIFaultyConstructor1() {
         boolean exceptionThrown = false;
         try{
             @SuppressWarnings("unused")
-            XmldbURI xmldbURI = XmldbURI.xmldbUriFor("xmldb:///db");
+			XmldbURI xmldbURI = XmldbURI.xmldbUriFor("exist:///db");
```
adamretter: When did we start using `exist:///`?

shabanovd: we didn't. Pay attention that test expecting `URISyntaxException` for it.

adamretter: Okay cool. Can you split into two tests with `@Test(expected = URISyntaxException.class)` syntax please

test/src/org/exist/xmldb/XmldbURITest.java
shabanovd: hmm, copy&paste issue from 08e29026831161db366dbbc571c128ad1989f39f?

test/src/org/exist/xmldb/XmldbURITest.java
shabanovd: This tests require much more cleanup. That is not target of this PR.

adamretter: So, I also want to understand why the intention of the test has changed.

It previously checked that `"xmldb:///db"`, would fail. Why does that now not fail?

I think my above question was meant to be, when did we start using `xmldb:///`?

test/src/org/exist/xmldb/XmldbURITest.java
dizzzz: I remember that @ljo added these system properties to be able to perform multiple tests in e.g. Jenkins. think of parallel  running tests. I am pretty sure this must be configurable, in one way or another

shabanovd: @dizzzz why should this be required for XmldbURI tests?

test/src/org/exist/xmldb/XmldbURITest.java
shabanovd: @adamretter what is wrong with this uri prefix? If that is valid uri prefix why it can't be supported?

test/src/org/exist/xmldb/XmldbURITest.java
adamretter: @dizzzz You can run this test in parallel without needing to make it configurable. It has no dependency on Jetty or an eXist server.

test/src/org/exist/xmldb/XmldbURITest.java
adamretter: As far as I am aware `xmldb:///` is not used in eXist. So supporting it without understanding what it is or how to use it will cause confusion. The current scheme is `xmldb:exist:///`, and adding a different but similar one would seem confusing to me and others I expect.

What is the intended purpose of `xmldb:///`?

shabanovd: Some url resolvers having problems of pattern `scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]` where `scheme` usually do not expect to have `:`.

Also, it's much shorter and possible to adopt to use as reference to "embedded" db (any, not just eXist-db).

shabanovd: btw, here old [code](https://github.com/eXist-db/exist/blob/005045fda2ee0df0c3871b12d1b90a79f086ffbc/src/org/exist/validation/resolver/unstable/ExistResolver.java#L52) with `xmldb:///`

dizzzz: I think `xmldb:///` refers to an embedded-mode location

test/src/org/exist/xmldb/XmldbURITest.java
```diff
@@ -23,24 +21,27 @@ public void xmldbURIConstructors() throws URISyntaxException {
         XmldbURI.xmldbUriFor("xmldb:exist:///db");
         XmldbURI.xmldbUriFor("xmldb:exist://localhost/db");
 
-        XmldbURI.xmldbUriFor("xmldb:exist://localhost:" + jettyPort + "/db");
-        XmldbURI.xmldbUriFor("//localhost:" + jettyPort + "/db");
+        XmldbURI.xmldbUriFor("xmldb:exist://localhost:8080/db");
```
dizzzz: I am quite concerned ... this 'flexible' port thing was introduced by @ljo to be able to run tests in parallel on a system.  Or run a test in parallel to an exist-db instance running on the same server. This is not possible anymore.....  Probably introduced with an other PR but now we are drifting away... @adamretter @wolfgangmm  ?

shabanovd:
@dizzzz only tests that require jetty. This tests do not start jetty or eXist.

dizzzz:
ahhhhhhhhhhhhh

test/src/org/exist/xmldb/XmldbURITest.java
shabanovd: yes, you are right

wolfgangmm:
Judging from the comments, nobody opposes this PR. I would thus merge unless there are objections?

PR #1343, opened 2017-03-10
avoid NPE on Permissions getOwner and getGroup calls

shabanovd:
it can happen when account or group can't be found by id because of storage corruption or missing account or group

```
java.lang.NullPointerException: null
	at org.exist.webdav.ExistCollection.initMetadata(ExistCollection.java:120) ~[exist-webdav.jar:?]
	at org.exist.webdav.MiltonCollection.<init>(MiltonCollection.java:123) ~[exist-webdav.jar:?]
	at org.exist.webdav.MiltonCollection.getCollectionResources(MiltonCollection.java:157) ~[exist-webdav.jar:?]
	at org.exist.webdav.MiltonCollection.getChildren(MiltonCollection.java:177) ~[exist-webdav.jar:?]
```

### DIFF
--- a/src/org/exist/security/UnixStylePermission.java
+++ b/src/org/exist/security/UnixStylePermission.java
@@ -24,6 +24,9 @@ import static org.exist.security.PermissionRequired.IS_DBA;
 import static org.exist.security.PermissionRequired.IS_MEMBER;
 import static org.exist.security.PermissionRequired.IS_OWNER;
 import static org.exist.security.PermissionRequired.IS_SET_GID;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.exist.security.internal.RealmImpl;
 import org.exist.storage.io.VariableByteInput;
 import org.exist.storage.io.VariableByteOutputStream;
@@ -41,6 +44,8 @@ import org.exist.storage.io.VariableByteOutputStream;
  * @author Adam Retter <adam@exist-db.org>
  */
 public class UnixStylePermission extends AbstractUnixStylePermission implements Permission {
+
+    public final static Logger LOG = LogManager.getLogger(SecurityManager.class);
 
     protected final SecurityManager sm;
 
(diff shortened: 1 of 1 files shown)


src/org/exist/security/UnixStylePermission.java
```diff
@@ -151,7 +153,11 @@ private void setOwnerId(final int ownerId) {
      */
     @Override
     public Group getGroup() {
-        return sm.getGroup(getGroupId());
+        Group group = sm.getGroup(getGroupId());
+        if (group == null) {
+            return sm.getDBAGroup();
```
dizzzz: is it safe to assume "no group" -> "DBA group" ?

src/org/exist/security/UnixStylePermission.java
```diff
@@ -78,7 +76,11 @@ public UnixStylePermission(final SecurityManager sm, final int ownerId, final in
      */
     @Override
     public Account getOwner() {
-        return sm.getAccount(getOwnerId());
+        Account account = sm.getAccount(getOwnerId());
+        if (account == null) {
+            return sm.getSystemSubject();
```
dizzzz: is it safe to assume "no subject" -> "system subject" ? Better to have 'guest'may be ?

src/org/exist/security/UnixStylePermission.java
shabanovd: Yes, because **only** user from DBA group will have access to it.

src/org/exist/security/UnixStylePermission.java
shabanovd: The problem  with "guest" that anyone will have access to it. Idea here to limit access to DBA only.

adamretter:
So the problem I have here is that this really just "Papering over the cracks". This issue occurs if there is a corruption in the security database. Ideally there should not be corruptions in the security database. Corruptions happen in the security database for 1 or 2 reasons: (1) bugs in the Configurator and or SecurityManager code and (2) corruptions in the database storage itself (.dbx) files.

Whilst I agree that we need to be able to handle such corruptions and continue operating as best as possible, hiding such corruptions is a terrible plan. As an absolute minimum such corruptions must be logged as a fatal level error.

Perhaps for this PR, adding something like:

```java
LOG.fatal("Detected a Security Database corruption. Could not find account for id: {}. Database may need to be repaired.", id);
```

shabanovd:
@adamretter done

PR #1344, opened 2017-03-10
use XmldbURI instead of manual encoding/decoding at java client

shabanovd:
address [issue](https://github.com/wolfgangmm/eXide/issues/151#issuecomment-285603580)

### DIFF
--- a/src/org/exist/client/ClientFrame.java
+++ b/src/org/exist/client/ClientFrame.java
@@ -1600,7 +1600,7 @@ public class ClientFrame extends JFrame implements WindowFocusListener, KeyListe
                 final ResourceDescriptor row = getRow(rowIndex);
                 switch (columnIndex) {
                     case 0:
-                        return row.getName();
+                        return row.getName().toString();
                     case 1:
                         return dateFormat.format(row.getDate());
                     case 2:
--- a/src/org/exist/client/ClientFrame.java
+++ b/src/org/exist/client/ClientFrame.java
@@ -1423,7 +1423,7 @@ public class ClientFrame extends JFrame implements WindowFocusListener, KeyListe
                 final ResourceDescriptor resource = resources.getRow(row);
                 if (resource.isCollection()) {
                     // cd into collection
-                    final String command = "cd " + '"' + URIUtils.urlDecodeUtf8(resource.getName()) + '"'; //$NON-NLS-1$
+                    final String command = "cd " + '"' + resource.getName() + '"'; //$NON-NLS-1$
                     display(command + "\n"); //$NON-NLS-1$
                     process.setAction(command);
                 } else {
--- a/src/org/exist/client/InteractiveClient.java
+++ b/src/org/exist/client/InteractiveClient.java
@@ -1,21 +1,21 @@
 /*
- *  eXist Open Source Native XML Database
- *  Copyright (C) 2001-2015 The eXist Project
- *  http://exist-db.org
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2017 The eXist Project
+ * http://exist-db.org
  *
- *  This program is free software; you can redistribute it and/or
- *  modify it under the terms of the GNU Lesser General Public License
- *  as published by the Free Software Foundation; either version 2
- *  of the License, or (at your option) any later version.
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
  *
- *  This program is distributed in the hope that it will be useful,
- *  but WITHOUT ANY WARRANTY; without even the implied warranty of
- *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- *  GNU Lesser General Public License for more details.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
  *
- *  You should have received a copy of the GNU Lesser General Public
- *  License along with this library; if not, write to the Free Software
- *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
  */
 package org.exist.client;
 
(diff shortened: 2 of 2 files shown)


shabanovd:
this require windows testing

adamretter:
@shabanovd This looks good to me, would like to get @joewiz just to confirm

@joewiz would you be able to test this PR to tell us if it resolves the issues you teased out in your previous testing. I could provide you with a build if need be.

joewiz:
Before proceeding, I'd like to confirm my understanding of how resource and collection URIs work in eXist. I couldn't find anything on this in the prose documentation. So, if I may, let me quote from the eXist book (p. 91):

> There is one extremely important thing you should be aware of when addressing resources and collections inside eXist’s database: eXist uses URL-encoded URIs for naming. This means that all reserved characters, according to the URL encoding rules (for more information, see http://tools.ietf.org/html/rfc3986#page-12), must be percent-encoded!
> For instance:
> • A document that shows up as `an example.xml` in the WebDAV browser must be
addressed programmatically as `an%20example.xml`.
> • A document in the collection `/db/my app (test)/test.xml` must be addressed
with `/db/my%20app%20%28test%29/test.xml`.
> • When you request a list of documents inside a collection (with the function xmldb:get-child-resources), the names returned are URIs, and as such they are URL-encoded.
> If you are not extremely careful with this, your application will quickly become a mess. Always distinguish internally between *names* (useful for displaying document/collection names to the user) and *URIs* (useful for addressing the documents/collections).

Is it correct, then, to make the following prescriptive statements about 3 different interaction scenarios?

1. When addressing collections and resources stored in the database, e.g., via `doc()` or other programmatic functions: any non-ASCII or reserved characters in path components of resource and collection names **must be percent-encoded**. 
1. When retrieving collection and resource URIs, e.g., via `base-uri()`, `xmldb:get-child-resources()`, or other programmatic functions: any non-ASCII or reserved characters in path components of resource and collection names **must be percent-encoded**. 
1. When prompting the user to supply the name of a collection or resource, e.g., in a Create New Collection or Rename Resource dialog: the user may enter non-ASCII or reserved characters in the supplied name, so these characters **must be percent-encoded** before the change is applied to the collection or resource.
1. When displaying collections and resources to the user while browsing the database, e.g., via Java admin client, WebDAV clients, Dashboard > Collection Browser, eXide File Manager and directory pane: the names of resources and collections that are percent-encoded  **must be decoded** for display to the user.

Is this correct? Could the wording be made more precise? 

If we can confirm these basic scenarios, I will develop some further test cases that we can apply to each scenario / client.

joewiz:
Also, @hungerburg [implied](https://github.com/wolfgangmm/eXide/issues/151#issuecomment-285625636) that the XmldbURI class is complex:

> XmldbURI class is so much above my head that I did not dare touch it, not even with a long stick; see https://github.com/eXist-db/exist/commit/2c00a7b85471400b9e556f36957e043ab483300a how I worked around it in RESTServer.java and EXistServlet.java by using its method `createInternal(path)`, that does no encoding, instead of `create(path)` and doing the necessary shenanigans upfront of it.

The source in question is https://github.com/eXist-db/exist/blob/develop/src/org/exist/xmldb/XmldbURI.java. 

Is this code in good shape? In other words, are there any underlying problems with the code here that need to be addressed before we debug further and look to ensuring consistency at the client/UI level?

shabanovd:
@joewiz I think that all that encoding/decoding requirement did get in because of file system (FS) limitations that is not a case anymore (or question of support/implementation for that "old" FS).

I'd say that front-end user (xquery apps or functions, java client) must return utf-8 (as default encoding). How eXist deal with it internally is question of implementation, not end-user headache.

I see it as [java.nio.file.Path](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html) interface  that give end-user readable way and deal with all FS implementations.

BTW, in same cases it may be different if that required, for example, webdav and/or http protocols may require encoding/decoding by specifications.

joewiz:
@shabanovd Interesting - that sounds like it could be a big change, no?

Here are some scenarios for consideration and where things currently work and break down:

1. `xmldb:create-collection('/db', 'my collection')` returns an error: 

    > err:FORG0001 failed to convert my test collection into an XmldbURI: Illegal character in path at index 2: my test collection [source: xquery version "3.1"; xmldb:create-collection('/db', 'my collection')]

    So the solution is to encode this, right? 

    `xmldb:create-collection('/db', xmldb:encode-uri('my collection'))` results in the correct collection name, `my%20collection`.

    But **how do you know** if you're supposed to encode collection names?  Say I wanted to create the collection `A=B` or Chinese characters like `中文`. The first, `A=B`, is a valid collection name without encoding, and xmldb:create-collection() already correctly encodes `中文` as `%E4%B8%AD%E6%96%87`. So applying `xmldb:encode-uri()` to these results in unnecessarily encoded `=` (`A%3DB`) and double-encoded `中文` (`%25E4%25B8%25AD%25E6%2596%2587`)!

    This is precisely the problem we're dealing with in eXide now. I had thought that applying `xmldb:encode-uri()` to all user-supplied resource/collection names would encode only when necessary. It turns out that this results in double-encoding in some cases. In other words, in some cases eXide's use of `xmldb:encode-uri()` is the only way to accept user input, while in other cases the same technique creates faulty resource/collection names. Is this use of `xmldb:encode-uri()` in eXide correct or not? Is there a better function or approach for **correctly** encoding user inputted resource or collection names? 

    Ultimately, the question is: What's wrong here? Why does `xmldb:encode-uri()` encode `=`? Why does `xmldb:create-collection()` refuse to encode spaces while it happily encodes `中文`? 

1. In the Java Admin Client (with your PR, at least), selecting File > "Create collection" and entering `my collection` is created without error and appears in the listing of the contents of the `/db` collection as percent-encoded:

    > my%20collection

    This is also how it appears in the File > Rename dialog. Similarly, creating a collection `中文` appears in the list and in the Rename dialog as percent-encoded:

    > %E4%B8%AD%E6%96%87

    In Transmit and eXide, these two collections appear as `my collection` and `中文` (not percent-encoded).

   `xmldb:get-child-collections('/db')` returns

    > "%E4%B8%AD%E6%96%87"
    > "my%20collection"

    It would be nice if the Java Admin Client could display the decoded collection names like Transmit and eXide do.

1. In the Java Admin Client, renaming the original `my collection` and `中文` collections by appending `2` to the end of the collection results in the collection name being double-encoded:

    > my%2520collection2
    > %25E4%25B8%25AD%25E6%2596%25872

    Renaming a 2nd time by appending yet another character, `3`, to the end, results in:

    > my%252520collection23
    > %2525E4%2525B8%2525AD%2525E6%252596%25258723

    In Transmit, the same renaming of these `my collection` and `中文` collections by appending `2` to the end of the collection works, and the collection names are updated correspondingly in the Java Admin Client, which lists the resulting collection names as:

    > my%20collection2
    > %E4%B8%AD%E6%96%872

    The Java Admin Client's renaming function is clearly doing the wrong thing.

1. `xmldb:create-collection('/db', 'A=B')` returns `/db/A=B`, and the Java Admin Client shows this as `A=B` - both in the main listing and in the rename dialog. 

1. In the Java Admin Client, creating `A=B` now works (with your patch in this PR), and both the main listing and the rename dialog show this collection as `A=B`. Same when creating `A=B` in Transmit.

hungerburg:
@joewiz Maybe I am just a lousy code reader; The author has been on the list recently. There is also https://github.com/eXist-db/exist/blob/develop/src/org/exist/xquery/util/URIUtils.java to consider.
@shabanovd From what I remember from back when I instrumented my build with log statements, that what gets passed around in xmldbURI class between String and URI mostly are in fact plain paths.

shabanovd:
@joewiz Internally eXist keep paths as string (utf-8, java default encoding). So, that is no changes at eXist storage. The problem that it's mess at front-end representation right now. We should clean it up, but because that changes will break backward compatibility full cleanup will go to `4.0.0` and just some fixes to `3.x.y`.

Can we merge this PR and copy your investigation results to issue?

shabanovd:
to be clean, this PR fix only `A=B` case.

shabanovd:
@adamretter @wolfgangmm @dizzzz @ljo should this go to 3.1.1?

adamretter:
@joewiz Please do open an issue for the remaining problems

shabanovd:
@adamretter it was not closed

adamretter:
@shabanovd Okay, well the old issue could be updated then. I just thought it easier to have a new issue which is clear about the outstanding problems. I will leave the decision up to @joewiz

joewiz:
Thanks @shabanovd and @adamretter. 

The original issue that @shabanovd referenced in opening this issue was https://github.com/wolfgangmm/eXide/issues/151#issuecomment-285603580 - which was limited in scope to the "A=B" example in eXide, where @hungerburg spotted the problem. Thanks to @shabanovd for fixing that issue for inclusion in 3.1.1!

In the thread here, I did some deeper research - which is stated in its most complete form above in https://github.com/eXist-db/exist/pull/1344#issuecomment-285901899. The problem seems wide ranging and deep - since it affects both the eXist core storage, XQuery functions, and clients that interact with eXist via XQuery, XML-RPC, and WebDAV. Given that this is potentially such a large, complex problem, it seems to me that we should initiate at least one new issue for this. 

My questions:

1. Should this be a single issue? If multiple issues, how should I break it up?
2. What should the issue(s) be called? Something like, "Problems with encoding and decoding of collection & resource names & URIs?"

PR #1345, opened 2017-03-10
do not set java location by "env.JAVA_HOME" property at yajsw config

shabanovd:
set local java location at yajsw config is wrong because it share with installer config and build fail if JAVA_HOME is not set

### DIFF
--- a/tools/yajsw/build.xml
+++ b/tools/yajsw/build.xml
@@ -38,20 +38,19 @@
     <target name="create-conf" unless="conf.available">
         <echo message="Creating wrapper config files for installer and direct local repo use..."/>
         
-	    <fail unless="env.EXIST_HOME" message="EXIST_HOME environment variable not set. Please set it and start the build again. If run from commandline just prepend it."/>
-	    <fail unless="env.JAVA_HOME" message="JAVA_HOME environment variable not set. Please set it and start the build again. If run from commandline just prepend it."/>
+        <fail unless="env.EXIST_HOME" message="EXIST_HOME environment variable not set. Please set it and start the build again. If run from commandline just prepend it."/>
 
         <filter token="app.name" value="eXist-db"/>
         <filter token="app.long.name" value="eXist-db Native XML Database"/>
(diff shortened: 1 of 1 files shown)


dizzzz:
I overlooked this PR. @ljo what is your idea?

ljo:
I do not get the motivation to change this. @shabanovd could you please elaborate on What this means: "set local java location at yajsw config is wrong because it share with installer config and build fail if JAVA_HOME is not set" The last part about JAVA_HOME I interpret as the intended behaviour.

shabanovd:
@ljo your change force to have `JAVA_HOME` (build fail if it was not set) and also at build time it create `wrapper.conf` with java location of the computer (OS) where it was build.

dizzzz:
@shabanovd that is not too bad is it? as long as in the installer/DMG no value is set? or do I overlook something?

shabanovd:
@dizzzz guess, you did.
 - build fail;
 - that only place where `JAVA_HOME` required;
 - installer have java location of OS where it was build.

ljo:
No. If you install and rebuild a wrapper you are really not talking about 1 percent. We need to keep both cases. Automated testing and installation will be completely hopeless without it. And if you are rebuilding why not simply provide your preferred values as requested? This is still the ant script we are talking about. If you are only installing from the installer your provided JAVA_HOME will be used.

shabanovd:
- why I should rebuild after install?
 - if I do not set JAVA_HOME where installer will take it?
 - how installer will know where you put java location if config file have no `JAVA_HOME`?
 - where automated tests for yajsw?
 - installation always require manual setup or automated by manually prepared configurations files that copied during installation.

dizzzz:
I am not (yet) convinced this is the correct way, as it removes the route te specify JAVA_HOME ?
(though there is a chance that the java executable use it)

adamretter:
@dizzzz @shabanovd Is there not a compromise here. i.e.

1) detect if JAVA_HOME is set.
2) if not see if we find `java` on the path
3) if none of the above, fail and complain appropriately to the user.

PR #1347, opened 2017-03-10
Adds --forking and --pidfile options to server.sh

funkyfuture:
This is supposed for standalone deployments to be controlled by systemd and alike services.

in order to let systemd supervise and control the process it needs to be aware of its pid which is not the case when the `server.sh` script launches Java as a subprocess.

in parts i also changed the indentation to be more comprehendable.

### DIFF
--- a/bin/server.sh
+++ b/bin/server.sh
@@ -5,11 +5,11 @@
 # $Id$
 # -----------------------------------------------------------------------------
 
-## @UNIX_INSTALLER_1@ 
+## @UNIX_INSTALLER_1@
 
 #
-# In addition to the other parameter options for the standalone server 
-# pass -j or --jmx to enable JMX agent.  The port for it can be specified 
+# In addition to the other parameter options for the standalone server
+# pass -j or --jmx to enable JMX agent.  The port for it can be specified
 # with optional port number e.g. -j1099 or --jmx=1099.
 #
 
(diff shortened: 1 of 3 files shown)


bin/server.sh
```diff
@@ -44,6 +44,14 @@ check_jmx_status;
 # save LANG
 set_locale_lang;
 
+if [ $FORKING == "1" ]; then
+    JAVA_RUN="exec ${JAVA_RUN}"
```
dizzzz: will this exit the script? other wise line 55 will be executed too...

funkyfuture: no and yes. it will be executed after string expansion on line 55 and that will exit the script by replacing it with the Java process. what is the intended behaviour.

adamretter:
Looks okay to me

shabanovd:
@dizzzz @adamretter @ljo @wolfgangmm should it go to 3.1.1?

dizzzz:
Depends if it is a bugfix or a feature. I guess 3.2.0 would be a better option

funkyfuture:
please do not merge this yet, i just realised that the `startup.sh` respectively the standalone mode is seemingly not maintained as it doesn't consider a provided `--http-port` option (and there's no stringent documentation concerning it). in opposite to the `startup.sh` does as one would expect. i will thus amend the changes to the `startup.sh` tomorrow.

funkyfuture:
i amended the same changes to `startup.sh`.

adamretter:
@dizzzz I don't see any harm in merging this, so I am going to accept it.

adamretter:
@funkyfuture thanks for your contribution :-)

PR #1358, opened 2017-03-17
make collections' and documents' id immutable

shabanovd:
That's step to collections no lock design.

### DIFF
--- a/src/org/exist/backup/ConsistencyCheck.java
+++ b/src/org/exist/backup/ConsistencyCheck.java
@@ -1,23 +1,21 @@
 /*
- *  eXist Open Source Native XML Database
- *  Copyright (C) 2001-07 The eXist Project
- *  http://exist-db.org
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2017 The eXist Project
+ * http://exist-db.org
  *
- *  This program is free software; you can redistribute it and/or
- *  modify it under the terms of the GNU Lesser General Public License
- *  as published by the Free Software Foundation; either version 2
- *  of the License, or (at your option) any later version.
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
  *
- *  This program is distributed in the hope that it will be useful,
- *  but WITHOUT ANY WARRANTY; without even the implied warranty of
- *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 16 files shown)


adamretter:
Without yet reading through the code, could you tell me how you deal with the 2^32 limit on document and collection ids. I am interested as this becomes more significant if document and collection ids are not recycled.

shabanovd:
@adamretter why do you think that ids do not recycled?

adamretter:
I was going on the title of the PR. If an ID is immutable, it cannot be recycled otherwise its reference is mutable.

shabanovd:
@adamretter you confuse yourself ;-)
"immutable reference" called uniqueness and it was implemented by metadata storage. Id immutability is subject of [java object immutability](https://docs.oracle.com/javase/tutorial/essential/concurrency/imstrat.html).

shabanovd:
@dizzzz if you do not create `BinaryDocument` directly then no needs for update. Search point that you have only one use at `org.exist.jms.replication.shared. MessageHelper`, right?

dizzzz:
@shabanovd that is right indeed, in this code a document is received and stored as a binary document.  *if* this code needs to be updated, it would be a bad thing. More extension can rely on this api, the community cannot afford yet another 'in-ex-ternal' API change; that means that devs would need to support immediately two versions of eXist. Myself, I don't have the time for that.  In this case, the original interface API should remain for some time, marked deprecated.

Or do I misread?

shabanovd:
@dizzzz are you saying that we should stop eXist improving because of 'in-ex-ternal' API change or I misread?

I did send already https://github.com/eXist-db/exist/pull/1324 and this PRs that removing lock-base design to non-blocking algorithm. And I did remove collection lock completely but I do not send it as one big PR because of complexity of that change. So, I split it into smaller change sets and send one by one for review.

_If_ I understand you comment above right you ask me to do not do that, right?

I have change that redesign store procedure from two passes strategy to single pass and that will break API for sure. Should I throw it out too because of 'in-ex-ternal' API change?

Do you have proposal for eXist improvement without 'in-ex-ternal' API change?

dizzzz:
My concrete question is.... do I need to change code in my extension? If not, I am happy. I can't judge though .....

dizzzz:
So am not asking you to stop any development or improvements! My question is quite simple, and again a forum/mail/.... seems to be a bad way of communicating :/

shabanovd:
@dizzzz ok, let's discuss it on monday call

dizzzz:
Please could you answer my question?

shabanovd:
@dizzzz I did - `if you do not create BinaryDocument directly then no needs for update.`

Issue #540, opened 2015-03-23
definition of NCName too strict


jensopetersen:
I would normally use ASCII characters only for xml:id's and suchlike, but for a specific project it makes sense to use Chinese characters as xml:id's. Here eXist appears to me to be too strict. While it accepts characters in the Basic Multilingual Plane (the second one below), it refuses later additions to Unicode. Other XQuery processors do accept these and documents containing them are valid according to Xerces.

	<graphs>
		<graph cp="3400" xml:id="&#x3400;"/>
		<graph cp="4E00" xml:id="&#x4e00;"/>
		<graph cp="20000" xml:id="&#x20000;"/>
		<graph cp="f0000" xml:id="&#xf0000;"/>
		<graph cp="100000" xml:id="&#x100000;"/>
	</graphs>

jensopetersen:
´<x xml:id=“1”/>´ is well-formed, but Xerces holds on to the original XML 1.0 definition of NCName  which was quite quickly superseded in XML 1.0 by the definition made in XML 1.1. Since eXist uses Xerces, this means that xml:id’s beginning with digits are held to be malformed in eXist, though they are well-formed according to XML 1.0.

adamretter:
@jensopetersen I wonder if the Xerces property `http://xml.org/sax/properties/document-xml-version` could help, or perhaps even the feature `http://xml.org/sax/features/xml-1.1`?

See:
https://xerces.apache.org/xerces2-j/properties.html
https://xerces.apache.org/xerces2-j/features.html

tuurma:
In the same vein see discussion at https://github.com/relaxng/jing-trang/issues/188

I have encountered the issue when wanting to use some polytonic Greek characters

ͷ   0377   GREEK SMALL LETTER PAMPHYLIAN DIGAMMA
ϝ   03DD   GREEK SMALL LETTER DIGAMMA
Ͷ   0376   GREEK CAPITAL LETTER PAMPHYLIAN DIGAMMA

which are perfectly kosher NameChar under XML Spec fifth edition https://www.w3.org/TR/REC-xml/#d0e804, alas not a Letter according to fourth https://www.w3.org/TR/2006/REC-xml-20060816/#NT-Letter

PR #1360, opened 2017-03-17
patches validCName

tasmo:
### Description:

Fixes the validCName problem while trying to install some XARs with the package manager.

~~~
/src/org/exist/Indexer.java
w/src/org/exist/Indexer.java

@@ -648,11 +648,6 @@ public class Indexer extends Observable implements ContentHandler, LexicalHandle
                     // an xml:id attribute. Normalize the attribute and set its
                     // type to ID
                     attr.setValue(StringValue.trimWhitespace(StringValue.collapseWhitespace(attr.getValue())));
-
-                    if (!XMLChar.isValidNCName(attr.getValue())) {
-                        throw new SAXException("Value of xml:id attribute is not a valid NCName: " + attr.getValue());
-                    }
-
                     attr.setType(AttrImpl.ID);
                 } else if (attr.getQName().equals(Namespaces.XML_SPACE_QNAME)) {
                     node.setPreserveSpace("preserve".equals(attr.getValue()));
~~~

### Reference:
Chat with @line-o

### DIFF
--- a/src/org/exist/Indexer.java
+++ b/src/org/exist/Indexer.java
@@ -649,10 +649,6 @@ public class Indexer extends Observable implements ContentHandler, LexicalHandle
                     // type to ID
                     attr.setValue(StringValue.trimWhitespace(StringValue.collapseWhitespace(attr.getValue())));
 
-                    if (!XMLChar.isValidNCName(attr.getValue())) {
-                        throw new SAXException("Value of xml:id attribute is not a valid NCName: " + attr.getValue());
-                    }
-
                     attr.setType(AttrImpl.ID);
                 } else if (attr.getQName().equals(Namespaces.XML_SPACE_QNAME)) {
                     node.setPreserveSpace("preserve".equals(attr.getValue()));


joewiz:
Relevant discussion on exist-open: http://markmail.org/message/62ym2duf3ugedecb

dizzzz:
Please could you add a test case  to this PR?

shabanovd:
Can this be configurable?

dizzzz:
I find the suggested change rather dangerous. I am sure that this check is/was there for a good reason. just removing code without a description why, how to reproduce, and a test case is a bad thing. Maybe.... the package manager needs to be fixed in stead.

dizzzz:
@wolfgangmm thank you for your valuable input.

joewiz:
The discussion on exist-open convinced me that we should relax this restriction. But reviewing the xml:id spec raised a few questions for me:

It's interesting to consider which eXist as "xml:id processor" as defined at https://www.w3.org/TR/xml-id/#terminology, and which aspects of eXist do the "xml:id processing" as defined at https://www.w3.org/TR/xml-id/#processing. If one rule that "xml:id processors" are supposed to adhere to ensuring that "The normalized value of the attribute is an NCName," which part of eXist does this? Or are we dropping this requirement, and thus dropping eXist's presumed claim to be a fully compliant "xml:id" processor? 

Assuming these questions don't get in the way of approving the PR, what would an appropriate test case be? Try to store a document with an `@xml:id` with a non-valid NCName (but a valid `@xml:id` value) and confirming that no error is raised? And then call `fn:id()` on this stored document and confirm that the operation **raises or doesn't raise an error**?

joewiz:
Would a test that raises errors on a document with two identical `xml:id` values be outside the scope of this PR? 

(I'm just looking at the other constraints as defined at https://www.w3.org/TR/xml-id/#processing.)

joewiz:
@wolfgangmm I'd be interested if you have the link to the discussion about this from the xslt list?

shabanovd:
@dizzzz @wolfgangmm why did you approve change without test?

dizzzz:
@shabanovd The change ansich is ok, but I asked for a test, no?

shabanovd:
@dizzzz `Approve` PR mean it can be merged. `Request changes` PR mean it can't be merge without that change. [Here](https://help.github.com/articles/about-pull-request-reviews/) details.

tasmo:
@dizzzz What must be absolutely tested?

dizzzz:
@tasmo as the PR actually changes behaviour of document storage, i'd suggest to add a test that stores an small XML document with 'code' that causes the conflict that is solved.

http://stackoverflow.com/questions/1631396/what-is-an-xsncname-type-and-when-should-it-be-used

something like:
```xml
<a>
<b id=":@$%&/+,'"/
</a>
```

joewiz:
@dizzzz Wouldn't it need to be `@xml:id`? I think eXist only recognizes `@xml:id` as an value, not `@id`. See http://exist-db.org/exist/apps/doc/indexing.xml#D3.2.2, item 7:

> `xml:id` Index : An index of all `xml:id` attribute values is automatically created. These values can be queried by `fn:id()`.

dizzzz:
@joewiz you are probably right :-) But this indicates that I do not have a good picture what the fix is about hence a test is essential.

tuurma:
To provide a bit of context:

Sometimes a need arises to use xml:ids that contain polytonic Greek characters. It works mostly fine, but for a few characters, namely 'digamma' symbols (these are what I did encounter).

ͷ   0377   GREEK SMALL LETTER PAMPHYLIAN DIGAMMA
ϝ   03DD   GREEK SMALL LETTER DIGAMMA
Ͷ   0376   GREEK CAPITAL LETTER PAMPHYLIAN DIGAMMA

According to [XML Specs - Fifth edition](http://www.w3.org/TR/REC-xml/#d0e804), these characters do fall within  [#x370-#x37D] and [#x37F-#x1FFF] ranges, so should be ok for NCNames. Alas, Jing for example uses 4th edition of XML Specs and raises error. Related issue there https://github.com/relaxng/jing-trang/issues/188

```xml
<nymList xmlns="http://www.tei-c.org/ns/1.0">
   <nym xml:id="nym_αεις">
      <form> ͷ ϝ Ͷ </form>
   </nym>
   <nym xml:id="nym_Θειόͷρυͷις"/>
   <nym xml:id="nym_Ͷαναξιμοῦς"/>
   <nym xml:id="nym_ϝόδαμος"/>
</nymList>
```

tuurma:
basically addresses #540

line-o:
A test should definitely be in place before removing checks, that may harm a running eXist instance. But in order to do that, what would the test be? What are potential problems?

1. store a document with an invalid CNAME
1. open it
1. check indices?

@dizzzz could you or someone else clarify what to test, because we really need to be able to store documents that will currently fail with this in place even though there is absolutely no problem after removing it.

line-o:
Is anyone able to respond to my above question? @dizzzz @shabanovd @wolfgangmm

dizzzz:
What needs to be proved with an automated test is that the fix actually improves the behavior; a test that before the patch fails and now runs ok....

adamretter:
@tasmo What is the status of this? If it is still relevant, can you please meet @dizzzz's request for a test.

joewiz:
Closed due to lack of response from original poster.

joewiz:
I just noticed that the README.md for https://gitlab.exist-db.org/tei-publisher/tei-publisher-docx#word-to-tei-conversion-package notes that this PR is needed for that package to work. @tasmo @tuurma @wolfgangmm If this PR is important, please reopen and supply a test as requested by @dizzzz and described by @line-o here: https://github.com/eXist-db/exist/pull/1360#issuecomment-289043379.

tasmo:
@line-o Can you help us testing?

PR #1362, opened 2017-03-18
do not hide IOException at webdav createCollection

shabanovd:
`IOException` hidden and `ConflictException` is not raise on `IOException`

### DIFF
--- a/extensions/webdav/src/org/exist/webdav/ExistCollection.java
+++ b/extensions/webdav/src/org/exist/webdav/ExistCollection.java
@@ -285,9 +285,6 @@ public class ExistCollection extends ExistResource {
             LOG.error(e);
             throw e;
 
-        } catch (IOException e) {
-            LOG.error(e);
-
         } catch (Throwable e) {
             LOG.error(e);
             throw new EXistException(e);


shabanovd:
> under what condition did you find this issue?

I do "manual" code review from time to time ;-)

> I see that the code is somewhat inconsistent... sometimes a Throwable is used as 'fall thru', sometimes not.

I just remove `IOException` because there was `Throwable` catch. But catching `Throwable` also quite [big issue](http://stackoverflow.com/questions/6083248/is-it-a-bad-practice-to-catch-throwable) because it also catch `InterruptedException` ([here](http://www.ibm.com/developerworks/java/library/j-jtp05236/index.html) details why it wrong).

dizzzz:
Yes, I can imagine the `InterruptedException` now. The reason I choose to use `Throwable` from time to time, is that I sometimes do not 'trust' third party libraries. Some devs just use runtime exceptions all over the place, which are never documented and not part of the interface contract.

joewiz:
@dizzzz In preparation for 3.2.0, where this is presumably going to be released, should we do extra testing of WebDAV clients to see if any operations that used to work raise errors? Are there any particular operations that you think we should focus tests on given these changes?

dizzzz:
good question. we should probably build some automated tests here, BUT.... each webdav client has its own undefined 'characteristics' regarding the spec, making it doubtful if such a test suite makes sense. But I have some ideas....

joewiz:
Cool, so in the absence of a full test suite, we should probably ask beta/RC testers to give special focus to WebDAV operations? 

As far as I recall, announcements about eXist RCs, etc., don't usually tell people what to test, but this might be a good practice, since we can't really anticipate what/if any effect this change will have on clients. Especially in this case, the core devs don't use all the WebDAV clients that the larger community uses.

shabanovd:
@joewiz this PR do not add any new error, IOException raise only on IO operations. So, it visible only when disk operation fail.

joewiz:
Disk operation on eXist's disk, right? For example, a remote WebDAV client tries to upload a file to eXist, but eXist has run out of free space. Is that the scenario?

shabanovd:
@joewiz yes, in current case that is `create collection` operation. No free space, folder exist under `data/fs/...` and so on

PR #1363, opened 2017-03-19
Slow memory leak

wolfgangmm:
On several servers I saw memory consumption increasing slowly over time. Stress testing revealed instances of BinaryValueFromInputStream staying in memory and building up over weeks until eXist would eventually crash.

Debugging showed the registered XQueryContext.BinaryValueCleanupTask is never executed because cleanupTasks list is cleared too early (in XQueryContext.reset). Thus when XQueryContext.runCleanupTasks is called, the list of tasks is already empty.

The fix is simple: clear cleanupTasks after tasks have been executed, not before.

### DIFF
--- a/src/org/exist/xquery/XQueryContext.java
+++ b/src/org/exist/xquery/XQueryContext.java
@@ -3643,6 +3643,8 @@ public class XQueryContext implements BinaryValueManager, Context
                 LOG.error("Cleaning up XQueryContext: Ignoring: " + t.getMessage(), t);
             }
         }
+        // now it is safe to clear the cleanup tasks list as we know they have run
+        // do not move this anywhere else
         cleanupTasks.clear();
     }
 }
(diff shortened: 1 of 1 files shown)


joewiz:
@wolfgangmm A promising development!

src/org/exist/xquery/XQueryContext.java
```diff
@@ -1418,8 +1418,6 @@ public void reset(final boolean keepGlobals) {
 
         clearUpdateListeners();
 
-        cleanupTasks.clear();
```
shabanovd: think that call `runCleanupTasks` instead with comment will be good? As minimum it'll avoid regression as it happen [here](https://github.com/ChiSamurai/exist/commit/0301cdfc768e5014bdf1132dce93cf467b58638a).

wolfgangmm: ok, I added a comment

Issue #1338, opened 2017-03-08
backup/restore not working


andray71:
### What is the problem
when executing command:
`bin/backup.sh --restore "$EXIST_BACKUP/backup-config-service.zip" -u admin -p "$EXIST_ADMIN_PASSW"`

getting error:
`Missing required arguments: [-b]null`

### What did you expect
database/collection restored. It worked in previous version 3.0

### Describe how to reproduce or add a test
just try to restore any collection using command line

### Context information
Please always add the following information
- eXist-db version: eXist-db 3.1.0 (running in Docker - FROM openjdk:latest...)
- Java version (e.g. Java8u121): 
openjdk version "1.8.0_121"
OpenJDK Runtime Environment (build 1.8.0_121-8u121-b13-1~bpo8+1-b13)
OpenJDK 64-Bit Server VM (build 25.121-b13, mixed mode)
- Operating system: Linux 4.9.12-moby amd64
- 64 bit
- Any custom changes in e.g. conf.xml: no

wolfgangmm:
Argument -b is for creating a backup. You should remove it for restore.

andray71:
the error I'm getting when i run restore, with out -b

StephanMa:
I can confirm that the restore is not working with the current develop-branch

`sudo ./backup.sh -u admin -p "pass" -P "pass" -r Path_To_File.zip -ouri=xmldb:exist://localhost:8181/xmlrpc`

Error as described: 
`Missing required arguments: [-b]null`

shabanovd:
@StephanMa, you can use `system:import` or `system:import-silently` xquery functions. (`system:import-silently` good for big backups, it redirect information to backup log file).

Here an example for your case: `system:import-silently("Path_To_File.zip", "pass", "pass")`.

wolfgangmm:
The following command line should work on a headless system:

```sh
bin/client.sh -l -s -x "system:restore('/path/to/backup.zip', 'admin-pass', 'restore-admin-pass')"
```

StephanMa:
Nice thank you!

andray71:
Thanks!!!

PR #1366, opened 2017-03-19
Some arguments for Backup/Restore should not be required

adamretter:
Closes https://github.com/eXist-db/exist/issues/1338

### DIFF
--- a/src/org/exist/backup/Main.java
+++ b/src/org/exist/backup/Main.java
@@ -109,11 +109,10 @@ public class Main {
     /* user/pass arguments */
     private static final Argument<String> userArg = stringArgument("-u", "--user")
             .description("set user.")
-            .required()
+            .defaultValue(DEFAULT_USER)
             .build();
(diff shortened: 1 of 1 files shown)


src/org/exist/backup/Main.java
```diff
@@ -76,6 +76,20 @@
...
+    private static final String USER_PROP = "user";
+    private static final String PASSWORD_PROP = "password";
+    private static final String URI_PROP = "uri";
+    private static final String CONFIGURATION_PROP = "configuration";
+    private static final String DRIVER_PROP = "driver";
+    private static final String CREATE_DATABASE_PROP = "create-database";
+    private static final String BACKUP_DIR_PROP = "backup-dir";
+
+    private static final String DEFAULT_USER = "admin";
+    private static final String DEFAULT_PASSWORD = "admin";
```
shabanovd: default password should be "" (empty string)

adamretter: Thanks, I just fixed that.

PR #1367, opened 2017-03-20
[bugfix] Memory leak: map constructor

wolfgangmm:
Analyzing a heap dump from one of @joewiz servers, I found another memory leak: the literal constructor for maps does not implement resetState for its subexpressions, so any data held in those expressions is not cleared.

Fix is quite simple: overwrite resetState and pass it on to all subexpressions.


joewiz:
@wolfgangmm Thank you for getting to the bottom of this! 

@shabanovd @adamretter Thank you for the review and merge!

Issue #1364, opened 2017-03-19
doc-available() should return false(), not an error, when encountering parsing errors


joewiz:
### What is the problem

The `fn:doc-available()` function is returning en error when the URI contains non-well-formed XML.

### What did you expect

The `fn:doc-available()` function should return `false()` when the URI cannot return a document node. See https://www.w3.org/TR/xpath-functions-31/#func-doc-available. 

### Describe how to reproduce or add a test

The following query:

```xquery
xquery version "3.1";

fn:doc-available("https://www.kingjamesbibleonline.org/Genesis-Chapter-1_Original-1611-KJV/")
```

... returns the following error:

> err:FODC0005 An error occurred while parsing https://www.kingjamesbibleonline.org/Genesis-Chapter-1_Original-1611-KJV/: Open quote is expected for attribute "lang" associated with an element type "html". [source: xquery version "3.1"; fn:doc-available("https://www.kingjamesbibleonline.org/Genesis-Chapter-1_Original-1611-KJV/")]

Saxon and BaseX both return `false()`. QT3's tests for this function (https://dev.w3.org/cvsweb/2011/QT3-test-suite/fn/doc-available.xml?rev=1.9;content-type=text%2Fx-cvsweb-markup) don't seem to include a test for this possibility, but I believe the expected behavior is clear enough from the spec.

### Context information

- eXist-db version: 3.1.0
- Java version: 1.8.0_121-b13
- Operating system: macOS
- 64 bit
- Any custom changes in e.g. conf.xml: no

shabanovd:
It was closed without test, reopen...

adamretter:
@shabanovd It is not possible to write an automated XQuery Test without introducing external dependencies. As such the test will need to be manual, which I have done, and which you could also do. Re-opening this issue makes no sense as the issue has been fixed, you can confirm that for yourself by testing it. So I am again closing this issue.

joewiz:
@adamretter I tested your fix in develop 0151784a7 and can confirm it fixes the issue as reported. Many thanks!

tuurma:
Seems there has been a regression here, testing on 3.5.0

adamretter:
@tuurma do you still have an issue on 3.5.0? If so, how do I reproduce it? Let me know, I am happy to fix...

tuurma:
@adamretter I tested using @joewiz's example above
```
[...]

PR #1368, opened 2017-03-20
Fix error handling in fn:doc-available

adamretter:
Closes https://github.com/eXist-db/exist/issues/1364

### DIFF
--- a/src/org/exist/xquery/functions/fn/FunDocAvailable.java
+++ b/src/org/exist/xquery/functions/fn/FunDocAvailable.java
@@ -33,6 +33,7 @@ import org.exist.xquery.value.Item;
 import org.exist.xquery.value.Sequence;
 import org.exist.xquery.value.SequenceType;
 import org.exist.xquery.value.Type;
+import org.xml.sax.SAXParseException;
 
 /**
  * Implements the XQuery's fn:doc-available() function.
@@ -83,23 +84,28 @@ public class FunDocAvailable extends Function {
             }
         }
 
-        final Sequence result;
+        Sequence result = BooleanValue.FALSE;
         final Sequence arg = getArgument(0).eval(contextSequence, contextItem);
-        if (arg.isEmpty()) {
-            result = BooleanValue.FALSE;
-        } else {
+        if (!arg.isEmpty()) {
             final String path = arg.itemAt(0).getStringValue();
             try {
                 result = BooleanValue.valueOf(DocUtils.isDocumentAvailable(this.context, path));
             } catch (final XPathException e) {
-                e.prependMessage(ErrorCodes.FODC0005, "");
-                throw e;
+                if (e.getCause() instanceof SAXParseException) {
+                    result = BooleanValue.FALSE;
+                } else if(e.getMessage().contains("is a binary resource, not an XML document")) {
+                    result = BooleanValue.FALSE;
+                } else {
+                    e.prependMessage(ErrorCodes.FODC0005, "");
+                    throw e;
+                }
             }
         }
 
         if (context.getProfiler().isEnabled()) {
             context.getProfiler().end(this, "", result);
         }
+
         return result;
     }
 
--- a/src/org/exist/xquery/functions/fn/FunDocAvailable.java
+++ b/src/org/exist/xquery/functions/fn/FunDocAvailable.java
@@ -1,6 +1,6 @@
 /*
  * eXist Open Source Native XML Database
- * Copyright (C) 2005-2009 The eXist Project
+ * Copyright (C) 2005-2017 The eXist Project
  * http://exist-db.org
  *
  * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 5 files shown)


src/org/exist/xquery/functions/fn/FunDocAvailable.java
```diff
@@ -35,79 +33,83 @@
...
+        if (!arg.isEmpty()) {
             final String path = arg.itemAt(0).getStringValue();
             try {
                 result = BooleanValue.valueOf(DocUtils.isDocumentAvailable(this.context, path));
             } catch (final XPathException e) {
-                e.prependMessage(ErrorCodes.FODC0005, "");
-                throw e;
+                if (e.getCause() instanceof SAXParseException) {
+                    result = BooleanValue.FALSE;
+                } else if(e.getMessage().contains("is a binary resource, not an XML document")) {
```
shabanovd: very bad hack

adamretter: I completely agree. However there is no better fix without either changing the API (which we can't do for a hotfix release) or duplicating code. I think a better fix could come in future. This was just designed for 3.1.1 and it improves the current situation.

dizzzz: :-)  can't we check in the EXpathException for an ErrorCode ?

shabanovd: yes, my proposal to use ErrorCode too

test/src/xquery/fn.xql
```diff
@@ -21,4 +41,17 @@ declare
 function fnt:normalize-unicode($normalization-form as xs:string) {
     count(string-to-codepoints(normalize-unicode($fnt:composed, $normalization-form))),
     count(string-to-codepoints(normalize-unicode($fnt:decomposed, $normalization-form)))
-};
\ No newline at end of file
+};
+
+declare
```
shabanovd: no test for external invalid xml resource?

src/org/exist/xquery/functions/fn/FunDocAvailable.java
adamretter: There is no difference between checking a constant error code and and a constant string here, both have the same fragility.

The solution is to refactor the code in `DocUtils` to be more granular and differentiate between different kinds of errors, however that cannot be done without API changes. This fix is just a temporary fix that improves the current state.

test/src/xquery/fn.xql
adamretter: There was no test before. I added the tests that I needed. Feel free to send a PR with more tests if you want them.

src/org/exist/xquery/functions/fn/FunDocAvailable.java
shabanovd: Huge difference, because _constant_ error have clear where-is-used. No needs for API change!

test/src/xquery/fn.xql
shabanovd: @adamretter you want to close #1364 but do not cover it by test

adamretter: How do you propose I test that without introducing an external dependency?

shabanovd: There is already number of it ... Just use url from issue report.

adamretter: @shabanovd That would introduce a dependency on an external website into the test suite. The problem is that if that website is offline our test suite fails. So I don't want an external dependency.

shabanovd: another option to add corrupted xml and reference it by `http://localhost:...`

shabanovd: BTW, [here](https://github.com/eXist-db/exist/blob/f7643dcb5d3c68c8c695b73d1ede6b0f8a8820b8/test/src/org/exist/xquery/XQueryTest.java#L1219) tests that have an external website dependency

joewiz: @shabanovd Good idea to reference some local file that is corrupted.

A more roundabout, but attractive, approach would be to submit a test to qt3, and then have eXist run this test along with all the others. (We'd need to get the qt3 runner working again though...)

adamretter: @shabanovd The XQuery Test Suite runner operates on an embedded XML Database, so there is no Jetty webserver.

adamretter: @shabanovd No tests should have external web dependencies as it has caused us pain in the past. That above `org.exist.xquery.XQueryTest` should be fixed in a separate PR.

shabanovd: remove `Closes 1364` and then it good to go

adamretter: Why? The PR does fix the issue in 1364.

shabanovd: where do you have test for available _external_ resource but invalid xml?

adamretter: I don't have a test for that which you know from our above discussion. However this does fix the issue, which you could confirm by testing yourself.

Issue #1237, opened 2017-01-27
 namespace constructor in computed constructor inconsistency


duncdrum:
I stumbled on namespace constructors in Walmsley *xQuery* 2nd ed. p. 167-8 (the output of her example is in *Example 10-9* Results).

```xquery
xquery version "3.0";

declare namespace ok="http://place-on-interwebz.com/a-ok";
declare namespace doh="http://also-on-interwebz.com/problem";

element root {namespace {""} {"http://also-on-interwebz.com/problem"}, 
    namespace ok {"http://place-on-interwebz.com/a-ok"},
    for $n in 1 to 3
    return
        element stuff {$n}
}
```

output:
```xml
<root xmlns:ok="http://place-on-interwebz.com/a-ok">
    <stuff>1</stuff>
    <stuff>2</stuff>
    <stuff>3</stuff>
</root>
```

expected:
```xml
<root xmlns="http://also-on-interwebz.com/problem" 
    xmlns:ok="http://place-on-interwebz.com/a-ok">
    <stuff>1</stuff>
    <stuff>2</stuff>
    <stuff>3</stuff>
</root
```

the empty namespace prefix leads to the whole doh namespace uri part to be ommited. 

Tested on eXist 2.2 via eXide and oxygen. Also tested with the ```xmldb:store``` active since there used to be a bug with how namespace appear in eXide's output vs. in stored documents.

dizzzz:
So, I tested in 3.0 in the Java admin: the result is:

```xml
<root xmlns:ok="http://place-on-interwebz.com/a-ok">
    <stuff>1</stuff>
    <stuff>2</stuff>
    <stuff>3</stuff>
</root>
```

which is consistent.  I tested the query also in Saxon (something one should always do when there is a doubt if the query engine in eXist-db is ok), resulting in:

```
Engine name: Saxon-PE XQuery 9.6.0.7
Severity: error
Description: XTDE0440: Cannot output a namespace node for the default namespace when the element is in no namespace
URL: http://www.w3.org/TR/xslt20/#err-XTDE0440
```

My conclusion is: yes you are right, eXist-db should give a different response, but it should yield into an error instead.

duncdrum:
@dizzzz I ll leave it to someone with a better understanding of the xQuery specs to see if it should raise an error or have different results. I don't understand namespaces (1.1, or other) well enough.   I just noticed that the current output could lead to problems, for documents with multiple namespaces.

duncdrum:
[...]

PR #1375, opened 2017-03-21
fix namespace constructor in computed constructor

shabanovd:
close https://github.com/eXist-db/exist/issues/1237

### DIFF
--- a/src/org/exist/dom/memtree/MemTreeBuilder.java
+++ b/src/org/exist/dom/memtree/MemTreeBuilder.java
@@ -428,7 +428,7 @@ public class MemTreeBuilder {
                 final String elemPrefix = (elemQN.getPrefix() == null) ? XMLConstants.DEFAULT_NS_PREFIX : elemQN.getPrefix();
                 if (checkNS
                     && XMLConstants.DEFAULT_NS_PREFIX.equals(elemPrefix)
-                    && "xmlns".equals(qname.getPrefix())
+                    && XMLConstants.XMLNS_ATTRIBUTE.equals(qname.getPrefix())
                     && "".equals(qname.getLocalPart())) {
 
                     throw new DOMException(
--- a/src/org/exist/dom/memtree/MemTreeBuilder.java
+++ b/src/org/exist/dom/memtree/MemTreeBuilder.java
@@ -28,6 +28,7 @@ import org.exist.dom.QName;
 import org.exist.dom.persistent.NodeProxy;
 import org.exist.xquery.Constants;
 import org.exist.xquery.XQueryContext;
+import org.w3c.dom.DOMException;
 import org.w3c.dom.Node;
 import org.xml.sax.Attributes;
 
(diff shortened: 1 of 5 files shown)


src/org/exist/dom/memtree/MemTreeBuilder.java
```diff
@@ -415,12 +416,26 @@ public int namespaceNode(final String prefix, final String uri) {
...
+    public int namespaceNode(final QName qname, boolean checkNS) {
         final int lastNode = doc.getLastNode();
         boolean addNode = true;
         if(doc.nodeName != null) {
             final QName elemQN = doc.nodeName[lastNode];
             if(elemQN != null) {
                 final String elemPrefix = (elemQN.getPrefix() == null) ? XMLConstants.DEFAULT_NS_PREFIX : elemQN.getPrefix();
+                if (checkNS
+                    && XMLConstants.DEFAULT_NS_PREFIX.equals(elemPrefix)
+                    && "xmlns".equals(qname.getPrefix())
```
adamretter: Consider using `XMLConstants.XMLNS_ATTRIBUTE`

src/org/exist/dom/memtree/MemTreeBuilder.java
```diff
@@ -415,12 +416,26 @@ public int namespaceNode(final String prefix, final String uri) {
...
+                    && "".equals(qname.getLocalPart())) {
```
adamretter: Should this not be:
```
&&  XMLConstants.DEFAULT_NS_PREFIX.equals(qname.getNamespaceURI())
```

src/org/exist/dom/memtree/MemTreeBuilder.java
shabanovd: done

src/org/exist/dom/memtree/MemTreeBuilder.java
shabanovd: nope, why it should check that `default namespace prefix` is equal to namespace uri?

adamretter: Sorry I made a copy-paste error. Should be:

```java
&&  XMLConstants.NULL_NS_URI.equals(qname.getNamespaceURI())
```
As checking the local part of a QName used for the namespace doesn't make any sense.

dizzzz: subtile :-)

shabanovd: @adamretter I think you didn't understand what and how was fixed here. I have no idea why you want to check namespace-URI to be "" (empty string) if original issue was in replacing default namespace with **any** namespace-URI

dizzzz: @adamretter do you still have a concern?

dizzzz: technically there is no difference.... it is the empty string ""

dizzzz: probably I just dont see it :-)

PR #1381, opened 2017-03-23
disable seqential token re-creation by default

chakl:
### Description:

Quick fix to disable sequential token re-creation in PersistentLogin, which was creating problems with AJAX. I left the token re-creation code, not sure if this should be made configurable.

### Reference:

https://trello.com/c/p8CLLdNs/46-review-persistent-login-module-and-create-best-practice-recommendation-to-use-in-our-apps

### Type of tests:

Compiled and manually verified. Also ran "./build.sh test"

### DIFF
--- a/extensions/modules/src/org/exist/xquery/modules/persistentlogin/PersistentLogin.java
+++ b/extensions/modules/src/org/exist/xquery/modules/persistentlogin/PersistentLogin.java
@@ -91,12 +91,18 @@ public class PersistentLogin {
             seriesMap.remove(tokens[0]);
             return null;
         }
+
+    // sequential token checking is disabled by default
+    if (data.seqBehavior) {
+        LOG.debug("Using sequential tokens");
         if (!data.checkAndUpdateToken(tokens[1])) {
             LOG.debug("Out-of-sequence request or cookie theft attack. Deleting session.");
             seriesMap.remove(tokens[0]);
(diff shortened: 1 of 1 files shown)


dizzzz:
@wolfgangmm please could you review as you are the original author?

JoernT:
I just had a talk with Olaf to verify the approach:

the goal of this PR is to fix the problems with the persistent login module. Certain situations occurred where the sequentiell tokens lead to problems and exceptions - even if they are not caused by CSRF attacks.

What it does:
with this PR applied session-tokens are not *re-*created any more meaning that an initial token is not refreshed any more. Thus the long session is left in place. The problematic area in the code is just un-armed. This is a working patch regardless of the need for a more general refactoring of the API which will come in a follow-up.

Rationale for the change:
sequentiell session tokens were introduced as a measure against CSRF attacks. But they don't work for asynchronous applications. This disqualifies sequentiell tokens for all kind of modern applications and is not a generic solutions any more. 

Furthermore CSRF (and other) attacks are more efficiently dealt with by using https by default along with some other detail security settings that can be done on a proxy.

The initial purpose for persistent login was to enable long-lasting sessions. This part has been left intact and should be useable now.

adamretter:
@dizzzz Likewise, I think this might have to be one for @wolfgangmm. A cursory glance makes me think it might be okay. Although I am not sure if `updateToken` needs to be called in non-sequential check mode too.

Issue #1313, opened 2017-03-02
Approve the new versioning scheme and release procedure document


joewiz:
With no further objections having been raised or amendments having been submitted, **I move that the eXist Community approve the draft versioning scheme and release procedure proposal** circulated and discussed at this week's Community Call and in followup Q&A via email since then. 

In so doing, we should ask @adamretter to apply changes as discussed and complete preparations for adopting the proposal in this repository, as follows:

1. Delete the section labeled "### TL;DR" and the entire portion stretching from "## Motivation" through the end of the document.

2. Add the note following the definition of major, minor, and patches in the section "### Product Releases" with the following text: "**NOTE:** For the purposes of this present document, we do not define eXist's API. This will be defined in a forthcoming amendment to this document or standalone document." To ensure that happens, we will add a topic for this to the list of future weekly calls, called "Defining the eXist API," and invite draft proposals to be written and circulated for that discussion. 

3. Paste the final text of the versioning document into a reply to this issue, to be moved into documentation at the soonest possible opportunity.

4. Initiate a PR forthwith applying the principles and proposals in the versioning scheme and release procedures document, in preparation for the code freeze for eXist 3.1.0. 

5. Allow 24 hours for comments on the PR and merge.

Then, looking ahead, we will work to carry out the planned review and release of eXist 3.1.0:

- Sunday 5th March 2017 at 12:00 UTC: Code freeze, allowing tests to begin
- Monday 6th March 2017 at 19:30 UTC: Community call to appoint a release manager and draft release notes
- Tuesday 7th March 2017: Release eXist 3.1.0

PR #1387, opened 2017-03-28
[doc] eXist Versioning Scheme and Release Process document

adamretter:
Closes https://github.com/eXist-db/exist/issues/1313

### DIFF
--- a/exist-versioning-release.md
+++ b/exist-versioning-release.md
@@ -1,0 +1,310 @@
+# eXist Versioning Scheme and Release Process
+
+![eXist Logo](https://github.com/eXist-db/exist/raw/develop/webapp/logo.jpg)
+
+## Overview
+
+This document describes a new Versioning Scheme and Release Process for eXist. These two topics are tightly connected, so both are covered in this document. 
+
+* The Versioning Scheme describes how eXist's source code and releases are to be named. The goal is that every new version number unambiguously inform users and developers about its significance and order relative to past and future versions.
+
+* The Release Process describes how the release manager should take a snapshot of eXist source code, apply the Versioning Scheme, assemble it, and publish the resulting products. The goal is to have a clear procedure for altering the version number to mark transitions in phases of development leading up to each release, and to ensure that releases are consistently sourced from a specific point in the project repository's history.
+
+The final section of the document explains the motivation for the recommendation here, for those who would like additional context. Assuming the proposal is adopted, this section could be removed from the final version of this document. 
+
+### TL;DR - The Motivation in Brief
+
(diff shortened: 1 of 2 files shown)


exist-versioning-release.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Where the version number is stored
+
+The version number is stored in the `$EXIST_HOME/build.properties` file, in a single property, `project.version`. The Semantic Versioning number `3.2.0-SNAPSHOT` would be stored as follows:
+
+```
+project.version = 3.2.0-SNAPSHOT
+```
+
+That version number must also should be copied into the `META-INF/MANIFEST.MF` file of any Jar packages that are built, using the standard manifest attributes: `Specification-Version` and `Implementation-Version`; This copying process should be automated as part of the build system.
```
joewiz: s/This/this

exist-versioning-release.md
```diff
@@ -0,0 +1,315 @@
...
+```
+project.version = 3.0.0
+```
+
+Once a stable release has been tagged, we will immediately initiate the next version, assuming a `MINOR` release, unless the core developers select a `PATCH` or `MAJOR` version; this next version will have a `LABEL` appended, e.g., `3.1.0-SNAPSHOT`, which will persist until 3.1.0 is released, unless a new `PATCH` or `MAJOR` version must be released first.
+
+We can call the installer (and/or packages) anything we want, but
+it would be sensible for them to reflect the version number clearly.
+So for simplicity we suggest just using the same version as is in
+project.version, i.e.:
```
joewiz: Consider removing the hard line breaks here and below, as no other paragraphs above use them.

exist-versioning-release.md
```diff
@@ -0,0 +1,315 @@
...
+eXist-db-setup-4.0.0-RC1.jar
+eXist-db-4.0.0-RC1.dmg
+
+exist-core-4.0.0-RC1.jar
+exist-core-4.0.0-RC1.pom
+```
+
+Having the git commit hash in any final release filenames is
+redundant. It only really made sense when we didn't have frequent
+releases.
```
joewiz: (More hard line breaks here.)

exist-versioning-release.md
adamretter: Okay fixed

exist-versioning-release.md
adamretter: Uggh. An artifact of copying and pasting Markdown in and out of GoogleDocs :-/
Just fixed it.

exist-versioning-release.md
adamretter: Just fixed them.

joewiz: Thanks for all the fixes!

adamretter:
@joewiz I think this could go in the documentation repo once that has been modernised.

PR #1397, opened 2017-04-07
General cleanup

adamretter:
This is almost just all minor cleanup stuff.

The most interesting thing is perhaps that I have now disabled AutoDeployment of EXPath packages in the tests (where not needed). This has the advantage that running individual tests that start and stop the database are much quicker. It also speeds up the JUnit test suite where tests delete the content of the *data dir* because XARs do not have to re-autodeployed.


src/org/exist/storage/CollectionCacheManager.java
```diff
@@ -49,8 +49,9 @@ public CollectionCacheManager(final BrokerPool pool, final CollectionCache cache
 
     @Override
     public void configure(final Configuration configuration) throws BrokerPoolServiceException {
-        final int cacheSize = Optional.of(configuration.getInteger(PROPERTY_CACHE_SIZE)).filter(size -> size > 0).orElse(DEFAULT_CACHE_SIZE);
-        this.maxCacheSize = cacheSize * 1024 * 1024;
+        final int cacheSize = Optional.of(configuration.getInteger(PROPERTY_CACHE_SIZE_BYTES)).filter(size -> size > 0)
```
shabanovd: this make it 4.0 changes

src/org/exist/util/Configuration.java
```diff
@@ -837,17 +837,33 @@ private void configureBackend( final Optional<Path> dbHome, Element con ) throws
...
+                } else if(collectionCache.endsWith("gb")) {
+                    collectionCacheBytes = 1024 * 1024 * 1024 * Integer.valueOf(collectionCache.substring(0, collectionCache.length() - 2));
+                } else {
+                    collectionCacheBytes = Integer.valueOf(collectionCache);
+                }
 
-            if( collectionCache.endsWith( "M" ) || collectionCache.endsWith( "m" ) ) {
-                collectionCache = collectionCache.substring( 0, collectionCache.length() - 1 );
-            }
+                config.put(CollectionCacheManager.PROPERTY_CACHE_SIZE_BYTES, collectionCacheBytes);
```
shabanovd: this make it 4.0 changes. Add "M" as deprecated.

test/src/org/exist/TestUtils.java
```diff
@@ -17,23 +16,42 @@
...
+ * Utility functions for working with tests
  */
 public class TestUtils {
-	
-    public static void cleanupDB() {
 
+    /**
+     * Default Admin username used in tests
+     */
+    public static final String ADMIN_DB_USER = "admin";
```
shabanovd: why can't be use constants from SecurityManager?

src/org/exist/util/Configuration.java
dizzzz: not sure if  I follow....

test/src/org/exist/storage/DirtyShutdownTest.java
```diff
@@ -63,7 +64,7 @@ public void storeRepeatedly() {
             for (int i = 0; i < 50; i++) {
                 try(final Txn transaction = transact.beginTransaction()) {
 
-                    Path f = Paths.get("samples/shakespeare/macbeth.xml");
+                    final Path f = TestUtils.resolveShakespeareSample("macbeth.xml");
```
shabanovd: why it can't be `TestUtils.shakespeareSample().resolve("macbeth.xml");`?

src/org/exist/util/Configuration.java
shabanovd: if configuration have "...M" it will not be supported anymore.

dizzzz: ah..... unless the collectionCache is always filled lowercase. Small m means milli, no :-=)

shabanovd: also, the difference between low and upper cases is huge. Here the [wiki](https://en.wikipedia.org/wiki/File_size) for that.

`k` equal 1000 bit and `K` - 1024 byte.

shabanovd: ok, what to do with "1000 bit" vs "1024 byte"?

dizzzz: @shabanovd now you go a bit too far :-)

adamretter: @shabanovd The units are KB and MB (i.e. 1024) not KiB or MiB (i.e. 1000).
I am certain nothing has changed in the API, everything is still  backwards compatible as @dizzzz showed. Or am I mistaken?

test/src/org/exist/TestUtils.java
adamretter: I could reference them, however they have different purposes; so changing the ones in the SecurityManager in future could cause unexpected changes in the tests; which is why I did not do that.

test/src/org/exist/storage/DirtyShutdownTest.java
adamretter: It could be, but that is not the design I chose. There are many tests that load shakespeare text so I felt it was more convenient to offer a function for it.

src/org/exist/util/Configuration.java
dizzzz: so the change is support from mb only to kb and even gb ; looks good to me?

test/src/org/exist/TestUtils.java
dizzzz: makes sense

test/src/org/exist/storage/DirtyShutdownTest.java
dizzzz: ok

src/org/exist/util/Configuration.java
adamretter: Yes I added support for units other than `M`.

shabanovd: @adamretter, fyi, KiB = 1024 byte and MiB = 1024 KiB.

the problem that you make no difference between: KB, kB, kb.

adamretter: @shabanovd I did what I thought was correct and best. So my mistake. Feel free to improve on my work

shabanovd: why mistake should go over review procedure?

shabanovd: eXist become single developer product ... enjoy it!

dizzzz: to be honest: as a simple end user, I'd only think in terms of gigabyte, megabyte or kilobyte, and I'd not care about the 1000 vs 1024, and KiB an MiB I'd not understand at all. Magnitude is important, in that sense @adamretter 's commit is actually too complex. If 1000000kb 1000000KB 1000000K, 1000MB 1000mb 1000m and 1GB 1gb 1g work, it should be ok.

PR #1408, opened 2017-04-14
in-memory embedded streams

shabanovd:
In-memory stream do not block DBBroker

### DIFF
--- a/conf.xml.tmpl
+++ b/conf.xml.tmpl
@@ -171,10 +171,13 @@
 		-->
 		<trigger class="org.exist.security.GnuCryptoJceProviderStartupTrigger"/>
 
-		<!--
-		    Trigger for registering eXists XML:DB URL handler with Java
-		-->
-		<trigger class="org.exist.protocolhandler.URLStreamHandlerStartupTrigger"/>
+                <!--
+                    Trigger for registering eXists XML:DB URL handler with Java
+                -->
+                <trigger class="org.exist.protocolhandler.URLStreamHandlerStartupTrigger">
+                    <parameter name="mode" value="memory"/>
+                    <!--<parameter name="mode" value="threads"/>-->
+                </trigger>
 
                 <!-- 
                     EXQuery RESTXQ trigger to load the RESTXQ Registry at startup time 
--- a/conf.xml.tmpl
+++ b/conf.xml.tmpl
@@ -175,7 +175,12 @@
                     Trigger for registering eXists XML:DB URL handler with Java
                 -->
                 <trigger class="org.exist.protocolhandler.URLStreamHandlerStartupTrigger">
+                    <!-- Keep stream's data in memory -->
                     <parameter name="mode" value="memory"/>
+                    <!-- low memory requirements, but also it have design bugs:
+                          - broker hold while io stream in "use";
+                          - require 2 threads per operation
+                    -->
                     <!--<parameter name="mode" value="threads"/>-->
                 </trigger>
 
--- a/src/org/exist/protocolhandler/Mode.java
+++ b/src/org/exist/protocolhandler/Mode.java
@@ -23,6 +23,15 @@ package org.exist.protocolhandler;
  * @author <a href="mailto:shabanovd@gmail.com">Dmitriy Shabanov</a>
  */
 public enum Mode {
+  /**
+   * low memory requirements, but it have design bugs:
+   *  - broker hold while io stream in "use";
+   *  - require 2 threads per operation.
+   */
   THREADS,
+
+  /**
+   * Keep stream's data in memory to solve THREADS design bugs.
+   */
   MEMORY
 }
--- a/src/org/exist/protocolhandler/Mode.java
+++ b/src/org/exist/protocolhandler/Mode.java
@@ -1,0 +1,28 @@
+/*
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2017 The eXist Project
+ * http://exist-db.org
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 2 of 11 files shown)


dizzzz:
I hoped that the embedded input stream would work for you?

shabanovd:
nope, it block DBBroker too

conf.xml.tmpl
```diff
@@ -171,10 +171,13 @@
 		-->
 		<trigger class="org.exist.security.GnuCryptoJceProviderStartupTrigger"/>
 
-		<!--
-		    Trigger for registering eXists XML:DB URL handler with Java
-		-->
-		<trigger class="org.exist.protocolhandler.URLStreamHandlerStartupTrigger"/>
+                <!--
+                    Trigger for registering eXists XML:DB URL handler with Java
```
dizzzz: it would me good to have details about the difference between two values....

src/org/exist/protocolhandler/Mode.java
```diff
@@ -0,0 +1,28 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.protocolhandler;
+
+/**
+ * @author <a href="mailto:shabanovd@gmail.com">Dmitriy Shabanov</a>
+ */
+public enum Mode {
+  THREADS,
```
dizzzz: javadoc ?

conf.xml.tmpl
dizzzz: please could you document the difference, why one should choose for one to the other? or explain me that i'll write it for you, but I have no clue yet.....  and the dangers...

src/org/exist/protocolhandler/Mode.java
dizzzz: please add javadoc, describing whet 'THREADS' and "MEMORY" is

src/org/exist/protocolhandler/URLStreamHandlerStartupTrigger.java
```diff
@@ -51,13 +51,21 @@
 
     @Override
     public void execute(final DBBroker sysBroker, final Map<String, List<? extends Object>> params) {
-        registerStreamHandlerFactory();
+        String mode = null;
+        if (params != null) {
+            List<?> list = params.get("mode");
+            if (list != null && list.size() == 1) {
```
dizzzz: you could use less restrictive code:

`(list != null &&  !list.isEmpty())`

wolfgangmm:
As I understood this PR tries to address an issue arising with complex XProc pipelines: using  EmbeddedInputStream, brokers are kept until the entire stream has been processed. To run a particular pipeline I thus had to increase the number of brokers to 2000 in the past. With Dmitriy's fix this is no longer necessary, so I can see the benefit of it, despite requiring more memory (though memory consumption is neglectable for the use case).

What I do not entirely understand yet is why we cannot change EmbeddedInputStream to overcome this limitation and need an alternative implementation instead? @shabanovd maybe you could explain?

shabanovd:
@wolfgangmm change EmbeddedInputStream mean only one solution at the end. Original implementation have very low memory requirements, but also it have design bugs: broker hold while io stream in "use" and require 2 threads per operation. New design solve design bugs but increase memory consumption. While it wasn't prove that new design stable I think that better to keep old implementation.

src/org/exist/protocolhandler/protocols/xmldb/ConnectionMemory.java
```diff
@@ -0,0 +1,89 @@
...
+import org.exist.protocolhandler.embedded.InMemoryOutputStream;
+import org.exist.protocolhandler.xmldb.XmldbURL;
+import org.exist.protocolhandler.xmlrpc.XmlrpcInputStream;
+import org.exist.protocolhandler.xmlrpc.XmlrpcOutputStream;
+
+/**
+ *  A URLConnection object manages the translation of a URL object into a
+ * resource stream.
+ */
+public class ConnectionMemory extends URLConnection {
```
wolfgangmm: Rename to InMemoryURLConnection?

src/org/exist/protocolhandler/protocols/xmldb/ConnectionThreads.java
```diff
@@ -47,14 +44,14 @@
  *
  * @author Dannes Wessels
  */
-public class Connection extends URLConnection {
+public class ConnectionThreads extends URLConnection {
```
wolfgangmm: Rename to EmbeddedURLConnection?

src/org/exist/protocolhandler/Mode.java
wolfgangmm: Yes, there should be an explanation of the two implementations. Maybe use what you responded to me in comments?

shabanovd:
@wolfgangmm done

wolfgangmm:
Following @shabanovd response, I'm ok to merge because the old implementation is not changed, so we can test extensively in real world. @dizzzz ok?

dizzzz:
I wanted to make one suggestion, since I am a Bit concerned about the scalability of the in memory caching.  I think it is better to have a overflow to disk for large documents, like the Deprecated virtualfile class or the deferreXXXstream from apache.org

If we make a issue out of this, that is ok.

Note: if we have still an issue validating a stored xml document with a schema in the same collection, this should be solved :$

dizzzz:
And: we should probably mark the old embeddedXstream classes as deprecated. The new names look good to me!

dizzzz:
pull in?!

Issue #1406, opened 2017-04-13
DefaultDocumentSet#contains always returns true


adamretter:
There is a bug in the implementation of `org.exist.dom.persistent.DefaultDocumentSet#contains(org.exist.dom.persistent.DocumentSet)` where it checks the contents of the set against itself.

See https://github.com/eXist-db/exist/blob/develop/src/org/exist/dom/persistent/DefaultDocumentSet.java#L187

I have no idea why we don't see real problems from this...

PR #1409, opened 2017-04-15
Bug fix and optimise in DefaultDocumentSet

adamretter:
1. Fixes a bug in `DefaultDocumentSet#contains`, see https://github.com/eXist-db/exist/issues/1406
2. Optimises `contains` between two `DeafultDocumentSets`
2. Optimises `equalDocs` between two `DeafultDocumentSets`

@wolfgangmm I would like your thoughts on why we have not encountered this bug in `DefaultDocumentSet#contains` before in XQuery.

### DIFF
--- a/src/org/exist/dom/persistent/DefaultDocumentSet.java
+++ b/src/org/exist/dom/persistent/DefaultDocumentSet.java
@@ -188,12 +188,11 @@ public class DefaultDocumentSet extends Int2ObjectHashMap implements MutableDocu
         if (other.getDocumentCount() > size()) {
             return false;
         }
-        for (int idx = 0; idx < tabSize; idx++) {
-            if (values[idx] == null || values[idx] == REMOVED) {
-                continue;
-            }
-            final DocumentImpl d = (DocumentImpl)values[idx];
-            if (!contains(d.getDocId())) {
+
+        final Iterator<DocumentImpl> otherDocumentIterator = other.getDocumentIterator();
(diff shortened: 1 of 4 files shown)


wolfgangmm:
The only place this method is called from is GeneralComparison.nodeSetCompare, which attempts to optimize the comparison, but would fall back to a less efficient approach. Query results remain the same though, which is why we didn't notice. 

I'll review the code...

adamretter:
@wolfgangmm Thanks. Yes I saw that it was called from `GeneralComparison.nodeSetCompare` as an optimization. Does that mean that that optimization path is not being applied for some reason then?

adamretter:
@wolfgangmm Any update please?

dizzzz:
👍🏼👍🏼👍🏼

adamretter:
@wolfgangmm I am still wondering about how this bug was never exposed, i.e. my earlier comment -

> @wolfgangmm Thanks. Yes I saw that it was called from GeneralComparison.nodeSetCompare as an optimization. Does that mean that that optimization path is not being applied for some reason then?

Any thoughts

wolfgangmm:
The optimization path was not applied anymore. It reduces the number of nodes to be compared, so seems reasonable to me.

adamretter:
@wolfgangmm Okay, so I suppose my question was/is, should that optimisation path be applied still? And if so, where do I look to find out/test why it is not working?

PR #1416, opened 2017-04-17
process all elements at selectFollowingSiblings

### LINKED ISSUE #1407: following-sibling axis not working correctly
`$items//b/following-sibling::a`
and
`$items//b[following-sibling::a]`
fail to return any nodes, even though the items in $items do contain such nodes. 

What does work is:

- the preceding-sibling axis used the same way, i.e. with an element name specified (no surprise here)
and the following workarounds
- `$items//b/following-sibling::*[name()='a']`
- `$items//b/self::*[following-sibling::a]` (so following-sibling with element name does work after all, but only under very special circumstances)

### Comment
I saw this bug reported in the mailing list a couple of years ago, but it seemed to receive surprisingly little attention, even though it seems to me like a really huge, mission-critical issue. (Well, luckily, there is the workaround with name(), but still.) Would someone with more experience kindly let me know if I am missing something obvious here?

### Context information
- eXist-db version: 2.2 and 3.1.1 release versions
- Java version: Java8u121
- Operating system: Windows 10 - 64 bit
- no custom changes to config.

shabanovd:
close #1407

### DIFF
--- a/src/org/exist/dom/persistent/NewArrayNodeSet.java
+++ b/src/org/exist/dom/persistent/NewArrayNodeSet.java
@@ -654,7 +654,7 @@ public class NewArrayNodeSet extends AbstractArrayNodeSet implements ExtNodeSet,
             final NodeId parentId = reference.getNodeId().getParentId();
             final int docIdx = findDoc(reference.getOwnerDocument());
             if(docIdx < 0) {
-                return NodeSet.EMPTY_SET;
+                continue;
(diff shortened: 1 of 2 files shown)


adamretter:
@shabanovd Excellent work!

joewiz:
Thanks so much, Dmitriy! I think this has bitten me in the past too, and I never got around to isolating the problem.

PR #1419, opened 2017-04-18
[bugfix] Add options to disable XQuery and XUpdate execution via submission

adamretter:
These are the options as previously discussed to disable submission of XQuery and/or XUpdate by the REST Server.

Feel free to comment/discuss

### DIFF
--- a/src/org/exist/http/RESTServer.java
+++ b/src/org/exist/http/RESTServer.java
@@ -62,6 +62,8 @@ import org.exist.dom.persistent.MutableDocumentSet;
 import org.exist.dom.QName;
 import org.exist.dom.persistent.XMLUtil;
 import static org.exist.http.RESTServerParameter.*;
+
+import org.exist.http.servlets.EXistServlet;
 import org.exist.http.servlets.HttpRequestWrapper;
 import org.exist.http.servlets.HttpResponseWrapper;
 import org.exist.http.servlets.ResponseWrapper;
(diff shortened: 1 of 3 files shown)


dizzzz:
Thank you @adamretter for picking this up!

joewiz:
@adamretter Looks like a great approach to the issue raised & discussed on exist-open (http://markmail.org/message/dwqnboz4txph7dak) and in https://github.com/eXist-db/documentation/issues/98, and really easy to configure for admins. Many thanks!

cc: @simar0at @rvdb @mathias-goebel @martindholmes @gspringtech

simar0at:
Sorry to state that, but: If these two features are enabled by default this is still a trap for anyone who does not use or even know about the "REST-API". I for example develop for exist-db for more than two years and I never used it. Not even eXide uses it. I would suggest to either set it to authenticate or disable and warn people in the release notes about this change. I guess there are many more instances out there on the web where this is enabled by accident but not used by the applications running there.

adamretter:
@simar0at It is currently enabled by default because we did not want to introduce an API change in a minor release (we follow the SemVer.org spec). We might disable it by default in the next major release. I would also note that a read of the documentation for eXist-db makes it clear that this XQuery submission feature is there and present.

joewiz:
@simar0at That is a reasonable proposal, but according to eXist's [versioning practices](https://github.com/eXist-db/exist/blob/develop/exist-versioning-release.md#versioning-scheme), the change you describe would likely be classified as an API change and would thus have to wait for a major version release (e.g., v4.0). As it is, @adamretter's commit is presented as a "hotfix," which can be released anytime (e.g., v3.1.2). I propose that we not wait until 4.0, but instead put our energies regarding this ticket into documenting this newly exposed control over these parameters, noting their default values, the reasons one might want to change the values, and how to change them. Thus I've opened a ticket at https://github.com/eXist-db/documentation/issues/109. I think this can be addressed through communication.

duncdrum:
Should we change the default for 4.0?

adamretter:
@duncdrum maybe, I am not sure. Secure by default is nice I guess, but it might break peoples expectations or make teaching harder?

Issue #1337, opened 2017-03-08
Getting "ERROR isSameNode(Node other) not implemented on other class org.exist.dom.persistent.DocumentImpl"


ebouladier:
This is something that was working with eXist-db version 3.0RC2 and doesn't work anymore with 3.1.0.

When executing an XQuery, from eXide for instance, there is an issue with the instruction : 

```xquery
transform:transform(
 doc($xmlFileUri), 
 doc(concat($xsltPath, '/', 'FancyTreeVisuDeclB30.xsl')), $params
)
```

We get this in return : 
`exerr:ERROR Exception while transforming node: isSameNode(Node other) not implemented on other class org.exist.dom.persistent.DocumentImpl`

When debugging an execution of our application embeddind eXist, we have seen that this message is coming from the method is SameNode of class ElementImpl : 
```java
    @Override
    public boolean isSameNode(final Node other) {
        // This function is used by Saxon in some circumstances, and this partial implementation is required for proper Saxon operation.
        if(other instanceof IStoredNode) {
            return (this.nodeId == ((IStoredNode<?>) other).getNodeId() &&
                this.ownerDocument.getDocId() == ((IStoredNode<? extends IStoredNode>) other).getOwnerDocument().getDocId());
        }
        throw new DOMException(DOMException.NOT_SUPPORTED_ERR,
            "isSameNode(Node other) not implemented on other class " + other.getClass().getName());
    }
```

So it seems that DocumentImpl should implement IStoredNode...

### Context information
[eXist Version : 3.1.0] 
[eXist Build : 20170307214949] 
[Git commit : c679cd8] 
[Operating System : Windows Server 2008 R2 6.1 amd64]

ebouladier:
Adding the console trace

`org.xmldb.api.base.XMLDBException: exerr:ERROR Exception while transforming node: isSameNode(Node other) not implemented on other class org.exist.dom.persistent.DocumentImpl [at line 47, column 6]
	at org.exist.xmldb.LocalXPathQueryService.execute(LocalXPathQueryService.java:195)
	at org.exist.xmldb.LocalXPathQueryService.lambda$doQuery$0(LocalXPathQueryService.java:151)
	at org.exist.xmldb.function.LocalXmldbFunction.apply(LocalXmldbFunction.java:46)
	at org.exist.xmldb.AbstractLocal.withDb(AbstractLocal.java:195)
[...]

PR #1426, opened 2017-04-25
[bugfix] Ensure that isSameNode is correctly implemented

adamretter:
Closes https://github.com/eXist-db/exist/issues/1337

### DIFF
--- a/src/org/exist/dom/persistent/ElementImpl.java
+++ b/src/org/exist/dom/persistent/ElementImpl.java
@@ -1734,17 +1734,6 @@ public class ElementImpl extends NamedNode implements Element {
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


src/org/exist/dom/persistent/ElementImpl.java
```diff
@@ -1734,17 +1734,6 @@ private XmldbURI calculateBaseURI() {
     }
 
     @Override
-    public boolean isSameNode(final Node other) {
```
dizzzz: without knowing too much of the details, I did not expect this to be removed.  Where is this 'detail' check done now?

dizzzz:
so `isSameNode()` is and more important  *was* only called by saxon?

PR #1436, opened 2017-05-09
Removes unused JETTYCONTAINER_OPTS

funkyfuture:
i was looking for a way to provide Jetty configuration per cli-options. this variable seemed to point into the right direction, but leads nowhere.


adamretter:
@funkyfuture Thanks for your contribution.

Most Jetty config properties can be overridden at startup time by setting suitable system properties from the command line, e.g. `-Djetty.port=9090`

PR #1444, opened 2017-05-17
fn:Sort function

shabanovd:
Implementation of [fn:sort](https://www.w3.org/TR/xpath-functions-31/#func-sort)

### DIFF
--- a/src/org/exist/xquery/functions/fn/FunSort.java
+++ b/src/org/exist/xquery/functions/fn/FunSort.java
@@ -24,16 +24,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import org.exist.dom.QName;
 import org.exist.util.Collations;
-import org.exist.xquery.AnalyzeContextInfo;
-import org.exist.xquery.Atomize;
-import org.exist.xquery.BasicFunction;
-import org.exist.xquery.Cardinality;
-import org.exist.xquery.Constants;
-import org.exist.xquery.ErrorCodes;
-import org.exist.xquery.Function;
-import org.exist.xquery.FunctionSignature;
-import org.exist.xquery.XPathException;
-import org.exist.xquery.XQueryContext;
+import org.exist.xquery.*;
 import org.exist.xquery.value.AtomicValue;
 import org.exist.xquery.value.FunctionParameterSequenceType;
 import org.exist.xquery.value.FunctionReference;
@@ -176,6 +167,8 @@ public class FunSort extends BasicFunction {
             }
           } else if (item1 instanceof AtomicValue && item2 instanceof AtomicValue) {
             try {
+              // throw type error if values cannot be compared with lt
+              ValueComparison.compareAtomic(collator, (AtomicValue)item1, (AtomicValue)item2, Constants.StringTruncationOperator.NONE, Constants.Comparison.LT);
               res = ((AtomicValue)item1).compareTo(collator, (AtomicValue)item2);
             } catch (XPathException e) {
               exception.set(e);
--- a/src/org/exist/xquery/functions/fn/FunSort.java
+++ b/src/org/exist/xquery/functions/fn/FunSort.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import org.exist.dom.QName;
 import org.exist.util.Collations;
 import org.exist.xquery.AnalyzeContextInfo;
+import org.exist.xquery.Atomize;
 import org.exist.xquery.BasicFunction;
 import org.exist.xquery.Cardinality;
 import org.exist.xquery.Constants;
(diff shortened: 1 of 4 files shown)


tuurma:
Thank you @shabanovd for quick job on this! Don't you think there should be more test for stability of the sorting and less trivial cases?

I'd like to ask @wolfgangmm and possibly @chakl for review, of course all volunteers welcome :-)

src/org/exist/xquery/functions/fn/FunSort.java
```diff
@@ -0,0 +1,223 @@
...
+
+  AnalyzeContextInfo cachedContextInfo;
+
+  public FunSort(XQueryContext context, FunctionSignature signature) {
+    super(context, signature);
+  }
+
+  @Override
+  public void analyze(AnalyzeContextInfo contextInfo) throws XPathException {
+    if (getContext().getXQueryVersion()<30) {
```
dizzzz: should probably be 31

joewiz: @dizzzz is correct. `fn:sort()` was added in XQuery 3.1 F&O; see https://www.w3.org/TR/xpath-functions-31/#substantive-changes.

shabanovd: done

shabanovd:
@wolfgangmm just add more tests

src/org/exist/xquery/functions/fn/FunSort.java
```diff
@@ -116,14 +119,16 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
...
-      keys.add(value);
+      keys.add(Atomize.atomize(value));
     }
 
     return sort(seq, keys, collator);
   }
 
   private Sequence sort(Sequence seq, ArrayList<Sequence> keys, Collator collator) throws XPathException {
 
+    final Holder<XPathException> exception = new Holder<>();
```
adamretter: I think you probably want 'Optional'

src/org/exist/xquery/functions/fn/FunSort.java
```diff
@@ -220,4 +240,16 @@ private FunctionReference function(Sequence[] args, int pos) throws XPathExcepti
       return null;
     }
   }
-}
+
+  static class Holder<T> {
```
adamretter: This makes no sense

src/org/exist/xquery/functions/fn/FunSort.java
shabanovd: `Optional` will not help here because it's not possible to update it value.

test/src/xquery/xquery3/fnSort.xql
```diff
@@ -0,0 +1,172 @@
...
+
+declare
+    %test:assertEquals("1","3","4","5","6")
+function testSort:single_parameter() {
+    fn:sort((1, 4, 6, 5, 3))
+};
+
+declare
+    %test:assertEquals("1","3","4","5","6")
+function testSort:single_parameter() {
```
joewiz: This is the 2nd function with the same name and arity.

test/src/xquery/xquery3/fnSort.xql
```diff
@@ -0,0 +1,172 @@
+xquery version "3.1";
+
+module namespace testSort="http://exist-db.org/xquery/test/function_sort";
+
+declare namespace test="http://exist-db.org/xquery/xqsuite";
+
+declare
+    %test:assertEquals("1","3","4","5","6")
+function testSort:single_parameter() {
+    fn:sort((1, 4, 6, 5, 3))
```
joewiz: The outputs should be of the same xs:integer type, not strings, no?

shabanovd: The purpose of this test to make sure that there is no different between string vs integers

test/src/xquery/xquery3/fnSort.xql
shabanovd: fixed

test/src/xquery/xquery3/fnSort.xql
joewiz: But if the input is an integer, shouldn't the output be the same type? Why would the test expect the result to be of a different type?

joewiz: I see that the test suite itself does some type conversion: "If the function returns an atomic type, the assertion argument is cast into the same type and the two are compared using the eq operator" (http://exist-db.org/exist/apps/doc/xqsuite.xml#D1.2.6). Unless we're testing the test suite's own ability to do this type conversion, I'm not sure this test is needed. The tests that compare like types (integer-to-integer, string-to-string) should suffice.

joewiz:
As discussed on yesterday's call, @wolfgangmm noted that collations need to be tested, and @ljo said he is working on adding some tests for cases he noticed.

joewiz:
@wolfgangmm @shabanovd Does the QT3 test suite contain tests for collations with `fn:sort()` that could be used here?

@ljo Could you briefly describe the tests you plan to add? 

My hope is to have this feature complete and included in the next release, 3.3.0, currently tentatively planned for July 2, I think .

joewiz:
@wolfgangmm @tuurma Any updates on uploading the final tests?

tuurma:
Great to see it merged, thanks @shabanovd and @wolfgangmm!

Issue #1124, opened 2016-09-13
Implement adaptive output serialization


joewiz:
The XQuery 3.1 serialization spec has added a new output option, called "adaptive output."  (See
http://www.w3.org/TR/xslt-xquery-serialization-31/#adaptive-output.)  The idea behind adaptive output is "to allow any valid XDM instance to be serialized without raising a serialization error."  So if your query returns XML, JSON, map, array, functions, etc., you will see a usable rendition of these items. 

I have implemented a poor man's adaptive output in eXide's pretty-print.xql module, but it somewhat memory intensive, suffers from whitespace bugs due to its attempt to color and indent results, and limited to eXide. It would be nice to have this serialization method built-in and available more generally.

(For discussion about this and links to the eXide implementation, see this exist-open thread: http://markmail.org/message/47vaxvjjggt6kixx.)

dizzzz:
link does not work for me....

joewiz:
@dizzzz Thanks for letting me know! Fixed: Markdown didn't know that the period wasn't part of either URL.

joewiz:
Work underway at #1448.

Issue #888, opened 2016-01-14
Implement Text serialization


gspringtech:
It would be greatly appreciated if Text serialization could be added as per the XQuery Serialization specification.

http://www.w3.org/TR/xslt-xquery-serialization-31/#text-output

In particular, the output:item-separator should be supported.

joewiz:
From @gspringtech's original post to exist-open (http://markmail.org/message/khfsbkb4s4syuz5n), here is some sample code that might be of use for testing:

```xquery
xquery version "3.0";

declare namespace output  = "http://www.w3.org/2010/xslt-xquery-serialization";
declare option output:method "text";
declare option output:item-separator "-";

("abc","123")
```

@gspringtech wrote:

> In Saxon 9.x the output is "abc-123" while eXist produces the sequence of two strings.
> 
> Currently I'm writing very unpleasant string-joins to achieve a CSV output - text serialisation would do it easily.

@adamretter responded:

> Open an issue with a feature request and I will implement it, as it is a trivial change...

adamretter:
@joewiz @gspringtech So I did spend some time looking into this previously. Unfortunately eXist's serializers are a bit crufty and fixing the `item-separator` was not as simple as I had envisaged.

joewiz:
@adamretter I see, thanks for looking into it!

joewiz:
Superseded by #1483.

PR #1448, opened 2017-05-25
Adaptive output

wolfgangmm:
Implement adaptive output mode as defined by [XQuery 3.1 serialization](https://www.w3.org/TR/xslt-xquery-serialization-31/#adaptive-output) spec. See tests for examples.

### DIFF
--- a/src/org/exist/util/serializer/AbstractSerializer.java
+++ b/src/org/exist/util/serializer/AbstractSerializer.java
@@ -51,7 +51,7 @@ public abstract class AbstractSerializer {
 
     public void setOutput(Writer writer, Properties properties) {
         if (properties == null) {
-            outputProperties = defaultProperties;
+            outputProperties = new Properties(defaultProperties);
         } else {
             outputProperties = properties;
         }
--- a/src/org/exist/util/serializer/AbstractSerializer.java
+++ b/src/org/exist/util/serializer/AbstractSerializer.java
@@ -30,7 +30,7 @@ public abstract class AbstractSerializer {
         new HTML5Writer()
     };
 
-    private final static Properties defaultProperties = new Properties();
+    protected final static Properties defaultProperties = new Properties();
 
     static {
         defaultProperties.setProperty(OutputKeys.ENCODING, "UTF-8");
--- a/src/org/exist/storage/serializers/EXistOutputKeys.java
+++ b/src/org/exist/storage/serializers/EXistOutputKeys.java
@@ -24,6 +24,11 @@ package org.exist.storage.serializers;
 
 public class EXistOutputKeys {
 
+    /**
+     * Parameter "item-separator" from the XQuery serialization spec 3.1
+     */
+    public final static String ITEM_SEPARATOR = "item-separator";
+
 	public final static String OUTPUT_DOCTYPE = "output-doctype";
 	 
 	public final static String EXPAND_XINCLUDES = "expand-xincludes";
--- a/src/org/exist/util/serializer/AdaptiveSerializer.java
+++ b/src/org/exist/util/serializer/AdaptiveSerializer.java
@@ -20,35 +20,14 @@
 
 package org.exist.util.serializer;
 
-import org.exist.Namespaces;
-import org.exist.dom.QName;
-import org.exist.dom.memtree.NamespaceNode;
 import org.exist.storage.DBBroker;
 import org.exist.storage.serializers.EXistOutputKeys;
-import org.exist.storage.serializers.Serializer;
-import org.exist.xquery.FunctionSignature;
-import org.exist.xquery.InlineFunction;
 import org.exist.xquery.XPathException;
-import org.exist.xquery.functions.array.ArrayModule;
-import org.exist.xquery.functions.array.ArrayType;
-import org.exist.xquery.functions.map.AbstractMapType;
-import org.exist.xquery.functions.map.MapModule;
-import org.exist.xquery.functions.math.MathModule;
-import org.exist.xquery.value.*;
-import org.w3c.dom.Attr;
+import org.exist.xquery.value.Sequence;
 import org.xml.sax.SAXException;
-import org.xml.sax.SAXNotRecognizedException;
-import org.xml.sax.SAXNotSupportedException;
 
-import javax.xml.transform.OutputKeys;
 import javax.xml.transform.TransformerException;
-import java.io.IOException;
 import java.io.Writer;
-import java.text.DecimalFormat;
-import java.text.DecimalFormatSymbols;
-import java.util.Iterator;
-import java.util.Locale;
-import java.util.Map;
 import java.util.Properties;
 
 /**
@@ -62,7 +41,7 @@ public class AdaptiveSerializer extends AbstractSerializer {
     private final static String DEFAULT_ITEM_SEPARATOR = "\n";
 
     private final DBBroker broker;
-    private Writer writer;
+    private AdaptiveWriter adaptiveWriter;
 
     public AdaptiveSerializer(final DBBroker broker) {
         super();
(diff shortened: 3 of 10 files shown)


wolfgangmm:
Ok, I added another test for a sequence of mixed types and sure enough stumbled across a bug in maps :-)

joewiz:
There's also a result in the qt3 test suite that is admittedly weird but I thought I'd mention. Here's the test case:

```xml
<test-case name="Serialization-adaptive-31" covers="adaptive-serialization">
    <description>Test the adaptive serialization output method - map with duplicate
        keys</description>
    <created by="Michael Kay" on="2015-07-13"/>
    <modified by="Michael Kay" on="2015-07-18" change="Implement spec change"/>
    <dependency type="spec" value="XQ31+"/>
    <test><![CDATA[

declare option output:method "adaptive";

map{ "2015":1, 2015:2, xs:gYear('2015'):3 }

]]></test>
    <result>
        <all-of>
            <serialization-matches><![CDATA["2015":1]]></serialization-matches>
            <serialization-matches><![CDATA[2015:2]]></serialization-matches>
            <serialization-matches><![CDATA[xs:gYear\("2015"\):3]]></serialization-matches>
        </all-of>
    </result>
</test-case>
```

BaseX returns the following:

```xquery
map {
  xs:gYear("2015"): 3,
  "2015": 1,
  2015: 2
}
```

eXist returns:

```xquery
map{2015:3,2015:2,"2015":1}
```

adamretter:
@joewiz I think your indentation inheritance is probably the correct interpretation of the intention of the spec authors.

I also note in your example from the qt3, that eXist might be producing the wrong output for an `xs:gYear` map key.

@wolfgangmm thoughts?

joewiz:
@adamretter Thanks, yes, the `xs:gYear` output is what concerned me. Another similar example:

```xquery
map { xs:dayTimeDuration("P1D"): "one day" }
```

should return the same, but eXist returns:

```xquery
map{P1D:"one day"}
```

It doesn't just affect map keys, though - it affects values too:

```xquery
map { "one day": xs:dayTimeDuration("P1D") }
```

returns:

```xquery
map{"one day":P1D}
```

joewiz:
Also, since eXist now supports the `item-separator` parameter for this serialization method, perhaps this test should be added too (it passes!):

```xml
<test-case name="Serialization-adaptive-15" covers="adaptive-serialization">
    <description>Test the adaptive serialization output method - item-separator</description>
    <created by="Andrew Coleman" on="2015-02-11"/>
    <dependency type="spec" value="XQ31+"/>
    <test><![CDATA[

declare option output:method "adaptive";
declare option output:item-separator "-";

(1,2,3)

]]></test>
    <result>
        <serialization-matches>1-2-3</serialization-matches>
    </result>
</test-case>
```

joewiz:
Speaking of `item-separator`, see https://github.com/eXist-db/exist/issues/888. 

@wolfgangmm With this PR have you also basically implemented the text serialization method?

wolfgangmm:
@joewiz @adamretter: I was thinking about indentation while implementing the feature, but came to the opposite conclusion: the spec says that the indentation parameter would be passed on to delegate output methods. This is true for XML and handled by my implementation. But: there are explicit rules for maps and arrays, which are clearly not output in JSON notation. Consequently, formatting maps and arrays is not delegated to the JSON serializer, but handled by the adaptive output method itself. And for this the spec says no indentation applies. To the contrary: the text is quite clear about how a map or array are to be written. So according to my interpretation, writing "map {" instead of "map{" as given in the specs would be wrong.

joewiz:
@wolfgang I see your point, and I also could find no explicit basis for indenting maps and arrays in the spec. The qt3 tests also show no whitespace in map or array constructors. I might put the question to xquery-talk. But for the purposes of this PR, I withdraw my points about indentation.

joewiz:
@wolfgangmm Liam Quinn responded to my question about this on xquery-talk, saying, "So as I read it you have as implementor a license to pass the indent parameter down and to use JSON serialization for maps and arrays and hence to allow the readable indenting."  He also noted that his response was his personal opinion, and that if we want an official working group response, we should post to the w3c bugzilla tracking system.  I'm happy to do so if you would like.  Assuming we get a similar response, though, would you like to tackle indent="yes" in this PR, or in a separate one?

By the way, here was my note to xquery-talk:

> I'm involved in commissioning the addition of adaptive serialization [1] to eXist [2], and I would like to confirm my reading of the spec.  It appears that adaptive serialization has no facility for indenting maps and arrays.  The ser/method-adaptive.xml tests in the qt3 test suite reflect this understanding, esp. tests 32-34.
> 
> In other words, `map { "a": "b", "c": "d" }` must be serialized as `map{"a":"b","c":"d"}` without any whitespace inserted inside the map constructor or between the map entries, and it would be *incorrect* to serialize this map as:
> 
> ```
> map { 
>     "a": "b",
>     "c": "d"
> }
> ```
> 
> Is there room for implementations to use such indentation and still call it "adaptive"?  If so, what basis is there in the spec for such behavior?
> 
> Thanks in advance for any insights into this question,
> Joe
> 
> [1] https://www.w3.org/TR/xslt-xquery-serialization-31/#adaptive-output
> [2] https://github.com/eXist-db/exist/issues/1124

And here is Liam's full response:

> Here's a personal response - use bugzilla (see the status section of
> the spec, https://www.w3.org/TR/xslt-xquery-serialization-31/
> for details) if you want a Working Group response.
> 
> The Working Group will probably close soon, however.
> 
> So, https://www.w3.org/TR/xslt-xquery-serialization-31/#JSON_INDENT
> has, for JSON,
> [[
> The indent parameter controls whether the serializer adjusts the
> whitespace in the serialized result so that a person will find it
> easier to read.
> ]]
> (the spec doesn't say _which_ person!).
> 
> OK, and now we see, in 10.1,
> [[
> For some item types the Adaptive output method will delegate
> serialization to other output methods. [...] serialization parameters, 
> if set, will be passed down to the serialization method that is
> applied  to each item in the supplied sequence.
> ]]
> 
> So as I read it you have as implementor a license to pass the indent
> parameter down and to use JSON serialization for maps and arrays and
> hence to allow the readable indenting.
> 
> Certainly that's in the _spirit_ of the adaptive method. And using a
> serialization parameter would also let you pass the tests :) although
> I don't know how important that is to you.
> 
> Again, this is based on my own reading of the spec, and being at the
> meetings and on the calls, but it's not an "official" response.
> 
> Best,
> Liam

wolfgangmm:
@joewiz thanks for clarification. At least we now know that applying indentation is not strictly forbidden :-)

This means I need to refactor this PR. It is not possible to simply delegate to the JSON writer, because not only the syntax for maps and arrays differs. Both of them may recursively contain other values which have to be output according to the adaptive method, not the JSON method (just think about a function item in a map). So I will likely create an AdaptiveWriter.

joewiz:
@wolfgangmm Indeed! Knowing that eXist uses Jackson for JSON serialization (and indentation), I can definitely understand why we can't simply delegate map and array serialization to that method.

p.s. For the record here was my reply to Liam and subsequent responses from him and Michael Kay:

> Many thanks for this helpful response about the "spirit" of the spec here.  I have one last question - do you have a sense why section 10.1.4 makes this statement?
> 
>> The indent and suppress-indentation parameters are not directly applicable to the Adaptive output method. [1]
> 
> (What throws me off a bit here, of course, is that Adaptive is the only output method that serializes maps *as* maps, so there isn't anywhere else to look for direct guidance on indentation of maps.)
> 
> I may well proceed with the bugzilla request, but if you have any further insight on this question, it would help me understand whether the current phrasing was an unintentional oversight or whether there was some other consideration that led the WG to state the above.
> 
> [1] https://www.w3.org/TR/xslt-xquery-serialization-31/#ADAPTIVE_INDENT

Michael Kay's reply:

> I think the key to this is the word "directly". For many things the adaptive methods delegates to other output methods, and in cases where the adaptive method does not itself use a serialization parameter, but the method to which it delegates does, the phrase "is not directly applicable" is used. This is explained in section 10.1 of the serialization spec.

Liam's reply:

>> I have one last question - do you have a sense why section 10.1.4
>> makes
>> this statement?
>>
>>> The indent and suppress-indentation parameters are not directly applicable to the Adaptive output method. [1]
> 
> it's saying that if the implementation delegates to another
> serialization method for part or all of the output, the indent and
> suppress-indentation parameters are to be passed on to that method.
> 
>>
>> (What throws me off a bit here, of course, is that Adaptive is the
>> only
>> output method that serializes maps *as* maps, so there isn't anywhere
>> else
>> to look for direct guidance on indentation of maps.)
> 
> In principle doesn't JSON do that too? Which is why I think delegating
> map serialization to the JSON method might make sense.
> 
> If you do plan to file an issue against the spec, do it soon - I think
> the XQuery Working Group might only have two or three more meetings
> before it closes. XSLT will be going perhaps a little longer but is
> focused on XSLT 3.

dizzzz:
Is this PR still work-in-progress?

joewiz:
Yes.  Wolfgang will be adding support for indentation.  (I've applied the "in progress" label to help clarify the current status of the PR.)

wolfgangmm:
I refactored the code and added a lot more tests from the XQ3.1 test suite. The PR is now ready.

joewiz:
The added tests look good, and I've built and tested adaptive serialization with and without indentation. I also have a quickly hacked version of eXide* that uses the new "native" adaptive serialization. It's snappy and works well from my perspective. I defer to the core devs on the code review. 

\* All eXide (and atom) need now are a way to apply xqlint's syntax highlighting to the results.

src/org/exist/util/serializer/AbstractSerializer.java
```diff
@@ -30,7 +30,7 @@
         new HTML5Writer()
     };
 
-    private final static Properties defaultProperties = new Properties();
+    protected final static Properties defaultProperties = new Properties();
```
adamretter: I am wondering if the fact that this is static would cause concurrent serializations to screw with each others properties? e.g. `AdaptiveSerialization#setOutput` I think could be called by concurrent XQueries?

wolfgangmm: The defaultProperties are only used if no other set of properties has been specified and they are not modified. If you call setOutput, you will pass in your own properties.

adamretter: Okay, I think that maybe the following change from:

```java
if (properties == null) {
    outputProperties = defaultProperties;
```

to:

```java
if (properties == null) {
    outputProperties = new Properties(defaultProperties);
```

might be a good idea to prevent accidental modification in AdaptiveWriter or other places.

wolfgangmm: Ok, I incorporated this small change.

joewiz:
Many thanks to @wolfgangmm for adding the feature and to @adamretter for the code review. I'm looking forward to the 3.3.0 release!

joewiz:
I think I found a little bug. eXist should reject `fn:concat#1` with error `XPST0017` (https://www.w3.org/TR/xpath-30/#ERRXPST0017). Instead, it serializes `fn:concat#1` as `fn:concat#-1`.

dizzzz:
yes please

joewiz:
@dizzzz Okay, great. I've now filed the issue: https://github.com/eXist-db/exist/issues/1467.

PR #1449, opened 2017-05-25
[bugfix] fn:replace with arrow operator

### LINKED ISSUE #1357: Unexpected error "function argument placeholder not expanded" with arrow operator
### What is the problem

An unexpected error is raised when using the arrow operator:

```xquery
xquery version "3.1";

"hello" => replace("llo", "y")
```

The error (in eXide):

> exerr:EXXQDY0001 Internal error: function argument placeholder not expanded. [source: xquery version "3.1"; "hello" => replace("llo", "y")] In function: replace(xs:string?, xs:string, xs:string) [-1:-1:String]

### What did you expect

I expected the result `hey` - the result of the equivalent expression `replace("hello", "llo", "y")`.

See https://www.w3.org/TR/xquery-31/#id-arrow-operator.

See above. I also tested in BaseX, which returns the expected result, `hey`.

### Context information
- eXist-db version: 3.1.0
- Java version: 1.8.0_121-b13
- Operating system: macOS 10.12.3
- 64 bit
- Any custom changes in e.g. conf.xml: no

wolfgangmm:
[bugfix] fn:replace arguments not cleared when dynamically called. Closes #1357

### DIFF
--- a/src/org/exist/xquery/functions/fn/FunReplace.java
+++ b/src/org/exist/xquery/functions/fn/FunReplace.java
@@ -112,6 +112,7 @@ public class FunReplace extends FunMatches {
 	 * @see org.exist.xquery.Function#setArguments(java.util.List)
 	 */
 	public void setArguments(List<Expression> arguments) throws XPathException {
+	    steps.clear();
         Expression arg = arguments.get(0);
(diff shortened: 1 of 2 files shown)


joewiz:
Perfect, I can confirm this fixes the issue as originally reported. Many thanks!

PR #1459, opened 2017-06-22
[feature] Provide an extension mechanism for extension modules to provide valid XPath/XQuery error codes

adamretter:
Exceptions thrown by XQuery functions are eventually wrapped up into an `XPathException` which has an `ErrorCodes.ErrorCode`, this change allows the originating exception to provide an `ErrorCode` if one is not otherwise explicitly provided.

Such a mechanism allows extension modules to provide correct error codes. An example usage is the `expath-bin-module` extension, you can see it in use here: https://github.com/eXist-db/expath-bin-module/blob/master/src/main/java/org/exist/xquery/modules/expath/bin/RegionFilterInputStream.java#L166

### DIFF
--- a/src/org/exist/xquery/XPathErrorProvider.java
+++ b/src/org/exist/xquery/XPathErrorProvider.java
@@ -1,0 +1,35 @@
+/*
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2017 The eXist Project
+ * http://exist-db.org
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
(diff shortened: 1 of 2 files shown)


src/org/exist/xquery/XPathException.java
```diff
@@ -43,7 +43,7 @@
     private Source source = null;
 
     /**
-     * @deprecated Use a constructor with errorCode
+     * @deprecated Use a constructor with errorCodex
```
dizzzz: typo?

adamretter: Yup. One moment...

src/org/exist/xquery/XPathErrorProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author Adam Retter <adam@exist-db.org>
+ */
+public interface XPathErrorProvider {
+
+    /**
+     * Gets the XPath Error code
+     *
+     * @return The error code or {@link ErrorCodes#ERROR} if there is no defined error code
+     */
+    ErrorCodes.ErrorCode getErrorCode();
```
dizzzz: You could consider to provide a 'default' implementation that returns ErrorCodes#ERROR when not implemented...

http://www.oracle.com/webfolder/technetwork/tutorials/obe/java/JavaSE8DefaultMethods/JavaSE8DefaultMethods.html

adamretter: Hmm... my experience with this has lead to interoperability issues with other JVM languages.

src/org/exist/xquery/XPathException.java
adamretter: Fixed!

adamretter:
@dizzzz Should be good to go

src/org/exist/xquery/XPathErrorProvider.java
dizzzz: hmmmmmmm ok. interesting....

PR #1440, opened 2017-05-15
fix lucene inline elements in case of not yet applied whitespaces

ubbo:
### Description:
The inline element configuration for lucene fails if an empty element is before the element to be inlined. 

For elements like this:
```xml
        <p><em>trampel</em>tier</p>
        <p><a name="1"></a><em>trampel</em>tier</p>
```
with a config like this:
```xml
            <lucene>
                <text qname="p">
                    <inline qname="em"/>
                </text>
            </lucene>
```

Only one match for the query `//p[ft:query(., $query)]` will be found.

This happens because in the DefaultTextExtractor the whitespace is not applied before, but in the middle of the word. 

This patch fixes this behaviour by resetting the variable "adSpaceBeforeNext" when an inline element starts.

### Reference:

https://github.com/eXist-db/exist/issues/1439

### Type of tests:

```xquery
xquery version "3.1";

module namespace fti="http://exist-db.org/xquery/ft-inline/test";

declare namespace test="http://exist-db.org/xquery/xqsuite";
declare namespace stats="http://exist-db.org/xquery/profiling";

declare variable $fti:COLLECTION_CONFIG := 
    <collection xmlns="http://exist-db.org/collection-config/1.0">
        <index xmlns:xs="http://www.w3.org/2001/XMLSchema">
            <lucene>
                <text qname="p">
                    <inline qname="em"/>
                </text>
            </lucene>
        </index>
    </collection>;

declare variable $fti:DATA :=
    <root>
        <p>trampeltier</p>
        <p><em>trampel</em>tier</p>
        <p><a name="1"></a><em>trampel</em>tier</p>
    </root>;
    
declare variable $fti:COLLECTION_NAME := "inlinetest";
declare variable $fti:COLLECTION := "/db/" || $fti:COLLECTION_NAME;

declare
    %test:setUp
function fti:setup() {
    xmldb:create-collection("/db/system/config/db", $fti:COLLECTION_NAME),
    xmldb:store("/db/system/config/db/" || $fti:COLLECTION_NAME, "collection.xconf", $fti:COLLECTION_CONFIG),
    xmldb:create-collection("/db", $fti:COLLECTION_NAME),
    xmldb:store($fti:COLLECTION, "test.xml", $fti:DATA)
};

declare
    %test:tearDown
function fti:cleanup() {
    xmldb:remove($fti:COLLECTION),
    xmldb:remove("/db/system/config/db/" || $fti:COLLECTION_NAME)
};

(:~
[...]

PR #1460, opened 2017-06-23
rework: fix lucene inline elements in case of not yet applied whitespaces

dizzzz:
rework of https://github.com/eXist-db/exist/pull/1440

added testcase


dizzzz:
inline_elements_xql_-_exist_-____development_github_exist_

dizzzz:
@wolfgangmm I am not sure that when the test suite is executed, these lucene tests are executed to. Im a wrong? I don't see the "xquery/ft-inline/test" suite namespace in the testlog; the new one or the existing ones

Issue #1437, opened 2017-05-10
Changes in white space handling 3.0.0 - 3.2.0


wsalesky:
Change in white space handling when accessing files with webdav. Added white space after ‘pretty printed’ elements, where there should only be a line break. 

eXistdb versions:  3.0, 3.1.0 and 3.1.1 and 3.2.0, fine before version 2.2

Behavior noticed when files are edited via webdav, notably in oXygen. To reproduce the problem, deploy this small sample repository: 

https://github.com/wsalesky/white-space

Open the templates/page.html via webdav, make any small edit and re-save. You will notice the SVG icon now has added spaces and no longer aligns correctly. The icon code is on line 47 and is automatically pretty printed even if I save all the span elements as a single line. 

Correct:
<img width="361" alt="screen shot 2017-05-09 at 10 47 41 pm" src="https://cloud.githubusercontent.com/assets/1168206/25881222/ba1c8a2c-3509-11e7-93ba-16421d4ae46b.png">

Incorrect, after editing file via oXygen:
<img width="503" alt="screen shot 2017-05-09 at 10 47 25 pm" src="https://cloud.githubusercontent.com/assets/1168206/25881234/c8a4bbc8-3509-11e7-9fd5-1f9572010209.png">

dizzzz:
is the change also visible in the webdav interface, or only when rendering?

dizzzz:
@ljo ?

wsalesky:
I only see it when rendering, not in the webdav interface. I think it happens when the record is saved to the db via webdav. I tested this by opening the record in oxygen. Close without any changes, the white space is fine. Make a small change anywhere in the document and save and the white space issue appears.

joewiz:
@wsalesky Could you try editing `extensions/webdav/webdav.properties` to change [line 4](https://github.com/eXist-db/exist/blob/develop/extensions/webdav/webdav.properties#L4) of this file to:

    indent=no

(Removing the `#` to uncomment the line, of course.)

Then restart eXist, open your test file, make your edits, and save via WebDAV. I believe the problem you were seeing will go away. If so, then we can confirm that what you're seeing is a result of the WebDAV interface's serialization policy—which unless specified in the `webdav.properties` file is inherited from conf.xml's `serializer/@indent=yes` setting. 

I've been gathering some notes on whitespace-related settings and options in eXist, which I've posted here: https://gist.github.com/joewiz/b53612a8b7744b8d1d24d2028770a79d. 

[...]

PR #1466, opened 2017-06-26
WebDAV: use global serialisation parameters

dizzzz:
The webDAV code used hardcoded settings which could by (optionally) changed by using a properties files. With this change the default serialisation parameters are used as defined in config.xml

Fixes https://github.com/eXist-db/exist/issues/1437 reported by @wsalesky

### DIFF
--- a/extensions/webdav/src/org/exist/webdav/ByteCountOutputStream.java
+++ b/extensions/webdav/src/org/exist/webdav/ByteCountOutputStream.java
@@ -25,7 +25,6 @@ import java.io.IOException;
 import java.io.OutputStream;
 
 /**
- *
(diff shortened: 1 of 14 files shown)


joewiz:
@dizzzz Sounds like a nice improvement! To confirm, with your change, webdav will now use eXist's defaults (e.g., as defined in `$EXIST_HOME/conf.xml` under `serializer/@indent`), except as overridden by a `$EXIST_HOME/webdav.properties` file (modeled on the file in `$EXIST_HOME/extensions/webdav` of the same name). Is that correct?

dizzzz:
@joewiz Yes that is exactly what has been modified.....

PR #1469, opened 2017-06-27
[bugfix] Fix JMX interface

wolfgangmm:
After change to mxbean (in 5edf64625aa0df4f825be70630cfaf7444166cd4), information on running queries, query history, jobs, locks etc. was no longer provided. This not only affected eXist's JMXServlet but also jconsole and other JMX clients.

### DIFF
--- a/src/org/exist/management/client/JMXtoXML.java
+++ b/src/org/exist/management/client/JMXtoXML.java
@@ -392,6 +392,8 @@ public class JMXtoXML {
         if (object instanceof TabularData) {
             serialize(builder, (TabularData) object);
 
+        } else if (object instanceof CompositeData[]) {
+            serialize(builder, (CompositeData[]) object);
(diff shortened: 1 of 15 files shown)


dizzzz:
Is it possible to test this automagically?

wolfgangmm:
Thought about it, but it would be difficult to write an automated test as useful information is only provided if there's a certain load on the server which would need to be triggered concurrently.

adamretter:
I am not quite clear why all these changes were required. Exporting the existing data via Tabular Data worked just fine. Is it rather that you also fixed the issue and then enhanced the output by using Java Beans instead of TabularData?

wolfgangmm:
It looks like directly returning TabularData is not possible using mxbeans. Instead of the data, you only got the type information. Anyway, using Java Beans results in more readable code, so there's a benefit in the end.

PR #1470, opened 2017-06-27
[bugfix] Do not destory LocalVariables if they are referenced in the `return` expr

adamretter:
Previously an `org.exist.xquery.LocalVariable` could have been incorrectly destroyed during a running XQuery if it was considered *"out of scope"*.

This destruction occurred in `XQueryContext#popLocalVariables(LocalVariable, Sequence)`. However, it failed to take into account whether the variable was referenced in the `return` expression. In practice, this was only implemented for `xs:base64binary`, and so was rarely seen.

This PR modifies `XQueryContext#popLocalVariables(LocalVariable, Sequence)`, so that it visits every sub-expression of the `return` expression to determine if the variable is referenced, if it is then it cannot be destroyed, otherwise it would not be available when it is needed.

@wolfgangmm I was unsure of the best way to fix this, although my solution appears to work. I would appreciate your input, as if I am not mistaken your commit 39f07de may have introduced this bug. Please let me know if there is a better (more efficient solution).

I have test cases within the [expath-bin-module](https://github.com/eXist-db/expath-bin-module) to prove the correctness of this, please see the following two unit tests:

1. https://github.com/eXist-db/expath-bin-module/blob/master/src/test/java/org/exist/xquery/modules/expath/bin/BasicFunctionsTest.java#L779

2. https://github.com/eXist-db/expath-bin-module/blob/master/src/test/java/org/exist/xquery/modules/expath/bin/BasicFunctionsTest.java#L820


dizzzz:
@adamretter any idea about the build issue?

adamretter:
@dizzzz This seems to have affected `group by`, however I think it might indicate a more endemic bug in the implementation of `group by`. I am hoping that @wolfgangmm can offer some input?

dizzzz:
@adamretter it is one a dependancy issue between jargo and the jar.,.. from jargo's `pom.xml` : 

inline_elements_xql_-_exist_-____development_github_exist_

adamretter:
@dizzzz Nope, that issue was fixed previously. This is a different one.

src/org/exist/xquery/XQueryContext.java
```diff
@@ -2566,24 +2567,42 @@ public LocalVariable markLocalVariables( boolean newContext )
...
      *
-     * @param  var
-     *
+     * @param var The variable
+     * @param returnExpr the return expression or null
+     * @param resultSeq the result sequence or null
      */
-    public void popLocalVariables(LocalVariable var, Sequence resultSeq)
-    {
+    public void popLocalVariables(final LocalVariable var, @Nullable final Expression returnExpr, @Nullable final Sequence resultSeq) {
```
wolfgangmm: ModuleContext inherits from XQueryContext and delegates popLocalVariables to its parent context. The newly introduced methods should thus be overwritten there as well to delegate to parentContext.

This is causing the failures in the group by test (which have nothing to do with group by itself).

adamretter:
@wolfgangmm I made the suggested changes. All tests now pass. Any updates from you please?

wolfgangmm:
I need some help as I cannot run the tests you referenced. I'd like to understand where the issue really is. Because popLocalVariables is easily called a million times during an average query, we need to make sure we limit additional overhead to where it is really necessary.

adamretter:
@wolfgangmm What seems to be wrong with the tests I referenced?

It should be as simple as:
```bash
$ git clone https://github.com/eXist-db/expath-bin-module.git
$ cd expath-bin-module
$ mvn clean compile test
```

What is the problem that you are experiencing?

PR #1477, opened 2017-07-09
Fix memory leaks

wolfgangmm:
Closes several memory leaks found by extensive stress testing. This affects nearly all applications using library modules, but because the leaked objects are small on average, the time it takes for the growing memory usage to become manifest depends on average load and complexity of the app (typically a server may run for several weeks until it becomes slow and eventually crashes).  Another leak applies to inline functions, which are quite common now in XQuery 3.1 code.

The fixes in detail:

1. function and variable declarations in imported modules not cleared: resetState was only called on functions and variables actually referenced from the importing module. Make sure it is called on *all* functions and variable declarations
2. inline function closure variables not cleared: a new function call is returned by every evaluation of the inline expression and closure variables need to freed on each call after query execution
3. some internal modules like the http request, response and session modules may declare variables or otherwise keep state during query execution. Always clear state afterwards

The last issue above requires a minor API change to org.exist.xquery.Module and would thus break compatibility for some external packages. To preserve backwards compatibility, the PR only deprecates the old method (with one argument).


src/org/exist/xquery/ExternalModuleImpl.java
```diff
@@ -270,6 +270,11 @@ public boolean moduleIsValid(DBBroker broker) {
         return mSource != null && mSource.isValid(broker) == Source.VALID;
     }
 
+    @Override
+    public void reset(XQueryContext context) {
```
adamretter: How about adding an `@deprecated` annotation? to the interface and `@Deprecated` javadoc?

wolfgangmm: The interface has @deprecated already.

src/org/exist/xquery/ExternalModuleImpl.java
```diff
@@ -268,9 +270,25 @@ public boolean moduleIsValid(DBBroker broker) {
...
+            needsReset = false;
+            mContext.reset(keepGlobals);
+            if (!keepGlobals) {
+                mStaticVariables.clear();
+                // reset state of variable declarations
+                mGlobalVariables.values().forEach(v -> v.resetState(true));
+            }
+            // reset state of declared functions
+            mFunctionMap.values().forEach(f -> f.resetState(true));
+            needsReset = true;
```
adamretter: Can you explain why when `needsReset` is `true`, it is set to `false` and then back to `true` again?

src/org/exist/xquery/InlineFunction.java
```diff
@@ -41,6 +42,7 @@
 	public final static QName INLINE_FUNCTION_QNAME = new QName("", "");
 	
 	private UserDefinedFunction function;
+	private Stack<FunctionCall> calls = new Stack<>();
```
adamretter: Do you really mean `Stack`? `ArrayDeque` is much cheaper and provides the same interface if you don't need synchronization.

src/org/exist/xquery/ExternalModuleImpl.java
wolfgangmm: Nested module imports would lead to an infinite loop without setting flag to prevent this.

src/org/exist/xquery/InlineFunction.java
```diff
@@ -96,6 +96,12 @@ public int returnsType() {
     public void resetState(boolean postOptimization) {
         super.resetState(postOptimization);
         // clear closure variables set on inline function
-        function.setClosureVariables(null);
+        calls.forEach(call -> {
```
adamretter: I think this could be reformulated to be more efficient, i.e.:

```java
if (!postOptimization) {
  calls.forEach(call -> call.getFunction().setClosureVariables(null));
}
```

src/org/exist/xquery/XQueryContext.java
```diff
@@ -2045,9 +2030,10 @@ public boolean isVarDeclared( QName qname )
      * 
      * @return currently visible local variables as a stack
      */
-    public List<Variable> getLocalStack() {
-    	final List<Variable> variables = new ArrayList<Variable>(10);
-    	
+    public List<ClosureVariable> getLocalStack() {
+
+        final List<ClosureVariable> closure = new ArrayList<>(6);
```
adamretter: Why `10` before and now why `6`?

src/org/exist/xquery/ExternalModuleImpl.java
adamretter: I see :-)

src/org/exist/xquery/XQueryContext.java
wolfgangmm: This is an arbitrary choice. Depends on how many let/for or function parameters you expect on average.

adamretter: Ok

PR #1479, opened 2017-07-13
Ensure that we unlock the same documents we lock

adamretter:
`DocumentSet#lock` can throw a lock exception. If that occurs, it may have only locked some of the documents within the document set, the corresponding call to `DocumentSet#unlock` would attempt to release locks on all documents; that is incorrect as it could release locks that it doesn't own!

We now record the actual locks taken, and only release those that we acquired.


adamretter:
This changes an internal API, so should be part of a Major release?

dizzzz:
I cannot see the consequence of this change. For me, if my extension keep working I am happy :-)
I'll check this branch....

PR #1490, opened 2017-07-25
Fix random crashes on Mac

wolfgangmm:
eXist's system tray launcher used the yajsw Java libraries to control eXist as a service. On a Mac this led to random crashes of the Java VM, likely due to library dependencies required by yajsw.  The .dmg install was thus plain unusable for most Mac users (for the last two releases of eXist).

This PR refactors the launcher to not rely on yajsw directly and thus avoid any jar dependencies on it. Instead it manages services by calling the external shell/bat scripts. Removing the yajsw dependency fixed the instability on Mac.

### DIFF
--- a/src/org/exist/launcher/Launcher.java
+++ b/src/org/exist/launcher/Launcher.java
@@ -327,9 +327,18 @@ public class Launcher extends Observable implements Observer {
             }
 
             popup.addSeparator();
-            quitItem = new MenuItem("Quit (and stop server)");
+            quitItem = new MenuItem("Quit");
             popup.add(quitItem);
-            quitItem.addActionListener(actionEvent -> shutdown(false));
+            quitItem.addActionListener(actionEvent -> {
+                if (serviceManager.isInstalled()) {
+                    if (tray != null) {
+                        tray.remove(trayIcon);
+                    }
+                    System.exit(SystemExitCodes.OK_EXIT_CODE);
+                } else {
+                    shutdown(false);
+                }
+            });
         }
         return popup;
     }
@@ -384,7 +393,7 @@ public class Launcher extends Observable implements Observer {
     protected void shutdown(final boolean restart) {
         utilityPanel.setStatus("Shutting down ...");
         SwingUtilities.invokeLater(() -> {
-            if (serviceManager.isRunning()) {
+            if (serviceManager.isRunning() && restart) {
                 if (serviceManager.stop()) {
                     serviceManager.start();
                     trayIcon.displayMessage(null, "Database stopped", TrayIcon.MessageType.INFO);
--- a/src/org/exist/launcher/Launcher.java
+++ b/src/org/exist/launcher/Launcher.java
@@ -50,6 +50,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.*;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * A launcher for the eXist-db server integrated with the desktop.
@@ -92,6 +93,8 @@ public class Launcher extends Observable implements Observer {
         SwingUtilities.invokeLater(() -> new Launcher(args));
     }
 
+    private final ReentrantLock serviceLock = new ReentrantLock();
+
     private ServiceManager serviceManager;
     private SystemTray tray = null;
     private TrayIcon trayIcon = null;
@@ -102,7 +105,7 @@ public class Launcher extends Observable implements Observer {
     private ConfigurationDialog configDialog;
     private boolean isInstallingService = false;
 
-    public Launcher(final String[] args) {
+    Launcher(final String[] args) {
         if (SystemTray.isSupported()) {
             tray = SystemTray.getSystemTray();
         }
@@ -112,7 +115,7 @@ public class Launcher extends Observable implements Observer {
         this.jettyConfig = getJettyConfig();
 
         serviceManager = new ServiceManager(this);
-
+        
         boolean initSystemTray = true;
         if (isSystemTraySupported()) {
             initSystemTray = initSystemTray();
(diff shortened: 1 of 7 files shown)


tools/yajsw/bin/installDaemon.sh
```diff
@@ -36,7 +36,7 @@ fi
 if [ -z "${JAVA_HOME}" ]; then
   echo -e "\nNo JAVA_HOME environment variable found!"
   echo "Attempting to determine JAVA_HOME (if this fails you must manually set it)..."
-  java_bin=$(readlink -f `which java`)
+  java_bin=$(readlink `which java`)
```
adamretter: Why was `-f` removed?

`-f` on Linux (GNU CoreUtils) is quite different to Mac (BSD). I cannot see an equivalent on Mac. I think this needs to be switched based on OS, so that we still use `-f` on Linux but on Mac we do whatever the equivalent is.

src/org/exist/launcher/ServiceManager.java
```diff
@@ -0,0 +1,249 @@
...
+
+        pb.directory(home.orElse(Paths.get(".")).toFile());
+        pb.redirectErrorStream(true);
+        if (consumer == null) {
+            pb.inheritIO();
+        }
+        try {
+            final Process process = pb.start();
+            if (consumer != null) {
+                final StringBuilder output = new StringBuilder();
```
adamretter: It would seem to me that Java's NIO.2 [`Files.readAllLines`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#readAllLines%28java.nio.file.Path,%20java.nio.charset.Charset%29) or even [`Files.readAllBytes`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#readAllBytes%28java.nio.file.Path%29) would be a better choice here.

src/org/exist/launcher/ServiceManager.java
```diff
@@ -0,0 +1,249 @@
...
+
+    boolean canUseServices() {
+        return canUseServices;
+    }
+
+    void queryState() {
+        if (canUseServices) {
+            runWrapperCmd("query", (code, output) -> {
+                if (code == 0) {
+                    Pattern statusRegex = Pattern.compile("^Installed\\s*:\\s*(.*)$|^Running\\s*:\\s*(.*)$", Pattern.MULTILINE);
```
adamretter: `Pattern` can be `static final` as a class member

src/org/exist/launcher/Launcher.java
```diff
@@ -450,70 +381,15 @@ private void setServiceState() {
...
-                                break;
-                            }
-                        }
-                        if (!service.isRunning() && service.start()) {
-                            trayIcon.displayMessage(null, "Database restarted", TrayIcon.MessageType.INFO);
-                        } else {
-                            trayIcon.displayMessage(null, "Failed to restart. Please start service manually.", TrayIcon.MessageType.INFO);
-                        }
-                    }
+            if (serviceManager.isRunning()) {
```
adamretter: If I am not mistaken doesn't `SwingUtilities.invokeLater` cause the `Runnable` to be executed on a different thread (i.e. the AWT event dispatcher)? If so, I think that `ServiceManager` needs to be made thread-safe as its state is read and modified by different threads?

src/org/exist/launcher/Launcher.java
```diff
@@ -97,17 +92,15 @@ public static void main(final String[] args) {
         SwingUtilities.invokeLater(() -> new Launcher(args));
     }
 
+    private ServiceManager serviceManager;
     private SystemTray tray = null;
     private TrayIcon trayIcon = null;
-    private Optional<WrappedService> runningAsService;
     private SplashScreen splash;
     private Optional<JettyStart> jetty = Optional.empty();
```
adamretter: It looks like `jetty` is read and written from different states. It needs some sort of synchronization mechanism, it might be possible to get away with `volatile` I haven't checked the semantics of all the read/write paths to it, otherwise and `AtomicReference` or `synchronized` is needed.

src/org/exist/launcher/ServiceManager.java
wolfgangmm: We're reading the input stream of a process here, so Files.readXXX does not match.

src/org/exist/launcher/Launcher.java
wolfgangmm: The only write to the Optional may happen during initialization while the splash screen is shown. Additionally, the called methods of JettyStart itself are synchronized and conflicting menu items for start/stop are greyed out.

src/org/exist/launcher/Launcher.java
wolfgangmm: As far as I can see, all calls to ServiceManager outside the Launcher constructor will be running in the AWT event dispatcher, but it doesn't cost much to synchronize this, so I'll do.

src/org/exist/launcher/ServiceManager.java
adamretter: Doh!

src/org/exist/launcher/Launcher.java
adamretter: If I search for `serviceManager.` then I see several places where it looks to me like the internal state of `ServiceManager` is called from outside the AWT event dispatcher.

src/org/exist/launcher/Launcher.java
adamretter: How about `Launcher#installAsService`?

wolfgangmm: Ok, I added a serviceLock to protect calls to either jetty or serviceManager.

src/org/exist/start/start.config
```diff
@@ -104,10 +104,4 @@ tools/jetty/lib/*                                  mode == standalone
...
 tools/ircbot/lib/*                                 mode == jetty
 tools/aspectj/lib/*                                always
-tools/yajsw/wrapperApp.jar                         mode == other
-tools/yajsw/wrapper.jar                            mode == other
-tools/yajsw/lib/extended/velocity/velocity-%latest%.jar         mode == other
-tools/yajsw/lib/core/commons/commons-configuration2-%latest%.jar mode == other
-tools/yajsw/lib/core/commons/commons-lang3-%latest%.jar          mode == other
-tools/yajsw/lib/core/netty/netty-all-4.0.35.Final.jar           mode == other
-tools/yajsw/lib/core/jna/jna-platform-%latest%.jar              mode == other
+tools/yajsw/lib/core/commons/commons-configuration2-%latest%.jar mode == other
```
wolfgangmm: yajsw expects this inside its directory, so it cannot be moved without major effort.

dizzzz: This is the sole jar referenced from start.conf? Intriguing :-)

dizzzz:
Happy with the change, though reviewing is a bit difficult for me.

dizzzz:
All concerns have been addressed...

src/org/exist/start/start.config
adamretter: This is not correct. I will send a PR. Unfortunately this means 3.4.0 is somewhat broken for developers, i.e. it won't work in Eclipse or NetBeans or outside of YAJSW (i.e. when `mode == other`). It also has an affect on the Maven artifacts, as they were not updated to reflect this change either.

I think we will need a 3.4.1... :-/

dizzzz:
Ow cr@p  :(

src/org/exist/start/start.config
adamretter: I have opened a PR here - https://github.com/eXist-db/exist/pull/1494

wolfgangmm: I don't understand where the problem is: I only removed some jar dependencies for one launch mode, not added a new one. commons-configuration2 was already referenced.

adamretter: Apologies, I thought you had added a YAJSW dependency on commons-configuration2 to start.config. I see now that it was already there and just moved. The problem was there before your change.

The problem is that `LauncherWrapper.java` which is part of the core of exist (i.e. it is compiled into `exist.jar`), has both a compile-time and run-time dependency on commons-configuration2. However, that dependency is not present in `lib/core`, it is also missing from the IDE files and also from the Maven POM. Copying the file into `lib/core` and adding the missing declarations will fix it (#1494), alongside updating the pom.xml for exist-core.

PR #1494, opened 2017-07-26
[bugfix] Fix a dependency issue in 3.4.0

adamretter:
Unfortunately a small regression was introduced in 7c92b78, which creates problems for developers wishing to use eXist-db 3.4.0. See the discussion here - https://github.com/eXist-db/exist/pull/1490/files#r129312822


dizzzz:
Should we update the list of Webstart JAR files for this?

https://github.com/eXist-db/exist/blob/develop/src/org/exist/webstart/JnlpJarFiles.java#L51

Only if it is required to run the client.....

adamretter:
@dizzzz I am pretty certain that is not required to run the client.

Issue #1475, opened 2017-07-05
Error to start Java Client 3.3.0 from Dashboard/web browser 


andray71:
Thank you for reporting your issue!

To be able to better understand you problem or suggestion, please add as much information as possible to this ticket. When there is no or just little information, it will be too difficult to work on it (and hence we will not be able to solve it)

Please fill in the following sections:

### What is the problem
> Trying to Open/Start Java Admin Client from Dashboard
Exception:
```
JNLPException[category: Security Error : Exception: null : LaunchDesc: 
<jnlp spec="7.0" codebase="http://localhost:8080/exist/webstart/" href="exist.jnlp" version="3.3.0">
  <information>
    <title>eXist XML-DB client</title>
    <vendor>exist-db.org</vendor>
    <homepage href="http://exist-db.org"/>
    <description>Integrated command-line and gui client, entirely based on the XML:DB API and provides commands for most database related tasks, like creating and removing collections, user management, batch-loading XML data or querying.</description>
    <description kind="short">eXist XML-DB client</description>
    <description kind="tooltip">eXist XML-DB client</description>
    <icon href="jnlp_logo.jpg"/>
    <icon href="jnlp_icon_128x128.gif" width="128" height="128"/>
    <icon href="jnlp_icon_64x64.gif" width="64" height="64"/>
    <icon href="jnlp_icon_32x32.gif" width="32" height="32"/>
  </information>
  <security>
    <all-permissions/>
  </security>
  <resources>
    <property name="jnlp.packEnabled" value="true"/>
    <property name="java.util.logging.manager" value="org.apache.logging.log4j.jul.LogManager"/>
    <java version="1.8+"/>
    <jar href="log4j-jul-2.8.2.jar" size="26272"/>
    <jar href="commons-codec-1.10.jar" size="307249"/>
    <jar href="cglib-nodep-3.1.jar" size="358483"/>
    <jar href="jcip-annotations-1.0.jar" size="4005"/>
    <jar href="pkg-java-fork.jar" size="87631"/>
    <jar href="commons-logging-1.2.jar" size="66579"/>
    <jar href="jargo-0.4.2-SNAPSHOT-7fa1d04.jar" size="200065"/>
    <jar href="xmlrpc-client-3.1.3.jar" size="65537"/>
    <jar href="rsyntaxtextarea-2.6.1.jar" size="1201107"/>
    <jar href="slf4j-api-1.7.21.jar" size="46264"/>
[...]

PR #1495, opened 2017-07-27
Fix the signing of core Jar files

adamretter:
Closes https://github.com/eXist-db/exist/issues/1475

### DIFF
--- a/build/scripts/build-impl.xml
+++ b/build/scripts/build-impl.xml
@@ -48,7 +48,7 @@
 
     <!-- setup conditional properties -->
     <available file="${xmldb.src}" type="dir" property="has.xmldb.src"/>
-    <available file="key.store" property="key.store.present"/>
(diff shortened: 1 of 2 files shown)


dizzzz:
I see that some echo's are removed that describe how to have the keystore file created. That might confuse some ppl, or.... are the instructions somewhere else these days?

adamretter:
@dizzzz The instructions that I removed are obsolete. They don't hold true for quite some time

dizzzz:
Are you sure? Not so long it worked fine and it was the easiest way to generate a keystore......

dizzzz:
Ah I remember issues reported on exist-open ; somehow I thought these were fixed.

The solution is very easy: just add

```xml
<property file="${basedir}/build.properties" />
<property file="${basedir}/local.build.properties" />
```

to 'jarsigner.xml' and things work again

PR #1508, opened 2017-08-01
Cache the compilation of Regular Expressions

adamretter:
We now cache the compiled `java.util.regex.Pattern` used in `fn:analyze-string`, `fn:matches`, `fn:replace` and `fn:tokenize`. Results in a significant performance improvement for repetitive invocations.


dizzzz:
please could you resolve the conflicts ? It looks like the 'optional' jars are removed?

adamretter:
@dizzzz Done.

p.s. I also use the Caffeine library in my rewrite of eXist locking and Collection caching. So whether it goes in in this PR or one following soon, I would like to have it.

dizzzz:
Ok, thnx for explaining !

joewiz:
I tested this PR and can confirm that it measurably speeds subsequent calls of regex functions. For my test query at https://gist.github.com/joewiz/5283efd0344deda0c30eff3f2d143bb7, which parses CSV files using `fn:tokenize`, `fn:replace`, and `fn:analyze-string`, the average query time was .255s. After, the average was .231s - an improvement of .024s or just under 10%. I imagine the effect would be greater as the amount of distinct regex calls in a query increases.

It's also worth noting that the PR adds caching to the regex used in `util:get-fragment-between`.

adamretter:
@joewiz The approach in `util:get-fragment-between` is not exactly caching. Rather the Regular Expression used by that function is static, so we simply just compile it once and share it between every instance of `util:get-fragment-between`.

adamretter:
@dizzzz Mergy mergy?

ben-manes:
fyi, the setting `weakValues` means that the entry is eligible for garbage collection when the pattern is not in use. The exact time depends on the GC, but it can occur in a minor collection as I think it is done eagerly. This means you might thrash a bit where it is only cached within a burst of activity and dropped when a GC occurs with inactivity in this path (references are kind of magical and their lifetimes confusing). Most likely the size bounding is already good enough since you gave a threshold and `Pattern` is a lightweight object.

adamretter:
@ben-manes So you are suggesting that I needn't use weak values here at all?

ben-manes:
I think so. I prefer weak values when I can't safely evict an entry by another constraint, e.g. due to being a stateful resource. In your case the value is immutable and you also evict by size, so weak values doesn't seem to buy you much.

PR #1512, opened 2017-08-06
Fix two regressions introduced by memory leak PR

wolfgangmm:
This PR fixes two regressions introduced by #1477:

1. in certain scenarios, the HTTP request, response and session variables were cleared before an XQuery was completed, thus causing errors. This mainly applies to the use of `util:eval`. This PR solves the issue and chooses a more general approach, by cleaning up *any* global variable declared in an internal module, thus making sure variables are always cleared (even if the module author did not think of it).
2. the changes to module cleanup in #1477 result in `resetState` being repeatedly called on module functions, which turned out to be very expensive (for complex queries there is now a measurable delay between finishing the query and returning the HTTP response). Further investigations showed that explicitely calling `resetState` on all module functions is actually unnecessary, because all used functions are already cleared via the call chain. The corresponding line in `AbstractInternalModule#resetState` can thus be removed. Module *variables* are still reset and need to be to avoid the memory leak.

### DIFF
--- a/src/org/exist/xquery/AbstractInternalModule.java
+++ b/src/org/exist/xquery/AbstractInternalModule.java
@@ -211,5 +211,8 @@ public abstract class AbstractInternalModule implements InternalModule {
     public void reset(XQueryContext xqueryContext, boolean keepGlobals) {
         // call deprecated method for backwards compatibility
         reset(xqueryContext);
+        if (!keepGlobals) {
+            mGlobalVariables.clear();
+        }
     }
 }
--- a/src/org/exist/xquery/ExternalModuleImpl.java
+++ b/src/org/exist/xquery/ExternalModuleImpl.java
@@ -285,8 +285,6 @@ public class ExternalModuleImpl implements ExternalModule {
                 // reset state of variable declarations
                 mGlobalVariables.values().forEach(v -> v.resetState(true));
             }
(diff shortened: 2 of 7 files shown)


dizzzz:
Nice work @wolfgangmm !

joewiz:
Wonderful, Wolfgang! Thank you!

dizzzz:
@adamretter pease could you have a look?

wolfgangmm:
@dizzzz well, this is quite simple because 1) fixes an obvious, reproducible issue reported by several users, and 2) removes unnecessary code. The test shows that despite the changes, relevant objects are still cleared as intended by #1477. So it's just a minor correction to that PR.

adamretter:
@dizzzz You asked me to review, and then only gave me 20 hours to do so (please keep in mind that we are on very different timezones). 

@dizzzz @wolfgangmm My review would have said: that this looks okay, but really must have some tests as this is trying to fix a regression for which there were previously no tests. As users have demonstrated there are easily reproducible test cases for this (esp. involving `util:eval`). In addition I wonder if the SQL Module, and in-fact any module which uses `ModuleUtils#storeObjectInContextMap`, may also need to have its state cleanup refactored?

@wolfgangmm Would you be able to add a few of the test cases that you used for diagnosing this please?

PR #1513, opened 2017-08-10
[bugfix] Improvements to Core Level 1 DOM compliance

adamretter:
I have fixed a number of bugs to improve the compliance of eXist's DOMs with respect to the W3C DOM Test Suites.

I have in-particular focused on eXist's persistent DOM and Level 1 Core compliance. In this area we have moved from ~18% (eXist 3.3.0) compliance to ~36% compliance.

The results can be seen here - https://github.com/exist-db/exist-domts#exist-db-201708081936-snapshot-from-adamretterexisttreebugfixdom

This PR is a mix of bug fixes and implementing missing functions. Arguably this could go into a minor release (e.g. eXist 3.5.0) as there are no API changes. However, apps which either relied on eXist's broken DOM behaviour or instigated workarounds may require some minor revisions.

In addition this PR also fixes two concurrency issues in internal pools that were found along the way.

Tests have been added where necessary to prove the implementation, but otherwise the tests of the W3C DOM Test suite (https://github.com/exist-db/exist-domts) were used to verify behaviour.


joewiz:
@adamretter Whether here or in an upcoming community call discussion, I'd be interested to know more about what motivated you to integrate this W3C DOM test suite, what problems you're uncovering, and what better conformance with the DOM will mean for eXist.

adamretter:
@joewiz So by way of further explanation:

>what motivated you to integrate this W3C DOM test suite

I needed to be able to use [XMLUnit](http://www.xmlunit.org/) with eXist-db, for the purposes of comparing XML Documents between those on a filesystem and copies of them which have been stored in the database. The underlying reason for this is a series of recovery tests that I have constructed, that seem to show problems in eXist's recovery mechanism under certain circumstances.

> what problems you're uncovering
XMLUnit makes use of the standard Java DOM and JAXP mechanisms, so it can work with any compliant DOM implementation. XMLUnit would not work correctly with eXist-db, because regrettably eXist-db's DOM implementation was lacking and quite weak where implemented (when compared against the W3C DOM Test Suites).

> what better conformance with the DOM will mean for eXist
Better interoperability with other libraries that wish to operate using JAXP against eXist. I suspect that I have also indirectly fixed a number of bugs in eXist's XQuery implementation which relies heavily on the DOM subsystem.

dizzzz:
Globally it looks OK :-) but there are many many files :-)
same changes everywhere, so I am happy :-)

PR #1516, opened 2017-08-14
Fixes for compilation on Java 9

adamretter:
This now gives  us the ability to now compile and run eXist-db on a Java 9 JVM.

eXist-db is still written in Java 8 and will still run on a Java 8 JVM as well.


src/org/exist/xquery/functions/util/FunctionFunction.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.logging.log4j.Logger;
 import org.exist.dom.QName;
 import org.exist.xquery.*;
+import org.exist.xquery.Module;
```
dizzzz: Double import??

adamretter: In Java 9 there is a `java.lang.Module`, so we have to explicitly import our one now.

dizzzz: makes sense :-)

adamretter:
@joewiz @windauer Okay this gives us the ability to now compile and run on Java 9 (alongside our existing Java 8 support).

@wolfgangmm I think this one is ready to land now.

dizzzz:
Are you sure the correct xerces and xalan versions are still on the classpath for java8 ?

dizzzz:
And loaded and used :)

adamretter:
I believe so, at runtime this code:

```java
reader.getClass().getProtectionDomain().getCodeSource().getLocation()
```

Shows me that the `XMLReader` that we use for parsing documents was loaded from the Jar file: `file:/Users/aretter/code/exist-for-release/lib/endorsed/xercesImpl-2.11.0.jar`

On a side note, you might also be interested to know that it seems that JDK 9 now includes Xerces 2.11.0, see: https://bugs.openjdk.java.net/browse/JDK-8044086

dizzzz:
Yes I know that java9 has an ‘uptodate’ version of xerces. That said, I’d vote or recommend to use the jboss/ wildfly maintained version in stead. This version has a few additional bug fixes and security fixes.

adamretter:
@joewiz Doh! I forgot that you could start eXist that way! The launcher has its own stuff, so I just fixed that and pushed it as well. Seems to work for me now.

adamretter:
I think the warnings:

`WARNING: An illegal reflective access operation has occurred`

We don't need to worry about too much, I see some from betterForm too. We can certainly fix them before they become errors.

joewiz:
@adamretter I've just tested your fix, and I can confirm that eXist now starts correctly on macOS 10.13 via `open start.jar`, `java -jar start.jar`, and `bin/startup.sh` after building with Java 9.0.1+11. This should be good to go now. @windauer or @wolfgangmm, would you care to review and merge please?

windauer:
thanks @adamretter! I just did a first test with Java 9 (macOS 10.13, 9.0.1)
* eXistdb compiles and starts \o/ 
* eXistdb tests: only DateTests.runXQueryBasedTests failed because of my German locale. Looking at the error message its obvious that the test is green as well: _arrays first differed at element [0]; expected:<11:14 [pm on Tuesday, June 26th], 2012> but was:<11:14 [ on Dienstag, Juni 26.], 2012>_ 
* Only thing I noticed is that I'm receiving following warnings in the console whenever test database is shutdown (turns up in the logs when running _build.sh test_ as well as shutting down eXistdb with _Ctrl + c_ after it was started via _bin/startup.sh_.

```
[junit] WARNING: An illegal reflective access operation has occurred
[junit] WARNING: Illegal reflective access by org.exist.storage.BrokerPool (file:/Users/windauer/dev/exist/exist/exist.jar) to field java.lang.Thread.threadLocals
[junit] WARNING: Please consider reporting this to the maintainers of org.exist.storage.BrokerPool
[junit] WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
[junit] WARNING: All illegal access operations will be denied in a future release
```

PR #1520, opened 2017-08-15
A more performant implementation of immutable maps

adamretter:
My research leads me to believe that this is the fastest implementation available for the JVM.

Could a couple of people give this a test and see if it speeds up their XQuery Maps code?

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
@@ -58,11 +58,7 @@ public class MapType extends AbstractMapType {
         keyValues.forEachRemaining(kv -> map.put(kv._1, kv._2));
         this.map = map.forked();
 
-        //TODO(AR) this is incorrect all types to be checked!
-        final long size = this.map.size();
-        if (size > 0) {
-            setKeyType(this.map.nth(size - 1).key().getType());
-        }
+        setKeyType(map);
     }
 
     public MapType(final XQueryContext context, final IMap<AtomicValue, Sequence> other, @Nullable final Integer type) {
@@ -77,11 +73,7 @@ public class MapType extends AbstractMapType {
         if (type != null) {
             this.type = type;
         } else {
-            //TODO(AR) this is incorrect all types to be checked!
-            final long size = this.map.size();
-            if (size > 0) {
-                setKeyType(this.map.nth(size - 1).key().getType());
-            }
+            setKeyType(map);
         }
     }
 
@@ -230,6 +222,20 @@ public class MapType extends AbstractMapType {
         }
         else if (type != newType) {
             type = Type.ITEM;
+        }
+    }
+
+    private void setKeyType(final IMap<AtomicValue, Sequence> newMap) {
+        for (final AtomicValue newKey : newMap.keys()) {
+            final int newType = newKey.getType();
+
+            if (type == Type.ANY_TYPE) {
+                type = newType;
+            }
+            else if (type != newType) {
+                type = Type.ITEM;
+                break; // done!
+            }
         }
     }
 
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
@@ -24,7 +24,15 @@ public class MapType extends AbstractMapType {
 
     // TODO(AR) future potential optimisation... could the class member `map` remain `linear` ?
     private IMap<AtomicValue, Sequence> map;
-    private int type = Type.ANY_TYPE;
+
+    /**
+     * The type of the keys in the map,
+     * if not all keys have the same type
+     * then this is set to {@link #MIXED_KEY_TYPES}.
+     *
+     * Uses integer values from {@link org.exist.xquery.value.Type}.
+     */
+    private int keyType = UNKNOWN_KEY_TYPE;
 
     private static IMap<AtomicValue, Sequence> newMap(@Nullable final Collator collator) {
         return new Map<>(KEY_HASH_FN, (k1, k2) -> keysEqual(collator, k1, k2));
@@ -43,7 +51,7 @@ public class MapType extends AbstractMapType {
     public MapType(final XQueryContext context, @Nullable final Collator collator, final AtomicValue key, final Sequence value) {
         super(context);
         this.map = newMap(collator).put(key, value);
-        this.type = key.getType();
+        this.keyType = key.getType();
     }
 
     public MapType(final XQueryContext context, @Nullable final Collator collator, final Iterable<Tuple2<AtomicValue, Sequence>> keyValues) {
@@ -61,7 +69,7 @@ public class MapType extends AbstractMapType {
         setKeyType(map);
     }
 
-    public MapType(final XQueryContext context, final IMap<AtomicValue, Sequence> other, @Nullable final Integer type) {
+    public MapType(final XQueryContext context, final IMap<AtomicValue, Sequence> other, @Nullable final Integer keyType) {
         super(context);
 
         if (other.isLinear()) {
@@ -70,15 +78,15 @@ public class MapType extends AbstractMapType {
 
         this.map = other;
 
-        if (type != null) {
-            this.type = type;
+        if (keyType != null) {
+            this.keyType = keyType;
         } else {
             setKeyType(map);
         }
     }
 
     public void add(final AbstractMapType other) {
-        setKeyType(other.key() != null ? other.key().getType() : Type.ANY_TYPE);
+        setKeyType(other.key() != null ? other.key().getType() : UNKNOWN_KEY_TYPE);
 
         if(other instanceof MapType) {
             map = map.union(((MapType)other).map);
@@ -102,15 +110,15 @@ public class MapType extends AbstractMapType {
         // create a transient map
         IMap<AtomicValue, Sequence> newMap = map.linear();
 
-        int prevType = type;
+        int prevType = keyType;
         for (final AbstractMapType other: others) {
             if (other instanceof MapType) {
                 // MapType - optimise merge
                 final MapType otherMap = (MapType) other;
                 newMap = map.union(otherMap.map);
 
-                if (prevType != otherMap.type) {
-                    prevType = Type.ITEM;
+                if (prevType != otherMap.keyType) {
+                    prevType = MIXED_KEY_TYPES;
                 }
             } else {
                 // non MapType
@@ -118,7 +126,7 @@ public class MapType extends AbstractMapType {
                     final AtomicValue key = entry.key();
                     newMap = newMap.put(key, entry.value());
                     if (prevType != key.getType()) {
-                        prevType = Type.ITEM;
+                        prevType = MIXED_KEY_TYPES;
                     }
                 }
             }
@@ -147,7 +155,7 @@ public class MapType extends AbstractMapType {
     @Override
     public AbstractMapType put(final AtomicValue key, final Sequence value) {
         final IMap<AtomicValue, Sequence> newMap = map.put(key, value);
-        return new MapType(this.context, newMap, type == key.getType() ? type : Type.ITEM);
+        return new MapType(this.context, newMap, keyType == key.getType() ? keyType : MIXED_KEY_TYPES);
     }
 
     @Override
@@ -179,7 +187,7 @@ public class MapType extends AbstractMapType {
         }
 
         // return an immutable map
-        return new MapType(context, newMap.forked(), type);
+        return new MapType(context, newMap.forked(), keyType);
     }
 
     @Override
@@ -217,11 +225,11 @@ public class MapType extends AbstractMapType {
     }
 
     private void setKeyType(final int newType) {
-        if (type == Type.ANY_TYPE) {
-            type = newType;
-        }
-        else if (type != newType) {
-            type = Type.ITEM;
+        if (keyType == UNKNOWN_KEY_TYPE) {
+            keyType = newType;
+
+        } else if (keyType != newType) {
+            keyType = MIXED_KEY_TYPES;
         }
     }
 
@@ -229,20 +237,20 @@ public class MapType extends AbstractMapType {
         for (final AtomicValue newKey : newMap.keys()) {
             final int newType = newKey.getType();
 
-            if (type == Type.ANY_TYPE) {
-                type = newType;
-            }
-            else if (type != newType) {
-                type = Type.ITEM;
-                break; // done!
+            if (keyType == UNKNOWN_KEY_TYPE) {
+                keyType = newType;
+
+            } else if (keyType != newType) {
+                keyType = MIXED_KEY_TYPES;
+                break; // done, we only have to detect this once!
             }
         }
     }
 
     private AtomicValue convert(final AtomicValue key) {
-        if (type != Type.ANY_TYPE && type != Type.ITEM) {
+        if (keyType != UNKNOWN_KEY_TYPE && keyType != MIXED_KEY_TYPES) {
             try {
-                return key.convertTo(type);
+                return key.convertTo(keyType);
             } catch (final XPathException e) {
                 return null;
             }
@@ -252,7 +260,7 @@ public class MapType extends AbstractMapType {
 
     @Override
     public int getKeyType() {
-        return type;
+        return keyType;
     }
 
     public static <K, V> IMap<K, V> newLinearMap() {
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
@@ -76,6 +76,38 @@ public class MapType extends AbstractMapType {
         }
     }
 
+    @Override
+    public AbstractMapType merge(final Iterable<AbstractMapType> others) {
+
+        // create a transient map
+        IMap<AtomicValue, Sequence> newMap = map.linear();
+
+        int prevType = type;
+        for (final AbstractMapType other: others) {
+            if (other instanceof MapType) {
+                // MapType - optimise merge
+                final MapType otherMap = (MapType) other;
+                newMap = map.union(otherMap.map);
+
+                if (prevType != otherMap.type) {
+                    prevType = Type.ITEM;
+                }
+            } else {
+                // non MapType
+                for (final IEntry<AtomicValue, Sequence> entry : other) {
+                    final AtomicValue key = entry.key();
+                    newMap = newMap.put(key, entry.value());
+                    if (prevType != key.getType()) {
+                        prevType = Type.ITEM;
+                    }
+                }
+            }
+        }
+
+        // return an immutable map
+        return new MapType(context, newMap.forked(), prevType);
+    }
+
     public void add(final AtomicValue key, final Sequence value) {
         setKeyType(key.getType());
         map = map.put(key, value);
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
@@ -55,7 +55,7 @@ public class MapType extends AbstractMapType {
     }
 
     public void add(final AbstractMapType other) {
-        setKeyType(other.getKey() != null ? other.getKey().getType() : Type.ANY_TYPE);
+        setKeyType(other.key() != null ? other.key().getType() : Type.ANY_TYPE);
 
         if(other instanceof MapType) {
             //TODO(AR) is the union in the correct direction i.e. keys from `other` overwrite `this`
@@ -67,8 +67,8 @@ public class MapType extends AbstractMapType {
             // create a transient map
             final IMap<AtomicValue, Sequence> newMap = map.linear();
 
-            for (final java.util.Map.Entry<AtomicValue, Sequence> entry : other) {
-                newMap.put(entry.getKey(), entry.getValue());
+            for (final IEntry<AtomicValue, Sequence> entry : other) {
+                newMap.put(entry.key(), entry.value());
             }
 
             // return to immutable map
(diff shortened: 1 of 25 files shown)


joewiz:
I've written [a test query](https://gist.github.com/joewiz/31419f45e3b597c7e2f6851cbbed50d9) designed to exercise map functionality (namely, `map:merge`, `map:entry`, `map:keys`, `map:for-each`), and after running this test several times with and without profiling enabled, I'm afraid I'm not seeing an improvement with this PR. Below is a summary of the results and screenshots of the query profiling window.

In contrast, BaseX executes the same query in an average of .550s - so there is ample room for improvement in both 3.4.1 and this PR.

|version|avg. execution time|avg. with profiling enabled|
|--|--|--| 
|3.4.1|6.9s|20.8s|
|This PR|8.3s|23.4s|

**3.4.1**
> screen shot 2017-08-17 at 12 04 50 pm

**This PR**
> screen shot 2017-08-17 at 12 01 59 pm

adamretter:
@joewiz Thanks Joe, that is some good work.

The fact that this PR is slower, makes me think that there is something wrong with my use of the new library rather than the library itself. I will need to break out the Java profiler and investigate...

So the fact that BaseX is much faster here also needs some investigation. eXist's XQuery implementation is much heavier in general (because of various additional monitoring, reporting and features). However, we should try and find out why we can't be as fast or faster than BaseX.

joewiz:
@adamretter Thank you on all points. I agree on the goal to meet/beat peer implementations like BaseX. As always, please let me know if there's anything I can do to help.

adamretter:
I need to update this for the latest Capsule - https://github.com/usethesource/capsule

adamretter:
Another options instead of Capsule which might be faster would be [Bifurcan](https://github.com/lacuna/bifurcan/blob/master/doc/comparison.md)

joewiz:
Wow, Bifurcan looks promising!

adamretter:
@joewiz I have done some work to update this... you might want to take it for a spin again!

wolfgangmm:
I'll stress test, but it will take a few days as this is time consuming.

adamretter:
@wolfgangmm As soon as you can let me know, this is the last thing I want before I do the next release.

wolfgangmm:
After checking this PR out locally, I immediately get an error on most of the default apps, e.g. documentation, monex, function docs ...

```
err:XPDY0002 undefined value for variable '$param-from-context' [at line 267, column 21, source: /Users/wolfgang/Library/Application Support/org.exist/expathrepo/shared-0.8.4/content/templates.xql]
In function:
	templates:map-argument(element(), map(*), function(*)) [249:19:/Users/wolfgang/Library/Application Support/org.exist/expathrepo/shared-0.8.4/content/templates.xql]
```

adamretter:
@wolfgangmm Uggh looks like we have a lack of tests then... I will take a look

adamretter:
@wolfgangmm Okay I fixed that one by breaking apart a much larger commit. Should be good to test now...

exist-core/src/test/xquery/maps/mapsLookup.xql
```diff
@@ -13,7 +13,7 @@ function mlt:postfix_lookup_on_map() {
 };
 
 declare
-    %test:assertEquals("Jenna", "Scott")
+    %test:assertEqualsPermutation("Jenna", "Scott")
```
line-o: That is a handy test annotation!

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
```diff
@@ -1,169 +1,252 @@
...
+                setKeyType(this.map.nth(size - 1).key().getType());
+            }
+        }
+    }
+
+    public void add(final AbstractMapType other) {
+        setKeyType(other.key() != null ? other.key().getType() : Type.ANY_TYPE);
+
+        if(other instanceof MapType) {
+            //TODO(AR) is the union in the correct direction i.e. keys from `other` overwrite `this`
```
line-o: Are you planning to remove / tackle these TODOs before this PR is merged?
We should have an answer especially for this one _before_ merging.

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
```diff
@@ -1,169 +1,252 @@
...
+        final Sequence result = map.get(key, null);
+        return result == null ? Sequence.EMPTY_SEQUENCE : result;
     }
 
     @Override
-    public AbstractMapType put(AtomicValue key, final Sequence value) throws XPathException {
-        return new MapType(this.context, this.map.assoc(key, value), type);
+    public AbstractMapType put(final AtomicValue key, final Sequence value) {
+        final IMap<AtomicValue, Sequence> newMap = map.put(key, value);
+        return new MapType(this.context, newMap, type == key.getType() ? type : Type.ITEM);
```
line-o: In what case does the key not have a type? And shouldn't it be `Type.ATOMIC` everywhere? 
Couldn't `Type.ITEM` be a node or element, which would not qualify for a map key?

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
```diff
@@ -1,169 +1,252 @@
...
-                try {
-                    PersistentTreeMap tmap = PersistentTreeMap.create(getComparator(null), null);
-                    for (final Map.Entry<AtomicValue, Sequence> entry : map) {
-                        tmap = tmap.assoc(entry.getKey(), entry.getValue());
-                    }
-                    map = tmap;
-                } catch (final XPathException e) {
-                }
-            }
+//            try {
```
line-o: Is this commented out code necessary? If not, please remove it.

exist-core/src/test/xquery/maps/mapsLookup.xql
adamretter: Right - I had to look into the source code to find it - because the docs on exist-db.org are out of date :-(

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
adamretter: I have resolved this now.

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
adamretter: Right. I just followed the type stuff that was previously there. That is too broad, I have now narrowed the types as you suggested.

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
adamretter: I have now removed it.

exist-core/src/test/xquery/maps/mapsLookup.xql
line-o: Is it documented or should we open a ticket on the documentation?

adamretter: It's documented. But eXist-db.org hasn't been updated :-(

joewiz:
@adamretter Thanks for continuing to push this forward! I see you've updated the eXist timings in my [earlier post](https://github.com/eXist-db/exist/pull/1520#issuecomment-323119325) with your results. Your new timings show that your PR improves performance compared to eXist 5.2.0 by 5% with Capsule and 10% with Bifurcan (from 2s to 1.9 or 1.8s, respectively). 

For comparison, on my system (iMac Mid 2011, 3.4 GHz Intel Core i7, 16 GB RAM), the PR performs as follows:

|version|avg. execution time|avg. with profiling enabled|
|--|--|--| 
| BaseX 9.3.1 | .6s | n/a |
| eXist-db 5.2.0 | 3.6s | 18s |
| eXist-db This PR (Bifurican 0.2.0-alpha1) | 4.5s | 18.5s |

This suggests a performance regression (of ~30%), but I don't know what to attribute it to other than the older chipset.

Upon further reflection, while my test allows for a fair, apples-to-apples comparison between BaseX and eXist, it does not test the full range of native map functionality that an XQuery 3.1 processor should have. Specifically, eXist lacks native support for the `map:find()` function, so some portion of my test is spent processing my XQuery implementation of that function, the `mf:map-find()` function. BaseX's performance improves significantly if we replace `mf:map-find()` with `map:find()`: it completes the query in .1s - 6x improvement! This isn't to point out a flaw in this PR, but rather to point out a flaw in the benchmark - and thus question the utility of this benchmark in judging the PR. On the other hand, if `map:find()` could be implemented in eXist, then the benchmark would be more useful.

joewiz:
Running the test in Saxon EE 9.9.1.5 illustrates the dilemma regarding this benchmark even more starkly:

|version|original test|test using native `map:find()`|
|--|--|--|
| BaseX 9.3.1 | .6s | .1s |
| Saxon EE 9.9.1.5 | 40s | .2s |

I wouldn't say Saxon is inefficient at performing this test. The test doesn't make use of Saxon's standards-compliant capabilities. I would conclude, therefore, that my benchmark is incapable of judging eXist's potential performance improvements with this PR.

adamretter:
@joewiz I want to leave the `map:find` stuff out of the equation right now. That is a further optimisation that we could look at in future, but right now thinking about that just muddies the water further.

I don't understand how you can see `3.6s` on 5.2.0 and then `4.5s` on this PR. I would like to get to the bottom of that. How are you doing your testing?

My methodology is to start the Java Admin Client in embedded mode using `bin/client.sh`, to paste your code into the query window, and then run that 10 or 20 times, noting the execution times and then taking an average.

adamretter:
@joewiz kind of out of scope for this PR I think, but if I rewrite your XQuery to reduce the amount of function creation I can also get a 10% improvement without this PR:

```
declare function mf:map-find($input as item()*, $key as xs:anyAtomicType) as array(*) {
	let $fn-adam := function($k, $v) { 
	     if ($k eq $key) then 
	         $v 
	     else 
	         (), 
	     mf:map-find-recurse($v, $key)
     } return

    array { mf:map-find-recurse($input, $fn-adam) }
};


(:~ 
 : A utility function that recurses through the input searching for map entries with the supplied key.
 : @param  $input An input sequence
 : @param  $key A key
 : @return  The value of any entries found to have the key
 :)
declare %private function mf:map-find-recurse($input as item()*, $fn-adam) as item()* {
    for $i in $input
    return
        if ($i instance of array(*)) then 
            mf:map-find-recurse($i?*, $fn-adam)
        else if ($i instance of map(*)) then 
            map:for-each($i, $fn-adam)
        else ()
};
```

What I am learning is that function evaluation in eXist-db seems to be very slow.

wolfgangmm:
@adamretter Trying this PR again, I did run into the following exception:

```
java.util.NoSuchElementException: null
	at java.util.ArrayList$Itr.next(Unknown Source) ~[?:?]
	at org.exist.xquery.value.ArrayListValueSequence$ArrayListValueSequenceIterator.nextItem(ArrayListValueSequence.java:436) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.EnclosedExpr.eval(EnclosedExpr.java:113) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.PathExpr.eval(PathExpr.java:280) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.ElementConstructor.eval(ElementConstructor.java:330) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.ConditionalExpression.eval(ConditionalExpression.java:100) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.DebuggableExpression.eval(DebuggableExpression.java:58) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
```

I don't have a reproducer, but hope the stack trace already helps. I'm also seeing some other failures, but maybe they are related, so one thing at a time ;-)

adamretter:
> I don't have a reproducer, but hope the stack trace already helps

Yup, I just pushed a fix.

>  also seeing some other failures

Just let me know what they are, or better yet send some tests ;-)

wolfgangmm:
@adamretter ok, the exception is fixed! I still see another issue for which I'll try to extract a test later today. It only happens occasionally but is related to maps.

wolfgangmm:
@adamretter I'm unable to reproduce my other issue, but I figured out that my code smells a bit, because I was passing in an element into the map function. This should theoretically work (and did before) as the argument should be cast to an atomic value. Anyway, if I change this by converting to string, the issue is gone.

So I can finally work on updating my stress tests and run them :-) I include my attempt to reproduce the issue below, but I don't see it standalone.

```xquery
let $model := map {
    "foo": "bar"
}
let $model := map:merge(
    (
        $model,
        map { "app": "myapp" }
    )
)
let $node := <a href="${{app}}/index.html">Start</a>
return
    element { node-name($node) } {
        for $attr in $node/@*
        return
            if (matches($attr, "\$\{[^\}]+\}")) then
                attribute { node-name($attr) } {
                    string-join(
                    let $parsed := analyze-string($attr, "\$\{([^\}]+?)(?::([^\}]+))?\}")
                    for $token in $parsed/node()
                    return
                        typeswitch($token)
                            case element(fn:non-match) return $token/string()
                            case element(fn:match) return
                                let $paramName := $token/fn:group[1]
                                let $default := $token/fn:group[2]
                                let $found := [
                                    request:get-parameter($paramName, $default),
                                    $model($paramName)
                                ]
                                return
                                    array:fold-right($found, (), function($in, $value) {
                                        if (exists($in)) then $in else $value
                                    })
                            default return $token
                    )
                }
            else
                $attr
    }
```

Changing

```xquery
let $paramName := $token/fn:group[1]
```

to

```xquery
let $paramName := $token/fn:group[1]/string()
```

fixes the problem I had. Unfortunately the code snippet above always shows the correct result.

joewiz:
@adamretter Using the Java admin client `bin/client.sh -l`, and running several times to warm up the db, then saving 5 runs and averaging the results, I am now seeing a ~13% improvement, comparing this PR to 5.2.0 with my original test code:

|version|avg. execution time|
|--|--|
| BaseX 9.3.1 | .6s |
| eXist-db 5.2.0 | 3.9s |
| eXist-db This PR (Bifurican 0.2.0-alpha1) | 3.4s |

joewiz:
Restarted the 1 failing CI test from AppVeyor.

adamretter:
@joewiz as requested, XQTS results:

* eXist-db 5.2.0: `74.48%`,
* eXist-db This PR (Bifurican 0.2.0-alpha1): `76.56%`

Whilst this is a small improvement, they are actually lower than I expected. Unfortunately eXist-db has several problems around comparing map keys, and therefore exhibits problems in `map:get`, `map:contains`, and `map:remove`. Something that needs further improvement in future I guess...

joewiz:
I've restarted the failing Travis and AppVeyor CI builds - one each.

wolfgangmm:
I just saw another exception popping up in the eXist log:

```
	at org.exist.xquery.FunctionCall.resetState(FunctionCall.java:348) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.UserDefinedFunction.resetState(UserDefinedFunction.java:229) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.InlineFunction.resetState(InlineFunction.java:100) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.functions.map.MapExpr$Mapping.resetState(MapExpr.java:123) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.functions.map.MapExpr$Mapping.access$0(MapExpr.java:121) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.functions.map.MapExpr.lambda$0(MapExpr.java:109) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at java.util.ArrayList.forEach(ArrayList.java:1257) ~[?:1.8.0_181]
	at org.exist.xquery.functions.map.MapExpr.resetState(MapExpr.java:109) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
```

wolfgangmm:
And here's a reproducible bug with `map:merge`:

```xquery
map:merge((
    map { "filter-rewrite": "true" },
    map { "genre": "letter" },
    map { "fields": "date" }
))
```

If you comment out any of the 3 sub-maps, you get the correct result. But you never get all 3 merged properly.

adamretter:
@wolfgangmm okay sure I will take a look. I am surprised we don't have test coverage already for something so simple!

adamretter:
@wolfgangmm Okay the`map:merge` issue was simple. I have fixed that now...

adamretter:
Latest XQTS results:

| XQTS Version    | eXist-db Version | Result | 
|-----------------|------------------|--------|
| W3C (pre) QT3.1 | 5.2.0            | 17.65% |
| W3C (pre) QT3.1 | This PR          | 19.61% |
| HEAD            | 5.2.0            | 74.48% |
| HEAD            | This PR          | 76.56% |

* Fro the `W3C` version, the following XQTS args were used:
    ```
    --xqts-version 3.1 --test-set map-collation,map-contains,map-entry,map-get,map-keys,map-new,map-remove,map-size
    ```
* For the `HEAD` version, the following XQTS args were used:
    ```
    --xqts-version HEAD --test-set map-contains,map-entry,map-find,map-for-each,map-get,map-keys,map-merge,map-put,map-remove,map-size
    ```

**NOTE**: The W3C (pre) QT3.1 results are not particularly useful as some of the test cases in there are out-of-date with regards to the final published release of XQuery 3.1.

adamretter:
@wolfgangmm Could you provide an updated stacktrace for your NPE and details of how to reproduce?

wolfgangmm:
@adamretter The stacktrace was thrown by an async process running in the background. I'm trying to reproduce it, but need the evening to inspect what exactly was going on.

Apart from this the PR appears to be stable for me now. The issues I reported are fixed. 

Also, @windauer was running the rather extensive stress tests based on the data and apps of @joewiz. This looked all good.

joewiz:
I've restarted the one failing AppVeyor CI run.

Any idea why Travis is suddenly missing from the CI listings here? I only see Codacy and Appveyor.

wolfgangmm:
@adamretter here's another stack trace with more plausible line numbers. Unfortunately this exception is thrown by a background thread and I can neither reproduce it at will nor can I relate the stack trace to a particular query.  It's just too random, happening only every now and then.

I will have a fresh look at the Java code tomorrow morning.

```
        at org.exist.xquery.AnalyzeContextInfo.<init>(AnalyzeContextInfo.java:70) ~[exist-core-5.3.0-SNAPSH
OT.jar:5.3.0-SNAPSHOT]
        at org.exist.xquery.InlineFunction.eval(InlineFunction.java:82) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3
.0-SNAPSHOT]
        at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
        at org.exist.xquery.functions.map.MapExpr.eval(MapExpr.java:61) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
        at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
        at org.exist.xquery.PathExpr.eval(PathExpr.java:280) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
        at org.exist.xquery.functions.array.ArrayConstructor.eval(ArrayConstructor.java:49) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
        at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
        at org.exist.xquery.PathExpr.eval(PathExpr.java:280) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
        at org.exist.xquery.VariableDeclaration.eval(VariableDeclaration.java:131) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
```

adamretter:
@wolfgangmm from your latest stack trace, I can't see how an NPE is possible:

`AnalyzeContextInfo.java` lines 69 and 70:
```java
public AnalyzeContextInfo(AnalyzeContextInfo other) {
        this.parent = other.parent;
```

`InlineFunction.java` line 82:
```java
call.analyze(new AnalyzeContextInfo(cachedContextInfo));
```

The only way there would be an NPE is if `other` is null at line 70 of `AnalyzeContextInfo.java` but the caller `InlineFunction` line 82, shows that that is never the case!

Let me know what you find...

wolfgangmm:
I traced it down to a variable declaration containing a map expression containing an inline function. In this particular case `analyze` is not passed down the hierarchy. Looks like some change introduced this recently, but maybe not even in this PR. Investigating further ...

wolfgangmm:
The issue was caused by #3265 which pollutes the context, so `analyze` does not correctly process the hierarchy and we get an NPE in the end.

This is hard to find but easy to fix. I'll open a separate PR.

PR #1523, opened 2017-08-17
Updated the EXPath HTTP Client

adamretter:
Fixes a number of issues and adds a couple new features:

1. [bugfix] Ensure chunked transfer encoding for HTTP 1.1

2. [bugfix] Avoid "null" in some error messages, as the message from Apache HTTP Client may be on the exception cause

3. [bugfix] It is an error if the user manually sets the Content-Type or Transfer-Encoding headers as these will be determined by Apache HTTP Client

4. [refactor] Switched away from using deprecated Apache HTTP Components methods

5. [feature] Implemented connection pooling for improved performance.

6. [feature] Support for GZip'ing data during transfer, just add a `gzip="true"` attribute to the `http:request` element.

7. Added support for SNI (Server Name Identification).

The connection pooling may also fix the issue that some users have had with needing to manually set a `Connection: close` header in the request.


grantmacken:
Thanks Adam for working on this. 
 My experience of using the expath  httpclient has not been good due to it's apparent lack of SNI support. 
https://github.com/fgeorges/expath-http-client-java/issues/5
My suggestion, was to add some tests for some known SNI sites.
Note the first site is no longer up.
Other tests and examples could round out the test suite.
 - HTTP2 requests, 
- Authenticating using Bearer Tokens
- etc

dizzzz:
@grantmacken HTTP/2 is supported from [http-components v5 ](https://hc.apache.org/httpcomponents-core-5.0.x/index.html)  whis not a stable release yet.

adamretter:
@dizzzz I just updated the `.classpath` and `nbproject/project.properties` files.

adamretter:
@grantmacken I replied to your SNI concern in fgeorges/expath-http-client-java#5

joewiz:
It's a little odd that the TravisCI failures are in the XQuery Update tests!

    [junit] Running org.exist.xquery.update.IndexIntegrationTest
    [junit] Tests run: 3, Failures: 1, Errors: 4, Skipped: 0, Time elapsed: 7.112 sec

adamretter:
@joewiz Indeed. I just restarted Travis to make sure it wasn't an anomaly

joewiz:
@adamretter That did the trick! How strange that TravisCI does this occasionally.

adamretter:
More likely I think we have tests that intermittently fail due to a real
issue hiding somewhere

joewiz:
It sounds like that would be tricky to root out... intermittent and inconsistent failures are the worst.

PR #1535, opened 2017-08-30
put db settings in build.properties

wshager:
I put the database settings for `conf.xml` (in some locations) in the `build.properties file`. This is probably a good idea, and would've saved me the time to find out how this works.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -7,8 +7,8 @@ project.name = eXist-db
 project.version = 3.5.0-SNAPSHOT
 
 # db settings
-dataDir = webapp/WEB-INF/data
-cacheSize = 256
+config.dataDir = webapp/WEB-INF/data
+config.cacheSize = 256
 
 # build settings
 build.debug = on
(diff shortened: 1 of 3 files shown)


dizzzz:
i'd like to propose to call these `config.dataDir` and `config.cacheSize`;

wshager:
@dizzzz sounds good.

wshager:
@dizzzz like dizzz?

adamretter:
@wshager Looks good to me, but if you add `dataDir` you should also add `journalDir` as by default they both point to `webapp/WEB-INF/data`. See `//recovery/@journal-dir` in `conf.xml.tmpl`.

Likewise, with `cacheSize` you will probably also need collection cache size, as the later is allocated from the former. See `//db-connection/@collectionCache`.

adamretter:
@wshager Would you be able to make the requested changes before Sept 25th Sept (which is our next release date)?

wshager:
@adamretter I simply corrected the parameters that were already present, but you're asking me to add new ones. However, I'm not entirely sure on the nomenclature, so perhaps this is better left to someone who is.

adamretter:
Superseded by https://github.com/eXist-db/exist/pull/1535

PR #1537, opened 2017-08-30
Correct classpath an typos to compile with Eclipse.

RemiKoutcherawy:
To compile with Eclipse

1. Correct .classpath :
replace 'lib/core/cglib-nodep-**2.2.2**.jar' with "lib/core/cglib-nodep-**3.1**.jar"/>
replace 'lib/core/commons-io-**2.4**.jar' with "lib/core/commons-io-**2.5**.jar"/> ...
replace kind="**src**" path="lib/core/j8fu-**1.0**.jar" with kind="**lib**" path="lib/core/j8fu-1.4.1.jar"..
Then Eclipse complains about typos :

2. Logger cannot be resolved to a type /eXist/src/org/exist/util/HtmlToXmlParser.java
   Replace `import org.apache.log4j.LogManager;` by `import org.apache.**logging**.log4j.LogManager;`

3. Logger cannot be resolved to a type /exist/launcher/ServiceManager.java
   Replace `import org.apache.log4j.LogManager;` by `import org.apache.**logging**.log4j.LogManager;`

4. The type NodeTest is ambiguous /eXist/src/org/exist/xquery/value/ValueSequence.java    line 563
   Add `import org.exist.xquery.NodeTest; // Needed by Eclipse` to disambiguate

5. Syntax error on token ";", delete this token /eXist/src/org/exist/xquery/functions/xmldb/XMLDBStore.java
   Replace ;; by ;
Now Eclipse reports only warnings.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -55,7 +55,7 @@
 	<classpathentry kind="lib" path="lib/core/log4j-core-2.8.2.jar"/>
 	<classpathentry kind="lib" path="lib/core/log4j-slf4j-impl-2.8.2.jar"/>
 	<classpathentry kind="lib" path="lib/core/log4j-jul-2.8.2.jar"/>
-	<classpathentry kind="lib" path="lib/core/jargo-0.4.2-SNAPSHOT-7fa1d04.jar" sourcepath="lib/core/jargo-src.zip"/>
+	<classpathentry kind="lib" path="lib/core/jargo-0.4.2-SNAPSHOT-7fa1d04.jar"/>
 	<classpathentry kind="lib" path="lib/core/xmldb.jar"/>
 	<classpathentry kind="lib" path="lib/optional/commons-httpclient-3.1.jar"/>
 	<classpathentry kind="lib" path="lib/core/commons-io-2.5.jar"/>
--- a/.classpath
+++ b/.classpath
@@ -11,11 +11,8 @@
 	<classpathentry kind="src" path="extensions/indexes/ngram/src"/>
 	<classpathentry kind="src" path="extensions/indexes/ngram/test/src"/>
 	<classpathentry excluding="org/exist/xquery/modules/cqlparser/**|org/exist/xquery/modules/jfreechart/**|org/exist/xquery/modules/memcached/**|org/exist/xquery/modules/oracle/**|org/exist/xquery/modules/xmlcalabash/**|org/exist/xquery/modules/xmpp/**|org/exist/xquery/modules/xslfo/**" kind="src" path="extensions/modules/src"/>
-	<classpathentry kind="src" path="extensions/xslt/src"/>
-	<classpathentry kind="src" path="extensions/xslt/test/src"/>
 	<classpathentry kind="src" path="src"/>
 	<classpathentry kind="src" path="test/src"/>
-	<classpathentry kind="src" path="tools/wrapper/src"/>
 	<classpathentry kind="src" path="samples/src"/>
 	<classpathentry kind="src" path="extensions/xqdoc/src"/>
 	<classpathentry kind="src" path="extensions/xqdoc/test/src"/>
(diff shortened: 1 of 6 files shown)


.classpath
```diff
@@ -52,34 +50,31 @@
 	<classpathentry kind="lib" path="extensions/exquery/restxq/lib/exquery-restxq-1.0-SNAPSHOT.jar"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
 	<classpathentry kind="lib" path="lib/core/antlr-2.7.7.jar"/>
-	<classpathentry kind="lib" path="lib/core/excalibur-cli-1.0.jar"/>
 	<classpathentry kind="lib" path="lib/core/jline-0.9.94.jar"/>
 	<classpathentry kind="lib" path="lib/core/log4j-api-2.8.2.jar"/>
 	<classpathentry kind="lib" path="lib/core/log4j-core-2.8.2.jar"/>
 	<classpathentry kind="lib" path="lib/core/log4j-slf4j-impl-2.8.2.jar"/>
 	<classpathentry kind="lib" path="lib/core/log4j-jul-2.8.2.jar"/>
+	<classpathentry kind="lib" path="lib/core/jargo-0.4.2-SNAPSHOT-7fa1d04.jar" sourcepath="lib/core/jargo-src.zip"/>
```
dizzzz: for me, lib/core/jargo-src.zip does not exist?

src/org/exist/xquery/value/ValueSequence.java
```diff
@@ -29,6 +29,7 @@
 import org.exist.numbering.NodeId;
 import org.exist.util.FastQSort;
 import org.exist.xquery.*;
+import org.exist.xquery.NodeTest; // Needed by Eclipse to disambiguate
```
dizzzz: Any idea what the conflict is? maybe a unit directory is not marked as a test directory? For me, this is kinda suspicious.

src/org/exist/launcher/ServiceManager.java
```diff
@@ -42,7 +43,7 @@
 
 public class ServiceManager {
 
-    private final static Logger LOG = Logger.getLogger(ServiceManager.class);
+    private final static Logger LOG = LogManager.getLogger(ServiceManager.class);
```
dizzzz: Though this change is good, I think that in .classpath the file is ` lib/user/log4j-1.2-api-2.8.2.jar` is missing. Maybe more?

dizzzz:
Highly appreciated!

.classpath
RemiKoutcherawy: Yes, remove it. 
I used it to debug then removed it, and forgot to correct classpath.

src/org/exist/xquery/value/ValueSequence.java
RemiKoutcherawy: Eclipse error : 
The type NodeTest is ambiguous	/eXist/src/org/exist/xquery/value	ValueSequence.java	line 564	Java Problem

Eclipse suggest to choose between many :
import com.sun.org.apache.xalan.internal.xsltc.compiler.NodeTest;
import com.sun.org.apache.xpath.internal.patterns.NodeTest;
import net.sf.saxon.pattern.NodeTest;
import org.apache.xalan.xsltc.compiler.NodeTest;
import org.apache.xpath.patterns.NodeTest;
...
import org.exist.dom.persistent.NodeTest;
import org.exist.fluent.NodeTest;
import org.exist.xquery.NodeTest;

Only the last one fits.

dizzzz:
I'd like to discuss the PR with the team; my ideas
- we should not change code because an IDE seems to have an issue. Instead we should fix the IDE configuration files
 - the NodeTest import.... when building with 'plain' java the current code base builds fine (but the changes are ok)
 - the actual problem  is that an JAR import is missing. -- for consistency this must be fixed 
- PR should be clean, roughly addressing one issue ; the last PR has a change not related to the Eclipse changes, and should therefore be in an other PR to be able to judge. In depth: I don't agree with the change as the use-case is unclear, and the original use case rather strong.

build/scripts/installer.xml
```diff
@@ -103,7 +103,7 @@
         </fetch>
     </target>
 
-    <target name="xars" depends="download-xars" description="Scan apps directory and include all .xar files into installer">
+    <target name="xars" description="Scan apps directory and include all .xar files into installer">
```
adamretter: I think this is a good idea in general, however I think there might be a problem...

Unfortunately I think this means the Xars will not be correctly downloaded and included when building the installer from a clean build.

May I instead suggest, that you keep the `depends="download-xars"` dependency and instead add a check to the `download-xars` target to skip doing anything if the Xars are already downloaded.

src/org/exist/launcher/ServiceManager.java
adamretter: I would ideally suggest not adding the legacy Jar to the `.classpath`. It is not needed for building and may only confuse people by suggesting to import legacy packages when doing auto-complete etc.

src/org/exist/xquery/value/ValueSequence.java
adamretter: This sounds like an iffy problem in the Eclipse Java compiler. If it helps Eclipse users, then adding the import seems fine to me.

adamretter:
@dizzzz Apart from the issue with the `download-xars` target this looks fine to me.

@RemiKoutcherawy Would you be willing to fix the `download-xars` issue that I commented on?

dizzzz:
thank you @RemiKoutcherawy for your contribution! highly appreciated.....

PR #1539, opened 2017-09-05
[bugfix] Undo change to autogenerated code caused by 967f48a

adamretter:
The commit 967f48a directly modified some generated code, however you have to modify the `.g` file which generates the code and not the code itself.


dizzzz:
I understand you change, but.... during the build these files are not regenerated, are they?

adamretter:
@dizzzz yes this one is

PR #1541, opened 2017-09-07
[bugfix] Fix selection of processing instructions by target

wolfgangmm:
This is a regression caused by changes to the DOM implementation. Test added.


adamretter:
@wolfgangmm Thanks Wolfgang :-)

PR #1542, opened 2017-09-09
Previously Mixed-Content was not correctly serialized to JSON

adamretter:
The tests added in `test/src/org/exist/util/serializer/json/JSONWriterTest.java` show that `JSONWriterTest#serializesMixedContent_whenAttrsPresent()` would previously correctly serialize mixed content, but if you removed the attributes from the elemnents, as in `JSONWriterTest#serializesMixedContent`, then the text nodes would be incorrectly skipped. This bugfix fixes that issue.


dizzzz:
@wolfgangmm ?

adamretter:
@dizzzz all tests pass :-)

PR #1546, opened 2017-09-12
handle multiple x-forwarded-for IP addresses

wshager:
It seems there was an additional requirement, as the X-Forwarded-For header can have multiple IP addresses, separated by commas. This happens when the client is in a chain of proxies. I've added a split on comma, and take the last of the IPs in the list (following https://en.wikipedia.org/wiki/X-Forwarded-For).


extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
```diff
@@ -63,12 +63,15 @@ protected void doPost(final HttpServletRequest request, final HttpServletRespons
 
 
         // Get reverse proxy header when available, otherwise use regular IP address
-        String ipAddress = request.getHeader("X-Forwarded-For");
-        if (ipAddress == null) {
+        String ipAddress = request.getHeader("X-Forwarded-For").replaceAll("\\s","");
```
adamretter: I am afraid you run the risk of a NPE here as `ServletRequest#getHeader(String)` can return `null` if the header is not present - http://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html#getHeader(java.lang.String)

Perhaps instead consider something like:

```java
final String ipAddress =
    Optional.ofNullable(request.getHeader("X-Forwarded-For"))
    .map(str -> str.replaceAll("\\s", ""))
    .filter(str -> !str.isEmpty)
    .map(str -> str.split(","))
    .filter(strs -> strs.length > 1)
    .map(strs -> strs[strs.length - 1]
    .orElseGet(() -> request.getRemoteAddr())
```

wshager:
@adamretter already found that out, yes. Does this functional approach have the same performance though? I wouldn't mind imperative null checks in this place.

adamretter:
I haven't tested the performance, I just provided it as an example

extensions/security/iprange/src/org/exist/security/realm/iprange/IPRangeServlet.java
```diff
@@ -63,12 +63,15 @@ protected void doPost(final HttpServletRequest request, final HttpServletRespons
...
+        // there may be a comma-separated chain of proxies
+        if (!ipAddress.isEmpty()) {
+            String[] xFFs = ipAddress.split(",");
+            if(xFFs.length > 1) ipAddress = xFFs[xFFs.length - 1];
```
dizzzz: please could you describe which address you want to fetch? 1st? last?

wshager: The last IP address should be used, because the values are only *added* when chained proxies are used. If no additional proxies are used, the XFF header only contains the remote address, which is the default.

wshager: > In a chain of proxy servers where this is fully utilized, the first "for" parameter will disclose the client where the request was first made, followed by any subsequent proxy identifiers.  The last proxy in the chain is not part of the list of "for" parameters.  The last proxy's IP address, and optionally a port number, are, however, readily available as the remote IP address at the transport layer.  It can, however, be more relevant to read information about the last proxy from preceding "Forwarded" header field's "by" parameter, if present.

https://tools.ietf.org/html/rfc7239#page-6

wshager:
@adamretter I've added the null check

adamretter:
@wshager thanks for another contribution :-)

PR #1551, opened 2017-09-20
[bugfix] Range index sets incorrect type

### LINKED ISSUE #1454: Combined range index causes query to fail
### What is the problem

If a collection has an element-attribute combined index, certain type checks on function signature return values will erroneously fail. If the index is removed, the type check is performed correctly.

### What did you expect

I expected the index configuration to have no impact on function signature type checking. 

Here is a test that demonstrates the error in eXide. It performs the same query on the same data in two collections - one collection has the range index applied, and one has no index applied. The query fails on the indexed collection. Not shown in the test is a 3rd case: if you change the `<create>` element to a simple value like `<create qname="@type" type="xs:string"/>`, the test will pass. So it isn't all range indexes that cause the problem - it's the "combined" type, with `create/field`. 

```xquery
xquery version "3.1";

module namespace rt="combined-range-function-signature-test";

declare variable $rt:COLLECTION_CONFIG := 
            <range>
                <create match="//note">
                    <field name="note-type" match="@type" type="xs:string"/>
                </create>
            </range>
        </index>
    </collection>;

declare variable $rt:DATA :=
    <root>
        <div>
            <head>Telegram 294530 From the Department of State to the Consulate in Montreal<note
                    type="source">Source: National Archives, RG 59, Central Files 1967-69, POL 27
                    ARAB-ISR. Unclassified. Drafted by Loy (E/TT), cleared by Roger P. Davies (NEA) and
                    Styles (E/OA/AVP), and approved by Rusk.</note>
                <note type="summary">A message from Deputy Assistant Secretary of State Loy to IATA
                    Director-General Knut Hammarskjold expressing U.S. concern about recent hijackings.
                    Loy condemned Israel for the attack on Khaldeh Airport in Lebanon and noted that the
                    U.S. had registered disapproval of the Athens hijacking of an El Al aircraft.</note>
            </head>
        </div>
    </root>;

declare variable $rt:INDEXED_COLLECTION_NAME := "test-indexed";
declare variable $rt:INDEXED_COLLECTION := "/db/" || $rt:INDEXED_COLLECTION_NAME;

declare variable $rt:NON_INDEXED_COLLECTION_NAME := "test-non-indexed";
declare variable $rt:NON_INDEXED_COLLECTION := "/db/" || $rt:NON_INDEXED_COLLECTION_NAME;

declare
    %test:setUp
function rt:setup() {
    xmldb:create-collection("/db/system/config/db", $rt:INDEXED_COLLECTION_NAME),
    xmldb:store("/db/system/config/db/" || $rt:INDEXED_COLLECTION_NAME, "collection.xconf", $rt:COLLECTION_CONFIG),
    xmldb:create-collection("/db", $rt:INDEXED_COLLECTION_NAME),
    xmldb:store($rt:INDEXED_COLLECTION_NAME, "test.xml", $rt:DATA),
    xmldb:create-collection("/db", $rt:NON_INDEXED_COLLECTION_NAME),
    xmldb:store($rt:NON_INDEXED_COLLECTION_NAME, "test.xml", $rt:DATA)
};

declare
    %test:tearDown
function rt:cleanup() {
    xmldb:remove($rt:INDEXED_COLLECTION_NAME),
    xmldb:remove("/db/system/config/db/" || $rt:INDEXED_COLLECTION_NAME),
    xmldb:remove($rt:NON_INDEXED_COLLECTION_NAME),
    xmldb:remove("/db/system/config/db/" || $rt:NON_INDEXED_COLLECTION_NAME)
};

declare function rt:get-note($div as element(div)) as element(note)  {
    $div//note[@type='summary']
};

(:~
 : Check indexed collection. This query should return one element. 
 :)
declare
    %test:assertXPath('/@type = "summary"')
function rt:test-indexed-collection() {
    rt:get-note(collection($rt:INDEXED_COLLECTION_NAME)//div)
};

(:~
 : Check non-indexed collection. This query should return one element. 
 :)
declare
    %test:assertXPath('/@type = "summary"')
function rt:test-non-indexed-collection() {
    rt:get-note(collection($rt:NON_INDEXED_COLLECTION_NAME)//div)
};
```

The test suite returns the following results:

```xml
<testsuites>
    <testsuite package="range-tests" timestamp="2017-06-09T16:17:01.251-04:00" failures="0"
        pending="0" tests="2" time="PT0.011S">
        <testcase name="test-indexed-collection" class="rt:test-indexed-collection">
            <error type="err:XPTY0004"
                message="It is a type error if, during the static analysis phase, an expression is found to have a static type that is not appropriate for the context in which the expression occurs, or during the dynamic evaluation phase, the dynamic type of a value does not match a required type as specified by the matching rules in 2.5.4 SequenceType Matching. err:XPTY0004: return type of function 'rt:get-note'. exerr:ERROR Type error: expected type: element(); got: node()"
            />
        </testcase>
        <testcase name="test-non-indexed-collection" class="rt:test-non-indexed-collection"/>
    </testsuite>
</testsuites>
```

### Context information

- eXist-db version: 3.3.0-DEVELOP a059dec22
- Java version: 8u121
- Operating system: macOS 10.12.5
- 64 bit

wolfgangmm:
New range index does not assign correct node type to results of a field query, thus leading to type errors. Closes #1454

### DIFF
--- a/extensions/indexes/range/src/org/exist/indexing/range/RangeIndexWorker.java
+++ b/extensions/indexes/range/src/org/exist/indexing/range/RangeIndexWorker.java
@@ -527,14 +527,16 @@ public class RangeIndexWorker implements OrderedValuesIndex, QNamedKeysIndex {
                 } else {
                     query = toQuery(field, qname, keys[0], operator, docs);
                 }
+                final short nodeType = qname.getNameType() == ElementValue.ATTRIBUTE ? Node.ATTRIBUTE_NODE : Node
+                        .ELEMENT_NODE;
 
                 if (contextSet != null && contextSet.hasOne() && contextSet.getItemType() != Type.DOCUMENT) {
                     NodesFilter filter = new NodesFilter(contextSet);
                     filter.init(searcher.getIndexReader());
                     FilteredQuery filtered = new FilteredQuery(query, filter, FilteredQuery.LEAP_FROG_FILTER_FIRST_STRATEGY);
-                    resultSet.addAll(doQuery(contextId, docs, contextSet, axis, searcher, null, filtered, null));
+                    resultSet.addAll(doQuery(contextId, docs, contextSet, axis, searcher, nodeType, filtered, null));
                 } else {
-                    resultSet.addAll(doQuery(contextId, docs, contextSet, axis, searcher, null, query, null));
+                    resultSet.addAll(doQuery(contextId, docs, contextSet, axis, searcher, nodeType, query, null));
                 }
             }
             return resultSet;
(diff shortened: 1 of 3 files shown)


joewiz:
@wolfgangmm Thank you!

PR #1552, opened 2017-09-21
RESTXQ handle multipart/related request and response

maanAlbalkhi:
### Description:
- Accept multipart/related http requests and convert to multi http:body and http:header nodes in a http:multipart element like:
`<http:multipart>
    <http:header name="content-id" value="..." />
    <http:body media-type="...">...</http:body>
    .
    . 
</http:multipart>`
Body content is represented as BASE64

- Serialize http:multipart element as response to a multipart response with a body part for each http:body element.

Multipart/related content type is mostly used for MTOM/XOP requests and responses
### Reference:
[ MIME Multipart/Related RFC 2387](https://tools.ietf.org/html/rfc2387#section-6.1)
[  The Multipart Content-Type RFC 1341 (MIME)](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html)
[ HTTP Client Module](http://expath.org/spec/http-client)

### DIFF
--- a/extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
+++ b/extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
@@ -173,7 +173,7 @@ class RestXqServiceImpl extends AbstractRestXqService {
                     }
                     MimeType mimeType = MimeTable.getInstance().getContentType(contentType);
 
-                    if(contentType.toLowerCase().equals("multipart/related")){
+                    if(contentType.equalsIgnoreCase("multipart/related")){
                     	
                     	// multipart/related request
                     	String encoding = request.getCharacterEncoding();
@@ -472,7 +472,7 @@ class RestXqServiceImpl extends AbstractRestXqService {
      * 
      * @see <a href="http://expath.org/spec/http-client">HTTP Client Module</a>
      */
-    private String buildBodyPartAsXmlStr(final Pattern typePtn, final Pattern transEncPtn, final String headers, String content) throws RestXqServiceException{
+    private String buildBodyPartAsXmlStr(final Pattern typePtn, final Pattern transEncPtn, final String headers, final String content) throws RestXqServiceException{
     	// find content type
     	final Matcher typeMtc = typePtn.matcher(headers);
 		String contentType = null;
@@ -492,11 +492,11 @@ class RestXqServiceImpl extends AbstractRestXqService {
 		}
 		// if the transfer encoding is BASE64 then return the the content in a http:body node
 		// and declare the media-type as binary
-		if(transferEncodeing != null && transferEncodeing.toLowerCase().equals("base64")){
+		if(transferEncodeing != null && transferEncodeing.equalsIgnoreCase("base64")){
 			return String.format("<http:body media-type=\"%s\">%s</http:body>", contentType, content);
 		}else{
-			content = Base64.getEncoder().encodeToString(content.getBytes());
-			return String.format("<http:body media-type=\"%s\">%s</http:body>", contentType, content);
+			final String encodedContent = Base64.getEncoder().encodeToString(content.getBytes());
+			return String.format("<http:body media-type=\"%s\">%s</http:body>", contentType, encodedContent);
 		}
     }
     
--- a/extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
+++ b/extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
@@ -26,17 +26,10 @@
  */
 package org.exist.extensions.exquery.restxq.impl;
 
-import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.util.ArrayList;
-import java.util.Base64;
 import java.util.List;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 import org.apache.commons.io.input.CloseShieldInputStream;
 import org.apache.commons.io.output.ByteArrayOutputStream;
--- a/extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
+++ b/extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
@@ -26,27 +26,29 @@
  */
 package org.exist.extensions.exquery.restxq.impl;
 
-import java.io.IOException;
-import java.io.InputStream;
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.ArrayList;
-import java.util.List;
 import java.util.Base64;
+import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+
 import org.apache.commons.io.input.CloseShieldInputStream;
+import org.apache.commons.io.output.ByteArrayOutputStream;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.exist.extensions.exquery.xdm.type.impl.BinaryTypedValue;
-import org.exist.extensions.exquery.xdm.type.impl.DocumentTypedValue;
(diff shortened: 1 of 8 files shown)


extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
```diff
@@ -164,9 +171,24 @@ protected Sequence extractRequestBody(final HttpRequest request) throws RestXqSe
                     if (contentType.contains(";")) {
                         contentType = contentType.substring(0, contentType.indexOf(";"));
                     }
-
                     MimeType mimeType = MimeTable.getInstance().getContentType(contentType);
-                    if (mimeType != null && !mimeType.isXMLType()) {
+
+                    if(contentType.equalsIgnoreCase("multipart/related")){
```
dizzzz: “Related” ????? Never seen that before?

dizzzz: NPE check please

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
```diff
@@ -297,4 +319,185 @@ private static StringValue parseAsString(final InputStream is, final String enco
...
+  	 *		<http:body media-type="...">...</http:body>
+	 *		.
+	 *		. 
+	 * 	</http:multipart>
+	 * 
+     */
+    private String extractBodyParts(final HttpRequest request, final InputStream is, final String encoding) throws RestXqServiceException, IOException{
+    	
+    	final String requestContentType = request.getContentType();
+    	final String boundary = getMultiPartBoundary(requestContentType);
```
dizzzz: I remember that the spec allows multiple boundary definitions in one request.
It is ok for me not to support this, but we must be aware.....

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
```diff
@@ -297,4 +319,185 @@ private static StringValue parseAsString(final InputStream is, final String enco
...
+    	final String rootId = getRootPartId(requestContentType);
+		final String content = getRequestContent(is, encoding);
+		
+		// build http:header and http:body for each body part within a http:mulitpart as XML String
+		// @see <a href="http://expath.org/spec/http-client">HTTP Client Module</a>
+		final StringBuilder xmlPartsAsStr = new StringBuilder();
+		
+		// compile part Content-Type pattern before going through the parts
```
dizzzz: Please move these out of the methods into class level (performance reasons)

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceSerializerImpl.java
```diff
@@ -82,10 +87,55 @@ protected void serializeBinaryBody(final Sequence result, final HttpResponse res
             }
         }
     }
-
+    
     @Override
     protected void serializeNodeBody(final Sequence result, final HttpResponse response, final Map<SerializationProperty, String> serializationProperties) throws RestXqServiceException {
-        try(final DBBroker broker = getBrokerPool().getBroker();
+    	// multipart/related serialization 
+    	if(serializationProperties.get(SerializationProperty.MEDIA_TYPE).equals("multipart/related")){
```
dizzzz: Risk for NPE, please swap

dizzzz:
Great to see multi-part support! It is also for receiving data no?

It would be nice to have testcases added, as there is quite some text/data fiddling and manipulating in the code.

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
maanAlbalkhi: Thank you for the quick response.
The MIME multipart/related is needed for SOAP-MTOM/XOP messages
[SOAP Message Transmission Optimization Mechanism](https://www.w3.org/TR/soap12-mtom/#xop-serialization)

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
maanAlbalkhi: That's true, in case of nested multipart entities "a Content-Type of multipart in a body part within another multipart entity": [Multipart RFC 1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html)
In this case further implementation is needed and a TODO task at the moment

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
maanAlbalkhi: That's plausible and i will work on it

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceSerializerImpl.java
maanAlbalkhi: will be done

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
maanAlbalkhi: it will be done

maanAlbalkhi:
@dizzzz  yes it's also for receiving data and not only soap messages.
I will implement and add testcases.

adamretter:
@maanAlbalkhi Thank you this is a valuable contribution however I feel that the approach is not quite so straight-forward. RESTXQ 1.0 does not support multi-part, so first we would need to define a RESTXQ 2.0 which does. Also, on a technical note, I don't think we should build XML as strings.

Would you be willing to work further with me on this?

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceImpl.java
maanAlbalkhi: contentType is checked for null value  [here](https://github.com/eXist-db/exist/pull/1552/files#diff-c29b4b471e6ee7611c4de8e42c6df4c6R169).
I hope we mean the same.

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/RestXqServiceSerializerImpl.java
maanAlbalkhi: This [commit](https://github.com/eXist-db/exist/pull/1552/commits/8b6d15bb2c0326fdd9da70eeb6caaf6905f77415) should solve the problem

maanAlbalkhi:
@dizzzz
Test cases are now part of the pull request. Thanks for pointing that out.
@adamretter
I totally agree. Having multipart/*-support in the RESTXQ specification would leed to a more intuative programming interface. I welcome your suggestion to work out a solution for multipart/* with you as I see demand for it.
I read the discussion on this topic you head with Josef (http://exist.2174344.n4.nabble.com/RESTXQ-Multipart-parsing-td4663106.html) and I wonder what your opinion is today on full support versus partial support of multipart/*. You pointed out, that HTTP parts could be of type multipart/* leading to a recursive structure. Of course, I don't have proof or any figures here, however, I do think that this a rare corner case. That is, I would vote for a short solution that supports the common "single-layered" multipart/* use cases like SOAP/XOP.

maanAlbalkhi:
@dizzzz, @adamretter  i agree that building XML as strings is not a good way. That's why i moved the multipart implementation to class level and multipart elements are now built with MemTreeBuilder instead of strings.

adamretter:
In case anyone is wondering... This is awaiting a formal definition of RESTXQ 2.0

adamretter:
Why is this closed?

line-o:
It was blocked for two years. Feel free to reopen when it is unblocked.

joewiz:
Seems a little rash to me to close issues without asking again for a status update.

line-o:
I remembered this to be wontfix. But that could have been a different PR. Sorry for the mixup.
Can we remove the milestone until it is unblocked?

line-o:
What about the high prio label on this? I would like to remove it since after two years of no action this is rather confusing.

line-o:
I removed the high-prio label after two years of inaction.

dizzzz:
agreed in community call to close.

Issue #1555, opened 2017-09-22
Bug with on-disk node test and following-sibling axis


joewiz:
### What is the problem

A query which returns expected results with in-memory node returns incorrect results when stored to the database. The problem is manifest when performing a node test using the following-sibling axis against the on-disk data.

My test query creates an in-memory node, stores it on-disk, then performs a series of queries against the data in-memory and on-disk. The queries select a node, traverse the *-sibling XPath axes, and apply node tests. The results become inconsistent when the query performs a node test with a named namespace on the following-sibling axis.

### What did you expect

I expected queries on in-memory and on-disk data to perform identically, and for queries involving namespaces and the following-sibling axis to return expected results. 

### Describe how to reproduce or add a test

My test query is as follows:

```xquery
xquery version "3.1";

declare namespace tei="http://www.tei-c.org/ns/1.0";

let $in-mem := 
    <TEI xmlns="http://www.tei-c.org/ns/1.0">
        <div>
            <div xml:id="d1"/>
            <div xml:id="d2"/>
            <div xml:id="d3"/>
        </div>
    </TEI>
let $store := xmldb:store("/db", "test.xml", $in-mem)
let $stored := doc($store)
return
    (
        exists($in-mem//tei:div[@xml:id eq "d2"]),
        exists($stored//tei:div[@xml:id eq "d2"]),
        exists($in-mem//tei:div[@xml:id eq "d2"]/preceding-sibling::*:div),
        exists($stored//tei:div[@xml:id eq "d2"]/preceding-sibling::*:div),
        exists($in-mem//tei:div[@xml:id eq "d2"]/preceding-sibling::tei:div),
        exists($stored//tei:div[@xml:id eq "d2"]/preceding-sibling::tei:div),
        exists($in-mem//tei:div[@xml:id eq "d2"]/following-sibling::*:div),
        exists($stored//tei:div[@xml:id eq "d2"]/following-sibling::*:div),
        exists($in-mem//tei:div[@xml:id eq "d2"]/following-sibling::tei:div),
        exists($stored//tei:div[@xml:id eq "d2"]/following-sibling::tei:div)
    )
```

The queries all return `true()` except the final one, which returns `false()`.

**Update:** My original test query used retrieved the source document via HTTP request, but in the interests of simplifying the test, I trimmed the test document to the bare minimum needed to reproduce the result.

### Context information

[...]

PR #1556, opened 2017-09-25
Add test for #1555 following-sibling bug

joewiz:
### Description:

Adds a minimal XQSuite test illustrating #1555.

Note: the test is annotated as `%test:pending`, so merging it will not break the build. Please remove the `%test:pending` when the bug is fixed to prevent future regressions!

With the pending annotation removed the test returns this failure:

```xml
<testcase name="following-sibling-stored-parent-child-same-name"
    class="axes:following-sibling-stored-parent-child-same-name">
    <failure
        message="assertEquals failed: wrong number of items returned by function. Expected: 1. Got: 0"
        type="failure-error-code-1">&lt;a/&gt;</failure>
    <output/>
</testcase>
```

### Reference:

See #1555.

### Type of tests:

XQSuite

### DIFF
--- a/test/src/xquery/axes.xql
+++ b/test/src/xquery/axes.xql
@@ -44,6 +44,14 @@ declare variable $axes:in-mem-doc2 :=
         </root>
     };
 
+declare variable $axes:in-mem-doc3 :=
+    document {
+        <a>
+            <a/>
+            <a/>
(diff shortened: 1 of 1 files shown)


joewiz:
Superseded by #2113.

Issue #1298, opened 2017-02-26
Arrow Operator syntax edge-case error


adamretter:
The following test-case for the arrow operator fails (also fails on Saxon and BaseX (but I don't think it should)):

``` xquery
xquery version "3.1";
declare namespace map = "http://www.w3.org/2005/xpath-functions/map";

function($k, $v) {
    "$k=" || $k
} =>
    map:for-each(
        map {
            "a" : "1",
            "b" : "2",
            "c" : "3"
        },
        ?
    )()
```

Previously described here: https://github.com/eXist-db/exist/pull/828#issuecomment-155055413, we should seek clarification from the XQuery WG.

dizzzz:
if it fails in saxon..... interesting.

joewiz:
@adamretter I tested after the merge of #1557, and your test still returns the following error:

> Cannot compile xquery: exerr:ERROR org.exist.xquery.XPathException: err:XPST0003 unexpected token: ( [at line 14, column 6]

(Line 14 is the last line of your query above.)

Perhaps you're right that clarification from the WG would help, especially given the observation about the same failure in Saxon and BaseX?

line-o:
The spec clearly states the `U => f(A,B)` will evaluate to `f(U,A,B)`. https://www.w3.org/TR/xquery-31/#id-arrow-operator

Also @adamretter the below example does work:

```xquery
function($k, $v) { "$k=" || $k } 
    => (map:for-each(map { "a": "1", "b": "2", "c": "3" }, ?))()
```

line-o:
@adamretter I believe this issue can be closed.

adamretter:
@line-o following the spec, as you mentioned `U => f(A,B)` will evaluate to `f(U,A,B)`:

Let's break this down into steps, if I am not mistaken, then:

1. `function($k, $v) {"$k=" || $k}` reduces to: `function(item()*, item()*) as xs:string`

2. `map:for-each(map {"a" : "1"}, ?)` reduces to: `function(function(xs:anyAtomicType, item()*) as item()*)) as item()*`.

So I don't see why I can't feed (1) into (2) as the first argument, e.g.

```
function(item()*, item()*) as xs:string => function(function(xs:anyAtomicType, item()*) as item()*)) as item()*()
```

What did I miss?

line-o:
@adamretter I totally agree, that this MAY/SHOULD work. I haven't checked how the other xquery implementations handle this, yet.
[...]

Issue #1553, opened 2017-09-21
Arrow operator expression failure


joewiz:
### What is the problem

Another arrow operator expression is failing, whereas the non-arrow operator version works. For example, the following expression throws an error:

```xquery
xquery version "3.1";

(1, 2, 3) => string-join("-")
```

The error from eXist:

> err:XPTY0004 xs:integer(1) is not a sub-type of xs:string

When expressing the query without the arrow operator, eXist does not throw an error:

```xquery
xquery version "3.1";

string-join((1, 2, 3), "-")
```

I'm not sure if the underlying problem is some kind of type promotion failure when using the arrow operator? 

Also, on a possibly related point (?), in version 3.1 of XQuery F&O spec appears changed the function signature for the first parameter of `fn:string-join()`, relaxing it from `xs:string*` to  `xs:anyAtomicType*`. eXist still restricts this parameter to `xs:string*`. See https://www.w3.org/TR/xpath-functions-31/#func-string-join vs. https://www.w3.org/TR/xpath-functions-30/#func-string-join vs. https://www.w3.org/TR/xpath-functions/#func-string-join vs. https://github.com/eXist-db/exist/blob/develop/src/org/exist/xquery/functions/fn/FunStringJoin.java#L58-L60. 

### What did you expect

The expected result, as produced correctly by BaseX and Saxon (and by eXist when using the variant of the query without the arrow operator), is "1-2-3".

### Describe how to reproduce or add a test

See the sample code above.

### Context information

- eXist-db version + Git Revision hash: eXist-db 3.5.0-SNAPSHOT+0875a60cc
- Java version: jdk1.8.0_144
- Operating system: macOS 10.12.6
- 32 or 64 bit: 64 bit
- Any custom changes in e.g. conf.xml: none

PR #1557, opened 2017-09-25
Fix arrow operator type errors

### LINKED ISSUE #1548: Arrow operator in user-defined function causes NPE
### What is the problem

In the current develop branch, the use of an arrow operator inside user-defined functions is causing an NPE. Adapting the query to not use the arrow operator eliminates the NPE.

### What did you expect

I expected the arrow operator not to cause an NPE in this circumstance. Saxon and BaseX return the expected result, 6.

The following query triggers the NPE:

```xquery
xquery version "3.1";

declare function local:foo($x) {
    local:foo($x, "bar")
};

declare function local:foo($x, $y) {
    ($x || $y) => string-length()
};

local:foo("foo")
```

The NPE, as seen in exist.log:

```text
2017-09-16 12:07:33,356 [qtp1229776026-79] ERROR (XQueryServlet.java [process]:534) - null 
java.lang.NullPointerException: null
	at org.exist.xquery.Function.analyze(Function.java:348) ~[exist.jar:3.5.0-SNAPSHOT]
	at org.exist.xquery.FunctionCall.analyze(FunctionCall.java:142) ~[exist.jar:3.5.0-SNAPSHOT]
	at org.exist.xquery.UserDefinedFunction.analyze(UserDefinedFunction.java:115) ~[exist.jar:3.5.0-SNAPSHOT]
	at org.exist.xquery.FunctionCall.analyze(FunctionCall.java:148) ~[exist.jar:3.5.0-SNAPSHOT]
	at org.exist.xquery.PathExpr.analyze(PathExpr.java:184) ~[exist.jar:3.5.0-SNAPSHOT]
	at org.exist.xquery.XQueryContext.analyzeAndOptimizeIfModulesChanged(XQueryContext.java:1609) ~[exist.jar:3.5.0-SNAPSHOT]
	at org.exist.xquery.XQuery.compile(XQuery.java:137) ~[exist.jar:3.5.0-SNAPSHOT]
	at org.exist.xquery.XQuery.compile(XQuery.java:79) ~[exist.jar:3.5.0-SNAPSHOT]
	at org.exist.xquery.XQuery.compile(XQuery.java:71) ~[exist.jar:3.5.0-SNAPSHOT]
	at org.exist.http.servlets.XQueryServlet.process(XQueryServlet.java:425) [exist-optional.jar:3.5.0-SNAPSHOT]
...
```

The following query is identical to the first, but it avoids the use of the arrow operator. It does not trigger an NPE:

```xquery
xquery version "3.1";

declare function local:foo($x) {
    local:foo($x, "bar")
};

declare function local:foo($x, $y) {
    string-length($x || $y)
};

local:foo("foo")
```

### Context information

- eXist-db version + Git Revision hash: eXist 3.5.0-DEVELOP fdaabad17
- Java version: 1.8.0_144
- Operating system: macOS 10.12.6
- 64 bit

wolfgangmm:
Closes #1548 and #1553.

### DIFF
--- a/test/src/xquery/xquery3/higher-order.xml
+++ b/test/src/xquery/xquery3/higher-order.xml
@@ -380,7 +380,7 @@ return
         </code>
         <expected>70</expected>
     </test>
-    <test output="text">
+    <test output="text" id="lookup-unknown">
         <task>fn:function-lookup on unknown function</task>
         <code><![CDATA[xquery version "3.0";
 
@@ -388,7 +388,7 @@ declare namespace ex="http://exist-db.org/xquery/ex";
 
 let $f1 := function-lookup(xs:QName("ex:add"), 2)
 return
-    if ($f1) then
+    if (exists($f1)) then
        $f1(20, 50)
     else
         ()
--- a/src/org/exist/xquery/ArrowOperator.java
+++ b/src/org/exist/xquery/ArrowOperator.java
@@ -161,7 +161,7 @@ public class ArrowOperator extends AbstractExpression {
         }
     }
 
-    private class ContextParam extends AbstractExpression {
+    private class ContextParam extends Function.Placeholder {
 
         private Sequence sequence;
 
--- a/src/org/exist/xquery/Function.java
+++ b/src/org/exist/xquery/Function.java
@@ -457,7 +457,7 @@ public abstract class Function extends PathExpr {
         return this;
     }
 
-    public final static class Placeholder extends AbstractExpression {
+    public static class Placeholder extends AbstractExpression {
 
         public Placeholder(final XQueryContext context) {
             super(context);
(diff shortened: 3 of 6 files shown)


test/src/xquery/xquery3/higher-order.xml
```diff
@@ -380,15 +380,15 @@ return
...
+    <test output="text" id="lookup-unknown">
         <task>fn:function-lookup on unknown function</task>
         <code><![CDATA[xquery version "3.0";
 
 declare namespace ex="http://exist-db.org/xquery/ex";
 
 let $f1 := function-lookup(xs:QName("ex:add"), 2)
 return
-    if ($f1) then
+    if (exists($f1)) then
```
adamretter: Hmm. Just wondering why the exists call is needed?

wolfgangmm: Because effectiveBooleanValue is not defined on function references, so the error eXist did throw for this test was correct. I checked against the spec and other implementations.

joewiz:
I can confirm this PR fixes the issues I reported in #1548 and #1553.

Many thanks for the fix, @wolfgangmm, and for the review & merge, @adamretter!

dizzzz:
Jay!

adamretter:
@joewiz I wonder if it also closed https://github.com/eXist-db/exist/issues/1298

Issue #1476, opened 2017-07-06
local-name() on empty set raises XPDY0002 (undef context item)


tuurma:
### What is the problem

```xquery

let $xml:= <a/>
return $xml/blah/local-name()
```
raises an error instead of returning empty sequence

err:XPDY0002 Undefined context item

### What did you expect

Expected behaviour is the same as if the context item was explicitly passed as local-name()'s argument.

```xquery
let $xml:= <a/>
return $xml/blah/local-name(.)
```

### Context information
Please always add the following information
tested on eXist-db 3.2.0 (Windows) and 3.0RC2 (Linux)

tuurma:
https://www.w3.org/TR/xpath/#function-local-name

>If the argument is omitted, it defaults to a node-set with the context node as its only member.

PR #1561, opened 2017-09-25
Fixed fn name functions to work with the Context Item

adamretter:
@tuurma this should fix your issue and then some ;-)
Closes https://github.com/eXist-db/exist/issues/1476


dizzzz:
@adamretter it looks good to me, but I need to have an explanation (to prevent i will make similar bugs in the future...)

adamretter:
@dizzzz the zero arity version of those functions, did not retrieve the context item correctly. Take a look at the tests that I added :-)

joewiz:
@adamretter Thank you!  I had occasionally run into this and didn't realize that some of these functions had zero-arity versions.  To add to the explanation in the test cases, here's what I would bump into:

While `/name()` would work, e.g.,:

```xquery
xquery version "3.1";

element tei:persName { "Dannes" }/name()
```

`node-name()` would return an error:

```xquery
element tei:persName { "Dannes" }/node-name()
```

would return:

> Unexpectedly received 0 parameter(s) in call to function 'node-name()'. Defined function signatures are: node-name($arg as node()?) xs:QName?

The workaround was to specify a context item, e.g., `/node-name(.)`. 

(I wonder if we need to do a review of eXist's function signatures against the spec? Or perhaps integrating the QT3 test suite, if only partially to focus on function signatures...)

joewiz:
Also, for context, here's the link to the current spec for this function: https://www.w3.org/TR/xpath-functions-31/#func-local-name.

PR #1580, opened 2017-10-12
Make SessionManager thread-safe

adamretter:
Previously the Session Manager which cached query results for the REST Server was not thread-safe, yet was accessed from at least two concurrent threads, and possibly more if there was more than one concurrent user. This PR both fixed the concurrency issues, and also simplifies the implementation.

### DIFF
--- a/src/org/exist/http/SessionManager.java
+++ b/src/org/exist/http/SessionManager.java
@@ -16,141 +16,78 @@
  *  You should have received a copy of the GNU Lesser General Public
  *  License along with this library; if not, write to the Free Software
  *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
- *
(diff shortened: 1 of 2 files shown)


src/org/exist/http/SessionManager.java
```diff
@@ -16,141 +16,78 @@
...
-            }
-        }
-
-        final Properties props = new Properties();
-        props.put(SESSION_MANAGER, this);
-        scheduler.createPeriodicJob(TIMEOUT_CHECK_PERIOD, task, 2000, props);
+        cache = cacheBuilder.build();
     }
 
     public int add(final String query, final Sequence sequence) {
```
dizzzz: queryString ?

adamretter: @dizzzz I have renamed that now

PR #1581, opened 2017-10-12
Make XML RPC Query Result Cache thread-safe

adamretter:
Previously the  cache of `XmldbRequestProcessorFactory` which cached query results for the XML RPC Server was not thread-safe.
This PR both fixed the concurrency issues, and also simplifies the implementation.

Very similar to https://github.com/eXist-db/exist/pull/1580

### DIFF
--- a/src/org/exist/xmlrpc/QueryResultCache.java
+++ b/src/org/exist/xmlrpc/QueryResultCache.java
@@ -19,99 +19,82 @@
  */
 package org.exist.xmlrpc;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import net.jcip.annotations.ThreadSafe;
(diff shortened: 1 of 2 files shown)


src/org/exist/xmlrpc/QueryResultCache.java
```diff
@@ -19,99 +19,82 @@
...
 public class QueryResultCache {
 
-    public final static int TIMEOUT = 180000;
-
-    private static final int INITIAL_SIZE = 254;
-
-    private AbstractCachedResult[] results;
-
     private static final Logger LOG = LogManager.getLogger(QueryResultCache.class);
+    private static final int TIMEOUT = 180000;
```
dizzzz: What is the unit? Msecs? Seconds?

dizzzz: Milliseconds thnx

PR #1582, opened 2017-10-12
Allow clean-default-data-dir to complete if any dirs are not present

joewiz:
### Description:

All `<delete>` tasks in clean-related targets use the `@failonerror="no"` attribute, to allow the task to proceed and complete even if any of the named directories are not present. This PR applies this attribute to the delete tasks in the `clean-default-data-dir` target. 

### Reference:

n/a

### Type of tests:

n/a


adamretter:
@joewiz there seems to be a small bug in your XML, can you check the Travis CI output please?

joewiz:
Sorry, I just noticed that. I've rebased and force-pushed the correction.

PR #1588, opened 2017-10-14
[feature] Added support for AppVeyor CI (Windows)

adamretter:
AppVeyor is a CI which runs on Windows platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 ## eXist Native XML Database
 
-[![Build Status](https://travis-ci.org/eXist-db/exist.png?branch=develop)](https://travis-ci.org/eXist-db/exist) [![Java 8](https://img.shields.io/badge/java-8-blue.svg)](http://java.oracle.com) [![License](https://img.shields.io/badge/license-LGPL%202.1-blue.svg)](https://www.gnu.org/licenses/lgpl-2.1.html)
+[![Build Status](https://travis-ci.org/eXist-db/exist.png?branch=develop)](https://travis-ci.org/eXist-db/exist) [![Build status](https://ci.appveyor.com/api/projects/status/27jtt2m7s4qvey11/branch/develop?svg=true)](https://ci.appveyor.com/project/AdamRetter/exist/branch/develop) [![Java 8](https://img.shields.io/badge/java-8-blue.svg)](http://java.oracle.com) [![License](https://img.shields.io/badge/license-LGPL%202.1-blue.svg)](https://www.gnu.org/licenses/lgpl-2.1.html)
 [![Download](https://api.bintray.com/packages/existdb/releases/exist/images/download.svg)](https://bintray.com/existdb/releases/exist/_latestVersion)
 [![HipChat](https://img.shields.io/badge/hipchat-eXist–db-blue.svg)](https://www.hipchat.com/gEBQ3SNfp)
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,7 @@
+version: '{branch}-{build}'
+environment:
+  EXIST_HOME: c:\projects\exist
+build_script:
+- cmd: build.bat
+test_script:
+- cmd: build.bat -Dexist.autodeploy=off -Dtest.haltonerror=true -Dtest.haltonfailure=true travis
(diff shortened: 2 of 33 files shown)


joewiz:
Hmm, interesting!  Is the idea to be able to ensure the test suite succeeds when building on Windows? 

You've probably seen this, but it looks like there's some problem on the Appveyor build with `BackupRestoreSecurityPrincipalsTest`:

```text
test-local:
     [echo] ------------------------------------------
     [echo] Running tests on a local database instance
     [echo] ------------------------------------------
    [junit] Running org.exist.backup.BackupRestoreSecurityPrincipalsTest
    [junit] Tests run: 1, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 5.757 sec
BUILD FAILED
C:\projects\exist\build\scripts\junit.xml:124: Test org.exist.backup.BackupRestoreSecurityPrincipalsTest failed
```

 I wonder what could be Windows-specific about this - perhaps nio or file system issues?

adamretter:
Exactly.

And yes, the test suite fails on Windows and I can confirm that on a local Windows installation here. I guess no one has run the test suite for a very long time on Windows, all the core-developers are on Mac or Linux AFAIK. So the Windows test failures actually reveals a subtle bug in shutdown where we are not closing some files that we have opened, I am preparing some fixes now.

joewiz:
Sounds like a valuable addition to eXist's CI infrastructure!

adamretter:
@joewiz Thanks, I would like to resolve the Windows test failures as part of this PR before we merge it now.

joewiz:
That sounds prudent.  Hmm, `java.nio.file.AccessDeniedException`... sounds tricky.

adamretter:
@joewiz It's another leaked file handle, this time from `util:binary-doc`. I can likely fix it fairly easily, I just have to finish something else first...

README.md
```diff
@@ -6,7 +6,7 @@
 
 ## eXist Native XML Database
 
-[![Build Status](https://travis-ci.org/eXist-db/exist.png?branch=develop)](https://travis-ci.org/eXist-db/exist) [![Java 8](https://img.shields.io/badge/java-8-blue.svg)](http://java.oracle.com) [![License](https://img.shields.io/badge/license-LGPL%202.1-blue.svg)](https://www.gnu.org/licenses/lgpl-2.1.html)
+[![Build Status](https://travis-ci.org/eXist-db/exist.png?branch=develop)](https://travis-ci.org/eXist-db/exist) [![Build status](https://ci.appveyor.com/api/projects/status/27jtt2m7s4qvey11/branch/develop?svg=true)](https://ci.appveyor.com/project/AdamRetter/exist/branch/develop) [![Java 8](https://img.shields.io/badge/java-8-blue.svg)](http://java.oracle.com) [![License](https://img.shields.io/badge/license-LGPL%202.1-blue.svg)](https://www.gnu.org/licenses/lgpl-2.1.html)
```
joewiz: @adamretter It looks like there's a way to use the organization rather than your personal account. See https://www.appveyor.com/docs/team-setup/#setting-up-appveyor-account-for-github-organization.

adamretter:
@joewiz @wolfgangmm Okay this fixes the eXist build on Windows.
If we merge this, then the AppVeyor CI should now pass. Any subsequent PRs may need rebasing atop this, as most of them are mine I am happy to do that as soon as this is merged... Thanks

It also fixes a number of file handle leaks that were identified when running eXist on Windows, it is likely these were also present on Linux and Mac as well (but not easily spotted).

joewiz:
@adamretter It's great to hear the test suite now passes on Windows! 

My only comment is rather surface-level: that it might be preferable in configuring the appveyor survice to use the eXist-db organization account rather than your personal account. See https://www.appveyor.com/docs/team-setup/#setting-up-appveyor-account-for-github-organization. I wouldn't hold up the PR for it, but that's the only improvement here I can think of.

adamretter:
@joewiz Unfortunately there is no specific "organization" thing in AppVeyor, it just relies on you creating a new account named after your organisation etc. Unfortunately I do not have the permissions to create the appropriate email account for exist-db.org, so I think if this is really desirable, we would need to have @wolfgangmm do this. However, the question then is where the emails should actually go for AppVeyor and if anyone will monitor them.

I would suggest that for now, we just merge this as is, and we can always easily change it later if need be.

joewiz:
Ah, okay, then let's get started with AppVeyor as you've configured it.

Issue #944, opened 2016-03-16
%test:args with a function argument that has either no explicit type or an explicit type of node() doesn't work


welblaud:
````xquery
xquery version "3.0";

module namespace cust = 'http://46.28.111.241:8081/exist/db/apps/karolinum-x/modules/cust';

declare namespace tei  = 'http://www.tei-c.org/ns/1.0';
declare namespace test = 'http://exist-db.org/xquery/xqsuite';
declare
    %test:args('<TEI xmlns="http://www.tei-c.org/ns/1.0">
                    <text>
                        <body>
                            <div n="1">
                                <head>Heading</head>
                                <p>paragraph</p>
                            </div>
                        </body>
                    </text>
                </TEI>')
    %test:assertXPath('$result/html')
function cust:transform($doc as node()*) as node() {
    let $styles := doc('/db/apps/karolinum-x/resources/xslt/style-web.xsl')
    let $document := 
        (
            <book n='1'>{($doc//tei:div[@n='1'])[1]}</book>
        )
    let $finale := transform:transform($document, $styles, ())
    return $finale
};
```
This test throws:

```xml
<testsuites>
    <testsuite package="http://46.28.111.241:8081/exist/db/apps/karolinum-x/modules/cust" timestamp="2016-03-15T12:53:16.5+01:00" failures="0" pending="0" tests="1" time="PT0.002S">
        <testcase name="transform" class="cust:transform">
            <error type="err:XPTY0004" message="It is a type error if,
                during the static analysis phase, an expression is found to have a
                static type that is not appropriate for the context in which the
                expression occurs, or during the dynamic evaluation phase, the dynamic 
                type of a value does not match a required type as specified by the 
                matching rules in 2.5.4 SequenceType Matching."/>
        </testcase>
    </testsuite>
</testsuites>
```
As a workaround if I change `$doc as node()*` to `$doc as element()*` then that does seem to work.

PietroLiuzzo:
I have encountered the same issue.

duncdrum:
@PietroLiuzzo can you try with 3.5.1, i think  #1593 might have something to do with this.

PietroLiuzzo:
I am on the latest stable release, you mean I should test with the latest dev version from git?

PietroLiuzzo:
the element() workaround did work for me as well

duncdrum:
Yes please, alternatively report back when the new version is released. Its just a hunch right now that the PR might have fixed this problem as well.

[...]

PR #1593, opened 2017-10-16
Fix fn:doc return type

joewiz:
### Description:

The return type of `fn:doc()` should be `document-node()?`. This PR corrects the return type in eXist, which was `node()?` - and could lead people to mistakenly believe that the function return other kinds of nodes, such as elements.

### Reference:

- https://www.w3.org/TR/xpath-functions/#func-doc
- https://www.w3.org/TR/xpath-functions-30/#func-doc
- https://www.w3.org/TR/xpath-functions-31/#func-doc

### Type of tests:

XQSuite

### DIFF
--- a/test/src/xquery/fn.xql
+++ b/test/src/xquery/fn.xql
@@ -16,7 +16,7 @@ function fnt:store() {
     let $col := xmldb:create-collection("/db", "fn-test")
     return
         (
-            xmldb:store($col, "test.xml", <something/>),
+            xmldb:store($col, "test.xml", <books><book/></books>),
             xmldb:store($col, "test.bin", "some binary", "application/octet-stream")
         )
 };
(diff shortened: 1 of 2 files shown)


test/src/xquery/fn.xql
```diff
@@ -57,6 +57,20 @@ function fnt:doc-available($filename as xs:string) {
...
+    %test:args("test.xml")
+    %test:assertEquals("true")
+function fnt:doc($filename as xs:string) {
+    fn:doc("/db/fn-test/" || $filename) instance of document-node()
+};
+
+declare
+    %test:args("test.xml")
+    %test:assertEmpty
+function fnt:doc($filename as xs:string) {
```
adamretter: @joewiz We seem to have two functions of the same name and arity, i.e. `fnt:doc#1`. This should cause the test suite to fail, although it seems to have passed on Travis at least?!?

joewiz: Drat, I had even given them unique names when testing in eXide but forgot to put this change into my comit. How would you recommend I fix this: push the fix to my branch and submit a new PR? I'd name these `fnt:doc-returns-document-node` and `fnt:doc-does-not-return-element`.

adamretter: A new PR would be great. I don't understand why the test suite doesn't currently fail though, are your tests being executed I wondered?

adamretter: @joewiz Any update on this please?

joewiz: @adamretter Thanks for the reminder. @duncdrum's comment on #944 reminded me too. A PR is coming shortly.

PR #1594, opened 2017-10-16
Changes the shebang line

tasmo:
Using `/bin/env` in favour of direct calling `/bin/bash`
Makes shell script suitable for Linux/Unix beside Debian/Ubuntu

### Description:
While trying to execute any shell script on other Linux machines which doesn't run Debian or Ubuntu I get an error that I run the wrong operating system.

### Reference:
```bash
Failed to execute process 'bin/client.sh'. Reason:
The file 'bin/client.sh' specified the interpreter '/bin/bash', which is not an executable command.
```

### Type of tests:
Running some scripts including `./build.sh test` successfully.


joewiz:
@tasmo Makes sense. 

For additional context, see:

- https://stackoverflow.com/questions/16365130/the-difference-between-usr-bin-env-bash-and-usr-bin-bash
- https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my

adamretter:
@tasmo I am not yet convinced that this is a good change. In one of the articles that @joewiz linked, it says:

> But in situations where security is the focus, the latter would be preferred, as it limits code injection possibilities.

In that instance the *"latter"* approach is: `/bin/bash`.

I would be interested to know on what machines you have actually experienced this problem? I personally have run eXist on many different Linux/Unix machines and so far have never experienced this problem on: CentOS, RHEL, Debian, Ubuntu, Mint, FreeBSD and Sun Solaris.

tasmo:
@adamretter I cannot argue against the mentioned security risk.

The two Linux distributions I'm using don't have bash in the path `/bin` - namely Arch Linux and NixOS. But env is in the path `/usr/bin` in every actual UNIX.

adamretter:
@tasmo I am looking at other ways to help you with this. I am just trying to understand why Arch Linux or NixOS wouldn't have bash in the usual place, so...

I just looked at a clean install of Arch 2017.10.01, and it does seem to have `/bin/bash`: 

What was the specific problem you saw with Arch Linux?

I will look into NixOS...

tasmo:
@adamretter You're right. On Arch my default shell appears as in `/usr/bin/bash` but there is the same binary installed in `/bin` as well.

adamretter:
@tasmo Okay I can see the issue with NixOS... hmmm.

So I think we have 2 options here:

1. We switch all our `.sh` scripts from `#!/bin/bash` to `#!/usr/bin/env bash` as you suggested, or

2. We advise users of NixOS to implement one of these workarounds:
    1. Just run `bash` first, e.g. `$ bash ./build.sh`
    2. Create a FHS-compatible lightweight sandbox as per the NixOS documentation: https://nixos.org/nixpkgs/manual/#sec-fhs-environments

I am really not sure which is better, I can see the attraction of changing to `#!/usr/bin/env bash`. However, I worry this might break things for other users, although I would be willing to try it and we could always revert it if there are any problems.

I would like really like some opinions from @dizzzz @wolfgangmm @ljo on this, before we merge...

tasmo:
Another solution for NixOS could be to use `#!/bin/sh`. This step would miss many features in the scripts which bash has.

tasmo:
@adamretter It's even worse:

```shell
$ bin/startup.sh                                                      
                                                                                           
/opt/exist/bin/functions.d/eXist-settings.sh: line 14: /usr/bin/dirname: No such file or directory
/opt/exist/bin/functions.d/eXist-settings.sh: line 14: /bin/pwd: No such file or directory
Unable to find start.jar. Please set EXIST_HOME to point to your installation directory.
```

adamretter:
@tasmo I think that we can just remove the paths for `pwd` and `dirname` as they should be available in the path

dizzzz:
Do we have the ambition to support every single deviating Linux/Unix distribution? IMO we can spend our time on more valueable issues.

adamretter:
@tasmo I feel confident that we can accept this. Thanks for your contribution :-)

PR #1595, opened 2017-10-16
WIP fix: prevent error in util:get-document-by-absolute-id

line-o:
### Description:

`util:get-document-by-absolute-id` returns an empty sequence if document was not found instead of throwing an error.

### Reference:

none

### Type of tests:

TBD

### DIFF
--- a/src/org/exist/storage/NativeBroker.java
+++ b/src/org/exist/storage/NativeBroker.java
@@ -2021,6 +2021,9 @@ public class NativeBroker extends DBBroker {
             //get the resource uri
             final Value key = new CollectionStore.DocumentKey(collectionId, resourceType, documentId);
             final VariableByteInput vbi = collectionsDb.getAsStream(key);
+            if (vbi == null) {
+                return null;
+            }
             vbi.readInt(); //skip doc id
             final String resourceUri = vbi.readUTF();
 
--- a/src/org/exist/xquery/functions/util/DocumentNameOrId.java
+++ b/src/org/exist/xquery/functions/util/DocumentNameOrId.java
@@ -163,6 +163,9 @@ public class DocumentNameOrId extends BasicFunction {
                     final int collectionId = absoluteId.and(BigInteger.valueOf(0xFFFFFFFF)).intValue();
                     
                     doc = context.getBroker().getResourceById(collectionId, resourceType, documentId);
+                    if (doc == null) {
+                        return Sequence.EMPTY_SEQUENCE;
+                    }
                     if(doc instanceof BinaryDocument) {
                         final BinaryDocument bin = (BinaryDocument) doc;
                         final InputStream is = context.getBroker().getBinaryResource(bin);


src/org/exist/storage/NativeBroker.java
```diff
@@ -2021,6 +2021,9 @@ public DocumentImpl getResourceById(final int collectionId, final byte resourceT
             //get the resource uri
             final Value key = new CollectionStore.DocumentKey(collectionId, resourceType, documentId);
             final VariableByteInput vbi = collectionsDb.getAsStream(key);
+            if (vbi == null) {
```
adamretter: I am not sure that this change is meant to be in this PR? It doesn't seem to be related to the topic of the PR. Can you explain please?

src/org/exist/xquery/functions/util/DocumentNameOrId.java
```diff
@@ -163,6 +163,9 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
                     final int collectionId = absoluteId.and(BigInteger.valueOf(0xFFFFFFFF)).intValue();
                     
                     doc = context.getBroker().getResourceById(collectionId, resourceType, documentId);
+                    if (doc == null) {
```
adamretter: Looks good. I think it would also make sense to add Javadoc on `DBBroker#getResourceById(int, byte, int)` so that it documents that the return type might be null. Would you be able to do that too please?

src/org/exist/storage/NativeBroker.java
```diff
@@ -2021,6 +2021,9 @@ public DocumentImpl getResourceById(final int collectionId, final byte resourceT
...
+                return null;
+            }
             vbi.readInt(); //skip doc id
```
line-o: This is the actual code that throws. So we must return null beforehand.

src/org/exist/storage/NativeBroker.java
line-o: It is necessary in order to avoid try catch around the call to 
```
 doc = context.getBroker().getResourceById(collectionId, resourceType, documentId);
```

src/org/exist/xquery/functions/util/DocumentNameOrId.java
line-o: Will do that

line-o:
As I am not allowed to push to this repo, I made the changes here
https://github.com/line-o/exist/tree/fix/get-document-by-absolute-id
Should I create a new PR?

src/org/exist/storage/NativeBroker.java
adamretter: Hmm... doesn't a `null` here imply that the `collections.dbx` has become corrrupted?

src/org/exist/storage/NativeBroker.java
adamretter: Hmm... doesn't a null here imply that the collections.dbx has become corrrupted?

adamretter:
I understand that @wolfgangmm is addressing with @line-o

adamretter:
@line-o @wolfgangmm I am going to close this as no progress is being made. Please feel free to re-open if you have new stuff

PR #1596, opened 2017-10-16
fix consistency check report

line-o:
### Description:

If a DB instance contains an XML document without any children running sanity-check on the command line will crash with an NPE.
This fix will prevent the app from crashing and report the name of the XML file.

### Reference:

None

### Type of tests:

None

### DIFF
--- a/src/org/exist/backup/ConsistencyCheck.java
+++ b/src/org/exist/backup/ConsistencyCheck.java
@@ -289,7 +289,7 @@ public class ConsistencyCheck
                 try {
                     final ElementImpl root = (ElementImpl)doc.getDocumentElement();
                     if (root == null) {
-                        return new ErrorReport.ResourceError(ErrorReport.RESOURCE_ACCESS_FAILED, "Failed to access document data");
+                        return new ErrorReport.ResourceError(ErrorReport.RESOURCE_ACCESS_FAILED, "Failed to access data of document " + doc.getDocumentURI() );
                     }
                 } catch( final Exception e ) {
                     e.printStackTrace();
@@ -324,6 +324,10 @@ public class ConsistencyCheck
                         EmbeddedXMLStreamReader reader = null;
                         try {
                             final ElementImpl             root            = (ElementImpl)doc.getDocumentElement();
+                            if (root == null) {
+                                return new ErrorReport.ResourceError(ErrorReport.RESOURCE_ACCESS_FAILED, "Failed to access data of document " + doc.getDocumentURI() );
+                            }
+
                             reader = (EmbeddedXMLStreamReader)broker.getXMLStreamReader( root, true );
                             NodeId                  nodeId;
                             boolean                 attribsAllowed  = false;


src/org/exist/backup/ConsistencyCheck.java
```diff
@@ -289,7 +289,7 @@ public Object start() {
                 try {
                     final ElementImpl root = (ElementImpl)doc.getDocumentElement();
                     if (root == null) {
-                        return new ErrorReport.ResourceError(ErrorReport.RESOURCE_ACCESS_FAILED, "Failed to access document data");
+                        return new ErrorReport.ResourceError(ErrorReport.RESOURCE_ACCESS_FAILED, "Failed to access data of document " + doc.getDocumentURI() );
```
adamretter: Nice. However, I think that `doc.getURI()` would be a safer choice than `doc.getDocumentURI()`. Could you change that please?

src/org/exist/backup/ConsistencyCheck.java
```diff
@@ -324,6 +324,10 @@ public Object start() {
                         EmbeddedXMLStreamReader reader = null;
                         try {
                             final ElementImpl             root            = (ElementImpl)doc.getDocumentElement();
+                            if (root == null) {
+                                return new ErrorReport.ResourceError(ErrorReport.RESOURCE_ACCESS_FAILED, "Failed to access data of document " + doc.getDocumentURI() );
```
adamretter: Looks good in principle, thank you.

I think we could improve the error message though, perhaps something like:

```java
ErrorReport.ResourceError(ErrorReport.NODE_HIERARCHY, "Document has no document element: " + doc.getURI());
```

joewiz: Or "Document has no root element"? (Not sure what a "document element" is?)

adamretter: @joewiz i.e. https://docs.oracle.com/javase/8/docs/api/org/w3c/dom/Document.html#getDocumentElement-- ;-)

joewiz: Sorry, I'm too XDM-centric! ;)

line-o:
Thank you for your feedback @adamretter. Looking at the changes it looks like they are actually handling the same error. One could even think about extracting that into a method.

```java
final ElementImpl root = (ElementImpl)doc.getDocumentElement();
if (root == null) {
    return new ErrorReport.ResourceError(
        ErrorReport.NODE_HIERARCHY, 
        "Document has no document element: " + doc.getURI()
    );
}
```
What do you think?

line-o:
As I can not push to this PR, I made the changes here
https://github.com/line-o/exist/tree/fix/consistency-check-report
Should I create a new PR and close this one?

adamretter:
@line-o if you push to the original branch, it will update the PR.

adamretter:
I understand that @wolfgangmm is addressing with @line-o

adamretter:
@line-o @wolfgangmm I am going to close this as no progress is being made. Please feel free to re-open if you have new stuff

Issue #1585, opened 2017-10-12
Add support for XQuery 3.1 html-ascii-case-insensitive collation


joewiz:
### What is the problem

According to the XQuery 3.1 F&O spec (https://www.w3.org/TR/xpath-functions-31/#html-ascii-case-insensitive-collation):

> The collation URI `http://www.w3.org/2005/xpath-functions/collation/html-ascii-case-insensitive` must be recognized by every implementation. 

eXist has not yet implemented this collation, which allows for a case-insensitive comparison of ASCII characters (`A-Z` = `a-z`). 

### What did you expect

This query should return true:

```xquery
xquery version "3.1";

declare default collation 'http://www.w3.org/2005/xpath-functions/collation/html-ascii-case-insensitive';

'HTML' = 'html'
```

### Describe how to reproduce or add a test

See the official QT3 test suite, namely, tests for fn:compare, fn:contains, fn:starts-with, fn:ends-with, etc. 

### Context information
Please always add the following information

- eXist-db version + Git Revision hash: eXist 3.5.0 12aea86
- Java version: 1.8.0_144
- Operating system: macOS 10.13
- 32 or 64 bit: 64 bit
- Any custom changes in e.g. conf.xml: none

joewiz:
See the BaseX implementation: https://github.com/BaseXdb/basex/blob/b0aac93628d5b3c192aab5e9499a8382e62b84c9/basex-core/src/main/java/org/basex/query/util/collation/NoCaseCollation.java

PR #1603, opened 2017-10-21
Implement the HTML ASCII Case Insensitive Collation as defined in XPath F&O 3.1

adamretter:
Closes https://github.com/eXist-db/exist/issues/1585

@joewiz I am certainly no collation expert, however I think this likely solves it.

### DIFF
--- a/src/org/exist/util/Collations.java
+++ b/src/org/exist/util/Collations.java
@@ -681,10 +681,12 @@ public class Collations {
     private static Collator getHtmlAsciiCaseInsensitiveCollator() throws Exception {
         Collator collator = htmlAsciiCaseInsensitiveCollator.get();
         if (collator == null) {
-            htmlAsciiCaseInsensitiveCollator.compareAndSet(null,
-                    new RuleBasedCollator("&a=A &b=B &c=C &d=D &e=E &f=F &g=G &h=H "
+            collator = new RuleBasedCollator("&a=A &b=B &c=C &d=D &e=E &f=F &g=G &h=H "
                     + "&i=I &j=J &k=K &l=L &m=M &n=N &o=O &p=P &q=Q &r=R &s=S &t=T "
-                    + "&u=U &v=V &w=W &x=X &y=Y &z=Z").freeze());
+                    + "&u=U &v=V &w=W &x=X &y=Y &z=Z");
+            collator.setStrength(Collator.PRIMARY);
+            htmlAsciiCaseInsensitiveCollator.compareAndSet(null,
+                    collator.freeze());
             collator = htmlAsciiCaseInsensitiveCollator.get();
         }
 
--- a/src/org/exist/util/Collations.java
+++ b/src/org/exist/util/Collations.java
@@ -26,6 +26,7 @@ import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
 import java.util.StringTokenizer;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 import com.ibm.icu.text.*;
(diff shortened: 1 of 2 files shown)


joewiz:
@adamretter This PR passes all of relevant QT3 tests except these two from `fn-contains`:

```xml
   <test-case name="liam-contains-001" covers="html-ascii-case-insensitive">
      <description>Check for presence of HTML ascii case-insensitive collation</description>
      <created by="Liam Quin" on="2009-09-08"/>      
      <modified by="Michael Dyck" on="2014-09-11" change="parenthesize IfExpr's condition"/>
      <modified by="Josh Spiegel" on="2014-09-16" change="Added XP31+ dependency."/>
      <modified by="Michael Kay" on="2014-09-16" change="Added XQ31+ dependency."/>
      <dependency type="spec" value="XP31+ XQ31+"/>
      <test>if (fn:contains('iNPut', 'pu', 'http://www.w3.org/2005/xpath-functions/collation/html-ascii-case-insensitive')) then 'pass' else 'huh'</test>
      <result>
         <assert-string-value>pass</assert-string-value>
      </result>
   </test-case>
   <test-case name="liam-contains-002" covers="html-ascii-case-insensitive">
      <description>Check for presence of HTML ascii case-insensitive collation</description>
      <created by="Liam Quin" on="2009-09-08"/>      
      <modified by="Michael Dyck" on="2014-09-11" change="parenthesize IfExpr's condition"/>
      <modified by="Josh Spiegel" on="2014-09-16" change="Added XP31+ dependency."/>
      <modified by="Michael Kay" on="2014-09-16" change="Added XQ31+ dependency."/>
      <dependency type="spec" value="XP31+ XQ31+"/>
      <test>if (fn:contains('iNPut', 'PU', 'http://www.w3.org/2005/xpath-functions/collation/html-ascii-case-insensitive')) then 'pass' else 'huh'</test>
      <result>
         <assert-string-value>pass</assert-string-value>
      </result>
   </test-case>
```

eXist returns `huh` instead of `pass` for these two tests.

adamretter:
@joewiz https://stackoverflow.com/questions/47022995/html-ascii-case-insensitive-icu-collator

dizzzz:
what is the status of this PR ?

joewiz:
@adamretter I'm not sure if this helps or not, but I checked how BaseX approaches this, and it appears they don't use UCA for this collator: 

- https://github.com/BaseXdb/basex/blob/master/basex-core/src/main/java/org/basex/query/util/collation/Collation.java#L23-L25
- https://github.com/BaseXdb/basex/blob/master/basex-core/src/main/java/org/basex/query/util/collation/NoCaseCollation.java

Their documentation backs this up - see http://docs.basex.org/wiki/XQuery_3.1#Collations:

> XQuery 3.1 provides a new default collation, which allows for a case-insensitive comparison of ASCII characters (A-Z = a-z)... 
>
> If the ICU Library is downloaded and added to the classpath, the full Unicode Collation Algorithm features get available in BaseX...

I read this as saying that case-insensitive ASCII comparison is available even if the ICU library isn't available.

adamretter:
@dizzzz Still just chasing a final issue.

adamretter:
@joewiz Indeed BaseX has its own simple implementation of a collator for this. However, I think for us it will be easier to use ICU4j if we can figure it out.

ljo:
A bit delayed, but I am on it now.

adamretter:
@ljo Any news please?

ljo:
Just to let you know it almost there, was hoping for last night already, but hopefully running the last test round now.

adamretter:
@dizzzz @ljo It seems I only needed the `asciiCollator.setStrength(Collator.PRIMARY);` bit. I have updated the PR.

@joewiz Hopefully this works for you now?

PR #1605, opened 2017-10-27
Provide unique names for XQSuite tests

joewiz:
### Description:

PR #1593 included two new XQSuite tests for the `fn:doc` function. Mistakenly, I gave these two tests the same name. This PR gives each test a unique name.

### Reference:

#1593.

### Type of tests:

n/a


adamretter:
@wolfgangmm Any idea why the test suite passed before?

PR #1617, opened 2017-11-15
[bugfix] Make sure the relative base URI is absolute

adamretter:
Fixes an issue when using automatic font detection.


joewiz:
@welblaud This may help you throw away some of your workarounds.

Issue #1421, opened 2017-04-20
Too many opened HttpSession


alebastrov:
Each HttpClient request to an application creates  a new HtppSession, which will be never used after. We should have possibility to release a HttpSession after having request processed. for instance with adding some header to HttpClient and in XQueryServlet we have to invalidate session in case this header is provided. It will allow us to save some memory during backup/restore procedure, because each file needs a single separate request, which creates a session

dizzzz:
Please can you fill in the template for issues?
I have no clue what you are referring to, we need more structured input.

Thnx

alebastrov:
Just do:
- add SessionCountListener to your App
- try to restore some backup
- watch how many HttpSessions awere opened but not closed.

In fact restoring each xml document will create an HttpSession, which is never closed (only by timeout)
Moreover, restoring a collections creates HttpSessions as well.

Of course I tested it on existDB 2.2 on Windows 7 Java 1.8.0 x64

dizzzz:
sorry this is from me far from sufficient to have a look at the issue; i really expect you to fill-in the template, and provide all we need that we can easily reproduce your issue in a few seconds. The description makes no sense to me and probably many of us.

we are all volunteers here, so help us to help you.

adamretter:
@alebastrov Can you provide the absolute minimum app which allows us to reproduce your issue please?

dizzzz:
As no response, i propose to close the issue;

adamretter:
@dizzzz We should probably allow 30 days. I think we could set a policy generally whereby if we don't have a response within 30 days, we send one reminder, and if no response in 14 days then we close and label as "abandoned". How does that sound?

alebastrov:
I wrote how to reproduce it in seconds. Unfortunately I do not know how to do it with tests

adamretter:
@alebastrov One of the pieces of information you did not provide was how you restored the backup.

I can however confirm that your issue happens when using the XML:DB / XML-RPC interfaces from the Java Admin Client.
Basically the Java Admin Client (which uses XML:DB over XML-RPC) is session-less , so every HTTP connection it makes to the server causes the server to create a new session.

I think we have two options here:
1) Configure the eXist-db Server not to create sessions at all on the XML:DB / XML-RPC interface.
2) Configure the eXist-db Java Admin Client to reuse the session id returned by the server from the first request.

I guess the question is whether the XML-RPC protocol should be stateful or not. I will do some looking into this and thinking about it...

[...]

PR #1624, opened 2017-11-21
XML-RPC Improvements

adamretter:
A few small improvements found whilst investigating https://github.com/eXist-db/exist/issues/1421


src/org/exist/xmldb/DatabaseImpl.java
```diff
@@ -303,6 +303,9 @@ private XmlRpcClient getRpcClient(final String user, final String password, fina
         final String key = user + "@" + url.toString();
         final XmlRpcClientConfigImpl config = new XmlRpcClientConfigImpl();
         config.setEnabledForExtensions(true);
+        config.setContentLengthOptional(true);
+        config.setGzipCompressing(true);
```
dizzzz: not sure.... does this force GZIP compression? If so, the client need be configured for this as well.
That said... do we need client settings?  in org.exist.xmldb.DatabaseImpl#getRpcClient we need to do the same, maybe?

dizzzz: (I reworked the PR)

dizzzz: indeed a bit concerned on backward compatibility when reading the java doc:

https://ws.apache.org/xmlrpc/apidocs/org/apache/xmlrpc/common/XmlRpcHttpRequestConfigImpl.html#setGzipCompressing(boolean)

adamretter: @windauer `DatabaseImpl.java` is client side. You are looking at the client side settings!

Backwards compatibility is not an issue as Apache XML-RPC uses a auto-detect mechanism on the server, to detect and support whatever the client requests. It is documented here - https://ws.apache.org/xmlrpc/extensions.html

dizzzz: @adamretter yes , I (not lars) checked indeed the client size; as you see, I linked to the same page earlier......

Issue #1618, opened 2017-11-16
expath http client issue 


windauer:
### What is the problem
NPE in 3.6.0-SNAPSHOT for EXPath HTTP request which works perfectly fine on 3.4.x

### What did you expect
Expected was a response sequnce with header information in response[1] and the response itself in response[2]

### Describe how to reproduce or add a test
```
import module namespace http="http://expath.org/ns/http-client";

let $user := "user"
let $pass:= "pass
let $url := "https://some-https-url-with-basic-auth"

let $request := <http:request method="GET" href="{$url}" timeout="120"  username="{$user}" password="{$pass}"
                        auth-method="basic" send-authorization="true" override-media-type="application/octet-stream">
            <http:header name="accept" value="application/xml"/>
        </http:request>

return
    <result>{ http:send-request($request) }</result>
```

From the log files: 

```
2017-11-16 16:27:24,267 [qtp1999415998-103] ERROR (XQueryServlet.java [process]:534) - null 
java.lang.NullPointerException: null
        at org.expath.httpclient.impl.ApacheHttpConnection.setCredentials(ApacheHttpConnection.java:415) ~[http-client-java-1.0-SNAPSHOT.jar:1.0-SNAPSHOT]
        at org.expath.httpclient.impl.ApacheHttpConnection.connect(ApacheHttpConnection.java:98) ~[http-client-java-1.0-SNAPSHOT.jar:1.0-SNAPSHOT]
        at org.expath.httpclient.impl.HttpRequestImpl.send(HttpRequestImpl.java:56) ~[http-client-java-1.0-SNAPSHOT.jar:1.0-SNAPSHOT]
        at org.expath.exist.SendRequestFunction.sendOnceWithAuth(SendRequestFunction.java:206) ~[exist-expath-20130805.jar:3.6.0-SNAPSHOT]
        at org.expath.exist.SendRequestFunction.sendOnce(SendRequestFunction.java:172) ~[exist-expath-20130805.jar:3.6.0-SNAPSHOT]
        at org.expath.exist.SendRequestFunction.sendRequest(SendRequestFunction.java:151) ~[exist-expath-20130805.jar:3.6.0-SNAPSHOT]
        at org.expath.exist.SendRequestFunction.eval(SendRequestFunction.java:133) ~[exist-expath-20130805.jar:3.6.0-SNAPSHOT]
[...]

PR #1627, opened 2017-11-21
Fixes a StackOverflow in the HistoryTrigger

adamretter:
Closes https://github.com/eXist-db/exist/issues/1618


adamretter:
@dizzzz Actually I added a test showing that `beforeCopyDocument` isn't needed for that scenario

PR #1629, opened 2017-11-22
Arrow operator fails to find function

### LINKED ISSUE #1622: Arrow operator "function not found" (XPST0017) error
### What is the problem

When (1) an arrow operator is used in function A and (2) the function that the arrow operator calls, function B, is declared after function A in the query prolog, eXist returns a "function not found" error.

### What did you expect

I expected eXist to locate function B when calling function B via the arrow operator. 

My tests below show that Saxon and BaseX do not have this problem.

My tests also show that eXist has no problem locating function B when we either (1) call function B without the arrow operator or (2) move function B above function A in the prolog.

Given the following query, eXist raises a "Function not found: local:B" error:

```xquery
xquery version "3.1";

declare function local:A($x) { 
    let $y := $x || $x 
    return 
        $y => local:B()
};

declare function local:B($x) {
    string-length($x)
};

local:A("hello")
```

The full error:

> Cannot compile xquery: err:XPST0017 Function not found: local:B [at line 6, column 12, source: xquery version "3.1"; declare function local:A($x) { let $y := $x || $x return $y => local:B() }; declare function local:B($x) { string-length($x) }; local:A("hello")]

The query runs without error, as expected, in Saxon and BaseX.

The following variation, which defines `local:B` first, completes without error in eXist:

```xquery
xquery version "3.1";

declare function local:B($x) {
    string-length($x)
};

declare function local:A($x) { 
    let $y := $x || $x 
    return 
        $y => local:B()
};

local:A("hello")
```

Similarly, calling local:B without the arrow operator completes without error in eXist:

```xquery
xquery version "3.1";

declare function local:A($x) { 
    let $y := $x || $x 
    return 
        local:B($y)
};

declare function local:B($x) {
    string-length($x)
};

local:A("hello")
```

### Context information

- eXist-db version + Git Revision hash: eXist-db 3.6.0 /  145d6526e
- Java version: 1.8.0_144
- Operating system: macOS 10.13.1
- 32 or 64 bit: 64 bit

wolfgangmm:
Defer function lookup until query has been parsed entirely. Fixes #1622.

### DIFF
--- a/src/org/exist/xquery/ArrowOperator.java
+++ b/src/org/exist/xquery/ArrowOperator.java
@@ -29,6 +29,7 @@ import org.exist.xquery.value.Type;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Supplier;
 
 /**
  * Implements the XQuery 3.1 arrow operator.
@@ -37,6 +38,7 @@ import java.util.List;
  */
 public class ArrowOperator extends AbstractExpression {
 
+    private QName qname = null;
     private Expression leftExpr;
     private FunctionCall fcall = null;
     private Expression funcSpec = null;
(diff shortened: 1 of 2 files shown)


joewiz:
@wolfgangmm Thank you very much for this fix! I can confirm this fixes the issue. 

@adamretter Thanks for reviewing and merging!

Issue #741, opened 2015-07-16
xs:QName not coerced by value/general comparison


telic:
Comparing an xs:QName to another atomic type results in a type error, even when explicit casts allow the comparison.

For example `'test' eq xs:QName('test')` results in `err:XPTY0004 cannot compare xs:string('test') with xs:QName('test')`, but `'test' eq xs:QName('test') cast as xs:string` results in `true()`.

This affects both value and general comparisons.  General comparisons are a particular problem, since `xs:QName` can not be compared to an empty sequence (which ends up being `xs:anyAtomicType`)!

Furthermore, a different error is reported by eXide depending on which side of the comparison the `xs:QName` appears on. If it is the left-hand operand, `exerr:ERROR" is raised, otherwise "err:XPT0004` is raised.  This has repercussions on try/catch clauses that attempt to mitigate the error.

adamretter:
> For example `'test' eq xs:QName('test')` results in `err:XPTY0004 cannot compare xs:string('test') with xs:QName('test')`, but `'test' eq xs:QName('test') cast as xs:string` results in `true()`.

Is correct behaviour in accordance wit the spec. You will find the same result in Saxon and BaseX also.


> This affects both value and general comparisons. General comparisons are a particular problem, since `xs:QName` can not be compared to an empty sequence (which ends up being `xs:anyAtomicType`)!

I am not sure I agree, the following seems to work fine:
```xquery
() = xs:QName('test')
```

> depending on which side of the comparison the xs:QName appears on. If it is the left-hand operand, exerr:ERROR" is raised, otherwise "err:XPT0004 is raised.

Ah that does seem to be a bug. Both should throw `err:XPT0004`.

PR #1630, opened 2017-11-22
[bugfix] Inability compare a QName should result in XPTY0004

adamretter:
Closes https://github.com/eXist-db/exist/issues/741


adamretter:
@dizzzz Ah right, I missed committing the test file. It is there for us now :-)

Issue #1620, opened 2017-11-19
Backup-restore failure : parsing errors


dizzzz:
### What is the problem
During a simple backup - restore (no database cleanup in between), some files haven't been restored due to errors reported by the XML parser:

```
Exception: ------------------------------------
Exception: ------------------------------------
Problems occured found during restore:

WARN: Failed to restore resource 'repo.xml'
from file '/path/eXist-backup.zip#db/apps/shared-resources/repo.xml'.
Reason: Failed to invoke method parse in class org.exist.xmlrpc.RpcConnection: org.xml.sax.SAXParseException: Element type "repo:meta" must be followed by either attribute specifications, ">" or "/>".

WARN: Failed to restore resource 'build.xml'
from file '/path/eXist-backup.zip#db/apps/webstart/build.xml'.
Reason: Failed to invoke method parse in class org.exist.xmlrpc.RpcConnection: org.xml.sax.SAXParseException: The string "--" is not permitted within comments.

WARN: Failed to restore resource 'repo.xml'
from file '/path/eXist-backup.zip#db/apps/monex/repo.xml'.
Reason: Failed to invoke method parse in class org.exist.xmlrpc.RpcConnection: org.xml.sax.SAXParseException: Element type "permissions" must be followed by either attribute specifications, ">" or "/>".

WARN: Failed to restore resource 'repo.xml'
from file '/path/eXist-backup.zip#db/apps/eXide/repo.xml'.
Reason: Failed to invoke method parse in class org.exist.xmlrpc.RpcConnection: org.xml.sax.SAXParseException: Element type "permissions" must be followed by either attribute specifications, ">" or "/>".
```

There is need a `--` in comments (issue 2), and all of the `repo.xml` files contain a root element with an attribute which name starts with an `#` character:

```xml
<repo:meta xmlns="http://exist-db.org/xquery/repo" 
    xmlns:repo="http://exist-db.org/xquery/repo" 
   #unknown="http://exist-db.org/xquery/repo">
    <repo:description>Shared resources used by other apps</repo:description>
    <repo:author>Wolfgang Meier</repo:author>
    <repo:website>https://github.com/eXist-db/shared-resources</repo:website>
    <repo:status>alpha</repo:status>
    <repo:license>GNU-LGPL</repo:license>
    <repo:copyright>true</repo:copyright>
    <repo:type>library</repo:type>
    <repo:target>shared-resources</repo:target>
    <repo:prepare/>
    <repo:finish/>
[...]

PR #1640, opened 2017-11-29
Fixes for MemTree namespace nodes

adamretter:
Closes https://github.com/eXist-db/exist/issues/1620

I also checked that this PR keeps the same DOM Test Suite score that we have for eXist 3.6.0 - https://github.com/exist-db/exist-domts#exist-db-360

### DIFF
--- a/src/org/exist/dom/memtree/MemTreeBuilder.java
+++ b/src/org/exist/dom/memtree/MemTreeBuilder.java
@@ -116,7 +116,7 @@ public class MemTreeBuilder {
      * @param attributes   DOCUMENT ME!
      * @return the node number of the created element
      */
-    public int startElement(final String namespaceURI, final String localName, final String qname, final Attributes attributes) {
+    public int startElement(final String namespaceURI, String localName, final String qname, final Attributes attributes) {
         final int prefixIdx = qname.indexOf(':');
 
         String prefix = null;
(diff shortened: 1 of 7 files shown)


dizzzz:
@adamretter great work, super fast! Question: does this address the double-dash issue as well?

`xml

`

adamretter:
@dizzzz Ah no. I missed that second issue in amongst the talk of the first one. I just added a follow-up in the issue for you for some more info ;-)

src/org/exist/dom/memtree/MemTreeBuilder.java
```diff
@@ -147,15 +147,15 @@ public int startElement(final QName qname, final Attributes attributes) {
 
             // parse attributes
             for(int i = 0; i < attributes.getLength(); i++) {
```
dizzzz: probably we could use another loop, without the "i" index. A simple for loop over the attributes?

src/org/exist/dom/memtree/SAXAdapter.java
```diff
@@ -125,7 +125,7 @@ public void startElement(final String namespaceURI, final String localName, fina
 
         for(int i = 0; i < atts.getLength(); i++) {
             if(atts.getQName(i).startsWith(XMLConstants.XMLNS_ATTRIBUTE)) {
-                final String prefix = atts.getLocalName(i);
+                final String prefix = null;
                 final String uri = atts.getValue(i);
                 if(namespaces == null || !namespaces.containsKey(prefix)) {
```
dizzzz: doesn't this yield into a NPE in line  130?

src/org/exist/dom/memtree/SAXAdapter.java
```diff
@@ -125,7 +125,7 @@ public void startElement(final String namespaceURI, final String localName, fina
 
         for(int i = 0; i < atts.getLength(); i++) {
```
dizzzz: what about a Java5 loop on Attributes without 'i' variable?

Issue #1642, opened 2017-12-03
NPE in org.exist.xquery.functions.fn.FunId#getId()


dizzzz:
Reported by craigberry  on exist-open

I have a web app that is a variant of TEI Simple and it has a search form that works in 3.4.0 but breaks in 3.5.0 (and is still broken the same way in 3.6.0).  Thanks to the "Copy as curl" feature in my browser, the form submission boils down to:

curl 'http://localhost:8080/exist/apps/shc/works/search.html?query=justice&field=text&lucene-query-mode=any&tei-target=tei-text' \
-XGET 

The result of that request is in the file attached here as search_test_350.html.  It consists mostly of a 2500-line Java stack dump. Changing GET to POST makes no difference as far as the error I get.  The heart of the matter seems to be somewhere around here:

Caused by: javax.servlet.ServletException: An error occurred while processing request to /exist/apps/shc/works/search.html: An unknown error occurred: null
	at org.exist.http.urlrewrite.XQueryURLRewrite.service(XQueryURLRewrite.java:381)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)

and here:

Caused by: javax.servlet.ServletException: An unknown error occurred: null
	at org.exist.http.servlets.EXistServlet.doPost(EXistServlet.java:478)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:707)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)

This problem has in common with GitHub issue #1612 (<https://github.com/eXist-db/exist/issues/1612>) blowing up on the same line in XQueryURLRewrite.java, but there it's an out of range exception rather than null exception as here, and I have no non-ASCII characters in my URL, which seems to be part of the problem there.  I should also note that other parts of the app, including other form submissions, seem to be fine.  A corollary of that is that I don't know how to make a smaller reproducer.

Since I'm getting an exception caused by a null in some code that was explicitly changed to return null in commit a9c76a0fc040ede, I'm going to take a wild guess that the problem is somehow related to that commit:

<https://github.com/eXist-db/exist/commit/a9c76a0fc040ede5a6e0d18b3452a11d918da4c7#diff-88285676e2d4cdf4bde12b4f92e821ca>

[...]

PR #1644, opened 2017-12-03
Fix for #1642 NPE in org.exist.xquery.functions.fn.FunId

dizzzz:
Fix for #1642

### DIFF
--- a/src/org/exist/xquery/functions/fn/FunId.java
+++ b/src/org/exist/xquery/functions/fn/FunId.java
@@ -23,43 +23,19 @@ package org.exist.xquery.functions.fn;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-
-import org.exist.dom.persistent.DefaultDocumentSet;
-import org.exist.dom.persistent.DocumentSet;
-import org.exist.dom.persistent.ExtArrayNodeSet;
-import org.exist.dom.persistent.MutableDocumentSet;
-import org.exist.dom.persistent.NodeProxy;
-import org.exist.dom.persistent.NodeSet;
 import org.exist.dom.QName;
 import org.exist.dom.memtree.DocumentImpl;
 import org.exist.dom.memtree.NodeImpl;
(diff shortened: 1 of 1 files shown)


craigberry:
I did a build with this and it seems to have fixed the form submission where the problem showed up.

src/org/exist/xquery/functions/fn/FunId.java
```diff
@@ -217,13 +193,15 @@ private void getId(NodeSet result, DocumentSet docs, String id) throws XPathExce
 
     private void getId(Sequence result, Sequence seq, String id) throws XPathException {
         final Set<DocumentImpl> visitedDocs = new TreeSet<>();
-        for (final SequenceIterator i = seq.iterate(); i.hasNext();) {
+        for (final SequenceIterator i = seq.iterate(); i.hasNext(); ) {
             final NodeImpl v = (NodeImpl) i.nextItem();
             final DocumentImpl doc = v.getOwnerDocument();
-            if (!visitedDocs.contains(doc)) {
+
+            if (doc != null && !visitedDocs.contains(doc)) {
```
adamretter: So the problem is that `org.exist.dom.memtree.NodeImpl#getOwnerDocument()` returns `null`? If I understand correctly, this will only happen when the `NodeImpl` is a `org.exist.dom.memtree.DocumentImpl`?

If that is the case, then this fix looks fine to me. I would have liked to have seen a test though ;-)

dizzzz: yes, in a commit from your hand, `org.exist.dom.memtree.NodeImpl#getOwnerDocument()` can return null.

dizzzz: I agree with that we need a test......

dizzzz: I studied your commit, where the `org.exist.dom.memtree.NodeImpl#getOwnerDocument()` can actually return null ; Being that a valid value, this loop should be resistent for it.

Issue #1639, opened 2017-11-28
general comparison to empty sequence sometimes raises an error


telic:
### What is the problem

eXist sometimes raises a type error when making a general comparison to an empty-sequence on the right-hand side (eg. `current-date() = ()`). It never raises this error if the empty sequence appears on the left-hand side (eg. `() = current-date()`).

This appears to be due to inconsistent practice in the implementation of the `compareTo(Collator, Comparison, AtomicValue)` method in each of the `org.exist.xquery.value` classes. Some explicitly check for `.isEmpty()` at the start, but `BooleanValue`, `AbstractDateTimeValue`, and `BinaryValue` do not (and there may be others that I missed).

### What did you expect

A general comparison to the empty sequence should always return `false()`, since no "[pair of atomic values](https://www.w3.org/TR/xquery-31/#id-general-comparisons)" exists.

### Context information
- eXist-db 3.6.0 (build 201711271542) / ceaab01
- Java 1.8.0_152-b16
- MacOS 10.13.1
- 64 bit

dizzzz:
<img width="835" alt="isempty" src="https://user-images.githubusercontent.com/1700062/33529936-cb2cc086-d877-11e7-81c5-ddddb413b2f5.png">

dizzzz:
I think we can fix this in the GeneralCOmparison : 

org/exist/xquery/GeneralComparison.java:1094

```java
 default: {
                    // If right value is () always return false
                    return( rv.isEmpty() ? false : lv.compareTo( collator, relation, rv ) );
                }
```

PR #1647, opened 2017-12-04
Bugfix/1639 fix empty sequence comparison

dizzzz:
Fix for #1639 : In general comparison when the right value is the empty sequence, always false() should be returned. (checked with Saxon)

### DIFF
--- a/test/src/xquery/namespaces.xql
+++ b/test/src/xquery/namespaces.xql
@@ -272,3 +272,27 @@ declare
 function nt:qname-lhs-compare() {
     xs:QName('test') eq 'test'
 };
+
+declare
+    %test:assertError("XPTY0004")
+function nt:int-string-compare() {
+    5 eq 'five'
+};
+
+declare
+    %test:assertError("XPTY0004")
+function nt:string-int-compare() {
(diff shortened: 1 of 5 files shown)


test/src/xquery/namespaces.xql
```diff
@@ -272,3 +272,15 @@ declare
 function nt:qname-lhs-compare() {
     xs:QName('test') eq 'test'
 };
+
```
adamretter: Erm I think these tests might be in the wrong file. They don't seem to be namespace related tests ;-)

dizzzz: I actually put them in this file because.... you put in a similar test here recently...

adamretter: Oh! Doesn't mean I got it right though ;-)

adamretter: Ah okay, so my tests were around `QNames`. Maybe could have gone somewhere else, they have a tenuous relationship to namespaces ;-)

dizzzz: I saw the compare thingy.... :-)

PR #1660, opened 2017-12-13
Enhancements for the Cache Module

adamretter:
Happy Christmas @joewiz 

1. Switches the underlying mechanism to Caffeine.
2. Adds LRU support via `maimumSize` and/or `expireAfterAccess` when creating a Cache with `cache:create($name, $config)`.
3. Adds some additional functions:
    1. `cache:names()`: returns the names of all active caches.
    2. `cache:list-keys($name)`: Lists the active keys of a named cache.
    3. `cache:destroy($name)`: Clears and removes the underlying object for a named cache.


joewiz:
I'm dumbfounded! Thank you, @adamretter!

This passes all tests except codacy - so it's ready for testing, right?

adamretter:
Yup

joewiz:
@adamretter I've been taking this for a spin, and really like the new functions and features that you've added - which make it much easier to probe caches for their keys and values.  A few questions:

1. What is the difference between your new `cache:destroy` and the original `cache:clear` function? Is it that the original one still leaves the cache object, just cleared of values, while your new function destroys the object altogether?
2. I notice that `cache:list` and your new `cache:list-keys` are both marked as "Deprecated as the operation is expensive. This function could be removed in the next major release version." Did you mean just to mark the former as such? 
3. Had you considered making `cache:list-keys` just the single-arity version of `cache:list`? So `cache:list#1` would list keys and `cache:list#2` would list values for a given key? Or perhaps, inspired by the convention from XQuery 3.0 for maps, it could be named just `cache:keys`. Between these two options, I have no strong opinion but favor the latter.
4. The function documentation for the `maximumSize` configuration option for `cache:create` describes this as "the maximum number of entries". "Entry" is not defined here though. Does entry refer to a cache key, or the values assigned to a key? In other words, if you create a cache whose maximumSize is 10, what happens when you create an 11th key, or when you put a 11th value into the 1st key?
5. The function documentation for `cache:create`'s 2nd parameter, `$options`, should probably express the sample entry as an XQuery map, with map constructors and quoted entry names, i.e., `map { "maximumSize": 1000, "expireAfterAccess": 120000, "permissions": map { "put-group": "group1", "get-group": "group2", "remove-group": "group3", "clear-group": "group4"} }`
6. The function documentation for `cache:put` indicates that values bound to a key are constrained to be `xs:anyType`. In my application's use of the original cache module, I've been putting XQuery maps into cache values (and maps [are of type `item()`](https://www.w3.org/TR/xpath-datamodel-31/#types-hierarchy), not `xs:anyAtomicType`). Your branch doesn't seem to reject maps as values for cache entries. Do you know if there is a reason why values are described as constrained to being xs:anyType? Or could the function signature be changed to `$value as item()*`?

Lastly, do you see any reason why we should keep this module disabled by default? Or would this newly reviewed and improved code be a good time to promote this module to being enabled by default?

adamretter:
@joewiz I believe I have made those changes you requested now.

joewiz:
@adamretter Thank you! I've now confirmed your answers by writing some tests. Would these be useful to include in the PR? 

- [cache.xql.zip](https://github.com/eXist-db/exist/files/1574666/cache.xql.zip)

Here are the results as I see them in eXide:

```xml
<testsuites>
    <testsuite package="http://exist-db.org/xquery/cache/test" timestamp="2017-12-20T02:05:46.473-06:00" failures="2" pending="0" tests="12" time="PT4.026S">
        <testcase name="create-already-exists" class="c:create-already-exists"/>
        <testcase name="create-maximumSize-success" class="c:create-maximumSize-success"/>
        <testcase name="create-simple-success" class="c:create-simple-success"/>
        <testcase name="exercise-expireAfterAccess" class="c:exercise-expireAfterAccess">
            <failure message="assertEquals failed." type="failure-error-code-1">0</failure>
            <output>1</output>
        </testcase>
        <testcase name="exercise-maximumSize" class="c:exercise-maximumSize">
            <failure message="assertEquals failed." type="failure-error-code-1">5</failure>
            <output>1</output>
        </testcase>
        <testcase name="get" class="c:get"/>
        <testcase name="keys" class="c:keys"/>
        <testcase name="list-all-values" class="c:list-all-values"/>
        <testcase name="list-one-value" class="c:list-one-value"/>
        <testcase name="names" class="c:names"/>
        <testcase name="remove" class="c:remove"/>
        <testcase name="remove-returns-previous-value" class="c:remove-returns-previous-value"/>
    </testsuite>
</testsuites>
```

I'm not quite sure why the two tests involving maximumSize and expireAfterAccess are failing; I'm giving them plenty of time to complete via `util:wait`. Could it be that the tests of these two features are best handled at the Java level, rather than the XQuery level?

I can think of some more tests; e.g., putting values of various types and confirming that the types are unchanged when getting values...  I'd consider adding others that folks are able to think of.  Adding tests for the permission options would be somewhat more involved - perhaps just a test to ensure guest can't access certain caches.

adamretter:
@joewiz - https://github.com/ben-manes/caffeine/issues/214

joewiz:
@adamretter Thanks for passing that question onto the creator of Caffeine.  If it turns out that the ability to accurately list or count keys isn't supported by Caffeine due to some underlying locking/performance design considerations, we can drop these tests.  One idea, though, based on https://github.com/ben-manes/caffeine/wiki/Cleanup: if we added a `cache:cleanup` function, this could force a sync and allow the tests to run.  

I don't know how you made that test suite commit in my name, but that was a cool trick! :)

adamretter:
@joewiz Incidentally your code had a bug which eXist failed to detect, which lead to intermittent failures when running. Basically you have two functions called `c:keys`. This should have led to `XQST0034`, however eXist seems to ignore the issue, and then at runtime choose (sometimes incorrectly) which `c:keys` was intended.

I have opened a new issue for that - https://github.com/eXist-db/exist/issues/1665

adamretter:
@joewiz Okay, I adjusted the function naming in your tests a little and implemented `cache:cleanup`. All tests are now green :-)

joewiz:
@dizzzz Thanks for the review & merge!

@adamretter Looks great! **Thank you** for these great additions!

Issue #1663, opened 2017-12-20
Update insert fails


dizzzz:
### What is the problem

Original report by Alexander Henket:

I get a `NullPointerException` when I try to “update insert” the first node into an element that has attributes.

```
org.xmldb.api.base.XMLDBException
	at org.exist.xmldb.LocalXPathQueryService.execute(LocalXPathQueryService.java:195)
	at org.exist.xmldb.LocalXPathQueryService.lambda$execute$1(LocalXPathQueryService.java:159)
	at org.exist.xmldb.function.LocalXmldbFunction.apply(LocalXmldbFunction.java:46)
	at org.exist.xmldb.AbstractLocal.withDb(AbstractLocal.java:196)
	at org.exist.xmldb.LocalXPathQueryService.execute(LocalXPathQueryService.java:158)
	at org.exist.client.QueryDialog$QueryThread.run(QueryDialog.java:573)
Caused by: java.lang.NullPointerException
	at org.exist.dom.persistent.ElementImpl.appendChildren(ElementImpl.java:587)
	at org.exist.xquery.update.Insert.eval(Insert.java:150)
	at org.exist.xquery.LetExpr.eval(LetExpr.java:99)
	at org.exist.xquery.LetExpr.eval(LetExpr.java:111)
	at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71)
	at org.exist.xquery.PathExpr.eval(PathExpr.java:276)
	at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71)
	at org.exist.xquery.XQuery.execute(XQuery.java:261)
	at org.exist.xquery.XQuery.execute(XQuery.java:185)
	at org.exist.xmldb.LocalXPathQueryService.execute(LocalXPathQueryService.java:192)
	... 5 more
Caused by: java.lang.NullPointerException
	at org.exist.dom.persistent.ElementImpl.appendChildren(ElementImpl.java:587)
	at org.exist.xquery.update.Insert.eval(Insert.java:150)
	at org.exist.xquery.LetExpr.eval(LetExpr.java:99)
	at org.exist.xquery.LetExpr.eval(LetExpr.java:111)
	at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71)
	at org.exist.xquery.PathExpr.eval(PathExpr.java:276)
	at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71)
[...]

PR #1664, opened 2017-12-20
Fix for "Update insert fails"

dizzzz:
See #1663

@adamretter You added the (now removed) lines some time ago....

- [x] add tests
- [x] address concerns


dizzzz:
Added testcases

dizzzz:
OK, thnx for the feedback..... I'll study....

dizzzz:
Hmmmm I think I do not understand the class then. As it is highly undocumented, I need help here from you or @wolfgangmm ....

dizzzz:
@adamretter I looked a long time at the code, the current "hasChildNodes()" looks to be OK (I annotated the meaning of the "children" variable).

I think the current code state should be ok, tests run all fine (in a previous PR-revision some XSLT tests failed)

dizzzz:
Action: run domts (need help from @adamretter  here)

adamretter:
Okay so this little program shows how Xerces interprets the XML sample you provided:

```xml
package org.exist.dom.persistent;

import org.w3c.dom.Document;
import org.w3c.dom.Element;
import org.xml.sax.SAXException;

import javax.xml.parsers.DocumentBuilder;
import javax.xml.parsers.DocumentBuilderFactory;
import javax.xml.parsers.ParserConfigurationException;
import java.io.*;

import static java.nio.charset.StandardCharsets.UTF_8;

public class XercesElementTest {

    public static void main(final String args[]) throws ParserConfigurationException, IOException, SAXException {
        final String xmlStr = "<root attr=\"1\"/>";

        final DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
        final DocumentBuilder builder = factory.newDocumentBuilder();
        try(final InputStream is = new ByteArrayInputStream(xmlStr.getBytes(UTF_8))) {
            final Document doc = builder.parse(is);

            final Element rootElem = doc.getDocumentElement();

            System.out.println("hasChildNodes=" + rootElem.hasChildNodes());
            System.out.println("getLastChild=" + rootElem.getLastChild());
        }
    }
}
```

It returns the output:
```
hasChildNodes=false
getLastChild=null
```

I can confirm that eXist-db's persistent Element implementation also returns the same results. So the problem seems to be that our `org.exist.dom.persistent#appendChildren(Txn, NodeList, int)` is no longer correct, since my DOM improvements.

dizzzz:
thnx for verifying my changes.... so in essence you say my proposed change is not ok? too bad :-/

adamretter:
@dizzzz Sadly the fix to `getLastChild` is not sufficient. I have a fix here, will incorprorate into your PR and your nice tests

PR #1667, opened 2017-12-28
Java9 requires "Permissions" to be set

dizzzz:
Java9 requires "Permissions" to be set; when left out the application will not start.

Partial copy from https://git…hub.com/Trundler/existdb-webstart/blob/master/build.xml


adamretter:
@dizzzz I have no idea if this is correct or not. What was the problem this PR tries to fix ? i.e. how do I reproduce the issue?

dizzzz:
valid question; I was not able to get the webstart client started in java9 ; according to the spec it is required, in my side project it really helped to get a clean webstart client start ;

i'll continue..

dizzzz:
I need to re-check

dizzzz:
it looks ok without the PR, will close it.

PR #1668, opened 2018-01-03
keep original source value

shabanovd:
keep original source information to better resolve relative urls


dizzzz:
Hi, is this part of a larger set of work? the source value is nowhere used; as the data is stored in the class, why did you choose not to use a getter?

shabanovd:
I use it at my own resolver and no needs for getter because it's `final`

dizzzz:
I don't see a good reason not to use a getter() here...

PR #1679, opened 2018-01-09
reload index controller when index configuration changed

shabanovd:
Use `System.currentTimeMillis()` to detect that  `IndexManager` configuration change and reinit `IndexController` at broker

(port 3ab8ac96dc444477839b17d023e728f47d9cf675)

### DIFF
--- a/src/org/exist/storage/BrokerPool.java
+++ b/src/org/exist/storage/BrokerPool.java
@@ -1042,6 +1042,11 @@ public class BrokerPool extends BrokerPools implements BrokerPoolConstants, Data
         return globalXUpdateLock;
     }
 
+    long config_ts = 0;
+    public void configurationChanged() {
+        config_ts = System.currentTimeMillis();
+    }
+
     /**
      * Creates an inactive broker for the database instance.
      *
@@ -1187,6 +1192,12 @@ public class BrokerPool extends BrokerPools implements BrokerPoolConstants, Data
                     }
             }
             broker = inactiveBrokers.pop();
+
+            if (broker.config_ts != config_ts) {
+                broker.config_ts = config_ts;
+                broker.initIndexModules();
+            }
+
             //activate the broker
             activeBrokers.put(Thread.currentThread(), broker);
 
--- a/src/org/exist/storage/BrokerPool.java
+++ b/src/org/exist/storage/BrokerPool.java
@@ -52,7 +52,6 @@ import org.exist.scheduler.impl.SystemTaskJobImpl;
 import org.exist.security.*;
 import org.exist.security.SecurityManager;
 import org.exist.security.internal.SecurityManagerImpl;
-import org.exist.storage.btree.DBException;
 import org.exist.storage.journal.JournalManager;
 import org.exist.storage.lock.DeadlockDetection;
 import org.exist.storage.lock.FileLockService;
@@ -1042,11 +1041,6 @@ public class BrokerPool extends BrokerPools implements BrokerPoolConstants, Data
         return globalXUpdateLock;
     }
 
-    long config_ts = 0;
-    public void configurationChanged() {
-        config_ts = System.currentTimeMillis();
-    }
-
     /**
      * Creates an inactive broker for the database instance.
      *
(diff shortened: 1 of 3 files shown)


dizzzz:
Please write "configTimestamp" in stead of "config_ts" for readability.

src/org/exist/storage/DBBroker.java
```diff
@@ -116,6 +116,8 @@
 
     protected IndexController indexController;
 
+    long config_ts = 0;
```
dizzzz: I'd like to discuss, but IMO these values need to be read/set with getters and setters.....

src/org/exist/storage/BrokerPool.java
```diff
@@ -1042,6 +1042,11 @@ public Lock getGlobalUpdateLock() {
         return globalXUpdateLock;
     }
 
+    long config_ts = 0;
```
dizzzz: please use longer names

dizzzz:
Could you think of a test? WHat is the problem that is actually solved with the PR?

src/org/exist/storage/DBBroker.java
adamretter: I would also like to see getters where the class is not an inner class; Otherwise I feel that it breaks encapsulation.

adamretter:
I am not really understanding the purpose of this change? Is it for dynamic loading of indexers, or for index configuration? if the the latter, should it not be per-Collection?

shabanovd:
@adamretter where do you see it per-Collection?

src/org/exist/storage/DBBroker.java
shabanovd: why do it needs getter or setter if that `protected` and for in-package use only?

shabanovd:
@dizzzz `ts` is known abbreviation for timestamp, no?

dizzzz:
@shabanovd  I have no idea what `ts` would be, it is not really standard I guess. As the code will be read by more people, use of abbreviations  should be limited. Probably only `nr` or `no` (number / of)

src/org/exist/storage/DBBroker.java
dizzzz: I guess there a quite reasons I could think of. Good API design is one of it. There is not really a good reason not to use getters/setters IMO.  These protected members are not accessed directly I assume?

shabanovd:
@dizzzz I refactor PR ... but you ask for setter mean that you don't understand the fix -(

dizzzz:
@shabanovd be please polite to me;  for sure you are right about the setter, but the remark comes from seeing two PRs without getters/setters this week.

shabanovd:
@dizzzz 2nd one don't need setter too

joewiz:
@shabanovd Please explain the purpose of this change?

shabanovd:
@joewiz index chain do not reload (invalidate) after changing index configuration

dizzzz:
I cannot judge the PR, sorry. @wolfgangmm  ?

joewiz:
@shabanovd Please provide a fuller explanation of the purpose of the PR besides a restatement of the information in the original PR.  Please help reviewers by better explaining the purpose of a PR from the start, particularly if you ignore [the default PR template](https://github.com/eXist-db/exist/blob/develop/.github/PULL_REQUEST_TEMPLATE.md). I shouldn't need to state this, but if reviewers ask you questions, please answer them all and in a direct, collegial fashion. We are still waiting for an answer to @adamretter's question above:

> Is [this PR] for dynamic loading of indexers, or for index configuration?

I look forward to a full, direct, and collegial response to these questions and any others that come up in the course of reviewing this PR. :-)

shabanovd:
@joewiz the purpose of this PR as simple as it was explained and I can't explain it anyhow differently. Also, @adamretter didn't answer my question, so I can't help here too.

src/org/exist/storage/DBBroker.java
```diff
@@ -116,6 +116,8 @@
 
     protected IndexController indexController;
 
+    long config_timestamp = 0;
```
wolfgangmm: config_timestamp is never updated?

joewiz:
After consulting with @wolfgangmm and @dizzzz we agree that this PR is not acceptable. Without a test we cannot be sure the PR indeed achieves it and does not destroy anything else.

@shabanovd Please consider submitting this work in a new PR that contains a test that explains the problem and its solution. On a more general point, your contributions are always welcome, as are anyone's, but when submitting PRs you must answer all questions from reviewers in a direct and collegial manner; ignoring them or refusing to answer them is outside the bounds of acceptable community conduct.

src/org/exist/storage/DBBroker.java
shabanovd: yeah, that is typo that pop up on  refactoring. Fixed.

shabanovd:
@joewiz by closing this PR you're saying that you don't want it

PR #1680, opened 2018-01-09
Update map -> map(*) and empty() -> empty-sequence() according to spec.

dizzzz:
### Description:

Update `map -> map(*)` and `empty() -> empty-sequence()` according to spec.
and: `element(*)`, `function(*)` and `array(*)`

### Reference:

https://www.w3.org/TR/xquery-31/#doc-xquery31-SequenceType
https://www.w3.org/TR/xquery-31/#doc-xquery31-MapTest

### Type of tests:

- existing tets suite --> OK
- pending: additional test to show progression

@wolfgangmm is this a correct approach?

### DIFF
--- a/src/org/exist/xquery/value/Type.java
+++ b/src/org/exist/xquery/value/Type.java
@@ -191,7 +191,7 @@ public class Type {
         defineBuiltInType(ITEM, "item()");
         defineBuiltInType(EMPTY, "empty-sequence()","empty()"); // keep empty() for backward compatibility
 
-        defineBuiltInType(ELEMENT, "element(*)","element()");
+        defineBuiltInType(ELEMENT, "element()");
         defineBuiltInType(DOCUMENT, "document-node()");
         defineBuiltInType(ATTRIBUTE, "attribute()");
         defineBuiltInType(TEXT, "text()");
--- a/src/org/exist/xquery/value/Type.java
+++ b/src/org/exist/xquery/value/Type.java
@@ -189,7 +189,7 @@ public class Type {
         //TODO use parentheses after the nodes name  ?
         defineBuiltInType(NODE, "node()");
         defineBuiltInType(ITEM, "item()");
-        defineBuiltInType(EMPTY, "empty()");
+        defineBuiltInType(EMPTY, "empty-sequence()","empty");
 
         defineBuiltInType(ELEMENT, "element()");
         defineBuiltInType(DOCUMENT, "document-node()");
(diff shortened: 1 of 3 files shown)


src/org/exist/xquery/value/Type.java
```diff
@@ -189,7 +189,7 @@
         //TODO use parentheses after the nodes name  ?
         defineBuiltInType(NODE, "node()");
         defineBuiltInType(ITEM, "item()");
-        defineBuiltInType(EMPTY, "empty()");
+        defineBuiltInType(EMPTY, "empty-sequence()","empty");
```
shabanovd: it must be `empty()`

dizzzz:
Nono according to the spec, as i wrote

shabanovd:
@dizzzz please, stop a bit and think before reply ... you're giving alternative name `empty` but it must be `empty()`

src/org/exist/xquery/value/Type.java
dizzzz: My mobile app did not su the correct comment' you are right, I'll update.
Te question remains... is this the correct change? is it an external interface change which means we need to hold it till eXist-db 4.0.0 ?

shabanovd: no needs to wait till 4.x as it support both, remove `empty()` or `map` will require 4.x

dizzzz:
@adamretter I was able to think of some tests..... things are not too difficult sometimes.
I found more examples where exist-db did not follow the spec. backward compliance is present.

joewiz:
@dizzzz Regarding your comment:

> I found more examples where exist-db did not follow the spec. backward compliance is present.

Can you think of any way to document which syntax is being supported for backward compatibility? This would be very helpful to be able to consult. Also, ideally, we should mark or declare such syntax that isn't compliant with the spec as deprecated - and then drop support for it in an upcoming major version of eXist.

dizzzz:
Marking stuff deprecated is not possible in the code; indeed I forgot about the array test :)

All comes from the 3.1 specs, I searched literally for (*) in there. Also element is marked this way.

I’ll updated the pr on your request ....

dizzzz:
In the end, this code looks only to be used for the xqdoc code; I actually found the issues with the fundocs app (Mongrel development)

adamretter:
@dizzzz @joewiz I am not sure about `element(*)`. I see both `element()` and `element(*)` being used throughout the spec. Unless anyone can suggest a good reason otherwise, I think we should stay with `element()`

dizzzz:
I reverted it to element() without the * in the last commit

dizzzz:
The title here https://github.com/eXist-db/exist/pull/1680/commits/3a44cc8de2b38d01c18efb8bb79eed91fa462d70 is wrong, sorry

joewiz:
Indeed, while this PR contains tests showing the correct syntax works, it didn't include tests demonstrating backward compatibility is maintained.

PR #1684, opened 2018-01-11
Small fixes

adamretter:


src/org/exist/collections/CollectionConfigurationManager.java
```diff
@@ -117,7 +117,7 @@ public void addConfiguration(Txn txn, final DBBroker broker, Collection collecti
 
             XmldbURI configurationDocumentName = null;
             // Replaces the current configuration file if there is one
-            final CollectionConfiguration conf = getConfiguration(broker, collection);
+            final CollectionConfiguration conf = getConfiguration(broker, confCol);
```
dizzzz: without this change... what went wrong?

adamretter: hmm good question. I think that was a mistake. I will correct

test/src/org/exist/TestUtils.java
```diff
@@ -59,28 +60,58 @@ public static void cleanupDB() {
...
-                broker.saveCollection(transaction, root);
-                for (Iterator<XmldbURI> i = root.collectionIterator(broker); i.hasNext(); ) {
-                    XmldbURI childName = i.next();
-                    if (childName.equals("system"))
-                        continue;
-                    Collection childColl = broker.getOrCreateCollection(transaction, XmldbURI.ROOT_COLLECTION_URI.append(childName));
-                    assertNotNull(childColl);
-                    broker.removeCollection(transaction, childColl);
+                Collection root = null;
+                try {
```
dizzzz: suggestion: 

```java
try(Txn transaction = pool.getTransactionManager().beginTransaction())){
```

adamretter: It is already in the `try-with-resources`:

```java
 try(final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()));
                final Txn transaction = pool.getTransactionManager().beginTransaction()) {
```

Note the semicolon after the broker acquisition!

PR #1693, opened 2018-01-19
Fixes some synchronization and reuse issues

adamretter:
1. Fixes a bug in the EXPath HTTP Client with connection pooling
2. Fixes a problem with bad thread synchronization in JMX
3. Explicitly names each thread that eXist creates for easier identification.


adamretter:
@wsalesky This PR includes a fix for your problems with the EXPath HTTP Client

Issue #1646, opened 2017-12-04
RESTServer returns 202 when a bad XQuery is posted


adamretter:
If you HTTP POST an invalid XQuery to the REST Server, it will return HTTP Status Code 202 (Accepted), whilst returning a HTTP Response Body describing the query error. This is obviously not correct.

It should most likely return HTTP 400, or maybe even 422.

PR #1696, opened 2018-01-19
Invalid query should result in HTTP 400

adamretter:
Closes https://github.com/eXist-db/exist/issues/1646

### DIFF
--- a/src/org/exist/http/RESTServer.java
+++ b/src/org/exist/http/RESTServer.java
@@ -851,10 +851,10 @@ public class RESTServer {
 
                         } catch (final XPathException e) {
                             if (MimeType.XML_TYPE.getName().equals(mimeType)) {
-                                writeXPathException(response, HttpServletResponse.SC_ACCEPTED,
+                                writeXPathException(response, HttpServletResponse.SC_BAD_REQUEST,
                                         encoding, null, path, e);
                             } else {
-                                writeXPathExceptionHtml(response, HttpServletResponse.SC_ACCEPTED,
+                                writeXPathExceptionHtml(response, HttpServletResponse.SC_BAD_REQUEST,
                                         encoding, null, path, e);
                             }
                         }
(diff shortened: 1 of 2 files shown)


test/src/org/exist/http/RESTServiceTest.java
```diff
@@ -33,6 +33,7 @@
 import java.net.URLEncoder;
 import javax.xml.parsers.ParserConfigurationException;
 
+import org.eclipse.jetty.http.HttpStatus;
```
dizzzz: better to use https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponse.html#SC_NOT_FOUND ?

adamretter: I don't see our dependency on Jetty going away, but I am happy to revise it. Do you wish me to do so?

src/org/exist/http/RESTServer.java
```diff
@@ -1854,26 +1854,24 @@ private void writeXPathException(final HttpServletResponse response,
...
+            writer.write("<message>");
+            final String message = e.getMessage() == null ? e.toString() : e.getMessage();
+            writer.write(XMLUtil.encodeAttrMarkup(message));
+            writer.write("</message>");
+            if (query != null) {
+                writer.write("<query>");
+                writer.write(XMLUtil.encodeAttrMarkup(query));
+                writer.write("</query>");
+            }
+            writer.write("</exception>");
```
dizzzz: Don't we need a flush() call here to be sure all data is sent to the http agent?

dizzzz: answering myself: no not according to the servlet spec :-)

adamretter:
@dizzzz I changed this back to 4.0.0 milestone because it is an API change.

dizzzz:
ah ok, sorry.

PR #1712, opened 2018-02-04
Ensure math functions report their ns prefix

joewiz:
### Description:

This PR fixes a problem with functions in the math module: they did not identify their namespace prefix to functions like `inspect:inspect-function`. 

For example, `inspect:inspect-function(math:pi#0)` would return `@name="pi"` instead of `@name="math:pi"`. In contrast, all modules (other than the default function namespace) module would return, e.g., `@name="array:size"`, `"map:merge"`, etc.

With this PR, math module functions now correctly report their namespace prefix to `inspect:inspect-function`, returning `@name="math:pi"`.

### Reference:

- https://www.w3.org/TR/xpath-functions-31/#namespace-prefixes
- https://www.w3.org/TR/xpath-functions-31/#trigonometry

### Type of tests:

- Well, the math module currently has none of its own tests whatsoever besides the QT3 tests which probably wouldn't test introspectability. I suppose some new test suite could be created, ensuring that `inspect:inspect-function` would include the prefix for functions in the `math`, `map`, and `array` modules. I'm open to suggestions here. I only know about how these functions should report themselves from XQuery and don't know how they should from a Java perspective.

### DIFF
--- a/src/org/exist/xquery/functions/math/NoParamFunctions.java
+++ b/src/org/exist/xquery/functions/math/NoParamFunctions.java
@@ -46,7 +46,7 @@ public class NoParamFunctions extends BasicFunction {
     public static final String PI = "pi";
 
     public final static FunctionSignature FNS_PI = new FunctionSignature(
-        new QName(PI, MathModule.NAMESPACE_URI),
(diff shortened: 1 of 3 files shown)


dizzzz:
Thnx !

joewiz:
@dizzzz Thanks!

PR #1718, opened 2018-02-05
Implement fn:unparsed-text and related functions

wolfgangmm:
Implements `fn:unparsed-text`, `fn:unparsed-text-lines` and `fn:unparsed-text-available` for resources stored in the db as well as retrieved from external URIs.

### DIFF
--- a/src/org/exist/source/Source.java
+++ b/src/org/exist/source/Source.java
@@ -25,6 +25,7 @@ package org.exist.source;
 import java.io.IOException;
 import java.io.Reader;
 import java.io.InputStream;
+import java.nio.charset.Charset;
 
 import org.exist.dom.QName;
 import org.exist.security.PermissionDeniedException;
@@ -96,7 +97,15 @@ public interface Source {
     InputStream getInputStream() throws IOException;
 
     String getContent() throws IOException;
-    
+
+    /**
+     * Returns the character encoding of the underlying source or null if unknown.
+     *
+     * @return the character encoding
+     * @throws IOException
+     */
+    Charset getEncoding() throws IOException;
+
     /**
      * Set a timestamp for this source. This is used
      * by {@link org.exist.storage.XQueryPool} to
--- a/src/org/exist/source/Source.java
+++ b/src/org/exist/source/Source.java
@@ -32,6 +32,8 @@ import org.exist.security.PermissionDeniedException;
 import org.exist.security.Subject;
 import org.exist.storage.DBBroker;
 
+import javax.annotation.Nullable;
+
 
 /**
  * A general interface for access to external or internal sources.
@@ -104,7 +106,7 @@ public interface Source {
      * @return the character encoding
      * @throws IOException
      */
-    Charset getEncoding() throws IOException;
+    @Nullable Charset getEncoding() throws IOException;
 
     /**
      * Set a timestamp for this source. This is used
--- a/src/org/exist/source/URLSource.java
+++ b/src/org/exist/source/URLSource.java
@@ -27,6 +27,11 @@ import java.net.ConnectException;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.HttpURLConnection;
+import java.nio.charset.Charset;
+import java.nio.charset.IllegalCharsetNameException;
+import java.nio.charset.UnsupportedCharsetException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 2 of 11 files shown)


src/org/exist/xquery/functions/fn/FunUnparsedText.java
```diff
@@ -0,0 +1,149 @@
...
+import java.net.URISyntaxException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.regex.Pattern;
+
+import static org.exist.xquery.FunctionDSL.*;
+
+public class FunUnparsedText extends BasicFunction {
+
+    private final static String LINE_REGEX = "\\r\\n|\\r|\\n";
```
adamretter: I would suggest to change this to:
```java
private static final Pattern LINE_REGEX = Pattern.compile("\\r\\n|\\r|\\n");
```

Using the `PatternFactory` cache here doesn't make much sense as the regex is static.

src/org/exist/source/URLSource.java
```diff
@@ -120,6 +126,26 @@ public Validity isValid(final Source other) {
 		return Validity.INVALID;
 	}
 
+	public Charset getEncoding() throws IOException {
+		if (connection == null) {
+			connection = url.openConnection();
+		}
+		final String contentType = connection.getContentType();
+		if (contentType != null) {
+			final Pattern pattern = PatternFactory.getInstance().getPattern("(?i)\\bcharset=\\s*\"?([^\\s;\"]*)");
```
adamretter: Using the PatternFactory cache here doesn't make much sense as the regex is static. Instead maybe just compile as a `private static final` member variable.

src/org/exist/source/URLSource.java
```diff
@@ -120,6 +126,26 @@ public Validity isValid(final Source other) {
...
+			final Matcher matcher = pattern.matcher(contentType);
+			if (matcher.find()) {
+				try {
+					return Charset.forName(matcher.group(1).trim().toUpperCase());
+				} catch (IllegalArgumentException e) {
+					// unknown or illegal charset
+					return null;
```
adamretter: The HTTP 1.1 spec specifies that if the `charset` parameter is not present and the media type is of a `text` subtype, then the client should assume `ISO-8859-1`, see: https://tools.ietf.org/html/rfc2616#section-3.7.1

src/org/exist/source/Source.java
```diff
@@ -96,7 +97,15 @@
...
     String getContent() throws IOException;
-    
+
+    /**
+     * Returns the character encoding of the underlying source or null if unknown.
+     *
+     * @return the character encoding
+     * @throws IOException
+     */
+    Charset getEncoding() throws IOException;
```
adamretter: Consider `@Nullable` or `Optional`

src/org/exist/xquery/functions/fn/FunUnparsedText.java
```diff
@@ -0,0 +1,149 @@
...
+    private String readContent(String uriParam, String encoding) throws XPathException {
+        try {
+            URI uri = new URI(uriParam);
+            if (uri.getScheme() == null) {
+                uri = new URI(XmldbURI.EMBEDDED_SERVER_URI_PREFIX + uriParam);
+            }
+            if (uri.getFragment() != null) {
+                throw new XPathException(this, ErrorCodes.FOUT1170, "href argument may not contain fragment identifier");
+            }
+            final StringWriter output = new StringWriter();
```
adamretter: Consider `try-with-resources`

src/org/exist/xquery/functions/fn/FunUnparsedText.java
```diff
@@ -0,0 +1,149 @@
...
+                if (isCalledAs("unparsed-text-available")) {
+                    return BooleanValue.FALSE;
+                }
+                throw e;
+            }
+            if (isCalledAs("unparsed-text-available")) {
+                return BooleanValue.TRUE;
+            }
+            if (isCalledAs("unparsed-text-lines")) {
+                final Pattern pattern = PatternFactory.getInstance().getPattern(LINE_REGEX);
```
adamretter: For large text files, it would be much more efficient to use `LineNumberReader` on the `InputStream` than to read the file into a string, split it into many strings (two copies of the data), and then return a `StringValue`.

src/org/exist/source/Source.java
dizzzz: I saw this one too, forgot to mention

wolfgangmm:
Ok, I applied the changes suggested by reviewers. In addition I disabled file access for non-dba users.

Issue #1721, opened 2018-02-06
Version mismatch jetty JARS monex vs v3.6.1 distribution


dizzzz:
The version of a number of jetty JARs do not match ; I'd recommend:

- to remove these JARS from the XAR into the default distribution 

![monex-0_9_11_xar](https://user-images.githubusercontent.com/1700062/35882331-181cecce-0b84-11e8-9566-3abdb85f33fc.jpg)

vs

```
ll tools/jetty/lib/ | grep jetty
-rw-r--r--  1 wessels  staff   78584 Jan  7 13:16 jetty-annotations-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff   52049 Jan  7 13:16 jetty-deploy-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff  164892 Jan  7 13:16 jetty-http-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff  128560 Jan  7 13:16 jetty-io-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff   39901 Jan  7 13:16 jetty-jaas-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff   25557 Jan  7 13:16 jetty-jmx-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff   41333 Jan  7 13:16 jetty-jndi-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff   54676 Jan  7 13:16 jetty-plus-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff   93173 Jan  7 13:16 jetty-security-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff  585344 Jan  7 13:16 jetty-server-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff  110246 Jan  7 13:16 jetty-servlet-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff  481910 Jan  7 13:16 jetty-util-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff  128578 Jan  7 13:16 jetty-webapp-9.4.8.v20171121.jar
-rw-r--r--  1 wessels  staff   52107 Jan  7 13:16 jetty-xml-9.4.8.v20171121.jar
```

adamretter:
Sounds like a very sensible idea.

PR #1722, opened 2018-02-06
Move monex' jar files into main repository.

dizzzz:
fix for #1721 ; need to update Monex too ! see https://github.com/eXist-db/monex/pull/50


adamretter:
Do you also need to add config files to `$EXIST_HOME/tools/jetty/etc/` and update `tools/jetty/etc/standard.enabled-jetty-configs`?

dizzzz:
not sure.... ? I could use some help here probably.

dizzzz:
answering: you are probably right, but as the original JAR files are loaded via the XAR... not sure how/if any changes are needed?

adamretter:
@dizzzz I think this should be okay as is

PR #1726, opened 2018-02-11
Add new user manager to distribution

wolfgangmm:
Version 0.4.9 of the dashboard drops the non-functional usermanager. As discussed during last teleconf, the new working usermanager has to be added to the distribution instead.


dizzzz:
@wolfgangmm does this require a change in the dashboard ?

wolfgangmm:
@dizzzz the change to the dashboard has been made with version 0.4.9 which is already in the public repo. We just need this PR for the installer.

dizzzz:
thx for explaining

PR #1732, opened 2018-02-13
Fixes a regression with XQuery Update insert

adamretter:
Closes https://github.com/eXist-db/exist/pull/1664


src/org/exist/dom/persistent/ElementImpl.java
```diff
@@ -1030,16 +1030,41 @@ public Node getFirstChild() {
...
+    /**
+     * Get the last child.
+     *
+     * @param attributesAreChildren In the DLN model attributes have child node ids,
+     *     however in the DOM model attributes are not child nodes. Set true for DLN
+     *     or false for DOM.
+     *
+     * @return the last child.
+     */
+    private Node getLastChild(final boolean attributesAreChildren) {
```
dizzzz: better to rename it to "isDLMmodel" ?

Issue #1682, opened 2018-01-11
From eXist 3.6.1 onwards, transform:transform() output of xmlns="" causes err:XQDY0102


finianoboyle:
In eXist 3.6.1, but not in eXist 3.6.0 or any previous versions, when `transform:transform()` outputs an element with `xmlns=""` which is embedded within prefixed or default namespace elements this causes the error:

```
org.xmldb.api.base.XMLDBException: err:XQDY0102 Cannot output a namespace node 
    for the default namespace when the element is in no namespace. 
```

If the same node sequence is constructed within XQuery it does not cause the error.

I've attached an XQuery which demonstrates the issue.

Finian

```xquery
xquery version "3.1";
(:
    In eXist-db 3.6.1 this XQuery causes the error
    
    "org.xmldb.api.base.XMLDBException: err:XQDY0102 Cannot output a namespace node 
    for the default namespace when the element is in no namespace. 
    [at line 28, column 1, source: xquery version "3.1";"
    
    In eXist-db 3.6.0, this XQuery returns the expected output:
    
    <package>
        <unprefixed xmlns="http://default-namespace">
            <prefixed:element xmlns:prefixed="http://prefixed-namespace">
                <dummy xmlns=""/>
            </prefixed:element>
        </unprefixed>
    </package

:)
let $xml := <anything/>
let $xslt := 
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" exclude-result-prefixes="#all" version="2.0">
    <xsl:template match="/">
        <unprefixed xmlns="http://default-namespace">
        <prefixed:element xmlns:prefixed="http://prefixed-namespace">
                <dummy xmlns=""/>
            </prefixed:element>
        </unprefixed>
    </xsl:template>
</xsl:stylesheet>

let $output := transform:transform($xml, $xslt, ())

let $package:=
<package>
{$output}
</package>

return $package
```

[test-an-error-in-361.txt](https://github.com/eXist-db/exist/files/1622074/test-an-error-in-361.txt)

dizzzz:
I actually get a different error (3.7.0-snapshot, almost identical to 3.6.1)

```
org.xmldb.api.base.XMLDBException: exerr:ERROR org.exist.xquery.XPathException: err:XPST0003 expecting "return", found 'null' [at line 37, column 11]

[...]

Issue #1586, opened 2017-10-13
Betterform integration in build target »dist«


mathias-goebel:
### What is the problem
when building eXist with the target "dist" (to place an executable version of the software in a single directory without the sources) the resulting software comes without a complete betterform integration, the initial observation is that the dashboard is not available (the browser reports 404 when requesting dojo.js from bfResources)

it seems that the `web.xml` misses the corresponding servlets configuration which get added by the build target `patchWebXml` from `extensions/betterform/build.xml`.

(if we manually call this task after the build is complete, we can correct this problem, thanks to @Lena-Luise who found this solution)

### What did you expect
a working instance of the database with a working dashboard - obviously.

### Describe how to reproduce or add a test
```bash
./build.sh dist
./dist/eXist*/bin/startup.sh
curl http://localhost:8080/exist/apps/bfResources/scripts/dojo/dojo.js
```

### Context information
Please always add the following information
- eXist-db version 3.6.0-SNAPSHOT commit f0d3647ec7b3fd66c642d3541a2645ffe244276e
- openjdk 1.8.0_131
- Linux, Ubuntu Server 64 bit
- no custom changes, just a cloned repo

inormann:
To make the `patchWebXml`  work it seems to be important to follow a certain build order. The following should be reproducable:
```
exist# git clone https://github.com/eXist-db/exist.git
exist# git checkout eXist-4.3.0
exist# ./build.sh
exist# ./build.sh dist
exist# cd dist/eXist-db-4.3.0/
exist/dist/eXist-db-4.3.0# ant -f extensions/betterform/build.xml patchWebXml
exist/dist/eXist-db-4.3.0# bin/startup.sh
```
The Dashboard should be now visible in the Browser.

adamretter:
@mathias-goebel @inormann So I think the following should work:


```
git clone https://github.com/eXist-db/exist.git
cd exist
git checkout eXist-4.3.0
./build.sh extension-betterform dist
```

The problem you have is that the Dashboard (and some apps) were designed in a way that lacked a separation of concerns, as such they require a component from betterForm which is actually not really anything to do with XForms or betterForm.

Unfortunately that means they will not work without betterForm being present. Normally when you compile eXist-db, it sets up  and installs betterForm for you by default. If you disable betterForm in `$EXIST_HOME/extensions/local.build.properties` and just build eXist from source - you will see the same problems with the Dashboard etc.

Another option may be to follow this process:

```
[...]

Issue #1352, opened 2017-03-14
error in XFormsFilter blocks deployment of eXist WAR in Tomcat-8.5


rvdb:
### What is the problem
Deploying an eXist-3.1.0 WAR build in Tomcat-8.5 fails due to errors caused by the XFormsFilter filter. Following errors are being reported:

    14-Mar-2017 14:26:53.792 SEVERE [localhost-startStop-1] org.apache.catalina.core.StandardContext.filterStart Exception starting filter XFormsFilter
     javax.servlet.ServletException: de.betterform.xml.config.XFormsConfigException: javax.xml.transform.TransformerException: javax.xml.transform.TransformerConfigurationException: net.sf.saxon.s9api.SaxonApiException: I/O error reported by XML parser processing file:/F:/ctb/CTB_webapps/CTBapps_dev/active/exist-3.1.0/WEB-INF/classes/META-INF/resources/xhtml.xsl: F:\tomcat-8.5.12\webapps\exist-3.1.0\WEB-INF\classes\META-INF\resources\xhtml.xsl (The system cannot find the path specified)
      at de.betterform.agent.web.filter.XFormsFilter.init(XFormsFilter.java:76)
      at org.apache.catalina.core.ApplicationFilterConfig.initFilter(ApplicationFilterConfig.java:279)
      at org.apache.catalina.core.ApplicationFilterConfig.getFilter(ApplicationFilterConfig.java:260)
      at org.apache.catalina.core.ApplicationFilterConfig.<init>(ApplicationFilterConfig.java:105)
      at org.apache.catalina.core.StandardContext.filterStart(StandardContext.java:4599)
      at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5242)
      at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
      at org.apache.catalina.core.ContainerBase.addChildInternal(ContainerBase.java:752)
      at org.apache.catalina.core.ContainerBase.addChild(ContainerBase.java:728)
      at org.apache.catalina.core.StandardHost.addChild(StandardHost.java:734)
      at org.apache.catalina.startup.HostConfig.deployDirectory(HostConfig.java:1107)
      at org.apache.catalina.startup.HostConfig$DeployDirectory.run(HostConfig.java:1841)
      at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
      at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

Issue #1335, opened 2017-03-07
infinispan.xml to be reset after rebuild?


pherk:
no big problem, but annoying.
I frequently forget to re-edit 
   webapp/WEB-INF/classes/infinispan.xml
after rebuild (bfSessionCache too small).

regards 
Peter

joewiz:
Thank you for reporting your issue!

To be able to better understand you problem or suggestion, please add as much information as possible to this ticket. When there is no or just little information, it will be too difficult to work on it (and hence we will not be able to solve it)

Please fill in the following sections:

### What is the problem
> Describe exactly what you see (e.g. an output of an XQuery) 

### What did you expect
> Describe what you expected to happen. Add for example a reference to a [specification](https://www.w3.org/TR/xquery-3/).

### Describe how to reproduce or add a test
> Describe how we can can reproduce the problem.

> The *best* way is to provide a 'self containing' test, a test that can be run without dependancies. The [XQSuite - Annotation-based Test Framework for XQuery](http://exist-db.org/exist/apps/doc/xqsuite.xml) makes it very easy for you to create tests. These tests can be executed from the [eXide editor](http://exist-db.org/exist/apps/eXide/index.html) (XQuery - Run as Test)

### Context information
Please always add the following information
- eXist-db version + Git Revision hash e.g. eXist-db 3.0 /  acd0c14 
- Java version (e.g. Java8u121)
- Operating system (Windows 7, Linux, MacOs)
- 32 or 64 bit
- Any custom changes in e.g. conf.xml

pherk:
I respect the template but thought it would be clear enough ;-)
My problem: parameter changes in infinispan.xml will be reset with each rebuild
  build.xml 
         copy $betterform.webapp/WEB-INF/classes/infinispan.xml to $exist.webapp/WEB-INF/classes
My Expection: changes should be preserved
Test: not needed, see above
Context:
- eXist-db newest develop
- java8
- Linux Ubuntu 12.04LTS, 64bit

PR #1736, opened 2018-02-17
Remove betterForm and Update to latest Saxon

adamretter:
I can no longer maintain the old betterForm extension. I have updated it previously for the past two releases of Saxon.

However, having again tried to update it for the latest version of Saxon, I have not managed to do this after several days of expended effort. There are just too many changes!
I am no longer prepared to maintain the betterForm extension and without anyone else be willing to do so, I see our only option is to remove it.

This PR both removes the betterForm extension, and updates us to the latest version of Saxon.

Closes https://github.com/eXist-db/exist/issues/1682
Closes https://github.com/eXist-db/exist/issues/1586
Closes https://github.com/eXist-db/exist/issues/1352
Closes https://github.com/eXist-db/exist/issues/1335
Closes https://github.com/eXist-db/exist/issues/560


dizzzz:
@JoernT can you elaborate on the future of betterForms? I "heard" rumours that there is some revision being done?

JoernT:
@adamretter first of all big thanks for your efforts. I generally not opposed to removing it as i see the saxon dependency as a big burden. However we have to check internally cause there's at least one customer project still relying on it. I'll report back how we can go on. Same might be true for people from the community that might use it - so, at least we should ask the list before we take the final step as we have no replacement for it.

@dizzzz well, that are only rumors ;) We're lacking the budget to actually do that cause it will take *at least* 3 months of work to do that and we have no justification for such a big investment - haven't heard much about XForms users lately. If i do a rewrite one day it would hardly conform to the standard but just take the best parts of it and port them into todays' world.

adamretter:
TODO - Add scripts to downgrade Saxon version and re-enable betterForm; as per-community call from 2017-03-12.

JoernT:
@all just a reminder - when removing betterFORM we also need to migrate to the new existdb-dashboard, usermanager, packagemanager etc. as there will be no Dojo any more to run the old dashboard. I'm happy to assist necessary changes

PR #1737, opened 2018-02-17
Enable & repair extensions XQuery tests

adamretter:
For some reason the tests written in XQuery for some extension modules were not enabled by default. I have now re-enabled.

1. I have fixed broken tests where I could.
2. I have also fixed a bug in QName Wildcard handling in the Lucene index.
3. There are some Lucene tests which are still broken, and need attention from the authors:

    1. `plain-store.xml`. @dizzzz these have you listed as the author. The two that fail in that suite seem to me that they could never succeed. Can you take a look and either fix or remove please?

    2. `LuceneFT+range_indexRetrievalTest.xml`. @rvdb these have you listed as the author. Weirdly if I copy and paste the code and data into the Java Admin Client, the tests seem to pass, but not when run from the test suite. Perhaps @wolfgangmm (who I think wrote the test suite code) has an idea?

    3. `analyzers.xml`. @wolfgangmm these have you listed as the author. The two failing tests are related to the `KeywordAnalyzer`, I can't see a problem with the tests themselves, perhaps either the `KeywordAnalyzer` is not the right choice here, or there is a bug in Lucene? Can you take a look and either fix, make as ignore, or remove please?


dizzzz:
@adamretter I see my name, but looking at the [git history](https://github.com/eXist-db/exist/commits/develop/extensions/indexes/lucene/test/src/xquery/lucene/plain-store.xml) .....  I am not really sure if I actually contributed to this test

rvdb:
Hi @adamretter , I'm afraid I can't be of much help; can't even get the test running anymore :worried: 

> Function t:run-testSet() is not defined in namespace 'http://exist-db.org/xquery/testing'

(even though the module seems to be enabled in eXist-4.0.0's conf.xml) . From the (earliest history of that test file, I see how the test used to pass, though...

adamretter:
@rvdb Can you tell me how you are trying to run the test please?

rvdb:
@adamretter Ah, I forgot to call `t:run-testSet()` with the extra required parameter which had been introduced later, it seems. Since no documentation for the old test suite seems to exist (anymore?), I had to look up how to run such tests in my email archive, where I found a reference to http://markmail.org/message/kya6e6flj3dwqbbm, which records how this extra parameter had bitten me in the past already.  

Anyway, I can now get the test running with following XQuery (either in eXide or Java Client):

    import module namespace t="http://exist-db.org/xquery/testing";
    let $testSet := doc('https://raw.githubusercontent.com/eXist-db/exist/develop/extensions/indexes/lucene/test/src/xquery/lucene/LuceneFT%2Brange_indexRetrievalTest.xml')/*
    return t:run-testSet($testSet, ())

This indeed shows 2 failing tests: 

    <TestSet>
        <testName>tests for index retrieval of Lucene FT and range indexed nodes</testName>
        <description>
            <p>these tests test whether the util:index-keys() function returns correct nodes for the Lucene FT and range indexex</p>
            <author>Ron Van den Branden</author>
        </description>
        <test n="1" pass="true"/>
        <test n="2" pass="true"/>
        <test n="3" pass="false">
            <task>index scan on path-based range indexed nodes</task>
            <expected>
                <entry>
                    <term>some text inside a paragraph</term>
                    <frequency>2</frequency>
                    <documents>2</documents>
                    <position>1</position>
                </entry>
                <entry>
                    <term>some text inside an element</term>
                    <frequency>2</frequency>
                    <documents>2</documents>
                    <position>2</position>
                </entry>
            </expected>
            <result>
                <entry>
                    <term>
                        some text inside an element
                    </term>
                    <frequency>2</frequency>
                    <documents>2</documents>
                    <position>1</position>
                </entry>
                <entry>
                    <term>some text inside a paragraph</term>
                    <frequency>2</frequency>
                    <documents>2</documents>
                    <position>2</position>
                </entry>
            </result>
        </test>
        <test n="4" pass="false">
            <task>index scan on qname-based range indexed nodes</task>
            <expected>
                <entry>
                    <term>some text inside a paragraph</term>
                    <frequency>2</frequency>
                    <documents>2</documents>
                    <position>1</position>
                </entry>
                <entry>
                    <term>some text inside an element</term>
                    <frequency>2</frequency>
                    <documents>2</documents>
                    <position>2</position>
                </entry>
            </expected>
            <result>
                <entry>
                    <term>
                        some text inside an element
                    </term>
                    <frequency>2</frequency>
                    <documents>2</documents>
                    <position>1</position>
                </entry>
                <entry>
                    <term>some text inside a paragraph</term>
                    <frequency>2</frequency>
                    <documents>2</documents>
                    <position>2</position>
                </entry>
            </result>
        </test>
    </TestSet>

(same results if I replace the `$testSet` URL with a reference to the test file on my FS.) Yet, if I just replace `$testSet` in the XQuery above with the content of the test file, and run that in eXide / Java Client, all tests pass.

From these differences, it seems a whitespace issue is at play, causing differences when elements are read with `doc()` function, or declared inline in a XQuery variable. 

Actually, I'm starting to doubt whether the test should pass, since indeed the range index here is defined on the `<pPath>` and `<pQname>` elements, which *do* contain extra whitespace, that should be indexed as well in the range index. Indeed, if I remove the extra WS from those nodes in the file at my FS and load it with the `doc()` function, the tests pass as well.

So, with apologies for the long answer to a simple question:

1. I do think the current test file is incorrect, and I'm willing to correct it immediately
2. Yet, I'm wondering if this difference doesn't indicate there could be a problem with the way WS is interpreted in XML fragments delcared in XQuery variables?

adamretter:
@rvdb It may be that how white-space handling is managed has changed. I think @ljo and/or @wolfgangmm would be best to comment on that. Perhaps once we have their input, we know whether to change the test, or the data, or that there is another issue at play here.

Thanks for your quick response.

adamretter:
@dizzzz So it looks from the Git history that @wolfgangmm added the test but with your name in. Perhaps you submitted tests to him directly? or perhaps it was a copy-paste error?

dizzzz:
@adamretter I have no really idea..... It looks like that the tests assume a certain order, which is not ok. @wolfgangmm can you elaborate? Otherwise i'd suggest to delete the [test file](https://github.com/eXist-db/exist/blob/develop/extensions/indexes/lucene/test/src/xquery/lucene/plain-store.xml).

joewiz:
@adamretter Perhaps you were aware, but it looks like there is a conflict in `extensions/indexes/lucene/build.properties`.

adamretter:
@joewiz Sure I will rebase shortly.

joewiz:
I've just raised the priority of this PR, as the problem it fixes prevented us from detecting regressions that were introduced into the source base and included in the 4.0.0 release; e.g., https://github.com/eXist-db/exist/issues/1744#issuecomment-368348252. Once this PR is merged and those issues are fixed, we should consider a 4.0.1 hotfix release.

dizzzz:
is `plain-store.xml` the only test that is now failing? +1 for me to gave it removed

PR #1747, opened 2018-02-23
XSuite - Proper integration between JUnit, and XQSuite and XML XQuery tests

adamretter:
Previously with `xquery.TestRunner` we would have only two JUnit test results (one for XQSuite based tests and one for `test.xq` based tests) regardless of how many XQuery tests were really being executed by XQSuite or `test.xql`.

This PR offers a proper integration with JUnit, it supports both XQSuite and test.xql tests and reports every test execution to JUnit. We have named this integration `XSuite`.

We now provide a JUnit Suite and Runners in `org.exist.test.runner`.

Migrating from `TestRunner` to XSuite is very simple, consider this old TestRunner declaration for Map tests:

```java
package xquery.maps;

import xquery.TestRunner;

public class MapTests extends TestRunner {
    @Override
    protected String getDirectory() {
        return "test/src/xquery/maps";
    }
}
```

The updated code for XSuite looks like:

```java
package xquery.maps;

import org.exist.test.runner.XSuite;
import org.junit.runner.RunWith;

@RunWith(XSuite.class)
@XSuite.XSuiteFiles({
    "test/src/xquery/maps"
})
public class MapTests {
}
```

The `XSuiteFiles` annotation takes one or more paths to either directories containing XQSuite test files, or test.xql XML test files.  If you need to be more specific, XSuite also supports paths to test files instead of directories through the same mechanism.


dizzzz:
would this enable running xquery tests in the maven build?

adamretter:
@dizzzz can do yes

src/org/exist/source/SourceFactory.java
```diff
@@ -177,15 +177,42 @@ else if (location.startsWith("file:/") || !location.contains(":/")) {
 
             if (source == null) {
                 /*
-                 * Lastly we try to load it using EXIST_HOME as the reference point
+                 * Try to load from the contextPath URL folder
```
dizzzz: Is this safe (enough)? I can't estimate the risks...

adamretter: Should be. All the tests pass

dizzzz: I meant: this adds file loading outside EXIST_HOME right? Will this change able ppl to break out an load e.g.  `/etc/shadow` ?

adamretter: We already had external loading before.  But none of it will parse as an XQuery do it won't load

adamretter:
@dizzzz Thanks :-)

dizzzz:
thank you for this improvement !

PR #1751, opened 2018-02-25
Further removal of test logging to StdOut/StdErr

adamretter:


test/src/org/exist/xmldb/ResourceTest.java
```diff
@@ -254,7 +254,6 @@ public void queryRemoveResource() throws XMLDBException {
 
         for (ResourceIterator iterator = rs.getIterator(); iterator.hasMoreResources();) {
             Resource r = iterator.nextResource();
-            System.err.println("Resource id=" + r.getId() + " xml=" + r.getContent());
```
dizzzz: no fail() replacement ?

adamretter: It was not previously considered a fail. I suspect, it was previously meant to say `System.out.println`, but well both in tests apart from debugging is bad.

adamretter: @dizzzz We good?

dizzzz: yes, sorry.

PR #1759, opened 2018-03-01
Fix parsing of qnames in XQuery

### LINKED ISSUE #1725: Function call as map value (official syntax)
## What is the problem
When trying to update some older code that uses maps, I found that exist struggles with the xpath3.1 map syntax `:` instead of `:=`. The sample query below returns the following error in exist.
`xquery: err:XPST0081 No namespace defined for prefix pos:fn`

This looks like a lexer problem the old (`:=`) syntax returns the desired maps in exist. 

```
map {
    1: 5
}

map {
    2: 3
}
```

~~However, in the problematic case checking with Saxon`9.7.0.19`, I see: 
 `Cannot add a map to an XML tree` 
which suggests I m trying to do something that I shouldn't that has nothing to do with namespaces.~~

~~The old syntax won't run:
`The ':=' notation is no longer accepted in map expressions: use ':' instead`~~

~~So I am not sure what the correct response from eXist should be, ie if the query should return a result of if even the old syntax should be changed to raise an error.~~ 

### What did you expect
Replacing `:=` with `:`according to [Xpath 3.1 specification](https://www.w3.org/TR/xpath-31/#id-maps-and-arrays) to not result in broken queries.

this might be of interest from the specs:
>For instance, consider the expression map{a:b}. Although it matches the EBNF for MapConstructor (with a matching MapKeyExpr and b matching MapValueExpr), the "longest possible match" rule requires that a:b **be parsed as a QName, which results in a syntax error. Changing the expression to map{a :b} or map{a: b} will prevent this**, resulting in the intended parse.

```xquery
for $n at $pos in ("hello", "you")
return
    map {  $pos: fn:string-length($n) }
```

see #1510 

### Context information
- eXist-db `3.6.1`
- Java version `1.8.0_162`
- Operating system (MacOs)

wolfgangmm:
The XQuery parser ignored whitespace when parsing a qname, leading to a wrong interpretation of ':' in some contexts, e.g. map constructors. Moved qname parsing to the lexer so it gains precedence. Closes #1725.


joewiz:
@duncdrum requested:

> can someone put a documentation label on #1759 I ll need to check map code samples in the docs

I've just done this.

PR #1760, opened 2018-03-02
Some further test fixes

adamretter:
I saw some strange things with the Range Index tests written in XQuery and realised that XSuite (and previously TestRunner) were not reporting errors that occurred during `%test:setUp` or `%test:tearDown` to JUnit, in fact the tests appeared to pass fine.

I have now fixed that, and also that showed some problems with the cleanup of some tests written in XQuery, so I fixed that too!

Cheers.

### DIFF
--- a/src/org/exist/xquery/functions/map/AbstractMapType.java
+++ b/src/org/exist/xquery/functions/map/AbstractMapType.java
@@ -1,6 +1,8 @@
 package org.exist.xquery.functions.map;
 
 import com.ibm.icu.text.Collator;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.exist.dom.QName;
 import org.exist.xquery.*;
 import org.exist.xquery.functions.fn.FunDistinctValues;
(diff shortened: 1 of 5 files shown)


src/org/exist/xquery/functions/map/AbstractMapType.java
```diff
@@ -1,11 +1,15 @@
...
 import com.ibm.icu.text.Collator;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.exist.dom.QName;
 import org.exist.xquery.*;
 import org.exist.xquery.functions.fn.FunDistinctValues;
 import org.exist.xquery.value.*;
 
+import javax.xml.xpath.XPath;
+import java.util.AbstractMap;
```
wolfgangmm: I can't see where XPath or AbstractMap may be used in this class?

adamretter: @wolfgangmm oops. Thanks. I have now removed those erroneous imports and rebased

Issue #194, opened 2014-04-12
Index usage breaks text() selector


vogelsgesang:
I stumbled across the following bug while I was trying to setup indexing:

I have an easy Xquery:

``` xquery
xquery version "3.0";
declare default element namespace 'http://www.loc.gov/MARC21/slim';
<collection>{subsequence((
   doc("/db/od/books_export.xml")/collection/record[datafield/subfield/text() = 'algebraic']
),0,2)}
</collection>
```

This query returns one matching result.

If I add the following file as "/db/system/config/db/od/collection.xconf" (without reindexing the database!), the above Xquery returns an empty set of results.

``` xml
<collection xmlns="http://exist-db.org/collection-config/1.0">
    <index xmlns:marc="http://www.loc.gov/MARC21/slim">
        <!-- Range indexes -->
        <range>
            <create qname="marc:subfield" type="xs:string"/>
        </range>
    </index>
</collection>
```

~~It seems that eXistDb tries to use the range index although it was not created so far. To me this is very confusing and this behaviour should be changed or at least documented.~~

duncdrum:
see #2363 
@dizzzz i m a bit lost trying to follow the conversation, could you post an update in here, if this is superseded, still an issue, … ?

daliboris:
I can test it.

daliboris:
Following XQTest checks most conditions specified in the previous [comment](https://github.com/eXist-db/exist/issues/194#issuecomment-40329315) - except reindexing collection after `collection.xconf` change. All tests pass on existdb/existdb:6.2.0 (Docker).

```xquery
xquery version "3.1";
module namespace t="http://exist-db.org/xquery/test";
declare namespace test="http://exist-db.org/xquery/xqsuite";
declare namespace exist = "http://exist.sourceforge.net/NS/exist";

(:
  See https://github.com/eXist-db/exist/issues/194#issuecomment-40329315
:)

declare variable $t:XML-test := document { 
<test>
  <existing>A</existing>
  <new>B</new>
</test>
};


declare variable $t:xconf-existing :=
    <collection xmlns="http://exist-db.org/collection-config/1.0">
    <index xmlns:xs="http://www.w3.org/2001/XMLSchema">
        <range>
         <create qname="existing" type="xs:string" />
        </range>
    </index>
</collection>;

declare variable $t:xconf-new :=
    <collection xmlns="http://exist-db.org/collection-config/1.0">
[...]

PR #1763, opened 2018-03-05
Range index returns invalid results

wolfgangmm:
Several users reported wrong nodes being returned as the result of an XQuery predicate using a range index. While the result looked ok on first sight, serializing it would show seemingly random nodes from the same document. Unfortunately the issue was not easy to reproduce. Several people tried to write tests, including @joewiz, @windauer and myself, but could not trigger the bug in a simple test scenario. 

I finally managed to download a complete database dump from one production server where the bug just became manifest and managed to track it down to deleted lucene documents being returned by queries.

The deleted lucene documents were picked up by a filter used by the range index to optimize queries on a single context node. As a consequence, the range index returned pointers to invalid (deleted) nodes into the XQuery result. It randomly chose one of the found lucene entries, which explains why the bug was unpredictable.

The optimization achieved by the filter appeared to be rather minor in my tests. I thus would vote for dropping this filter altogether as it apparently is the cause of the issues.


joewiz:
From https://ci.appveyor.com/project/AdamRetter/exist/build/develop-354:

```
3272    [junit] Using temporary storage location: C:\Users\appveyor\AppData\Local\Temp\1\org.exist.test.ExistEmbeddedServer4141105031809044933
3273    [junit] Running xquery.arrays.ArrayTests
3274    [junit] Tests run: 142, Failures: 0, Errors: 1, Skipped: 2, Time elapsed: 1.984 sec
3275
3276BUILD FAILED
3277C:\projects\exist\build\scripts\junit.xml:153: Test xquery.arrays.ArrayTests failed
3278
3279Total time: 7 minutes 51 seconds
3280Command exited with code 1
```

dizzzz:
Interesting to see that the non-windows build succeeds

duncdrum:
I think we should revisit #194 in light of this PR to see if it still applies.

dizzzz:
I am really waiting for this fix.....

wolfgangmm:
@dizzzz If I can get the windows build fixed ...

dizzzz:
I understand. is it clear which test fails? or..... if there is only one...

wolfgangmm:
At the moment I still have 1 failure and 1 error. Both are very weird.

dizzzz:
I merged-in the recent changes of 'develop', rerunning tests..

dizzzz:
I did not see the problems anymore after this fix !

PR #1772, opened 2018-03-16
[bugfix] Support more locales for month and day names in format-date.

wolfgangmm:
format-date and format-dateTime only supported a small number of locales for the names of months and days of week. Java 8 has added java.time.Month and DayOfWeek for this purpose, which allow us to replace some custom code and support languages like French or Spanish.


wolfgangmm:
@brihaye Thanks Pierrick! The test explicitely asks for an ordinal number in `[D1o]`. Running the following through saxon:

```xquery
format-date(xs:date("2002-07-01"), "[FNn] [D1o] [MNn] [Y]", "fr", (), ())
```

returns

```
Lundi 1er Juillet 2002
```

Would this be a valid date in French or shall we always assume it has to be written without ordinal suffix?

brihaye:
Le 20/03/2018 à 21:07, Wolfgang Meier a écrit :

> |Lundi 1er Juillet 2002 |
> 
> Would this be a valid date in French

Yes.

Tell me now about :

Mardi 2 juillet 2002
Mercredi 3 juillet 2002 :-)

French is strange...

The suffix is only needed for "first"... Jeudi 11, vendredi 21, and 
samedi 31 don't need any suffix !

Enjoy :-)

p.b.

wolfgangmm:
@brihaye Thanks Pierrick, I modified the code according to your suggestion.

PR #1775, opened 2018-03-19
Random 0-byte XML docs in db

wolfgangmm:
A long standing bug in `Indexer` caused 0-byte documents to appear in database collections. The bug occurs rather randomly and is hard to reproduce. As I finally figured out, the 0-byte document was created due to the element stack used by Indexer being messed up. Closer inspection revealed a weird if condition checking the element qname in endElement (and only there) in a wrong way. This (probably very old) code seems to be completely unnecessary if we assume that a document with incorrect element hierarchy should be rejected by the parser. I thus vote for removing it completely.

The PR also fixes a small issue in package deployment which unfortunately did hide the 0-byte bug during deployment. The code for handling HTML5 documents catched all exceptions.

### DIFF
--- a/src/org/exist/Indexer.java
+++ b/src/org/exist/Indexer.java
@@ -398,7 +398,6 @@ public class Indexer extends Observable implements ContentHandler, LexicalHandle
     @Override
     public void endElement(final String namespace, final String name, final String qname) {
         final ElementImpl last = stack.peek();
-        if (last.getNodeName().equals(qname)) {
             processText(last, ProcessTextParent.ELEMENT_END);
             stack.pop();
             XMLString elemContent = null;
@@ -424,7 +423,6 @@ public class Indexer extends Observable implements ContentHandler, LexicalHandle
             currentPath.removeLastComponent();
             setPrevious(last);
             level--;
-        }
     }
 
     /**
--- a/src/org/exist/repo/Deployment.java
+++ b/src/org/exist/repo/Deployment.java
@@ -786,22 +786,26 @@ public class Deployment {
 
                 try(final Txn transaction = mgr.beginTransaction()) {
                     if (mime.isXMLType()) {
+                        final InputSource is = new FileInputSource(file);
+                        IndexInfo info = null;
                         try {
-                            final InputSource is = new InputSource(file.toUri().toASCIIString());
-                            final IndexInfo info = targetCollection.validateXMLResource(transaction, broker, name, is);
+                            info = targetCollection.validateXMLResource(transaction, broker, name, is);
+                        } catch (EXistException | PermissionDeniedException | LockException | SAXException | IOException e) {
+                            //check for .html ending
+                            if(mime.getName().equals(MimeType.HTML_TYPE.getName())){
+                                //store it as binary resource
+                                storeBinary(targetCollection, file, mime, name, requestedPerms, transaction);
+                            } else {
+                                // could neither store as xml nor binary: give up and report failure in outer catch
+                                throw new EXistException(FileUtils.fileName(file) + " cannot be stored");
+                            }
+                        }
+                        if (info != null) {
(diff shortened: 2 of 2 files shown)


src/org/exist/Indexer.java
```diff
@@ -398,7 +398,6 @@ private void processText(ElementImpl last, ProcessTextParent ptp) {
     @Override
     public void endElement(final String namespace, final String name, final String qname) {
         final ElementImpl last = stack.peek();
-        if (last.getNodeName().equals(qname)) {
```
adamretter: @wolfgangmm So this check was incorrect somehow? But is it safe to actually remove it?

wolfgangmm: @adamretter The test doesn't properly respect namespaces. But I see no reason why the node name passed to endElement would ever be different than the one startElement received on the same hierarchical level - unless the XML is actually invalid. In the latter case the XML parser would have complained during the validation step though. We could certainly fix the check, but in my opinion it is completely unnecessary and smells like old code left behind.

I did not see any side effects despite having tested this PR quite extensively. The 0-byte documents this bug produces are a major showstopper for many users and have been reported occasionally since at least half a year.

adamretter: Okay sounds cool

adamretter:
@wolfgangmm Can you rebase please so that AppVeyor is passing...

dizzzz:
done

adamretter:
This seems to fail OracleJDK 9

dizzzz:
```
    [junit] WARNING: All illegal access operations will be denied in a future release
    [junit] Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
    [junit] Running org.exist.xquery.modules.sql.ExecuteFunctionTest
    [junit] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.59 sec
    [junit] Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
    [junit] Running org.exist.xquery.update.IndexIntegrationTest
    [junit] WARNING: An illegal reflective access operation has occurred
    [junit] WARNING: Illegal reflective access by org.easymock.cglib.core.ReflectUtils$1 (file:/home/travis/build/eXist-db/exist/lib/test/easymock-3.5.1.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
    [junit] WARNING: Please consider reporting this to the maintainers of org.easymock.cglib.core.ReflectUtils$1
    [junit] WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
    [junit] WARNING: All illegal access operations will be denied in a future release
    [junit] Tests run: 3, Failures: 1, Errors: 4, Skipped: 0, Time elapsed: 6.179 sec
BUILD FAILED
/home/travis/build/eXist-db/exist/build/scripts/junit.xml:153: Test org.exist.xquery.update.IndexIntegrationTest failed
```

dizzzz:
easy mock / java9 issue ; 
- https://docs.oracle.com/javase/9/migrate/toc.htm#JSMIG-GUID-2F61F3A9-0979-46A4-8B49-325BA0EE8B66
- https://stackoverflow.com/questions/46671472/illegal-reflective-access-by-org-springframework-cglib-core-reflectutils1

dizzzz:
in intellij i see the following error with java9u04:

```
Error:Module 'eXist-db' production: org.aspectj.bridge.AbortException: can't determine superclass of missing type java.lang.Object
 [Xlint:cantFindType]
```

wolfgangmm:
Testing locally with Java 9 I did not get any test failures. I have rerun the travis build and now it appears to be green as well.

adamretter:
@wolfgangmm Okay then, let's put it down to a Travis glitch then.

PR #1777, opened 2018-03-19
Closing inputstreams and improve ByteArrayOutputStream usage

dizzzz:
As inspired my a recent change of @wolfgangmm I found these input streams not being closed....

Not sure if the assumptions are correct though, eg in the context of BinaryValues ...

### DIFF
--- a/extensions/modules/src/org/exist/xquery/modules/image/CropFunction.java
+++ b/extensions/modules/src/org/exist/xquery/modules/image/CropFunction.java
@@ -132,11 +132,10 @@ public class CropFunction extends BasicFunction {
         String formatName = mimeType.substring(mimeType.indexOf("/")+1);
 
         //TODO currently ONLY tested for JPEG!!!
-        Image image = null;
         BufferedImage bImage = null;
         try {
             //get the image data
-            image = ImageIO.read(((BinaryValue)args[0].itemAt(0)).getInputStream());
+            Image image = ImageIO.read(((BinaryValue)args[0].itemAt(0)).getInputStream());
             //			image = ImageModule.getImage((Base64BinaryValueType)args[0].itemAt(0));
             //      			image = ImageIO.read(new ByteArrayInputStream(getImageData((Base64BinaryValueType)args[0].itemAt(0))));
             if(image == null) {
@@ -149,6 +148,7 @@ public class CropFunction extends BasicFunction {
             if(cropImage instanceof BufferedImage) {
                 // just in case cropImage is allready an BufferedImage
                 bImage = (BufferedImage)cropImage;
+
             } else {
                 bImage = new BufferedImage(cropImage.getHeight(null),
                 cropImage.getWidth(null),BufferedImage.TYPE_INT_RGB);
--- a/extensions/modules/src/org/exist/xquery/modules/image/CropFunction.java
+++ b/extensions/modules/src/org/exist/xquery/modules/image/CropFunction.java
@@ -33,6 +33,7 @@ import java.awt.image.BufferedImage;
 import java.awt.image.CropImageFilter;
 import java.awt.image.FilteredImageSource;
 import java.io.ByteArrayInputStream;
+import java.io.InputStream;
 import javax.imageio.ImageIO;
 
 import org.exist.dom.QName;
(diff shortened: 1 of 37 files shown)


extensions/modules/src/org/exist/xquery/modules/image/CropFunction.java
```diff
@@ -132,11 +132,10 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
         String formatName = mimeType.substring(mimeType.indexOf("/")+1);
 
         //TODO currently ONLY tested for JPEG!!!
-        Image image = null;
         BufferedImage bImage = null;
         try {
             //get the image data
-            image = ImageIO.read(((BinaryValue)args[0].itemAt(0)).getInputStream());
+            Image image = ImageIO.read(((BinaryValue)args[0].itemAt(0)).getInputStream());
```
adamretter: This `InputStream` needs to be closed when it is finished with.

From the javadoc:

> This method <em>does not</em> close the provided
> <code>InputStream</code> after the read operation has completed;
> it is the responsibility of the caller to close the stream, if desired.

extensions/modules/src/org/exist/xquery/modules/mail/SendEmailFunction.java
```diff
@@ -343,22 +344,17 @@ public SMTPException(Throwable cause)
...
 
-        try
-        {
-            //Create a Socket and connect to the SMTP Server
-            smtpSock = new Socket(SMTPServer, TCP_PROTOCOL_SMTP);
+        try (//Create a Socket and connect to the SMTP Server
+             Socket smtpSock = new Socket(SMTPServer, TCP_PROTOCOL_SMTP);
+
+             //Create a Buffered Reader for the Socket
+             BufferedReader smtpIn = new BufferedReader(new InputStreamReader(smtpSock.getInputStream()));
```
adamretter: `smtpSock.getInputStream` should be bound to it's own variable before the `BufferedReader` to ensure that it is actually closed. Same with the `OutputStream`:

e.g.
```java
try (
  final InputStream is = smtpSock.getInputStream();
  final Reader smtpIn = new BufferedReader(new InputStreamReader(is));
  final OutputStream os = smtpSock.getOutputStream();
  final OutputStreamWriter osw = new OutputStreamWriter(os, charset)
  final Writer smtpOut = new PrintWriter(osw)
) {
```

samples/src/org/exist/examples/http/PostExample.java
```diff
@@ -74,10 +74,11 @@ private void doPost(String request) throws IOException {
 		os.write(request.getBytes(UTF_8));
 		connect.connect();
 
-        BufferedReader is = new BufferedReader(new InputStreamReader(connect.getInputStream()));
-        String line;
-        while((line = is.readLine()) != null)
-            System.out.println(line);
+        try (BufferedReader is = new BufferedReader(new InputStreamReader(connect.getInputStream()))) {
```
adamretter: Split these out again:
e.g.
```java
try (
  final InputStream is = connect.getInputStream();
  final Reader isr = new InputStreamReader();
  Reader is = new BufferedReader(isr))) {
```

samples/src/org/exist/examples/http/PutExample.java
```diff
@@ -86,10 +86,12 @@ public static void main(String[] args) {
             connect.setRequestMethod("GET");
             connect.connect();
             System.out.println("Result:");
-            BufferedReader bis = new BufferedReader(new InputStreamReader(connect.getInputStream()));
-            String line;
-            while ((line = bis.readLine()) != null) {
-                System.out.println(line);
+
+            try (BufferedReader bis = new BufferedReader(new InputStreamReader(connect.getInputStream())) ) {
```
adamretter: Split these out like above

dizzzz:
Reworked PR ; added some additional improvements

extensions/modules/src/org/exist/xquery/modules/mail/SendEmailFunction.java
dizzzz: is this for style, or does the JVM/compiler not recognise the original way of coding? (wondering what the spec is)

dizzzz:
Work is finished now :-)

adamretter:
Why the switch to the Apache `ByteArrayOutputStream`?

dizzzz:
It is a drop in replacement with some  api additions that makes things more efficient. Internally it does do a much better job than the java.io version regarding memory reallocations.

dizzzz:
In the past I have migrated more exist classes to use the commons-io equivalent

dizzzz:
the main thing that using the apache ByteArrayOutputStream solves: fill ByteArrayOutputStream  -> get ByteArray (expensive!)-> read data via ByteArrayInputStream  

Benefits Apache's ByteArrayOutputStream :+1: 
- no copy to byte[] is required ... saves 1/2 of memory !
- ByteArrayOutputStream  does do some smart stuff with internal memory allocations.

[Quote](https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/output/ByteArrayOutputStream.html): "In contrast to the original it doesn't reallocate the whole memory block but allocates additional buffers. This way no buffers need to be garbage collected and the contents don't have to be copied to the new buffer. This class is designed to behave exactly like the original."

adamretter:
@dizzzz Looks good. I have just done some reading on the subject too. It seems we could actually do better than Apache Commons `ByteArrayOutputStream` as it still has `synchronized` access which we don't need if operating in a single-threaded call. I will do some experiments...

adamretter:
I think for the time being we can merge your changes. We can do more in a future PR...

PR #1779, opened 2018-03-22
Create move.yml

duncdrum:
Enable move bot in main repo


duncdrum:
so win and linux pass but macOS fails thats a new one. I think we should restart travis, adding files to `.github/` should not cause any functional changes to exist whatsoever.

PR #1783, opened 2018-03-27
[doc] Correct typo in method signature

adamretter:


adamretter:
@dizzzz Apologies I also bought over the API differences from Granite, I have fixed this now.

PR #1786, opened 2018-03-27
Update libs to latest versions

adamretter:


dizzzz:
I am not sure about the YAJSW part; I checked all scripts and configuration files, and some things have changed in the original project: there is a separate directory for .bat files, the configuration file has been extended a bit, the .sh scripts have pup/push statements in stead of `cd` etc.

see https://github.com/eXist-db/exist/pull/1789

Regarding my PR I have an additional question, I'll add it right away.

adamretter:
@dizzzz Regards YAJSW - I found that they were actually missing some files in their release that are reference by their shell scripts. I have opened an issue about that. As such I figured to just take the Java updates, as that should be fine in itself. The shell scripts don't seem to need to be changed for that as-far-as-I-can-see...

dizzzz:
@adamretter yes I saw your recent contributions and remarks. If you are certain that replacing the JAR files is sufficient, then we can do my 'overhaul' a bit later. 

With my PR we are more inline with the default distribution, scripts wise, and the strange 'change user' constructs in our code (added by ljo) are in the PR gone. probably a good thing.

PR #1794, opened 2018-04-02
Small fixes and cleanup in Lucene index code

adamretter:


extensions/indexes/lucene/src/org/exist/indexing/lucene/LuceneIndexWorker.java
```diff
@@ -300,7 +300,7 @@ public MatchListener getMatchListener(DBBroker broker, NodeProxy proxy) {
         boolean needToFilter = false;
         Match nextMatch = proxy.getMatches();
         while (nextMatch != null) {
-            if (nextMatch.getIndexId() == LuceneIndex.ID) {
```
dizzzz: Need to think about this; probably we can use object reference comparison here which is faster

adamretter: Object reference comparison is not really a safe idea here, as these are strings and strings are immutable copies.

dizzzz: for sure you are right, strictly speaking. But actually here a string reference check would be ok enough :-)

PR #1799, opened 2018-04-04
Remove deprecated VirtualTempFile

adamretter:
`VirtualTempFile` has always had a very sketchy design. It was likely misunderstood/misused and could lead to un-released file handles.

I have now removed `VirtualTempFile` and replaced it with the more modern `TemporaryFileManager` and/or the `FilterInputStreamCache`.


extensions/webdav/src/org/exist/webdav/MiltonDocument.java
```diff
@@ -250,96 +235,39 @@ public Long getContentLength() {
...
-                    // default value, but not suitable for MacOsX Finder.
-                    size = existDocument.getContentLength();
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug(String.format("Serializing XML to /dev/null to determine size (%s) MacFinder=%s",
+                            resourceXmldbUri, isMacFinder));
                 }
 
+                // Stream document to '/dev/null' and count bytes
+                final ByteCountOutputStream counter = new ByteCountOutputStream();
+                try {
```
dizzzz: WHat about putting this stream as try-resource and put `size = counter.getByteCount();` in the try-catch ?

src/org/exist/http/servlets/HttpServletRequestWrapper.java
```diff
@@ -332,14 +273,15 @@ public String getParameter(final String name) {
...
-            String encoding = request.getCharacterEncoding();
-            if (encoding == null) {
-                encoding = "UTF-8";
-            }
-            return new BufferedReader(new InputStreamReader(getContentBodyInputStream(), encoding));
-        } else {
-            return request.getReader();
+        String encoding = request.getCharacterEncoding();
+        if (encoding == null) {
+            encoding = "UTF-8";
```
dizzzz: use literal from NIO ?

src/org/exist/xmlrpc/SerializedResult.java
```diff
@@ -19,26 +19,28 @@
...
+
 /**
  * Simple container for the results of a query. Used to cache
  * query results that may be retrieved later by the client.
  *
  * @author jmfernandez
  */
 public class SerializedResult extends AbstractCachedResult {
-    protected VirtualTempFile result;
+    protected Path result;
```
dizzzz: so here we have a 'real' difference: VirtualTempFile could hold a complete document in memory as where Path always will have a resource on disk. @adamretter you are convinced that typically the OS cache will handle this transparently right?  Only when a virus scanner is involved, having a file might have additional performance impact no?

extensions/webdav/src/org/exist/webdav/MiltonDocument.java
adamretter: Done.

src/org/exist/http/servlets/HttpServletRequestWrapper.java
adamretter: Not much point here, as HttpServletRequest works with a String

src/org/exist/xmlrpc/SerializedResult.java
adamretter: Well actually `VirtualTempFile` could have both an on-disk temp-file and an in-memory copy.

The memory use is now controllable because we are using `CachingFilterInputStream` and you can choose the cache implementation in `conf.xml`.

Issue #1709, opened 2018-01-28
"Socket is closed" error using httpclient


wolfgangmm:
### What is the problem

On latest develop, when using httpclient to GET a resource which is returned as xs:base64Binary, a following util:binary-to-string will fail with a "Socket is closed" error.

### What did you expect
Sample query below works fine in eXist 3.6.1 and returns a JSON string.

### Describe how to reproduce or add a test
```xquery
xquery version "3.1";

import module namespace http="http://expath.org/ns/http-client";

let $url := "https://www.ssrq-sds-fds.ch/places-db-edit/views/loc-search.xq?query=sevelen"
let $request :=
    <http:request method="GET" href="{$url}"/>
let $response := http:send-request($request)
let $json := util:binary-to-string($response[2])
return
    $json
```

### Context information

- 3.7.0-SNAPSHOT 8685d2d4e 
- java version "1.8.0_162"

Relevant stacktrace:

```
Caused by: java.net.SocketException: Socket is closed
        at sun.security.ssl.SSLSocketImpl.checkEOF(SSLSocketImpl.java:1542) ~[?:1.8.0_162]
        at sun.security.ssl.AppInputStream.read(AppInputStream.java:95) ~[?:1.8.0_162]
        at org.apache.http.impl.io.SessionInputBufferImpl.streamRead(SessionInputBufferImpl.java:137) ~[httpcore-4.4.9.jar:4.4.9]
        at org.apache.http.impl.io.SessionInputBufferImpl.fillBuffer(SessionInputBufferImpl.java:153) ~[httpcore-4.4.9.jar:4.4.9]
        at org.apache.http.impl.io.SessionInputBufferImpl.read(SessionInputBufferImpl.java:206) ~[httpcore-4.4.9.jar:4.4.9]
        at org.apache.http.impl.io.ContentLengthInputStream.read(ContentLengthInputStream.java:176) ~[httpcore-4.4.9.jar:4.4.9]
        at org.apache.http.conn.EofSensorInputStream.read(EofSensorInputStream.java:135) ~[httpclient-4.5.4.jar:4.5.4]
        at org.exist.util.io.AbstractFilterInputStreamCache.read(AbstractFilterInputStreamCache.java:118) ~[exist.jar:3.7.0-SNAPSHOT]
        at org.exist.util.io.CachingFilterInputStream.read(CachingFilterInputStream.java:176) ~[exist.jar:3.7.0-SNAPSHOT]
[...]

PR #1800, opened 2018-04-04
Fix EXPath HTTP Client socket-closed regression

adamretter:
Closes https://github.com/eXist-db/exist/issues/1709


src/org/exist/xquery/value/BinaryValueFromFile.java
```diff
@@ -82,6 +92,7 @@ public boolean isClosed() {
     @Override
     public void close() throws IOException {
         channel.close();
+        closeListener.ifPresent(cl -> cl.accept(file));
```
dizzzz: if ` channel.close();` throws an exception... should we try to execute line 95 as well?

extensions/expath/src/org/expath/httpclient/model/exist/EXistResult.java
```diff
@@ -94,8 +92,13 @@ public void add(final Reader reader) throws HttpClientException {
     @Override
     public void add(final InputStream is) throws HttpClientException {
         try {
-            result.add(BinaryValueFromInputStream.getInstance(context, new Base64BinaryValueType(), is));
-        } catch(final XPathException xpe) {
+            // we have to make a temporary copy of the data stream, as the socket will be closed shortly
+            final TemporaryFileManager temporaryFileManager = TemporaryFileManager.getInstance();
+            final Path tempFile = temporaryFileManager.getTemporaryFile();
```
dizzzz: Will this for small files, a in-memory operation? I ask for performance reasons.... (IO overhead to disk makes things slow)

adamretter: I don't think you will notice the difference. As Network is so much slower than disk in the first place

src/org/exist/xquery/value/BinaryValueFromFile.java
adamretter: Good idea. I will fix...

adamretter: Fixed.

joewiz:
Excellent! Thanks **so** much, Adam and Dannes!

PR #1808, opened 2018-04-08
Fix regression in chunked transfer of XML-RPC

adamretter:
@dizzzz I think this is the issue you described to me recently; I just found it myself.


dizzzz:
I am not sure what I reported :-) but thnx for the fix :-)

dizzzz:
tricky code, but I think it looks good ... thnx for the tests!

PR #1809, opened 2018-04-09
[bugfix] Align xml:id handling with saxon and basex

wolfgangmm:
Improves #1360 and addresses #540. Testing against saxon and basex showed that both are much more permissive about xml:id in constructed documents, which makes sense because it is quite common e.g. when writing XSLT stylesheets to declare an xml:id whose value is not strictly a valid NCName. Storing documents like this in eXist resulted in an error, causing trouble for XSLT users. Also, the check in eXist (based on xerces code), did not allow some unicode characters which were added in later versions of the spec.

The XQuery spec says an implementation MAY trigger an error on an invalid xml:id when constructing a document, but does not require it. Included tests have been run against saxon and basex to make sure we see the same behaviour on all implementations.

### DIFF
--- a/src/org/exist/Indexer.java
+++ b/src/org/exist/Indexer.java
@@ -653,10 +653,6 @@ public class Indexer extends Observable implements ContentHandler, LexicalHandle
                     // type to ID
                     attr.setValue(StringValue.trimWhitespace(StringValue.collapseWhitespace(attr.getValue())));
 
-                    if (!XMLChar.isValidNCName(attr.getValue())) {
-                        throw new SAXException("Value of xml:id attribute is not a valid NCName: " + attr.getValue());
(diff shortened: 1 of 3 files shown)


brihaye:
Maybe an option from the config file in order to keep the old behaviour would have been useful ?

wolfgangmm:
Why would you want to keep the old behavior? The previous behavior rejected documents which were reported valid by all other tools (XML parsers, oXygen ...), thus causing quite some confusion. If you really want strict validation, there are better ways to achieve it.

brihaye:
Just to apply : an implementation MAY trigger an error :-)
Anyway. I won't fight for this...

dizzzz:
are we good to go?

PR #1810, opened 2018-04-10
Small cleanups

adamretter:
Just a couple small cleanups.

### DIFF
--- a/src/org/exist/collections/CollectionConfigurationManager.java
+++ b/src/org/exist/collections/CollectionConfigurationManager.java
@@ -58,21 +58,21 @@ public class CollectionConfigurationManager implements BrokerPoolService {
 
     private static final Logger LOG = LogManager.getLogger(CollectionConfigurationManager.class);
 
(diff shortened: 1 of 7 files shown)


src/org/exist/collections/CollectionConfigurationManager.java
```diff
@@ -200,13 +201,11 @@ public void testConfiguration(DBBroker broker, String config) throws CollectionC
...
      * the collection hierarchy for available configurations.
-     * 
-     * @param broker
+     *
      * @param collection
      * @return The collection configuration
-     * @throws CollectionConfigurationException
      */
-    protected CollectionConfiguration getConfiguration(DBBroker broker, Collection collection) throws CollectionConfigurationException {
+    protected @Nullable CollectionConfiguration getConfiguration(final Collection collection) {
```
shabanovd: when it can be `Nullable`?

adamretter: Right you are :-)

PR #1811, opened 2018-04-10
Fix an XML-RPC API regression with uploading binary files

adamretter:
Another... oops!

### DIFF
--- a/src/org/exist/xmldb/RemoteCollection.java
+++ b/src/org/exist/xmldb/RemoteCollection.java
@@ -64,9 +64,9 @@ public class RemoteCollection extends AbstractRemote implements EXistCollection 
 
     // Max size of a resource to be send to the server.
     // If the resource exceeds this limit, the data is split into
-    // junks and uploaded to the server via the update() call
+    // chunks and uploaded to the server via the update() call
     private static final int MAX_CHUNK_LENGTH = 512 * 1024; //512KB
-    private static final int MAX_UPLOAD_CHUNK = 10 * 1024 * 1024; //10 MB
+    public static final int MAX_UPLOAD_CHUNK = 10 * 1024 * 1024; //10 MB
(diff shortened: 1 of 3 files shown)


src/org/exist/xmldb/RemoteCollection.java
```diff
@@ -64,9 +64,9 @@
 
     // Max size of a resource to be send to the server.
     // If the resource exceeds this limit, the data is split into
-    // junks and uploaded to the server via the update() call
+    // chunks and uploaded to the server via the update() call
```
joewiz: ;)

joewiz:
@dizzzz Thanks! I'm curious, were the CI failures unrelated to the PR?

adamretter:
@joewiz regards the performance the WevDav and XML-RPC Apis take very different approaches.

The WevDav Api is not chunked, and the XML-RPC Api adds one or two layers of compression to try and reduce network transfer size. Not to mention that it uses an XML envelope for the RPC and that binaries have to be base 64 encoded and decoded within that.  I don't have much time right now to improve the  the XML RPC performance. IMHO we would be better replacing the XML-RPC Api in the longer term with protobuf or similar

joewiz:
@adamretter How interesting, I hadn’t realized that about any of these inherent differences in the protocols.

dizzzz:
Deeply sorry, I somehow pulled in this PR without checking the whole PR; I thought that it was only a PR from joe with a typo fix. My mobile app made it look this way.

For now, @adamretter the 'develop' seems to be broken.......

(sorry again)

dizzzz:
@joewiz I saw a very slow "remote restore" process last week (using the java client for a restore). Probably related?

dizzzz:
after match: the change looks ok to me. Esp. the while loop looks better

PR #1817, opened 2018-04-13
XQueryURLRewrite should select closest controller in hierarchy

wolfgangmm:
Fixes an issue introduced in ec96c07b33077ccaf88a5f61b2edbc04fa7c7c85. XQueryURLRewrite should select the closest `controller.xql` it finds in the collection hierarchy. If `/db/apps/foo/baz/nested` does not have a controller, the one in `/db/apps/foo/baz` should be returned and so on.

### DIFF
--- a/src/org/exist/http/urlrewrite/XQueryURLRewrite.java
+++ b/src/org/exist/http/urlrewrite/XQueryURLRewrite.java
@@ -775,15 +775,14 @@ public class XQueryURLRewrite extends HttpServlet {
     private @Nullable
     DocumentImpl findDbControllerXql(final DBBroker broker, final XmldbURI collectionUri, final String[] pathComponents) {
         Collection collection = null;
+        DocumentImpl controller = null;
         try {
             collection = broker.openCollection(collectionUri, LockMode.READ_LOCK);
             if (collection == null) {
                 return null;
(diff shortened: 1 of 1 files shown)


wolfgangmm:
To reproduce: create a collection `/db/apps/controller-test` and a subcollection `/db/apps/controller-test/nested`. Store a `controller.xql` into `/db/apps/controller-test` with following content:

```xquery
xquery version "3.1";

<p>Controller response</p>
```

Upload an arbitary XML resource into the nested collection, e.g. `samples/shakespeare/hamlet.xml`. Access http://localhost:8080/exist/apps/controller-test/nested/hamlet.xml in the browser. You should always see "Controller response" and NOT the resource as the controller should intercept any request to paths below it.

I have no idea what TCE is. Feel free to change the fix accordingly.

adamretter:
Subsumed by https://github.com/eXist-db/exist/pull/1826

adamretter:
@dizzzz "include or absorb (something) in something else"

dizzzz:
so.... #1826 should be taken and this one not? or both in a specific order? or....

adamretter:
@dizzzz this  one should be ignored and the other one merged

PR #1818, opened 2018-04-13
Fix java.lang.IllegalArgumentException thrown by FastByteArrayOutputStream

wolfgangmm:
Fix exception in compression module: size of binary resource might be reported as -1 if unknown and should thus not be passed on the FastByteArrayOutputStream.


extensions/modules/src/org/exist/xquery/modules/compression/AbstractCompressFunction.java
```diff
@@ -432,7 +432,7 @@ private void compressResource(OutputStream os, DocumentImpl doc, boolean useHier
 		} else if (doc.getResourceType() == DocumentImpl.BINARY_FILE) {
 			// binary file
             try (final InputStream is = context.getBroker().getBinaryResource((BinaryDocument)doc);
-                 final FastByteArrayOutputStream baos = new FastByteArrayOutputStream((int)doc.getContentLength())) {
+                 final FastByteArrayOutputStream baos = new FastByteArrayOutputStream()) {
```
adamretter: Would you consider `doc.getContentLength() == -1 ? 1024 : (int)doc.getContentLength()` instead please

PR #1826, opened 2018-04-14
Fix regression with URL Rewrite controller selection

adamretter:
@wolfgangmm I added a Unit Test for this as well to prevent future regressions. Unfortunately the current test fails.

It seems that when URL Rewriting is run from JUnit, it seems to have some sort of static rewrite rule and so doesn't bother looking for a `controller.xql`. I could not figure out why that is. Perhaps you can help there? See `org.exist.http.urlrewrite.URLRewritingTest`.

If I test by starting eXist-db server and doing the same, then all passes just fine. So something weird happens with URL rewriting under JUnit.


wolfgangmm:
For me the test already fails in `URLRewritingTests#setup`, so I couldn't run it.

The only guess I would have for URL rewriting not applied during junit tests would be that `webapp/WEB-INF/controller-config.xml` is not found?

adamretter:
@wolfgangmm  no the controller-config.xml is still found and loaded. You didn't tell me what the error you get is the setup is, can you please share as that would fine here

test/src/org/exist/http/urlrewrite/URLRewritingTest.java
```diff
@@ -0,0 +1,107 @@
...
+        final String testDocument = "<hello>world</hello>";
+
+        HttpResponse response = executor.execute(Request
+                .Put(docUri)
+                .bodyString(testDocument, ContentType.APPLICATION_XML)
+        ).returnResponse();
+        assertEquals(HttpStatus.SC_CREATED, response.getStatusLine().getStatusCode());
+
+        response = executor.execute(Request
+                .Get(docUri)
```
wolfgangmm: URL rewriting will not be triggered if you go through the REST interface, so this line would need to request a path below `/exist/apps`:

```java
.Get(getServerUri() + "/apps/controller-test/nested/test.xml")
```

adamretter:
@wolfgangmm Okay should be fixed now.

PR #1830, opened 2018-04-15
Fix link to "quick start" docs

joewiz:
### Description:

Spotted a typo.

### Reference:

- The URL http://exist-db.org/quickstart.html returns a 404
- The correct URL is https://exist-db.org/exist/apps/doc/quickstart.xml

### Type of tests:

n/a

### DIFF
--- a/README
+++ b/README
@@ -3,7 +3,7 @@ eXist Native XML Database
 =========================
 
 For instructions on how to install and start eXist, please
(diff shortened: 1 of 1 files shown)


joewiz:
@duncdrum https://exist-db.org/exist/apps/doc/basic-installation returns a 404, so the link in this PR is the correct link for now. When exist-db.org is updated with the new version of the documentation, we'll need another PR for your change then.

PR #1838, opened 2018-04-18
Tests which show database integrity corruptions

adamretter:
Two test suites for database corruption, one for XML documents and one for Binary documents. Both test suites run the same tests but for the respective document type.

Currently there are:

* **4** failing test cases for Binary documents.
* **6** failing test cases for XML documents.

Each test case is executed on a clean instance of the database to ensure there are no "knock-on" effects from one test case to another.

Perhaps the most revealing of the test cases is `AbstractRecoveryTest#storeAndLoad()`.

I have previously fixed and can provide a PR for the binary issues!
Unfortunately the XML corruptions are much more serious, and difficult to understand, as they appear to be between eXist-db's BTree's and the Journal recovery.

I previously raised the issue of these corruptions in July 2017, but they have received little attention. I am opening this issue to make the tests more accessible and to increase visibility.

### DIFF
--- a/build/scripts/junit.xml
+++ b/build/scripts/junit.xml
@@ -311,6 +311,8 @@
             <formatter type="plain"/>
             <formatter type="xml"/>
             <test name="org.exist.storage.AllStorageTests" todir="${junit.reports.dat}"/>
+            <test name="org.exist.storage.RecoverBinaryTest" todir="${junit.reports.dat}"/>
+            <test name="org.exist.storage.RecoverXmlTest" todir="${junit.reports.dat}"/>
         </junit>
     </target>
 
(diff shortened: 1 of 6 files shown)


dizzzz:
impressive

wolfgangmm:
There seems to be a bug in the test: looking at AbstractRecoverTest.storeAndLoad, it deletes the temporary data dir via existEmbeddedServer.restart(), so recovery doesn't find a transaction log to recover from.

adamretter:
@wolfgangmm Right you are. I fixed that and just rebased. Still failures, but less of them now.

duncdrum:
needs a rebase

dizzzz:
Needs to be rebased

adamretter:
Closed by https://github.com/eXist-db/exist/pull/2241

PR #1857, opened 2018-05-05
Fixup PR-1679

adamretter:
Enhances and fixes some shortcoming of #1679

Closes #1679

### DIFF
--- a/src/org/exist/storage/DBBroker.java
+++ b/src/org/exist/storage/DBBroker.java
@@ -116,6 +116,8 @@ public abstract class DBBroker extends Observable implements AutoCloseable {
 
     protected IndexController indexController;
 
+    long config_timestamp = 0;
+
     public DBBroker(final BrokerPool pool, final Configuration config) {
         this.config = config;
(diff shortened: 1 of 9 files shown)


adamretter:
This has a small API modification around initialising DBBroker from BrokerPool. However, as I don't think anyone would ever directly initialise a DBBroker instance (i.e. they are always taken from BrokerPool) we should be okay to do this before a 5.0.0. Arguments anyone?

src/org/exist/storage/DBBroker.java
```diff
@@ -123,11 +124,27 @@ public DBBroker(final BrokerPool pool, final Configuration config) {
...
+         * has been called since the previous lease of this broker.
+         */
+        loadIndexModules();
+    }
+
+    /**
+     * Loads the index modules via an IndexController
+     */
+    protected void loadIndexModules() {
+        indexController.setIfExpiredOrNull(getBrokerPool().getIndexManager().getConfigurationTimestamp(), () -> new IndexController(this));
```
shabanovd: this looks as over-engineering to me because broker have `single thread` contract

dizzzz:
I agree, @wolfgangmm ? @ljo ?

adamretter:
@dizzzz @shabanovd Are we good to merge now?

PR #1858, opened 2018-05-06
Fixup PR-1056

adamretter:
Closes https://github.com/eXist-db/exist/pull/1056

**NOTE**: These changes affect the format of the `/db/system/security/config.xml`. I think the changes are forwards compatible i.e. old backups should continue to work with eXist-db. However newer backups once this PR is applied will likely not work with older versions of eXist-db.

**NOTE**: This PR includes an API change where I have corrected a typo and spelling of an existing function name.

As such this should be merged as part of the work for eXist-db 5.0.

### DIFF
--- a/src/org/exist/security/AbstractRealm.java
+++ b/src/org/exist/security/AbstractRealm.java
@@ -61,10 +61,8 @@ public abstract class AbstractRealm implements Realm, Configurable {
 
     protected final PrincipalDbByName<Account> usersByName = new PrincipalDbByName<>();
     protected final PrincipalDbByName<Group> groupsByName = new PrincipalDbByName<>();
-    
 
-    private SecurityManagerImpl sm;
-
+    private final SecurityManager sm;
     protected Configuration configuration;
 
     protected Collection collectionRealm = null;
@@ -73,7 +71,7 @@ public abstract class AbstractRealm implements Realm, Configurable {
     protected Collection collectionRemovedAccounts = null;
     protected Collection collectionRemovedGroups = null;
 	
-    public AbstractRealm(SecurityManagerImpl sm, Configuration config) {
+    public AbstractRealm(final SecurityManager sm, final Configuration config) {
         this.sm = sm;
         this.configuration = Configurator.configure(this, config);
     }
@@ -84,7 +82,7 @@ public abstract class AbstractRealm implements Realm, Configurable {
     }
 
     @Override
-    public SecurityManagerImpl getSecurityManager() {
+    public SecurityManager getSecurityManager() {
         return sm;
     }
     
--- a/src/org/exist/security/AbstractRealm.java
+++ b/src/org/exist/security/AbstractRealm.java
@@ -1,23 +1,21 @@
 /*
  *  eXist Open Source Native XML Database
- *  Copyright (C) 2010-2011 The eXist Project
+ *  Copyright (C) 2001-2016 The eXist Project
  *  http://exist-db.org
- *  
+ *
  *  This program is free software; you can redistribute it and/or
  *  modify it under the terms of the GNU Lesser General Public License
  *  as published by the Free Software Foundation; either version 2
  *  of the License, or (at your option) any later version.
- *  
+ *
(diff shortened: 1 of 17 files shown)


src/org/exist/security/SecurityManager.java
```diff
@@ -108,10 +112,6 @@
    @Deprecated
    Subject getSubjectBySessionId(String sessionid);
 
-   void addGroup(int id, Group group);
```
shabanovd: why is it require methods renaming?

test/src/org/exist/security/internal/BackupRestoreSecurityPrincipalsTest.java
```diff
@@ -0,0 +1,291 @@
+/*
```
shabanovd: that was backup test that also check security related staff. why did you move it out of backup related tests package?

src/org/exist/security/SecurityManager.java
adamretter: Good question! I don't know - it was a change that you made in your original PR - https://github.com/eXist-db/exist/pull/1056/files#diff-1ae708c6e5ca91cc9ddcfa7181910ca4L111

test/src/org/exist/security/internal/BackupRestoreSecurityPrincipalsTest.java
adamretter: Good question! Again, I don't know - as it was a change that you made in your original PR - https://github.com/eXist-db/exist/pull/1056/files#diff-14915bb0f81a3900e16d841b355f91b9

Although after some consideration, this does now feel like the correct place for it after some.

src/org/exist/security/SecurityManager.java
shabanovd: it must stay `addGroup`

test/src/org/exist/security/internal/BackupRestoreSecurityPrincipalsTest.java
shabanovd: right. guess, I had exactly same reasoning that time too

src/org/exist/security/SecurityManager.java
adamretter: Why?

shabanovd: because it's api change

adamretter: That's fine. This PR already changed the format of the backup, so I consider that it already has API changes and therefore will go in 5.0.0

src/org/exist/security/internal/SecurityManagerImpl.java
```diff
@@ -403,7 +395,7 @@ public Subject authenticate(final String username, final Object credentials) thr
             }
 
             final Subject subject = sessions.read(db1 -> {
-                final Session session = db1.get(credentials);
+                final Session session = db1.get(credentials.toString());
```
shabanovd: this is not correct because `credentials` can be `byte[]`

src/org/exist/security/SecurityManager.java
shabanovd: anyway, is it possible to rename it back ... I don't want to deal with another api "renaming"

src/org/exist/security/internal/SecurityManagerImpl.java
shabanovd: k

src/org/exist/security/AbstractRealm.java
```diff
@@ -65,7 +63,7 @@
     protected final PrincipalDbByName<Group> groupsByName = new PrincipalDbByName<>();
     
 
-    private SecurityManager sm;
+    private SecurityManagerImpl sm;
```
dizzzz: Why the Imp here and not the interface?

adamretter: That is one of the things I corrected from the original PR. Check the final state of the PR

dizzzz: Yes sorry.....

PR #1871, opened 2018-05-12
Tests for Journal of binary documents

adamretter:
Just tests...

@wolfgangmm this might be useful for you when trying to figure out the XML Document Recovery Corruptions. You could implement a `JournalXMLTest extends AbstractJournalTest` in s similar vein to how `test/src/org/exist/storage/journal/JournalBinaryTest.java` works.


dizzzz:
The PR shows issues....  needs rework?

adamretter:
@dizzzz Fixed. Sorry about that, junit.xml was executing the abstract class too which is not possible. I added it to the exclusions now and all works well.

PR #1880, opened 2018-05-21
Enabled JDK 9 on AppVeyor Ubuntu

adamretter:
Okay this one is now fixed and ready to land.


dizzzz:
For me, I actually rely on Travis, the AppVeyor is a nice and useful addition. For sure, if there are issues on windows, AppVeyor will show, so we can fix these. We were blind for these in the past.

PR #1883, opened 2018-05-25
Fix NPE when uploading (large) document from Java client

dizzzz:
Make a stream out of a string to avoid an NPE when uploading a file with the Java client. 
ToDo: add test

### Description:
See original issue : A NPE occurs when a (large) *XML* document is uploaded into the database.

### Type of tests:
junit test (to be added)


dizzzz:
Test added @adamretter

Issue #1878, opened 2018-05-18
Bug in day of week with format-date functions 


joewiz:
### What is the problem

In the current release and the develop branch, eXist returns the wrong day of the week when using the `[F]` picture string to output the day of the week with `format-date()` or `format-dateTime()`. It consistently returns the day of the week *one day after* what it should be returning. So, instead of saying that today is Friday, it says Saturday.

### What did you expect

I expected eXist to return the correct day of the week.

### Describe how to reproduce or add a test

The problem is evident in my tests with the eXist 4.1.0 release (919514c68) and later (including the current develop HEAD, f53dbceae), but not in the eXist 4.0.0 release (cc66ebc). When I copy https://github.com/eXist-db/exist/blob/develop/test/src/xquery/dates/format-dates.xql into my database, I get the failures shown below. A quick test anyone can run is this:

```xquery
"Is today really " || format-date(current-date(), "[FNn]") || "?"
```

Result of format-dates.xql on my machine (identical results with the 4.1.0 release and current develop):

```xml
<testsuites>
    <testsuite package="http://exist-db.org/xquery/test/format-dates" timestamp="2018-05-18T13:11:53.493-04:00" failures="14" pending="10" tests="88" time="PT0.268S">
        <testcase name="date-format-in-format-time" class="fd:date-format-in-format-time"/>
        <testcase name="format-date-Z" class="fd:format-date-Z"/>
        <testcase name="format-date-z" class="fd:format-date-z"/>
        <testcase name="format-dateTime" class="fd:format-dateTime">
            <failure message="assertEquals failed." type="failure-error-code-1">11:14 pm on Tuesday, June 26th, 2012</failure>
            <output>11:14 pm on Wednesday, June 26th, 2012</output>
        </testcase>
        <testcase name="format-dateTime-Z" class="fd:format-dateTime-Z"/>
        <testcase name="format-dateTime-Z-negative-double-digit" class="fd:format-dateTime-Z-negative-double-digit"/>
        <testcase name="format-dateTime-Z-negative-single-digit" class="fd:format-dateTime-Z-negative-single-digit"/>
        <testcase name="format-dateTime-Z-positive-double-digit" class="fd:format-dateTime-Z-positive-double-digit"/>
        <testcase name="format-dateTime-Z-positive-single-digit" class="fd:format-dateTime-Z-positive-single-digit"/>
        <testcase name="format-dateTime-Z-zero" class="fd:format-dateTime-Z-zero"/>
[...]

PR #1885, opened 2018-05-25
Written day of week was off by +1

adamretter:
Closes https://github.com/eXist-db/exist/issues/1878

Corrects a small issue when converting between Calendar bases.

### DIFF
--- a/src/org/exist/xquery/functions/fn/FnFormatDates.java
+++ b/src/org/exist/xquery/functions/fn/FnFormatDates.java
@@ -139,7 +139,7 @@ public class FnFormatDates extends BasicFunction {
         RETURN
     );
 
-    private static Pattern componentPattern = Pattern.compile("([YMDdWwFHhmsfZzPCE])\\s*(.*)");
+    private static final Pattern componentPattern = Pattern.compile("([YMDdWwFHhmsfZzPCE])\\s*(.*)");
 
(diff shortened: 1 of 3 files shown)


adamretter:
Yes this will still say Monday for you.

joewiz:
I just committed fixes to the failures - which were all in the tests, not in the fix (which is correct). 

The test suite absolutely should have failed on this PR, as demonstrated by https://github.com/adamretter/exist/blob/f2d40c60dcf88ff95698c3b9bddaee400877d7f4/test/src/xquery/dates/format-dates.xql#L572. The typo here - `date-str` instead of `$date-str` - could never have worked. 

I am puzzled how CI tests were passing on this PR. Are the CI servers somehow skipping this tests? If so, are they also skipping other XQuery tests?

adamretter:
@joewiz Travis has a limitation on how long a build can take. In the past I think that was 10 minutes, so we had to reduce the amount of tests we were executing. To do that rather than execute `build.sh test` we would execute `build.sh travis` which just ran a subset of the "most important" tests.

I think the limits have been increased, and so I have opened a PR here to see if we can now run the entire test suite on Travis and AppVeyor - https://github.com/eXist-db/exist/pull/1887

adamretter:
Thanks @dizzzz

PR #1886, opened 2018-05-26
Added %test:assumeInternetAccess to XQSuite

adamretter:
Some XQuery tests previously depended on various websites. Unfortunately that then causes those tests to fail when you don't have any internet access. I have now implemented an annotation `%test:assumeInternetAccess` which allows you to make sure that your test is only run when Internet Access is available.

### DIFF
--- a/src/org/exist/xquery/lib/xqsuite/xqsuite.xql
+++ b/src/org/exist/xquery/lib/xqsuite/xqsuite.xql
@@ -18,6 +18,7 @@ xquery version "3.0";
  :)
 module namespace test="http://exist-db.org/xquery/xqsuite";
 
+import module namespace http = "http://expath.org/ns/http-client";
 import module namespace map = "http://www.w3.org/2005/xpath-functions/map";
 import module namespace util = "http://exist-db.org/xquery/util";
 import module namespace system = "http://exist-db.org/xquery/system";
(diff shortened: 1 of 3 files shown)


duncdrum:
Please add a note about this annotation to the docs before merging.

adamretter:
Documentation label is here - https://github.com/eXist-db/documentation/pull/245

duncdrum:
all good from me then

adamretter:
@joewiz happy for someone to send a follow up PR for that. I was just modelling it so far on the JUnit approach, so that we can use it from XSuite.

joewiz:
@adamretter I'll be happy to do that. Which approach makes the most sense? (1) Adding failed assumptions to the existing count of pending results, or (2) Adding a new attribute reporting just the failed assumptions?

duncdrum:
@joewiz I would prefer a more neutral term `skipped` or s.th. `failed` should be reserved for test that run but fail, imv.

joewiz:
@duncdrum Oh, I wasn't saying we should add these to `failed` - I was asking if they should be added to the count of `pending` tests or if it's best to add a new category, `assumptions` - namely, here: https://github.com/eXist-db/exist/blob/develop/src/org/exist/xquery/lib/xqsuite/xqsuite.xql#L88-L92.

adamretter:
@joewiz, I am not certain really. I just followed the JUnit approach where assumptions that do not hold, are effectively reported as Ignore/Skipped tests

joewiz:
@adamretter Interesting, I'll have a look at the JUnit documentation to see where these might best be placed, if they need to be counted/reported at all.

An exemplary new feature - code with tests and documentation!

PR #1896, opened 2018-05-31
Changes to APIs for improved transaction support.

adamretter:
This PR forms the fundamental basis of improved transaction support as used in Granite. Whilst granite adds many more changes to transactions than just this, these changes have some direct benefit to eXist-db:

* If an existing transaction is in progress for a single thread, there is no need to open a second transaction. This greatly reduces the amount of transactions in-flight for a single thread. This has advantages both in terms of a reduction in size and processing overhead for the journal and also a reduction in in-memory Txn objects.

My motivation for offering this PR is not only to enhance eXist-db, but also to reduce the API differences between eXist-db and the version of eXist-db used by Granite.

There are also many changes we have made to eXist-db APIs for Granite so that transactions are explicitly passed to every function which performs a read or write against the database. With such changes it becomes easier for eXist-db to implement transaction isolation and exploit facilities such as MVCC. Whilst this does not provide an immediate win for eXist-db, it does open the APIs to many future improvements. If there is interest from @wolfgangmm I would be prepared to also back-port those API changes to this PR also.


test/src/org/exist/xquery/functions/validate/JingOnvdlTest.java
```diff
@@ -103,6 +104,7 @@ public void onvdl_valid() throws XPathException, IOException, XpathException, SA
         executeAndEvaluate(query,"valid");
     }
 
+    @Ignore("Can cause a deadlock in txn-thread branch")
```
dizzzz: :-(

extensions/fluent/src/org/exist/fluent/Database.java
```diff
@@ -104,20 +104,17 @@ static void configureRootCollection(Path configFile) {
...
 		try {
-			pool.getConfigurationManager().addConfiguration(tx.tx, tx.broker, tx.broker.getCollection(XmldbURI.ROOT_COLLECTION_URI), configXml.toString());
-			tx.commit();
-			DBBroker broker = db.acquireBroker();
-			try {
-				broker.reindexCollection(XmldbURI.ROOT_COLLECTION_URI);
-			} finally {
-				db.releaseBroker(broker);
-			}
+            try(final Transaction tx = db.requireTransactionWithBroker()) {
```
dizzzz: indenting is not OK here? [readability]

extensions/fluent/src/org/exist/fluent/Folder.java
```diff
@@ -839,7 +840,7 @@ void acquire(LockMode _lockMode, DBBroker _broker) {
 	
 	void release() {
```
dizzzz: this method should be formatted

extensions/fluent/src/org/exist/fluent/Node.java
```diff
@@ -199,8 +199,7 @@ public XMLDocument document() {
 			final StoredNode node = (StoredNode) getDOMNode();
 			return new ElementBuilder<Node>(namespaceBindings, true, new ElementBuilder.CompletedCallback<Node>() {
 				public Node completed(org.w3c.dom.Node[] nodes) {
-					Transaction tx = db.requireTransactionWithBroker();
-					try {
+					try(final Transaction tx = db.requireTransactionWithBroker();) {
```
dizzzz: `;` not needed here

test/src/org/exist/xquery/functions/validate/JingOnvdlTest.java
adamretter: Okay so I removed those @Ignores now as we fixed it here - https://github.com/eXist-db/exist/pull/1908

extensions/fluent/src/org/exist/fluent/Database.java
adamretter: It was a tabs vs spaces thing. I have fixed it.

extensions/fluent/src/org/exist/fluent/Folder.java
adamretter: Okay I reformatted it.

extensions/fluent/src/org/exist/fluent/Node.java
adamretter: Okay I removed it.

adamretter:
@dizzzz Thanks for your review. I have now rebased and pushed the updated PR.

PR #1897, opened 2018-06-01
Security fixes to mitigate Zip/Tar file exit attacks when decompressing archives.

adamretter:
Firstly I would like to thank the "**[Snyk Security Research Team](http://snyk.io)**" for reporting this issue to us. They provided both the insight to the problem and details of a suggested fix.

This PR also adds some additional functions to the Compression Module:

1. `compression:no-filter#2` and `compression:no-filter#3`. These functions can be used when the user does not want to filter anything from the archive they are uncompressing.

2. `compression:fs-store-entry3#1`, `compression:fs-store-entry4#1`, `compression:fs-store-entry3#1`, and `compression:db-store-entry4#1`. These functions can be used when the user wants to write the entries in the archive to either the filesystem or the database. Both of these functions offer in-built protection against the archive exit attack.

3. `compression:unzip#3` and `compression:untar3`, which are simplified versions of the existing unzip and untar functions for the most common use-cases.

**NOTE** users should be very careful when writing their own `$entry-data` functions for `compression:unzip` or `compression:untar` if they are writing the file entries to persistent storage using the name of the entry from the archive; Consideration should be given to employing further exit attack protections of their own.

### DIFF
--- a/extensions/modules/src/org/exist/xquery/modules/compression/AbstractExtractFunction.java
+++ b/extensions/modules/src/org/exist/xquery/modules/compression/AbstractExtractFunction.java
@@ -42,15 +42,7 @@ import org.exist.xquery.XPathException;
 import org.exist.xquery.XQueryContext;
 import org.exist.xquery.functions.xmldb.XMLDBAbstractCollectionManipulator;
 import org.exist.xquery.modules.ModuleUtils;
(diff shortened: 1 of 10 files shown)


dizzzz:
@wolfgangmm @ljo I need some help with the review of the PR. It looks OK for me, but there are a lot of changes here....

PR #1900, opened 2018-06-03
Fix an issue with recycling of temp files.

adamretter:
Memory Mapped files cannot be reused for normal I/O of filesystem operations on Microsoft Windows.

### DIFF
--- a/src/org/exist/util/io/TemporaryFileManager.java
+++ b/src/org/exist/util/io/TemporaryFileManager.java
@@ -27,10 +27,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package org.exist.util.io;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/org/exist/util/io/TemporaryFileManager.java
```diff
@@ -69,64 +92,46 @@ private TemporaryFileManager() {
...
+        will be recovered on restart by cleanupOldTempFolders
+         */
         tmpFolder.toFile().deleteOnExit();
 
         LOG.info("Temporary folder is: " + tmpFolder.toAbsolutePath().toString());
     }
 
     public final Path getTemporaryFile() throws IOException {
-        Path tempFile = available.poll();
+        final Path tempFile = Files.createTempFile(tmpFolder, FILE_PREFIX + '-', ".tmp");
```
dizzzz: why the dash?

PR #1903, opened 2018-06-03
Compressor now uses GZip encoding

adamretter:
Previously Zip file encoding was used, which has quite some overhead - due to having a directory layout format inside the archive. Now the just the data stream is compressed with GZip, which has a much smaller overhead as it is not an archive format.

NOTE: This introduces a non-backwards compatible change to the XML-RPC API (and by extension the XML:DB API and Java Admin Client).


duncdrum:
@adam i feel we should we add a note about this to [the docs](http://www.exist-db.org/exist/apps/doc/devguide_xmlrpc.xml?field=all&id=D3.3.2), despite there not being one right now

adamretter:
@duncdrum it at least needs to go in the release notes as a non-backwards compatible change

adamretter:
@duncdrum It is not configurable. It used to be zip, now it is gzip.

duncdrum:
@adamretter I understand that; but users can call `upload `vs `uploadCompressed` and the second doesn't appear in the docs about what users can or can't call.

src/org/exist/util/Compressor.java
```diff
@@ -1,94 +1,112 @@
...
-        try (final FastByteArrayOutputStream baos = new FastByteArrayOutputStream(length);
-                final ZipOutputStream gzos = new ZipOutputStream(baos)) {
-            gzos.setMethod(ZipOutputStream.DEFLATED);
-            gzos.putNextEntry(new ZipEntry(length + ""));
-            gzos.write(whatToCompress, 0, length);
-            gzos.closeEntry();
+    public static byte[] compress(final byte[] buf, final int len) throws IOException {
+        try (final FastByteArrayOutputStream baos = new FastByteArrayOutputStream(len);
+                final GZIPOutputStream gzos = new GZIPOutputStream(baos)) {
+            gzos.write(buf, 0, len);
```
dizzzz: `gzos.flush()` is required here, the `close()` will not do that, strangely.

src/org/exist/util/Compressor.java
```diff
@@ -1,94 +1,112 @@
...
+     * @exception IOException if an error occurs
+     */
+    public static void uncompress(final byte[] buf, final OutputStream os) throws IOException {
+        try (final FastByteArrayInputStream bais = new FastByteArrayInputStream(buf);
+             final InputStream gzis = new GZIPInputStream(bais)) {
+            final byte[] tmp = new byte[4096];
+            int read;
+            while ((read = gzis.read(tmp)) != -1) {
+                os.write(tmp, 0, read);
+            }
```
dizzzz: flush buffer here?

src/org/exist/util/Compressor.java
adamretter: Are you certain? Looking at the code, it appears to me that `finish()` (below) will do a very similar thing to a `flush()`.

src/org/exist/util/Compressor.java
adamretter: I don't think it is needed.

src/org/exist/util/Compressor.java
dizzzz: ah crap, sorry, i meant `finish()`. I guess it is the heat.

src/org/exist/util/Compressor.java
dizzzz: Hmmmmmmm ok

PR #1908, opened 2018-06-04
Fixes a deadlock that can occur with XML validation

adamretter:
Closes https://github.com/eXist-db/exist/issues/558

Previously to obtain an InputStream or OutputStream against any resource in the database, we would buffer in-memory between by using a thread to pipe from Input to Output or vice-versa. Unfortunately this thread also needed to take locks on the database resources, which could cause conflicts with other concurrent actions.

We now instead just buffer to a Temporary file on disk, which means that we don't need any additional threads, which means that there are less chance of deadlocks. In fact this has the advantage, that for binary files in the db, we don't even need a temporary file, we can just stream the file directly.

The new approach seems to greatly simplify the implementation IMHO.


conf.xml.tmpl
```diff
@@ -175,13 +175,11 @@
                     Trigger for registering eXists XML:DB URL handler with Java
                 -->
                 <trigger class="org.exist.protocolhandler.URLStreamHandlerStartupTrigger">
-                    <!-- Keep stream's data in memory -->
+                    <!-- Keeps stream data on disk (temporary files are used for XML documents) -->
```
dizzzz: one of the elements is not commented out. Is this OK?

adamretter: Yup.

Issue #1724, opened 2018-02-08
Namespace serialization bug over WebDAV or XML-RPC after XQuery Update


joewiz:
### What is the problem

After using XQuery Update to insert an element (a) in the empty namespace and (b) with an `xml:id` attribute into a document with a non-empty namespace, the empty namespace of the inserted element is not shown when reading the resulting document over WebDAV or XML-RPC. However, when reading the document via the REST interface, the `doc()` function, or eXide (via its load.xql module), the inserted empty namespace is preserved. 

This phenomenon is only evident when (a) and (b) are in place. In other words, the problem is not evident if you insert an element in a non-empty namespace, or if you insert an element without the `xml:id` attribute. Also, the phenomenon is not evident if you store a document with the same structure directly and without the use of XQuery update.

### What did you expect

I expected node identity to be preserved regardless of the interface through which the node is serialized or method with which it was stored into the database.

### Describe how to reproduce or add a test

Run the following query in eXide or Java admin client:

```xquery

xquery version "3.1";

declare namespace foo="foo";

let $in-memory := 
    <root xmlns="foo">
        <x/>
    </root> 
let $store := xmldb:store("/db", "test.xml", $in-memory)
let $on-disk := doc("/db/test.xml")
let $new-node := <x xml:id="aargh"/>
let $update := update insert $new-node into $on-disk/foo:root
return
    (
        $in-memory,
        $on-disk
    )
```

The 2nd item returned is as follows:

```xml
<root xmlns="foo">
    <x/>
    <x xmlns="" xml:id="aargh"/>
</root>
```

Then retrieve the file via REST interface, at <http://localhost:8080/exist/rest/db/test.xml>, and the result should be identical.

Then retrieve the file via XML-RPC interface (e.g., Java admin client or oXygen data source explorer) or WebDAV (e.g., Transmit or oXygen WebDAV data source), and the inserted element has lost its empty namespace:

```xml
<root xmlns="foo">
    <x/>
    <x xml:id="aargh"/>
</root>
```

This phenomenon is not evident if the `@xml:id` attribute is removed from the inserted element, if the inserted element is in a non-empty namespace, or if the document is stored whole via the following query:

```xquery
let $new-node := <x a="b"/>
```

```xquery
let $new-node := <x xmlns="bar"/>
```

```xquery
xmldb:store(
    '/db', 
    'test.xml', 
    <root xmlns="foo">
        <x/>
        <x xmlns="" xml:id="aargh"/>
    </root>
)
[...]

PR #1911, opened 2018-06-05
Fix empty namespace serialization issues

adamretter:
Closes https://github.com/eXist-db/exist/issues/1724

### DIFF
--- a/src/org/exist/storage/NativeBroker.java
+++ b/src/org/exist/storage/NativeBroker.java
@@ -2267,25 +2267,14 @@ public class NativeBroker extends DBBroker {
     @Override
     public void readBinaryResource(final BinaryDocument blob, final OutputStream os)
         throws IOException {
-        InputStream is = null;
-        try {
-            is = getBinaryResource(blob);
-            final byte[] buffer = new byte[BINARY_RESOURCE_BUF_SIZE];
-            int len;
-            while((len = is.read(buffer)) >= 0) {
-                os.write(buffer, 0, len);
-            }
-        } finally {
-            if(is != null) {
-                is.close();
-            }
-        }
+        final Path binFile = getBinaryFile(blob);
+        Files.copy(binFile, os);
     }
 
     @Override
     public long getBinaryResourceSize(final BinaryDocument blob)
         throws IOException {
-        final Path binFile = getCollectionFile(getFsDir(), blob.getURI(), false);
+        final Path binFile = getBinaryFile(blob);
         return Files.size(binFile);
     }
 
@@ -2297,7 +2286,7 @@ public class NativeBroker extends DBBroker {
     @Override
     public InputStream getBinaryResource(final BinaryDocument blob)
         throws IOException {
-        return Files.newInputStream(getCollectionFile(getFsDir(), blob.getURI(), false));
+        return Files.newInputStream(getBinaryFile(blob));
     }
 
     //TODO : consider a better cooperation with Collection -pb
--- a/src/org/exist/util/serializer/XMLWriter.java
+++ b/src/org/exist/util/serializer/XMLWriter.java
@@ -373,27 +373,40 @@ public class XMLWriter {
         }
     }
 
-    public void cdataSection(final char[] ch, final int start, final int len) throws TransformerException {
+    public void startCdataSection() throws TransformerException {
         if(tagIsOpen) {
             closeStartTag(false);
         }
-        
+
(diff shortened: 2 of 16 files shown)


joewiz:
@adamretter Thanks for working on this head scratcher!

I just checked out and built your branch:

```text
project.version=4.2.0-SNAPSHOT
project.built=201806051135
scm.branch=hotfix/empty-namespace-serialization
scm.revision=5ef8b7ac2
```

Unfortunately, when I try the test from my original issue, I am still seeing the same namespace serialization problem. In other words, when I open the resulting `/db/test.xml` via WebDAV or XML-RPC clients (oXygen WebDAV and Java Admin Client), respectively, the document is serialized incorrectly, as:

```xml
<root xmlns="foo">
    <x/>
    <x xml:id="aargh"/>
</root>
```

... instead of as:

```xml
<root xmlns="foo">
    <x/>
    <x xmlns="" xml:id="aargh"/>
</root>
```

... which is how the REST server serializes the document. Specifically, the problem is that `xmlns=""` is omitted from the 2nd `<x>` element when loading this document over WebDAV and XML-RPC.

adamretter:
@joewiz are you sure? I added your example withinor changes to SerialisationTest.java in the PR and that passes now with my changes!

src/org/exist/util/serializer/XMLWriter.java
```diff
@@ -373,47 +373,72 @@ public void comment(final CharSequence data) throws TransformerException {
         }
     }
 
-    public void cdataSection(final char[] ch, final int start, final int len) throws TransformerException {
+    public void startCdataSection() throws TransformerException {
```
dizzzz: should/could be private or package protected when required for unit tests

src/org/exist/util/serializer/XMLWriter.java
```diff
@@ -373,47 +373,72 @@ public void comment(final CharSequence data) throws TransformerException {
...
+
         try {
             writer.write("<![CDATA[");
-            writer.write(ch, start, len);
+        } catch(final IOException ioe) {
+            throw new TransformerException(ioe.getMessage(), ioe);
+        }
+    }
+
+    public void endCdataSection() throws TransformerException {
```
dizzzz: should/could be private or package protected when required for unit tests

src/org/exist/storage/NativeBroker.java
```diff
@@ -2297,7 +2286,7 @@ public Path getBinaryFile(final BinaryDocument blob) throws IOException {
     @Override
     public InputStream getBinaryResource(final BinaryDocument blob)
         throws IOException {
-        return Files.newInputStream(getCollectionFile(getFsDir(), blob.getURI(), false));
+        return Files.newInputStream(getBinaryFile(blob));
```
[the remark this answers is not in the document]
adamretter: Well actually that is not nessecarily the case, `Files.newInputStream` might return a buffered stream. Depends on the OS and JVM. Best to let them decide!

src/org/exist/util/serializer/XMLWriter.java
adamretter: Nope. It doesn't override anything.

src/org/exist/util/serializer/XMLWriter.java
adamretter: Nope it should be public.

src/org/exist/storage/NativeBroker.java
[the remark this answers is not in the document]
dizzzz: not according to the java doc ?

dizzzz:
waiting again for the slow AppVeyor

src/org/exist/storage/NativeBroker.java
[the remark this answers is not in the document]
adamretter: @dizzzz Wooow! Okay, I was totally mistaken there. Well there will be many places where this is the case in the codebase. I have not actually changed the buffering here AFAIK.

Perhaps we could cope with this in a future PR from someone... that could look at the uses of `Files.newInputStream` and add buffering where it makes sense for larger transfers?

adamretter:
So @joewiz @dizzzz. I am running the test suite locally, and if Travis-CI passes, I would ask that in this case we make an exception and merge without AppVeyor. There are no Windows specific changes or filesystem changes in this, and I would dearly like to see this go into the 4.2.0 release...

Issue #1714, opened 2018-02-04
Function signature mismatches in eXist vs. W3C F&O spec


joewiz:
### What is the problem

Besides [certain functions](http://exist-open.markmail.org/thread/67zw5ukqvzqaqcao), [expressions](https://www.w3.org/TR/xquery-31/#id-windows), [options](https://github.com/eXist-db/exist/issues/1483), etc. from the XQuery 3.0 and 3.1 specifications that eXist doesn't yet support, there are slight mismatches in the signatures of the functions that eXist does support from built-in function library when comparing eXist to the W3C XQuery 3.1 spec. @dizzzz created a test suite generator that pulls function signature information from the spec, and I created [a report](https://gist.github.com/joewiz/78b1827971a3c1c5dec8fc2d13dec8b2) that analyzes the results to highlight exactly where the differences are. An analysis of this report shows the following causes:

1. Cardinality mismatches (the bulk of mismatches, and the focus here). For example, `fn:starts-with()` should return `xs:boolean`, but eXist allows `xs:boolean?`. These are likely the result of changes to the spec since XQuery 1.0 (see changes between [1.0 and 3.0](https://www.w3.org/TR/xpath-functions-30/#back-compatibility) and [3.0 and 3.1](https://www.w3.org/TR/xpath-functions-31/#back-compatibility). Ideally, eXist would read the version declaration of the query and apply the correct cardinality checks. Does eXist support this type of analysis? If not, should we consider adopting 3.1's function signatures in a forthcoming major version, as 3.1 is the current recommendation? 

2. Insufficient detail reported by `inspect:inspect-function`. For example, the return for `fn:analyze-string` should be `element(fn:analyze-string-result)`, but eXist reports only `element()`. Similarly, the return for `fn:for-each-pair` should be `function(item(), item()) as item()*`, but eXist reports only `function(*)`. If eXist is applying these finer-grained type checks, should we consider updating the `inspect` module to report the checks and help users feel confident that eXist conforms to the spec? 

### What did you expect

I expected eXist to conform precisely to the signatures of functions it has implemented from the XQuery 3.1 spec.

### Describe how to reproduce or add a test

See the report at https://gist.github.com/joewiz/78b1827971a3c1c5dec8fc2d13dec8b2. Note that the results shown here assume that https://github.com/eXist-db/exist/pull/1712 and https://github.com/eXist-db/exist/pull/1713 have been incorporated. Those PRs were driven by eliminating large numbers of mismatches.

### Context information

[...]

PR #1918, opened 2018-06-07
[bugfix] index-of should return xs:integer*

joewiz:
### Description:

According to https://www.w3.org/TR/xpath-functions-31/#func-index-of, `fn:index-of` should return `xs:integer*` (`ZERO_OR_MORE`), but eXist had return `xs:integer?` (`ZERO_OR_ONE`). 

### Reference:

- XQuery F&O 1.0: https://www.w3.org/TR/xquery-operators/#func-index-of
- XQuery F&O 3.1 https://www.w3.org/TR/xpath-functions-31/#func-index-of
- Function mismatches in eXist: https://github.com/eXist-db/exist/issues/1714 and https://gist.github.com/joewiz/78b1827971a3c1c5dec8fc2d13dec8b2#file-signature-mismatches_results-xml-L351-L369
- exist-open thread: https://exist-open.markmail.org/thread/6h77llsoi76hyxgy

### Type of tests:

Forthcoming. As discussed with @wolfgangmm and @adamretter, the QT3 test suite will be integrated very soon.

### DIFF
--- a/src/org/exist/xquery/functions/fn/FunIndexOf.java
+++ b/src/org/exist/xquery/functions/fn/FunIndexOf.java
@@ -50,7 +50,7 @@ import org.exist.xquery.value.ValueSequence;
  */
 public class FunIndexOf extends BasicFunction {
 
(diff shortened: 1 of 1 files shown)


joewiz:
This is a port of https://github.com/eXist-db/exist/pull/1916 intended for the eXist 4.x.x branch of development.

Issue #1913, opened 2018-06-06
New issue with eXist 4.2.0: Restore indefinitely hangs with no error message


iampopuri:
### What is the problem 
> Problem: Restore indefinitely hangs with no error message
> Description: After importing very few files then it just hangs forever without any error message

### What did you expect
> Restore should continue till all files import is completed

### Describe how to reproduce or add a test
> Take small set of data backup from old eXist database and try to restore them via client or linux command line with ant script then it just hangs. The problem seems to be in RemoteCollection.java Line # 624. It might be going to indefinite while loop.
### Context information
Please always add the following information
- eXist-db version + Git Revision hash - eXist-db 4.2.0 
- Java version - 1.8.0_161-b14
- Operating system - (Windows 7, Linux) - 64 bit
- How is eXist-db installed? - JAR installer
- Any custom changes in e.g. conf.xml - No

![restore-hangs](https://user-images.githubusercontent.com/39523509/41048624-6d58762a-69af-11e8-8cc9-8244371c170b.jpg)

duncdrum:
phew I wonder if this is related to #1874 both times RemoteCollection seems to be throwing itself a party.

iampopuri:
Few more points from my colleague....might helpful for you.

Seems like some changes made to optimize the chunk size in the RemoveCollection.java are causing these issues: See below picture.

![code_bug](https://user-images.githubusercontent.com/39523509/41089585-7ae5cfa2-6a42-11e8-9418-ef02235b062c.jpg)

At line 602 the chunk size resolves to 0, therefore a database is created of 0 length.
When entering the wile loop at 612, the len will therefore always be 0 – as it never reads any bytes – and thus hangs indefinitely.

If we comment the chunk size changes and revert back to the original we get the following error at line 670 which seems to be a problem in the RpcConnection: 

![code_bug_1 jpg](https://user-images.githubusercontent.com/39523509/41089600-86bec108-6a42-11e8-8344-60c60d4e41f0.png)

09:16:47.070 [pool-3-thread-1] ERROR org.exist.backup.restore.RestoreHandler - networking error
org.xmldb.api.base.XMLDBException: networking error
       at org.exist.xmldb.RemoteCollection.uploadAndStore(RemoteCollection.java:704) ~[classes/:?]
       at org.exist.xmldb.RemoteCollection.storeResource(RemoteCollection.java:504) ~[classes/:?]
[...]

Issue #1874, opened 2018-05-14
Client unable to upload large XML files (4.1.0)


thammr:
When I try to upload a large XML file (>512k) using the client, I see the following error in the Java console:

Exception in thread "Thread-23" java.lang.NullPointerException
	at org.exist.xmldb.RemoteCollection.uploadAndStore(RemoteCollection.java:606)
	at org.exist.xmldb.RemoteCollection.storeResource(RemoteCollection.java:507)
	at org.exist.xmldb.RemoteCollection.storeResource(RemoteCollection.java:476)
	at org.exist.client.InteractiveClient.store(InteractiveClient.java:1763)
	at org.exist.client.InteractiveClient.parse(InteractiveClient.java:1673)
	at org.exist.client.ClientFrame$2.run(ClientFrame.java:980)

I tried to reconstruct the code to understand what the problem is.

It appears the res.getContent() returns a String (not a File or InputStream). Consequently the InputStream instance "is" is never instantiated and generates a null pointer exception.

This bug also causes restore to skip large files.

thammr:
Selecting the Add Document button on the toolbar gives the above error if the file is large. The upload dialog simply hangs.
Restoring a backup gives a "null" for each large file in the progress dialog. Each failed file is listed in the error dialog at the completion. The stack dumps in the java console indicate the same error location ie 606 in RemoteCollection.

dizzzz:
Please fill in the Issue template as a lot of essential information is missing here, making this a potential long ping-pong with questions. For next ticket please use the template otherwise we'll not be able to pickup the issues reported effectively.

thammr:
Where do I find the Issue Template?

joewiz:
You’ll see it whenever you create a new issue. It’s supplied from https://github.com/eXist-db/exist/blob/develop/.github/ISSUE_TEMPLATE.md.

thammr:
Problem

Client cannot upload large XML files. If I upload such a file by clicking on the add resource button, the upload fails to complete. The java console shows the stack-dump, I initially reported. 
More significantly, when restoring a backup, large XML files are skipped. The java console displays a stack-dump for each failed file. This stack-dump indicates the same error location. The error dialog displayed at the end of the backup apparently tries to list the skipped files but this dialog has no scroll-bars and is generally scrambled.
Looking at the code, it appears that any XML file larger than 512k will cause this error.
I can find no relevant errors in the eXist logs. 
All these files can be successfully uploaded using the Collection App. 

[...]

PR #1920, opened 2018-06-07
Fix an issue with uploading compressed zip entries with XML:DB Remote API

adamretter:
Closes https://github.com/eXist-db/exist/issues/1913
Closes https://github.com/eXist-db/exist/issues/1874


raducoravu:
👍

PR #1933, opened 2018-06-12
[bugfix] Only omit values for attributes which are considered boolean in HTML5Writer

### LINKED ISSUE #777: html5 method eating attribute value when value is eq to name
With very recent build on develop branch,

``` xquery
xquery version "3.0";

declare option output:method "html5";

<html>
    <div test="test">Hello</div>
</html>
```

will return 

``` xml

<html>
    <div test>Hello</div>
</html>
```

This causes the browser to report 

```
This page contains the following errors:

error on line 3 at column 13: Specification mandate value for attribute test
Below is a rendering of the page up to the first error.
```

... and the wiki app to behave strangely as some parameter will never be sent to the server, like `name` in : 

``` xml
 <input class="form-control" data-template="app:edit-name" type="text" name="name" placeholder="ShortName" required="required" pattern="[A-Za-zäöüÄÖÜ]{1}[\w-äöüßÄÖÜ]{2,}"/>
```

C.

devatwork:
This PR fixes #777 by introducing a list of attribute names which are considered [boolean ](http://w3c.github.io/html/infrastructure.html#boolean-attribute)as per the html 5 specification and are therefore collapsed it their value is equal to their name.

This PR also fixes a case where the attribute name is not written at all in the `HTML5Writer.attribute(QName, String)` function. Please let me know if you want me to split this up into a separate PR.

Finally, this PR adds test cases to test the new behavior.

### DIFF
--- a/src/org/exist/util/serializer/HTML5Writer.java
+++ b/src/org/exist/util/serializer/HTML5Writer.java
@@ -35,6 +35,88 @@ import java.io.Writer;
  */
 public class HTML5Writer extends XHTML5Writer {
 
+    /**
+     * Holds the names of the attributes that are considered boolean
+     * according to http://www.w3.org/TR/html51/single-page.html.
+     * 
+     * The value of these attributes are written if they equal the
+     * name of the attribute. For example: checked="checked" will be
+     * written as checked.
+     * 
+     * See https://github.com/eXist-db/exist/issues/777 for details. 
+     */
+    private final static ObjectHashSet<String> BOOLEAN_ATTRIBUTE_NAMES = new ObjectHashSet<String>(68);
+    static {
+        BOOLEAN_ATTRIBUTE_NAMES.add("allowFullscreen");
+        BOOLEAN_ATTRIBUTE_NAMES.add("async");
+        BOOLEAN_ATTRIBUTE_NAMES.add("autofocus");
+        BOOLEAN_ATTRIBUTE_NAMES.add("autoplay");
+        BOOLEAN_ATTRIBUTE_NAMES.add("badInput");
+        BOOLEAN_ATTRIBUTE_NAMES.add("checked");
+        BOOLEAN_ATTRIBUTE_NAMES.add("closed");
+        BOOLEAN_ATTRIBUTE_NAMES.add("commandChecked");
+        BOOLEAN_ATTRIBUTE_NAMES.add("commandDisabled");
+        BOOLEAN_ATTRIBUTE_NAMES.add("commandHidden");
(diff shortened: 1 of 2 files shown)


devatwork:
I ran all the test successfully via `build.bat test`.

The only thing left to do is to create and run a local build #777 is actually solved and does not require an update to the documentation (unless you want me to include a note documenting the behavior on older versions of eXist?).

Is there any documentation available how to run eXist from a source build? My knowledge of Java is pretty much non-existent.

duncdrum:
@devatwork wow that was quick.
[build from source instructions](https://github.com/eXist-db/exist/blob/develop/BUILD.md)

I m not a win person, so you could either follow the instructions for starting via [.bat on win] (http://www.exist-db.org/exist/apps/doc/exist-building.xml) or do it the [java way](http://www.exist-db.org/exist/apps/doc/advanced-installation) 
```
java -jar start.jar jetty
```

P.S.: Yes we need to consolidate the three different locations telling folks how to build from source

devatwork:
It just bothered me ;) and it is a good opportunity to learn about a different language and code base.

In any case, I ran the local build eXist instance succesfully with following steps:

```ps
build.bat clean-all
build.bat all
java -jar start.jar jetty
```

Next I followed the steps described in http://www.exist-db.org/exist/apps/doc/development-starter#D3.33.

In the old behavior, the value for the `name` attribute would have been omitted causing browser not to send the form values in the query string. With this PR these steps work as expected, so it is a fix for the problem described in #777.

From a documentation perspective, the tutorial will without any modifications when/if this PR is landed. Do I need to do something to get this merged? Also, it may be worth considering to backport this to the stable release branch.

adamretter:
@devatwork Hi Bert, thank you very much for your contribution :-) An excellent piece of work, complete with tests :-)

Sorry for the slow delay in getting back to you, I only arrived back into India yesterday afternoon, and all was quite a blur.

Thanks again. I will send a PR to backport this also...

devatwork:
@adamretter You're welcome, thanks for doing the back porting! And I'm glad you got home safe.

@duncdrum Thanks for your assistance!

joewiz:
@devatwork Thanks so much! To think it allows an existing code snippet in our documentation to work when it previously failed... A really important contribution!

adamretter:
Thanks again @devatwork. Your changes made it into the eXist-db 4.2.1 release (out today) - http://www.exist-db.org/exist/apps/wiki/blogs/eXist/eXistdb421

devatwork:
You're welcome @adamretter and thank you for the recognition in the release notes.

Issue #1927, opened 2018-06-08
NPE on inspect:module-functions unless location is prepended with xmldb:exist://


joewiz:
### What is the problem

`inspect:module-functions` does not work unless the location is prepended with `xmldb:exist://`.

### What did you expect

I expected the `inspect:module-functions` to work even if the location is not prepended with `xmldb:exist://`.

### Describe how to reproduce or add a test

The following two expressions should return identical results:

**Case 1**: *with* `xmldb:exist://` prefix

```xquery
inspect:module-functions(xs:anyURI("xmldb:exist:///db/apps/shared-resources/content/apputil.xql"))
```

**Case 2**: *without* `xmldb:exist://`prefix

```xquery
inspect:module-functions(xs:anyURI("/db/apps/shared-resources/content/apputil.xql"))
```

Case 1 returns the list of functions in the app-util module. Case 2 triggers an NPE:

```text
2018-06-08 16:57:54,365 [qtp1635955718-74] ERROR (XQueryServlet.java [process]:545) - null 
java.lang.NullPointerException: null
```

Work was recently done on this code to address https://github.com/eXist-db/exist/issues/1827 (closed by https://github.com/eXist-db/exist/pull/1833). 

### Context information

- eXist-db version + Git Revision hash: eXist 4.2.0
- Java version: 1.8.0_172-b11
- Operating system: macOS 10.13.4
- 32 or 64 bit: 64 bit
- How is eXist-db installed: cloned from GitHub
- Any custom changes in e.g. conf.xml: no

adamretter:
*NOTE* Also affects `5.0.0-RC1`.

PR #1934, opened 2018-06-13
Make sure that Source Factory correctly and safely resolves sources

adamretter:
Closes https://github.com/eXist-db/exist/issues/1927


joewiz:
Thank you @adamretter!

PR #1935, opened 2018-06-13
(5.0.0) Make sure that Source Factory correctly and safely resolves sources

adamretter:
Back-port of https://github.com/eXist-db/exist/pull/1934

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -4,7 +4,7 @@
 #
 # $Id$
 project.name = eXist-db
-project.version = 5.0.0-RC1
+project.version = 5.0.0-SNAPSHOT
 
 # db settings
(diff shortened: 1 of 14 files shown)


windauer:
Dannes approved, Travis is green and I read through the code as well... will merge

joewiz:
@windauer Just noticed the AppVeyor tests (our only tests for Windows) hadn't completed yet for this PR. Hopefully there's no Windows problem, but in the future we should wait for all the tests to complete before merging, not just Travis.

adamretter:
@joewiz - completely agree. In this case though I already gave @lars special permission to move forward before the AppVeyor results ;-)

PR #1946, opened 2018-06-16
replace exist specific util:parse with xpath function

### LINKED ISSUE #1944: replace custom function util:parse() with xpath fn:parse-xml()
`parse-xml()` is part of Xpath specs, we should replace all occurrences of `util:parse()` with the official function and mark `util:parse` for deprecation in 5.0.0 and then remove it in a later major version update. 

When in doubt, always stick with the standard function

see #1510

- [x] replace function in exist-db core apps (see open PRs)
- [x] set deprecation warning
- [ ] do a wider search on github and inform folks of planned changes

duncdrum:
This is a PR for 5.0.0 only.

replaces calls to `util:parse` with xpath `parse-xml`

there are i think two files that only need to be modified, once we actually remove the function (presumably in 6.0)

@adamretter please take a look at my added depreciation warning, in `src/org/exist/xquery/functions/util/Parse.java` i m not sure how to make sure that it only covers `util:parse` but not` util:parse-html`

close #1944

### DIFF
--- a/src/org/exist/xquery/functions/util/Parse.java
+++ b/src/org/exist/xquery/functions/util/Parse.java
@@ -2,21 +2,21 @@
  *  eXist Open Source Native XML Database
  *  Copyright (C) 2001-09 The eXist Project
  *  http://exist-db.org
- *  
+ *
  *  This program is free software; you can redistribute it and/or
  *  modify it under the terms of the GNU Lesser General Public License
  *  as published by the Free Software Foundation; either version 2
  *  of the License, or (at your option) any later version.
- *  
+ *
  *  This program is distributed in the hope that it will be useful,
  *  but WITHOUT ANY WARRANTY; without even the implied warranty of
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  *  GNU Lesser General Public License for more details.
- *  
+ *
  *  You should have received a copy of the GNU Lesser General Public License
  *  along with this program; if not, write to the Free Software
  *  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
- *  
+ *
  *  $Id$
  */
 package org.exist.xquery.functions.util;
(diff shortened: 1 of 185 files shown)


src/org/exist/xquery/functions/util/Parse.java
```diff
@@ -49,17 +49,22 @@
...
+		* @deprecated use parse-xml instead
+		*/		
     public final static FunctionSignature signatures[] = {
         new FunctionSignature(
             new QName( "parse", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
+						"!This function will be deprecated in a future version!" +
+						"use parse-xml instead." +
             "Parses the passed string value into an XML fragment. The string has to be " +
             "well-formed XML. An empty sequence is returned if the argument is an " +
             "empty string or sequence. If the XML is not well-formed, the function throws an " +
```
adamretter: Hi @duncdrum I like what you are doing...

If you want to deprecate an XQuery function which is implemented in Java though, there is a better way than manually changing the description. The steps to follow are:

1. The `@Deprecated` java annotation should be placed on the entire class (if there is a java class to xquery function one to one equivalence). e.g. https://github.com/eXist-db/exist/blob/develop/src/org/exist/xquery/functions/xmldb/XMLDBChmodCollection.java#L48

2. You can reference the deprecated function as the last argument of the function signature. e.g. https://github.com/eXist-db/exist/blob/develop/src/org/exist/xquery/functions/xmldb/XMLDBChmodCollection.java#L60

The example code links above are for `xmldb:chmod-collection` which is *deprecated by* `security:chmod`.

duncdrum: thx for the hint adam, my problem is actually with how to do  2. would that be `ParsingFunctions.xml-parse` ? I m unclear about the signature [here](https://github.com/eXist-db/exist/blob/21f75b7f0e7ba6f3025bc0b65cd94e40d5882245/src/org/exist/xquery/functions/fn/ParsingFunctions.java#L66)

duncdrum:
hmm I'd say thats a race condition in the branch test on appveyor, it runs fine on travis and on the pr…

adamretter:
@duncrum I restarted the PR. We will also need a backport for 5.0.0, would you be able to open a second PR for that?

duncdrum:
guess i should have created a second branch for the backport *snap*

PR #1948, opened 2018-06-17
Use ErrorCodes for XQuery Errors

adamretter:
Avoids the horrible and incorrect `exerr:ERROR` in a number of situations, by using the correct error codes.


adamretter:
Will backport once the CI passes!

PR #1950, opened 2018-06-17
(4.x.x) Fixes an issue with restoring zero byte binaries via XML:DB Remote API

adamretter:
I will send the backport PR's after CI and after @windauer confirms that it fixes his issues.


windauer:
Tested with a backup that could not be restored using eXist-db 4.2.1 which works perfectly fine now! Thanks a lot @adamretter ! \o/

PR #1949, opened 2018-06-17
(4.x.x) Fix an NPE with XQuery Update and in-memory documents

adamretter:
Closes https://github.com/eXist-db/exist/issues/1761

PR #1951, opened 2018-06-17
(5.0.0) Fix an NPE with XQuery Update and in-memory documents

adamretter:
Forward port of https://github.com/eXist-db/exist/pull/1949


dizzzz:
@adamretter what is the relation with #1952 ?

PR #1956, opened 2018-06-18
(5.0.0) Fixes an issue with restoring zero byte binaries via XML:DB Remote API

adamretter:
Forward port of https://github.com/eXist-db/exist/pull/1950

### DIFF
--- a/src/org/exist/xmldb/RemoteCollection.java
+++ b/src/org/exist/xmldb/RemoteCollection.java
@@ -622,36 +622,37 @@ public class RemoteCollection extends AbstractRemote implements EXistCollection 
                 chunk = new byte[MAX_UPLOAD_CHUNK];
             }
             try {
-                int len;
+
                 String fileName = null;
-                while ((len = is.read(chunk)) > -1) {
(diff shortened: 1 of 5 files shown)


src/org/exist/xmldb/RemoteCollection.java
```diff
@@ -622,36 +622,37 @@ private void uploadAndStore(final Resource res) throws XMLDBException {
...
-                if (fileName == null) {
-                    final byte[] compressed = Compressor.compress(new byte[0], 0);
+                } else {
+                    // Zero length stream? Let's get a fileName!
                     final List<Object> params = new ArrayList<>();
-                    params.add(compressed);
+                    params.add(chunk);
                     params.add(0);
-                    fileName = (String) xmlRpcClientLease.get().execute("uploadCompressed", params);
+                    fileName = (String) xmlRpcClientLease.get().execute("upload", params);
```
dizzzz: this makes a java client using the xmlrpc interface not compatible with a different version of the database, correct?

PR #1973, opened 2018-06-26
Fixes an issue when constructing MemTrees via SAX parsers

adamretter:
Previously if you used the `SAXAdapter` for the `MemtreeBuilder` with a SAX parser which is not namespace aware, then the elements in the Memtree would all have empty element names, which is obviously invalid! This fixes that.


adamretter:
@duncdrum I don't think so

PR #1974, opened 2018-06-26
Basic support for available documents of Dynamic Context

adamretter:
Implements basic support for https://www.w3.org/TR/xpath-31/#dt-available-docs

We need this to be able to execute the XQTS correctly.


adamretter:
Apart from the tests I added. I will be directly calling it from the XQTS driver.

PR #1987, opened 2018-07-01
(5.0.0) Basic support for available collections of Dynamic Context

adamretter:
Implements basic support for https://www.w3.org/TR/xpath-31/#dt-available-collections

We need this to be able to execute the XQTS correctly.

### DIFF
--- a/src/org/exist/xquery/XQueryContext.java
+++ b/src/org/exist/xquery/XQueryContext.java
@@ -232,11 +232,18 @@ public class XQueryContext implements BinaryValueManager, Context
     private Map<String, TriFunctionE<DBBroker, Txn, String, Either<org.exist.dom.memtree.DocumentImpl, org.exist.dom.persistent.DocumentImpl>, XPathException>> dynamicDocuments = null;
 
     /**
-     * The available test resources of the dynamic context.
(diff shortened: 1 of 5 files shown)


src/org/exist/xquery/XQueryContext.java
```diff
@@ -1249,6 +1264,23 @@ public DocumentSet getStaticDocs() {
         return docSupplier.apply(getBroker(), getBroker().getCurrentTransaction(), uri).fold(md -> (Sequence)md, pd -> (Sequence)pd);
     }
 
+    /**
+     * Get's a document from the "Available collections" of the
+     * dynamic context.
+     */
+    public @Nullable Sequence getDynamicallyAvailableCollection(final String uri) throws XPathException {
+        if (dynamicCollections == null) {
+            return null;
```
dizzzz: just a q: why didn't you choose for Optional ?

adamretter: I could have equally used `Optional` here, but I felt that in this instance `@Nullable` was a better fit.

dizzzz:
@adamretter Resolve issue :-(

PR #1989, opened 2018-07-03
Support the Collation used in XQTS 3.1

adamretter:


adamretter:
@duncdrum no one should ever need to know that this new collation option is present. It is just a bridge for running the XQTS.

duncdrum:
@adamretter I don't really follow the logic of known unknowns when it comes to the docs. Support for `caseblind` is a feature, so it should be mentioned in the documentation where it explicitly states which options exist supports, which already includes `html-ascii-case-insensitive`.

adamretter:
@duncdrum, my point is that the Collation's URI is - `http://www.w3.org/2010/09/qt-fots-catalog/collation/caseblind` i.e. it is part of XQTS - indicated by the `qt-fots-catalog` part and so should not be used from anywhere apart from XQTS; To ensure that, I actually want it to be undocumented, so that people don't start using it ;-)

PR #1990, opened 2018-07-03
(5.0.0) Support the Collation used in XQTS 3.1

adamretter:
Backport of https://github.com/eXist-db/exist/pull/1989


adamretter:
@ljo yes as I suspect there are some subtle differences, but I cannot discern them yet. I just want to get the basics there so the XQTS tests won't immediately error.

ljo:
@adamretter So regarding the milestone change. Do you want me to pull this after 4.3.0 or should I do it now since it is really a minor change as is?

adamretter:
@ljo this PR is against the develop-5.0.0 branch, so you can pull it now if you like. It won't affect 4.3.0 at all.

PR #2016, opened 2018-07-12
Add further JDK's to Travis

adamretter:


adamretter:
@dizzzz both are JDK 11 versions. Looks like we have problems with the build on Java 11

adamretter:
@dizzzz this is ready to land.

dizzzz:
Adam: yes that would be very beneficial, as ppl will start using eXist with these java versions too.....

dizzzz:
thnx @adamretter !

PR #2038, opened 2018-07-16
Fixes another issue with zero byte binary resources in the XML:DB Remote API

adamretter:


src/org/exist/xmldb/RemoteCollection.java
```diff
@@ -647,10 +647,12 @@ private void uploadAndStore(final Resource res) throws XMLDBException {
                             fileName = (String) xmlRpcClientLease.get().execute("uploadCompressed", params);
                         }
                     }
-                } else {
+                }
+
+                if (fileName == null) {
```
dizzzz: `fileName` is set in line 642 or line 647 ...  which checking for null? probably there is an indenting confusion for me...

PR #2062, opened 2018-07-25
Updated the function defintions for scheduling jobs via XQuery

MikeJ1971:
### Description:

I've updated the function definitions to indicate that submitting a scheduled job via XQuery isn't persistent.

### Reference:

https://sourceforge.net/p/exist/mailman/message/36375333/

### Type of tests:

None. Should existing tests suffice?


adamretter:
@MikeJ1971 Excellent. Thank you very much :-)

PR #2075, opened 2018-07-29
Fix issues with IzPack and shutdown on Windows

adamretter:
As reported by Dominic Martin on the exist-open mailing list.

### DIFF
--- a/installer/install.xml.tmpl
+++ b/installer/install.xml.tmpl
@@ -3,7 +3,7 @@
         <appname>@name@</appname>
         <appversion>@version@</appversion>
         <url>http://exist-db.org</url>
-        <javaversion>1.6</javaversion>
+        <javaversion>1.8</javaversion>
 		<requiresjdk>no</requiresjdk>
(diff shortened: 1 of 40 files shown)


dizzzz:
the other build of 4.x.x was ok, but for this build I see

image

adamretter:
@dizzzz actually I am wondering if I introduced an issue with the lock events. I think the `LockTable` needs to become per-`BrokerPool`. It is likely that we start and destroy `BrokerPools` within the same VM, but that the LockTable processor is destroyed when the first BrokerPool is, and never recreated - which could lead to us collecting lots of lock events which are never processed.

adamretter:
@dizzzz Assuming CI passes, I think this is now good to go.

dizzzz:
off and on some tests fail, ut in general I think the PR is OK..

duncdrum:
codacy discovered an unused import, that you might want to remove

adamretter:
@dizzz indeed I think the CI failures are transient.

@duncdrum okay, I have made that that change. Thanks

PR #2100, opened 2018-08-06
(4.x.x) Consistency check should check the entire document

adamretter:
A very simple two line fix, that ensures the consistency check considers all nodes in the document and not just the document element.


adamretter:
@dizzzz The XMLStreamReader will then read over all other nodes of the document starting from the first node, that includes descendants.

PR #2103, opened 2018-08-11
Consistency check should check the entire document

adamretter:
Forward port of https://github.com/eXist-db/exist/pull/2100

### DIFF
--- a/src/org/exist/backup/ConsistencyCheck.java
+++ b/src/org/exist/backup/ConsistencyCheck.java
@@ -21,6 +21,7 @@
  */
 package org.exist.backup;
 
+import org.exist.dom.persistent.*;
 import org.exist.security.Account;
 import org.exist.security.Group;
 import org.exist.security.Permission;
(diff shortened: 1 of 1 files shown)


src/org/exist/backup/ConsistencyCheck.java
```diff
@@ -322,8 +319,8 @@ public ErrorReport checkXMLTree( final DocumentImpl doc )
                     public Object start() {
                         EmbeddedXMLStreamReader reader = null;
                         try {
-                            final ElementImpl             root            = (ElementImpl)doc.getDocumentElement();
-                            reader = (EmbeddedXMLStreamReader)broker.getXMLStreamReader( root, true );
+                            final Node              root            = doc.getFirstChild();
```
dizzzz: First child.... should we do this for all children? Is a comment also a node ?

adamretter: First child of the document is just the starting point for the scan. The XMLStreamReader will then read over all other nodes. So yes, the first node could be a comment.

PR #2105, opened 2018-08-11
Some general code cleanup

adamretter:
1. Small mechanical code cleanup
2. Switch from `java.util.Stack` to more performant `java.util.ArrayList`.

PR #2107, opened 2018-08-11
(4.x.x) Some general code cleanup

adamretter:
Backport of https://github.com/eXist-db/exist/pull/2105


adamretter:
@dizzzz Ah I missed a small think in the rebase. Should be good now...

bin/functions.d/eXist-settings.sh
```diff
@@ -1,6 +1,8 @@
-# -*-Shell-script-*-
-# Common eXist script functions and settings
-# $Id:eXist-settings.sh 7231 2008-01-14 22:33:35Z wolfgang_m $
+#!/usr/bin/env bash
```
dizzzz: A wise (?) man (you?) told me once that is this is a dangerous construct... makes the script vulnerable ?

adamretter:
@dizzzz Rebased as requested

bin/functions.d/eXist-settings.sh
adamretter: Hmmm. I am not certain about that.

However it seems that these days, `#!/usr/bin/env bash` is recommended over `#!/bin/bash` for portability (not all systems have bash in `/bin` e.g. NixOS).

dizzzz:
:-( the process is really silent for 10+ minutes. De we have a deadlock ?

adamretter:
@dizzzz It was a dangling lock problem.
Somehow whilst merging during the back-port I removed a line that released a lock. I have now reinstated that, rebased and pushed.

All should be good now :-)

Issue #1691, opened 2018-01-16
Reindexing collection causes xpath in xsl transform to fail.


paulmer:
### What is the problem
Reindexing a collection causes XSL transformations run by the `transform:transform` function to fail.  XPath expressions and keys generated on a document included by the `xsl:doc` function behave strangely and generate different results after the reindex requested from the Java client.

### What did you expect
The attached example which counts the nodes in a document illustrates the problem.  Prior to the reindex, the node count reported is `1`, after the reindex, the count is `0`.  Editing and saving the XML document restores the count to `1`.  Removing either or both of the two comments at the start of the XML document appears to eliminate the problem as well.

### Describe how to reproduce or add a test
Using the Java client:
1. Create a test collection, `/db/test`. 
2. Upload the `test.xql`, `xml.xml` and `xsl.xsl` documents attached to this collection using the Java client.  
3. Open `test.xql` and choose submit.  The result should be `1`.  (`<param-count expr="count($xml//*)">1</param-count>` specifically.)
4. Choose the `/db/test` collection and reindex.
5. Submit `test.xql` again.  The result is `0`.
6. Open `xml.xml` and choose save.
7. Submit `test.xql` again.  The result is `1`.  Repeating step 4 and 5 will cause the result to become `0` again.
8. Remove either comment at the beginning of `xml.xml` and save the file.  
9. Submit `test.xql`.  The result will be `1`.
10. Reindex test again, and resubmit `test.xql`.  With the comment removed, the result will be `1` after the reindex.

Although this test doesn't demonstrate it, I also found this behavior:  If the XML document consists of two nodes:  `<a><b/></a>`, then after reindexing, the expression `count(doc('xml.xml')//*)` returns `0` , but `count(doc('xml.xml')/a//*)` returns `1` if evaluated inside a transformation.  If these expressions are evaluated in the XQuery, they return the expected results before and after the reindex.  Only expression evaluation inside a transformation are affected.

[test.zip](https://github.com/eXist-db/exist/files/1636765/test.zip)

### Context information
Please always add the following information
- eXist-db version: 3.6.1
- Java version: Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
- Operating system: MacOS & Linux
- 64 bit
- Any custom changes in e.g. conf.xml: N/A (problem exists on customized and uncustomized instances)

joewiz:
Very interesting, Paul. Nice job whittling down the test case.

[...]

PR #2113, opened 2018-08-13
Various fixes for preceding-sibling and following-sibling axis

adamretter:
Closes https://github.com/eXist-db/exist/issues/1691

1. This includes many fixes for `preceding-sibling` and `following-sibling` axis in eXist-db. Particularly at the document level.
2. It also offers an optimisation for `following-sibling` where scanning can commence from the current node; Previously scanning would start from the parent node, and traverse all nodes before the current node.
3. It fixes a number of issues with node selection at the document level.
4. Fixes an issue with index data loss for nodes at the document level when reindexing documents.
5. XML:DB Remote API when calling `#getContentAsDOM()` will now return an element node where appropriate, and not always a document node.
6. `DocumentImpl#getNode(NodeId)` now correctly returns all node types, not just elements.


src/org/exist/xmldb/RemoteXMLResource.java
```diff
@@ -96,12 +120,14 @@ public RemoteXMLResource(
             final int handle,
             final int pos,
             final XmldbURI docId,
-            final Optional<String> id)
+            final Optional<String> id,
+            final Optional<String> type)
```
dizzzz: shouldn't Optional's in parameters of constuctors/methods be avoided?

adamretter: I have seen some advice that says that yes, and some that doesn't. I am not 100% convinced yet...

extensions/indexes/lucene/src/org/exist/xquery/modules/lucene/Query.java
```diff
@@ -316,13 +318,21 @@ protected Properties parseOptions(Sequence contextSequence, Item contextItem) th
         Sequence optSeq = getArgument(2).eval(contextSequence, contextItem);
         NodeValue optRoot = (NodeValue) optSeq.itemAt(0);
         try {
-            XMLStreamReader reader = context.getXMLStreamReader(optRoot);
+            final int thisLevel = optRoot.getNodeId().getTreeLevel();
+            final XMLStreamReader reader = context.getXMLStreamReader(optRoot);
+            reader.next();
```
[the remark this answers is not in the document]
dizzzz: @adamretter ?

duncdrum:
Ehm win builds are failing?

dizzzz:
Ow?

PR #2117, opened 2018-08-16
Correct jmx option parsing

hungerburg:
Enable jmx and try to parse port ONLY when option matches at beginning of argument.

### DIFF
--- a/bin/functions.d/getopt-settings.sh
+++ b/bin/functions.d/getopt-settings.sh
@@ -45,19 +45,19 @@ is_jmx_switch() {
 	if substring "${JMX_OPTS}" "|$1|"; then
 		JMX_ENABLED=1;
 		return 0;
-	elif substring "|$1|" "$JMX_SHORT_EQUAL"; then
+	elif substring "|X$1|" "X$JMX_SHORT_EQUAL"; then
 		JMX_ENABLED=1;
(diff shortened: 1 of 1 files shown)


adamretter:
@hungerburg awesome thanks :-)

Could you send PR's for the `develop-4.x.x` and `develop-5.0.0` as well please?

adamretter:
@hungerburg did you mean to close those?

hungerburg:
@adamretter I tested the patch here, works fine; Is it still available -- I did a git flow command that I do not understand

joewiz:
@adamretter @dannes Is anyone still using Git Flow for eXist? I recall it being part of a README when we first switched to GitHub, struggled with it, stopped using it, and forgot about it until now. I see it is still mandated at https://github.com/eXist-db/exist/blob/develop/CONTRIBUTING.md. I don’t use it or see others using it. Am I mistaken, or shall we remove git flow from the contributing guidelines?

adamretter:
@joewiz I can update the `CONTRIBUTING.md`. It is not necessary to use the git flow command, but basically the workflow is the same if you use it or not.

dizzzz:
Myself I don't use the tool, but I do adhere to the fundamentals....

duncdrum:
i d say put the git-flow question on the community call agenda so more folks can chime in. I think the cons outweigh the pros but it seems easier to discuss this there then in the comments of this PR

Issue #2151, opened 2018-09-04
backup/restore not working


andray71:
### What is the problem

executing command:

```bash
bin/backup.sh -u admin -p [password] -r /backup/backup-config-service.zip
```

getting errors:

```bash
/opt/exist/bin/functions.d/eXist-settings.sh: line 74: $'-Dfile.encoding=UTF-8;\n    fi\n\n    OS=Linux\n    if [  == Darwin ]; then\n\tCLIENT_NAME=Client\n        JAVA_OPTIONS=': command not found
/opt/exist/bin/functions.d/eXist-settings.sh: line 34: locale: command not found
/opt/exist/bin/functions.d/eXist-settings.sh: line 35: locale: command not found
Using locale: en_US.UTF-8
```

### What did you expect

restored content from the zip file (old backup)

works fine when i restore it using Java Admin UI

older versions (3.*) worked fine.

### Describe how to reproduce or add a test

running the eXist in alpine docker image

### Context information

- eXist-db version  - 4.3.1
- Java version - 1.8.0_171
- Operating system - Linux 4.9.93-linuxkit-aufs amd64
- 64 bit
- How is eXist-db installed? Jar inside alpine docker image
- Any custom changes - no

duncdrum:
From the error message I would say this is a problem with how Java is configured inside alpine. You could try with the exist-db provided images. Or we need more info about your docker configuration, whats the base image, how do you install java, how do you install exist, etc.

omkarnix:
If no one else is working on this issue, I would like to give a try to investigate.

joewiz:
@omkarnix That would be great, thank you!

duncdrum:
@andray71 is this still a problem? without more info I'm afraid this will be a nofix

dizzzz:
2 years no response. please feel free to re-open/add details

PR #2152, opened 2018-09-04
4.x.x: Upgrade jetty to version 9.4.12

dizzzz:
from the release notes:

Notably, Jetty 9.4.12 includes compatibility for JDK 11. Additionally, TLS 1.3 support has been implemented. While full functionality for new JDK features is not yet supported, this release has been built and tested for compatibility with the latest releases from Oracle.

All changes, including 9.4.11: https://github.com/eclipse/jetty.project/blob/27208684755d94a92186989f695db2d7b21ebc51/VERSION.txt


dizzzz:
```
Finished restore of backup.
3939 [ERROR] XQueryURLRewrite: Error while processing /xmlrpc: 501: Unsupported Content-Encoding
org.eclipse.jetty.http.BadMessageException: 501: Unsupported Content-Encoding
	at org.eclipse.jetty.server.Request.extractContentParameters(Request.java:460) ~[jetty-server-9.4.12.v20180830.jar:9.4.12.v20180830]
	at org.eclipse.jetty.server.Request.getParameters(Request.java:386) ~[jetty-server-9.4.12.v20180830.jar:9.4.12.v20180830]
	at org.eclipse.jetty.server.Request.getParameterMap(Request.java:1039) ~[jetty-server-9.4.12.v20180830.jar:9.4.12.v20180830]

```

duncdrum:
hmm looks eerily similar too #2151

dizzzz:
image

dizzzz:
@duncdrum no, with the current version of jetty the tests work just fine.

duncdrum:
@dizzzz I was more thinking along the lines of possible changes to how locals are handled. Could just be coincidence, but two errors within 24h about backup and content encoding make me suspicious.

joakime:
https://github.com/eXist-db/exist/blob/be7cb57cbc410a3e29a32de46518f7584bcffb7a/src/org/exist/http/urlrewrite/XQueryURLRewrite.java#L1048-L1053

Why are you reading the entire Request ParameterMap in a Request Wrapper?
That will prevent any use of Servlet 3.1 Async I/O as you are forcing the entire request input stream to be read in a blocking fashion during the constructor of the RequestWrapper.

Also, that wrapper doesn't handle the charset order properly for Readers.
That wrapper also allows you to set the inputstream and/or reader without a regard for the rules of the Servlet spec.
That wrapper also doesn't use the getInputStream() properly from the getReader()
The wrapper also doesn't setup the getCharacterEncoding() properly. 
Nor does it handle the complexity of getContentType() properly.
Ordered / Priority / Quality based getContentType() isn't supported by your wrapper.
It doesn't handle Parts properly (or mime multipart).

The CachingResponseWrapper is counterproductive to the use of GzipHandler and Servlet 3.1 Async I/O.
You'd have far better performance using Async I/O for your responses then attempting to trade memory and cache the responses the way you are doing.
Using Async I/O you'll use less memory, use less threads, not hold memory for slow clients, and be able to support more active connections.

In short, you have a lot of fundamental problems with your various wrappers.
Follow the servlet spec.
Eg: you can't change getCharacterEncoding or getContentType or getMethod after a call to getParameterMap() for example.
You cannot modify the getParameterMap also, as that will mess up RequestDispatcher support during include/forward.
Also if you ever hope to gain the benefits from Servlet 3.1 and Async I/O, then you should never wrap the ServletInputStream or ServletOutputStream (or associated readers/writers).

dizzzz:
@joakime thnx for your input, I will study your remarks;  I am not sure if we have the ambition to support the async-io in our code, but fellow devs might think different. @adamretter @wolfgangmm

adamretter:
@dizzzz I think we would need to understand why we have the caching wrappers in the first place. I probably used to know, but it has been too long and I have forgotten...

dizzzz:
@adam I need to think too @wolfgangmm

dizzzz:
waiting for https://github.com/eclipse/jetty.project/pull/3031 to be released

joakime:
@dizzzz Jetty version `9.4.13.v20181111` is now available on Maven Central.

Issue #2157, opened 2018-09-15
Module import via remote URL is broken


joewiz:
### What is the problem

eXist used to support loading modules over the web, but doing so with 4.3.1 raises an error:

```xquery
xquery version "3.1";

import module namespace functx = "http://www.functx.com" 
    at "http://www.xqueryfunctions.com/xq/functx-1.0-doc-2007-01.xq";

"hello"
```

The error raised is:

> Invalid URI: xmldb URI scheme does not start with xmldb:: http://www.xqueryfunctions.com/xq/functx-1.0-doc-2007-01.xq

Under 4.0.0 (release), this query does not raise an error, but instead returns the expected result, `"hello"`.

### What did you expect

Loading modules via URI is discussed at https://exist-db.org/exist/apps/doc/xquery#module-system.

### Describe how to reproduce or add a test

See test query above. The stack trace is:

```text
2018-09-15 15:16:56,737 [qtp2000015086-35] ERROR (XQueryServlet.java [process]:552) - Invalid URI: xmldb URI scheme does not start with xmldb:: http://www.xqueryfunctions.com/xq/functx-1.0-doc-2007-01.xq 
java.lang.IllegalArgumentException: Invalid URI: xmldb URI scheme does not start with xmldb:: http://www.xqueryfunctions.com/xq/functx-1.0-doc-2007-01.xq
	at org.exist.xmldb.XmldbURI.create(XmldbURI.java:219) ~[exist.jar:4.3.1]
	at org.exist.xmldb.XmldbURI.append(XmldbURI.java:595) ~[exist.jar:4.3.1]
	at org.exist.source.SourceFactory.getSource(SourceFactory.java:98) ~[exist.jar:4.3.1]
	at org.exist.xquery.XQueryContext.importModule(XQueryContext.java:2846) ~[exist.jar:4.3.1]
	at org.exist.xquery.parser.XQueryTreeParser.importDecl(XQueryTreeParser.java:6312) ~[exist.jar:4.3.1]
	at org.exist.xquery.parser.XQueryTreeParser.prolog(XQueryTreeParser.java:5340) ~[exist.jar:4.3.1]
	at org.exist.xquery.parser.XQueryTreeParser.mainModule(XQueryTreeParser.java:4042) ~[exist.jar:4.3.1]
	at org.exist.xquery.parser.XQueryTreeParser.module(XQueryTreeParser.java:3987) ~[exist.jar:4.3.1]
[...]

PR #2161, opened 2018-09-20
Fix a regression for loading modules from non-XMLDB URI

adamretter:
Closes https://github.com/eXist-db/exist/issues/2157


joewiz:
@adamretter Thank you! This is a huge help!

My only question is whether there are already adequate tests for this, or if not, should they be XQSuite tests or would they need to be java? For example, we could have a query like the one in the original issue that tries to import functx; the test could use the `%test:assumeInternetAccess` annotation. I could also imagine tests that store modules in the database root or in subcollections and try to import these from/via various contexts (`/db`, `xmldb:exist:/db`, `xmldb:exist:///db`, `xmldb:exist://localhost/db`, `xmldb:exist://blah/db`, etc.).

adamretter:
@joewiz so there are some tests for this which were actually failing, unfortunately XQSuite was reporting them as passing. Those things will be addressed in the other Open PR around XQSuite etc.

PR #2164, opened 2018-09-20
Add some missing files to the dist

adamretter:
The `.tar.gz` and `.tar.bz` distributions were missing some source code files which are needed for rebuilding from source.

### DIFF
--- a/build/scripts/dist.xml
+++ b/build/scripts/dist.xml
@@ -242,6 +242,7 @@
                 <include name="*.bat"/>
                 <include name="*.sh"/>
                 <include name="conf.xml"/>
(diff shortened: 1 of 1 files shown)


dizzzz:
@duncdrum the distribution files should be containing all code so eXist-db can be built entirely from the archive...

duncdrum:
@dizzzz imv, this explains the `*.tmpl`, but not necessarily the test directory`/test/src/`
in any case maven will hopefully soon make this a non issue

dizzzz:
not sure; the directory with the test sources are integral part of exist-db ; the build scripts will fail without them. When switching to maven, nothing changes right?

PR #2168, opened 2018-09-23
(4.x.x) Fix handling of / when context item was declared

### LINKED ISSUE #2156: Bug in eXist's handling of / when a context item has been declared
### What is the problem

Support for the [context item declaration](https://www.w3.org/TR/xquery-31/#id-context-item-declarations) was added in #1571. eXist's implementation correctly redefines the [context item expression](https://www.w3.org/TR/xpath-31/#id-context-item-expression) `.` to be the item from the declaration. 

However, eXist's implementation does not carry this through to redefine the node selected by `/` (which the XPath spec [defines](https://www.w3.org/TR/xpath-31/#id-path-expressions) as "an abbreviation for the initial step (`fn:root(self::node())` treat as `document-node()`)"). 

Instead, eXist treats `/` as `collection()`, even when a context item has been declared.

This was first reported by @martin-honnen in https://github.com/eXist-db/eXide/issues/204.

### What did you expect

I expected eXist to return the same result for `/` as its expanded form `fn:root(self::node())` when a context item has been declared.

```xquery
xquery version "3.1";

declare context item := document {
    <root>
        <item>foo</item>
        <item>bar</item>
        <item>foo</item>
    </root>
};

. instance of document-node(),
self::node() instance of document-node(),
root(self::node()) instance of document-node(),
(/) instance of document-node()
```

This query returns `true(), true(), true(), false()`. The final result should also be `true()`.

(Unfortunately, XQSuite doesn't respect a context item declaration ... so a proper test would probably need to be implemented in java alongside the other tests for this feature in https://github.com/eXist-db/exist/blob/develop/test/src/org/exist/xquery/XQueryDeclareContextItemTest.java.)

### Context information

- eXist-db version + Git Revision hash: eXist 4.3.1 release
- Java version: 10.0.2+13
- Operating system: macOS 10.13.6
- 32 or 64 bit: 64 bit
- How is eXist-db installed: DMG installer
- Any custom changes in e.g. conf.xml: None

wolfgangmm:
Closes #2156

### DIFF
--- a/src/org/exist/xquery/RootNode.java
+++ b/src/org/exist/xquery/RootNode.java
@@ -30,9 +30,7 @@ import org.exist.numbering.NodeId;
 import org.exist.storage.UpdateListener;
 import org.exist.util.LockException;
 import org.exist.xquery.util.ExpressionDumper;
-import org.exist.xquery.value.Item;
-import org.exist.xquery.value.Sequence;
-import org.exist.xquery.value.Type;
+import org.exist.xquery.value.*;
 
 import java.util.Iterator;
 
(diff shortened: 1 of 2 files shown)


brihaye:
Absolutely no importance, but a slight typo : "getContextItemDeclartion"

joewiz:
@wolfgangmm Many thanks!! It really helps having this working. I developed this as an alternative for readers without oXygen (and its XPath toolbar) to help readers XPath learn to write simple path expressions against a single XML document. Quoting from the text: 

> To follow along with the examples in this section, we recommend that you use an XML editor like oXygen that has an XPath toolbar that allows you to explore a document with XPath. You'll definitely learn more quickly if you copy the XML document above into the editor and try out these XPath exercises yourself in the XPath toolbar. If you don't have oXygen, we can achieve the same goal by pasting the following "skeleton" XQuery into your editor and following the directions in the comments. We'll use this skeleton, ignoring the syntax of the XQuery until we get to the next chapter:
>
> ```xquery
> xquery version "3.1";
> 
> declare context item := document {
>     (: Enter your XML document on the following line :)
> 
> };
> 
> (: Enter your XPath expression on the following line :)
```

adamretter:
@joewiz Interesting I used a similar approach in my class at the XML Summer School. However I bound to a variable rather than the context item. Your approach is nicer.

joewiz:
@adamretter I'd also originally gone with the variable binding approach, but as I was reviewing the manuscript I realized that we could use this, which is far less fiddly and a really nice use case for the "context item" feature. (Thanks again for adding that in the first place!) 

I could imagine leveraging this to build a document-focused query tool - one pane for the source document, one pane for an XPath expression (or in practice, a short XQuery expression, since I don't think we have any way of limiting util:eval to pure XPath...), and one pane for results. Sort of like http://xqueryfiddle.liberty-development.net/ but built for eXist. (In fact, I bet @martin-honnen was thinking of just this when he reported the initial issue, https://github.com/eXist-db/eXide/issues/204.)

dizzzz:
@wolfgangmm looks good, please could you make PRs for 5.x and develop? If you lack time please ping me, I'll do so.

dizzzz:
no echo, two Pas created.

PR #2169, opened 2018-09-24
Phase1 of "Allow more elaborate XPath expressions in the Lucene index spec in collection.xconf"

sten1ee:
### Description:
We need finer control over which nodes get selected for FullText indexing with specific Analyzer.
More specifically, the text match attribute of collection.xconf should support more elaborate XPath expressions, e.g.:
```html
<text match="//tei:teiHeader/tei:fileDesc//tei:title[@xml:lang='Sa-Ltn']"/>
```
Currently, attempting to use the above spec. fragment results in:
```
*WARN  (LuceneConfig.java [parseConfig]:311) - Invalid lucene configuration element: Lucene module: invalid qname in configuration:
Illegal QName: 'tei:title[@xml:lang='Sa-Ltn']' not a valid local name. *
```

### Reference:

Here is a message from the original eXist-open thread that raised the issue of this feature being missing:
https://sourceforge.net/p/exist/mailman/message/36392026/

### Type of tests:
Note that this is phase 1 of the feature implementation, where only some design/structural changes are introduced; changes that are deemed necessary for the clean implementation of the feature.
Thus it is only necessary that the existing tests pass, i.e. the changes comprising this PR do not break anything.

### DIFF
--- a/src/org/exist/source/SourceFactory.java
+++ b/src/org/exist/source/SourceFactory.java
@@ -77,14 +77,9 @@ public class SourceFactory {
         Source source = null;
 
         /* resource: */
-        if (location.startsWith(ClassLoaderSource.PROTOCOL)) {
-            source = new ClassLoaderSource(location);
-        } else if (contextPath != null && contextPath.startsWith(ClassLoaderSource.PROTOCOL)) {
-            // Pretend it is a file on the local system so we can resolve it easily with URL() class.
-            final String conPathNoProtocol = contextPath.replace(ClassLoaderSource.PROTOCOL, "file://");
-            String resolvedURL = new URL(new URL(conPathNoProtocol), location).toString();
-            resolvedURL = resolvedURL.replaceFirst("file://", ClassLoaderSource.PROTOCOL);
-            source = new ClassLoaderSource(resolvedURL);
+        if (location.startsWith(ClassLoaderSource.PROTOCOL)
+                || (contextPath != null && contextPath.startsWith(ClassLoaderSource.PROTOCOL))) {
+            source = getSource_fromClasspath(contextPath, location);
         }
 
         /* xmldb */
@@ -152,6 +147,26 @@ public class SourceFactory {
                 .create(path)
                 .getPathSegments()[0]
                 .getRawCollectionPath();
+    }
+
+    private static Source getSource_fromClasspath(final String contextPath, final String location) throws IOException {
+        if (location.startsWith(ClassLoaderSource.PROTOCOL)) {
+            return new ClassLoaderSource(location);
+        }
+
+        final Path rootPath = Paths.get(contextPath.substring(ClassLoaderSource.PROTOCOL.length()));
+
+        // 1) try resolving location as child
+        final Path childLocation = rootPath.resolve(location);
+        try {
+            return new ClassLoaderSource(ClassLoaderSource.PROTOCOL + childLocation.toString().replace('\\', '/'));
+        } catch (final IOException e) {
+            // no-op, we will try again below
+        }
+
+        // 2) try resolving location as sibling
+        final Path siblingLocation = rootPath.resolveSibling(location);
+        return new ClassLoaderSource(ClassLoaderSource.PROTOCOL + siblingLocation.toString().replace('\\', '/'));
     }
 
     /**
--- a/test/src/org/exist/source/SourceFactoryTest.java
+++ b/test/src/org/exist/source/SourceFactoryTest.java
@@ -134,6 +134,53 @@ public class SourceFactoryTest {
     }
 
     @Test
+    public void getSourceFromResource_contextFolderUrl_locationRelative() throws IOException, PermissionDeniedException {
+        final String contextPath = "resource:org/exist/source";
+        final String location = "library.xqm";
+
+        final Source source = SourceFactory.getSource(null, contextPath, location, false);
+
+        assertTrue(source instanceof ClassLoaderSource);
+        assertEquals(getClass().getResource("library.xqm"), source.getKey());
+    }
+
+    @Test
+    public void getSourceFromResource_contextFolderUrl_locationAbsoluteUrl() throws IOException, PermissionDeniedException {
+        final String contextPath = "resource:org/exist/source";
+        final String location = "resource:org/exist/source/library.xqm";
+
+        final Source source = SourceFactory.getSource(null, contextPath, location, false);
+
+        assertTrue(source instanceof ClassLoaderSource);
+        assertEquals(getClass().getResource("library.xqm"), source.getKey());
+    }
+
+    @Test
+    public void getSourceFromResource_contextFolderUrl_locationRelativeUrl() throws IOException, PermissionDeniedException {
+        final String contextPath = "resource:org/exist/source";
+        final String location = "library.xqm";
+
+        final Source source = SourceFactory.getSource(null, contextPath, location, false);
+
+        assertTrue(source instanceof ClassLoaderSource);
+        assertEquals(getClass().getResource("library.xqm"), source.getKey());
+    }
+
+    @Test
+    public void getSourceFromResource_contextFolderUrl_locationRelativeUrl_basedOnSource() throws IOException, PermissionDeniedException {
+        final String contextPath = "resource:org/exist/source";
+        final String location = "library.xqm";
+
+        final Source mainSource = SourceFactory.getSource(null, "", contextPath, false);
+        assertTrue(mainSource instanceof ClassLoaderSource);
+
+        final Source relativeSource = SourceFactory.getSource(null, ((ClassLoaderSource)mainSource).getSource(), location, false);
+
+        assertTrue(relativeSource instanceof ClassLoaderSource);
+        assertEquals(getClass().getResource(location), relativeSource.getKey());
+    }
+
+    @Test
     public void getSourceFromXmldb_noContext() throws IOException, PermissionDeniedException {
         final String contextPath = null;
         final String location = "xmldb:exist:///db/library.xqm";


src/org/exist/storage/NodePath2.java
```diff
@@ -0,0 +1,87 @@
...
+import org.exist.dom.persistent.IStoredNode;
+import org.w3c.dom.Element;
+
+/**
+ * @author Stanislav Jordanov <stenlee@gmail.com>
+ * @version 1.0
+ *
+ * This is an extension of class NodePath, that keeps track of a *real* node/element path, not just a QName path
+ * as its base class -- NodePath does.
+ * This is required in order to implement the feature requested/discussed here:
```
adamretter: I am a little confused by your terminology here.

What do you man by a "real node/element path"?

I think you need to use the terms that we would expect to find in XQuery. For example do you actually mean an XPath Expression? Predicates, etc. etc.?

adamretter:
@sten1ee I am interested to see more. I think you should keep adding commits to this PR.

At the moment this PR doesn't actually do anything apart from rename some stuff. So I don't think it needs to be merged yet. I am interested to see how you are actually going to evaluate these more complex index expressions during indexing.

I have some experience with such things here - https://github.com/BCDH/exist-algolia-index#configuration
My experience is that the indexing pipeline is SAX like (event based), so it is very hard to evaluate complex XPath expressions for the index definitions (which typically rely on a DOM interface). I settled on just allowing a subset of XPath expression.

Look forward to seeing what you come up with...

sten1ee:
That's the thing -- the best option I had was to guess what a "NodePath" is
supposed to represent.
Apart from /** @author wolf */ there's no other class comment.
I am not acquainted with the XPath terminology, excuse me about that.
But there is no "node path" definition in the XQuery 3.1 specification
either ( https://www.w3.org/TR/xpath-31 )
So, I had to investigate.
It turned out NodePath is used to represent at least these 2 distinct ...
err, let's call them 'concepts':
1. Holding the value of a element's path starting from the root of the doc
(akin to 'absolute file path' in Linux)
2. A limited XPath expression - as the one that is parsed by
NodePath.init() method and allow  '*', '//', '/' and QNames only (in
particular -- no predicates)
In my opinion "1" and "2" should be represented by two distinct classes,
the way a 'java.lang.String' and 'java.util.regex.Pattern' are two distinct
classes, but anyway.
My initial attempt at implementing the feature in question was to start
hacking NodePath to make it aware of node attributes and predicates on
attributes.
Guess what happened?
This is what happened -
https://sourceforge.net/p/exist/mailman/message/36412983/

In short -- I figured out class NodePath is used to represent too many
things and thus it is almost *untouchable.*
Phase 1 deals mostly with this untouchable-ness of NodePath :) -- it gives
'room to breath' without breaking stuff that expects NodePath to behave as
it used to.
I believe (I may be wrong, of course) that what is left now, in order to
implement the feature in question,
is to properly implement NodePathExpression's init() (i.e. parse XPath) and
match(NodePath(2)) methods, which now will have access not only to QNames,
but also to array of Element(s) forming the path from the document root to
current element, attributes included.

sten1ee:
Apart from that - yes - this PR does not accomplish anything functionally.
I am more concerned with you being ok with the structural (not functional
so far) changes introduced, so I can continue building on them.
I am more concerned with that than with getting this PR merged.

sten1ee:
How to 'nudge' an integration test that is stuck?

If you see the last travis test -- 2 out of the 8 tests seems to have gone
awry.
E.g. the apple build failed due to:
java.net.UnknownHostException: mirror.ox.ac.uk

duncdrum:
@sten1ee i restarted your build. We're looking into ways so you can do this yourself from here on out. stay tuned ...

sten1ee:
Oopsy daisy.

It is only now that I understand what Adam's comment means:

_My experience is that the indexing pipeline is SAX like (event based), so it is very hard to evaluate complex XPath expressions for the index definitions (which typically rely on a DOM interface). I settled on just allowing a subset of XPath expression._

:/

sten1ee:
Request for help:
Could someone that is familiar with the test cases for Lucene indexing,
craft a test case for the feature implementation by this pull request.
I am pasting here the relevant collection.xconf and then the contents of a
document to be indexed:

    <index xmlns:tei="http://www.tei-c.org/ns/1.0">
        
        <lucene diacritics="no">
            <analyzer
class="org.apache.lucene.analysis.standard.StandardAnalyzer"/>
            <text field="title" match="//tei:title[@xml:lang='Sa-Ltn']"
boost="2.0"/>
        </lucene>
    </index>
</collection>

<?xml-model href="../schema/1.0/tei.rng" schematypens="
http://relaxng.org/ns/structure/1.0"?>
<TEI xmlns="http://www.tei-c.org/ns/1.0">
   <teiHeader>
      <title type="longTitle" xml:lang="Sa-Ltn">Buddha</title> 
      <title type="longTitle"        lang="Sa-Ltn">Dharma</title> 
      <title type="longTitle" xml:lang="En">       Sangha</title> 
      <title type="longTitle">                              Dakini
 </title> 
   </teiHeader>
</TEI>


Kind regards,
/St.

adamretter:
@sten1ee did you get to the bottom of your test problems?

sten1ee:
Yes, yes.
Now the issue with OpenJdk10 is on the radar.

sten1ee:
>
> Answer to Question 1 - What does it do?
>
> You can add a small attribute equals predicate, e.g.:
>
> ```xml
> <text match="//p[@xml:lang = 'Sa-Ltn']"/>
>
> Is that correct?
>
Correct.

> Answer to Question 2 - Multiple predicates?
>
> Is it possible with your current approach to specify multiple predicates,
> e.g.:
>
> <text match="/div[@id = '123']/p[@xml:lang = 'Sa-Ltn']"/>
>
> Yes, (though none of the TC demonstrates this)

> Suggestion 1
>
> At the moment your = operator as really actually eq, and so should be
> renamed as such.
>
> We can also easily support ne, =, and !=.
>
> Later we should consider some further extensions, perhaps adding some
> simple things like:
>
>    1. contains()
>    2. matches()
>    3. lt, le, gt and ge.
>    4. Type constructors, e.g. /p[xs:int(@xml:id) = 123] or
>    `/published[xs:date(@Date <https://github.com/Date>) lt
>    xs:date("2018-09-01")]
>
> For the contains and matches functions, we could abstract and reuse the
> code that operates on Java strings from FunContains.java and
> FunMatches.java into a common utility class. Likely the same thing goes
> for the type constructors.
>
Yes. Support for eq, ne, = and != can be added now in half an hour.
Generally the difficult part of it is crafting a fully fledged XPath parser
to replace the NodePattern.parseXPathExpression() & NodePattern.addSegment()

> Suggestion 2
>
> If I understand correctly, you are caching a copy of the key/value pairs
> of all elements into a HashMap in NodePath2. I am wondering if this is
> really necessary, it seems to me that we will likely memonize things that
> we don't need. Perhaps it would be more efficient to calculate these
> matches in a step fashion in LuceneStreamListener#startElement. Where we
> just eliminate any configs that don't match from a list of applicable
> configs?
>
You're right that this way all the attributes in the current NodePath2 are
being memorized,
which (in case there are *way too many attributes) *will slow down the
(re)indexing process.
In my initial approach, NodePath2 was keeping a reference to each
Node/IStoredNode along the path, so as to make this Node's attributes
available to the match later.
But then I figured out (empirically) that (sadly enough) at the time of
invoking the NodePattern.match() method, the Node attributes were
gone/wiped-out. (feel free to laugh at me here, but I don't have previous
experience with SAX parsers and such). Hence the HashMap.
This is actually the only part of the implementation that makes me go
'Hmmm' and this is why I wanted a code review of the PR as early as
possible.

OTOH the approach that you suggest is beyond 'complicated to implement' (at
least to me), as it requires a major intervention in several big chunks of
code.
And this is only a guess -- I don't fully grok what needs to be done in
order to implement it.

sten1ee:
I have the (bad) feeling that (again) something is expected from me :)
What is it now -- documentation (only) ?

adamretter:
Nope your waiting on us. Sorry I'm very busy and about to move to Malaysia. Should have some time next week. However perhaps @wolfgangmm or @dizzzz could help?

duncdrum:
@sten1ee mostly what adam said, but new features or functionality indeed require documentation. 

so an associated PR [here](https://github.com/eXist-db/documentation/blob/master/src/main/xar-resources/data/lucene/lucene.xml) introducing the new capabilities is needed. You can find some general hints on how to do so in our [author guide](http://exist-db.org/exist/apps/doc/author-reference). If you need any help with the docs feel free to open another issue in the documentation repo and I'll do my best to help you along.

dizzzz:
I am probably not the best guy to judge the functionality in the PR. Process wise all look ok, code too.

extensions/indexes/lucene/src/org/exist/indexing/lucene/NodePathPattern.java
```diff
@@ -0,0 +1,230 @@
...
+        qnPath = new NodePath();
+        parseXPathExpression(namespaces, matchPattern);
+    }
+
+    public NodePathPattern(final QName qname) {
+        qnPath = new NodePath(qname);
+    }
+
+    private void parseXPathExpression(final Map<String, String> namespaces, final String matchPattern) {
+        //TODO : compute better length
```
dizzzz: ToDo?

extensions/indexes/lucene/src/org/exist/indexing/lucene/NodePathPattern.java
```diff
@@ -0,0 +1,230 @@
...
+            qnPath.addComponent(NodePath.WILDCARD);
+        }
+        else {
+            qnPath.addComponent(namespaces, qname);
+        }
+        predicates.add(pred);
+    }
+
+    private Predicate  parsePredicate(String input) {
+        if (!input.startsWith("[") || !input.endsWith("]") || input.charAt(1) != '@')
```
dizzzz: what happens if string length == 0 ? or is that not possible?

extensions/indexes/lucene/src/org/exist/indexing/lucene/NodePathPattern.java
```diff
@@ -0,0 +1,230 @@
...
+            throw new IllegalArgumentException("Bad predicate spec: " + input + "\nOnly [@attr=value] is supported");
+
+        // So far we're supporting only [@attr=value] predicates:
+        int eqIdx = input.indexOf('=');
+        if (eqIdx < 0)
+            throw new IllegalArgumentException("Bad predicate spec: " + input + "\nOnly [@attr=value] is supported");
```
dizzzz: code style: this is not in line with codebase.
```suggestion
if (eqIdx < 0){
            throw new IllegalArgumentException("Bad predicate spec: " + input + "\nOnly [@attr=value] is supported");
            }
```

extensions/indexes/lucene/src/org/exist/indexing/lucene/NodePathPattern.java
```diff
@@ -0,0 +1,230 @@
...
+
+        String name = input.substring(2, eqIdx).trim(); // 2 is to skip the leading [@
+        String val  = input.substring(eqIdx + 1, input.length() - 1).trim(); // -1 is to skip the trailing ]
+
+        if (!(val.startsWith("\'") && val.endsWith("\'") || val.startsWith("\"") && val.endsWith("\"")))
+            throw new IllegalArgumentException("Bad predicate spec: " + input + "\nAttribute value not in quotes");
```
dizzzz: code style ; please use { and }'s

extensions/indexes/range/src/org/exist/indexing/range/RangeIndexWorker.java
```diff
@@ -291,19 +292,19 @@ public ReindexMode getMode() {
 //            return null;
         if (config == null)
             return null;
-        NodePath p = new NodePath(path);
+        NodePath2 p = new NodePath2((NodePath2)path);
         boolean reindexRequired = false;
         if (node.getNodeType() == Node.ELEMENT_NODE && !includeSelf)
-            p.removeLastComponent();
+            p.removeLastNode();
```
dizzzz: { and }'s

src/org/exist/storage/NodePath2.java
```diff
@@ -0,0 +1,144 @@
...
+
+    public NodePath2() {
+        super();
+    }
+
+    public NodePath2(NodePath2 o) {
+        super(o);
+        attribs = new HashMap[o.length()];
+        for (int i = 0; i < attribs.length; i++)
+            attribs[i] = o.attribs(i);
```
dizzzz: { }'s

extensions/indexes/lucene/src/org/exist/indexing/lucene/NodePathPattern.java
sten1ee: NodePathPattern.java is partial copy-paste-edit from NodePath.java
In particular, parseXPathExpression is copy-paste-edit of NodePath's
   private void init(final Map<String, String> namespaces, final String path) {
        //TODO : compute better length
        final FastStringBuffer token = new FastStringBuffer(path.length());

So you see ... I just inherited the TODO from there.
Should have deleted it.
Delete it if you can, please.

extensions/indexes/lucene/src/org/exist/indexing/lucene/NodePathPattern.java
sten1ee: Not possible.
The only code snippet that invokes parsePredicate(...) is this:
    private void  addSegment(final Map<String, String> namespaces, final String segment) {
        String qname;
        int predBeg = segment.indexOf('[');
        Predicate pred = null;
        if (predBeg >= 0) {
            qname = segment.substring(0, predBeg);
            pred = parsePredicate(segment.substring(predBeg));
        }
//
But all in all -- perhaps, I should have added assert input.indexOf('[') >= 0
This way it would have been more informative

extensions/indexes/lucene/src/org/exist/indexing/lucene/NodePathPattern.java
sten1ee: Which is not inline? Throwing IllegalArgumentException ?
I see you have a suggestion here, but I can't understand what the suggestion is (blame it on GitHub)

sten1ee: Aaa, ok, I haven't put braces around

sten1ee: Ok, I don't understand -- you have 
Codacy/PR Quality Review 
and it does not complain about missing braces.
Why is that?

sten1ee:
ok, done

> together with a bug fix

sten1ee:
Travis CI is not green only for the Oracle JDK 10 build failing with:

**oraclejdk10 is deprecated. See https://www.oracle.com/technetwork/java/javase/eol-135779.html for more details. Consider using openjdk10 instead.**

sten1ee:
Travis CI is not green only for the Oracle JDK 10 build failing with:

**oraclejdk10 is deprecated. See https://www.oracle.com/technetwork/java/javase/eol-135779.html for more details. Consider using openjdk10 instead.**

dizzzz:
yes I saw it too :-( I tried to file a ticket to have this fixed but github was kinda down today

dizzzz:
yes I saw it too :-( I tried to file a ticket to have this fixed but github was kinda down today

dizzzz:
yes I saw it too :-( I tried to file a ticket to have this fixed but github was kinda down today

sten1ee:
Travis CI is not green *only* because of Oracle JDK 10 build failing with:

**oraclejdk10 is deprecated. See
https://www.oracle.com/technetwork/java/javase/eol-135779.html for more
details. Consider using openjdk10 instead.**

duncdrum:
@sten1ee the last thing that is missing from my end is the documentation of  the new xpath expressions preferably with usage-examples in the [configuring indexes](http://www.exist-db.org/exist/apps/doc/lucene.xml?field=all&id=D3.15#D3.19) section of the lucene documentation.

sten1ee:
Regarding "needs documentation":
I plan to insert description of this feature somewhere here: https://exist-db.org/exist/apps/doc/lucene#D3.19.58

I have one problem though -- I can't figure out how to locate the doc in question on github :D
Help?

duncdrum:
the article location on GH [https://github.com/eXist-db/documentation/blob/master/src/main/xar-resources/data/lucene/lucene.xml](https://github.com/eXist-db/documentation/blob/master/src/main/xar-resources/data/lucene/lucene.xml) see also this section for [documentation authors](http://exist-db.org/exist/apps/doc/author-reference#D3.17)

sten1ee:
(Not sure if this is the proper way to do it), but I've created the above mentioned PR in documentation.
If that's not the way to do it -- explain, please.

duncdrum:
@dizzzz @sten1ee I thought this would have made it into the last release. There don't seem to be any outstanding issues except for backporting to 4.x and 5.x branches ?

sten1ee:
@duncdrum i am not supposed to backport this to 4.x.x and 5.0.0, am I?

duncdrum:
@sten1ee sort of yes I m afriad. While we are working on a release candidate, ongoing releases are cut from these two branches. To backport to e.g. ` 4.x.x` you need to create  feature branches  of `develop-4.x.x` e.g.  `4-Phase1` and cherry pick the commits from this PR into it. Then open a PR referencing this PR in the commit message `backport of #2169`

Then rinse and repeat for 5.

As I said before you might want to rebase this PR first and squash a few intermediate commits.

domlatham:
@duncdrum is there some documentation of how we get the feature into a release please? We're a little confused. We are currently on 3.4.1 and we have that working but of course we don't want to get stuck. So we would like it to get into a future release at some point but we don't necessarily need to backport to 4.x.x. Will it eventually end up in 5.something if we just leave it or do we have to add a PR for 5.x.x to get it included at all?
Thanks.

duncdrum:
@domlatham our [release documentation](https://github.com/eXist-db/exist/blob/develop/exist-versioning-release.md) is here in the repo

As I said before, while we are in release-candidate mode, releases are cut from the branches `develop-4.x.x` and `develop-5.0.0`, this is the exception but necessary for 5.0.0

For the long term maintainability of your contribution, and since you currently have a conflict to resolve, you should first [rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) this PR (make sure you have a backup). When you do I would recommend to clean up the commit history to make it shorter at the same time. Commits like `ooops` aren't very helpful when understanding a code contribution. 

Once you wiggled this down to a few commits, you can cherry-pick these onto develop-4, and develop-5 branches. Once this is done, you should open two more PRs (1 per branch). The next release from develop branch is still ways off, and given that you already have a conflict, you ll have to rebase before we can merge this then anyway.

Tl;dr: Your options: 
-   rebase now and keep rebasing until 5.0.0 is released and we switch back to just the develop branch (bad idea), 
-   rebase now and prepare actionable backport forwardpot PRs to go into the next pair of releases (better idea).

sten1ee:
@duncdrum
When you say:
```you should first rebase this PR (make sure you have a backup)```
You mean (clean &) rebase onto develop, right?

duncdrum:
@sten1ee yes, that would be the simplest way imv

adamretter:
I am still concerned about the memory use that I raised in "Suggestion 2" here - https://github.com/eXist-db/exist/pull/2169#issuecomment-430505011

Perhaps it is better just to accept this and fix it later if anyone complains? @wolfgangmm your thoughts please? You have been noticeably absent from this ;-)

PR #2175, opened 2018-09-27
5xx: Fix handling of / when context item was declared, closes #2156

### LINKED ISSUE #2156: Bug in eXist's handling of / when a context item has been declared

dizzzz:
sync of #2156

### DIFF
--- a/src/org/exist/xquery/RootNode.java
+++ b/src/org/exist/xquery/RootNode.java
@@ -32,9 +32,7 @@ import org.exist.storage.UpdateListener;
 import org.exist.storage.lock.ManagedDocumentLock;
 import org.exist.util.LockException;
 import org.exist.xquery.util.ExpressionDumper;
-import org.exist.xquery.value.Item;
-import org.exist.xquery.value.Sequence;
-import org.exist.xquery.value.Type;
+import org.exist.xquery.value.*;
 
 import java.util.Iterator;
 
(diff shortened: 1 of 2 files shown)


dizzzz:
where are the travis checks?

dizzzz:
travis is OK: https://travis-ci.org/eXist-db/exist/builds/434311433

joewiz:
@dizzzz @wolfgangmm Thank you!

Issue #2195, opened 2018-10-08
Error on taking a subsequence of stored nodes


ptrckb:
Thank you for reporting your issue!

To be able to better understand you problem or suggestion, please add as much information as possible to this ticket. When there is no or just little information, it will be too difficult to work on it (and hence we will not be able to solve it)

Please fill in the following sections:

### What is the problem
> Describe exactly what you see (e.g. an output of an XQuery) 

Since #2138 I am experiencing problems with taking a subsequence of nodes from a resource (stored?).

### What did you expect
> Describe what you expected to happen. Add for example a reference to a [specification](https://www.w3.org/TR/xquery-3/).

No error, the text node of the second 'x' element with the code below.

### Describe how to reproduce or add a test
> Describe how we can can reproduce the problem.

> The *best* way is to provide a 'self containing' test, a test that can be run without dependancies. The [XQSuite - Annotation-based Test Framework for XQuery](http://exist-db.org/exist/apps/doc/xqsuite.xml) makes it very easy for you to create tests. These tests can be executed from the [eXide editor](http://exist-db.org/exist/apps/eXide/index.html) (XQuery - Run as Test)

Running the following script:
```xquery
xquery version "3.1";

let $x := doc('/db/apps/test.xml')/test/x
let $xx := subsequence($x, 2, 1)
return $xx/text()
```

on the document (`/db/apps/text.xml`):
```xml
<test>
    <x>1</x>
    <x>2</x>
    <x>3</x>
</test>
```

creates the error:
```
exerr:ERROR Type error: the sub-sequence cannot be converted into a MemoryNodeSet. It contains nodes from stored resources.
```

With any build from source prior to ace6379f4d034fb1b5db61d1160f5a6c764d59af it returns:
```
2
```

### Context information
Please always add the following information
- eXist-db version + Git Revision hash e.g. eXist-db 3.0 /  acd0c14 
4.4.0 or develop-4.x.x (608991ac3e4ada8e887a516f58cac41af3ea4809)
- Java version (e.g. Java8u121)
1.8.0
- Operating system (Windows 7, Linux, MacOs)
Linux
- 32 or 64 bit
64 bits
- How is eXist-db installed? (JAR installer, DMG, WAR file, clone from GitHub)
clone from GitHub
- Any custom changes in e.g. conf.xml
none

PR #1133, opened 2016-09-22
XPTY0018 should only be raised for Path Operations

adamretter:
Closes https://github.com/eXist-db/exist/issues/1121

PR #2196, opened 2018-10-08
Fixes an issue with mixed types in a subsequence

adamretter:
Closes https://github.com/eXist-db/exist/issues/2195

### DIFF
--- a/src/org/exist/xquery/PathExpr.java
+++ b/src/org/exist/xquery/PathExpr.java
@@ -297,7 +297,7 @@ public class PathExpr extends AbstractExpression implements CompiledXQuery,
 
             final boolean allowMixedNodesInReturn;
             if(prev != null) {
-                allowMixedNodesInReturn = prev.allowMixedNodesInReturn() | expr.allowMixedNodesInReturn();
+                allowMixedNodesInReturn = prev.allowMixedNodesInReturn() /*| expr.allowMixedNodesInReturn()*/;
             } else {
                 allowMixedNodesInReturn = expr.allowMixedNodesInReturn();
             }
--- a/src/org/exist/xquery/PathExpr.java
+++ b/src/org/exist/xquery/PathExpr.java
@@ -297,7 +297,7 @@ public class PathExpr extends AbstractExpression implements CompiledXQuery,
 
             final boolean allowMixedNodesInReturn;
             if(prev != null) {
-                allowMixedNodesInReturn = prev.allowMixedNodesInReturn() /*| expr.allowMixedNodesInReturn()*/;
+                allowMixedNodesInReturn = prev.allowMixedNodesInReturn() | expr.allowMixedNodesInReturn();
             } else {
                 allowMixedNodesInReturn = expr.allowMixedNodesInReturn();
             }
--- a/src/org/exist/xquery/Function.java
+++ b/src/org/exist/xquery/Function.java
@@ -439,12 +439,6 @@ public abstract class Function extends PathExpr {
     }
 
     @Override
-    public boolean allowMixedNodesInReturn() {
-        final int returnType = mySignature.getReturnType().getPrimaryType();
-        return returnType == Type.ANY_TYPE || returnType == Type.ITEM || returnType == Type.UNTYPED;
-    }
-
-    @Override
     public String toString() {
         final StringBuilder result = new StringBuilder();
         result.append(getName());
--- a/src/org/exist/xquery/Function.java
+++ b/src/org/exist/xquery/Function.java
@@ -439,6 +439,12 @@ public abstract class Function extends PathExpr {
     }
 
     @Override
+    public boolean allowMixedNodesInReturn() {
+        final int returnType = mySignature.getReturnType().getPrimaryType();
+        return returnType == Type.ANY_TYPE || returnType == Type.ITEM || returnType == Type.UNTYPED;
+    }
+
+    @Override
     public String toString() {
         final StringBuilder result = new StringBuilder();
         result.append(getName());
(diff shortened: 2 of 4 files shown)


src/org/exist/xquery/PathExpr.java
```diff
@@ -297,7 +297,7 @@ public Sequence eval(Sequence contextSequence, final Item contextItem) throws XP
 
             final boolean allowMixedNodesInReturn;
             if(prev != null) {
-                allowMixedNodesInReturn = prev.allowMixedNodesInReturn() | expr.allowMixedNodesInReturn();
+                allowMixedNodesInReturn = prev.allowMixedNodesInReturn() /*| expr.allowMixedNodesInReturn()*/;
```
line-o: @wolfgangmm it is unclear why this line needed to be changed for the test to succeed.

wolfgangmm: @adamretter running the test on my install, above comment was not needed. The test passed without it. 

I did not contribute to the code calling allowMixedNodesInReturn, but I can see some sense in it. It's not vital though. I would thus say, if your test passes without commenting out the test above, we should leave PathExpr untouched. If it gets in your way though, we need to reconsider.

adamretter:
@juri not intentionally!

src/org/exist/xquery/PathExpr.java
adamretter: @wolfgangmm Sure the contributed test is fine. But one of the existing tests in the test suite fails without commenting that out (due to my override in `Function.java`).

juri:
@adamretter wrong person in the mention, I think…

src/org/exist/xquery/PathExpr.java
wolfgangmm: Ok, so the error expected by `XQueryTest#pathOperatorContainingNodesAndNonNodes()` is in line with what saxon reports. I wonder why you had to overwrite `Function#allowMixedNodesInReturn()`? It does not seem related to the original issue.

wolfgangmm: Discard the question, I see it. This is quite a puzzle ...

wolfgangmm:
@adamretter ok, after some hours I think I found the correct fix. Please try to change `SubSequence.getItemType()` to return the correct type of the underlying sequence:

```java
@Override
    public int getItemType() {
        return sequence.getItemType();
    }
```

With this in place you do not need to overwrite `Function.allowMixedNodesInReturn()` or `PathExpr`. Tests should pass. If this fails lazy evaluation, we will need another way to determine the expected type.

adamretter:
@wolfgangmm So I don't think that will work for the very common `ValueSequence` which will still return `Type.ITEM`.

Is there anything wrong with my proposed changes. No tests fail. Are there things that you know that this breaks? If so, we should start by adding tests for those things...

wolfgangmm:
`ValueSequence` does compute its type, so should be fine.

You wanted me to look into this because you had to comment out the check in `PathExpr`. This was originally added by yourself, so I assume you had a reason to do so. I have thus tried to find a solution which does not require the changes to `Function` and `PathExpr`.

All in all returning the correct type seems a more correct approach.

adamretter:
@wolfgangmm Okay, I had no idea I added that check. I guess I did so to fix some broken test.

Regards`ValueSequence` it uses a common super-type. So if you have a `node()` and an `xs:string` in the value sequence, the common super type is `item()`... no?

wolfgangmm:
See #1133, which introduced the changes in question to `PathExpr`. I'm fine to just drop code nobody can explain, but you asked about that particular line, so I thought it is not ok and we need a solution.

wolfgangmm:
If the sequence is indeed mixed, it is correct to return `Type.ITEM` and in this case the check in `PathExpr` should fail as it would.

adamretter:
@wolfgangmm I just spotted a problem with your suggestion to change `SequenceType#getItemType()` to `return sequence.getItemType();`.

When given code like:
```xquery
let $seq := (<a/>, <b/>, 1, 2)
return
  subsequece($seq, 1,2)
```
Your change would mean that we return `Type.ITEM` from the underlying value sequence, but the subsequence should actually return `Type.NODE` (or even `Type.ELEMENT`).

We cannot infer the type from the subequence without actually iterating all of the items in the subsequence and checking their common super type. We really want to avoid that, as a lazy approach is much more performant and memory kind.

Thoughts?

adamretter:
> So even if you have to iterate, you do not need to continue beyond item 2, which means the performance impact should be small?

heh. Sure that works for a small subsequence, but what about a sequence of 20,000 items, where you want the first `999` ?

wolfgangmm:
> heh. Sure that works for a small subsequence, but what about a sequence of `20,000` items I , where you want the first `999` ?

In this case you only need to determine the type of the first 999, so there's no performance loss, right?

> I guess I am trying to figure out if we _have_ to take determine the type as you are suggesting, or if you think that the fix I proposed previously has a problem?

The whole issue was that using `subsequence` as a path step *should* result in an error if atomic values are returned. At least this is why `PathExpr` tries to determine the type. So `subsequence((<a>a</a>, "a"), 1, 2)/text()` would be an error. We could argue that those checks should actually be done elsewhere and it's not the job of subsequence to determine the type. 

However, I think having correct type information returned by expressions where possible is always preferable to not having the information.

wolfgangmm:
I guess what I'm thinking is: if there is an easy and fast way to determine the type of the result of subsequence, we should rather go for it instead of having to modify the superclass `Function` and `PathExpr`.

But if it means a performance loss, I'm rather for speed.

adamretter:
@wolfgangmm I have commited the change you suggested. It looks good. Thank you :-)

PR #2198, opened 2018-10-08
(4.x.x) Fixes an issue with mixed types in a subsequence

adamretter:
Backport of https://github.com/eXist-db/exist/pull/2196

### DIFF
--- a/src/org/exist/xquery/value/SubSequence.java
+++ b/src/org/exist/xquery/value/SubSequence.java
@@ -361,7 +361,26 @@ public class SubSequence extends AbstractSequence {
 
     @Override
     public boolean isPersistentSet() {
-        return false;
+        final SequenceIterator iterator;
+        try {
+             iterator = iterate();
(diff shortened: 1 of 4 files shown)


src/org/exist/xquery/value/SubSequence.java
```diff
@@ -361,7 +361,26 @@ public DocumentSet getDocumentSet() {
 
     @Override
     public boolean isPersistentSet() {
-        return false;
+        final SequenceIterator iterator;
+        try {
+             iterator = iterate();
+        } catch (final XPathException e) {
+            throw new RuntimeException(e); // should never happen!
```
dizzzz: isn't this dangerous ?

dizzzz: as in: better to log a bing ERROR and return false ?

dizzzz: Oki

Issue #2186, opened 2018-10-03
REST interface put method failing


eriksiegel:
### What is the problem
We have code that copies files from one server to the other using the standard REST interface functionality eXist provides. Here is a dressed down code example:

```
let $full-path-source := '/db/TEST/somefile.xml'
let $full-path-dest := $full-path-source

let $server := 'server-name-or-ip' 

let $request :=
  <hc:request href="http://{$server}:8080/exist/rest{$full-path-dest}" method="put" auth-method="basic" 
      username="admin" password="adminpassword" send-authorization="true">
    <hc:body media-type="{xmldb:get-mime-type(xs:anyURI($full-path-source))}"/>
  </hc:request>

return
  hc:send-request($request, (), doc($full-path-source))
```
The environment is all ok as far as I can see. The source file exists, is well-formed, the destination collection exists and is writable.

Since using 4.4.0 there are problems with this:

- When the file to send is bigger than some threshold (an XML file of 84Kb fails, a very small one succeeds, usually...) an error pops up
- The error is, rather puzzling: Error serializing the body content (on the hc:send-request()). 
- But there is nothing wrong with the file and why wouldn't it be able to serialize? It seems like a wrong message for some other problem occurring.
- Very rarely the bug does not happen and the file transfers...
- The source is a WIndows machine, the destination a Unix machine, all running 4.4.0
- To make it even more strange: When I do the same with a Windows machine as destination, my whole source system hangs up
- There is nothing in the exist.log about this
- We run the script from within oXygen but also using a rest call to it
- We had no problems with 4.3.1

### What did you expect
File copied to remote system

### Describe how to reproduce or add a test

Have two separate servers running eXist, preferably the source on Windows and the destination on Unix
Adapt the script above with the right stuff (address, password, etc).
Try it to run on the source machine

### Context information
Please always add the following information
- eXist version: 4.4.0 (Build 201809211803)
- Java version 1.8.0_181
- Operating system W10 (source) and Unix (dest)
- 64 bit
- JAR installed
- Any custom changes in e.g. conf.xml? 
  - Data and Journal directory moved to outside eXist installation directory
  - Global cache module enables
  - RestXQ and expath auto deploy trigger disabled

duncdrum:
Hi Eric,  just to make sure:  are these machines behind a reverse-proxy,  could that limit the max size allowed for file transfers (i know nginx does that by default)?

eriksiegel:
[...]

PR #2207, opened 2018-10-11
Fix for preemptive authentication in the EXPath HTTP Client

adamretter:
Related to: https://github.com/eXist-db/exist/issues/2186


dizzzz:
nice catch

Issue #2203, opened 2018-10-08
map with duplicate keys should throw an error


line-o:
### What is the problem

The following code

```xquery
map { 'a': 1, 'a': 2 }
```

evaluates to

```xquery
map { 'a': 2 }
```

### What did you expect

An error should be raised, that key 'a' was entered twice.
BaseX does exactly that:

```
XQUERY map {'a':1, 'a':2}
Stopped at ., 1/5:
[XQDY0137] Key "a" already exists in map (values: 1 vs. 2).
```

This should not affect `map:put` or `map:merge` where this actually overwrites the given key in the returned, new map.

### Context information

- eXist-db 5.0.0-RC3 /  bfdcc8406 
- Java version: 1.8.0_152
- Operating system: MacOS (Mojave)
- build from source / clone from GitHub

duncdrum:
Yup, the [specs](https://www.w3.org/TR/xquery-31/#id-map-constructors) are pretty unambiguous.

PR #2213, opened 2018-10-11
Map Constructor should not allow duplicate keys

adamretter:
Closes https://github.com/eXist-db/exist/issues/2203

### DIFF
--- a/src/org/exist/xquery/ErrorCodes.java
+++ b/src/org/exist/xquery/ErrorCodes.java
@@ -131,8 +131,9 @@ public class ErrorCodes {
             "Bind the prefix xmlns to any namespace URI.\n" +
             "Bind a prefix to the namespace URI http://www.w3.org/2000/xmlns/.\n" +
             "Bind any prefix (including the empty prefix) to a zero-length namespace URI.");
-    public static final ErrorCode XQDY0138 = new W3CErrorCode("XQDY0138", "Position n does not exist in this array");
     public static final ErrorCode XQDY0102 = new W3CErrorCode("XQDY0102", "If the name of an element in an element constructor is in no namespace, creating a default namespace for that element using a computed namespace constructor is an error.");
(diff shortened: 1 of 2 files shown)


.idea/libraries/expath.xml
```diff
@@ -4,7 +4,9 @@
       <root url="file://$PROJECT_DIR$/extensions/expath/lib" />
     </CLASSES>
     <JAVADOC />
-    <SOURCES />
+    <SOURCES>
+      <root url="file://$PROJECT_DIR$/../expath-http-client-java/http-client-java/src/main/java" />
+    </SOURCES>
```
joewiz: Was this intended to be included?

adamretter: Oops. Have now fixed.

joewiz: ;)

.idea/libraries/expath.xml
```diff
@@ -7,4 +7,4 @@
     <SOURCES />
     <jarDirectory url="file://$PROJECT_DIR$/extensions/expath/lib" recursive="false" />
   </library>
-</component>
\ No newline at end of file
+</component>
```
joewiz: No biggie, but apparently a small whitespace change remained...

wolfgangmm:
Somehow all the travis builds seem to be hanging?

adamretter:
@wolfgangmm It is not hung it is queued. See - https://travis-ci.com/eXist-db/exist/builds

I am not sure if after moving from travis-ci.org to travis-ci.com we have less build workers available, but it looks like that way, all tasks used to build in parallel, but now it looks like just 2 workers. Maybe @duncdrum knows something?

duncdrum:
@adam I noticed that as well. I m not aware of any previous arrangements to get more workers? We are now at what I understand to be the default for oss.

PR #2216, opened 2018-10-14
(4.x.x) Fix thread naming

adamretter:
This PR address the naming of threads and thread grouping.

Threads which are per-database instance (i.e. per BrokerPool) and now:
1. Placed into a thread group specific to that BrokerPool. The thread group is named like: `exist.db.${instanceId}`.
2. Threads are named like: `db.${instanceId}.${threadName}`.

Threads which are global, i.e. across broker pools, are named like: `global.${threadName}`.

This makes identifying threads when running multiple eXist-db instances within the same JVM much simpler.

I also added a test `ConcurrentBrokerPoolTest` which shows that this should all work just fine.


src/org/exist/client/InteractiveClient.java
```diff
@@ -188,6 +188,9 @@
     protected XmldbURI path = XmldbURI.ROOT_COLLECTION_URI;
     private Optional<Writer> lazyTraceWriter = Optional.empty();
 
+    private static final NamedThreadGroupFactory clientThreadGroupFactory = new NamedThreadGroupFactory("javaadminclient");
```
dizzzz: better name = java-adminclient ?

dizzzz: just a suggestion

adamretter: Okay fixed. I changed it to "java-admin-client"

src/org/exist/xquery/functions/util/Eval.java
```diff
@@ -147,7 +136,8 @@
             new QName("eval-async", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
             "Dynamically evaluates an XPath/XQuery expression asynchronously. The ID of the executing thread is returned.",
             new SequenceType[] { EVAL_ARGUMENT },
-            RETURN_NODE_TYPE),
+            RETURN_NODE_TYPE,
+            "The implementation of eval-async was unreliable. eval-async now defers to eval!"),
```
dizzzz: i did not realise that the implementation is not stable.....

adamretter: It would stabily screw you up ;-)

PR #2241, opened 2018-10-26
Fix several Journalling and Recovery issues for XML and Binary documents

adamretter:
Closes https://github.com/eXist-db/exist/pull/1838

This fixes several journalling and recovery issues for both XML and Binary documents. Most issues were previously identified in #1838. Further tests have been added to this PR.

* This PR does not address the issue with crash-safe concurrent access to resources in the database. Those issues are documented as `@Ignore` in both `test/src/org/exist/storage/journal/AbstractJournalTest.java` and `test/src/org/exist/storage/AbstractRecoverTest.java`.

    I plan to address such issues shortly in an separate PR which will lock resources for the duration of a concurrent transaction.

Most issues in this PR have had at least some brief discussion between myself and @wolfgangmm, so there should be no real surprises!

**NOTE**: This PR changes the format of eXist-db's Journal file on disk. To upgrade to this version of eXist-db, the user should perform a full-backup and restore if they have `.log` files in `$EXIST_HOME/webapp/WEB-INF/data`.


joewiz:
> This PR changes the format of eXist-db's Journal file on disk

Given this, is this PR still ok for a minor release, or does it need to be deferred to a major release?

duncdrum:
@adamretter so this could potentially kill all containers that run with persistent data volumes? We need to coordinate the release very closely, between here and the docker registry.

adamretter:
@wolfgangmm Oops. The NPE was actually masking the informative message that tells you that the Journal version is wrong. I think I have fixed that now.

wolfgangmm:
Ok, works fine for me now. I tested crashing my database a few times on different machines.

PR #2250, opened 2018-10-31
Develop: Update location of JAI downloads

dizzzz:
textual change


joewiz:
Hasn't cleared CI, but this is just a markdown edit, so I'll merge.

PR #2254, opened 2018-11-01
Journal Checksums

adamretter:
This PR adds a checksum to the Journal for each entry. The checksum of the entry is then compared when reading the Journal back for recovery. This enables us to ensure that the Journal file is not corrupt.

I have used the xxhash-64 algorithm from the lz4 project, which in my testing for the typical size of journal entries was faster than lz4's xxhash-32, Java's Adler32, and OpenHFT's no-allocation xxhash.

fixed-hash-benchmarks

**NOTE** This PR also increments the Journal file format version to version `3`.


src/org/exist/storage/journal/JournalReader.java
```diff
@@ -218,23 +235,51 @@ Loggable readEntry() throws LogException {
...
+            // read entry backLink
+            final short backLink = payload.getShort();
+            if (backLink != size + LOG_ENTRY_HEADER_LEN) {
+                LOG.error("Bad pointer to previous: backLink = " + backLink + "; size = " + size +
                         "; transactId = " + transactId);
                 throw new LogException("Bad pointer to previous in entry: " + loggable.dump());
             }
+
+            // update the checksum for the entry data and backLink
+            if (payload.hasArray()) {
```
dizzzz: looks like almost identical code of line 211 ?

dizzzz:
see my remark; I'd guess some longer stress testing might needed here?

src/org/exist/storage/journal/JournalReader.java
adamretter: I could abstract it, but I would need access to several local vars, as it was small I just reproduced it.

adamretter:
@dizzzz some replies to your comments:
> Wondering if a journal is not completely written due e.g. a JVM crash, will the loader robust for this? 

No. At the moment, we just detect the failure during recovery and abort. However, recovery will then only have been done as far as the checksum failure. This could lead to inconsistencies in the recovered data. Consider this an improvement on what we had, whereby previously we might ignore journal corruptions and report that recovery completed.

We have two other options which would be more robust:

1. Do a pre-scan of each journal file before recovery, checking that the checksums are valid.
2. and/or switch to a fixed block size based journal format such as that used in LevelDB, this enables us to more easily detect corruptions due to the framing.

> wouldn't it be safer to write the checksum before the larger payload data in writeToLog() ?

I am not sure it makes much difference. If there is a corruption, regardless of whether the checksum is before or after the payload, it will not match.

> how large can the payload become?

The older journal code is a bit yucky and won't cope with payloads greater than 32 KB. In reality the payloads written by eXist-db are always under 4 KB which is the B-Tree page size.

PR #2262, opened 2018-11-05
Improve the provenance of the XML:DB API dependency

adamretter:
Jar Binary from the established source code provenance - github.com/exist-db/xmldb-org


src/org/exist/xmldb/LocalResourceSet.java
```diff
@@ -87,6 +87,13 @@ public void addResource(final Resource resource) throws XMLDBException {
         resources.add(resource);
     }
 
+    @Override
```
dizzzz: why is this needed?

adamretter: The "official" API has this, it was really the only difference apart from `int` vs `long` for some sizes.

PR #2271, opened 2018-11-07
Fix verbose logging regression in ExistSAXParserFactory

adamretter:


adamretter:
@joewiz Yes that is because they are separate PR's ;-)

PR #2273, opened 2018-11-07
(4.x.x) Fix verbose logging regression in ExistSAXParserFactory

adamretter:
Backport of https://github.com/eXist-db/exist/pull/2271

### DIFF
--- a/src/org/exist/util/ExistSAXParserFactory.java
+++ b/src/org/exist/util/ExistSAXParserFactory.java
@@ -53,7 +53,9 @@ public class ExistSAXParserFactory {
 
         } catch (final ClassNotFoundException ex) {
             // quick escape
-            LOG.debug(className + ": " + ex.getMessage(), ex);
+            if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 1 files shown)


src/org/exist/util/ExistSAXParserFactory.java
```diff
@@ -53,7 +53,9 @@ public static SAXParserFactory getSAXParserFactory(final String className) {
 
         } catch (final ClassNotFoundException ex) {
             // quick escape
-            LOG.debug(className + ": " + ex.getMessage(), ex);
+            if (LOG.isDebugEnabled()) {
+                LOG.debug(className + ": " + ex.getMessage(), ex);
```
dizzzz: probably is with only

`LOG.debug("{}: {}", className, ex.getMessage(),ex)` almost as fast...?

adamretter: Nope. Without the `if` getMessage is always evaluated even when debug logging is disabled.

PR #2140, opened 2018-09-03
XQSuite+XSuite was not correctly reporting all test failures

adamretter:
This fixes an issue raised by @dizzzz where not all test failures were reported from XQSuite correctly, which caused XSuite to show tests as passing when they were actually failing.

PR #2279, opened 2018-11-08
Feature/lucene search leading wildcard

tuurma:
### Description:

ft:search currently doesn't allow specifying options, which in particular makes leading-wildcard queries impossible on computed index fields

### Reference:

no issue has been registered so far afaik

### Type of tests:

Relevant test has been attached with this PR

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
@@ -669,7 +669,7 @@ public class LuceneIndexWorker implements OrderedValuesIndex, QNamedKeysIndex {
      * @param queryText
      * @return search report
      */
-    public NodeImpl search(final XQueryContext context, final List<String> toBeMatchedURIs, String queryText, String[] fieldsToGet) throws XPathException, IOException {
+    public NodeImpl search(final XQueryContext context, final List<String> toBeMatchedURIs, String queryText, String[] fieldsToGet, Properties options) throws XPathException, IOException {
 
         return index.withSearcher(searcher -> {
             // Get analyzer : to be retrieved from configuration
@@ -677,7 +677,11 @@ public class LuceneIndexWorker implements OrderedValuesIndex, QNamedKeysIndex {
 
             // Setup query Version, default field, analyzer
             final QueryParserWrapper parser = getQueryParser("", searchAnalyzer, null);
-            final Query query = parser.parse(queryText);
+            try {
+                setOptions(options, parser.getConfiguration());
+            } catch (ParseException e) {
+                throw new XPathException("Lucene query syntax error: " + e.getMessage());
+            }       final Query query = parser.parse(queryText);
 
             // extract all used fields from query
             final String[] fields;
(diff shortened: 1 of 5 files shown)


extensions/indexes/lucene/test/src/xquery/lucene/binary.xql
```diff
@@ -65,4 +65,12 @@ function luct:check-visibility-collection-pass() {
     system:as-user("admin", "", 
         ft:search("/db/lucenetest/sub/", "title:admin")/search/@uri/string()
     )
+};
+
+declare
+    %test:assertEquals("Lorem")
```
joewiz: A very useful addition! One question: wouldn't this test return the URI `/db/lucenetest/test.txt` rather than the contents of the field?

tuurma: true, interesting that the test passes...

joewiz: I bet the failure is being masked by https://github.com/eXist-db/exist/pull/2140.

duncdrum: @joewiz maybe…

adamretter: @joewiz Likely indeed that #2140 has to be fixed first

extensions/indexes/lucene/src/org/exist/indexing/lucene/LuceneIndexWorker.java
```diff
@@ -669,15 +669,19 @@ public void writeNonXML() {
...
             final Analyzer searchAnalyzer = new StandardAnalyzer(LuceneIndex.LUCENE_VERSION_IN_USE);
 
             // Setup query Version, default field, analyzer
             final QueryParserWrapper parser = getQueryParser("", searchAnalyzer, null);
-            final Query query = parser.parse(queryText);
+            try {
+                setOptions(options, parser.getConfiguration());
+            } catch (ParseException e) {
+                throw new XPathException("Lucene query syntax error: " + e.getMessage());
+            }       final Query query = parser.parse(queryText);
```
dizzzz: formatting is not ok

extensions/indexes/lucene/test/src/xquery/lucene/binary.xql
adamretter: @tuurma Would you be able to get @wolfgangmm to take a look at #2140 first please?

joewiz:
@tuurma Will you be able to prepare ports of this PR, or do you need assistance with this? It would be great to be able to merge this in, but we need all ports in place beforehand.

dizzzz:
if you want I can do this.... it is very little work

tuurma:
Backports provided in #2467 and #2468

adamretter:
Requires #2140 first!

adamretter:
@tuurma #2140 is now merged. Could please rebase this on the develop HEAD? We can then look to get this PR merged.

adamretter:
@tuurma I am afraid that your rebase doesn't look like it was correct, this PR now incorporates another PR from @windauer. I think it is @windauer's PR which is also now causing CI to fail for this.

@tuurma would you like some help rebasing?

tuurma:
@adamretter I don't think there's any problem with the rebase, it does incorporate @windauer's commit because he was the one to write the tests

Reason why tests fail here is there has been a problem with them, only resolved with 4f60eeea0e9a20fe97b1e3fe5721877a59ae3732 I'm not sure what's the correct approach now.

tuurma:
Sorry @adamretter, now I see what the problem is

PR #2292, opened 2018-11-12
(4.x.x) Remove the 2GB log file limit for the Journal

adamretter:
Previously eXist-db would silently write bad entries to the Journal if the Journal file exceeded 2 GB. I recently added a check which would detect the situation and raise an error to avoid silent corruption.

I had two reports from different individuals, both saying they had encountered the error! It seems that it is relatively easy to store enough data into eXist-db in a single transaction to exceed the 2 GB limit. 

Previously the limit was caused by the use of 4 bytes for storing the file position of the Journal entry within the LSN (Log Sequence Number). I have now increased that to 8 bytes, which allows for Journal files as large as the JVM can support, i.e. 8,191 PB. It is unlikely that will ever be exceeded! ;-)

Unfortunately fixing this was not straightforward due to a number of assumptions made in some of eXist-db's older code.

**NOTE:** This patch increments the file version of all `.dbx` and `.log` files by one, as the storage format needed to change to accomodate the larger LSN. A full database backup and restore is required to use this branch.

### DIFF
--- a/extensions/indexes/ngram/src/org/exist/indexing/ngram/NGramIndex.java
+++ b/extensions/indexes/ngram/src/org/exist/indexing/ngram/NGramIndex.java
@@ -42,7 +42,7 @@ import java.nio.file.Path;
  */
 public class NGramIndex extends AbstractIndex implements RawBackupSupport {
 
-    public static final short FILE_FORMAT_VERSION_ID = 13;
+    public static final short FILE_FORMAT_VERSION_ID = 14;
 
     public final static String ID = NGramIndex.class.getName();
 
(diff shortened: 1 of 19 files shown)


duncdrum:
> which allows for Journal files as large as the JVM can support, i.e. 8,191 PB. It is unlikely that will ever be exceeded! 

challenge acc…. nvm

linxd:
reindex on large number of file and values my cause this error occur.

linxd:
reindex on large number of files with many index value can cause this error

adamretter:
@linxd does this fix your problem?

duncdrum:
still needs 5.x.x and develop ports

adamretter:
@linxd Please stay on topic. You are mixing this with other problems!

linxd:
Thanks. now it's ok

在 2018-11-15 00:08:22，"Adam Retter" <notifications@github.com> 写道：


@linxd Please stay on topic. You are mixing this with other problems!

PR #2142, opened 2018-09-03
(4.x.x) XQSuite+XSuite was not correctly reporting all test failures

adamretter:
Backport of https://github.com/eXist-db/exist/pull/2140

PR #2296, opened 2018-11-15
(4.x.x) Fix failing xml:id tests by using proper checks for xs:NCName

wolfgangmm:
Fixes the failing tests on xml:id related functions demonstrated by #2142.

* `fn:id` lookups on in-memory documents basically never worked
* tests for NCName were wrong and are replaced by an implementation more closely following the character ranges defined in the specs (https://www.w3.org/TR/REC-xml/#NT-Name)

*Do **not** merge before all issues in #2142 have been addressed!*


adamretter:
After discussion with @wolfgangmm during a Community Call, it was decided these could be merged sooner than the XQSuite fix. I also agreed to provide the backports.

PR #2297, opened 2018-11-16
(4.x.x) XQSuite tests relying on the query trace log were failing

wolfgangmm:
This PR fixes most of the failing tests concerning query optimization as revealed by #2142. Reason for the failure was that we were instantiating a different `Profiler` instance for the dynamically imported module, so stats were not available to the main query running XQSuite.

*Do **not** merge before all issues in #2142 have been addressed!*


adamretter:
After discussion with @wolfgangmm during a Community Call, it was decided these could be merged sooner than the XQSuite fix. I also agreed to provide the backports.

Issue #2304, opened 2018-11-20
Insufficient synchronization in TransactionManager


adamretter:
The internal state of the transaction manager is read and modified by multiple threads, sometimes with a lock and sometimes without. This should be fixed so that it is consistent and correct memory affects are seen.

1. `TransactionManager#transactions` is potentially accessed without sufficient synchronisation:
    1. https://github.com/eXist-db/exist/blob/eXist-4.4.0/src/org/exist/storage/txn/TransactionManager.java#L200
    2. https://github.com/eXist-db/exist/blob/eXist-4.4.0/src/org/exist/storage/txn/TransactionManager.java#L253
    3. https://github.com/eXist-db/exist/blob/eXist-4.4.0/src/org/exist/storage/txn/TransactionManager.java#L264
    4. https://github.com/eXist-db/exist/blob/eXist-4.4.0/src/org/exist/storage/txn/TransactionManager.java#L269
    5. https://github.com/eXist-db/exist/blob/eXist-4.4.0/src/org/exist/storage/txn/TransactionManager.java#L272
    6. https://github.com/eXist-db/exist/blob/eXist-4.4.0/src/org/exist/storage/txn/TransactionManager.java#L299

2. `TransactionManager#nextTxnId` is potentially accessed without sufficient synchronisation:
    1. https://github.com/eXist-db/exist/blob/eXist-4.4.0/src/org/exist/storage/txn/TransactionManager.java#L225
    2. https://github.com/eXist-db/exist/blob/eXist-4.4.0/src/org/exist/storage/txn/TransactionManager.java#L260

PR #2305, opened 2018-11-20
(4.x.x) New implementation of the Transaction Manager

adamretter:
Closes https://github.com/eXist-db/exist/issues/2304

1. Fixes the potential concurrency safety issues in the current transaction manager.
2. New design is non-blocking lock-free.
3. New design does not need to acquire a broker from the Broker Pool for every transaction operation which improves throughput and reduces contention on the Broker Pool.
4. New design allows concurrent  begin, commit, abort operations reducing contention on the Transaction Manager and improving concurrency.

I believe that the design is correct and I have tried to explain it in the Javadoc and with liberal commenting of the code. I recommend that the core developers carefully review my use of Atomics in `TransactionManager` to make sure I have not accidentally introduced new race conditions.

Some testing in a heavily concurrent system would be good too. @wolfgangmm perhaps you could use your concurrent JMeter tests?

### DIFF
--- a/src/org/exist/storage/txn/TransactionManager.java
+++ b/src/org/exist/storage/txn/TransactionManager.java
@@ -160,6 +160,7 @@ public class TransactionManager implements BrokerPoolService {
                     final long thisThreadId = Thread.currentThread().getId();
                     if (systemThreadId.compareAndSet(thisThreadId, thisThreadId)) {
                         // our thread is executing system tasks, allow reentrancy from our thread!
+
                         // done... return from CAS loop!
                         return doBeginTransaction();
 
@@ -196,8 +197,8 @@ public class TransactionManager implements BrokerPoolService {
         }
 
         /*
-         * NOTE: we call intentionally increment the txn counter
-         *     to set the counter to 1 to represent the TxnStart
+         * NOTE: we intentionally increment the txn counter here
+         *     to set the counter to 1 to represent the TxnStart,
          *     as that will not be done
          *     by {@link JournalManager#journal(Loggable)} or
          *     {@link Journal#writeToLog(loggable)}.
@@ -494,8 +495,8 @@ public class TransactionManager implements BrokerPoolService {
     private int uncommittedTransaction() {
         final Integer uncommittedCount = transactions.reduce(1000,
                 (txnId, txnCounter) -> {
-                    if (txnCounter.counter > 0) {
-                        LOG.warn("Found an uncommitted transaction with id " + txnId + ". Pending operations: " + txnCounter.counter);
+                    if (txnCounter.getCount() > 0) {
+                        LOG.warn("Found an uncommitted transaction with id " + txnId + ". Pending operations: " + txnCounter.getCount());
                         return 1;
                     } else {
                         return 0;
@@ -550,7 +551,15 @@ public class TransactionManager implements BrokerPoolService {
      * during shutdown.
      */
     private static final class TxnCounter {
-        private long counter = 0;
+        /**
+         * The counter variable is declared volatile as it is only ever
+         * written from one thread (via {@link #increment()} which is
+         * the `transaction` for which it is maintaining a count, whilst
+         * it is read from (potentially) a different thread
+         * (via {@link #getCount()} when {@link TransactionManager#shutdown()}
+         * calls {@link TransactionManager#uncommittedTransaction()}.
+         */
+        private volatile long counter = 0;
 
         public TxnCounter increment() {
             counter++;
--- a/src/org/exist/storage/txn/TransactionManager.java
+++ b/src/org/exist/storage/txn/TransactionManager.java
@@ -1,82 +1,131 @@
 /*
- *  eXist Open Source Native XML Database
- *  Copyright (C) 2001-04 The eXist Project
- *  http://exist-db.org
- *  
- *  This program is free software; you can redistribute it and/or
- *  modify it under the terms of the GNU Lesser General Public License
- *  as published by the Free Software Foundation; either version 2
- *  of the License, or (at your option) any later version.
- *  
- *  This program is distributed in the hope that it will be useful,
- *  but WITHOUT ANY WARRANTY; without even the implied warranty of
- *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- *  GNU Lesser General Public License for more details.
- *  
- *  You should have received a copy of the GNU Lesser General Public License
- *  along with this program; if not, write to the Free Software
- *  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
- *  
- *  $Id$
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2018 The eXist Project
+ * http://exist-db.org
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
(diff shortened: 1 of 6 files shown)


duncdrum:
@adamretter since wolfgangs tests take quite some time, you might want to address codacy `unused import`

adamretter:
@wolfgangmm just wondering if you have seen this?

adamretter:
@wolfgangmm It's been 1 month, any chance you could take a look please?

src/org/exist/storage/txn/TransactionManager.java
```diff
@@ -1,82 +1,131 @@
...
- * This is the central entry point to the transaction management service.
- * 
- * There's only one TransactionManager per database instance that can be
- * retrieved via {@link BrokerPool#getTransactionManager()}. TransactionManager
- * provides methods to create, commit and rollback a transaction.
+ * The Transaction Manager provides methods to begin, commit, and abort
+ * transactions.
+ *
+ * This implementation of the transaction manager is non-blocking lock-free.
+ * It makes use of several CAS variables to ensure thread-safe concurrent
```
dizzzz: CAS ?

src/org/exist/storage/txn/TransactionManager.java
```diff
@@ -1,82 +1,131 @@
...
+     */
+    private final AtomicInteger state = new AtomicInteger(STATE_IDLE);
+
+    /**
+     * Id of the thread which is executing system tasks when
+     * the {@link #state} == {@link #STATE_SYSTEM}. This
+     * is used for reentrancy when system tasks need to
+     * make transactional operations.
+     */
+    private final AtomicLong systemThreadId = new AtomicLong(-1);
```
dizzzz: -1 should be STATE_SYSTEM ?

src/org/exist/storage/txn/TransactionManager.java
```diff
@@ -1,82 +1,131 @@
...
+
+    /**
+     * State for when the Transaction Manager has executing system tasks.
+     */
+    private static final int STATE_SYSTEM = -1;
+
+    /**
+     * State for when the Transaction Manager is idle, i.e. no active transactions.
+     */
+    private static final int STATE_IDLE = 0;
```
dizzzz: could be initialised as `new AtomicLong(0)` so you could potentially perform a faster object identity comparison in certain conditions ? (in stead of value comparison) #notsure :-)

src/org/exist/storage/txn/TransactionManager.java
```diff
@@ -85,63 +134,164 @@ public TransactionManager(final BrokerPool pool, final Optional<JournalManager>
...
     }
     
     /**
      * Commit a transaction.
      * 
-     * @param txn
-     * @throws TransactionException
+     * @param txn the transaction to commit.
+     *
+     * @throws TransactionException if the transaction could not be committed.
```
dizzzz: should `@thows NullPointerException` be added as well?

src/org/exist/storage/txn/TransactionManager.java
adamretter: Compare and Swap - e.g. the atomics in Java.

src/org/exist/storage/txn/TransactionManager.java
adamretter: @dizzzz I am not sure I follow your reasoning here. Do you want a call to discuss?

src/org/exist/storage/txn/TransactionManager.java
adamretter: No. This is the ID of the thread which entered the system state, not the state indicator. -1 just means undefined, as it is an invalid thread id.

src/org/exist/storage/txn/TransactionManager.java
adamretter: I don't think that is necessary. Almost every Java method is capable of throwing an NPE.

wolfgangmm:
Using rather simple jmeter tests, I'm able to trigger deadlocks. See stack trace below. I'm trying to see if the issue is reproducible and send the jmeter test if it is.

[jstack.txt](https://github.com/eXist-db/exist/files/2732724/jstack.txt)

wolfgangmm:
The following jmeter test reproducibly triggers the issue for me. Because you need at least one longer running request to encounter the issue, the tests were created against one document in the current version of TEI Publisher (install from public repo). One test generates an epub and will thus lock the document for a while. It completes, but the other waiting threads will never wake up.

Testing against develop-4.x.x all threads complete reliably.

[status.jmx.txt](https://github.com/eXist-db/exist/files/2732939/status.jmx.txt)

adamretter:
@wolfgangmm It's not a deadlock, there are no locks in the new Transaction Manager!

It's a broker starvation issue. I think you have too many incoming request threads for the amount of brokers that are available.

The stack trace shows many threads waiting for a system task to complete, they do that by spinning around here:

```
java.lang.Thread.State: TIMED_WAITING (sleeping)
	at java.lang.Thread.sleep(Native Method)
	at org.exist.storage.txn.TransactionManager.beginTransaction(TransactionManager.java:170)
	at org.exist.xmldb.AbstractLocal.withDb(AbstractLocal.java:194)
```

Then there is a single thread, which is executing a system task:

```
"db.exist.scheduler.quartz-worker-1" #25 prio=5 os_prio=31 tid=0x00007fe9c63b3000 nid=0x11603 in Object.wait() [0x000070000b914000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(Native Method)
	at java.lang.Object.wait(Object.java:502)
	at org.exist.storage.BrokerPool.get(BrokerPool.java:1204)
	- locked <0x0000000740ca6798> (a org.exist.storage.BrokerPool)
	at org.exist.storage.SystemTaskManager.processTasks(SystemTaskManager.java:64)
	- locked <0x0000000740db43a8> (a java.util.ArrayDeque)
	at org.exist.storage.txn.TransactionManager.processSystemTasks(TransactionManager.java:536)
	at org.exist.storage.txn.TransactionManager.triggerSystemTask(TransactionManager.java:521)
```

The system task cannot proceed because there are no brokers available for it to use.

Before this PR each of the waiting threads would (as they still do in this PR) take a broker in `AbstractLocal.withDb` and then block when they enter `TransactionManager#beginTransaction` waiting on an exclusive lock in the transaction manager. The difference is that before this PR, a DBBroker would be taken in `TransactionManager#triggerSystemTasks` before the exclusive lock was taken in the transaction manager.

There are two solutions to solving the problem in the PR:

1. `TransactionManager#processSystemTasks` must take a broker before switching into the SYSTEM state, so that the system task cannot be starved of brokers. This is the simplest but a bit yucky, as the transaction manager shouldn't have to be concerned with this. A small change to the API where the broker is passed to `SystemTaskManager#processTasks` could make this slightly less yucky through making the broker contract explicit.

2. We reserve a broker in the broker pool which is only used for system tasks, and use that broker. This requires a bit more work but is probably a cleaner approach.

Right now, I am short on time and so could implement (1).

wolfgangmm:
Ok, so this explains why I need 20 threads to bring the system to a halt and why the same works reliably on develop-4.x.x. I'll investigate and see if I could work on option 2).

adamretter:
@wolfgangmm I think it's better just to do (1) now and focus on other things. I will send a commit for (1) I have it tested and working here.

adamretter:
@wolfgangmm okay it is pushed. Could you test again please?

If you're then happy with it, I will cherry-pick the additional commit to the other branches...

wolfgangmm:
I tested again and now everything runs fine, thanks!

dizzzz:
buildfailure can be ignored.

PR #2310, opened 2018-11-24
(4.x.x) Copying an XML Resource must remove the nodes of any XML resource it overwrites!

adamretter:
When copying an XML Document, if that document was replacing an existing document in another Collection, the nodes of the Document that it replaced were never removed from the Indexes or DOM Store.

This means that the DOM Store and Indexes will continue to grow on each copy operation, and that performance will deteriorate as more inaccessible nodes have to be searched and skipped over on each operation.


adamretter:
@wolfgangmm after you have reviewed and you think this is a sane approach, let me know, and then I will create back/forward ports for the other branches.

src/org/exist/storage/NativeBroker.java
```diff
@@ -2532,44 +2483,117 @@ public void copyResource(final Txn transaction, final DocumentImpl doc, final Co
...
+    private void doCopyDocument(final Txn transaction, final DocumentTrigger trigger,
+                                final DocumentImpl srcDoc, final Collection dstCol, final XmldbURI dstDocName)
+            throws TriggerException, LockException, PermissionDeniedException, IOException, ReadOnlyException, EXistException {
+
+        // URI of the destination document in the database
+        final XmldbURI dstDocUri = dstCol.getURI().append(dstDocName);
+
+        trigger.beforeCopyDocument(this, transaction, srcDoc, dstDocUri);
+
+        final DocumentImpl currentDstDoc = dstCol.getDocumentWithLock(this, dstDocName, LockMode.READ_LOCK);
```
shabanovd: can it simply call `delete` operation for destination document? Mean, before start copy operation.

adamretter: That would be more expensive.

shabanovd: it'll be the proper way! you should not only index inform about document remove but also triggers.

adamretter: @shabanovd I am not sure I follow you. Can you be clearer please? I have not changed how the triggers work here.

shabanovd: and that is the problem. When overwriting resource it must be deleted by "remove" procedure to inform index and fire trigger events.

adamretter: @shabanovd My goal in this PR, was not to change on which events triggers were fired for from how they were before this PR. If you would like to change that, could you submit another PR once this one is merged please?

src/org/exist/storage/NativeBroker.java
```diff
@@ -2847,6 +2851,39 @@ private void dropIndex(final Txn transaction, final DocumentImpl document) throw
...
+                    @Override
+                    public Object start() {
+                        final NodeHandle node = (NodeHandle) document.getFirstChild();
+                        domDb.removeAll(transaction, node.getInternalAddress());
+                        return null;
+                    }
+                }.run();
+            }
+        } catch(NullPointerException npe0) {
+            LOG.error("Caught NPE in DOMTransaction to actually be able to remove the document.");
```
dizzzz: as there is no recovery needed, better to log as WARNing ?

adamretter: This code was just moved. TBH I have no idea why we try and catch an NPE there, I assumed there was some reason that was forgotten in the past.

adamretter: @shabanovd Ah okay. Thanks :-)

adamretter: @dizzzz seems like a legit error to me.

adamretter:
@wolfgangmm It's been 1 month, any chance you could take a look please?

Issue #2317, opened 2018-11-30
array:get IndexOutOfBoundsException


median1st:
While accessing an out-of-bound index of array in XQuery using the parentheses notation correctly throws XQuery error, using array:get() function for the same purpose ends with an IndexOutOfBoundsException thrown within the servlet.

**Steps to reproduce:**

1.  Open new XQuery 3.1 window
2. evaluate the following expression:
```
  let $a := [ 1, 2, 3 ]
  return $a(10)
```
This correctly ends with **err:FOAY0001 Position 10 does not exist in this array.**  being reported.
3.   Change the second line of the expression to
 ` return array:get($a, 10)`
4. Evaluate the expression again. This time a successful query returing 0 items is repoted, while the log of the database contains an exception dump with this cause:
```
2018-11-30 13:08:35,603 [qtp1502843539-99] ERROR (XQueryServlet.java [process]:552) - null 
java.lang.IndexOutOfBoundsException: null
        at com.github.krukow.clj_lang.PersistentVector.arrayFor(PersistentVector.java:126) ~[clj-ds-0.0.4.jar:?]
        at com.github.krukow.clj_lang.PersistentVector.nth(PersistentVector.java:130) ~[clj-ds-0.0.4.jar:?]
        at org.exist.xquery.functions.array.ArrayType.get(ArrayType.java:67) ~[exist.jar:4.4.0]
        at org.exist.xquery.functions.array.ArrayFunction.eval(ArrayFunction.java:252) ~[exist.jar:4.4.0]
        at org.exist.xquery.BasicFunction.eval(BasicFunction.java:74) ~[exist.jar:4.4.0]
        at org.exist.xquery.InternalFunctionCall.eval(InternalFunctionCall.java:41) ~[exist.jar:4.4.0]
        at org.exist.xquery.DebuggableExpression.eval(DebuggableExpression.java:58) ~[exist.jar:4.4.0]
        at org.exist.xquery.LetExpr.eval(LetExpr.java:111) ~[exist.jar:4.4.0]
        at org.exist.xquery.LetExpr.eval(LetExpr.java:111) ~[exist.jar:4.4.0]
        at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71) ~[exist.jar:4.4.0]
        at org.exist.xquery.PathExpr.eval(PathExpr.java:276) ~[exist.jar:4.4.0]
[...]

PR #2328, opened 2018-12-04
Fix an ArrayOutOfBoundException with array:get

adamretter:
Closes https://github.com/eXist-db/exist/issues/2317

### DIFF
--- a/test/src/xquery/arrays/arrays.xql
+++ b/test/src/xquery/arrays/arrays.xql
@@ -136,10 +136,26 @@ function arr:function-item-invalid() {
 
 declare
     %test:assertError("FOAY0001")
-function arr:get-item-out-of-bounds() {
+function arr:get-item-out-of-bounds-positive() {
     let $a := [13, 10, 14]
     return
         array:get($a, 22)
+};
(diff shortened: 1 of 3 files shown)


test/src/xquery/arrays/arrays.xql
```diff
@@ -134,6 +134,14 @@ function arr:function-item-invalid() {
         $a("x")
 };
 
+declare
```
dizzzz: what about testing [0] [-1] [1] and [3] ? these are cornercases...

adamretter: I have now incorporated your further tests.

dizzzz: but the logic is still not throwing the correct errors, right?

adamretter: I added further tests. The logic was correct, but the error code was wrong, I have corrected that now.

adamretter:
@dizzzz the error was thrown, it just had the wrong code! I have fixed this and rebased, hopefully good to go now.

PR #2330, opened 2018-12-04
(5.0.0) Fix an ArrayOutOfBoundException with array:get

adamretter:
Backport of https://github.com/eXist-db/exist/pull/2328

### DIFF
--- a/test/src/xquery/arrays/arrays.xql
+++ b/test/src/xquery/arrays/arrays.xql
@@ -136,10 +136,26 @@ function arr:function-item-invalid() {
 
 declare
     %test:assertError("FOAY0001")
-function arr:get-item-out-of-bounds() {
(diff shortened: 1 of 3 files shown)


test/src/xquery/arrays/arrays.xql
```diff
@@ -136,12 +136,28 @@ function arr:function-item-invalid() {
 
 declare
     %test:assertError("FOAY0001")
-function arr:get-item-out-of-bounds() {
+function arr:get-item-out-of-bounds-positive() {
     let $a := [13, 10, 14]
     return
         array:get($a, 22)
```
dizzzz: I'd prefer to test cornercase with value 4 (that is maybe related to your int -1 change) in stead of 22 of which that should fail anyway. for "4" we might find an/the implementation issue.

Issue #2282, opened 2018-11-08
Unexpected output in Java GUI client


djbpitt:
### What is the problem

The output of `ls` in the command-line window at the bottom of the Java GUI client does not include the owner, group, or permissions, and does not match the output that I see when I launch the client in non-GUI mode from a command line. What I see in the command-line in the bottom of the GUI client is:

```
exist:/db/apps/shakespeare> ls
data	templates	modules	resources	view-work.html	Fri Oct 26 01:18:06 EEST 2018	
expath-pkg.xml	Fri Oct 26 01:18:06 EEST 2018	collection.xconf	Fri Oct 26 01:18:06 EEST 2018	icon.png	Fri Oct 26 01:18:06 EEST 2018	index.html	Fri Oct 26 01:18:06 EEST 2018	error-page.html	Fri Oct 26 01:18:06 EEST 2018	
build.xml	Fri Oct 26 01:18:06 EEST 2018	demo-queries.html	Fri Oct 26 01:18:06 EEST 2018	controller.xql	Fri Oct 26 01:18:06 EEST 2018	view-play.html	Fri Oct 26 01:18:06 EEST 2018	pre-install.xql	Fri Oct 26 01:18:06 EEST 2018	
ajax.html	Fri Oct 26 01:18:06 EEST 2018	search.html	Fri Oct 26 01:18:06 EEST 2018	search-help.html	Fri Oct 26 01:18:06 EEST 2018	repo.xml	Fri Oct 26 01:18:12 EEST 2018	
```

Note that owner, group, and permissions are missing.

### What did you expect

I expect output like the following, which was produced at the command line when I launched the client in non-GUI mode:

```
exist:/db/apps/shakespeare>ls
drwxr-xr-x admin     dba       data
drwxr-xr-x admin     dba       templates
drwxr-xr-x admin     dba       modules
drwxr-xr-x admin     dba       resources
-rw-r--r--	admin	dba	view-work.html	Fri Oct 26 01:18:06 EEST 2018
-rw-r--r--	admin	dba	expath-pkg.xml	Fri Oct 26 01:18:06 EEST 2018
-rw-r--r--	admin	dba	collection.xconf	Fri Oct 26 01:18:06 EEST 2018
-rw-r--r--	admin	dba	icon.png	Fri Oct 26 01:18:06 EEST 2018
-rw-r--r--	admin	dba	index.html	Fri Oct 26 01:18:06 EEST 2018
[...]

PR #2336, opened 2018-12-06
(4.x.x) Improve Java Admin Client Shell

adamretter:
Fixes and improves a few things with the Java Admin Client shell.

Closes https://github.com/eXist-db/exist/issues/2282
Closes https://github.com/eXist-db/docker-existdb/issues/45

@duncdrum this should fix your Docker issue with the Java Admin Client :-)


adamretter:
@duncdrum I think it should "just work" now in Docker. Could you test? If not, you might like to then test after removing the `jansi` jar from lib/core which is only required for native terminal support on Windows.

dizzzz:
should the webstart jar list be extended?

PR #2240, opened 2018-10-25
unattended installation (closes eXist-db/exist#2239)

mathias-goebel:
by using variables present in the env the script supports unattended installation

PR #2341, opened 2018-12-10
unattended installation with YAJSW

mathias-goebel:
see #2240


mathias-goebel:
ahh, no, in this way it doesnt make sense

PR #2348, opened 2018-12-13
Move from Ubuntu Trusty to Xenial on Travis-CI

adamretter:
I am not certain why, but after Travis-CI's move from Ubuntu Trusty on their container architecture to Ubuntu Trusty on their GCE architecture, the build fails. Using Travis-CI's Ubuntu Xenial on GCE seems to work fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: java
 
+dist: xenial
+
 matrix:
   include:
     - jdk: openjdk8
-    - jdk: oraclejdk8
     - jdk: openjdk9
     - jdk: openjdk10
     - jdk: openjdk11


.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: java
 
+dist: xenial
+
 matrix:
   include:
     - jdk: openjdk8
-    - jdk: oraclejdk8
```
dizzzz: :'-(

duncdrum: actually its necessary, on to the breach 11 here we come

adamretter: @dizzzz Xenial doesn't have Oracle JDK8 available, we could write some script to install it. As OracleJDK 8 and OpenJDK 8 have been very stable and are basically "done", I don't see them diverging at any point in the future. I suspect if our tests pass on OpenJDK 8, we will have no problem with Oracle JDK8

dizzzz: @adamretter I think it is a fair assumption. Let's not invest time in the oracle jvm here.

PR #2351, opened 2018-12-16
Remove legacy backup conversion code

adamretter:
This code converts an old style eXist-db user database to a new style one when restoring a backup. This stuff is from pre-2.2 days. I really don't think we need this anymore. Not to mention that it does not actually work correctly if the backup and the database have different admin passwords.


duncdrum:
we should note this change for prosperity in `upgrading.xml` in the docs. So users know that they should long jump 1.x-4.x and then move on to 5. (there is small section talking about upgrading to 1.4)

adamretter:
@duncdrum docs as requested - https://github.com/eXist-db/documentation/pull/294

PR #2362, opened 2019-01-03
(4.x.x) Improve classpath modules resolution

adamretter:
Backport of https://github.com/eXist-db/exist/pull/2360

### DIFF
--- a/src/org/exist/source/SourceFactory.java
+++ b/src/org/exist/source/SourceFactory.java
@@ -77,14 +77,9 @@ public class SourceFactory {
         Source source = null;
 
         /* resource: */
-        if (location.startsWith(ClassLoaderSource.PROTOCOL)) {
-            source = new ClassLoaderSource(location);
-        } else if (contextPath != null && contextPath.startsWith(ClassLoaderSource.PROTOCOL)) {
-            // Pretend it is a file on the local system so we can resolve it easily with URL() class.
-            final String conPathNoProtocol = contextPath.replace(ClassLoaderSource.PROTOCOL, "file://");
-            String resolvedURL = new URL(new URL(conPathNoProtocol), location).toString();
-            resolvedURL = resolvedURL.replaceFirst("file://", ClassLoaderSource.PROTOCOL);
-            source = new ClassLoaderSource(resolvedURL);
+        if (location.startsWith(ClassLoaderSource.PROTOCOL)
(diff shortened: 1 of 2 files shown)


adamretter:
@duncdrum Ah balls!
I will have to take a look into that...

src/org/exist/source/SourceFactory.java
```diff
@@ -163,6 +158,26 @@ private static String firstPathSegment(final String path) {
                 .getRawCollectionPath();
     }
 
+    private static Source getSource_fromClasspath(final String contextPath, final String location) throws IOException {
+        if (location.startsWith(ClassLoaderSource.PROTOCOL)) {
+            return new ClassLoaderSource(location);
+        }
+
+        final Path rootPath = Paths.get(contextPath.substring(ClassLoaderSource.PROTOCOL.length()));
```
dizzzz: this will resolve a *file* on literal `/path/to/file` correct? is this intended? does this mean that if I would write `resource:/etc/shadow` on a linux box that I would see all users and password hashes?

test/src/org/exist/source/SourceFactoryTest.java
```diff
@@ -134,6 +134,53 @@ public void getSourceFromResource_contextAbsoluteFileUrl_locationRelativeUrl_bas
         assertEquals(getClass().getResource(location), relativeSource.getKey());
     }
 
+    @Test
+    public void getSourceFromResource_contextFolderUrl_locationRelative() throws IOException, PermissionDeniedException {
+        final String contextPath = "resource:org/exist/source";
```
dizzzz: what if `final String contextPath = "resource:/org/exist/source";` (note the /)

adamretter:
@duncdrum Windows issue should be fixed now. Let's see what the CI says...

src/org/exist/source/SourceFactory.java
adamretter: No. `resource:` limits the resolution to the classpath. So you can't resolve things which are not already on the classpath. The `Path` class is just used to make it easier to manipulate the paths for the classpath correctly.

test/src/org/exist/source/SourceFactoryTest.java
adamretter: It won't make any difference. However in practice that doesn't happen, we have many tests for this in SourceFactoryTest.java.

adamretter:
@dizzzz I don't think there are any security concerns, this only resolves things which you have already placed on the classpath.

PR #836, opened 2015-11-18
Support for resources in repo.xml (replaces PR #727)

zwobit:
support for resources directories in repo.xml, if storing html fails try as binary in deployment.

**Cherry picked version of https://github.com/eXist-db/exist/pull/727 with current develop:**

repo.xml sample:
```
<?xml version="1.0" encoding="UTF-8"?>
<meta xmlns="http://exist-db.org/xquery/repo">
    <description>eXistdb some app</description>
    <author>eXist Solutions</author>
    <website/>
    <status>alpha</status>
    <license>BSD-2-Clause</license>
    <copyright>true</copyright>
    <type>application</type>
    <target>someapp</target>
    <resources path="bower_components" type="binary"/>
    <resources path="app/elements" type="binary"/>
</meta>
```

PR #2368, opened 2019-01-06
HTML is most likely not valid XML, unless it is XHTML

adamretter:
Updates the `mime-type.xml`.

### DIFF
--- a/mime-types.xml.tmpl
+++ b/mime-types.xml.tmpl
@@ -45,10 +45,6 @@
         <description>XHTML document</description>
         <extensions>.xhtml,.xht</extensions>
     </mime-type>
-    <mime-type name="text/html" type="xml">
-        <description>HTML document</description>
-        <extensions>.html,.htm</extensions>
-    </mime-type>
     <mime-type name="application/atom+xml" type="xml">
         <description>Atom Feed Document</description>
         <extensions>.atom</extensions>
@@ -75,6 +71,11 @@
     </mime-type>
     
     <!-- Binary mime types -->
+    <mime-type name="text/html" type="binary">
+        <!-- HTML is most likely not valid XML, unless it is XHTML -->
+        <description>HTML document</description>
+        <extensions>.html,.htm</extensions>
+    </mime-type>
     <mime-type name="application/exi" type="binary">
         <description>Efficient XML Interchange</description>
         <extensions>.exi</extensions>


duncdrum:
uh wouldn't that raise errors when storing / serving actual html5 from exist? At the moment that is possible, and should come with the mime-type `html`

dizzzz:
@adamretter i remember that we have `nekohtml` in place to be sure that html is xml-ified automagically. If that is not working.... we should fix that.

All XAR/apps, with the templating engine, expect the HTML files to be XML files, afaik. @wolfgangmm  please  step in to confirm (or not :-) )

joewiz:
@dizzzz Good point about eXist HTML template files that use the `.html` extension by convention, not `.xhtml`; HTML templating code certainly assumes that `.html` files are accessed as XML documents, so this PR certainly needs to be tested. 

I don't think neko is applied automatically when storing files. Perhaps you're thinking of how the HTTP Client tidies non-well-formed XML resources when fetching them?

adamretter:
@dizzzz Tidying of HTML to XHTML is not employed automatically when storing documents.

@joewiz The templating module might need an update then to use `.xhtml` files.

joewiz:
@adamretter Not only would the templating module need to be changed, I think, but every `.html` file in every app would need to be renamed, and every `controller.xql` that takes requests for resources with an `.html` file extension and passes the *presumably found* `.html` file through a view module would need to be adjusted. 

I do see the merits of the PR - using the `x` in `.xhtml` to signal that the contents are well-formed XML and dropping any hope of storing well-formed `.html` files, doing away with the current ambiguity. Such a change would obviate the need for https://github.com/eXist-db/exist/pull/836 (see original motivation stated explicitly in https://github.com/eXist-db/exist/pull/716 and also past wrestlings with this issue such as https://markmail.org/message/jd7r2z2nel4cslyy). 

But for a wide swath of apps hoping for a smooth upgrade to 5.0, this change of default would either require a significant update to their apps, or perhaps more directly a local reversion of `mime-types.xml`. I fear that without a well-researched, broadly coordinated, and straightforward solution, I fear that this PR could harm adoption of 5.0. 

I'm also curious to hear @wolfgangmm's take.

adamretter:
When considering this PR, I think we need to be careful to separate 3 things here:

1. Whether eXist-db stores a resource as a Binary or XML document.

    Previously eXist-db would attempt to store a file with a `.html` extension and the media-type `text/html` as an XML document. This would fail 90% of the time, as most HTML is **not** XML. When

2. The media-type.
    1. The media-type determined from the file extension,  whether eXist-db stores the resource as a binary of XML document.
    2. When the resource is stored, a copy of the media-type is also stored alongside the resource. This is not connected to the file extension and can be changed.
  3. When serving a resource (e.g. via REST) the media-type stored with the resource is used, which is not necessarily the same as that which was determined when the resource was stored.

3. The file extension.
    This just gives eXist-db a hint about the media-type when storing documents, not when serving them.

Yes, I agree that this might require some changes to the templating module. However, whichever way we cut it, the situation sucks...

Previously we tried to store `.html` as XML, as time has moved on and HTML 5 (which is non-XML) is now the predominant standard, storing HTML into eXist-db fails most of the time.

With this PR, storing HTML into eXist-db will always work, whether it is valid or invalid XML. Applications that previously (wrongly) assumed that a HTML file would be valid XML may need some improvements.

line-o:
I would appreciate getting HTML as HTML. By using the workaround my logs often look like this: 
screenshot 2019-01-10 at 18 33 14

wolfgangmm:
@line-o that's something to be fixed on the client. The gulp plugin does it wrong. The mime type should be set correctly, even when uploading as binary. The Atom plugin does it right and I don't get those messages.

line-o:
@wolfgangmm would you like to elaborate on this? Does the atom-sync change the mime-type in a subsequent request?

wolfgangmm:
@line-o it calls xmldb:store-as-binary, which sets the correct mime type on the resource:

https://github.com/eXist-db/atom-editor-support/blob/011d535ed5cda71312e9818cf4ad2d0b1f55708e/store.xql#L89

line-o:
@wolfgangmm Thanks for sharing. I will try to set the mime-type separately and see if that works, too.

line-o:
Regarding the actual topic of this PR:

* HTML is HTML and not XML, but always parseable into a DOM for querying. 
* Anything that must be valid XML should say so explicitly. 
* There is no better timing, than an upcoming breaking change, to address this.

line-o:
I understand the backwards compatibility issue @wolfgangmm.
Applying the change @adamretter proposed here in the instances I have to work with, seems to be the way to go.

adamretter:
So I disagree with @wolfgangmm on this. However, I don't see this PR making any progress. Should I close it?

wolfgangmm:
As far as I can see, @duncdrum, @joewiz and @dizzzz all raised major concerns against this change.

PR #2369, opened 2019-01-06
(5.0.0) HTML is most likely not valid XML, unless it is XHTML

adamretter:
Backport of https://github.com/eXist-db/exist/pull/2368


adamretter:
So I disagree with @wolfgangmm on this. However, I don't see this PR making any progress. Should I close it?

Issue #2233, opened 2018-10-22
Implement cdata-section-elements serialization parameter


joewiz:
### What is the problem

eXist has not implemented the `cdata-section-elements` serialization parameter.

### What did you expect

I expected eXist to support https://www.w3.org/TR/xslt-xquery-serialization-31/#XML_CDATA-SECTION-ELEMENTS.

See discussion in exist-open at https://markmail.org/message/lintubwl3wiowa3i.

### Describe how to reproduce or add a test

The following query should return the `<copy>` element with a CDATA section:

```xquery
xquery version "3.1";

declare namespace output="http://www.w3.org/2010/xslt-xquery-serialization";

declare option output:method "xml";
declare option output:cdata-section-elements "copy";

<copy><![CDATA[
<p>We&#8217;ve always known Dean Wendy Perdue was an outstanding leader for Richmond Law.</p>
]]></copy>
```

Expected result:

```xml
<copy><![CDATA[
<p>We&#8217;ve always known Dean Wendy Perdue was an outstanding leader for Richmond Law.</p>
]]></copy>
```

BaseX and Saxon correctly return the expected result.

Only eXist ignores the serialization parameter and returns the following incorrect result:

```xml
<copy>
&lt;p&gt;We&amp;#8217;ve always known Dean Wendy Perdue was an outstanding leader for Richmond Law.&lt;/p&gt;
</copy>
```

### Context information

- eXist-db version + Git Revision hash: 4.4.0 (release)
- Java version: 11
- Operating system (Windows 7, Linux, MacOs): macOS 
- 32 or 64 bit: 64 bit
- How is eXist-db installed? DMG installer
- Any custom changes in e.g. conf.xml: n/a

joewiz:
Closed by #2374.

Issue #2081, opened 2018-07-30
Incorrectly serialized brackets in HTML/JavaScript


wsalesky:
### What is the problem
Saving HTML files to the database (tested with webdav, eXide and notices on deploying a .xar)  eXist serializes brackets, even those inside CDATA sections. So `>` becomes` &gt;`

### What did you expect
This does not happen in the 3.x.x branch, the brackets inside CDATA sections are preserved. This is important for the JavaScript sections embedded in HTML.

### Describe how to reproduce or add a test
Save the below into eXide as html:

```
<div xmlns="http://www.w3.org/1999/xhtml">
    <h1>Test CDATA</h1>
    <p>Error: When this file is saved to the database (tested with webdav, eXide and notices on deploying a .xar) 
        eXist serializes brackets (> becomes &gt;), even those inside CDATA sections.</p>
    <p>eXist 4.0.0 seems to be okay, error exists in 4.2.0, 4.3.0, 4.3.1 and eXist-5.0.0-RC2</p>
    <script type="text/javascript"><![CDATA[  > Test < ]]></script>
</div>

```

### Context information
Please always add the following information
- eXist-db Tested in eXist 4.2.0, 4.3.0, 4.3.1 and eXist-5.0.0-RC2 seems to be okay in version 4.0.0
- Operating system MacOs
- How is eXist-db installed? clone from GitHub
- Any custom no custom changes

duncdrum:
Confirmed on 5.0.0 RC2 3eb07483b

joewiz:
I'm able to reproduce this result with 4.3.1 with curl - ruling out eXide as the source of the behavior, presumably, which also simply submits a PUT request to the database.

```bash
curl 'http://localhost:8080/exist/apps/eXide/store/db/temp/test.html' -X PUT -H 'Content-Type: application/xml' --data-binary $'<div xmlns="http://www.w3.org/1999/xhtml">\n    <h1>Test CDATA</h1>\n    <p>Error: When this file is saved to the database (tested with webdav, eXide and notices on deploying a .xar) \n        eXist serializes brackets (> becomes &gt;), even those inside CDATA sections.</p>\n    <p>eXist 4.0.0 seems to be okay, error exists in 4.2.0, 4.3.0, 4.3.1 and eXist-5.0.0-RC2</p>\n    <script type="text/javascript"><![CDATA[  > Test < ]]></script>\n</div>'
```

[...]

PR #2374, opened 2019-01-09
(4.x.x) Fix CDATA handling and implement output:cdata-section-elements

adamretter:
Closes https://github.com/eXist-db/exist/issues/2233
Closes https://github.com/eXist-db/exist/issues/2081

Thanks to the University of Richmond for sponsoring,

### DIFF
--- a/src/org/exist/dom/memtree/InMemoryXMLStreamReader.java
+++ b/src/org/exist/dom/memtree/InMemoryXMLStreamReader.java
@@ -24,12 +24,15 @@ package org.exist.dom.memtree;
 import org.exist.dom.QName;
 import org.exist.numbering.NodeId;
 import org.exist.stax.ExtendedXMLStreamReader;
+import org.exist.util.hashtable.Object2IntHashMap;
 import org.w3c.dom.Node;
 
+import javax.annotation.Nullable;
 import javax.xml.namespace.NamespaceContext;
 import javax.xml.stream.Location;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamReader;
+import java.util.*;
 
 
 /**
@@ -41,9 +44,12 @@ public class InMemoryXMLStreamReader implements ExtendedXMLStreamReader {
 
     private final DocumentImpl doc;
     private final NodeImpl rootNode;
+    private final InScopeNamespaces inScopeNamespaces = new InScopeNamespaces();
     private int currentNode;
 
     private int state = XMLStreamReader.START_DOCUMENT;
+
+
 
     public InMemoryXMLStreamReader(final DocumentImpl doc, final NodeImpl node) {
         this.doc = doc;
@@ -74,6 +80,9 @@ public class InMemoryXMLStreamReader implements ExtendedXMLStreamReader {
 
                 if(next < 0) { // no child nodes
                     state = XMLStreamReader.END_ELEMENT;
+
+                    inScopeNamespaces.pop(getQName());
+
                     return state;
                 }
             }
@@ -89,12 +98,23 @@ public class InMemoryXMLStreamReader implements ExtendedXMLStreamReader {
                         state = XMLStreamReader.END_ELEMENT;
                     }
                     currentNode = next;
+
+                    if (state == XMLStreamReader.END_ELEMENT) {
+                        inScopeNamespaces.pop(getQName());
+                    }
+
                     return state;
                 }
             }
             currentNode = next;
         } else {
             currentNode = rootNode.getNodeNumber();
+        }
+
+        if (state == XMLStreamReader.START_ELEMENT) {
+            inScopeNamespaces.push(getQName());
+        }  else if (state == XMLStreamReader.END_ELEMENT) {
+            inScopeNamespaces.pop(getQName());
         }
 
         switch(doc.nodeKind[currentNode]) {
@@ -173,7 +193,7 @@ public class InMemoryXMLStreamReader implements ExtendedXMLStreamReader {
 
     @Override
     public String getNamespaceURI(final String prefix) {
-        return null;
+        return getNamespaceContext().getNamespaceURI(prefix);
     }
 
     @Override
@@ -362,7 +382,7 @@ public class InMemoryXMLStreamReader implements ExtendedXMLStreamReader {
 
     @Override
     public NamespaceContext getNamespaceContext() {
-        throw new UnsupportedOperationException();
+        return inScopeNamespaces;
     }
 
     @Override
@@ -477,5 +497,96 @@ public class InMemoryXMLStreamReader implements ExtendedXMLStreamReader {
     @Override
     public String getPIData() {
         return new String(doc.characters, doc.alpha[currentNode], doc.alphaLen[currentNode]);
+    }
+
+    /**
+     * Allows access to namespaces that are in-scope
+     * by using a reference counting mechanism as
+     * opposed to a stack.
+     */
+    private static class InScopeNamespaces implements NamespaceContext {
+        private final Object2IntHashMap<QName> namespaces = new Object2IntHashMap<>(4);
+
+        // uri -> prefix(s)
+        private final Map<String, Set<String>> byUri = new HashMap<>();
+
+        // prefix -> url
+        private final Map<String, String> byPrefix = new HashMap<>();
+
+        /**
+         * Push an in-scope namespace.
+         *
+         * The namespace.
+         */
+        public void push(@Nullable final QName qname) {
+            if (qname == null) {
+                return;
+            }
+
+            final int count;
+            if (namespaces.containsKey(qname)) {
+                count = namespaces.get(qname) + 1;
+            } else {
+                count = 1;
+            }
+            namespaces.put(qname, count);
+
+            byUri.compute(qname.getNamespaceURI(), (k,v) -> {
+                if (v == null) {
+                    v = new HashSet<>();
+                }
+                v.add(qname.getPrefix());
+                return v;
+            });
+
+            byPrefix.putIfAbsent(qname.getPrefix(), qname.getNamespaceURI());
+        }
+
+        /**
+         * Pop an in-scope namespace.
+         */
+        public void pop(@Nullable final QName qname) {
+            if (qname == null) {
+                return;
+            }
+
+            if (!namespaces.containsKey(qname)) {
+                return;
+            }
+
+            final int newCount = namespaces.get(qname) - 1;
+            if (newCount >= 1) {
+                namespaces.put(qname, newCount);
+            } else {
+                namespaces.remove(qname);
+                byUri.remove(qname);
+                byPrefix.remove(qname);
+            }
+        }
+
+        @Override
+        public String getNamespaceURI(final String prefix) {
+            return byPrefix.get(prefix);
+        }
+
+        @Override
+        public String getPrefix(final String namespaceURI) {
+            final Set<String> prefixes = byUri.get(namespaceURI);
+            if (prefixes == null) {
+                return null;
+            }
+
+            return prefixes.iterator().next();
+        }
+
+        @Override
+        public Iterator getPrefixes(final String namespaceURI) {
+            final Set<String> prefixes = byUri.get(namespaceURI);
+            if (prefixes == null) {
+                return Collections.EMPTY_SET.iterator();
+            } else {
+                return prefixes.iterator();
+            }
+        }
     }
 }
--- a/src/org/exist/collections/CollectionConfigurationManager.java
+++ b/src/org/exist/collections/CollectionConfigurationManager.java
@@ -22,6 +22,7 @@ package org.exist.collections;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.exist.EXistException;
+import org.exist.Namespaces;
 import org.exist.dom.persistent.DocumentImpl;
 import org.exist.dom.memtree.SAXAdapter;
 import org.exist.security.PermissionDeniedException;
@@ -161,6 +162,7 @@ public class CollectionConfigurationManager implements BrokerPoolService {
             try {
                 reader = parserPool.borrowXMLReader();
                 reader.setContentHandler(adapter);
+                reader.setProperty(Namespaces.SAX_LEXICAL_HANDLER, adapter);
                 reader.parse(src);
             } finally {
                 if (reader != null) {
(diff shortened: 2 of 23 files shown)


adamretter:
@joewiz So I suspect that we have two conflicting things here:

1) the XDM Serialization, whereby you have to use `cdata-section-elements` to get a result like:
`<root><![CDATA[  > Test < ]]></root>`.  I think we now have this bit correct.

2) plain XML document serialization, whereby when requesting a document (when not using XQuery) you should get back what you put in. It looks to me that we are using the same serializer and options for both. I will have to see if I can set a flag for this to preserve the CDATA.

I hadn't considered this... but yeah I guess it should be in scope :-/

joewiz:
@adamretter I'm with you on both parts, except I'm not sure #1 is fully working. If you store `/db/temp.xml` as in my test, I'm unable to get the CDATA serialized even if I set the `cdata-section-elements` parameter to the name of the element:

```xquery
xquery version "3.1";

declare option output:method "xml";
declare option output:cdata-section-elements "root";

doc("/db/temp/test.xml")
```

This still returns results without CDATA, even though I'm specifying `cdata-section-elements="root"`.

In addition, I'm unable to use the `cdata-section-elements` option with the `fn:serialize()` function:

```xquery
xquery version "3.1";

fn:serialize(
    doc("/db/temp/test.xml"),
    map {
        "method": "xml",
        "cdata-section-elements": xs:QName("root")
    }
)
```

This returns:

```xml
<root>  &gt; Test &lt; </root>
```

adamretter:
@joewiz Are you certain? I wrote tests to prove that this works on persistent and in-memory (both external and constructed) documents. The tests also cover `fn:serialize` - https://github.com/eXist-db/exist/pull/2374/files#diff-a16ab00d5175e422339ff68b756cc09aR2139

Where are you executing the XQuery from?

joewiz:
@adamretter Your question about where I was executing the query from led me to try storing the XML document via a different interface. I had been using PUT via curl to eXide's "store" endpoint, which must somehow strips CDATA in the process—though it really just performs `xmldb:store()` on the value returned by `request:get-data()`; see https://github.com/eXist-db/eXide/blob/develop/modules/store.xql#L55-L98.  

I had success with your PR when I tried storing the XML into eXist over the WebDAV interface (via oXygen) and over the XML-RPC interface (via Java admin client) (though I should note that the CDATA blocks are stripped when re-opening the file via either interface). Here are my steps:

1. Save the following file via WebDAV or XML-RPC to `/db/test/cdata.xml`:

    ```xml
    <elem1><![CDATA[<p>Hello there Bob?</p>]]></elem1>
    ```

2. Save the following query to `/db/test/test.xq` and run it via `http://localhost:8080/exist/rest/db/test/test.xq`:

    ```xquery
    xquery version "3.1";
    
    declare option output:method "xml";
    declare option output:cdata-section-elements "elem1";
    
    doc("/db/test/cdata.xml")
    ```

3. The result is correct:

    ```xml
    <elem1><![CDATA[<p>Hello there Bob?</p>]]></elem1>
    ```

The variant for `fn:serialize()` (declaring text output in the prolog to avoid double-escaping of the string returned by the query body):

    ```xquery
    xquery version "3.1";
    
    declare option output:method "text";

    fn:serialize( 
        doc("/db/test/cdata.xml"),
        <output:serialization-parameters>
            <output:method value="xml"/>
            <output:cdata-section-elements value="elem1"/>
    )
    ```

However, the map variant of this raises an unexpected error. The query:

    ```xquery
    xquery version "3.1";
    
    fn:serialize( 
        doc("/db/test/cdata.xml"),
        map {
            "method": "xml",
            "cdata-section-elements": xs:QName("elem1")
        }
    )
    ```

Saxon returns the expected result without error, but eXist returns an error:

> Argument is not a URIQualifiedName

The same error is raised if we the construct the QName via `QName((), "elem1")` or `Q{}elem1`.

The relevant portion of the trace:

```text
2019-01-11 14:34:30,702 [qtp711572640-161] ERROR (XQueryServlet.java [process]:552) - Argument is not a URIQualifiedName 
java.lang.IllegalArgumentException: Argument is not a URIQualifiedName
	at org.exist.dom.QName.fromURIQualifiedName(QName.java:177) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.util.serializer.XMLWriter.parseCdataSectionElementNames(XMLWriter.java:132) ~[exist.jar:4.6.0-SNAPSHOT]
	at com.evolvedbinary.j8fu.lazy.LazyVal.get(LazyVal.java:61) ~[j8fu-1.21.jar:1.21]
	at org.exist.util.serializer.XMLWriter.startCdataSection(XMLWriter.java:420) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.util.serializer.SAXSerializer.startCDATA(SAXSerializer.java:444) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.util.serializer.ReceiverToSAX.cdataSection(ReceiverToSAX.java:147) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.XIncludeFilter.cdataSection(XIncludeFilter.java:195) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.NativeSerializer.serializeToReceiver(NativeSerializer.java:368) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.NativeSerializer.serializeToReceiver(NativeSerializer.java:273) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.NativeSerializer.serializeToReceiver(NativeSerializer.java:130) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.NativeSerializer.serializeToReceiver(NativeSerializer.java:90) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.Serializer.toReceiver(Serializer.java:1053) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.dom.memtree.DocumentImpl.startNode(DocumentImpl.java:1342) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.dom.memtree.DocumentImpl.streamTo(DocumentImpl.java:1264) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.dom.memtree.NodeImpl.streamTo(NodeImpl.java:687) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.Serializer.serializeToReceiver(Serializer.java:1081) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.Serializer.serializeToReceiver(Serializer.java:1067) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.Serializer.itemToSAX(Serializer.java:1028) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.storage.serializers.Serializer.toSAX(Serializer.java:921) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.util.serializer.XQuerySerializer.serializeXML(XQuerySerializer.java:70) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.util.serializer.XQuerySerializer.serialize(XQuerySerializer.java:50) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.util.serializer.XQuerySerializer.serialize(XQuerySerializer.java:36) ~[exist.jar:4.6.0-SNAPSHOT]
	at org.exist.xquery.functions.fn.FunSerialize.eval(FunSerialize.java:74) ~[exist.jar:4.6.0-SNAPSHOT]
```

BaseX raises an error with the QName cast, saying that the parameter must be passed as `xs:string`, and returns the expected result if the QName cast is removed. Whether implementations should require xs:QName is unclear to me. The spec [reads](https://www.w3.org/TR/xslt-xquery-serialization-31/#XML_CDATA-SECTION-ELEMENTS):

> The `cdata-section-elements` parameter contains a list of expanded QNames.

The spec then [explains "expanded QNames"](https://www.w3.org/TR/xslt-xquery-serialization-31/#dt-expanded-qname) as follows:

> An expanded QName consists of an optional namespace URI and a local name. An expanded QName also retains its original namespace prefix (if any), **to facilitate casting the expanded QName into a string**.

(Emphasis added.) Thus, I think strings should be accepted, just as [they are](https://www.w3.org/TR/xslt-xquery-serialization-31/#serparam) for the method parameter:

> `method`: An expanded QName with a non-null namespace URI, or with a null namespace URI and a local name equal to one of `xml`, `xhtml`, `html`, `text`, `json`, or `adaptive`. If the namespace URI is non-null, the parameter specifies an implementation-defined output method.

So, I'd say your PR works, but there are indirectly related issues that we should perhaps address if only by documenting deficiencies or filing new issues:

1. CDATA blocks can be serialized over the REST interface by setting/declaring serialization parameters with the `cdata-section-elements` parameter containing the list of elements. There's no way to serialize CDATA blocks in all nodes.
2. CDATA blocks are dropped when retrieving database contents via XML-RPC or WebDAV. There's no way to serialize CDATA blocks *at all* via these interfaces.
3. When using `fn:serialize()` with map form of the 2nd parameter, the `cdata-section-elements` entry raises an error. You must use the XML form for expressing these serialization parameters.

Lastly, I haven't tested backup/restore, but I worry that CDATA blocks could be stripped during the process of writing out backups or reading them back in.

adamretter:
> @adamretter Your question about where I was executing the query from led me to try storing the XML document via a different interface. I had been using PUT via curl to eXide's "store" endpoint, which must somehow strips CDATA in the process—though it really just performs `xmldb:store()` on the value returned by `request:get-data()`; see https://github.com/eXist-db/eXide/blob/develop/modules/store.xql#L55-L98.
>

I don't tend to test with eXide as I find that there are two many layers of complexity there to really know where a problem might occur. In the past eXide has exhibited several problems with serialization which doesn't inspire confidence in me. As eXide is developed atop the REST interface, it is enough for me if things work via the REST interface, anything else is likely a bug in eXide which should be addressed separately.

I see that you tested with XML-RPC and WebDAV, we should also test directly with the REST Server. i.e. HTTP PUT/POST the document, and then HTTP GET.

> I had success with your PR when I tried storing the XML into eXist over the WebDAV interface (via oXygen) and over the XML-RPC interface (via Java admin client) (though I should note that the CDATA blocks are stripped when re-opening the file via either interface). 

Yes, that is the issue we discovered in (2) [above](https://github.com/eXist-db/exist/pull/2374#issuecomment-452589949).


> However, the map variant of this raises an unexpected error. The query:
> 
> ```
> ```xquery
> xquery version "3.1";
> 
> fn:serialize( 
>     doc("/db/test/cdata.xml"),
>     map {
>         "method": "xml",
>         "cdata-section-elements": xs:QName("elem1")
>     }
> )
> ```

> > Argument is not a URIQualifiedName

Ah okay, that is definitely a bug. I can fix that easily. Unfortunately the parsing of the XML options variant and the Map options variant, require almost the same code twice, but it is just different enough that we have to have to maintain it twice in subtly different ways.

> The same error is raised if we the construct the QName via `QName((), "elem1")` or `Q{}elem1`.

It should not be possible to use a QName here. That is likely a more general bug in the parser or the static analysis phase.

> BaseX raises an error with the QName cast, saying that the parameter must be passed as `xs:string`, and returns the expected result if the QName cast is removed.

Rightly so!

> So, I'd say your PR works, but there are indirectly related issues that we should perhaps address if only by documenting deficiencies or filing new issues:
> 
> 1. CDATA blocks can be serialized over the REST interface by setting/declaring serialization parameters with the `cdata-section-elements` parameter containing the list of elements. There's no way to serialize CDATA blocks in all nodes.

Not sure I follow... Is this a REST issue of an eXide issue?

> 2. CDATA blocks are dropped when retrieving database contents via XML-RPC or WebDAV. There's no way to serialize CDATA blocks _at all_ via these interfaces.

CDATA blocks should not be dropped here IMHO. Although there is no "specification" for what eXist-db should do when retrieving a document, I think we can assume that people want out what they put in. We should also consider REST (not eXide).

> 3. When using `fn:serialize()` with map form of the 2nd parameter, the `cdata-section-elements` entry raises an error. You must use the XML form for expressing these serialization parameters.

I will fix that.

> Lastly, I haven't tested backup/restore, but I worry that CDATA blocks could be stripped during the process of writing out backups or reading them back in.

The backup/restore will do whatever the XML:DB interface does, which is likely the same result as the XML-RPC interface.

joewiz:
Agreed on all points. And in my last comment I did not rely on eXide at all - REST meant REST.

adamretter:
@joewiz I think I have it sorted now. I will send an updated PR shortly. Just testing...

adamretter:
@joewiz Okay let's see what the CI says. I have added tests for the REST and XML:DB APIs. I also added unit tests for the WebDAV API but disabled them as there seems to be some problem with the WebDAV client. Perhaps @dizzzz knows about this?

adamretter:
@joewiz I think this should be good to go now. Let me know what you find...

joewiz:
@adamretter I think you've nailed this. Retrieving the cdata.xml file via REST (opening http://localhost:8080/exist/rest/db/test/cdata.xml), WebDAV (oXygen), or XML-RPC (Java admin client), this file's CDATA blocks are preserved. 

I'm curious, is the [new "xdm-serialization" output key](https://github.com/eXist-db/exist/pull/2374/files#diff-8306e021a9cba43b9bcca4bc964ab294) accessible from XQuery? I tried `declare option exist:serialize "xdm-serialization=yes|no";` in a variant of the test query above, to see if I could enable/disable CDATA blocks without naming the individual elements, but I didn't have any luck.

joewiz:
@adamretter I see the logic there. The downside, however, is that XQuery-based APIs (including IDEs like Pebble) will be second class citizens in their ability to preserve CDATA blocks when performing store or load actions on XML documents. It would be nice for those applications to have a method to become CDATA-aware instead of CDATA-blind. Perhaps not a significant concern...

src/org/exist/dom/memtree/InMemoryXMLStreamReader.java
```diff
@@ -74,6 +80,11 @@ public int next() throws XMLStreamException {
 
                 if(next < 0) { // no child nodes
                     state = XMLStreamReader.END_ELEMENT;
+
+                    if (state == XMLStreamReader.END_ELEMENT) {
```
dizzzz: @adamretter the same value has been added two lines up.....  so maybe a } is misplaced?

adamretter:
> The downside, however, is that XQuery-based APIs (including IDEs like Pebble) will be second class citizens in their ability to preserve CDATA blocks when performing store or load actions on XML documents. It would be nice for those applications to have a method to become CDATA-aware instead of CDATA-blind

By Jove! You are of course right... such things step outside of the spec really. Could you open a new issue and reference your comment in this so we can think about doing something in future... such as either exposing the `xdm-serialization` option or something better

adamretter:
@dizzzz I just pushed an update that removes the redundant `if` statement you identified. Thank you :-)

Issue #1348, opened 2017-03-10
"except" operator misbehaving in certain cases


olvidalo:
Say I have a `$parentCollection` with two subcollections, `$col1` and `$col2`, each containing a document with `<root>col1Root<root>` and `<root>col2Root<root>` respectively.

The query 

```
$parentCollection/root except $col2/root
```

would return only the node from $col1, whereas if I use an asterisk to select the child nodes instead of the node name, like so:

```
$parentCollection/* except $col2/*
```

it returns both nodes.

I would expect it to only return the node from $col1, like in the first query.
Or am I missing something?

See

See this XQuery test for an actual example: https://gist.github.com/olvidalo/1f0a7aa7edbd134a93d05eecc1fdaad6

Relevent spec section is here: https://www.w3.org/TR/xquery/#combining_seq

Using the current develop at commit 50c66474793c95ef3a91d3159a86019ea46f244a.

dizzzz:
Hmmm don't you mean

```xquery
$col1/root except $col2/root
```

as `$parentCollection` does not contain a document? or

```xquery
$col1//root except $col2/root
```

olvidalo:
Depends if `$parentCollection` has been assigned using `collection()` (recursive) and `xmldb:xcollection()`(not recursive), I think.

In the (quick and hypothetical...) example above, `$parentCollection` would contain the documents from both subcollections.

But either way, shouldn't these two queries return the same thing if `$parentCollection` only contains `root` elements?

duncdrum:
@olvidalo would you mind creating a PR with the test script you posted marked as pending? thank you.

duncdrum:
We could still use the PR to avoid regressions, but i m glad this is working properly now.

PR #2383, opened 2019-01-11
add tests for except operator

olvidalo:
as requested in #1348


test/src/xquery/exceptOperator.xql
```diff
@@ -0,0 +1,66 @@
...
+
+declare
+ %test:tearDown 
+ function excepttest:cleanup() { 
+ xmldb:remove($excepttest:COLLECTION_NAME)
+ }; 
+
+
+declare
+%test:assertEquals("Act 1")
```
dizzzz: I am not sure if the tested result is OK

test/src/xquery/exceptOperator.xql
```diff
@@ -0,0 +1,66 @@
...
+function excepttest:except-wildcard() {
+ ($excepttest:COLLECTION/* except $excepttest:SUBCOL2/*)//tei:head/text()
+};
+
+declare
+%test:assertEquals("Act 1")
```
dizzzz: I am not sure if the tested result is OK

olvidalo:
> I think the test is not so strong. Having a look at https://www.w3.org/TR/xpath-30/#combining_seq (with the A,B,C example) we could be able to have a better test.

This test was not meant to be an exhaustive test for the except operator. I made it solely for demonstrating an issue I had that I reported in #1348, which has since be resolved. @duncdrum thought it would be beneficial to have it in the test suite to catch possible regressions, so I made this PR. I really don't mind what you do with it :)

adamretter:
Thanks for your contribution @olvidalo 

Requires https://github.com/eXist-db/exist/pull/2140 first

duncdrum:
@dizzzz with 2140 in the green I m inclined to merge this. What's in this PR is better than nothing. Improvements are of course always welcome, and can be added later.

adamretter:
@olvidalo #2140 is now merged. Could please rebase this on the develop HEAD? We can then look to get this PR merged.

olvidalo:
done

PR #2391, opened 2019-01-15
Fix lucene term range queries

tuurma:
Forward port of #2386

A lucene query like ft:search("/db/binary/", "para:[even TO foobaar]") would result in an UnsupportedOperationException. Fixed both: queries on standard and constructed indexes, added tests.


adamretter:
Requires https://github.com/eXist-db/exist/pull/2140 first

PR #2398, opened 2019-01-18
(4.x.x) Add util:eval-and-serialize function

adamretter:
New function `util:eval-and-serialize`. This is designed for external applications which wish to use XQuery to execute XQuery and need absolute control over the serialization and paging of results.

@joewiz as discussed this could benefit several tools such as Pebble, or even eXide or atom-exist.


adamretter:
@joewiz Good spot! I have now fixed that.

PR #2408, opened 2019-01-21
[feature] dashboard is dead; long live dashboard

joewiz:
### Description:

- This PR replaces the [old dashboard](https://github.com/eXist-db/dashboard) with the [new dashboard](https://github.com/eXist-db/existdb-dashboard) for the purpose of populating the autodeploy directory during a normal and installer builds.
- It also updates the webapp default redirect of requests for "/" to "/exist/apps/existdb-dashboard" and similarly updates the URL associated with dashboard in the menubar/taskbar and tool window.

Many developers have been using and testing the new dashboard and, as discussed in the January 7, 2018 community call, recommend that the new dashboard be included in the next 5.0.0-RC. (There are no current plans to backport this to the 4.x.x series.)

### Reference:

- [January 7, 2018 community call](https://docs.google.com/document/d/1-5iWMTIdnev3eaiRwAJPm7e2OmfbDroi9e6pVhJIq6A/edit?usp=sharing)
- Numerous [completed action items relating to Dashboard](https://docs.google.com/spreadsheets/d/10uqjXLrhWxvcZ9WlJzhT6IeD8gpy6a5abfmGVLw7LwM/edit?usp=sharing)

### Type of tests:

- I tested regular and installer builds and confirmed that with these PR changes, eXist starts correctly; the new dashboard loads; all panes in the app open without error; the new dashboard is the target of the redirect when requesting "/"; and the new dashboard is the target when selecting dashboard entries from menubar/taskbar and tool window.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -30,7 +30,7 @@ keystore.validity = 100000
 # The EnsureLockAspect enforces the locking contracts by annotation
 enable.ensurelocking.aspect=false
 
-autodeploy=dashboard,shared,eXide,monex,functx,usermanager
+autodeploy=existdb-dashboard,packageservice,shared,eXide,monex,functx
 autodeploy.repo=http://demo.exist-db.org/exist/apps/public-repo
 use.autodeploy.feature=true
 
--- a/exist-versioning-release.md
+++ b/exist-versioning-release.md
@@ -236,7 +236,7 @@ Once development on a new stable version is complete, the following steps will p
     
     6. Create a XAR for the website: `$ git checkout eXist-3.1.0 && ant`.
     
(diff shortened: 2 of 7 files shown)


adamretter:
One concern, is that the URLs are changing, this applies to anyone who has given out a URL to their Dashboard website or to the public website Dashboard, e.g.

`http://www.exist-db.org/exist/apps/dashboard/index.html` is becoming `http://www.exist-db.org/exist/apps/existdb-dashboard/index.html`.

I think that's a bad thing. If the new Dashboard is final, why don't we just import the code into the old repo. Git will save everything for us so we can always go back, we can then also stick with the `dashboard` name and URI.

adamretter:
@joewiz I think for me it is a showstopper.

joewiz:
Makes sense.

PR #2424, opened 2019-01-22
Fixed #2387, BackupSystemTask is not limiting the backups as folders.

### LINKED ISSUE #2387: BackupSystemTask is not limiting the backups as folders
### What is the problem
A backup job as below is not limiting the number of backups in case when suffix is not equal with ".zip
```xml
<job type="system" class="org.exist.storage.BackupSystemTask" cron-trigger="0 0 3 ? * * *">
        <parameter name="dir" value="/home/claudius/exist-db/backup" />
        <parameter name="prefix" value="full-backup-" />
        <parameter name="collection" value="/db" />
        <parameter name="user" value="admin" />
        <parameter name="password" value="password" />
        <parameter name="zip-files-max" value="2" />
</job>
```
### What did you expect
I would expect only two backups to be maintained.

Anyone can reproduce this issue, by using the above backup job in conf.xml.

### Context information
- eXist-db 4.4.0 / ab5716180
- java version "1.8.0_191"
- ubuntu 18
- 64 bit
- How is eXist-db installed? - JAR installer

claudius108:
Fix for issue #2387.

If core team agrees, I would extract
```java
Predicate<Path> filter = path -> {
	String entryName = path.getFileName().toString();
	return entryName.startsWith(prefix) && entryName.endsWith(suffix);
};
```
and
```java
Comparator<Path> timestampComparator = new Comparator<Path>() {
	public int compare(Path path1, Path path2) {
		int result = 0;
		try {
			result = Files.getLastModifiedTime(path1).compareTo(Files.getLastModifiedTime(path2));
		} catch (IOException e) {
			LOG.error("Cannot compare files by timestamp: " + path1 + ", " + path2, e);
		}
		return result;
	}
};
```
into org.exist.util.FileUtils.

### DIFF
--- a/src/org/exist/storage/BackupSystemTask.java
+++ b/src/org/exist/storage/BackupSystemTask.java
@@ -16,9 +16,6 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
 import java.util.*;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-import java.util.function.Predicate;
 
 /**
  * BackupSystemTask creates an XML backup of the current database into a directory
@@ -140,14 +137,8 @@ public class BackupSystemTask implements SystemTask {
 
     public void purgeZipFiles() throws IOException {
         if (LOG.isDebugEnabled()) {LOG.debug("starting purgeZipFiles()");}
-        
-		Predicate<Path> filter = path -> {
-			String entryName = path.getFileName().toString();
-
-			return entryName.startsWith(prefix) && entryName.endsWith(suffix);
-		};
 		
-		List<Path> entriesPaths = FileUtils.list(directory, filter);
+		List<Path> entriesPaths = FileUtils.list(directory, FileUtils.getPrefixSuffixFilter(prefix, suffix));
 		int entriesNumber = entriesPaths.size();
 		int numberOfEntriesToBeDeleted = entriesNumber - zipFilesMax + 1;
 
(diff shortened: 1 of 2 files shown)


claudius108:
Thanks.

What do you think about moving the Predicate<Path> filter snippet of code to org.exist.util.FileUtils?

In the eXist's tests, there are 9 occurrences of quite similar filters, which can be reduced to the above one.

In the current form, no PR for documentation is needed, as I just improved the case of parameter "zip-files-max" to consider backups as collections, too.

A PR for documentation would be needed in a second stage, if we agree on a new parameter, called "max-backups" or so, as being more suggestive.

adamretter:
@claudius108 Yes, always happy to see common code abstracted.

duncdrum:
the only time that `zip-file-max` occurs in the current docs is at `documentation/src/main/xar-resources/data/configuration/listings/listing-6.xml` 

As the parameter is effectively undocumented I agree with @dizzzz  that it should be. 
@claudius108 even if this is just a refactor, could you please add a pr to the docs about this (and the `max-backup`) parameter to the article that adam posted?  one or two lines about where users can find it, and what its supposed to do should do the trick. Thanks

claudius108:
I added the function FileUtils.getPrefixSuffixFilter(), abstracting this filter to be used 9 unit tests.

adamretter:
LGTM I guess we just need docs and backports now?

claudius108:
Meanwhile I tested it with v. 4.4.0, and v. 4.5.0. It works fine. It is finally helping us in production. :)

I will make PRs for documentation and backport, respectively.

dizzzz:
why did you close this PR? it is not pulled in, right?

claudius108:
I will redo it, as the structure of source code changed a lot.

PR #2430, opened 2019-01-24
Fixes OOM by adding memory options and GC tweaks

reinhapa:
### Description:
Fixes the OutOfMemory error by specify start/maximum heap size and tweak some GC settings.


duncdrum:
@reinhapa thanks this looks great.
@eXist-db/core one question if we effectively need to run the concurrency test with `5G` mem, does that mean we should also increase recommend system specs?

reinhapa:
@duncdrum I can not say if that is the same configuration that fits all clients... I will do some more investigations together with the guys from jClarity to see what the optimal GC settings would be. The maximum amount of memory as well as the GC settings depend on the type of usage and actual document sizes...

adamretter:
@reinhapa Thank you so much for fixing this :-)

PR #2431, opened 2019-01-24
Improve restore progress window 

dizzzz:
Make sure the restore progress window scan be enlarged whilst the messages pane and the items on the right scale well (or at least better)

before:
pasted_image_24_1_19__21_58

after:
restoring_data_____and_exist_admin_client


adamretter:
So what has changed? The window is just bigger?

dizzzz:
no, the content of the window now scales with the size of the window.

PR #2439, opened 2019-01-27
(4.x.x) Re-enable restore progress bar

dizzzz:
image

fixes #2434.

Also applied some improvements as suffused by Intellij (to make code more accessible)

Best reviewed while [ignoring whitespace](https://github.com/eXist-db/exist/pull/2439/files?utf8=✓&diff=split&w=1)

PR #2441, opened 2019-01-27
(5.x.x) Re-enable restore progress bar 

dizzzz:
Port of #2439

### DIFF
--- a/src/org/exist/backup/FileSystemBackupDescriptor.java
+++ b/src/org/exist/backup/FileSystemBackupDescriptor.java
@@ -21,6 +21,8 @@
  */
 package org.exist.backup;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.exist.repo.RepoBackup;
 import org.exist.util.EXistInputSource;
 import org.exist.util.FileInputSource;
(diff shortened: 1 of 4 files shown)


src/org/exist/backup/FileSystemBackupDescriptor.java
```diff
@@ -33,9 +35,13 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Properties;
+import java.util.stream.Stream;
 
 
 public class FileSystemBackupDescriptor extends AbstractBackupDescriptor {
+
+    private final static Logger LOG = LogManager.getLogger();
```
adamretter: Don't we need a class here for the logger? e.g. `getLogger(AbstractBackupDescriptor.class)`?

src/org/exist/backup/ZipArchiveBackupDescriptor.java
```diff
@@ -42,6 +44,8 @@
 
 public class ZipArchiveBackupDescriptor extends AbstractBackupDescriptor {
 
+    private final static Logger LOG = LogManager.getLogger();
```
adamretter: Don't we need a class here for the logger? e.g. `getLogger(ZipArchiveBackupDescriptor.class)`?

dizzzz: AFAIK this is not needed (anymore). in the era of log4j 1.2.x is was indeed required for some time.

see https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/LogManager.html#getLogger()

src/org/exist/backup/FileSystemBackupDescriptor.java
dizzzz: AFAIK this is not needed (anymore). in the era of log4j 1.2.x is was indeed required for some time.

see https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/LogManager.html#getLogger()

PR #2450, opened 2019-02-01
Ci adjustments

duncdrum:
This PR adjusts our combined Travis and Appveyor Ci matrix: 
-   on Appveyor we only test java8 and java11 aka lts releases java9 and 10 are outdated
-   Travis is on ubuntu 16.04, i see little point in duplicating this on appveyor so i bumped appveyor linux builds to 18.04
-   macOS on travis now uses macOS 10.13 aka HighSierra
-   removed a duplicate ubuntu openjdk 11 build. openjdk is on appveyor, oraclejdk on travis. 

This makes test runs faster, while still broadening our system coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,10 @@ matrix:
     - jdk: openjdk8
     - jdk: openjdk9
     - jdk: openjdk10
-    - jdk: openjdk11
     - jdk: oraclejdk11
 
     - os: osx
-      osx_image: xcode9.2
+      osx_image: xcode9.3
       env: JAVA_HOME=$(/usr/libexec/java_home)
 
 script: ./build.sh -Dexist.autodeploy=off -Dtest.haltonerror=true -Dtest.haltonfailure=true clean clean-all all test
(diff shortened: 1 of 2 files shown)


duncdrum:
If there are no objections, i ll prepare the ports

duncdrum:
so i screwed up and accidentally pushed to develop-5.0.0 instead of a separate branch for the back port. While i can revert the changes i m a bit worried that it might lead to even more trouble @adamretter @wolfgangmm @dizzzz revert and prepare a branch or just leave it be and feel really stupid.

dizzzz:
@duncdrum For me a revert is not really needed, unless @adamretter can solve it quickly and safely :-)

adamretter:
Two questions - 

1) Why are we removing OpenJDK11 from Travis? OpenJDK is most important
2) What version of Ubuntu was AppVeyor previously using?

duncdrum:
@adam previously both travis and appveyor where using ubuntu 16.04 with openJdk 11, and one additional ubuntu 16.04 with oracleJdk 11 on travis. 
Given our somewhat tight instance budget for both appveyor and travis, i fail to see the value of testing the same OS JDK version twice, and adding 3h (2x30min x3 branches) of wait times to ci runs across the org for it. 
With this PR we keep openJDK11 on ubuntu 18.04 on appveyor, and oracleJDK11 on ubuntu 16.04 on travis. I think we learn more from testing against both oracle and open, instead of testing openJDK on two different ubuntu versions. 

Thanks for the rewind will open another branch then.

PR #2457, opened 2019-02-04
Improve serializer performance

adamretter:
Fixes a performance issue in the serializer.


dizzzz:
How does this improve performance?

adamretter:
@dizzzz creating at XSLT processor is an expensive operation. The serializer only needs an XSLT processor when there is an XSL Processing Instruction present in the XML; This is rarely used.

Before this change the serializer would always create a new XSLT processor

PR #2466, opened 2019-02-09
(4.x.x) Deduplicating BLOB Store

joewiz:
Rebase of https://github.com/eXist-db/exist/pull/2314

(Please test/confirm if I got this right!)


joewiz:
Clearly that hasty attempt failed to fix it. I think reverting the merged PR and reapplying the merges in the correct order is the way to go.

PR #2507, opened 2019-02-20
(4xx) Make sure mime-types.xml is in jar file

### LINKED ISSUE #2506: Some resources are not handled correctly in latest releases
Exist-db 4.6 and 5.0.0RC6

the following items need to be adressed:

- mime-types.xml  is copied to `eXist-db\src\org\exist\util\mime-types.xml` but should end in `eXist-db\exist-core\src\main\resources` (or we should git rid of the one in EXIST_HOME at all, of have it copied from the resources
- `eXist-db\exist-core\src\main\resources\org\exist\xquery\parser` should only contain the .g files as the rest should be generated. 

image

and

image

dizzzz:
Add xml file to mime type. Fixes #2506


joewiz:
Should the generated files removed in this PR be added to https://github.com/eXist-db/exist/blob/develop/.gitignore?

dizzzz:
@joewiz not yet ; it is covered by the `mime-types.xml` pattern

Issue #2533, opened 2019-03-08
Disable recovery is not possible


ohumbel:
### What is the problem
I am embedding eXist, and I am turning off recovery by setting `/exist/db-connection/recovery/@enabled` to `false` in `conf.xml`.
This leads to a `java.util.NoSuchElementException: No value present` in `BrokerPool`.

### What did you expect
I expected to be able to create a local connection.

### Describe how to reproduce or add a test
I will ad a pull request shortly, with a [BrokerPoolNoRecoveryTest](https://github.com/ohumbel/exist/blob/a0c74f30570b4668ec00fbbb7781adaaeddf07de/exist-core/src/test/java/org/exist/storage/BrokerPoolNoRecoveryTest.java) reproducing the problem.

### Context information
I tested it on branches `develop` and `develop-4.x.x`. The problem is present on both.
My Java version is `Java8u202`.
It is happening on Windows, Linux and Mac, 64 bit.
I embedded `exist-core`, `exist-optional` and `exist-start` into my application.
I set `/exist/db-connection/recovery/@enabled` to `false` in `conf.xml`.

PR #2534, opened 2019-03-08
Start fine with recovery disabled (Fixes issue #2533)

ohumbel:
### Description:
In `BrokerPool`, there are several accesses to the `Optional` `journalManager`. All except one are checking it for presence before calling `get()`. The fix adds the missing one which was causing the `java.util.NoSuchElementException: No value present` exception.

### Reference:
This fixes [Issue 2533](https://github.com/eXist-db/exist/issues/2533).

### Type of test:
`JUnit` test

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/util/LogFunction.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/util/LogFunction.java
@@ -26,7 +26,6 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.exist.dom.QName;
 import org.exist.source.StringSource;
-import org.exist.storage.serializers.Serializer;
 import org.exist.util.serializer.XQuerySerializer;
 import org.exist.xquery.*;
 import org.exist.xquery.value.*;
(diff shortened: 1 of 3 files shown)


exist-core/src/main/java/org/exist/xquery/functions/util/LogFunction.java
```diff
@@ -147,13 +146,21 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
...
+                    XQuerySerializer xqs = new XQuerySerializer(context.getBroker(), props, writer);
+                    xqs.serialize(next.toSequence());
+                    buf.append(writer.toString());
+
+                } catch (IOException | SAXException e) {
+                    throw new XPathException(this, e.getMessage());
+                }
             }
 
         }
```
wolfgangmm: Looks like this file was added to the PR by mistake due to your branch not being in sync with upstream? The change to util:log has already been merged.

ohumbel: sorry try to fix that

adamretter: As soon as that is fixed, I am happy to merge this.

PR #2543, opened 2019-03-09
Fix some issues with equality in the EnsureLockingAspect

adamretter:


exist-core/src/main/java/org/exist/storage/lock/EnsureLockingAspect.java
```diff
@@ -624,6 +624,11 @@ private boolean hasCollectionLock(final LockManager lockManager, final XmldbURI
                     throw new UnsupportedOperationException("Currently only READ or WRITE lock modes are supported");
             }
 
+            if (uri.numSegments() == 2 && uri.getCollectionPath().equals("/db")) {
```
dizzzz: should this be ported to 4.x.x ? @adamretter

PR #2551, opened 2019-03-09
Fix the YAJSW installation

adamretter:
Fixes a regression introduced by https://github.com/eXist-db/exist/pull/2240


adamretter:
@mathias-goebel could you please test this and report back please?

Issue #2446, opened 2019-01-30
NPEs for user without primary group


opax:
### What is the problem

If there is a database user without a primary group, NPEs may be thrown in at least three places:

* org.exist.security.PermissionFactory.getDefaultCollectionPermission(PermissionFactory.java:74)
* org.exist.xquery.value.StringValue.equals(StringValue.java:757)
* org.exist.util.serializer.AdaptiveWriter.escapeQuotes(AdaptiveWriter.java:274)

I suspect there are more places, these three I just happened to stumble upon.

### What did you expect

If there is a strong assumption that a user always has a primary group, it should not be possible to create a user without one. If the assumption is that the primary group is optional, the code should be changed accordingly.

### Describe how to reproduce or add a test

The first NPE happened to me when accessing http://localhost:8080/exist/ when authenticated as a user without primary group. This may be database-dependent, though.

The second NPE may be triggered by executing (as admin)

```xquery
sm:create-account('foo', 'secret', '', ())
```
to create user 'foo'. Then, in a different query, the following triggers the exception:

```xquery
sm:get-user-primary-group('foo')
```

The third NPE can be triggered by executing the following query in eXide:
```xquery
let $user := 'foo2'
return (
    sm:create-account($user, 'secret', '', ()), 
    sm:get-user-primary-group($user)
)
```
The Usermanager in the dashboard also allows to create users without primary group - just do not assign any group when creating a new user.

### Context information

Reproducible in

- eXist-db 4.5.0 (official docker image, no further changes), i.e. eXist 4.5.0 / e29b4099c / build date 201901270353
- eXist-db 5.0.0-RC5 (official docker image, no further changes), i.e. eXist 5.0.0-RC5 / 4414b1a73 / build date 201811231138

duncdrum:
this could explain something i encountered before. as for the primary design of group less users @eXist-db/core any thoughts
also  @JoernT looks like usermanager should indeed have an additional check to make sure it only lets users implement whatever the design intended

adamretter:
Primary groups are not optional, which is why the parameter cardinality is `EXACTLY_ONE`. The parameter is required, setting an empty-string is subverting the system!

opax:
Even with that patch, it is still trivially possible to trigger the NPE, just create the user with

```xquery
sm:create-account('foo', 'secret', 'bar', ())
```

[...]

PR #2564, opened 2019-03-10
Avoid NPE when empty string is given as Primary Group

adamretter:
Closes https://github.com/eXist-db/exist/issues/2446


exist-core/src/main/java/org/exist/xquery/functions/securitymanager/AccountManagementFunction.java
```diff
@@ -209,7 +206,11 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
                         subGroups = getGroups(args[2]);
                     } else {
                         //add the primary group as the primary group
-                        user.addGroup(args[2].getStringValue());
+                        final String primaryGroup = args[2].getStringValue();
+                        if (primaryGroup == null || primaryGroup.isEmpty()) {
```
dizzzz: !StringUtils.isBlank()

Issue #2437, opened 2019-01-25
dynamic declared variables in xquery module lead to NPE


line-o:
### What is the problem

It is possible to declare variables, whose values will be (lazily?) evaluated at runtime.
Using a function or arrow expression leads to an NPE.

### What did you expect

one of the three:

1. the value to be evaluated correctly
2. the xquery parser to raise an error
3. the xquery runtime to raise an error

The xquery 3.1 spec states:

> [Definition: If a variable declaration includes an expression (VarValue or VarDefaultValue), the expression is called an initializing expression. The static context for an initializing expression includes all functions, variables, and namespaces that are declared or imported anywhere in the Prolog, other than the variable being declared.]

and includes following example

```xquery
declare variable $a := local:f(); 
declare variable $b := 1;
declare function local:f() { $b }; 
```

### Describe how to reproduce or add a test

Run the following test suite:

```xquery
xquery version "3.1";

module namespace devar="http://exist-db.org/xquery/test/declared-variables";
declare namespace test="http://exist-db.org/xquery/xqsuite";

declare variable $devar:static := 'I am static.';
declare variable $devar:dynamic := replace($devar:static, 'static', 'dynamic');
declare variable $devar:function := (function ($a) {
    replace($a, 'static', 'the return of a function expression')
})($devar:static);

declare variable $devar:arrow := $devar:static 
    => replace('static', 'the return of an arrow expression')
;

declare
    %test:assertEquals('I am static.')
function devar:read-static () {
    $devar:static
};

declare
    %test:assertEquals('I am dynamic.')
function devar:read-dynamic () {
    $devar:dynamic
};

declare
    %test:assertEquals('I am the return of a function expression.')
function devar:read-function () {
    $devar:function
};

declare
    %test:assertEquals('I am the return of a arrow expression.')
function devar:read-arrow () {
    $devar:arrow
};
```

### Context information
Please always add the following information

- eXist-db version 5.0.0 / 3d8272ad5fb706cb6291cabf8d1e3cef31566453 
- Java(TM) SE Runtime Environment (build 1.8.0_152-b16)
- Operating system MacOs Mojave
- How is eXist-db installed? GitHub
- Any custom changes in e.g. conf.xml none

joewiz:
I can confirm the result described here. The test suite returns the following output:

```xml
<testsuites>
    <testsuite package="http://exist-db.org/xquery/test/declared-variables"
[...]

PR #2577, opened 2019-03-12
Dynamically loaded module vars not analyzed

### LINKED ISSUE #1550: regression with closures in maps (since 3.4.0)
### What is the problem
Starting with eXist version 3.4.0 inline functions within maps will not inherit variables declared outside its function body. The problem has been discussed on [eXist-open](https://sourceforge.net/p/exist/mailman/message/36045046/) and was confirmed by @joewiz by testing against Saxon and BaseX.

### What did you expect

I expected all three tests to pass without errors (as they do in eXist version 3.3.0).

```xquery
xquery version "3.1";

module namespace ct="http://exist-db.org/xquery/closures/test";

declare variable $ct:model := function($param) {
    let $func := function() {
       $param
    }
    return
        map {
            'test1' : $func,
            'test2' : $func(),
            'test3' : function() {
                $param
            }
        }
};

declare %test:assertEquals('Hello World') function ct:test1-function-reference() as xs:string {
    try {
        $ct:model('Hello World')?test1()
    }
    catch * {$err:code || ' ' || $err:description}
};

declare %test:assertEquals('Hello World') function ct:test2-function-call() as xs:string {
    try {
        $ct:model('Hello World')?test2
    }
    catch * {$err:code || ' ' || $err:description}
};

declare %test:assertEquals('Hello World') function ct:test3-inline-function() as xs:string {
    try {
        $ct:model('Hello World')?test3()
    }
    catch * {$err:code || ' ' || $err:description}
};
```

### Context information
- eXist-db version 3.5.0-SNAPSHOT /  fdaabad
- Java 1.8.0_131 [Oracle Corporation]
- Operating system MacOs Capitan 64 bit
- vanilla build from develop branch

wolfgangmm:
### Description:

Trying to reproduce #1550 it turned out that globally declared variables are not properly analyzed when a module is loaded dynamically, e.g. via `load-xquery-module`. A statically loaded module would be analyzed during compilation. For dynamically loaded modules we thus need to call analyze one time on all declared variables. Not doing so causes a NPE on expressions used within the variable declaration, which depend on the analysis context. #2437 demonstrates the XQuery issue itself, while #1550 reports a wrong failure in XQSuite due to the same bug.

### Reference:

Closes #1550, #2437

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ExternalModuleImpl.java
+++ b/exist-core/src/main/java/org/exist/xquery/ExternalModuleImpl.java
@@ -242,8 +242,7 @@ public class ExternalModuleImpl implements ExternalModule {
 
     public void analyzeGlobalVars() throws XPathException {
         for (final VariableDeclaration decl : mGlobalVariables.values()) {
-            decl.resetState(false);
-            decl.analyze(new AnalyzeContextInfo());
+            decl.analyzeExpression(new AnalyzeContextInfo());
         }
     }
 
--- a/exist-core/src/main/java/org/exist/xquery/functions/util/PrologFunctions.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/util/PrologFunctions.java
@@ -27,6 +27,7 @@ import org.apache.logging.log4j.Logger;
 import org.exist.dom.QName;
 import org.exist.xquery.*;
 import org.exist.xquery.Module;
+import org.exist.xquery.functions.fn.LoadXQueryModule;
 import org.exist.xquery.value.FunctionParameterSequenceType;
 import org.exist.xquery.value.Sequence;
 import org.exist.xquery.value.SequenceType;
@@ -48,7 +49,7 @@ public class PrologFunctions extends BasicFunction {
 				new FunctionParameterSequenceType("location", Type.ANY_URI, Cardinality.EXACTLY_ONE, "The location of the module")
 			},
 			new SequenceType(Type.ITEM, Cardinality.EMPTY),
-			"Prefer fn:load-xquery-module"),
+			LoadXQueryModule.LOAD_XQUERY_MODULE_2),
 		new FunctionSignature(
 			new QName("declare-namespace", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
 			"Dynamically declares a namespace/prefix mapping for the current context.",
--- a/exist-core/src/main/java/org/exist/xquery/functions/util/PrologFunctions.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/util/PrologFunctions.java
@@ -47,7 +47,8 @@ public class PrologFunctions extends BasicFunction {
 				new FunctionParameterSequenceType("prefix", Type.STRING, Cardinality.EXACTLY_ONE, "The prefix to be assigned to the namespace"),
 				new FunctionParameterSequenceType("location", Type.ANY_URI, Cardinality.EXACTLY_ONE, "The location of the module")
 			},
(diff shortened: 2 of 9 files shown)


exist-core/src/test/xquery/xquery3/closure-in-maps.xql
```diff
@@ -0,0 +1,45 @@
+(:~
+ : See https://github.com/eXist-db/exist/issues/1550
```
duncdrum:
```suggestion
 : @see https://github.com/eXist-db/exist/issues/1550
```

exist-core/src/test/xquery/xquery3/closure-in-maps.xql
```diff
@@ -0,0 +1,45 @@
...
+
+declare variable $ct:test := "123";
+
+declare variable $ct:model := function($param) {
+    let $func := function() {
+        $param
+    }
+    return
+        map {
+        'test1' : $func,
```
duncdrum:
```suggestion
        'test1': $func,
```

exist-core/src/test/xquery/xquery3/closure-in-maps.xql
```diff
@@ -0,0 +1,45 @@
...
+        'test2' : $func(),
```
duncdrum:
```suggestion
        'test2': $func(),
```

exist-core/src/test/xquery/xquery3/closure-in-maps.xql
```diff
@@ -0,0 +1,45 @@
...
+        'test3' : function() {
```
duncdrum:
```suggestion
        'test3': function() {
```

exist-core/src/main/java/org/exist/xquery/ExternalModuleImpl.java
```diff
@@ -242,8 +242,7 @@ public Variable resolveVariable(QName qname) throws XPathException {
 
     public void analyzeGlobalVars() throws XPathException {
         for (final VariableDeclaration decl : mGlobalVariables.values()) {
-            decl.resetState(false);
-            decl.analyze(new AnalyzeContextInfo());
+            decl.analyzeExpression(new AnalyzeContextInfo());
```
adamretter: Just to check - is it okay to not call `decl.ResetState(false)`?

exist-core/src/main/java/org/exist/xquery/functions/util/PrologFunctions.java
```diff
@@ -47,7 +47,8 @@
 				new FunctionParameterSequenceType("prefix", Type.STRING, Cardinality.EXACTLY_ONE, "The prefix to be assigned to the namespace"),
 				new FunctionParameterSequenceType("location", Type.ANY_URI, Cardinality.EXACTLY_ONE, "The location of the module")
 			},
-			new SequenceType(Type.ITEM, Cardinality.EMPTY)),
+			new SequenceType(Type.ITEM, Cardinality.EMPTY),
+			"Prefer fn:load-xquery-module"),
```
adamretter: It is better to link by class reference that name, e.g. change `"Prefer fn:load-xquery-module"` ->`LoadXQueryModule.class`

adamretter:
@wolfgang #2140 is now merged. Could please rebase this on the develop HEAD? We can then look to get this PR merged.

exist-core/src/main/java/org/exist/xquery/ExternalModuleImpl.java
wolfgangmm: Yes, we're calling this method right after the module was compiled, so there's no state to reset yet.

PR #2587, opened 2019-03-14
Adopt FastUtil primitive maps

adamretter:
eXist's implementations have problems with hash-collisions and chaining.


dizzzz:
nice.... 2 q's:
- what is "lib/core/fastutil-8.2.2-min.jar.md" about ... the contents is strange
- missing licence file ?

wolfgangmm:
I was happy when we finally got fastutil removed. Need some testing before feeling well about reintroducing it.

joewiz:
@dizzzz See https://github.com/eXist-db/exist/commit/649108f09570f07d4650714aa49b28d6cb826707.

adamretter:
@joewiz @wolfgangmm that was 15 years ago!!! Fastutil has for sure moved on since then and is well maintained. It also doesn't have the collision bugs I discovered in eXist-db's

adamretter:
@dizzz the md (markdown) file contains the details of how I produced a "min" jar from the official fastutils release... which is inline with the advice on the fastutils website

wolfgangmm:
Yeah, it's a long time, but still doesn't hurt to be cautious. I remember it was a monster library back then. For my sanity I just need to be confident it has changed.

adamretter:
@wolfgangmm sure the full library is 18mb. Which is why I used autojar as they suggest to just take what we need. Hence the "-min" postfix and ".md" file. Our min library is just 1.8mb

wolfgangmm:
I quickly checked memory after restoring a larger backup on 4.x.x and develop. Nothing suspicious to be seen. I did not run stress tests yet, but if this PR is needed on the way to fix 5.0.0, I would agree to merge it.

PR #2590, opened 2019-03-18
Improve the performance of the LockTable

adamretter:
Hopefully this resolves the performance issue of eXist-db 5.0.0 compared to 4.x.x.

For production, users also have the option of disabling the lock table by using the system property `-Dexist.locktable.disable=true`.


adamretter:
@dizzzz okay I fixed it :-)

PR #2600, opened 2019-03-25
Fix the Spatial Index

adamretter:
Fixes the compilation and operation of the Spatial Index.

There is still one test failure, the `update` test. However, I cannot see that this could ever have worked.

So it seems the Spatial Index is restored, but at present does not support XUpdate or XQuery Update.

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -7,23 +7,6 @@
   <component name="CompilerConfiguration">
     <option name="DEFAULT_COMPILER" value="ajc" />
     <excludeFromCompile>
-      <file url="file://$PROJECT_DIR$/extensions/indexes/spatial/src/main/java/org/exist/xquery/modules/spatial/FunGeometricProperties.java" />
-      <file url="file://$PROJECT_DIR$/extensions/indexes/spatial/src/main/java/org/exist/indexing/spatial/GMLHSQLIndexWorker.java" />
-      <file url="file://$PROJECT_DIR$/extensions/indexes/spatial/src/main/java/org/exist/indexing/spatial/AbstractGMLJDBCIndexWorker.java" />
-      <file url="file://$PROJECT_DIR$/tools/izpack/src/org/exist/izpack/CommandLineInstaller.java" />
-      <file url="file://$PROJECT_DIR$/extensions/indexes/spatial/src/main/java/org/exist/xquery/modules/spatial/FunGMLProducers.java" />
-      <file url="file://$PROJECT_DIR$/extensions/modules/src/main/java/org/exist/xquery/modules/svn/SVNLog.java" />
-      <file url="file://$PROJECT_DIR$/extensions/modules/src/main/java/org/exist/xquery/modules/xslfo/RenderFunction.java" />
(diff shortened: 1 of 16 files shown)


brihaye:
Err... in  exist-core/src/main/java/org/exist/storage/structural/NativeStructuralIndexWorker.java, we have :

index.btree == null

The spatial index is database driven. What btree are we dealing with ?

brihaye:
Also... all tests did work in 2008 :-) The spatial index has been thoroughly tested...

duncdrum:
@brihaye @adamretter there is a lengthy section in the docs about the spatial index here: (devguide_indexes.xml)[https://github.com/eXist-db/documentation/blob/53a12e322df49173ae816e516935f2d46ecc4b1a/src/main/xar-resources/data/devguide_indexes/devguide_indexes.xml]

Before we merge could you please take a look and make the necessary updates

brihaye:
> @brihaye @adamretter there is a lengthy section in the docs about the spatial index here: [devguide_indexes.xml](https://github.com/eXist-db/documentation/blob/53a12e322df49173ae816e516935f2d46ecc4b1a/src/main/xar-resources/data/devguide_indexes/devguide_indexes.xml)

> Before we merge could you please take a look and make the necessary updates

This text corresponds to https://exist-db.org/exist/apps/doc/devguide_indexes.xml?field=all&id=D3.19.31.2

It dates from the period where I and Wolfgang decided to implement indexes on eXist. He worked on the N-Gram index, I worked on the spatial one... Is this stuff still valuable ?

I could write something about use of the spatial index though...

duncdrum:
TBH there is a bit of overlap in this article (hence its length), in my mind we still need:
-   an article explaining the existence, usage, and basics of the spatial index
-   dito for N-gram
-   an article explaining how to write indexes for exist

I m open for rethinking if this should happen in three separate articles, in different sections, or … some refactoring seems in order. 

As far as this PR is concerned, the code base should match the examples in the docs. I ll open a separate issue in the docs for how to declutter `devguide_indexes.xml`. @brihaye  Your contributions would be very much welcome.

adamretter:
@brihaye The change to `NativeStructuralIndexWorker.java` is not directly related to the Spatial Index.

@brihaye I am not sure why Update no longer works. But it looks to me like it tries to remove the document node, which is odd as eXist-db has never stored document nodes in the DOM.

exist-core/src/main/java/org/exist/plugin/PluginsManagerImpl.java
```diff
@@ -94,7 +94,6 @@ public void prepare(final BrokerPool brokerPool) {
 		//Temporary for testing
 		addPlugin("org.exist.scheduler.SchedulerManager");
 		addPlugin("org.exist.storage.md.MDStorageManager");
-		addPlugin("org.exist.monitoring.MonitoringManager");
```
dizzzz: should this be in this PR?

adamretter:
@duncdrum I updated the docs as requested - https://github.com/eXist-db/documentation/pull/319

duncdrum:
thx @adamretter i d suggest a rebase to make sure the linux tests on appveyor have a chance to run properly.

adamretter:
@dizzzz Okay buddy, I have now rebased it.

PR #2614, opened 2019-03-28
(4.x.x) Fast faceted search

wolfgangmm:
**Update**: the PR is now considered feature complete. Documentation PR: https://github.com/eXist-db/documentation/pull/325

This adds fast faceted search to the lucene-based full text index. A full text index defined in `collection.xconf` can now specify additional facets using an arbitrary xpath rooted in the indexed element. The facets are created automatically at indexing time.

```xml
<lucene>
    <text qname="tei:body">
        <facet dimension="from" expression="ancestor::tei:TEI//tei:sourceDesc/vg:letDesc/vg:letHeading/tei:author"/>
        <facet dimension="to" expression="ancestor::tei:TEI//tei:sourceDesc/vg:letDesc/vg:letHeading/vg:addressee"/>
    </text>
</lucene>
```

All result sequences obtained from a query with `ft:query` will have facet counts attached, which can be retrieved for display to the user. The user may then select a particular facet to drill down into the result. Drill down is super fast on large result sets as no additional computation needs to be done at runtime. See the test cases for usage examples.

With the new features, facets can be integrated into a search results view as in below example:

image


adamretter:
@wolfgangmm Very cool!

Do you plan to also extend faceting to the Range Index?

joewiz:
Very exciting! Here's a screenshot of my app using this branch:

Screen Shot 2019-03-27 at 5 32 49 PM

extensions/indexes/lucene/src/test/xquery/lucene/facets.xql
```diff
@@ -0,0 +1,310 @@
...
+            map { "method": "json" }
+        )
+};
+
+declare
+    %test:args("place:hamburg")
+    %test:assertEquals(1)
+    %test:args("place:berlin")
+    %test:assertEquals(2)
+    %test:args("from:rudi and place:berlin")
```
joewiz: For this and the next few argument sets, the boolean `AND` should be expressed in all caps. Otherwise, it is likely dropped as a stop word and thus fails to perform the expected function. See https://lucene.apache.org/core/2_9_4/queryparsersyntax.html#Boolean%20operators.

wolfgangmm:
As discussed in the community call, this PR should maybe go into 5.0 and develop only, not into 4.x.x. Though it does not introduce breaking changes and preserves full compatibility, it's a major feature, which seems more appropriate for a major release. It may also be an incentive for users to update to 5.

Issue #2592, opened 2019-03-18
Re-implement VirtualTempFile for remote resources


reinhapa:
### The content for each remote Resource is stored in real file on the client side
When switching from the version 1.4.2 to the current version 4.6.x version we found a relative big performance drop when reading a lot of small XML resource via remote XMLDB connection from exist compared to the old version.

We found that the old VirtualTempFile storing the remote resource data up to a certain amount (256 KB) in memory instead of saving it to the disk storage.

### What did you expect
I would like to restore the [VirtualTempFile](https://github.com/eXist-db/exist/blob/eXist-1.4.x/src/org/exist/util/VirtualTempFile.java) functionality back in the actual versions of eXist database.

dizzzz:
please could you pinpoint to the interfaces/code where this change should be implemented?
@adamretter convinced me that  modern OS should provide a transparent buffering for this...

reinhapa:
The change would replace the `Path contentFile` within the `AbstractRemoteResource` with a specialized implementation buffering a certain amount of data in the HEAP before storing to the disk.

reinhapa:
After testing an initial implementation of a `VirtualTempPath` the impact exporting 56701 documents less than 8 KB is as follows before:
```
[2019-03-25 10:31:42,616] INFO  [EmbeddedConnection] found 56701 documents
[2019-03-25 10:31:43,539] INFO  [TemporaryFileManager] Temporary folder is: C:\Users\PATRIC~1.REI\AppData\Local\Temp\exist-db-temp-file-manager-2459278453317762022
[2019-03-25 10:53:57,947] INFO  [EmbeddedConnection] export collection dbXworkspaceXrepositoryXStandardXfield done
[2019-03-25 10:53:57,947] INFO  [EmbeddedConnection] export complete in 1336073ms
```
Using the re-implemented VirtualTempPath:
```
[2019-03-25 10:54:39,648] INFO  [EmbeddedConnection] export collection dbXworkspaceXrepositoryXStandardXfield
[2019-03-25 10:54:39,873] INFO  [EmbeddedConnection] found 56701 documents
[2019-03-25 10:54:40,682] INFO  [TemporaryFileManager] Temporary folder is: C:\Users\PATRIC~1.REI\AppData\Local\Temp\exist-db-temp-file-manager-3740725248214970142
[2019-03-25 11:07:37,898] INFO  [EmbeddedConnection] export collection dbXworkspaceXrepositoryXStandardXfield done
[...]

PR #2621, opened 2019-03-29
(4.x.x) Implements VirtualTempPath as described in #2592

reinhapa:
### Description:
Adds in-memory buffer for all remote `XMLResource` including JUnit tests for the newly added functionality.


exist-core/src/main/java/org/exist/xmldb/AbstractRemoteResource.java
```diff
@@ -34,30 +34,29 @@
...
 import org.xmldb.api.base.XMLDBException;
 
-import javax.annotation.Nullable;
-
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.exist.util.io.InputStreamUtil.copy;
 
 public abstract class AbstractRemoteResource extends AbstractRemote
         implements EXistResource, ExtendedResource, Resource {
+    private static final int DEFAULT_IN_MEMOTY_SIZE = 0x400_0000; // 4 MB
```
dizzzz: typo: MEMO R Y_SIZE

reinhapa:
Fixed so far, the failing "quality" gates are a bit blind sided in my opinion.

- the static factory method `static AutoReleaseLock autoRelease(Lock lock)` of the `AutoReleaseLock` class is intentionally package private and could not be made private nor do I wanted it to be public.
- the test methods `public void close()` and `public void flush()` of `OverflowToDiskStreamTest` do not have either a `fail` nor `assertXXX` methods due the fact as they are tested by a recorded EasyMock expectation

dizzzz:
the mac build seems to be too slow... there are always timeouts

exist-core/src/main/java/org/exist/util/io/AutoReleaseLock.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.exist.util.io;
+
+import java.util.concurrent.locks.Lock;
+
+/**
+ * @author Patrick Reinhart <patrick@reini.net>
+ */
+final class AutoReleaseLock implements AutoCloseable {
```
adamretter: I think the `AutoReleaseLock` is a lovely idea, but we already have an equivalent in the `develop` branch - `ManagedLock`. So I think having this might make creating the forward ports a bit harder for you.

exist-core/src/main/java/org/exist/util/io/ByteArrayConsumer.java
```diff
@@ -0,0 +1,34 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package org.exist.util.io;
+
+/**
+ * @author Patrick Reinhart <patrick@reini.net>
+ */
+@FunctionalInterface
+interface ByteArrayConsumer {
```
adamretter: I am just wondering why you didn't just use a `Consumer(byte[])`

exist-core/src/main/java/org/exist/util/io/MemoryContentsImpl.java
```diff
@@ -0,0 +1,219 @@
...
+import static java.lang.Math.max;
+import static java.lang.Math.min;
+import static org.exist.util.io.AutoReleaseLock.autoRelease;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import org.apache.commons.logging.Log;
```
adamretter: In the core, we use the LOG4j2 API directly.

exist-core/src/main/java/org/exist/util/io/MemoryContentsInputStream.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+        boolean success = false;
+        int read = 0;
+        while (!success) {
+            long positionBefore = POSITION_UPDATER.get(this);
+            read = this.memoryContents.read(b, positionBefore, off, len);
```
adamretter: I am afraid that I think you have some race conditions in this class.

You have two invariants: `memoryContents` and `position` which I think have to be kept in sync. They are not read or written as an atomic unit, you have two different synchronisations. So it is possible when two or more threads are operating for the `position` not to reflect the state of the `memoryContents` as seen by another thread.

exist-core/src/main/java/org/exist/util/io/MemoryContentsOutputStream.java
```diff
@@ -0,0 +1,49 @@
...
+    private final byte[] singleByteBuffer = new byte[1];
+    private final MemoryContents memoryContents;
+
+    MemoryContentsOutputStream(MemoryContents memoryContents) {
+        this.memoryContents = memoryContents;
+    }
+
+    @Override
+    public void write(int b) throws IOException {
+        synchronized (singleByteBuffer) {
```
adamretter: I don't understand the purpose of buffering a single byte here?

exist-core/src/main/java/org/exist/util/io/OutputStreamSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+@FunctionalInterface
+interface OutputStreamSupplier {
+    /**
+     * Returns a newly created {@link OutputStream} instance. The receiver is
+     * responsible to close the stream after use.
+     *
+     * @return a new output stream instance
+     * @throws IOException if a error while creating the steam occurs
+     */
+    OutputStream get() throws IOException;
```
adamretter: Do we need this class? Why not just use `Supplier<OutputStream>`?

exist-core/src/main/java/org/exist/util/io/OverflowToDiskStream.java
```diff
@@ -0,0 +1,125 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package org.exist.util.io;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+import org.apache.commons.logging.Log;
```
adamretter: Please use the Log4j2API

exist-core/src/main/java/org/exist/util/io/VirtualTempPath.java
```diff
@@ -0,0 +1,122 @@
...
+import java.nio.file.Path;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.exist.util.FileUtils;
+
+/**
+ * @author Patrick Reinhart <patrick@reini.net>
+ */
+public final class VirtualTempPath implements AutoCloseable {
```
adamretter: I am wondering what the concurrency for this class is? Is it intended to be `@NotThreadSafe` or `@ThreadSafe`?

I note that the `content` member looks like it is designed to be thread-safe, but this class itself is not-thread-safe, due to its unsynchronized `inMemorySize`.

exist-core/src/main/java/org/exist/xmldb/AbstractRemoteResource.java
```diff
@@ -354,6 +358,14 @@ protected void getRemoteContentIntoLocalFile(final OutputStream os, final boolea
         }
     }
 
+    private static int getInMemorySize(Properties properties) {
```
adamretter: You could also consider using a LazyVal for this. e.g.

```java
private final LazyVal<Integer> inMemoryBufferSize = new new LazyVal<>(() -> Integer.parseInt(properties.getProperty("in-memory-buffer-size", "4194304")));
```

exist-core/src/main/java/org/exist/util/io/ByteArrayConsumer.java
reinhapa: This interface is no longer needed get removed with my next commit

exist-core/src/main/java/org/exist/util/io/OutputStreamSupplier.java
reinhapa: The Java functions do not support declared exception as the `IOException` in this case...

adamretter: You can use `SupplierE<T, IOException>` ;-)

exist-core/src/main/java/org/exist/util/io/VirtualTempPath.java
reinhapa: Hmm, technically it is `@ThreadSafe`... Also how should the final field `inMemorySize` set within the Constructor be a problem? Did I miss something here?

adamretter: Sorry. `inMemorySize` is correct as it is `final`. But what about the visibility of `content` and `contentFile`? They are lazily initialised and not as an an atomic unit. If I am not mistaken, they need to be atomic references.

exist-core/src/main/java/org/exist/util/io/VirtualTempPath.java
```diff
@@ -41,73 +42,102 @@
...
 
     public OutputStream newOutputStream() throws IOException {
-        if (inMemorySize <= 0) {
-            contentFile = tempFileManager.getTemporaryFile();
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("In memory buffering disabled writing to " + contentFile.toAbsolutePath());
+        long stamp = lock.readLock();
+        try {
+            if (inMemorySize <= 0) {
+                contentFile = tempFileManager.getTemporaryFile();
```
adamretter: You have a write to `contentFile` here happening under a read lock!?!

It might be useful to add `@GuardedBy` annotations to `content` and `contentFile` members so that FindBugs plugin will show you such issues ;-)

adamretter:
Good work @ptrckb. Thanks for your contributions :-)

PR #2628, opened 2019-04-08
Fast faceted search

wolfgangmm:
This adds fast faceted search to the lucene-based full text index. A demo showing the new features in action can be viewed here: https://teipublisher.com/exist/apps/vangogh/ (facets shown in the sidebar to the right).

A full text index defined in `collection.xconf` can now specify additional facets using an arbitrary xpath rooted in the indexed element. The facets are created automatically at indexing time.

```xml
<lucene>
    <text qname="tei:body">
        <facet dimension="from" expression="ancestor::tei:TEI//tei:sourceDesc/vg:letDesc/vg:letHeading/tei:author"/>
        <facet dimension="to" expression="ancestor::tei:TEI//tei:sourceDesc/vg:letDesc/vg:letHeading/vg:addressee"/>
    </text>
</lucene>
```

All result sequences obtained from a query with `ft:query` will have facet counts attached, which can be retrieved for display to the user. The user may then select a particular facet to drill down into the result. Drill down is super fast on large result sets as no additional computation needs to be done at runtime. See the test cases for usage examples.

In addition to supporting facets, this PR also enables fields to be defined on a parent index in the same way as facets. Fields are used to index content which is not directly contained in the XML of the parent node or needs to be pre-processed to be useful for searching. A typical example would be dates which need to be normalized to enable sensible searching or sorting. Fields may optionally be stored along the indexed parent node, allowing speedy sorting or display of computed information. They also provide a means to support multi-lingual content.

Documentation: https://github.com/eXist-db/documentation/pull/325.

Port of #2614 to develop. The original PR will be closed as the feature is not planned to be supported on 4.x.x.


extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneConfig.java
```diff
@@ -219,6 +240,10 @@ public FieldType getFieldType(String name){
         return fieldTypes.get(name);
     }
 
+    public List<ModuleImport> getImports() {
+        return imports;
```
dizzzz: the value van be null .... is t an option to use `Optional<>` here?

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
```diff
@@ -0,0 +1,112 @@
...
+                final String value = i.nextItem().getStringValue();
+                if (value.length() > 0) {
+                    luceneDoc.add(new FacetField(dimension, value));
+                }
+            }
+        }
+    }
+
+    @Override
+    protected void processText(CharSequence text, Document luceneDoc) {
```
dizzzz: can `text` be null ?

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFieldConfig.java
```diff
@@ -0,0 +1,260 @@
...
+                    return new LongField(fieldName, tl, LongField.TYPE_STORED);
+                case Type.DATE_TIME:
+                    DateTimeValue dtv = new DateTimeValue(content);
+                    String dateStr = dateTimeToString(dtv);
+                    return new TextField(fieldName, dateStr, Field.Store.YES);
+                default:
+                    return new TextField(fieldName, content, store ? Field.Store.YES : Field.Store.NO);
+            }
+        } catch (NumberFormatException | XPathException e) {
+            // wrong type: ignore
```
dizzzz: add logging on TRACE logging?

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
wolfgangmm: Nope. The class is called from a single place and neither `text` nor `luceneDoc` can be null.

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFieldConfig.java
wolfgangmm: Ok, done.

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneMatch.java
```diff
@@ -0,0 +1,235 @@
...
+            return values[0].getValue(type);
+        }
+        final ValueSequence result = new ValueSequence(values.length);
+        for (FieldValue value: values) {
+            result.add(value.getValue(type));
+        }
+        return result;
+    }
+
+    // DW: missing hashCode() ?
```
dizzzz: is it missing? required for comparable interface?

PR #2632, opened 2019-04-09
Fix use of deprecated functions to restore dashboard etc.

wolfgangmm:
Replace xmldb:is-admin-user with sm:is-dba to restore packaged apps for develop and 5.0.

### DIFF
--- a/build/scripts/extensions-common.xml
+++ b/build/scripts/extensions-common.xml
@@ -360,6 +360,7 @@ to add: -pre-compile steps
             <path refid="classpath.extensions"/>
             <path refid="classpath.test"/>
             <path refid="classpath.aspectj"/>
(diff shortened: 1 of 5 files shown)


adamretter:
As this is part of the core, coukd you add some XQuery test please, so that we know when this breaks in future?

joewiz:
I have tested this PR and confirmed that it allows apps that otherwise couldn't start (such as dashboard, monex, and packagemanager) to work. 

I added the "needs 4.x.x backport" just as a suggestion. It isn't strictly necessary, but it would be nice to keep the source aligned where possible across branches/versions.

wolfgangmm:
Ok, added a junit test to cover this module.

dizzzz:
@adamretter ?

PR #2633, opened 2019-04-09
(5.0.0) Fix use of deprecated functions to restore dashboard etc.

wolfgangmm:
Port of #2632


adamretter:
Could you put the test into the correct module please? I think you accidentally put it in `core`.

adamretter:
@wolfgangmm I am afraid that your test seems to fail...

PR #2651, opened 2019-04-23
EXPath HTTP should respect charset in override-media-type

adamretter:


adamretter:
@duncdrum That sort of testing should happen in the upstream project not here

duncdrum:
@adamretter agreed, so [here](https://github.com/adamretter/expath-http-client-java/commit/6d5e8373b2e86aa9264dd545ba3569d10b3d1d70) then?

Issue #2278, opened 2018-11-08
String index out of range when using UCA collation to sort >8,192 strings


joewiz:
### What is the problem

When using the `numeric=yes` option with the UCA collation in an `order by` clause of a FLWOR expression, sorting more than >8,192 strings will raise an "String index out of range" error.

### What did you expect

I expected there to be no fixed limit to the number of strings that could be sorted.

For the documentation on the `numeric=yes` option, see https://www.w3.org/TR/xpath-functions-31/#uca-collations. Support for the UCA collation was added in https://github.com/eXist-db/exist/pull/1575.

### Describe how to reproduce or add a test

The following query and the variant described below illustrate how to reproduce:

```xquery
xquery version "3.1";

let $terms := (1 to 8193) ! (. cast as xs:string)
for $term in $terms
order by $term collation "http://www.w3.org/2013/collation/UCA?numeric=yes"
return 
    $term
```

This raises no error and works correctly, returning the "terms" (strings) in *numeric* order (1, 2, 3) rather than in string order (1, 10, 100). **However,** if you change "8192" to "8193", you will trigger an "String index out of range" error [1]. 

Saxon 9.8.0.12 (run via oXygen 20.1) and BaseX 9.1 (with icu4j-63_1.jar added to the classpath as per http://docs.basex.org/wiki/XQuery_3.1#Collations) returns the expected results without error.

Moreover, if you submit the query several times in succession, one of the queries will hang instead of returning an error quickly, and monex will perpetually report the stack is in a state of waiting. [2] However, memory and CPU usage is idle, and the query cannot be killed (e.g., via monex). The only way out is to force kill eXist.

**[1] exist.log**

```text
2018-11-07 22:27:30,775 [qtp1159770377-88] ERROR (XQueryServlet.java [process]:552) - String index out of range: 2 
java.lang.StringIndexOutOfBoundsException: String index out of range: 2
	at java.lang.StringLatin1.charAt(StringLatin1.java:47) ~[?:?]
	at java.lang.String.charAt(String.java:693) ~[?:?]
	at com.ibm.icu.impl.coll.UTF16CollationIterator.handleNextCE32(UTF16CollationIterator.java:107) ~[icu4j-59_1.jar:59.1]
	at com.ibm.icu.impl.coll.CollationIterator.nextCE(CollationIterator.java:247) ~[icu4j-59_1.jar:59.1]
[...]

PR #2667, opened 2019-05-03
Fix sorting sequences of more than 8192 items with collations

adamretter:
Closes https://github.com/eXist-db/exist/issues/2278

Also adds a couple of small optimisations.


dizzzz:
ideally there would be testcase.. for the corner case?

adamretter:
@dizzzz Indeed. Unfortunately any such test case which shows the problem actually causes eXist-db Query thread to hang due to an issue with the design of parallel sorting in the JDK. So if we don't want the test suite to hang, it is best avoided ;-)

PR #2367, opened 2019-01-05
Allow more elaborate XPath expressions in the Lucene index spec (original PR #2169)

sten1ee:
Allow more elaborate XPath expressions in the Lucene index spec in collection.xconf (original PR #2169)
For more information, see the discussion in the original PR: https://github.com/eXist-db/exist/pull/2169

PR #2680, opened 2019-05-05
Descendant nodes wrongly indexed, fix PR #2367

wolfgangmm:
Fixes an issue introduced by PR #2367: a lucene index configured with a match attribute should only be created for nodes matching the path, not all its descendant nodes.


adamretter:
Good catch!

PR #2681, opened 2019-05-05
Fix two issues with Lucene facets

wolfgangmm:
This fixes two issues related to Lucene facets:

1. calling `ft:query` with an empty query parameter (to match all indexed elements) would pass an `MatchAllDocsQuery` to Lucene. This is wrong though as it would include Lucene matches in other elements than the current context node. While the XPath result was correct, facet counts got messed up. The PR fixes this by taking the context node into account.
2. `ft:facet` expected a single node as input sequence to return facet counts for. This is inconvenient if  you want to retrieve facet counts for a sequence of nodes resulting from **multiple** lucene queries. `ft:facet` will now accept an entire sequence and merge the facet counts for each lucene query referenced from nodes in the sequence.

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Facets.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Facets.java
@@ -21,6 +21,7 @@
 package org.exist.xquery.modules.lucene;
 
 import org.apache.lucene.facet.FacetResult;
+import org.apache.lucene.search.Query;
 import org.exist.dom.QName;
 import org.exist.dom.persistent.Match;
 import org.exist.dom.persistent.NodeProxy;
@@ -32,6 +33,8 @@ import org.exist.xquery.functions.map.MapType;
 import org.exist.xquery.value.*;
 
 import java.io.IOException;
+import java.util.IdentityHashMap;
+import java.util.Map;
 
 public class Facets extends BasicFunction {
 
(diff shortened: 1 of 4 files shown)


extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Facets.java
```diff
@@ -42,13 +45,13 @@
                         "to the entire sequence returned by ft:query, so the same map will be returned for all nodes in the sequence. " +
                         "It is thus sufficient to specify one node from the sequence as first argument to this function.",
                 new SequenceType[] {
-                        new FunctionParameterSequenceType("node", Type.NODE, Cardinality.EXACTLY_ONE,
+                        new FunctionParameterSequenceType("node", Type.NODE, Cardinality.ZERO_OR_MORE,
```
adamretter: Description needs to be updated to match the new cardinality. Also is this meant to be `ZERO_OR_MORE` instead of `ONE_OR_MORE`?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Facets.java
```diff
@@ -75,7 +78,7 @@
                     "to the entire sequence returned by ft:query, so the same map will be returned for all nodes in the sequence. " +
                     "It is thus sufficient to specify one node from the sequence as first argument to this function.",
             new SequenceType[] {
-                    new FunctionParameterSequenceType("node", Type.NODE, Cardinality.EXACTLY_ONE,
+                    new FunctionParameterSequenceType("nodes", Type.NODE, Cardinality.ZERO_OR_MORE,
                             "A single node resulting from a call to ft:query for which facet information should be retrieved. " +
```
adamretter: Again the description needs updating. Not sure about the cardinality (as above).

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Facets.java
wolfgangmm: Yes, indeed. I corrected the description. `ZERO_OR_MORE` is correct. The function will return an empty map if an empty sequence is passed in.

dizzzz:
build is ok.....

dizzzz:
@adamretter are you OK ?

adamretter:
@dizzzz LGTM now. Just awaiting CI...

PR #2683, opened 2019-05-05
(4.x.x) Descendant nodes wrongly indexed, fix PR #2367

wolfgangmm:
Backport of #2680


joewiz:
Codacy is the only CI failure, and it looks like a glitch on their side.

PR #2687, opened 2019-05-09
Reduce mutable state in MutableCollection

adamretter:
This removes two fields from `MutableCollection` which were unnecessarily located there as they are not persisted. These are:

* `userReader` - this can now instead be supplied when calling `Collection#validateXMLResource` and `Collection#store`.
* `triggersEnabled` - this can now be set on the `DBBroker` instead.

`triggersEnabled` was only modified when restoring a backup which contained deleted entries from a system export. To pass `triggersEnabled` to a `DBBroker` across the XML:DB (remote) API, required refactoring eXist-db's restore code, so that `__contents__.xml` parsing always happens server-side. To this end we have introduced the XML:DB service `RestoreService`. This actually puts us in a better position in future to be able to offer a more consistent database backup/restore.


adamretter:
@wolfgangmm from the user perspective of performing backup/restore there are no changes.

The Java Admin Client is no longer compatible between 4.x.x and 5.0.0 but I suspect it hasn't been for all possible operations for quite some time. Where should we document this? In the release notes?

joewiz:
As discussed during the Community Call, we would appreciate if folks would test this PR and see if you can restore your backups, particularly larger ones.

dizzzz:
I have a large backup set here, I can try

duncdrum:
@adamretter pre-emptively upgrade instructions should appear in the [upgrading article](https://github.com/eXist-db/documentation/blob/a9bcb918d2ed9aaa0d0a5a5b62bfc75f0695f2af/src/main/xar-resources/data/upgrading/upgrading.xml)

duncdrum:
see #1358 ??

adamretter:
@duncdrum there is some overlap, but its not the same

adamretter:
@dizzzz Yup. At that point I had only added the test and not the fix. All fixed now...

duncdrum:
Is there still a need for adjusted documentation @wolfgangmm @adamretter ?

adamretter:
@duncdrum AFAIK, the changes should be invisible to the user.

Issue #2685, opened 2019-05-06
IllegalMonitorState in LockTable


wolfgangmm:
I encountered the following exception using current develop. Any idea what could cause it?

```java
Caused by: java.lang.IllegalMonitorStateException
        at java.util.concurrent.locks.StampedLock.unlockRead(StampedLock.java:563) ~[?:1.8.0_181]
        at org.exist.storage.lock.LockTable$Entries.unmerge(LockTable.java:385) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.storage.lock.LockTable.event(LockTable.java:230) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.storage.lock.LockTable.released(LockTable.java:126) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.storage.lock.LockManager.lambda$null$2(LockManager.java:201) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.storage.lock.LockManager.unlockAll(LockManager.java:249) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.storage.lock.LockManager.lambda$acquireCollectionReadLock$3(LockManager.java:201) ~[exist.jar:
5.0.0-SNAPSHOT]
        at org.exist.storage.lock.ManagedLock.close(ManagedLock.java:186) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.collections.LockedCollection.close(LockedCollection.java:90) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.storage.NativeBroker.getXMLResource(NativeBroker.java:2233) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.source.DBSource.isValid(DBSource.java:91) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.xquery.ExternalModuleImpl.moduleIsValid(ExternalModuleImpl.java:267) ~[exist.jar:5.0.0-SNAPSHO
T]
        at org.exist.xquery.XQueryContext.checkModulesValid(XQueryContext.java:1487) ~[exist.jar:5.0.0-SNAPSHOT]
        at org.exist.xquery.PathExpr.isValid(PathExpr.java:526) ~[exist.jar:5.0.0-SNAPSHOT]
[...]

PR #2689, opened 2019-05-09
Fix an issue with LockTable Entries#unmerge under high concurrency

adamretter:
Closes https://github.com/eXist-db/exist/issues/2685

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/lock/LockTable.java
+++ b/exist-core/src/main/java/org/exist/storage/lock/LockTable.java
@@ -344,6 +344,7 @@ public class LockTable {
                 }
             }
 
+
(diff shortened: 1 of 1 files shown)


exist-core/src/main/java/org/exist/storage/lock/LockTable.java
```diff
@@ -357,13 +358,10 @@ public Entry unmerge(final String id, final LockType lockType, final LockMode lo
 
                             long writeStamp = entriesLock.tryConvertToWriteLock(stamp);
                             if (writeStamp != 0l) {
```
dizzzz: 0l .. better to make a magic number? or write as 0L ?

PR #2692, opened 2019-05-13
Opening a collection using full XmldbURI (with scheme) fails.

### LINKED ISSUE #2691: Controller fails on first access to app
5.0/develop: the first time one tries to access the root page of any of the installed apps, only a collection is returned. The controller is not evaluated/fails due to the collection not being found.

wolfgangmm:
Loading an existing collection fails on develop/5.0 when calling `DBBroker.openCollection` with a full XmldbURI (i.e. with scheme part). This manifests itself in issue #2691, where accessing the root of an app shows just a collection listing. The cause for this is that the collection is not found when searching for the controller. It is found later by the REST servlet though, because this uses an XmldbURI without scheme part.

The issue does not apply to 4.x.x. The uri is correctly processed there.

Closes #2691

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/NativeBroker.java
+++ b/exist-core/src/main/java/org/exist/storage/NativeBroker.java
@@ -650,7 +650,7 @@ public class NativeBroker extends DBBroker {
      * @throws TriggerException If a CollectionTrigger throws an exception
      */
     private Tuple2<Boolean, Collection> getOrCreateCollectionExplicit(final Txn transaction, final XmldbURI path, final Optional<Tuple2<Permission, Long>> creationAttributes) throws PermissionDeniedException, IOException, TriggerException {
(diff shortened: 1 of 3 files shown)


adamretter:
@wolfgangmm so you are trying to open a collection with a URI like `xmldb:exist:///db/collection`?

wolfgangmm:
@adamretter yes, passing in a FullXmldbURI used to work in 4.x.x and I see no reason why it should no longer.

adamretter:
Ah okay!

That won't work I am afraid. As the URI is used as the object lock. This is because the URIs `xmldb:exist:///db` and `/db` are not the same thing.

I think you will have to fix this at the source.

wolfgangmm:
No, I have checked this: for locking and caching, we always use the URI without scheme. By normalizing the incoming URI as we did before we're fine.

adamretter:
@wolfgangmm Well okay then! :-)

PR #2701, opened 2019-05-21
Make doc id in DocumentImpl immutable

adamretter:
Closes https://github.com/eXist-db/exist/pull/1358 when considered alongside https://github.com/eXist-db/exist/pull/2687

### DIFF
--- a/exist-core/src/main/java/org/exist/dom/persistent/BinaryDocument.java
+++ b/exist-core/src/main/java/org/exist/dom/persistent/BinaryDocument.java
@@ -23,9 +23,10 @@
 package org.exist.dom.persistent;
 
 import org.exist.collections.Collection;
+import org.exist.security.Permission;
+import org.exist.security.PermissionFactory;
 import org.exist.storage.BrokerPool;
 import org.exist.storage.blob.BlobId;
-import org.exist.storage.btree.Paged.Page;
 import org.exist.storage.io.VariableByteInput;
 import org.exist.storage.io.VariableByteOutputStream;
 import org.exist.xmldb.XmldbURI;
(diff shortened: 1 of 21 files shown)


wolfgangmm:
Could you add some explanation of what the PR is expected to achieve? I find it difficult to judge from the code...

adamretter:
@wolfgangmm The goal is to make more data in `DocumentImpl` immutable, in this instance the `docId` field.

This complements the immutable changes I recently made to `MutableCollection` (https://github.com/eXist-db/exist/pull/26870).

I am just trying to reduce the amount of mutable state that is shared between transactions via the Collection Cache. This reduces the number of likely concurrency issues.

This work is actually back-port from some foundation work done previously in FusionDB. It does have relevance for improving transaction isolation in future in eXist-db though

adamretter:
@wolfgangmm I also now back-ported a couple more changes from FusionDB whereby the immutability identifies and fixes a problem in the WebDAV locking code. You can see the issue here in `4.7.0` - https://github.com/eXist-db/exist/blob/eXist-4.7.0/extensions/webdav/src/main/java/org/exist/webdav/MiltonResource.java#L211

It should call `setType` and not `setScope` but because mutable bytes were used for all the flag types it compiled anyway, switching to immutable enum flags in this PR identified the problem, and fixes it.

exist-core/src/main/java/org/exist/dom/persistent/BinaryDocument.java
```diff
@@ -45,39 +46,45 @@
...
      * Creates a new persistent binary Document instance.
      *
      * @param pool The broker pool
      * @param collection The Collection which holds this document
+     * @param docId the id of the document
      * @param fileURI The name of the document
      */
-    public BinaryDocument(final BrokerPool pool, final Collection collection, final XmldbURI fileURI) {
-        super(pool, collection, fileURI);
+    public BinaryDocument(final BrokerPool pool, final Collection collection, final int docId, final XmldbURI fileURI) {
```
dizzzz: I think for some of my extensions this is an API change.
Not a problem for me as I'd check and refactor for v5.x anyway

exist-core/src/main/java/org/exist/dom/persistent/LockToken.java
```diff
@@ -146,29 +123,136 @@ public static String generateUUID() {
...
-        ostream.writeByte(resourceType);
+        ostream.writeByte(resourceType.getValue());
     }
 
-    public void read(final VariableByteInput istream) throws IOException {
-        type = istream.readByte();
-        depth = istream.readByte();
-        scope = istream.readByte();
-        owner = istream.readUTF();
+    public static LockToken read(final VariableByteInput istream) throws IOException {
```
dizzzz: Ideally... this meta data would be stored in a non-shared dbx file or so? I mean, per document/collection quite a lot of bytes are preserved PLUS a string representation of the lock-owner.

dizzzz:
Huge commit! And... the code is not alone about immutability? The changes ok to me, smaller remarks....

PR #2703, opened 2019-05-22
Lucene: fix AlreadyClosedException

### LINKED ISSUE #2700: "this IndexReader is closed" error during reindex with facets
### What is the problem

When using the facets feature in 5.0.0 (develop) with stock TEI Publisher (master), I am hitting a “this IndexReader is closed” error when loading a page that queries facets during a reindex. For example, running `xmldb:reindex("/db/apps/tei-publisher")` and then requesting http://localhost:8080/exist/apps/tei-publisher/modules/facets.xql?sort=title before the reindex is complete will yield a 500 error, culminating in:

```
org.apache.lucene.store.AlreadyClosedException: this IndexReader is closed
    at org.apache.lucene.index.IndexReader.ensureOpen(IndexReader.java:276)
    at org.apache.lucene.index.SegmentReader.getBinaryDocValues(SegmentReader.java:501)
    at org.apache.lucene.facet.taxonomy.FastTaxonomyFacetCounts.count(FastTaxonomyFacetCounts.java:54)
    at org.apache.lucene.facet.taxonomy.FastTaxonomyFacetCounts.<init>(FastTaxonomyFacetCounts.java:49)
    at org.apache.lucene.facet.taxonomy.FastTaxonomyFacetCounts.<init>(FastTaxonomyFacetCounts.java:39)
    at org.exist.indexing.lucene.LuceneIndexWorker.lambda$getFacets$3(LuceneIndexWorker.java:554)
    at org.exist.indexing.lucene.LuceneIndex.withSearcher(LuceneIndex.java:259)
    at org.exist.indexing.lucene.LuceneIndexWorker.getFacets(LuceneIndexWorker.java:554)
    at org.exist.xquery.modules.lucene.Facets.addFacetsToMap(Facets.java:151)
    at org.exist.xquery.modules.lucene.Facets.eval(Facets.java:141)
```

This seems to describe the problem and a possible solution too:

https://stackoverflow.com/questions/41437582/lucene-alreadyclosedexception-this-indexreader-is-closed

### What did you expect

I expected to be able to make requests to the application during reindexing without encountering a 500 error. I hope this is surmountable. Otherwise, production instances cannot perform reindexing or even perhaps loading large amounts of data.

See description above.

### Context information

- eXist-db version + Git Revision hash: 5.0.0 (current develop)
- Java version: OpenJDK 12.0.1+12
- Operating system: macOS 10.14.5
- 32 or 64 bit: 64 bit
- How is eXist-db installed? Built from source
- Any custom changes in e.g. conf.xml: no

wolfgangmm:
Fix AlreadyClosedException when retrieving facets during a reindex. Previously we were postponing the actual facet computation until the user requested it via `ft:facets`. Internally lucene would potentially use an outdated reader for the computation though, which led to the exception.

Instead of postponing, compute facets immediately within the same lucene search context and thus using the same reader. Memory and performance-wise, the overhead is minimal.

Closes #2700

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneMatch.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneMatch.java
@@ -20,6 +20,7 @@
 
 package org.exist.indexing.lucene;
 
+import org.apache.lucene.facet.Facets;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.search.Query;
 import org.exist.dom.persistent.Match;
@@ -41,7 +42,7 @@ public class LuceneMatch extends Match {
     private float score = 0.0f;
     private final Query query;
 
-    private LuceneIndexWorker.ExtFacetsCollector facetsCollector;
+    private LuceneIndexWorker.LuceneFacets facets;
 
     private Map<String, FieldValue[]> fields = null;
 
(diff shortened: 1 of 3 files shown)


extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneMatch.java
```diff
@@ -41,25 +42,25 @@
     private float score = 0.0f;
     private final Query query;
 
-    private LuceneIndexWorker.ExtFacetsCollector facetsCollector;
+    private LuceneIndexWorker.LuceneFacets facets;
```
adamretter: Could this be `final`?

adamretter:
The code looks okay but there are no tests. So I will need to rely on @joewiz's review to determine if it fixes his problem...

PR #2706, opened 2019-05-22
Fix the logo image path in README.md

adamretter:


adamretter:
@joewiz Okay, I corrected that as you suggested.

Issue #1248, opened 2017-02-02
"eXist-db" app remains running after first install on macOS


joewiz:
When I first run eXist-db.app using current nightly builds on macOS, the eXist-db icon that appears in my dock doesn't close when I quit eXist via the menu bar.  It seems the only way to get this icon to go away is to control-click on it and select quit.  

This is odd, since it's not associated with the eXist-db process. It seems harmless but will certainly confuse users. 

The "normal" behavior I would expect from eXist-db is that the dock icon appears briefly and then disappears after the menu bar icon appears.  

Since this persistent dock icon only appears once - on first run of a new installation - my hunch is that this is somehow related to something that only happens on first run and not subsequent runs.  But I'm not sure what it is.

duncdrum:
From the activity monitor screenshot in #1275 it appears to be doing something….

adamretter:
@dizzzz I don't think I am the correct person for this; I always wanted to drop the IzPack installer and Mac DMG etc ;-)

joewiz:
(I've added a proposed topic for a future call about dropping the IzPack installer and the Mac DMG. See the current draft of the [2017-02-27 - eXist Community Call](https://docs.google.com/document/d/116O2kVM5_7QNsL7DVe21y-GFM7OQKi1FjfD8NLnqFXQ/edit?usp=sharing) under Agenda section 5.a. Any topic is ok to propose, but I'm especially trying to capture knotty issues like this for future discussion.)

adamretter:
@joewiz Awesome, I think I would prefer to see platform native packages, but on Mac maybe use your HomeBrew, but have it build from a .tar.gz.

duncdrum:
Just a comment on the homebrew vs  .jar vs .dmg topic for mac OS. How the app gets into the applications folder is less of an issue then the different default data directories of each installer. This could make for some very unpleasant upgrading with either an old directory collecting dust in a hidden systems folder, or with accidental overwrites which could break restorability. Especially going from 2.2 to 3.x.

duncdrum:
same with 3.1 installer

joewiz:
I just noticed several things that might explain why first-runs of 3.0 (release) and now also 3.1.0 result in the eXist-db app icon remaining active even after the app has completed launching the eXist-db process, whereas this was not the case with 2.2 or 3.0RC1. The differences are evident when comparing 3.0RC1 to 3.1.0, while monitoring the dock, menu bar, and Activity Monitor during first launch. 

[...]

PR #2729, opened 2019-05-23
LauncherWrapper not exiting after first start

wolfgangmm:
Fixes issue #1248. During first start, the launcher shows a configuration dialog and restarts after it is closed. The wrapper process kept running though.

Checked on OSX and Windows 10.

### DIFF
--- a/exist-core/src/main/java/org/exist/launcher/LauncherWrapper.java
+++ b/exist-core/src/main/java/org/exist/launcher/LauncherWrapper.java
@@ -55,6 +55,8 @@ public class LauncherWrapper {
             System.out.println("First launch: opening configuration dialog");
             ConfigurationDialog configDialog = new ConfigurationDialog(restart -> {
                 wrapper.launch();
+                // make sure the process dies when the dialog is closed
+                System.exit(0);
             });
             configDialog.open(true);
             configDialog.requestFocus();
--- a/exist-core/src/main/java/org/exist/launcher/ConfigurationDialog.java
+++ b/exist-core/src/main/java/org/exist/launcher/ConfigurationDialog.java
@@ -107,6 +107,9 @@ public class ConfigurationDialog extends JDialog {
         }
         setVisible(true);
         requestFocus();
+        // make sure the process dies when the dialog is closed
+        // we cannot set this before the dialog is open though
+        setDefaultCloseOperation(EXIT_ON_CLOSE);
     }
 
     /**
--- a/exist-core/src/main/java/org/exist/launcher/ConfigurationDialog.java
+++ b/exist-core/src/main/java/org/exist/launcher/ConfigurationDialog.java
@@ -107,9 +107,6 @@ public class ConfigurationDialog extends JDialog {
         }
         setVisible(true);
         requestFocus();
-        // make sure the process dies when the dialog is closed
-        // we cannot set this before the dialog is open though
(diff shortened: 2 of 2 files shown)


joewiz:
Looks like the Appveyor CI failure was on one of the four images, and it looks like a timeout - died at 1 hr, in contrast to the 22-38 min taken up by the other 3 tests.

wolfgangmm:
Wait, small change required as the option was removed in Java 11...

duncdrum:
sounds like we should set up an integration test, to avoid future regressions

wolfgangmm:
@duncdrum I have absolutely no idea how you would test this ;-)

wolfgangmm:
Ok, updated the PR. The more elegant approach of calling `setDefaultCloseOperation` unfortunately doesn't work reliably across platforms and JVMs (exception with Oracle 11 on windows, though not with Zulu 11), so we have to exit explicitly.

exist-core/src/main/java/org/exist/launcher/LauncherWrapper.java
```diff
@@ -55,6 +55,8 @@ public final static void main(String[] args) {
             System.out.println("First launch: opening configuration dialog");
             ConfigurationDialog configDialog = new ConfigurationDialog(restart -> {
                 wrapper.launch();
+                // make sure the process dies when the dialog is closed
+                System.exit(0);
```
adamretter: Won't this kill both processes? You are exiting the parent, so I think it will also kill the child i.e. the launcher...

wolfgangmm: No, a new wrapper is started in the line before, so we safely can and must stop the current one.

adamretter: @wolfgangmm I am afraid that I don't agree... at least on Unix systems. To be sure, I wrote a little test program to demonstrate the problem:

```java
public static void main(final String args[]) throws IOException, InterruptedException {
        ProcessBuilder pb = new ProcessBuilder();
        pb.command("read");  // read will run forever as it is listening to stdin
        final Process p = pb.start();

        p.waitFor(2, TimeUnit.SECONDS);

        System.out.println("sub-process `read` isAlive=" + p.isAlive());

        try {
            final Field f = p.getClass().getDeclaredField("pid");
            if (f != null) {
                f.setAccessible(true);
            }

            System.out.println("sub-process `read` pid=" + f.get(p));

        } catch (final NoSuchFieldException | IllegalAccessException e) {
            System.out.println("Could not determine pid of `read` sub-process");
        }

        System.exit(0);
        
        // when the main process exits the sub-process is also killed
        // you can check this by doing a `ps -ef | grep <pid>` where
        // <pid> is the process id reported above for the read sub-process
    }
```

This will show you that when the parent process exists the sub-process is killed.

adamretter: Actually never mind. I see my error now. In the example above the stdin is inheirted from the parent process and so when it exits, it closes stdin, which causes `read` to exit.

If I replace "read" with "/tmp/forever.sh":
```bash
#!/bin/bash
while :; do sleep 2073600; done
```

Then I see that the sub-process continues after the parent process exists. The sub-process then acquires the root process `1` as its parent.

So... LGTM sorry for the confusion.

PR #2737, opened 2019-05-24
Installing service fails with OpenJDK

### LINKED ISSUE #2731: Installing service fails with OpenJDK
### What is the problem
Using windows 10, installing eXist as a service fails when using one of the OpenJDK-based Java distributions (tried Zulu and Amazon Corretto). According to procrun log, it fails to create Java and prints a "file not found".

wolfgangmm:
Auto-detection of `jvm.dll` needed by procrun does not work reliably across different JDK distributions (ok with Oracle, fails with Zulu and Amazon Corretto). We can easily determine the correct location from `JAVA_HOME` though.

Closes #2731

Tested on Windows 10 with Zulu 8 and 11, Amazon Corretto 11 and Oracle JDK 11.

### DIFF
--- a/exist-core/src/main/java/org/exist/launcher/WindowsServiceManager.java
+++ b/exist-core/src/main/java/org/exist/launcher/WindowsServiceManager.java
@@ -109,7 +109,7 @@ class WindowsServiceManager implements ServiceManager {
                 "--LogPrefix=service",
                 "--PidFile=service.pid",
                 "--Startup=auto",
-                "--Jvm=auto",
+                "--Jvm=" + findJvm().orElse("auto"),
                 "--Classpath=\"" + existHome.resolve("lib").toAbsolutePath().toString().replace('\\', '/') + "/*\"",
                 "--JvmMs=" + minMemory,
                 "--StartMode=jvm",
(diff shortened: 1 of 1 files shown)


joewiz:
Sounds a lot like the phenomenon we're having with appbundler finding JDKs other than the Oracle JDK: https://bitbucket.org/infinitekind/appbundler/pull-requests/55.

exist-core/src/main/java/org/exist/launcher/WindowsServiceManager.java
```diff
@@ -259,6 +259,25 @@ public void showNativeServiceManagementConsole() throws UnsupportedOperationExce
...
 
+    /**
+     * Try to find jvm.dll, which should either reside in `bin/client` or `bin/server` below
+     * JAVA_HOME. Autodetection does not seem to work with OpenJDK-based Java distributions.
+     *
+     * @return Path to jvm.dll or empty Optional
+     */
+    private Optional<String> findJvm() {
+        final Path javaHome = Paths.get(System.getProperty("java.home")).toAbsolutePath();
+        Path jvm = javaHome.resolve("bin/client/jvm.dll");
```
adamretter: If this is windows, should Paths not be written `\` instead of `/`? Also better not to use the separator at all and instead use `.resolve("bin").resolve("client").resolve("jvm.dll")

exist-core/src/main/java/org/exist/launcher/WindowsServiceManager.java
```diff
@@ -109,7 +109,7 @@ public void install() throws ServiceManagerException {
                 "--LogPrefix=service",
                 "--PidFile=service.pid",
                 "--Startup=auto",
-                "--Jvm=auto",
+                "--Jvm=" + findJvm().orElse("auto"),
```
adamretter: Do you plan to submit an issue upstream to procrun too?

exist-core/src/main/java/org/exist/launcher/WindowsServiceManager.java
dizzzz: @adamretter yep that is better

wolfgangmm: Ok, changed.

exist-core/src/main/java/org/exist/launcher/WindowsServiceManager.java
wolfgangmm: There already is a feature request and some discussion around the topic, which suggests the solution used in this PR. The documented behaviour of `--Jvm=auto` though is to use the registry to determine the location of `jvm.dll`, which for the mentioned Java distributions does not work correctly. It is disputable who's right or wrong here, but since we can easily determine the correct location, we should do so.

exist-core/src/main/java/org/exist/launcher/WindowsServiceManager.java
adamretter: Thanks

exist-core/src/main/java/org/exist/launcher/WindowsServiceManager.java
adamretter: Okay cool. Totally agree

Issue #2728, opened 2019-05-23
comment nodes incorrectly unescaped?


merenyics:
### What is the problem

Create a document with the following contents:
```
<root>
    <!-- text <a> &lt;b&gt;  -->
</root>
```

Save it in exide or upload to db via webdav.

Reopen the document in exide or via webdav. The result will be:

```
<root>
    <!-- text <a> <b>  -->
</root>
```

### What did you expect
As far a I know, there is no need to escape anything inside XML comments, all characters are literal, the only restriction is the sequence '--' is not allowed. Therefore I expected the original contents of the document returned by existdb.

### Context information
- eXist-db version 4.4
- Java version 1.8.0_161
- Operating system Windows 10
- 64 bit
- How is eXist-db installed? JAR installer

PR #2739, opened 2019-05-24
The content of comments should not be escaped

adamretter:
Closes https://github.com/eXist-db/exist/issues/2728


adamretter:
@duncdrum the problem with eXist-db/dashboard#88 is that you cannot have `--` inside a comment.

duncdrum:
@adamretter and in joerns build he escaped the `—`  inside a comment which somehow gets expanded into an not well-formed xml inside exist

dizzzz:
isn't there a 4.x port needed?

adamretter:
@dizzzz I understood from last-weeks community call that 4.7.0 was meant to be the last in 4.x.x line?

dizzzz:
make sense.... but maybe some bug fixes when it is really needed? (eg patch it in codebase so at least ppl can use it?)

PR #2756, opened 2019-05-27
Fixes tests failing if system language it not english

reinhapa:
Fixes all test failing if the system language is other than English.


joewiz:
@reinhapa Thank you very much!

PR #2757, opened 2019-05-27
(5.0.0) Fixes tests failing if system language it not english

reinhapa:
Fixes all test failing if the system language is other than English for `develop-5.0.0`


joewiz:
Port of #2756.

PR #2774, opened 2019-06-01
(5.0.0) Don't quieten Maven output on Travis-CI

adamretter:
Backport of https://github.com/eXist-db/exist/pull/2773

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ cache:
   directories:
   - $HOME/.m2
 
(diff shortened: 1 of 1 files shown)


joewiz:
AppVeyor failed...

adamretter:
@joewiz Unfortunately the public repo seems to be offline. Which leads to this in the AppVeyor log:
```
[ERROR] Received HTTP 502 when trying to access: http://demo.exist-db.org/exist/apps/public-repo/find?processor=5.0.0-RC8-SNAPSHOT&abbrev=dashboard&info=true
```
So we cannot build the `exist-distribution` as it fails to download the XARs...

PR #2792, opened 2019-06-09
Feature/implement fn:contains-token

tuurma:
implements fn:contains-token with relevant tests

### DIFF
--- a/exist-core/src/test/xquery/xquery3/fn.xql
+++ b/exist-core/src/test/xquery/xquery3/fn.xql
@@ -122,3 +122,37 @@ function fnt:outermost() {
             ","
         )
 };
+
+declare
+      %test:args('red green blue ', 'red')          %test:assertTrue
+      %test:args("red, green, blue", ' red ')       %test:assertFalse
+      %test:args('', 'red')                         %test:assertFalse
+      %test:args('red green blue', '')              %test:assertFalse
+      %test:args('red green blue', ' ')             %test:assertFalse
+      %test:args('red green blue ', 'RED ')         %test:assertFalse
+function fnt:contains-token-tests($input, $token) {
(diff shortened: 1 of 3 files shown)


exist-core/src/test/xquery/xquery3/fn.xql
```diff
@@ -122,3 +122,37 @@ function fnt:outermost() {
             ","
         )
 };
+
+declare
+      %test:args('red green blue ', 'red')          %test:assertTrue
+      %test:args("red, green, blue", ' red ')       %test:assertFalse
+      %test:args('', 'red')                         %test:assertFalse
```
duncdrum: can we add a test for `empty-sequence` here, i don't think that is the same as `' '` the specs say this should be false, so we shouldn't see a type error. For future maintainability i would also add a link to the relevant [spec page](https://www.w3.org/TR/xpath-functions-31/#func-contains-token) in an xqdoc comment

tuurma: @duncdrum https://docs.google.com/document/d/1xPZEa-dCcwt5vqZlTQR2oVc9A_sXdkvyCmWYVFxAKK0/edit# there already is a test for empty input sequence or did you mean for the token too?

duncdrum: Is empty string `””` the same as empty sequence ‘()’? If so all is good

tuurma: @duncdrum fnt:contains-token-tests-empty-input-sequence already tests the empty input sequence, see l.149. Thanks for spotting the ;;

tuurma:
just created documentation PR and ported to 5.0.0

PR #2800, opened 2019-06-10
Increase Travis timeout while silencing the logs

duncdrum:
Bump travis timeout for install phase to 20min, but reign in the noise. We can increase further but at some point we ll run into overall stalled build timeout limit of 50min.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ cache:
   directories:
   - $HOME/.m2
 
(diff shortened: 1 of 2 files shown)


joewiz:
@duncdrum Thanks! One question: Is there a reason the 1st commit is here - related to the hsqldb?

duncdrum:
actually no, let me clean this up

Issue #2686, opened 2019-05-09
WeakLazyStripes.drainClearedReferences kills performance


wolfgangmm:
### What is the problem

I was testing latest develop by creating a collection with roughly 30000 documents. Running some queries unrelated to the documents stored, I found query performance to break down on many expressions rooted in a `fn:collection` call. Sometimes performance is good (< 100ms), but if you repeat the same expression, it will break down completely (> 3s). This renders the version largely unusable for mid-sized collections.

YourKit profiler shows that nearly 80% of the query time are spent in `WeakLazyStripes.drainClearedReferences`:

![image](https://user-images.githubusercontent.com/607433/57451594-0b181580-7262-11e9-846f-f26de0d0f4eb.png)

### Describe how to reproduce or add a test
Running a `count(collection("/db"))` in the Java admin client should be sufficient to see the effect, but you need a sufficiently large number of documents stored in a subcollection. The data size does not seem to matter, just the number of docs.

My original query was:

```xquery
xquery version "3.1";

declare namespace tei="http://www.tei-c.org/ns/1.0";
declare namespace expath="http://expath.org/ns/pkg";

collection(repo:get-root())//expath:package[@name = "http://exist-db.org/apps/eXide"]
```

### Context information
Please always add the following information
- eXist version 5.0.0-SNAPSHOT (8b4e94b75)
- Java version "1.8.0_162"
- MacOS 10.14.3

wolfgangmm:
Note: commenting out the call to `drainClearedReferences` results in performance to improve to 100 to 200ms for each execution.

joewiz:
I can also confirm that commenting out https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/util/WeakLazyStripes.java#L110-L113 speeds up queries on collections with 20,000 documents that seemed to drag.

adamretter:
For reference, running this simple query a couple times is enough to reproduce the issue for me:

```xquery
for $i in (1 to 30000) return
	doc("/db/doc" || $i || ".xml")
```

PR #2830, opened 2019-06-19
Improve the performance of WeakLazyStripes

adamretter:
Closes https://github.com/eXist-db/exist/issues/2686

### DIFF
--- a/exist-core/src/main/java/org/exist/util/Holder.java
+++ b/exist-core/src/main/java/org/exist/util/Holder.java
@@ -1,0 +1,37 @@
+/*
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2019 The eXist Project
+ * http://exist-db.org
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


exist-core/src/main/java/org/exist/util/Holder.java
```diff
@@ -0,0 +1,37 @@
...
+package org.exist.util;
+
+/**
+ * Simple strong reference holder.
+ *
+ * @param <T> the type held.
+ *
+ * @author Adam Retter <adam@evolvedbinary.com>
+ */
+public final class Holder<T> {
```
shabanovd: btw, `AtomicReference` can be used instead

adamretter: Indeed. But as this is single-threaded it is cheaper not to involve memory fencing or atomics

shabanovd: it's exactly same in terms of memory usage & `set`/`get` methods calls

PR #2836, opened 2019-06-24
Disable bad optimization

### LINKED ISSUE #2786: XPath general comparison optimisation bug
### What is the problem
xpath bug.  Result sets are incorrect.

first xpath is returning more than one item - ("topicValue","topicValue")
second xpath is not returning an empty set - ("topicValue")

### What did you expect
first xpath should have only returned 1 value - ("topicValue")
second xpath should have returned an empty set - ()

```
let $dita := 
    <dita>
        <topic>
            <prolog>
                <metadata>
                    <data name="topicType" value="topicValue"/>
                </metadata>
            </prolog>
            <topic>
                <prolog>
                    <metadata>
                        <data name="topicType" value="topicValue"/>
                    </metadata>
                </prolog>
            </topic>
        </topic>
    </dita>
let $xpath1 := $dita/topic/prolog//data[@name eq 'topicType']/@value
let $xpath2 := $dita/topic/topic/topic/prolog//data[@name eq 'topicType']/@value
```

### Context information
- eXist-db versions: 5.0.0-RC8-SNAPSHOT, 4.7.0, 4.6.1, 4.3.1, 3.4.1, 2.2
- Java version: jdk 1.8.0_144
- Operating system: Win 8.1, Win 10
- 64 bit
- install method: JAR installer
- custom changes: none

wolfgangmm:
Removes a buggy query rewrite, which can't work the way it was. This affects expressions like /a/b//c[d = 123]. The main optimization on c[d = 123] will still work as shown by the tests, but /a/b will be fully evaluated.

If users want a similar optimisation, where it is appropriate they can rewrite their query manually.

Closes #2786

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/Optimizer.java
+++ b/exist-core/src/main/java/org/exist/xquery/Optimizer.java
@@ -126,44 +126,7 @@ public class Optimizer extends DefaultExpressionVisitor {
                 
                 // Replace the old expression with the pragma
                 path.replace(locationStep, extension);
-                
-                // Check if there are additional steps before the optimizable expression and
-                // rewrite them to use the ancestor axis. This will change //a/b//c[d = "D"] into
-                // //c[d = "D"][ancestor::b/parent::a]
-                int reverseAxis = reverseAxis(locationStep.getAxis());
-                Expression previous = path.getPrevious(extension);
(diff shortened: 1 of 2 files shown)


dizzzz:
rerunning builds

PR #2842, opened 2019-06-24
Adds VirtualTempPath on set content methods

reinhapa:
### Description:
Adds `VirtualTempPath` usage to also update methods within `RemoteXMLResource` and others like `RemoteResourceSet` in order to use less file file I/O for small data values. This will increase the the performance mainly on Windows systems and generally reduce the disk I/O.

### Reference:
The implementation is a follow up on the already implemented issue #2592


reinhapa:
@duncdrum What's the reason for closing my PR?

duncdrum:
I m so sorry i must have mis-clicked did not intend to close

reinhapa:
@duncdrum never mind, I've got also a nervous fingers sometimes :-)

PR #2849, opened 2019-06-25
(4.x.x) Remove openjdk 9 and 10 from build

dizzzz:
Openjdk 9 and 10 are EOL, we will not make fixes specific for these versions.

Proposal: get rid of them

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,7 @@ dist: xenial
 matrix:
   include:
     - jdk: openjdk8
-    - jdk: openjdk9
-    - jdk: openjdk10
-    - jdk: oraclejdk11
+    - jdk: openjdk11
 
     - os: osx
       osx_image: xcode9.3


dizzzz:
J12... well it all depends on if we plan to make any code changes for this coming 1-2 years ... at it is a release to stay shortly?

duncdrum:
yes its not LTS, but I think we should be looking ahead to see where errors will occur down the line. Also sooner or later it will be the default Java on some distros (just like J10 did on some ubuntus), if exist breaks on those i d like to know about it.

adamretter:
@duncdrum @dizzzz at present eXist-db does not compile on JDK 12, so that would have to be addressed first/separately.

duncdrum:
@adamretter @dizzzz I think that we should stick with J9 & J10 until we are ready to replace them with J12 on Travis.  We could do that now and set the J12 jobs to be allowed to fail, or do the switch when we are ready to address J12 build problems.  If the more people want to get rid of J9 & J10, i ll go along, i don't feel strongly about this.

dizzzz:
imo there is nothing wrong with saying 'no' to non LTS versions, as they are floating targets. The update frequency is (IMO) to high for serious business. 

But well if others say openjdk9/10 should be tested, i will accept :-) but I really dislike the overhead in travis.

PR #2857, opened 2019-06-30
feat (xquery): implement fn:parse-ietf-date()

### LINKED ISSUE #2820: implement fn:parse-ietf-date
### What is the problem

`fn:parse-ietf-date($value as xs:string?) as xs:dateTime?` is not defined in xpath-functions module.
This function is a helper to deal with date time strings received in HTTP headers like `Date`, `Last-Modified` and `If-modified-since' etc.

I was able to implement this partly in xquery. The code is in this [gist](https://gist.github.com/line-o/13c8224f7fa8938f80b6a640da134eeb)
Note that `imfd:to-dateTime` is limited to handle strings in the fix-date subset of  [Internet Message Format](https://tools.ietf.org/html/rfc5322).

### What did you expect

The function to be implemented.

BaseX has implemented it:

`XQUERY fn:parse-ietf-date('Tue, 18 Jun 2019 13:10:56 GMT')`
returns `2019-06-18T13:10:56Z`

spec: https://www.w3.org/XML/Group/qtspecs/specifications/xpath-functions-31/html/Overview.html#func-parse-ietf-date

Eval `fn:parse-ietf-date('Tue, 18 Jun 2019 13:10:56 GMT')` in eXide

### Context information
- eXist-db 5.0.0-RC8 - commit 0b53dd1ba552cc329e0737dd6869ceb348ac1a4b 
- Java(TM) SE Runtime Environment (build 1.8.0_152-b16)
- Operating system MacOs
- 64 bit
- clone from GitHub
-  no custom changes in e.g. conf.xml

line-o:
# Description:

Fully [spec](https://www.w3.org/XML/Group/qtspecs/specifications/xpath-functions-31/html/Overview.html#func-parse-ietf-date) compliant implementation of `fn:parse-ietf-date()` in java.

- add error code FORG0010
- add function to fn namespace

### Reference:

closes #2820

### Type of tests:

Xquery testsuite

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunParseIetfDate.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunParseIetfDate.java
@@ -1,0 +1,395 @@
+package org.exist.xquery.functions.fn;
+
+import org.exist.dom.QName;
+import org.exist.xquery.*;
+import org.exist.xquery.value.*;
+
+import javax.xml.datatype.DatatypeConstants;
+import javax.xml.datatype.XMLGregorianCalendar;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+
+import java.util.Map;
+import java.util.HashMap;
+import java.util.Arrays;
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Parses a string containing the date and time in IETF format,
+ * returning the corresponding xs:dateTime value.
+ *
+ * @author Juri Leino (juri@existsolutions.com)
+ */
+public class FunParseIetfDate extends BasicFunction {
+
+    private static FunctionParameterSequenceType IETF_DATE =
+            new FunctionParameterSequenceType(
+                    "value", Type.STRING, Cardinality.ZERO_OR_ONE, "The IETF-dateTime string");
+
+    private static FunctionReturnSequenceType RETURN =
+            new FunctionReturnSequenceType(
+                    Type.DATE_TIME, Cardinality.ZERO_OR_ONE, "The parsed date");
+
+
+    public final static FunctionSignature FNS_PARSE_IETF_DATE = new FunctionSignature(
+            new QName("parse-ietf-date", Function.BUILTIN_FUNCTION_NS),
+            "Parses a string containing the date and time in IETF format,\n" +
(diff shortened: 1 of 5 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/FunParseIetfDate.java
```diff
@@ -0,0 +1,427 @@
...
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ *  $Id$
```
dizzzz: can be left out

dizzzz:
yes, this is a nice and robust solution.

duncdrum:
@line-o can you remove the function from the list of unsupported features in the documenation pls. [here](https://github.com/eXist-db/documentation/blob/1b81fb207669d4b89e95f94e4ce479d77267908e/src/main/xar-resources/data/xquery/xquery.xml#L146)

line-o:
@duncdrum It is a little unclear where exactly I should remove that function from the list of unsupported ones. (master, 1b81fb, 5.0.0-RC1, ...)

line-o:
That last issue Codacy is complaining about, complexity in tzoffset(), is a little picky in my opinion.
Any proposals how to make that service happy?

line-o:
Also, could someone re-run the travis tests? This time Java8 and Java9 builds failed for no apparent reason.

line-o:
*sigh*, I was able to simplify `tzoffset()` but IDEA warned me to make the FunctionSignature package-private which Codacy is now complaining about. Any advice which one of them is right?

line-o:
Also, I ran some speed tests and it turned out that `ZonedDateTime.parse()` was actually a tad slower than the custom parser (~1 microsecond).
I removed it therefore, which further simplifies the implementation.

PR #2876, opened 2019-07-13
Fix a number of security issues around user groups

adamretter:
Previously there were several problems whereby accounts could end up without a group. Various parts of the system assumed that a group always had at least 1 group (i.e. a primary group). This has now been fixed, and accounts will always have at least 1 group, even if that is `nogroup`.

These previous problems could lead to NPEs and system instability.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/securitymanager/GroupManagementFunction.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/securitymanager/GroupManagementFunction.java
@@ -76,7 +76,7 @@ public class GroupManagementFunction extends BasicFunction {
 
     public final static FunctionSignature FNS_REMOVE_GROUP = new FunctionSignature(
         qnRemoveGroup,
-        "Remove a User Group. Any resources owned by the group will be moved to the 'guest' group.",
+        "Remove a User Group.",
         new SequenceType[]{
             new FunctionParameterSequenceType("group-name", Type.STRING, Cardinality.EXACTLY_ONE, "The group-id to delete")
         },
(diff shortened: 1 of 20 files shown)


exist-core/src/main/java/org/exist/xquery/functions/securitymanager/GroupManagementFunction.java
```diff
@@ -76,7 +76,7 @@
 
     public final static FunctionSignature FNS_REMOVE_GROUP = new FunctionSignature(
         qnRemoveGroup,
-        "Remove a User Group. Any resources owned by the group will be moved to the 'guest' group.",
```
dizzzz: so what happens after PR?

adamretter:
@duncrum nope. This stuff is very internal.

@dizzzz this should be very backwards compatible, as that was my starting point

duncdrum:
and we d like our users to be able to get to know exist's internals, via the docs?

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/GroupManagementFunction.java
adamretter: Same as before.

adamretter:
@duncdrum docs as requested - https://github.com/eXist-db/documentation/pull/362

joewiz:
From July 15, 2019 community call: 

> AR - I don’t plan to backport this to 4.x.x - others are free to do so if they wish

joewiz:
@dizzzz All tests are passing. Do you approve?

PR #2879, opened 2019-07-15
WIP: support lucene suggestions

wolfgangmm:
Lucene provides a package to generate suggestion lists based on an index, which can be presented to users to help formulating a query. This PR integrates the feature into eXist's lucene-based indexing. It also refactors the index package to get a cleaner separation between the lucene full-text index and the range index.

While eXist always had a feature to read index terms via XQuery, this method often fails if the text was passed through an analyzer, e.g. to apply language-dependent stemming to the tokenized text. Lucene's own suggestion package does address this problem, taking analyzers into account properly.

**Work in progress,** do not merge.

See tests in `extensions/indexes/lucene/src/test/xquery/lucene/suggestions.xql`

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Suggest.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Suggest.java
@@ -1,0 +1,76 @@
+/*
+ *  eXist Open Source Native XML Database
+ *  Copyright (C) 2001-2019 The eXist Project
+ *  http://exist-db.org
+ *
+ *  This program is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public License
+ *  as published by the Free Software Foundation; either version 2
+ *  of the License, or (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.modules.lucene;
+
+import org.apache.lucene.search.suggest.Lookup;
+import org.exist.dom.QName;
+import org.exist.indexing.lucene.LuceneIndex;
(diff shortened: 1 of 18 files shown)


extensions/indexes/lucene/src/test/xquery/lucene/suggestions.xql
```diff
@@ -0,0 +1,93 @@
...
+            <lucene>
+                <analyzer class="org.exist.indexing.lucene.analyzers.NoDiacriticsStandardAnalyzer" id="nodiacritics"/>
+                <analyzer class="org.apache.lucene.analysis.core.KeywordAnalyzer" id="keyword"/>
+                <analyzer class="org.apache.lucene.analysis.de.GermanAnalyzer" id="german"/>
+                <analyzer class="org.apache.lucene.analysis.en.EnglishAnalyzer" id="english"/>
+                <analyzer class="org.apache.lucene.analysis.standard.StandardAnalyzer" id="standard"/>
+                <text qname="document">
+                    <suggest id="document-content"/>
+                    <field name="content" analyzer="standard">
+                        <suggest id="document-text" type="freetext"/>
```
joewiz: Could you add a test corresponding to this `freetext` suggestion type?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Suggest.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+public class Suggest extends BasicFunction {
+
+    public final static FunctionSignature[] signatures = {
+            new FunctionSignature(
+                    new QName("suggest", LuceneModule.NAMESPACE_URI, LuceneModule.PREFIX),
+                    "Return a map of facet labels and counts for the result of a Lucene query.",
```
joewiz: Wrong function description ;)

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Suggest.java
```diff
@@ -0,0 +1,75 @@
...
+                    new SequenceType[]{
+                            new FunctionParameterSequenceType("field", Type.STRING, Cardinality.EXACTLY_ONE,
+                                    "Name of the field"),
+                            new FunctionParameterSequenceType("query", Type.STRING, Cardinality.EXACTLY_ONE,
+                                    "The string to pass to the suggester")
+                    },
+                    new FunctionReturnSequenceType(Type.STRING, Cardinality.ZERO_OR_MORE,
+                            "A map with each suggestion as key and the score as value")
```
joewiz: Wrong return description ;)

line-o:
Should also support distinct indexing and query analysers. Could be implemented as in #2991 (#3227).

PR #2881, opened 2019-07-15
Implements isOpen() / close() methods on RemoteCollection

reinhapa:
### Description:
Implements the `Collection.isOpen()` method behaviour and also moved duplicated `XmlRpcClient` acces into a central method on the `RemoteCollection` eliminating repetitive implementations all over multiple places reducing the total amount of code.

### Type of tests:
Adds a new JUnit tests for the new implemented parts as the enhanced parts from `Leasable`

### DIFF
--- a/exist-core/src/main/java/org/exist/util/Leasable.java
+++ b/exist-core/src/main/java/org/exist/util/Leasable.java
@@ -24,6 +24,7 @@ import net.jcip.annotations.GuardedBy;
 import net.jcip.annotations.ThreadSafe;
 
 import javax.annotation.Nullable;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
 
@@ -135,10 +136,20 @@ public class Leasable<T> {
         }
     }
 
+    public boolean isClosed() {
+        lock.readLock().lock();
+        try {
+            return Leasable.this.closed;
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
+
     /**
      * Represents a lease of the leasable object.
      */
     public class Lease implements AutoCloseable {
+        private final AtomicBoolean leaseClosed = new AtomicBoolean();
 
         /**
          * Access to the leased object.
@@ -146,6 +157,9 @@ public class Leasable<T> {
          * @return the leased object.
          */
         public T get() {
+            if (leaseClosed.get()) {
+                throw new IllegalLeasableState("Lease is closed");
+            }
             lock.readLock().lock();
             try {
                 if (closed) {
@@ -167,25 +181,31 @@ public class Leasable<T> {
          */
         @Override
         public void close() {
-            lock.writeLock().lock();
-            try {
-                if (closed) {
-                    throw new IllegalLeasableState("Object is closed");
-                }
+            if (leaseClosed.compareAndSet(false, true)) {
+                lock.writeLock().lock();
+                try {
+                    if (Leasable.this.closed) {
+                        throw new IllegalLeasableState("Object is closed");
+                    }
 
-                if(leases == 0) {
-                    throw new IllegalLeasableState("Lease was already returned");
-                }
+                    if(leases == 0) {
+                        throw new IllegalLeasableState("Lease was already returned");
+                    }
 
-                leases--;
+                    leases--;
 
-                if (leases == 0 && closer != null) {
-                    closer.accept(object);
-                    closed = true;
+                    if (leases == 0 && closer != null) {
+                        closer.accept(object);
+                        Leasable.this.closed = true;
+                    }
+                } finally {
+                    lock.writeLock().unlock();
                 }
-            } finally {
-                lock.writeLock().unlock();
             }
+        }
+
+        public boolean isClosed() {
+            return leaseClosed.get();
         }
     }
 
--- a/exist-core/src/main/java/org/exist/xmldb/AbstractRemoteResource.java
+++ b/exist-core/src/main/java/org/exist/xmldb/AbstractRemoteResource.java
@@ -27,7 +27,6 @@ import java.util.zip.DataFormatException;
 import java.util.zip.Inflater;
 
 import com.evolvedbinary.j8fu.lazy.LazyVal;
-import org.apache.xmlrpc.XmlRpcException;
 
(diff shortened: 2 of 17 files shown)


dizzzz:
@duncdrum there is a docker issue (java11): any idea?

image

exist-core/src/main/java/org/exist/xmldb/AbstractRemoteResource.java
```diff
@@ -156,11 +154,7 @@ public void setLastModificationTime(final Date dateModified) throws XMLDBExcepti
             params.add(path.toString());
             params.add(dateModified.getTime());
 
-            try {
-                xmlRpcClientLease.get().execute("setLastModified", params);
-            } catch (final XmlRpcException e) {
-                throw new XMLDBException(ErrorCodes.UNKNOWN_ERROR, e.getMessage(), e);
-            }
+            collection.execute("setLastModified", params);
```
dizzzz: should we make this string a (shared) constant?

exist-core/src/main/java/org/exist/xmldb/AbstractRemoteResource.java
```diff
@@ -487,20 +479,16 @@ protected long getStreamLengthInternal(final Object obj)
...
-            try {
-                final Map<?, ?> table = (Map<?, ?>) xmlRpcClientLease.get().execute("describeResource", params);
-                if (table.containsKey("content-length-64bit")) {
-                    final Object o = table.get("content-length-64bit");
-                    if (o instanceof Long) {
-                        retval = ((Long) o);
-                    } else {
-                        retval = Long.parseLong((String) o);
-                    }
+            final Map<?, ?> table = (Map<?, ?>) collection.execute("describeResource", params);
```
dizzzz: shared constant?

exist-core/src/main/java/org/exist/xmldb/AbstractRemoteResource.java
dizzzz: nope, there are too many of them?

dizzzz:
@adamretter do you want to have this in 5.x as well?

exist-core/src/main/java/org/exist/xmldb/AbstractRemoteResource.java
reinhapa: @dizzzz you mean `"setLastModified"`? Where should that be put to?

dizzzz: yes. but forget about my remark as there are soo many, and... they are only used once right?

reinhapa: yes

dizzzz:
please could you rebase ? the docker issues are biting the PR :-(

exist-core/src/main/java/org/exist/util/Leasable.java
```diff
@@ -132,17 +133,30 @@ public boolean isLeased() {
         }
     }
 
+    public boolean isClosed() {
+        lock.readLock().lock();
+        try {
+            return Leasable.this.closed;
```
adamretter: I think just `return closed;` is fine.

exist-core/src/main/java/org/exist/xmldb/RemoteCollection.java
```diff
@@ -110,37 +110,47 @@ public static RemoteCollection instance(final Leasable<XmlRpcClient> leasableXml
...
+        try {
+            return xmlRpcClientLease.get().execute(pMethodName, pParams);
+        } catch (final XmlRpcException xre) {
+            throw new XMLDBException(ErrorCodes.VENDOR_ERROR, xre.getMessage(), xre);
+        }
+    }
+
     @Override
     public void close() {
+        if (xmlRpcClientLease.isClosed()) {
```
adamretter: I don't think this is required as calling closed is `idempotent`

adamretter:
I think it would be useful to have a call to discuss the design of this ;-)

exist-core/src/main/java/org/exist/xmldb/RemoteCollection.java
reinhapa: No, because subsequent call on xmlRpcClientLease.close() will throw an error, that I don't want to have here. The `close()` method should gracefully ignore subsequent calls here.

reinhapa:
@adamretter we could have a call to discuss this change any time you like...

adamretter:
@reinhapa sorry for the delay. The scale of this change has put me off a little. I will try and find some time for it ASAP

reinhapa:
@adamretter I will check if I can find all the tests that will need to close the collections. Also I found only the following classes where I added the `RemoteCollection` for the simplification of the remote call so far (maybe I missed an other one):
`RemoteDatabaseInstanceManager`
`RemoteRestoreService`
for those I introduced a `RemoteCallSite` functional interface that specifies the needed method signature and does so hides away the actual needed RPC client implementation. There is only one spot for changing the password, where I would suggest to introduce a "virtual" method name for changing the password in order to remove the need for passing in the client reference, but I think I will do that in a next PR.

duncdrum:
i have the feeling that we'll only know the true extend on missing `close` calls, if we move this one ahead, to give the wider community a chance to locate problem areas.

reinhapa:
I will rebase the source in order to tidy up all changes into one commit

exist-core/src/main/java/org/exist/xmldb/RemoteXMLResource.java
```diff
@@ -437,11 +429,7 @@ public void setDocType(final DocumentType doctype) throws XMLDBException {
             params.add(doctype.getPublicId() == null ? "" : doctype.getPublicId());
             params.add(doctype.getSystemId() == null ? "" : doctype.getSystemId());
 
-            try {
-                xmlRpcClientLease.get().execute("setDocType", params);
-            } catch (final XmlRpcException e) {
-                throw new XMLDBException(ErrorCodes.UNKNOWN_ERROR, e.getMessage(), e);
-            }
+            collection.execute("setDocType", params);
```
dizzzz: maybe we can share `"setDocType"` as magic string in collection?

dizzzz:
@adamretter please could you mark your concerns on 'resolved' if you agree with the answers?

adamretter:
@reinhapa thanks for your contribution and sorry for the delays.

Issue #2887, opened 2019-07-19
Remove draft notation for map constructor entries


joewiz:
### What is the problem

The draft syntax for map entries is still supported in eXist, which allows users to write bad code. 

A major version such as 5.0 would be a good opportunity to remove this syntax—just as we removed other draft-era functions like `map:new`.

### What did you expect

Map constructor entry syntax is as follows: https://www.w3.org/TR/xquery-31/#prod-xquery31-MapConstructorEntry

However, according to http://exist-db.org/exist/apps/doc/xquery#legacy-features:

> Map notation: During the drafting of XQuery 3.0, the specification changed slightly and so did the implementation in eXist-db. To keep backwards compatibility, eXist allows the older notation in map constructors: `key:=value` in addition to the new one, `key: value`.

### Describe how to reproduce or add a test

Code such as `map { "name" := "joe" }` should raise an error.

### Context information

- eXist-db version: current 4.x and 5.x 
- Java version: n/a
- Operating system: n/a
- 32 or 64 bit: n/a
- How is eXist-db installed? n/a
- Any custom changes in e.g. conf.xml: n/a

adamretter:
@joewiz Seems like a good idea to me

joewiz:
As @duncdrum mentioned during the Community Call, this PR would require a new release of shared-resources, since the HTML templating module contains one use of the old syntax: https://github.com/eXist-db/shared-resources/blob/master/content/templates.xql#L94-L97. 

Also, many HTML templating apps' `view.xql` files would need to be updated, e.g., https://gitlab.existsolutions.com/tei-publisher/tei-publisher-app/blob/master/modules/view.xql#L26-29.

duncdrum:
exide's app-generator uses the proper syntax since 2016, generator apps always used the correct syntax in `view.xql`. 
I think this counts as sufficient time, for users to escape the tyranny of the defaults.

line-o:
@joewiz oh wow, this beauty is hiding in a lot more places than I thought.

joewiz:
@line-o Sure is! The most common culprit is view.xql of HTML Templating apps. I wonder if @craigberry could figure out a way to extend his code audit script to find these. I used the regex `map\s*\{[^\}]+?:=` over all of the XQuery files stored on my system. There were some false positives, but on the whole, it helped.

joewiz:
[...]

PR #2890, opened 2019-07-22
Remove XQuery draft Map constructor

line-o:
### Description:

Fixes: #2887 

Constructing a map with `map { 'a': 1 }` works.
Using the draft syntax with `map { 'a':= 1}` will fail with
```
error found while executing expression: org.exist.xquery.XPathException: err:XPST0003 unexpected token: null [at line 1, column 10]
```
### Reference:

https://www.w3.org/TR/xquery-31/#prod-xquery31-MapConstructor

### Type of tests:

Added test to `exist-core/src/test/xquery/maps/maps.xql`

### DIFF
--- a/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
+++ b/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
@@ -1338,7 +1338,13 @@ mapConstructor throws XPathException
 
 mapAssignment throws XPathException
 	:
-	exprSingle COLON^ ( EQ! )? exprSingle
+    (exprSingle COLON! EQ!) => exprSingle COLON^ eq:EQ^ exprSingle
+    {
+        throw new XPathException(#eq.getLine(), #eq.getColumn(), ErrorCodes.XPST0003,
+               "The ':=' notation is no longer accepted in map expressions: use ':' instead.");
+    }
+    |
+	exprSingle COLON^ exprSingle
 	;
 
 arrayConstructor throws XPathException
--- a/exist-core/src/test/xquery/maps/maps.xql
+++ b/exist-core/src/test/xquery/maps/maps.xql
@@ -2,10 +2,9 @@ xquery version "3.0";
 
 module namespace mt="http://exist-db.org/xquery/test/maps";
 
-import module namespace map="http://www.w3.org/2005/xpath-functions/map";
 import module namespace test="http://exist-db.org/xquery/xqsuite" at "resource:org/exist/xquery/lib/xqsuite/xqsuite.xql";
 
-declare variable $mt:daysOfWeek :=
+declare %private function mt:getMapFixture () {
     map {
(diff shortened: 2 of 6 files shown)


line-o:
`map { 'a': 1 }` works `map { 'a':= 1}` will fail with
```
error found while executing expression: org.exist.xquery.XPathException: err:XPST0003 unexpected token: = [at line x, column y]
```

joewiz:
@line-o Sounds like you found the right hook! 

For comparison, here's Saxon's error:

> XPath failed due to: The ':=' notation is no longer accepted in map expressions: use ':' instead

BaseX's error:

> [XPST0003] Invalid value, simple expression expected.

line-o:
@joewiz Yes, ideally we would error with a meaningful message. But do you know where to do that?

joewiz:
@line-o In the same file, if you search for `exception catch [RecognitionException e]`, you'll see how you can create custom error responses. Something like this might do the trick:

```java
    exception catch [RecognitionException e]
        {
        	if (e.getMessage().contains("unexpected token: =")) {
	            throw new XPathException(e.getLine(), e.getColumn(), ErrorCodes.XPST0003, 
                       "The ':=' notation is no longer accepted in map expressions: use ':' instead.");
	        } else {
	        	throw e;
	        }
        }
	;
```

duncdrum:
I m a bit surprised that our own tests still use the wrong syntax, as i seem to recall that we already did an audit, but apparently we didn't catch them all.

line-o:
@joewiz with a little help from @wolfgangmm I got the error message (by adding an alternative syntax). The RecognitionException was not thrown in this case.

line-o:
@duncdrum really ^^ which ones? I haven't checked. Is that why the build failed?

duncdrum:
yup check the travis logs

line-o:
*sigh* then I'll fix them as well.

line-o:
ready to be merged

line-o:
Travis reports
```
[ERROR] Errors: 
[ERROR]   MapTests.initializationError » XPath err:XQST0045 Error while loading module /...
```
This does not happen when I run `mvn clean package` locally. Any pointers❓

line-o:
likely because of this construct
``` 
    util:eval('map { 1 := 1 }')
```
to test for a static error.

line-o:
never mind - found the issue

line-o:
@joewiz there will be test cases failing, I am currently fixing them

line-o:
Test failing due to `:=` in `extensions/modules/cache/xquery/modules/cache/cache.xqm`.
Where is the repo for that one?

joewiz:
@line-o https://github.com/eXist-db/exist/blob/develop/extensions/modules/cache/src/test/xquery/modules/cache/cache.xqm

line-o:
@joewiz Ah, I got confused as the file gets copied next to the other test files at some point.

Issue #2895, opened 2019-07-25
(4.x.x) Unattended installation fails on Windows only


ohumbel:
### What is the problem
Starting an unattended installation on Windows by specifying an `autoinstaller.xml` file leads to the following error:
```
[ Starting processing ]
Starting process Setting data dir... (1/2)
Error: main class net.sf.saxon.Transform not found
```
Please note that this error only occurs on the `develop-4.x.x` branch (and only on Windows).

### What did you expect
I expected the unattended installation to run without errors, as it does on Linux and Mac

### Describe how to reproduce or add a test
Start the unattended installation under Windows with the 4.8.0-SNAPSHOT installer, as described in detail in the [pull request 2896](https://github.com/eXist-db/exist/pull/2869)

### Context information
- eXist-db version: eXist-db 4.8.0-SNAPSHOT  /  commit bddb39173b20c55446b565b7550e241764a93f56
- only occurs on the `develop-4.x.x` branch
- Java version Java8u121 (but does not matter)
- Operating system: Windows *only*
- 64 bit (but does not matter)
- How is eXist-db installed? Unattended, see [pull request 2896](https://github.com/eXist-db/exist/pull/2869)
- No custom changes

joewiz:
Fixed by #2896.

PR #2896, opened 2019-07-25
(4.x.x) Use the correct name for the endorsed Saxon-HE.jar on Windows, too.

ohumbel:
### Description:
This PR fixes a typo in the `Saxon-HE.jar` name for `<os family="windows"/>` in  `installer/jobs.xml`

### Reference:
Fixes [Issue #2895](https://github.com/eXist-db/exist/issues/2895)

### Type of tests:
Now the unattended installation under Windows (using an `autoinstaller.xml` file) passes.
Verified on our site on different windows machines.

### DIFF
--- a/installer/jobs.xml
+++ b/installer/jobs.xml
@@ -5,7 +5,7 @@
         <os family="windows"/>
         <executefile name="java">
             <arg>-cp</arg>
-            <arg>$INSTALL_PATH\lib\endorsed\Saxon-HE-Saxon-HE-9.6.0-7.jar</arg>
+            <arg>$INSTALL_PATH\lib\endorsed\Saxon-HE-9.6.0-7.jar</arg>
             <arg>net.sf.saxon.Transform</arg>
             <arg>-s:$INSTALL_PATH\conf.xml</arg>
             <arg>-xsl:$INSTALL_PATH\bin\conf.xslt</arg>


ohumbel:
@duncdrum Many thanks for approving :+1:

ohumbel:
@joewiz Many thanks for approving :+1: 

We (@reinhapa and me) heavily depend on the unattended installation. So we only had to check the differences introduced by recent commits ;-)

joewiz:
While CI waits to finish... I wonder if there is an automated test we could add to the test suite that would prevent this regression from returning in the future?

ohumbel:
Regarding an automated test:

We do one real physical installation on each product build, several times a day. It takes a good minute on our internal Jenkins CI System. Would this be affordable?

In my opinion, only an integral test doing a real installation and checking the result would prevent the unattended installation from failing for whatever reason. But it takes time.

joewiz:
@ohumbel While I think it would be great to have an automated test for unattended Windows installation, that does sound somewhat resource intensive. If I'm understanding what actually caused this failure, it's that because of the name mismatch, Saxon wasn't available on the classpath during the operation in question. Would it be feasible to implement a test for this? I'd welcome comments from others in @eXist-db/core about this, as I'm not a testing expert - just asking a question we tend to ask, if we can find a way to prevent this regression from recurring in the future.

ohumbel:
@joewiz Yes I think it would be feasible to write a test which parses the `jobs.xml` file and checks if this `saxon.jar` is really present.

duncdrum:
I like the idea of parsing a simple xml file to check if saxon is working very much. I could also see a larger automated test as part of a maven integration test phase, which is still on general post maven TODO list.

PR #2908, opened 2019-07-31
Allow distinct Lucene Analyzers to be specified and then used during indexing vs. querying

sten1ee:
Allow distinct Lucene Analyzers to be specified and then used during indexing vs. querying

In our usage of eXist, we're utilising a 'State of the art' LuceneAnalyzer for Sanskrit language developed by our colleagues from [BDRC](https://www.tbrc.org/) It has several modes of operation, but the one that we use - 'lenient mode' - requires one Analyzer to be used during the indexing phase, but then a different (or a differently configured) one to be used during querying.
Currently, eXist does not provide support for this use case, but this PR is aiming to achieve this feat.

There is no eXist issue related to this PR.
Neither there are tests yet, but I will shortly come up with some.

Note: the current PR adds the described feature *only* to [named Lucene indexes](https://exist-db.org/exist/apps/doc/lucene#named-indexes), i.e. those that have a _field_ attribute in the specification of the elements to be indexed. It is not accessible to _qname_ indexes or the 'main' Lucene index - the one that is not associated with a particular field. The reason for this omission is that being an outsider to eXist, I had to do a lot of 'reverse engineering'  (i.e. debugging) to figure out how and where to implement that feature. And in our use case of eXist - we don't have any qname indexes ... If some eXist insider can easily transfer the feature to qname indexes - he is most welcome to do so.

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneConfig.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneConfig.java
@@ -136,7 +136,8 @@ public class LuceneConfig {
         return analyzers.getDefaultAnalyzer();
     }
 
-    public Analyzer getAnalyzer(NodePath nodePath) {
+    // Notice: this method is never used!
+    public Analyzer __getAnalyzer(NodePath nodePath) {
         if (nodePath.length() == 0)
             throw new RuntimeException();
         LuceneIndexConfig idxConf = paths.get(nodePath.getLastComponent());
@@ -161,12 +162,24 @@ public class LuceneConfig {
 
     public Analyzer getAnalyzer(String field) {
         LuceneIndexConfig config = namedIndexes.get(field);
-        if (config != null) {
-            String id = config.getAnalyzerId();
-            if (id != null)
-                return analyzers.getAnalyzerById(config.getAnalyzerId());
+        if (config == null)
+            return analyzers.getDefaultAnalyzer();
+
+        String id = config.getAnalyzerId();
+        if (id == null)
+            return analyzers.getDefaultAnalyzer();
+
+        final String indexSuffix = ":index";
+        if (id.endsWith(indexSuffix)) {
+            // Substitute <analyzer-id>:index with <analyzer-id>:query
+            String qid = id.substring(0, id.length() - indexSuffix.length()) + ":query";
+            Analyzer queryAnalyzer = analyzers.getAnalyzerById(qid);
+            if (queryAnalyzer != null)
+                return queryAnalyzer;
+
+            LOG.warn(String.format("Failed to substitute %s with %s analyzer", id, qid));
         }
-        return analyzers.getDefaultAnalyzer();
+        return analyzers.getAnalyzerById(config.getAnalyzerId());
     }
 
     public Analyzer getAnalyzerById(String id) {


dizzzz:
thnx for the PR! having a test would certainly help to understand (how to use) the change, and to avoid future regression. for inspiration on tests see https://github.com/eXist-db/exist/tree/develop/extensions/indexes/lucene/src/test/xquery/lucene (eg analyzers.xq)

sten1ee:
Couple of questions here:

1. I spotted an unused method in the vicinity of the code changed to implement the feature - more precisely, on LuceneConfig.java, line 140. What is the policy with dead code? Leave it 'as is' or prune it?

2. Strangely enough, running the tests locally ```./build.sh test)``` produces an avalanche of failed tests. My local eXist was cloned, updated, built (by running ```./build.sh```) and then the Server and the client start and run without any problem both through ```bin/startup.sh``` as well as through IntelliJ Idea. When running the tests, the server should not be started separately, right?
What could be causing this behaviour?

dizzzz:
@sten1ee can you post some logs? (attach file)

dizzzz:
`__getAnalyze(nodepath)` might be used by code that is not being build by default? or third party? I have no idea yet... in the develop/5xx branch I cannot see any usage either

brihaye:
Hello,

Le 31/07/2019 à 16:18, sten1ee a écrit :

> Allow distinct Lucene Analyzers to be specified and then used during 
> indexing vs. querying

I fully support this request.

As I have explained years ago, a text stream can be indexed in several 
ways, i.e. pass through more than one (many ?) analyzers.

In the same way, serialization of an indexed stream could (should ?) 
pass through other analyzers...

That's what I did years ago, maybe before I knew about eXist. My 
arabic/english analyzer, is able to query a text in arabic and... give 
the result in (very, very, very bad) english...

See : 
http://www.nongnu.org/aramorph/english/lucene.html#The+english+glosser+for+Lucene

I hope to see this issue solved... ASAP ;-)

Cheers,

p.b.

wolfgangmm:
The proposed solution is straightforward and I understand the need for using different analyzers at indexing and query time. Two notes though:

* as already suggested, there should be a test along the lines of the existing analyzer tests
* eXist 5 takes a different approach to creating fields (with per-field analyzers) and I would therefore consider the existing support for named fields in 4.x.x deprecated. The PR is still welcome, but it would be good to have a similar PR and tests for develop/5.x.x before merging it.

sten1ee:
@wolfgangmm 
regarding your comment:

> eXist 5 takes a different approach to creating fields (with per-field analyzers) and I would therefore consider the existing support for named fields in 4.x.x deprecated. The PR is still welcome, but it would be good to have a similar PR and tests for develop/5.x.x before merging it.

I have no plans to come up with a similar PR for a (supposedly quite different) develop/5.x.x branch.
Will someone else volunteer to do that?
If not -- will this PR be still merged (when I supply the required parts that are missing, e.g. tests) ?

adamretter:
> I have no plans to come up with a similar PR for a (supposedly quite different) develop/5.x.x branch.

@sten1ee It is unlikely there will be a future release of eXist-db in the 4.x.x branch. If your code went into only 4.x.x then nobody would get to use it.

We will be releasing eXist-db 5.0.0 at the end of August. As such, may I suggest that you target eXist-db 5 for new features. There is little change around Lucene and Indexes in 5.x.x.

sten1ee:
@adamretter: Thank you for clarifying this, Adam.
_We will be releasing eXist-db 5.0.0 at the end of August._
When do you expect to close this release for new feature PRs?

adamretter:
4.x.x is closed for features, only hotfixes might be ported back if absolutely nessecary.

Ideally I don't want to be taking any new features between RC8 and 5.0.0, however we could discuss your PR for inclusion in a Community Call.

wolfgangmm:
As I already wrote, I consider the mechanism this PR plugs into as deprecated in 5.x.x, which offers much broader support for defining field indexes. As is it does not make much sense to apply this PR to 5.x.x and would need an update.

sten1ee:
@wolfgangmm: I've got couple of (easy to answer, I hope) questions:

1.  How can I get acquainted to the  _much broader support for defining field indexes_ in 5.x.x ?
1. What is the Lucene version that 5.x.x utilizes?

wolfgangmm:
@sten1ee 

1. in the documentation for 5.0.0, see the section about "Facets and Fields" (http://localhost:8080/exist/apps/doc/lucene#facets-and-fields in your local install). 5.x.x unifies various features ("named indexes", "constructed indexes" and "facets") into a single mechanism and configuration. This can result in a major performance boost for many applications.
2. the lucene version has not changed. It is still 4.10.4.

wolfgangmm:
@sten1ee as an example says more than a thousand words, maybe have a look at the [app I created as a test case](https://teipublisher.com/exist/apps/vangogh/) while refactoring the lucene index. [Source code](https://gitlab.existsolutions.com/tei-publisher/vangogh) is also available, so you can look at the index definitions etc.

sten1ee:
Superseded by PR#2991

Issue #2880, opened 2019-07-15
uber.jar might be missing monex websocket stuff


duncdrum:
On the docker images the `rconsole` url consistently reports a `404`. I think that this might be related to monex websocket jars not being in the flattened `uber.jar` thats on the classpath for the `entrypoint` of the image. 

monex requires
```
jackson-annotations-2.9.8.jar
jackson-databind-2.9.8.jar
javax.websocket-api-1.1.jar
```

and i can't locate them in the uber jar. When `jar tf ./exist.uber.jar > jar-contents.txt` [jar-contents.txt](https://github.com/eXist-db/exist/files/3393603/jar-contents.txt)
@adamretter should they be included by our current architecture, do we need to adjust how / when the uber jar is generated?

adamretter:
So which Jars do we talk about?

duncdrum:
`exist/lib/exist.uber.jar`

adamretter:
@duncdrum According to the `pom.xml` for Monex, Jackson annotations should be in the Monex XAR - https://github.com/eXist-db/monex/blob/master/pom.xml#L103

PR #2928, opened 2019-08-05
Fix the WebSocket server when run from Uber Jar (e.g. Docker)

adamretter:
Closes https://github.com/eXist-db/exist/issues/2880


duncdrum:
I don't quite understand what you did, but it works :) thanks

PR #2944, opened 2019-08-10
Restore should not overwrite newer apps by default

### LINKED ISSUE #2914: Restoring a backup overwrites newer apps
When restoring a backup, if the backup contains older apps, then can easily overwrite the newer apps, or worse you end up with a mess of old apps with some orphaned newer files.

Restore should detect app versions and offer the user a choice about whether to replace the newer version with the older version, or ignore the older version.

Example:
---
**NOTE**: EXPath Apps shipped with previous version of eXist-db, are not guaranteed to be forwards compatible with RC8. As such when restoring a backup to RC8 you should at least exclude the following from your `/db/apps` folder:
1. `/db/apps/dashboard`
2. `/db/apps/doc`
3. `/db/apps/eXide`
4. `/db/apps/fundocs`
5. `/db/apps/markdown`
6. `/db/apps/monex`
7. `/db/apps/packageservice`
8. `/db/apps/shared-resources`

If in doubt, you should consider excluding `/db/apps` entrirely from your backup archive during the restore, and then subsequently installing the latest versions of the Apps you need via the Dashboard's Package Manager.

wolfgangmm:
When restoring a backup, check if apps already installed in the db are newer than those in the backup. Do not overwrite newer apps unless user asks for it. Closes #2914

Porting this back to 4.x.x would require quite some work as the differences between the versions are huge, but since apps are compatible within the 4.x.x series, users will not face the same severe issues as when updating from 4 to 5.

## Tests

JUnit test included

## Documentation

https://github.com/eXist-db/documentation/pull/372


exist-core/src/main/java/org/exist/backup/Main.java
```diff
@@ -118,6 +118,10 @@
             .description("rebuild the EXpath app repository after restore.")
             .defaultValue(false)
             .build();
+    private static final Argument<Boolean> overwriteAppsArg = optionArgument("-a", "--overwrite")
```
adamretter: Can we call this `--overwrite-apps` so that it is a bit more specific?

exist-core/src/main/java/org/exist/backup/Restore.java
```diff
@@ -56,7 +57,7 @@
     private static final byte[] ZIP_FILE_MAGIC_NUMBER = {0x50, 0x4B, 0x03, 0x04};
 
     public void restore(final DBBroker broker, @Nullable final Txn transaction, final String newAdminPass, final Path f,
-                        final RestoreListener listener) throws EXistException, IOException, SAXException, PermissionDeniedException {
+                        final RestoreListener listener, boolean overwriteApps) throws EXistException, IOException, SAXException, PermissionDeniedException {
```
adamretter: Could be `final`?

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+        final List<AppDetail> apps = getAppsFromBackup(broker, descriptors);
+        final Set<String> paths = new HashSet<>();
+        final Optional<ExistRepository> repo = broker.getBrokerPool().getExpathRepo();
+        if (repo.isPresent()) {
+            for (final AppDetail app: apps) {
+                final Packages packages = repo.get().getParentRepo().getPackages(app.name);
+                if (packages != null) {
+                    final Package latest = packages.latest();
+                    try {
+                        final Semver version = Semver.parse(latest.getVersion());
```
adamretter: This will only work for apps which have valid SemVer 2.0 version numbers. @joewiz and I just fixed this elsewhere.
I am not certain, but I would that that could be a concern here too as there are some older packages with non-Semver 2.0 compatible naming?

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+     * The method attempts to be fail safe to make sure even bad backups can be restored. Errors
+     * reading package descriptors are thus only logged and should not abort the process.
+     *
+     * @param broker the broker to use for parsing the descriptor and obtaining the app root
+     * @param descriptors a queue of backup descriptors to inspect
+     * @return list of application details
+     */
+    private static List<AppDetail> getAppsFromBackup(DBBroker broker, Deque<BackupDescriptor> descriptors) {
+        final String appRoot = getAppRoot(broker);
+        final List<AppDetail> result = new ArrayList<>(17);
```
adamretter: Why `17`? Seems unnecessary

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+     * found.
+     *
+     * The method attempts to be fail safe to make sure even bad backups can be restored. Errors
+     * reading package descriptors are thus only logged and should not abort the process.
+     *
+     * @param broker the broker to use for parsing the descriptor and obtaining the app root
+     * @param descriptors a queue of backup descriptors to inspect
+     * @return list of application details
+     */
+    private static List<AppDetail> getAppsFromBackup(DBBroker broker, Deque<BackupDescriptor> descriptors) {
```
adamretter: args could be `final`.

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+     * paths may then be ignored during a restore.
+     *
+     * The method attempts to be fail safe to make sure even bad backups can be restored. Errors
+     * reading package descriptors are thus only logged and should not abort the process.
+     *
+     * @param broker the broker used for reading the backup and retrieving the expath repo
+     * @param descriptors a queue of backup descriptors to inspect
+     * @return a set of paths for which newer versions exist in the database. may be empty.
+     */
+    public static Set<String> checkApps(DBBroker broker, Deque<BackupDescriptor> descriptors) {
```
adamretter: args could be `final`

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+                                    if (PKG_NAMESPACE.equals(uri) && "package".equals(localName)) {
+                                        final String version = attributes.getValue("version");
+                                        final String name = attributes.getValue("name");
+                                        if (StringUtils.isEmpty(version) || StringUtils.isEmpty(name)) {
+                                            LOG.warn("Invalid package descriptor for " + app.getSymbolicPath());
+                                            return;
+                                        }
+                                        try {
+                                            final AppDetail detail = new AppDetail(app.getSymbolicPath(), name, Semver.parse(version));
+                                            result.add(detail);
```
adamretter: This will only work for apps which have valid SemVer 2.0 version numbers. @joewiz and I just fixed this elsewhere.
I am not certain, but I would that that could be a concern here too as there are some older packages with non-Semver 2.0 compatible naming?

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        return result;
+    }
+
+    private static List<String> getSubcollectionNames(XMLReaderPool parserPool, BackupDescriptor apps) {
```
adamretter: args could be `final`.

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+        final List<String> collections = new ArrayList<>();
+        try {
+            apps.parse(parserPool, new DefaultHandler() {
+                @Override
+                public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
```
adamretter: args could be `final`.

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+                for (final String collection : collections) {
+                    final BackupDescriptor app = apps.getChildBackupDescriptor(collection);
+                    final InputSource is = app.getInputSource("expath-pkg.xml");
+                    if (is != null) {
+                        XMLReader reader = null;
+                        try {
+                            reader = parserPool.borrowXMLReader();
+                            reader.setContentHandler(new DefaultHandler() {
+                                @Override
+                                public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
```
adamretter: args could be `final`.

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+        return collections;
+    }
+
+    /**
+     * Get the database root path for applications, removing /db and trailing slash.
+     * @param broker the broker to get the configuration from
+     * @return the root path for applications
+     */
+    private static String getAppRoot(DBBroker broker) {
```
adamretter: args could be `final`.

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+            appRoot = appRoot.substring(XmldbURI.ROOT_COLLECTION.length() + 1);
+        }
+        if (appRoot.endsWith("/")) {
+            appRoot = appRoot.substring(0, appRoot.length() - 1);
+        }
+        return appRoot;
+    }
+
+    final static class AppDetail {
+        protected final String path;
```
adamretter: Why `protected`?

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
```diff
@@ -0,0 +1,180 @@
...
+        protected final String name;
+        protected final Semver version;
+
+        AppDetail(String path, String name, Semver version) {
```
adamretter: args could be `final`.

exist-core/src/main/java/org/exist/backup/restore/RestoreHandler.java
```diff
@@ -405,7 +407,11 @@ private void restoreSubCollectionEntry(final Attributes atts) throws SAXExceptio
         //parse the sub-collection descriptor and restore
         final BackupDescriptor subDescriptor = descriptor.getChildBackupDescriptor(name);
         if(subDescriptor != null) {
-
+            if (pathsToIgnore != null && pathsToIgnore.contains(subDescriptor.getSymbolicPath())) {
+                listener.info("Skipping app path " + subDescriptor.getSymbolicPath() + ". Newer version " +
```
adamretter: I think the message could be more generic, as this class is only concerned with skipping collections, and does not need to know about 'apps', maybe just a simple:

`listener.info("Automatically skipping path: " + subDescriptor.getSymbolicPath());`

exist-core/src/main/java/org/exist/xmldb/LocalRestoreService.java
```diff
@@ -49,11 +49,12 @@ public String getVersion() {
 
     @Override
     public void restore(final String backup, final @Nullable String newAdminPassword,
-            final RestoreServiceTaskListener restoreListener) throws XMLDBException {
+            final RestoreServiceTaskListener restoreListener, boolean overwriteApps) throws XMLDBException {
```
adamretter: arg could be `final`.

exist-core/src/main/java/org/exist/xmldb/RemoteRestoreService.java
```diff
@@ -64,7 +64,7 @@ public String getVersion() {
 
     @Override
     public void restore(final String backup, @Nullable final String newAdminPassword,
-            final RestoreServiceTaskListener restoreListener) throws XMLDBException {
+            final RestoreServiceTaskListener restoreListener, boolean overwriteApps) throws XMLDBException {
```
adamretter: arg could be `final`.

exist-core/src/main/java/org/exist/xquery/functions/system/Restore.java
```diff
@@ -52,23 +52,40 @@
...
+                    "Restore the database or a section of the database (admin user only).",
+                    new SequenceType[] {
+                            new FunctionParameterSequenceType("dir-or-file", Type.STRING, Cardinality.EXACTLY_ONE,
+                                    "This is either a backup directory with the backup descriptor (__contents__.xml) or a backup ZIP file."),
+                            new FunctionParameterSequenceType("admin-pass", Type.STRING, Cardinality.ZERO_OR_ONE,
+                                    "The password for the admin user"),
+                            new FunctionParameterSequenceType("new-admin-pass", Type.STRING, Cardinality.ZERO_OR_ONE,
+                                    "Set the admin password to this new password.")
+                    },
+                    new FunctionReturnSequenceType(Type.NODE, Cardinality.EXACTLY_ONE, "the restore results")),
```
adamretter: Could reduce the copy-paste here by extracting common params into static final variables... or use the FunctionDSL.

exist-core/src/main/java/org/exist/xquery/functions/system/Restore.java
```diff
@@ -82,6 +99,11 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
         if (args[2].hasOne())
                 {adminPassAfter = args[2].getStringValue();}
 
+        boolean overwriteApps = false;
```
adamretter: Consider: `final boolean overwriteApps = args.length == 4 && args[3].effectiveBooleanValue();`

exist-core/src/test/java/org/exist/backup/RestoreAppsTest.java
```diff
@@ -0,0 +1,211 @@
...
+        try(final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()));
+            final Txn transaction = pool.getTransactionManager().beginTransaction()) {
+            Restore restore = new Restore();
+            TestRestoreListener listener = new TestRestoreListener();
+            restore.restore(broker, transaction, null, backup, listener, false);
+
+            assertEquals(0, listener.info.size());
+        }
+    }
+
```
adamretter: I think we need two more tests where a package version is invalid Semver 2.0, first test should be where the invalid version is already present in the db, second test should be where the invalid version is in the backup.

exist-core/src/test/java/org/exist/backup/RestoreAppsTest.java
```diff
@@ -0,0 +1,211 @@
...
+    private void createAndInstallApp(String version) throws IOException, PackageException, EXistException {
+        String descriptor =
+                "<package xmlns=\"http://expath.org/ns/pkg\" name=\"http://existsolutions.com/apps/backup-test\"\n" +
+                "   abbrev=\"backup-test\" version=\"" + version + "\" spec=\"1.0\">\n" +
+                "   <title>Backup Test App</title>\n" +
+                "   <dependency processor=\"http://exist-db.org\" semver-min=\"5.0.0-RC8\"/>\n" +
+                "</package>";
+        Path xarFile = temporaryFolder.newFile().toPath();
+        ZipOutputStream zos = new ZipOutputStream(Files.newOutputStream(xarFile, StandardOpenOption.WRITE));
```
adamretter: `try-with-resources` to avoid dangling file handles on exception.

exist-core/src/test/java/org/exist/backup/RestoreAppsTest.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        @Override
+        public void restoredResource(String resource) {
+            // unused
+        }
+
+        @Override
+        public void info(String message) {
+            System.out.println(message);
```
adamretter: Should avoid `System.out` in tests. Instead can you write to log4j

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
wolfgangmm: As far as I know the version attribute in the `expath-pkg.xml` always had to be a valid semver - thus the exception being thrown by Semver. Maybe you're confusing this with dependencies and `semver`, `semver-max` etc?

For invalid semvers no version check can be made, so corresponding packages can only be ignored by this code.

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
wolfgangmm: Just to make codacity stop getting on my nerves ;-)

exist-core/src/main/java/org/exist/backup/restore/AppRestoreUtils.java
duncdrum: actually expath failed to properly implement the intended checks. So while documentation said `semver` what it actually let through was not always proper semver. I fixed that a while ago.

adamretter: Also you can still do comparisons by coercion. Which is exactly what @joewiz and I implemented for the public-repo and packageservice.

@joewiz thoughts/comments please?

wolfgangmm: for `semver` I'm aware there are no checks, but we're only dealing with `version` here. In any case it is unclear what the expected behaviour would be? We can assume that the eXist core apps which are required for proper operation have a valid semver. For user-supplied apps it is the users job to sort out the problem, which is easy if you have a working dashboard.

adamretter: I don't think we can make such assumptions - we should do the same coercion here...

adamretter: @wolfgangmm It looks like I might have already build coercion into the version you are using - https://github.com/expath/expath-pkg-java/blob/master/pkg-java/src/test/java/org/expath/pkg/repo/deps/SemverTest.java#L26

Maybe we are lucky? But we should add those two suggested tests around that

adamretter:
@wolfgangmm this needs to be extended to cover EXPath libraries as well as apps; as reported by Craig in https://github.com/eXist-db/exist/issues/2914#issuecomment-521053816

wolfgangmm:
Ok, so with the latest commits, library packages (which are not deployed into the db) are version-checked before being restored as well: if a newer version is already installed, do not restore the package into /db/system/repo, so if you run the expathrepo repair script, it will not attempt to install the older version over the newer one.

exist-core/src/main/java/org/exist/backup/restore/RestoreHandler.java
wolfgangmm: I guess for users it is important to know why the path was skipped - thus the verbose message.

Issue #2014, opened 2018-07-10
Performance of 5.0.0-RC1 - XPath expressions


joewiz:
### What is the problem

Comparing the current stable release (4.3.0) to 5.0.0-RC1, the performance of an XPath expression like `/package:expath` is measurably slower in 5.0.0-RC1—by a factor of 4-5x. This performance gap persists as the database size increases. 

I believe this issue contributes to the slower performance I'm seeing under 5.0.0-RC1 in real-world apps running on a database with a substantial number of documents and collections. Though there may be other factors too, I wanted to report this one, as I was able to create an easily reproducible test (see below). 

I'd welcome others to report their results when running the same test query below against their fully populated databases.

### What did you expect

I expected the performance of XPath expressions to be roughly the same in 5.0.0-RC1 as the current stable version (4.3.0).

### Describe how to reproduce or add a test

The following test runs an XPath expression 100 times, then returns the timing information. 

```xquery
xquery version "3.1";

declare namespace expath="http://expath.org/ns/pkg";

let $query := "/expath:package"
let $hits := count(/expath:package)
let $iterations := 100
let $tests :=
    for $x in 1 to $iterations
    let $start-time := util:system-time()
    let $eval := util:eval($query)
    return
        map { "duration": seconds-from-duration(util:system-time() - $start-time) * 1000 }
let $total-ms := sum($tests?duration)
let $average-ms := $total-ms div $iterations
return
    map {
        "query": $query,
        "hits": $hits,
        "iterations": $iterations,
        "total-ms": $total-ms, 
        "average-ms": $average-ms,
        "exist-db-version": system:get-version() || "-" || system:get-revision()
    }
```

The results in 4.3.0 vs. 5.0.0-RC1, after submitting the query 10 times first after startup to allow the database to "warm up", are as follows:

```json
[
    {
        "hits": 8,
        "query": "/expath:package",
        "total-ms": 50,
        "iterations": 100,
        "average-ms": 0.5,
        "exist-db-version": "4.3.0-3d6c60327"
    },
    {
        "hits": 8,
        "query": "/expath:package",
        "total-ms": 235,
        "iterations": 100,
        "average-ms": 2.35,
        "exist-db-version": "5.0.0-RC1-141b5c7"
    }
]
```

In other words, an XPath expression like `/expath:package` takes .5ms in 4.3.0 but 2.35ms in 5.0.0-RC1. This is with a stock system containing just the 8 built-in xars. 

[...]

PR #2962, opened 2019-08-18
Improve Lock Table lease recording performance

adamretter:
See https://github.com/eXist-db/exist/issues/2014#issuecomment-522358491


adamretter:
I have updated the docs in - https://github.com/eXist-db/documentation/pull/380

PR #2969, opened 2019-08-23
Fix an NPE under load recently introduced in WeakLazyStripes

adamretter:
Fixes a regression introduced in https://github.com/eXist-db/exist/pull/2962


duncdrum:
If this is regression with nope territory shouldn’t t we add a test?

adamretter:
@duncdrum it is quite hard to reproduce deterministically, however if you study the code closely you can see an NPE is possible.

PR #2992, opened 2019-08-29
Addition to #2367: added [@attr != value] predicate

sten1ee:
Added [@attr != value] predicate in XPath spec for Lucene index spec.
Previously only [@attr = value] was supported
Tests added too.
( see original https://github.com/eXist-db/exist/pull/2367 )

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/NodePathPattern.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/NodePathPattern.java
@@ -186,10 +186,6 @@ public class NodePathPattern {
             input = input.substring("fn:not".length()).trim();
             negate = true;
         }
-        else if (input.startsWith("not(")) {
-            input = input.substring("not".length()).trim();
-            negate = true;
-        }
 
         if (negate) {
             if (!input.startsWith("(") || !input.endsWith(")")) {
--- a/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/NodePathPattern.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/NodePathPattern.java
@@ -64,11 +64,18 @@ public class NodePathPattern {
         }
     };
 
-    static class SimpleAttrEqValuePredicate implements Predicate {
+    enum PredicateCode {
+        EQUALS,
+        NOT_EQUALS
+    }
+
+    static class SimpleAttrValuePredicate implements Predicate {
+        private final PredicateCode pcode;
         private final String attrName;
         private final String attrVal;
 
-        SimpleAttrEqValuePredicate(String attrName, String attrVal) {
+        SimpleAttrValuePredicate(PredicateCode pcode, String attrName, String attrVal) {
+            this.pcode = pcode;
             this.attrName = attrName;
             this.attrVal = attrVal;
         }
@@ -76,7 +83,15 @@ public class NodePathPattern {
         @Override
         public boolean evaluate(NodePath2 nodePath, int elementIdx) {
             String val = nodePath.attribs(elementIdx).get(attrName);
-            return Objects.equals(val, attrVal);
+            switch (pcode) {
+                case EQUALS:
+                    return Objects.equals(val, attrVal);
+                case NOT_EQUALS:
+                    return !Objects.equals(val, attrVal);
+                default:
+                    assert false: "PredicateCode " + pcode + " not handled!";
+                    return false;
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


adamretter:
Just checking, do you mean `!=` or `ne`? As they are subtly different!

sten1ee:
@adamretter elaborate please - what's the difference? (I am a XPath newbie)
None of the values in this case is a _node-set_
so the implemented semantics of [@attr != val] is effectively:
_A node will match this if and only if it has not attribute with name attr, and value val_

adamretter:
@sten1ee See slides 30, and 37 to 42: http://www.adamretter.org.uk/presentations/xquery-and-xml-databases_xml-summer-school_oxford_20180913.pdf

I think it is quite unlikely that you want `!=`. You probably mean `[fn:not(@attr = val)]` or `[@attr ne val]`, neither of which is the same as `[@attr != val]`.

sten1ee:
@adamretter thank you for clarifying this for me.
I think I can implement all the 4 options here: == and != and 'ne' and 'eq'
but let me see if I am getting this correctly:

1. Your slide 30 says that == and != are meant for sequences (i.e. when one of the operands is a sequence?) while 'ne' and 'eq' are meant for atomic values (i.e. when both operands are atomic values), right or wrong ?

1. In my case the left operand **@attr** is a sequence of length 1 or 0 (i.e. the value of that attribute when it is present, or the empty sequence when that particular attribute is not present at all), right or wrong?

1. In my case the right operand **value** is always an atomic value (and in particular - string, as I don't try to interpret other possibilities), right or wrong?

1. Hence '!=' should behave like 'ne' **except** when that particular attribute is not present - then '!=' should return _false_ (as no element from the empty sequence on the left hand side is different from the atomic value on the right hand side), whereas 'ne' should return _true_, right or wrong?

sten1ee:
Last commit adds 'ne' and 'eq' to '=' and '!='
Plus there is option for negation now:
[fn:not(@attr='value')]
(I am well aware that this joke-of-a-parsePredicate should be replaced with a proper (ANTLR?) generated parser, nut the perceived amount of work in this overwhelms me)

adamretter:
@sten1ee See below:

> 1. Your slide 30 says that == and != are meant for sequences (i.e. when one of the operands is a sequence?) while 'ne' and 'eq' are meant for atomic values (i.e. when both operands are atomic values), right or wrong ?

Yes, although with `==` and `!=`, one or *both* of the operands are allowed to be sequences.

> 2. In my case the left operand @attr is a sequence of length 1 or 0 (i.e. the value of that attribute when it is present, or the empty sequence when that particular attribute is not present at all), right or wrong?

In XQuery a sequence of 1 item is indistinguishable from an atomic value. So you can use it as either ;-)

> 3. In my case the right operand value is always an atomic value (and in particular - string, as I don't try to interpret other possibilities), right or wrong?

AFAIK from what you have said.

> 4. Hence '!=' should behave like 'ne' except when that particular attribute is not present - then '!=' should return false (as no element from the empty sequence on the left hand side is different from the atomic value on the right hand side), whereas 'ne' should return true, right or wrong?

Nope. For != you have to compare each item on the left with each item on the right, if there is one pair that does not match then you get fn:true.

sten1ee:
@adamretter:
> Nope. For `!=` you have to compare each item on the left with each item on the right, if there is one pair that does not match then you get fn:true.

Well, I thought I have stated exactly what you have stated, so to me the "Nope" is quite undeserved.
'Each item on the left' is:
- either the attribute value (if that attribute is present) and then it is a single value
- or none (aka 'empty sequence') (if that attribute is not present) and then '!=' should return true

sten1ee:
@adamretter: 

> Hence '!=' should behave like 'ne' except when that particular attribute is not present - then '!=' should return false (as no element from the empty sequence on the left hand side is different from the atomic value on the right hand side), whereas 'ne' should return true, right or wrong?

If you read carefully this (previously item number 4 from my list of 4 statements/question), I _believe_ it is correct (unless my understanding of English is incorrect).
I agree that I have screwed up part of my previous comment (the one that you have edited and :-p -ed at).

sten1ee:
More precisely, I've screwed up by writing this: _and then '!=' should return true_
Whereas I've meant: _and then '!=' should return false_
But if you please check the current implementation of those predicates in the code it says:
```java
        public boolean evaluate(NodePath2 nodePath, int elementIdx) {
            String val = nodePath.attribs(elementIdx).get(attrName);
            switch (pcode) {
                case EQUALS: // = 
                case EQ: // eq
                    return Objects.equals(val, attrVal);
                case NOT_EQUALS: // !=
                    // actual attr val should be present but different:
                    return val != null && !Objects.equals(val, attrVal);
                case NE: // ne
                    // actual attr val may be null (i.e. not present) or present but different:
                    return !Objects.equals(val, attrVal);
                default:
                    assert false;
                    throw new IllegalArgumentException("PredicateCode " + pcode + " not handled!");
            }
        }
```
so the implemented behaviour is as you've described it.
Ok?

sten1ee:
@adamretter: What is preventing this one from being accepted and merged in a subsequent 5.x.x release?

dizzzz:
only the macOS test is failing; should be fixed with a rebase.

sten1ee:
Rebased.
"All checks have passed"
Merge?

Issue #3004, opened 2019-09-05
Can't use module relying on jar code after 5.0.0 migration


gmella:
### What is the problem
The installation of some modules coming with jar/java code inside are not working. 
I notived that on one module which we developed and see after some test that it also occurs on expath `ft-client`

Error shown is `exerr:ERROR error found while loading module ft-client: Unable to instantiate module from EXPathrepository: org.expath.exist.ftclient.ExistExpathFTClientModule`
I can provide more log but reproducer is quite easy.

### What did you expect
Get common distributed modules and my mvn module (freshly compiled with exist-core 5.0.0 dependency ) back again.

### Describe how to reproduce or add a test
install step:
```
xquery version "3.1";
repo:install-and-deploy("http://expath.org/ns/ft-client", "1.1.6", "http://demo.exist-db.org/exist/apps/public-repo/modules/find.xql")
```
use step:

```
xquery version "3.1";
import module namespace ft-client="http://expath.org/ns/ft-client";
ft-client:connect(xs:anyURI('ftp://ftp.de.debian.org/debian/'))
```

### Context information

eXist Version: | 5.1.0-SNAPSHOT
eXist Build: | 20190904093321
Operating System: | Linux 4.9.0-8-amd64 amd64
Java Version: | 1.8.0_222

I also tried with Docker versions latest, 5.0.0 and 4.7.1 and a more classical one with 5.0.0 installer. Was ok only on 4.7.1 and previous.

I changed `conf.xml` but that is not relevant for this issue.
let me share how in my docker-compose
```
oival:
  image: existdb/existdb:5.0.0
  ports:
    - "8080:8080"
  volumes:
    - ./jmmc-resources/build/jmmc-resources-0.23.xar:/exist/autodeploy/jmmc-ressources.xar
    - ./oival/build/oival-2.9.xar:/exist/autodeploy/oival.xar
    # overwrite conf.xml for output-size-limit="10000000" and enable-java-binding="yes"
    - ./docker/exist/conf.xml:/exist/etc/conf.xml
  user: 0:1000
  restart: always
```

and address a very big Thank you for all of your work!

dizzzz:
I think there are parallels with https://github.com/eXist-db/exist/pull/2949

adamretter:
@gmella where can I find the source code for `ft-client`?

gmella:
@dizzz yes I saw this issue but prefer to open a new one...
[...]

PR #3013, opened 2019-09-10
Add classNotFoundException as throwable in log to get full stacktrace info

gmella:
nhance logging pattern not to swallow exception information.

This could help users to pinpoint some classloader issue as discussed with #3004

A couple of other log report still swallow such classNotFoundException. Please tell me if this is relevant to provide larger PR on this aspect.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -1569,8 +1569,8 @@ public class XQueryContext implements BinaryValueManager, Context {
             if (LOG.isDebugEnabled()) {
                 LOG.debug("module " + module.getNamespaceURI() + " loaded successfully.");
             }
-        } catch (final ClassNotFoundException e) {
-            LOG.warn("module class " + moduleClass + " not found. Skipping...");
+        } catch (final ClassNotFoundException cnfe) {
+            LOG.warn("module class " + moduleClass + " not found. Skipping...", cnfe);
         }
         return module;
     }


exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -1569,8 +1569,8 @@ Module initBuiltInModule(final String namespaceURI, final String moduleClass) {
             if (LOG.isDebugEnabled()) {
                 LOG.debug("module " + module.getNamespaceURI() + " loaded successfully.");
             }
-        } catch (final ClassNotFoundException e) {
-            LOG.warn("module class " + moduleClass + " not found. Skipping...");
+        } catch (final ClassNotFoundException cnfe) {
+            LOG.warn("module class " + moduleClass + " not found. Skipping...", cnfe);
```
adamretter: @gmella I would like to avoid putting the full stack-trace in the log, as this is at `WARN` level.
Rather, I think that `cnfe.getMessage()` might contain enough detail of the missing class name?

gmella: No experience to say if such case may be considered as an error and add such stack trace.
`grep -re "catch*ClassNotFoundException" -A3 src` shows us some other cases that do not throw any exception with non consistent rules. 

In any case showing the root Cause would be the best compromise to help understanding. This may be done using a helper function for summarizing exception causes ?
We use such [kind of code](https://github.com/JMMC-OpenDev/jMCS/blob/764395f1c0d67ea36b0b70ba68889e29456d4ec7/src/main/java/fr/jmmc/jmcs/gui/component/MessagePane.java#L153) to report feedback from our applications.

adamretter: Any idea if we need that complexity, or if `getMessage()` would suffice here?

gmella: yes adding `getMessage()` probably is a small enhancement here. But in any case such report would appear after previous warning log that could explain this classnotfoundexception. We can ignore this PR. 

Now I understand that my analysis was not clear because I focused more on the error than the warning as shown in the next log extract:

> 
> 2019-09-11 11:29:46,407 [qtp1366811528-40] WARN  (XQueryContext.java [initBuiltInModule]:1573) - module class fr.jmmc.exist.OIExplorerModule not found. Skipping... 
> 2019-09-11 11:29:46,407 [qtp1366811528-40] ERROR (XQueryServlet.java [process]:559) - Cannot compile xquery: err:XPST0081 error found while loading module jmmc-oiexplorer: Error while loading module /db/apps/jmmc-resources/content/jmmc-oiexplorer.xql: Invalid qname oi:to-xml. No namespace defined for prefix oi. QName is invalid: INVALID_PREFIX 
> ...

PR #3034, opened 2019-10-02
Allow multiple transactions per broker

adamretter:
Simple change to allow multiple transactions per-broker.


exist-core/src/main/java/org/exist/storage/txn/TransactionManager.java
```diff
@@ -208,9 +208,9 @@ private Txn doBeginTransaction() {
         transactions.put(txnId, new TxnCounter().increment());
         final Txn txn = new Txn(this, txnId);
 
-        // TODO(AR) ultimately we should be doing away with DBBroker#setCurrentTransaction
+        // TODO(AR) ultimately we should be doing away with DBBroker#addCurrentTransaction
```
dizzzz: is the comment correct?

adamretter: Yup

PR #3035, opened 2019-10-03
Avoid failing macOS builds on travis

dizzzz:
let's check


dizzzz:
the result:

image

dizzzz:
@adamretter please accept this PR as the never passing builds on mac are very unfriendly and time-wasting while checking the results.

adamretter:
As long as we still run the dependency check on Linux then I am happy.

PR #3064, opened 2019-10-22
Fix a race condition in WeakLazyStripes

adamretter:
Thanks very much to @paulmer for digging in and finding the cause:

Quoting @paulmer regarding `WeakLazyStripes`:
>The "get" method allocates a new lock if it finds that the weak reference matching the key has expired.  It then saves this new lock in the stripes hash (see getOptimistic, for example).  Then it proceeds to run drainClearedReferences().  If an expired reference with the same name as the lock just requested is on the refererenceQueue, drainClearedReferences will remove the reference to the newly created lock from the hash table.  A subsequent thread can then obtain a new lock for the same key even while the previous thread is still holding onto its lock.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,15 +41,9 @@ after_test:
 
 on_failure:
   - cmd: 7z a exist-core-surefire-reports.zip %APPVEYOR_BUILD_FOLDER%\exist-core\target\surefire-reports\ && appveyor PushArtifact exist-core-surefire-reports.zip
-  - cmd: 7z a exist-core-test-logs.zip %APPVEYOR_BUILD_FOLDER%\exist-core\target\test-logs\ && appveyor PushArtifact exist-core-test-logs.zip
+  - cmd: 7z a exist-core-test-logs.zip %APPVEYOR_BUILD_FOLDER%\exist-core\target\test-logs-* && appveyor PushArtifact exist-core-test-logs.zip
   - sh: tar czvf exist-core-surefire-reports.tgz $APPVEYOR_BUILD_FOLDER/exist-core/target/surefire-reports/ && appveyor PushArtifact exist-core-surefire-reports.tgz
-  - sh: tar czvf exist-core-test-logs.tgz $APPVEYOR_BUILD_FOLDER/exist-core/target/test-logs/ && appveyor PushArtifact exist-core-test-logs.tgz
-
-artifacts:
-  - path: exist-core\target\surefire-reports
-    name: exist-core-surefire-reports
-  - path: exist-core\target\test-logs
-    name: exist-core-test-logs
+  - sh: tar czvf exist-core-test-logs.tgz $APPVEYOR_BUILD_FOLDER/exist-core/target/test-logs-* && appveyor PushArtifact exist-core-test-logs.tgz
 
 cache:
   - '%USERPROFILE%\.m2'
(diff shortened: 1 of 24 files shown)


joewiz:
The only CI failure is AppVeyor's jdk11 build.

@adamretter Do you chalk this up to a CI glitch, or is it possibly caused by this PR?

```
[ERROR] concurrent(org.exist.xmldb.concurrent.ConcurrentAttrUpdateTest)  Time elapsed: 7.97 s  <<< ERROR!
java.util.concurrent.ExecutionException: java.lang.NullPointerException

[ERROR] transformReindexTransform(org.exist.xquery.functions.transform.TransformTest)  Time elapsed: 0.525 s  <<< ERROR!
org.exist.xquery.XPathException: exerr:ERROR Unable to set up transformer: null [at line 6, column 9]
	at org.exist.xquery.functions.transform.TransformTest.transform1(TransformTest.java:306)
	at org.exist.xquery.functions.transform.TransformTest.transformReindexTransform(TransformTest.java:267)
	at org.exist.xquery.functions.transform.TransformTest.transform1(TransformTest.java:306)
	at org.exist.xquery.functions.transform.TransformTest.transformReindexTransform(TransformTest.java:267)
```

adamretter:
@joewiz I will restart that job - it looks like two tests running in parallel both tried to write to the same log4j log file simultaneously which caused a problem.

appveyor.yml
```diff
@@ -41,15 +41,9 @@ after_test:
 
 on_failure:
   - cmd: 7z a exist-core-surefire-reports.zip %APPVEYOR_BUILD_FOLDER%\exist-core\target\surefire-reports\ && appveyor PushArtifact exist-core-surefire-reports.zip
-  - cmd: 7z a exist-core-test-logs.zip %APPVEYOR_BUILD_FOLDER%\exist-core\target\test-logs\ && appveyor PushArtifact exist-core-test-logs.zip
+  - cmd: 7z a exist-core-test-logs.zip %APPVEYOR_BUILD_FOLDER%\exist-core\target\test-logs-* && appveyor PushArtifact exist-core-test-logs.zip
```
dizzzz: originally the test-logs directory was bundled... now only files?

joewiz:
@adamretter When I checked a couple of hours ago, I saw the same AppVeyor jdk11 CI profile had failed after just a few minutes. I restarted that run, but while this one lasted longer (1h), it ended with this error:

```
ERROR] Tests run: 96, Failures: 0, Errors: 1, Skipped: 6, Time elapsed: 12.25 s <<< FAILURE! - in org.exist.xquery.XQueryTest
[ERROR] namespace(org.exist.xquery.XQueryTest)  Time elapsed: 0 s  <<< ERROR!
java.lang.NullPointerException
	at org.exist.xquery.XQueryTest.namespace(XQueryTest.java:821)
```

adamretter:
@joewiz these failures on AppVeyor are very strange. As far as I can see it should be impossible to have an NPE on line 821 of XQueryTest.

joewiz:
The good news is that the AppVeyor jdk11 build passed, but “APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2017, JAVA_HOME=C:\Program Files\Java\jdk1.8.0” timed out after 60 min. I just restarted that build. Fingers crossed.

Issue #3071, opened 2019-10-27
fn:random-number-generator not spec compliant


line-o:
### What is the problem

The function `fn:random-number-generator` is not compliant to the xquery 3.1 specification.

Specifically:
1. The seed can be of type `xs:anyAtomic` but only `xs:number` is supported because it is internally cast to `java.long` https://github.com/eXist-db/exist/blob/cdefdee3263b3e68e37abc3ad6893b30fc7682c6/exist-core/src/main/java/org/exist/xquery/functions/fn/FnRandomNumberGenerator.java#L43
2. There is a side-effect with the function returned in the `next` key.
https://github.com/eXist-db/exist/blob/cdefdee3263b3e68e37abc3ad6893b30fc7682c6/exist-core/src/main/java/org/exist/xquery/functions/fn/FnRandomNumberGenerator.java#L84

### What did you expect

The implementation to be spec compliant.

### Describe how to reproduce or add a test

You can find the tests in #3072 

### Context information

- eXist-db 5.0.x /   
- Java
- MacOs 10.14.x (64 bit)
- clone from latest develop
- no custom changes

PR #3072, opened 2019-10-27
[bugfix] spec compliant random-number-generator

line-o:
### Description:

- [x] add tests for `fn:random-number-generator`
- [x] xs:anyAtomic allowed as a seed value
- [x] calls to `fn:random-number-generator()?next()` are deterministic

### Reference:

fixes https://github.com/eXist-db/exist/issues/3071

### Type of tests:

xqsuite

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FnRandomNumberGenerator.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FnRandomNumberGenerator.java
@@ -6,6 +6,7 @@ import org.exist.xquery.*;
 import org.exist.xquery.functions.map.MapType;
 import org.exist.xquery.value.*;
 
+import java.util.Optional;
 import java.util.Random;
 
 import static org.exist.xquery.FunctionDSL.*;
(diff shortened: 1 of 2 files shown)


line-o:
Yes, this is will fail until the function is implemented according to spec.

duncdrum:
To merge this now i d suggest expecting the error and linking to #3071 in an inline comment, this way we would track changes in behaviour moving forward.  

Unless you have a fix to he underlying java implementation ready to go, so we could add it to this PR ?

line-o:
I would rather leave this open until the the fix is in place, @duncdrum .

adamretter:
@line-o happy to take a look at this with you when we meet during XML Prague 2020.

line-o:
@adamretter I just checked the spread of the random numbers (`next()` called 10000 times). The results looks OK.
The delta of each tenth is below one hundred in each segment (-100, 100).

adamretter:
@line-o you should find that they are better than java.util.Random

adamretter:
@line-o sounds good, but as you and I both worked on it - you will have to find someone else to review. The rule is - "you never merge your own PRs"

line-o:
I know :)

joewiz:
I’m curious to know, are we passing more of the qt3 tests for this function after the bug fix?

adamretter:
@joewiz after modifying `build.sbt` to reflect the correct eXist-db version, running:

```bash
$ java -jar exist-xqts-runner-assembly-1.0.0.jar --xqts-version HEAD --test-set fn-random-number-generator
```

Yields the results:
1. 45.45% for eXist-db 5.2.0
2. 56.82% for eXist-db 5.3.0-SNAPSHOT

Not great I know! But... in this PR I only focused on fixing the two bugs that @line-o reported to me.

line-o:
Well these two amounted for more than 10%, I think it was worth the effort. 
@adamretter could you publish the failing ones here?

adamretter:
@line-o No, you can run the XQTS.

I think you will particularly groan at the `fn-random-number-generator-30` failure though, as it has implications on your xBow work too!

adamretter:
See also - https://github.com/eXist-db/exist/pull/3249

Issue #2051, opened 2018-07-20
sibling axes used without element name very inefficient


merenyics:
### What is the problem
> whereas an XPath expression like
`preceding-sibling::a`
works as it should, more general expressions involving the sibling axes are _exremely_ inefficient.
`preceding-sibling::*
preceding-sibling::node()
preceding-sibling::text()
`
do seem to work, but take ages to evaluate both as steps in an XPath and as filter expressions. The same goes for following-sibling.

### What did you expect
> fast execution

### Describe how to reproduce or add a test
> On a collection containing just a few thousand elements `<a>`, run a query like
`$coll//a[following-sibling::*]`
sit back and relax (for 10-30 minutes!!!)

I would also like to add that these queries cannot be killed, or more precisely, killing them takes just as long as runnig them.

### Context information
Please always add the following information
- eXist-db version : all recent versions, I'm using 4.3 at the moment
- Java version (Java8u161)
- Operating system (Windows 10)
- 64 bit
- How is eXist-db installed? (JAR installer)

adamretter:
So my findings so far are:

1. the EQName Test query (`preceding-sibling::a`) takes a very different path to both, the Wildcard name test query (`preceding-sibling::*`), and the `AnyKindTest` query (`preceding-sibling::node()`). This disjunction happens at line 888 of https://github.com/eXist-db/exist/blob/develop/src/org/exist/xquery/LocationStep.java#L888
   1. The Wildcard and AnyKindTest queries are considered wildcard tests, these fall-back to doing a range scan over the `dom.dbx` file, which is known to be much slower than using an index.
   2. The EQName test query uses the Structural Index to lookup the siblings in the `structure.dbx` file, which is much faster than a range scan.

However, a profiler shows that the problem with the wildcard tests is not actually the `dom.dbx` range scan, but rather that almost all of the execution time is spent in `org.exist.util.FastQSort` for `NewArrayNodeSet`.

After some examination, it seems that `NewArrayNodeSet` is less than idea to use for sibling lookups (and other operations as well I suspect). The `LocationStep#PrecedingSiblingFilter` calls `NewArrayNodeSet#get` and then `NewArrayNodeSet#add` for every sibling it finds whilst scanning `dom.dbx`. This causes the `NewArrayNodeSet` to be re-sorted on every call to `NewArrayNodeSet#get`, i.e. once for every single preceding-sibling that matches the expression.

[...]

Issue #2527, opened 2019-02-28
xpath issue


eduarddrenth:
in exist 4.6 this xpath yields 4 results instead of 1:

//tei:entry[position()=1]/tei:form/tei:gram[@type='pos']

on this xml: [huis.zip](https://github.com/eXist-db/exist/files/2915540/huis.zip)

I can fix it using:

//tei:entry[position()=1]/tei:form[@type='lemma']/tei:gram[@type='pos']

joewiz:
I can confirm the report. I've reduced it to the following test case:

```xquery
<root>
    <a>
        <b>
            <c foo="bar"/>
        </b>
        <x>
            <y>
                <b>
                    <c foo="bar"/>
                </b>
            </y>
        </x>
        <x>
            <y>
                <b>
                    <c foo="bar"/>
                </b>
            </y>
            <y>
                <b>
                    <c foo="bar"/>
                </b>
            </y>
        </x>
    </a>
</root>
/a[1]/b/c[@foo="bar"]
```

Expected results: `<c foo="bar"/>`
Actual results: 4 of these.

Saxon and BaseX return the expected results. eXist 4.x.x-develop (e935f4e2) returns the erroneous results.

@eduarddrenth Thanks for reporting this, but next time, **please** complete the issue template, to help us best address your issue.

eduarddrenth:
sorry, I know, busy busy busy.... and I thought better you know it thsi way than not at all

keep up the good work

dizzzz:
@eduarddrenth thnx for the report! but.... we are all busy :-)

line-o:
#3075 fixed this

line-o:
There will likely be no back port to v4.x.x for this. I hope upgrading to v5.3.0 is an option for you @eduarddrenth

eduarddrenth:
thnks. I'm on 5.2.0 for quite some time

Issue #2129, opened 2018-08-21
wildcard following and preceding axes are slow


craigberry:
Some of this has previously been discussed in issue #2051 but here I focus specifically on the non-sibling axes.  

I compared the average of five trials of the following script in 4.3.1 and eXist-5.0.0-RC1-165-ge607599f4. The script basically starts from a known node in the middle of a largish document (getting to that point is not timed) and then times gathering the 5 preceding and 5 following nodes of the same element relative to that starting node. 4.3.1 averaged 0.439 seconds for the timed portion of the script, and the new development version 1.174 seconds, or 2.6 times worse than before. 

It seems to me that gathering 11 nodes in sequence should not take 400+ milliseconds (or more than a whole second) when you start from a known node in the middle of that sequence, and recent work on the wildcard axes has not helped.  In the example, the node I start from is node 25,000 out of 50,000.  The query runs faster if I start from an earlier node and slower if I start from a later node.  It sure sounds like something is iterating over every node in the document from the beginning, whereas it should be doing the exact same amount of work regardless of where I start, decrementing or incrementing from the location I give it as a starting point.

Here is the query.

xquery version "3.1";

let $words := for $i in (1 to 50000) return <w xml:id="{$i}">{$i}</w>

let $x := if (file:exists('/db/words.xml')) then xmldb:remove('/db/' , 'words.xml') else ()

let $docname := xmldb:store('/db', 'words.xml', document {<words>{$words}</words>})

let $w := doc($docname)//w[@xml:id='25000']

let $start-time := util:system-time()
let $before := for $i in (1 to 5) return $w/preceding::*[5 + 1 - $i][self::w]/text()
let $word := '[' || $w/text() || ']'
let $after := for $i in (1 to 5) return $w/following::*[$i][self::w]/text()
let $seconds := (util:system-time() - $start-time) div xs:dayTimeDuration("PT1S")

let $string := string-join(($before, $word, $after), ' ')

return "Took " || $seconds || " seconds to produce '" || $string || "'"

adamretter:
Do you have timings for eXist-db 2.2 please?

craigberry:
Average of 5 trials with 2.2 (after changing xquery version to "3.0" and xml:id to id) is 0.560 seconds, so not quite as good v4.3.1 but twice as good as the 5.x development stream.

adamretter:
@craigberry is this still an issue with 5.0.0-RC8? We did some work to improve performance there.

[...]

Issue #1892, opened 2018-05-30
Path expression with nested positional predicate fails on database node


joewiz:
### What is the problem

A path expression containing a nested positional predicate, which works on an in-memory node, fails on an identical node stored in the database.

### What did you expect

I expected a path expression on a structure to work uniformly whether it is in-memory or stored in the database.

### Describe how to reproduce or add a test

The following xqsuite test, adapted from @ChristianGruen's https://github.com/BaseXdb/basex/issues/1573, shows how the in-database function fails while the in-memory function works correctly.

```xquery
xquery version "3.1";

module namespace npt="http://exist-db.org/test/nested-positional-predicate";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare variable $npt:DATA :=
    document {
        <xml>
            <a>
                <b>A</b>
            </a>
            <a>
                <b>B</b>
                <c>correct</c>
            </a>
            <a>
                <b>B</b>
                <c>wrong</c>
            </a>
        </xml>
    };
    
declare
    %test:setUp
function npt:setup() {
    xmldb:create-collection("/db", "test"),
    xmldb:store("/db/test", "test.xml", $npt:DATA)
};

declare
    %test:tearDown
function npt:cleanup() {
    xmldb:remove("/db/test")
};

declare
    %test:assertEquals("<c>correct</c>")
function npt:in-memory() {
    $npt:DATA//c[../preceding-sibling::a[1]/b = 'A']
};

declare
    %test:assertEquals("<c>correct</c>")
function npt:in-database() {
    doc("/db/test.xml")//c[../preceding-sibling::a[1]/b = 'A']
};
```

The results:

```xml
<testsuites>
    <testsuite package="http://exist-db.org/test/nested-positional-predicate"
        timestamp="2018-05-29T23:01:23.898-05:00" failures="1" pending="0" tests="2" time="PT0.029S">
        <testcase name="in-database" class="npt:in-database">
            <failure
                message="assertEquals failed: wrong number of items returned by function. Expected: 1. Got: 0"
                type="failure-error-code-1">&lt;c&gt;correct&lt;/c&gt;</failure>
            <output/>
        </testcase>
        <testcase name="in-memory" class="npt:in-memory"/>
    </testsuite>
</testsuites>
```

### Context information

- eXist-db version + Git Revision hash: eXist-db 4.2.0-SNAPSHOT+201805281153 (e46c45c9d) 
- Java version: Oracle JDK 1.8.0_172-b11
[...]

PR #3075, opened 2019-10-28
Enhance optimisations for positional predicates

wolfgangmm:
### Description:

Frequently used XPath expressions like `$node/following-sibling::node()[1]` or `$node/following::*[1]` were not optimized in any way, thus requiring eXist to scan through the whole set of following nodes. 

* enhance the existing optimization to cover those cases and abort early
* clean up the code to make it more clear when those optimizations are applicable
* remove some large chunks of old, unused code
* improve performance of preceding-sibling axis steps

Because this part of the code has a long history, I limited the PR to small, carefully selected changes at a time, which were thoroughly tested. More will follow later.

### Reference:

#2051 

### Type of tests:

`exist-core/src/test/xquery/optimizer/positional.xql` checks if expressions are correctly optimized. Positional predicates in general are covered by dozens of tests elsewhere.


duncdrum:
@wolfgangmm can you chime in on the effects of this PR on #2129, #1892  it seems directly relevant

there is also #1168 and #2527 which might be effected.

wolfgangmm:
> @wolfgangmm can you chime in on the effects of this PR on #2129, #1892 it seems directly relevant
> 
> there is also #1168 and #2527 which might be effected.

#2129 should be fixed because my test cases are based on it. #1892 is related and I disabled some optimizations on nested filters because they were producing wrong results, but I need to check in detail. The other two are likely unrelated.

exist-core/src/main/java/org/exist/xquery/CastExpression.java
```diff
@@ -164,8 +164,8 @@ public int returnsType() {
 	 * @see org.exist.xquery.AbstractExpression#getDependencies()
 	 */
 	public int getDependencies() {
-//        return expression.getDependencies();
-		return Dependency.CONTEXT_SET | Dependency.CONTEXT_ITEM;
+        return expression.getDependencies() | Dependency.CONTEXT_ITEM;
+//		return Dependency.CONTEXT_SET | Dependency.CONTEXT_ITEM;
```
dizzzz: can be removed

wolfgangmm:
I confirm #1892 is fixed by this PR. I incorporated the tests.

craigberry:
In an average of five trials, the example in #2129 now takes approximately twice as long in the develop branch as of 20d228f45 as it did in 5.0.0.  So as far as wildcard following and preceding we seem to be moving in the wrong direction.

adamretter:
@craigberry Ah... shit! Hey @wolfgangmm and @duncdrum what are your thoughts on this? Do we need to roll-back and re-evaluate?

craigberry:
Probably best to make sure someone else can reproduce my results before rolling back, though I can't really think of anything I could have done that would mess up the results that consistently.  In another test I noticed that preceding is much faster than following, like 30 times faster, so perhaps the magic is working for `preceding::` but not `following::`.  Just a wild guess.

wolfgangmm:
@craigberry the tests in https://github.com/eXist-db/exist/blob/develop/exist-core/src/test/xquery/optimizer/positional.xql are based on the example you gave in #2129 and they clearly demonstrate that the optimization works for the cases covered. If you switch off the optimization, the test case should take minutes rather than seconds. I'm sure it's something small if your particular example doesn't show an improvement. I'll have a look tomorrow, but meanwhile you may want to check the test cases and see if we miss a case.

Also, `following::*[1]` must be many times faster than `preceding::*[1]` if the optimizer steps in.

craigberry:
@wolfgangmm thanks for your work on this.  The details of the changes are way over my head, so I'll limit myself to things I know how to test.

I have looked at positional.xql and some of it looks familiar, but I don't know how to run it -- it just declares a bunch of variables and functions.  In particular, I do not know how to time it in different versions of eXist.  Also, knowing whether a particular optimization has kicked in or not is obviously important for code coverage testing so you get correct results either way, but for me the bottom line is whether the new way of doing things is faster.

A slightly updated version of my test that now includes timings from preceding as well as total times so that one can compare` preceding:`: and `following::` performance, based on the reproducer from #2129, and one that anyone can run without any dependencies, looks like this:

```XQuery
xquery version "3.1";

let $words := for $i in (1 to 50000) return <w xml:id="X{$i}">{$i}</w>

let $x := if (file:exists('/db/words.xml')) then xmldb:remove('/db/' , 'words.xml') else ()

let $docname := xmldb:store('/db', 'words.xml', document {<words>{$words}</words>})

let $docname :='/db/words.xml'
let $w := doc($docname)//w[@xml:id='X25000']

let $start-time := util:system-time()
let $before := for $i in (1 to 5) return $w/preceding::w[5 + 1 - $i]/text()
let $seconds := (util:system-time() - $start-time) div xs:dayTimeDuration("PT1S")
let $before_time := $seconds

let $word := '[' || $w/text() || ']'
let $after := for $i in (1 to 5) return $w/following::w[$i]/text()
let $seconds := (util:system-time() - $start-time) div xs:dayTimeDuration("PT1S")

let $string := string-join(($before, $word, $after), ' ')

return "Took " || $before_time || " before and " || $seconds || " seconds total to produce '" || $string || "'"
```

This consistently takes 66 milliseconds with current develop on my machine, as opposed to 33 milliseconds in 5.0.0. Side note: I now put "X" at the beginning of all the numeric xml:id attributes to comply with the standard.

Here is another test, which has one dependency, the 65MB file [here](https://bitbucket.org/shcdemo/a01/raw/0c6fec6f22acf0d88d41500872b7bfa17cba8496/A01622.xml).  You can put this file anywhere you like in your eXist database and modify the let `$doc := ...` line in the script to point to it.  This file has close to a million nodes.  The test timing is about the same in eXist 5.0.0 and current develop, coming in at around 145 seconds (not milliseconds) on my 2014 MacBook Pro.   More than 140 of the 145 seconds involve the following axis -- the rest is the preceding axis, so following is much, much slower.  The goal here is to fetch 30 words of context around a particular word, and preferably to fetch a dozen or two sets of such context fast enough that a web user won't notice the delay, so some modest fraction of a second would be acceptable.

```XQuery
xquery version "3.1";
import module namespace console="http://exist-db.org/xquery/console" at "java:org.exist.console.xquery.ConsoleModule";

let $start-time := util:system-time()

let $doc := doc('/db/apps/EarlyPrintTexts/texts/A016/A01622.xml')
let $w:= $doc/id("A01622-005-a-01140")

let $seconds := (util:system-time() - $start-time) div xs:dayTimeDuration("PT1S")
let $x := console:log("got word at " || $seconds)

let $before := for $i in (1 to 15)
    return $w/preceding::*[self::tei:w or self::tei:pc][16 - $i]
let $seconds := (util:system-time() - $start-time) div xs:dayTimeDuration("PT1S")
let $x := console:log("got preceding context at " || $seconds)

let $after := for $i in (1 to 15)
    return $w/following::*[self::tei:w or self::tei:pc][$i]
let $seconds := (util:system-time() - $start-time) div xs:dayTimeDuration("PT1S")
let $x := console:log("got following context at " || $seconds)

let $context := ($before, $w, $after)
return element seg { $context }


```

duncdrum:
@adamretter first instinct is to not roll back, as there are other issues addressed by this pr. To understand the difference in performance what are everybody’s parameters (os, conf.xml, java ) ?

craigberry:
My timing tests have been run on MacBook Pro (Retina, 15-inch, Mid 2014), Oracle Java 1.8.0_201, macOS Mojave 10.14.6.  5.0.0 tests run from macOS DMG install.  develop tests run by doing:

$```
 git checkout develop
$ mvn -DskipTests package
$ ./exist-distribution/target/exist-distribution-5.1.0-SNAPSHOT-dir/bin/startup.sh

```
No changes to the default config.xml.

**Comment by adamretter** [comment]
So I think we need to look into @craigberry s tests... I rewrote the predicate stuff a year ago based on his feedback. Whilst it didn't "fix it" it was an improvement, so maybe we can combine @wolfgangmm s work with that?

**Comment by wolfgangmm** [comment]
@craigberry Ok, so there are two distinct issues here:  the first one is simple: static type checking for `$i` in your original query got in the way, so the optimization was not applied. This should be fixed by https://github.com/wolfgangmm/exist/tree/bugfix/positional-type.

Second, the discussion so far is mixing up different classes of expressions involving positional predicates. The class of expressions I tried to address are those using a positional predicate after a wildcard step on the following,  following-sibling, preceding or preceding-sibling axis. Expressions like this are common when you have to output an element differently depending on its immediate neighbour. For example, `earlyprint.odd` uses expressions like this and I can definitely see a performance gain here:

```xquery
not(@join = ('right', 'both')) and not(following-sibling::*[1][self::w and @join = ('left', 'both')]) and not(following-sibling::*[1][self::pc and not(@join='right')]) and not(parent::q/w[last()] = node()) and not(parent::quote/w[last()] = node())
```

In your second test query,  expressions of the kind

```xquery
$w/following::*[self::tei:w or self::tei:pc][$i]
```

may seem like belonging to the same class  as

```xquery
$w/following::*[$i][self::tei:w or self::tei:pc]
```

but there's an important semantic difference:  the first case selects the nth item resulting from evaluating `following::*[self::tei:w or self::tei:pc]` while the second returns the nth following element if it happens to be a w or pc.

The PR addresses the 2nd class of expressions. The first class would need a different kind of optimization.

So if I change the example to

```xquery
xquery version "3.1";

let $doc := doc('/db/A01622.xml')
let $w:= $doc/id("A01622-005-a-01140")

let $before := for $i in (1 to 15)
    return $w/preceding::*[16 - $i][self::tei:w or self::tei:pc]

let $after := for $i in (1 to 15)
    return $w/following::*[$i][self::tei:w or self::tei:pc]

let $context := ($before, $w, $after)
return $context
```

I do see a rather clear improvement with and without this PR applied. For sure this has slightly different semantics and may return less than the 15 nodes, though in practice it could be circumnavigated easily. Also, performance is still not what I ideally would like to see, but the PR clearly provides an improvement of factor > 10 for this class of expressions, so as a first step it's better than not having it.

craigberry:
@wolfgangmm that's good news on the predicates from hell we have in earlyprint.odd -- I didn't test those as I didn't have a timing test ready to go.

With the fix in the positional-type branch, my first test query is now almost twice as fast as in 5.0.0, so that's a big improvement and worth moving forward with.

On my second test, thanks for the explanation that this PR doesn't address that.  It is indeed about 40 times faster if I swap the predicates so the counter is first.  For the sake of future comparisons, it is even another order of magnitude faster than that if I eliminate wildcard axes entirely and write recursive routines as below.  Something feels wrong about that, i.e., that recursive XQuery is so much faster the XPath implementation, but that can be a topic for another day and it looks like this PR will be a big step forward.  Thanks for seeing it through.

```XQuery
xquery version "3.1";

declare
    %private
function local:get-preceding-context($word as node(),
                                     $context as node()*,
                                     $maxwords as xs:int,
                                     $current_word as xs:int) as node()* {
    (: Avoid pathologically slow wildcard axes by getting both w and pc
       and taking the later of the two in document order.
     :)
    let $next_w := $word/preceding::tei:w[1]
    let $next_pc := $word/preceding::tei:pc[1]
    let $next := if ($next_w >> $next_pc) then $next_w else $next_pc

    return
        if (empty($next) or $current_word ge $maxwords)
        then ($next, $context)
        else local:get-preceding-context($next, ($next, $context), $maxwords, $current_word + 1)
};


declare
    %private
function local:get-following-context($word as node(),
                                     $context as node()*,
                                     $maxwords as xs:int,
                                     $current_word as xs:int) as node()* {
    (: Avoid pathologically slow wildcard axes by getting both w and pc
       and taking the earlier of the two in document order.
     :)
    let $next_w := $word/following::tei:w[1]
    let $next_pc := $word/following::tei:pc[1]
    let $next := if ($next_w << $next_pc) then $next_w else $next_pc

    return
        if (empty($next) or $current_word ge $maxwords)
        then ($context, $next)
        else local:get-following-context($next, ($context, $next), $maxwords, $current_word + 1)
};

let $doc := doc('/db/apps/EarlyPrintTexts/texts/A016/A01622.xml')
let $w:= $doc/id("A01622-005-a-01140")
let $before := local:get-preceding-context($w, (), 15, 1)
let $after := local:get-following-context($w, (), 15, 1)

let $context := ($before, $w, $after)
return element seg { $context }
```

Issue #3044, opened 2019-10-08
xsl "document" function hangs when given an internal path to moderately sized XML file


paulmer:
### What is the problem
The transform:transform function called with an XSL transform that reads a document using the "document" function appears to bypass any eXist indexes if the document is referenced with a database path ("/db/.....") instead of an xmldb URL ("xmldb:exist///db/....").  This causes very poor performance and makes the database to appear to hang if the document is even moderately sized (5000-6000 nodes, 200KB)

The 

### What did you expect
I expected roughly the same response time evaluating the same XSL transformation on document("/db/a.xml") and document("xmldb:exist///db/a.xml"), but instead I see response times of 22 seconds vs. 85 milliseconds when q

### Describe how to reproduce or add a test

1. Start with an empty database (I deleted the data directory completely.)

2. Start the eXist client and connect to an embedded database.

3. Load the attached documents to create the collection /db/test containing data.xml and query.xqy.
[test.zip](https://github.com/eXist-db/exist/files/3703660/test.zip)

4. Open query.xqy and edit the value of $xmldbPrefixOpt.  Setting the value to 1 prepends "xmldb:exist://" to the file path that's used in the XSL transformation, setting it to any other value omits that prefix.

5.  Run the script.  It performs a simple query on data.xml to find an element with a specific xml:id attribute.  The query is run twice, once through an XSL transform passed to transform:transform and once in native XQuery, and the number of nodes found as well as the path used for the document are returned.  

On my system running with $xmldbPrefixOpt set to 1 completes in 0.1 to 1 second (depending on the cache state), while running with $xmldbPrefixOpt set to 2 takes about 22 seconds.

### Context information
Please always add the following information
eXist 5.0.0

- Java version (e.g. Java8u121) 1.8.0_191
- MacOS Sierra, Catalina, and RHEL 7
64 bit
- How is eXist-db installed? JAR installer
- Any custom changes in e.g. conf.xml: Altered sync time to 3 seconds, shutdown time to 4 seconds, added preserve-whitespace-mixed-content, disabled transformer caching and enabled full text indexing on attributes.

adamretter:
Indexes are never used within XSLT. eXist-db uses the Saxon XSLT engine for XSLT, and Saxon is not aware of the eXist-db indexes. I imagine what you are seeing here is some sort of URI resolution problem.

paulmer:
[...]

Issue #351, opened 2014-10-01
Feature request: Have Saxon resolve doc() and collection() from xmldb


wshager:
Saxon or other XSLT engines could benefit from resolving docs and collections from eXist's storage directly.

shabanovd:
This should work. @wshager can you confirm that this actually do not work in last develop branch?

wshager:
I'll check, but this issue was opened @adamretter 's request, so perhaps he had different intentions with the resolver, i.e. replacing the current one.

adamretter:
@wshager Do you have tests that show this doesn't work?

wshager:
AFAIK doc() does work, but collection() doesn't. Here's the stack trace for `<xsl:for-each select="collection('xmldb:///db/test/collection')"><xsl:copy-of select="."/></xsl:for-each>`

2015-04-08 22:03:53,211 [eXistThread-233] WARN  (Transform.java [fatalError]:816) - XSL transform reports fatal error: Error reported by XML parser
; Line#: -1; Column#: -1
net.sf.saxon.trans.XPathException: Error reported by XML parser
        at net.sf.saxon.lib.StandardErrorHandler.reportError(StandardErrorHandler.java:95)
        at net.sf.saxon.lib.StandardErrorHandler.fatalError(StandardErrorHandler.java:80)
        at org.apache.xerces.util.ErrorHandlerWrapper.fatalError(Unknown Source)
        at org.apache.xerces.impl.XMLErrorReporter.reportError(Unknown Source)
        at org.apache.xerces.impl.XMLErrorReporter.reportError(Unknown Source)
        at org.apache.xerces.impl.XMLErrorReporter.reportError(Unknown Source)
        at org.apache.xerces.impl.XMLVersionDetector.determineDocVersion(Unknown Source)
        at org.apache.xerces.parsers.XML11Configuration.parse(Unknown Source)
        at org.apache.xerces.parsers.XML11Configuration.parse(Unknown Source)
        at org.apache.xerces.parsers.XMLParser.parse(Unknown Source)
        at org.apache.xerces.parsers.AbstractSAXParser.parse(Unknown Source)
        at org.apache.xerces.jaxp.SAXParserImpl$JAXPSAXParser.parse(Unknown Source)
        at net.sf.saxon.event.Sender.sendSAXSource(Sender.java:405)
        at net.sf.saxon.event.Sender.send(Sender.java:178)
[...]

Issue #41, opened 2013-09-12
Need for a more flexible catalog-loader


lcahlander:
When trying to load the DocBook XSLT files, there are failures due to entities that are not recognized by eXist.

I tried to make another pass at creating an EXPath library that encapsulates the DocBook 5 XSLT files.  The entities in entities.ent are not recognized since they are not in eXist's catalog.xml.  It would be nice if the EXPath packaging allowed for a catalog.xml for the app? 

When trying to load the XSLT files, the following error occurs:

storing document autoidx-kimber.xsl (6 of 99) ...XMLDBException during parse: Failed to invoke method parse in class org.exist.xmlrpc.RpcConnection: java.io.FileNotFoundException: /Applications/eXist-db/entities.ent (No such file or directory)

joewiz:
So this is a "won't fix", it appears?

duncdrum:
maybe something like this could help, in essence write a catalog file dynamically to the filesystem -> execute xsl -> remove catalog.
https://gitlab.existsolutions.com/tei-publisher/tei-publisher-app/blob/master/modules/odd-by-example.xql#L70

line-o:
There were changes to catalog loading in the meantime. Is this issue resolved? @lcahlander please retest with eXist-db 6.2.0

dizzzz:
No response.

PR #3076, opened 2019-10-28
Add classes supporting a SAXSource for included database documents

paulmer:
### Description:

This PR includes classes that handle internal documents referenced in the course of transformations under the transform::transform function using a Serializer under a SAXSource subclass instead of a subclass of DOMSource.  It's a response to the relatively slow performance of the DOMSource previously returned by EXistURIResolver as [reported here](https://github.com/eXist-db/exist/issues/3044), and is intended to treat documents included via xsl:doc() and xsl:include() the same way that the initial XML and XSL parameters to transform::transform are treated.  The performance for included files under these modifications is similar to the performance reading the same files when passed as initial parameters.

The modifications pass the existing transformation tests in my build environment.

------

### Reference:
#3044

### DIFF
--- a/exist-core/src/main/java/org/exist/xslt/EXistDbInputSource.java
+++ b/exist-core/src/main/java/org/exist/xslt/EXistDbInputSource.java
@@ -1,0 +1,63 @@
+/*
+ * eXist Open Source Native XML Database
+ * Copyright (C) 2001-2019 The eXist Project
+ * http://exist-db.org
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xslt;
+
+import java.io.InputStream;
+import java.io.Reader;
+
+import org.exist.dom.persistent.DocumentImpl;
+import org.exist.storage.DBBroker;
+import org.xml.sax.InputSource;
+
+/**
+ * {@link org.xml.sax.InputSource} identifying a document within the eXist database.
+ *
+ * @author <a href="mailto:Paul.L.Merchant.Jr@dartmouth.edu">Paul Merchant, Jr.</a>
+ */
+
+public class EXistDbInputSource extends InputSource {
+    protected DBBroker broker;
+    protected DocumentImpl doc;
+    
+    public EXistDbInputSource(DBBroker broker, DocumentImpl doc) {
+        super();
+        
+        this.broker = broker;
+        this.doc = doc;
+    }
+    
+    public DBBroker getBroker() {
+        return this.broker;
+    }
+    
+    public DocumentImpl getDocument() {
+        return this.doc;
+    }
+    
+    @Override
+    public void setByteStream(InputStream stream) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void setCharacterStream(Reader stream) {
+        throw new UnsupportedOperationException();
+    }    
+}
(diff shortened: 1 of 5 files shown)


duncdrum:
see #351 
and somewhat see #41

dizzzz:
thank you for your PR

paulmer:
What kind of tests are you looking for beyond the transformation tests that already exist?

dizzzz:
As you added some new code, I'd  expect to see some progressions tests that show that the new code is actual functional.

adamretter:
@dizzzz this is just a performance optimisation. The behaviour should be the same as before.

paulmer:
@dizzzz I'm sorry, I'm really puzzled by your comment.  There are existing tests that drive this code in the transformation suite.  I know this because there were some details I had to add to make sure my modifications worked with those tests.  I noted this in the original commit.

I'm much more concerned with why the CI system reported a stack overflow in Java 11, but all tests work on my system under Java 11.  There are other testing issues I've brought up on Slack that are quite concerning too.

paulmer:
Hi Adam, no problem adding the conding conventions, but I'm not sure how source can be null in EXistDBSource or EXistDbXMLReader?   Source is initialized in the constructor of EXistDBSource and both EXistDBSource.setInputSource() and EXistDbXMLReader.parse() throw exceptions if not given the expected objects, so I think I have it covered.  Nevertheless, I can add a couple of null tests in if it'll make things clearer.

I'll clean up the handing of the SystemID in EXistDBSource.setSystemId(), since it's a little off-spec.  As written EXistDBSource isn't able to do everything a SaxSource can, since it can't create an InputSource when given a SystemID, but as that's not necessary for it's current obligations, I'll opt to throw an exception.

A new commit is coming shortly....

On Oct 31, 2019, at 1:16 PM, Adam Retter <notifications@github.com<mailto:notifications@github.com>> wrote:


@adamretter requested changes on this pull request.

Hi Paul,

In general this looks great :-)

Just a few minor style things if you don't mind

________________________________

In exist-core/src/main/java/org/exist/xslt/EXistDbInputSource.java<https://github.com/eXist-db/exist/pull/3076#discussion_r341078228>:

> +
+import java.io.InputStream;
+import java.io.Reader;
+
+import org.exist.dom.persistent.DocumentImpl;
+import org.exist.storage.DBBroker;
+import org.xml.sax.InputSource;
+
+/**
+ * {@link org.xml.sax.InputSource} identifying a document within the eXist database.
+ *
+ * @author <a href="mailto:Paul.L.Merchant.Jr@dartmouth.edu">Paul Merchant, Jr.</a>
+ */
+
+public class EXistDbInputSource extends InputSource {
+    protected DBBroker broker;


As an eXist-db coding style thing - all class fields, variables, and function arguments should be final where possible.

________________________________

In exist-core/src/main/java/org/exist/xslt/EXistDbSource.java<https://github.com/eXist-db/exist/pull/3076#discussion_r341078455>:

> +        return new EXistDbXMLReader();
+    }
+
+    @Override
+    public void setInputSource(InputSource inputSource) {
+        if (!(inputSource instanceof EXistDbInputSource)) {
+            throw new UnsupportedOperationException("EXistDbSource only accepts EXistDbInputSource");
+        }
+
+        this.source = inputSource;
+    }
+
+    @Override
+    public void setSystemId(String systemId) {
+        this.systemId = systemId;
+        this.source.setSystemId(systemId);


Add a check for source != null

________________________________

In exist-core/src/main/java/org/exist/xslt/EXistDbInputSource.java<https://github.com/eXist-db/exist/pull/3076#discussion_r341078728>:

> +import java.io.InputStream;
+import java.io.Reader;
+
+import org.exist.dom.persistent.DocumentImpl;
+import org.exist.storage.DBBroker;
+import org.xml.sax.InputSource;
+
+/**
+ * {@link org.xml.sax.InputSource} identifying a document within the eXist database.
+ *
+ * @author <a href="mailto:Paul.L.Merchant.Jr@dartmouth.edu">Paul Merchant, Jr.</a>
+ */
+
+public class EXistDbInputSource extends InputSource {
+    protected DBBroker broker;
+    protected DocumentImpl doc;


Do these member variables need to be protected rather than private.

________________________________

In exist-core/src/main/java/org/exist/xslt/EXistDbXMLReader.java<https://github.com/eXist-db/exist/pull/3076#discussion_r341079357>:

> +    }
+
+    @Override
+    public Object getProperty(String name) {
+        return null;
+    }
+
+    @Override
+    public void parse(InputSource input) {
+        if (!(input instanceof EXistDbInputSource)) {
+            throw new UnsupportedOperationException("EXistDbXMLReader only accepts EXistDbInputSource");
+        }
+
+        EXistDbInputSource source = (EXistDbInputSource) input;
+        try {
+            final Serializer serializer = source.getBroker().newSerializer();


source can be null - we should avoid a potential NPE here.

adamretter:
Thank you very much @paulmer

Issue #2627, opened 2019-04-08
As a developer I would like file:sync to also delete surplus files


line-o:
### What is the problem

Calling `file:sync()` only updates and adds files.   

This is a feature request. I would like to start a discussion on how to synchronise a collection to filesystem in the future.

Changing this might break existing workflows, that rely on the current behaviour of `file:sync`.
Also, oftentimes one would want to exclude certain files from deletion (e.g. `.git`).

Possible solutions:

- a new function: `file:sync-remove($exclude-pattern)`
- a additional signature with 2 parameters: `file:sync($remove, $exclude-pattern)`
- configuration parameter `sync-remove: true` in combination with `file:sync($exclude-patttern)`

### What did you expect

I would expect that syncing files would also delete files that are not present in the source collection.

### Describe how to reproduce or add a test

- Create a collection 'sync-test'
- create files 'a' and 'b'
- synchronise this collection to the filesystem in directory 'sync-test'
- delete file 'b' in eXist-DB
- synchronise again
- file 'b' is still existent in the filesystem

### Context information

- tested with eXist-db version 4.1.0  
- java version "1.8.0_151"
- Operating system: Ubuntu 14.04.5 LTS (GNU/Linux 4.4.0-130-generic x86_64)
- How is eXist-db installed? clone from GitHub
- custom changes, but the same applies for vanilla existdb

duncdrum:
There is some overlap with eXist-db/atom-existdb#40.
I know that there have been reservations about the wisdom of enabling deletions on the filesystem. So from a security standpoint a separate `trust-me-i-know-what-i-m-doing` function makes sense. Utlimately though i d like syncing to the fs to be smarter, i keep things in one folder on the fs e.g. `src` and in a collection at `collection/resources/...` in exist. it would be nice if exist could keep track of such paths.

line-o:
@duncdrum I am explicitly against a folder mapping. When a collection is synchronized to the filesystem I want to see it copied as verbatim as possible. That is likely due to my use case.
The only exception would be extra files on the filesystem excluded from deletion/ clean up.

line-o:
@duncdrum thinking about it file:sync could get an additional signature that accepts a single `$options` parameter.
```
<options delete='true' exclude='.git'><mapping in='./**' out='src/**' /></options>

duncdrum:
[...]

PR #3084, opened 2019-11-01
added optional 'prune' param to file:sync(...)

sten1ee:
### Description:
```file:sync(collection, targetPath, dateTime?)``` is now extended to accept another optional arg:
```file:sync(collection, targetPath, dateTime?, prune?)```
When not specified the function behaves as usual,
When specified and true, ```sync``` will take care to delete obsolete files and directories, i.e. it will delete any file/dir that does not correspond to a doc/collection currently in the DB.

### Reference:
There is no GitHub issue corresponding to this PR

### Type of tests:
There are no tests for the new feature.
I could not find the tests for file:sync anyway, so point me to those, pls. I will build on top of them.

### DIFF
--- a/extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
+++ b/extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
@@ -8,6 +8,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.attribute.FileTime;
+import java.net.URISyntaxException;
 import java.util.*;
 
 import javax.xml.XMLConstants;
@@ -40,6 +41,7 @@ import org.exist.xquery.FunctionSignature;
 import org.exist.xquery.XPathException;
 import org.exist.xquery.XQueryContext;
 import org.exist.xquery.value.DateTimeValue;
+import org.exist.xquery.value.BooleanValue;
 import org.exist.xquery.value.FunctionParameterSequenceType;
 import org.exist.xquery.value.FunctionReturnSequenceType;
 import org.exist.xquery.value.Sequence;
@@ -64,7 +66,10 @@ public class Sync extends BasicFunction {
                         Cardinality.EXACTLY_ONE, "The full path or URI to the directory"),
                 new FunctionParameterSequenceType("dateTime", Type.DATE_TIME, 
                         Cardinality.ZERO_OR_ONE, 
-                		"Optional: only resources modified after the given date/time will be synchronized.")
+                        "Optional: only resources modified after the given date/time will be synchronized."),
+                new FunctionParameterSequenceType("prune", Type.BOOLEAN,
+                        Cardinality.ZERO_OR_ONE,
+                        "Optional: delete any file/dir that does not correspond to a doc/collection currently in the DB.")
             },
             new FunctionReturnSequenceType(Type.BOOLEAN, Cardinality.EXACTLY_ONE, "true if successful, false otherwise")
         );
@@ -88,15 +93,21 @@ public class Sync extends BasicFunction {
 		}
 		
 		final String collectionPath = args[0].getStringValue();
+
+		final String target = args[1].getStringValue();
+		Path targetDir = FileModuleHelper.getFile(target);
+
 		Date startDate = null;
 		if (args[2].hasOne()) {
 			DateTimeValue dtv = (DateTimeValue) args[2].itemAt(0);
 			startDate = dtv.getDate();
 		}
-        
-		final String target = args[1].getStringValue();
-        Path targetDir = FileModuleHelper.getFile(target);
-        
+
+		boolean prune = false;
+		if (args[3].hasOne()) {
+			BooleanValue bv = (BooleanValue) args[3].itemAt(0);
+			prune = bv.getValue();
+		}
 		
 		context.pushDocumentContext();
 		final MemTreeBuilder output = context.getDocumentBuilder();
@@ -111,7 +122,7 @@ public class Sync extends BasicFunction {
 			output.addAttribute(new QName("collection", FileModule.NAMESPACE_URI), collectionPath);
 			output.addAttribute(new QName("dir", FileModule.NAMESPACE_URI), targetDir.toAbsolutePath().toString());
 			
-			saveCollection(XmldbURI.create(collectionPath), targetDir, startDate, output);
+			saveCollection(XmldbURI.create(collectionPath), targetDir, startDate, prune, output);
 			
 			output.endElement();
 			output.endDocument();
@@ -123,7 +134,7 @@ public class Sync extends BasicFunction {
 		return output.getDocument();
 	}
 
-	private void saveCollection(final XmldbURI collectionPath, Path targetDir, final Date startDate, final MemTreeBuilder output) throws PermissionDeniedException, LockException {
+	private void saveCollection(final XmldbURI collectionPath, Path targetDir, final Date startDate, final boolean prune, final MemTreeBuilder output) throws PermissionDeniedException, LockException {
 		try {
 			targetDir = Files.createDirectories(targetDir);
 		} catch(final IOException ioe) {
@@ -142,6 +153,9 @@ public class Sync extends BasicFunction {
 			if (collection == null) {
 				reportError(output, "Collection not found: " + collectionPath);
 				return;
+			}
+			if (prune) {
+				pruneCollectionEntries(collection, targetDir, output);
 			}
 			for (final Iterator<DocumentImpl> i = collection.iterator(context.getBroker()); i.hasNext(); ) {
 				DocumentImpl doc = i.next();
@@ -162,7 +176,7 @@ public class Sync extends BasicFunction {
 		
 		for (final XmldbURI childURI : subcollections) {
 			final Path childDir = targetDir.resolve(childURI.lastSegment().toString());
-			saveCollection(collectionPath.append(childURI), childDir, startDate, output);
+			saveCollection(collectionPath.append(childURI), childDir, startDate, prune, output);
 		}
 	}
 
@@ -170,6 +184,26 @@ public class Sync extends BasicFunction {
 		output.startElement(new QName("error", FileModule.NAMESPACE_URI), null);
 		output.characters(msg);
 		output.endElement();
+	}
+
+	private void pruneCollectionEntries(final Collection collection, final Path targetDir, final MemTreeBuilder output) {
+		try {
+			Files.walk(targetDir, 1).forEach(path -> {
+				try {
+					String fname = path.getFileName().toString();
+					XmldbURI dbname = XmldbURI.xmldbUriFor(fname);
+					if (!collection.hasDocument(context.getBroker(), dbname)
+					 && !collection.hasChildCollection(context.getBroker(), dbname)) {
+						Files.deleteIfExists(path);
+					}
+				} catch (IOException | URISyntaxException
+						| PermissionDeniedException | LockException e) {
+					reportError(output, e.getMessage());
+				}
+			});
+		} catch (IOException e) {
+			reportError(output, e.getMessage());
+		}
 	}
 
 	private void saveXML(final Path targetDir, final DocumentImpl doc, final MemTreeBuilder output) {

### CODE REVIEW

**Comment by sten1ee** [comment]
@adamretter Could you please 'assign' a person from the eXist team, who is going to answer my questions and help me get this PR accepted!

As a beginning I see these issues: 
1. appveyor has failed the Win build (due to timeout), someone has to restart it or ... remove the Win build altogether
1. On the invocation site  the new functionality will look like:
```xquery
file:sync(
             concat($common:data-path, '/', $collection), 
             concat('/', $repo-path, '/', $collection), 
             (),
             true
         )
```
which is far from informative. Are there ['keyword arguments'](https://thoughtbot.com/blog/ruby-2-keyword-arguments) in xquery, so that the last arg may be supplied in the form:
```
         prune: true
```

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -64,7 +66,10 @@
                         Cardinality.EXACTLY_ONE, "The full path or URI to the directory"),
                 new FunctionParameterSequenceType("dateTime", Type.DATE_TIME, 
                         Cardinality.ZERO_OR_ONE, 
-                		"Optional: only resources modified after the given date/time will be synchronized.")
+                        "Optional: only resources modified after the given date/time will be synchronized."),
+                new FunctionParameterSequenceType("prune", Type.BOOLEAN,
+                        Cardinality.ZERO_OR_ONE,
```
adamretter: The Cardinality for this should be EXACTLY_ONE. It doesn't really make sense for this to be optional as it is a boolean value.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -88,15 +93,21 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
 		}
 		
 		final String collectionPath = args[0].getStringValue();
+
+		final String target = args[1].getStringValue();
+		Path targetDir = FileModuleHelper.getFile(target);
```
adamretter: Could this variable be `final`?

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -88,15 +93,21 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
...
+
 		Date startDate = null;
 		if (args[2].hasOne()) {
 			DateTimeValue dtv = (DateTimeValue) args[2].itemAt(0);
```
adamretter: Could `dv` and `bv` be `final`?

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -172,6 +186,26 @@ private void reportError(final MemTreeBuilder output,final String msg) {
 		output.endElement();
 	}
 
+	private void pruneCollectionEntries(final Collection collection, final Path targetDir, final MemTreeBuilder output) {
+		try {
+			Files.walk(targetDir, 1).forEach(path -> {
+				try {
+					String fname = path.getFileName().toString();
+					XmldbURI dbname = XmldbURI.xmldbUriFor(fname);
```
adamretter: Can these variables be `final`?

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -143,6 +154,9 @@ private void saveCollection(final XmldbURI collectionPath, Path targetDir, final
 				reportError(output, "Collection not found: " + collectionPath);
 				return;
 			}
+			if (prune) {
+				pruneCollectionEntries(collection, targetDir, output);
+			}
 			for (final Iterator<DocumentImpl> i = collection.iterator(context.getBroker()); i.hasNext(); ) {
 				DocumentImpl doc = i.next();
 				if (startDate == null || doc.getMetadata().getLastModified() > startDate.getTime()) {
```
adamretter: You should READ_LOCK the document before reading from it

adamretter:
@sten1ee You can find tests in `exist/extensions/modules/file/src/test`. You can implement your sync tests in either Java or XQuery.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
sten1ee: I've made it "optional" with the single purpose that the API remains backward compatible.
Old style invocations of file:sync(...) will still work.
If the ```prune``` param is made non-optional, then:
1. Old-style invocations of that function will be rendered broken (i.e. will have to be revised)
2. IIRC the previous (3rd) param of the function will also need to become non-optional, as to the best of my understanding, you can't have optional param, followed by non-optional, right?

adamretter: That's not backwards compatible. There is no polymorphism in XQuery. You have to keep the original function signature, and then define additional function signatures with additional arities.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
sten1ee: Sure.
Btw that's only one (of the hundreds) of reasons why I prefer Kotlin to Java these days.

adamretter: Sure... But, eXist-db is in Java ;-)

sten1ee: Actually - no.
targetDir is (conditionally changed) on line 117.
And it is not something I introduced -- this code has been there for a long time.
Of course, this can be avoided (as the course in "Compiler construction" teaches, every program can be converted to equivalent SSA form, but the resultant code is not necessarily more readable.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
sten1ee: Done, but why do this for block-scoped vars, that go out of scope two lines further?

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
sten1ee: (I am assuming you are referring to _the collection_ when you say _the document_):
As far as I see on line 6 lines up (on line 152) the collection is locked and opened:
```try(final Collection collection = context.getBroker().openCollection(collectionPath, LockMode.READ_LOCK))```

File: extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java (hunk 4 above)
Comment (sten1ee) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: Done.

File: extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java (hunk 1 above)
Comment (sten1ee) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: Wait, wait, wait.
Then (in order not to break existing code) this extended 'prune' functionality has to come together with a new function, e.g. ```file:sync_ex(...)``` or ```file:sync_prune(...)``` or something.
Is it not so?

**Comment by sten1ee** [comment]
@adamretter : you did not answer my question:
Are there 'keyword arguments' in xquery, so that the last arg may be supplied in the form:
         ```prune: true```

And I also have this one:
So, an optional param to a xQuery function means that that param may be 'null', but this not exempt callers from having to explicitly specify 'null' as an argument for this param?

File: extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java (hunk 1 above)
Comment (adamretter) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: You don't need a new named function as they are on different arities. But you have to design two signatures, the original signature, and the new signature which has an additional argument. The new signature should then be added to the FileModule.java

File: extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java (hunk 2 above)
Comment (adamretter) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: Ah okay. No worries.

File: extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java (hunk 3 above)
Comment (adamretter) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: As a code style thing - we just try to prefer that all variables are `final` unless they can't be for some logical reason.

File: extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java (hunk 5 above)
Comment (adamretter) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: You are calling `doc.getMetadata().getLastModified()` which is on the document, so you should read-lock the document.

**Comment by adamretter** [comment]
> Are there 'keyword arguments' in xquery, so that the last arg may be supplied in the form:
prune: true

No.

> So, an optional param to a xQuery function means that that param may be 'null', but this not exempt callers from having to explicitly specify 'null' as an argument for this param?

There is no `null` in XQuery; The closest thing is likely an "empty sequence". So even if the cardinality of the parameter is specified as `ZERO_OR_ONE`, you still have to specify a value when calling even if it is the empty sequence.

File: extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java (hunk 5 above)
Comment (sten1ee) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: No, I am not.
Once again -- this is old code, I have not modified it.
So you see - I am trying to get a feature PR through and then it turns out the code _as it was_ before my PR was not exactly conformant.
I don't know what is the proper way to LOCK a doc in this case (and in general).
Could you suggest the relevant change to the code and I will take it from there.
(You know I've seen GitLab gives the reviewer a way to suggest changes in the code, there should be something equivalent here)

Comment (adamretter) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: `try (final ManagedLock lock = broker.getBrokerPool().getLockManager().acquireDocumentLock(docUri, LockMode.READ_LOCK) {
   ...
}`

Comment (sten1ee) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: Thanks.
You've missed one missing final - line 162: ```DocumentImpl doc = i.next();``` ;-)

Comment (sten1ee) [reply extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java]: Alternatively -- is it not a better idea (for the sake of 'Atomicity')
to read_lock the whole collection tree that is being sync-ed?

**Comment by sten1ee** [comment]
@adamretter in the Sync signature:
```java
                new FunctionParameterSequenceType("collection", Type.STRING,
                        Cardinality.EXACTLY_ONE, "The collection to sync."),
                new FunctionParameterSequenceType("targetPath", Type.ITEM, 
                        Cardinality.EXACTLY_ONE, "The full path or URI to the directory"),
```
```targetPath``` is declared like ITEM (as opposed to STRING) but is later treated as if it is string:
```java
		final String collectionPath = args[0].getStringValue();

		final String target = args[1].getStringValue();
		Path targetDir = FileModuleHelper.getFile(target);
```
is that a mistake?

sten1ee:
I need a (xquery) hand here, in order to craft the test for this sync/4 (sync with prune)
The only sensible XQuery test I can see (so far) is:
```xquery
declare
    %test:pending("need to mechanism to setup a temporary file to work with")
    %test:assertEquals("data", "true", "true", "true")
function serialization:overwrite() {

    let $node-set := text {"data"}
    let $path := system:get-exist-home() || "/test.txt"
    let $parameters := ()
    let $append := true()
    let $remove := file:delete($path)
    let $ser1 := 	file:serialize($node-set, $path, (), false())
    let $ser2 := 	file:serialize($node-set, $path, (), false())
    let $read := file:read($path)
    let $remove := file:delete($path)
    return ($read, $ser1, $ser2, $remove)
};
```
I am extremely short in XQuery skills, that's why I need help with the following test scenario:
```text
1. create a collection C containing two docs: A and B
2. sync(C) the collection to the disk
3. check both files are in target dir with expected content
4. update contents of doc A
5. sync(C ... mode='prune')
6. check both files are in target dir with expected content
7. remove/delete doc B from collection C
8. sync(C ... mode='no_prune')
9. check both files are in target dir with expected content
10. sync(C ... mode='prune')
11. check only file A is in target dir with expected content
```

8. 
5. check both files are in target dir with expected content

adamretter:
> targetPath is declared as Type.ITEM (as opposed to Type.STRING) but is later treated as if it is Type.STRING:

It should likely have been declared as `Type.String`, however we can't change it easily now, and the `getStringValue` will likely give the expected result anyway.

adamretter:
@duncdrum @joewiz any chance either of you could help @sten1ee with some XQuery for his test above?

duncdrum:
@joewiz can i pass creating an XQsuite test on to you? I'm swamped with more pressing issues at the moment. If so can you assign yourself the ticket.

sten1ee:
> We are traditionally hesitant with on-disc delete operations
> Is there anything preventing me to `prune C:\>`

You have a point here.
Security-wise I haven't changed anything in the existing code (the one that implemented `file:sync` but without option to `prune`).
I think exploit-wise *overwriting* is even more dangerous than *deleting* - imagine if file:sync gives you a chance to replace a system file.
OTOH disaster-wise deleting seems to be more dangerous.

Now back to the code:
- on the eXist side: You can do that (ask to delete/overwrite files in C:\>)
- on the OS side: hopefully, eXist is not running as a user, who is granted the right to do that?

line-o:
related issue https://github.com/eXist-db/exist/issues/2627

sten1ee:
@line-o, are you willing to help this PR get through by authoring the required TC?
Please, see my post from Nov 5th (the ones starting with _'I need a (xquery) hand here'_)
Perhaps the TC may be simplified -- in general it has to demonstrate that the functionality is absent before that PR and present with it.

sten1ee:
@duncdrum, perhaps you haven't been notified about my reply to your question about wiping out C:\> with 'prune'
Please, check my comment from Nov 13th and tell me what do we do with it - is the option with using OS user rights enough of a 'protection' against this disaster *or* if it is not -- what else do you suggest ?

line-o:
@sten1ee First of all, for me, just adding a prune argument is not enough. I need the ability to exclude files so that `.git` and other files are left untouched.

line-o:
@duncdrum `file:sync` is able to modify a folder on the file system based on OS permissions. For production systems this should always be a user with limited rights. Write access includes the ability to remove files. Therefore, I see no issue in removing files when syncing to the filesystem.

joewiz:
@line-o's suggestion about a mechanism for excluding file and folder paths and patterns makes a lot of sense, particularly given that we don't always store `.git` folders or other build artifacts in the database.

@sten1ee Would you be interested in adding this feature to this PR, or in a follow-on PR? I imagine this would take the form of a new `$excludes as xs:string*` parameter that would take 0-or-more strings containing the filenames or patterns. For a function that implements filename/patterns, see http://exist-db.org/exist/apps/fundocs/index.html?q=xmldb:store-files-from-pattern. 

I would be happy to help craft an XQSuite test.

sten1ee:
@joewiz I think I can do that, it shouldn't be that much of a work. I only need to coordinate this with (my) management as after all it is 84000 that's paying me.

Apart from that, it'd be very helpful for me if you go ahead and craft some XQSuite test for the already submitted code. Why is that? I am almost sure there will be things to fix in it, that I can't spot without a test.
If you craft the test keeping in mind the '_mechanism for excluding file and folder paths and patterns_' that is pending -- it will be quite easy to extend the test to cover this later.

I prefer we get all this done in a single PR, but it's not that important in the end.

sten1ee:
@joewiz, @line-o I were given 'green light' to implement the 'excluding mechanism', let's get it going.
I will include also @adamretter in the discussion:

1. Do we all agree that supporting the .gitignore path pattern matching format (or a subset of it) is a sane approach in this case? Or do we go for reusing the functionality of eXist's NodePathPattern for that job?
2. If we decide to go for .gitignore's format: @adamretter, do we already have in eXist a piece of code or a lib that does that? Or if not - is using a third party lib for that task an option? I found this one, for example (distributed under _Eclipse Distribution License - v 1.0_) - https://github.com/iloveeclipse/java-things

@adamretter i

sten1ee:
@joewiz
I am more or less done (this includes the `exclusion patterns` functionality.
When will you be able to craft some TCs ?

sten1ee:
@joewiz, are you still willing to contribute the XQSite test, so that we can move fwd with that feature?

joewiz:
@sten1ee Sorry for the delay, I haven't forgotten. Just working on some hard deadlines and will hope to be able to supply some tests soon. In the meantime, Duncan smartly added a [a skeleton XQSuite test](https://github.com/eXist-db/exist/blob/develop/.github/ISSUE_TEMPLATE/bug_report.md) to the issue template, so if that provides the scaffolding you needed to add a test, by all means. But if that's not sufficient, I will still follow through on creating tests as soon as I can.

line-o:
Just a heads up: @sten1ee your PR was discussed in todays Community Meeting. We are still committed to get it merged, but none of us has the time to prepare a test suite at the moment. As a result, this feature will not make it into the next release, which is around the corner.

adamretter:
@sten1ee Of course, if you want to send a test-suite yourself, then we can get it merged ASAP.

joewiz:
With https://github.com/eXist-db/exist/issues/3704, I now have a technique for creating and accessing temporary directories on the filesystem, which is needed for writing tests of file:sync. We could use the same technique for creating the xqsuite test for this PR. However, I see two challenges:

1. Some conflicts are present in the PR now and need to be resolved.
2. In light of https://github.com/eXist-db/exist/issues/3704, it's clear to me that file:sync needs a `$parameters` argument, corresponding to `file:serialize`, to allow developers to specify serialization parameters when syncing a collection to disk. See http://exist-db.org/exist/apps/fundocs/view.html?uri=http://exist-db.org/xquery/file#serialize.3. Should this `$parameters` argument come before or after the `$prune` argument? Or should we consolidate `$parameters` and `$prune` (and even `$dateTime`) into an `$options as map(*)*` map, like:

```xquery
file:sync(
    $source-collection, 
    $destination-directory, 
    map { 
        "dateTime": xs:dateTime("2021-01-12T11:47:08.12-05:00"),
        "prune": true(),
        "indent": false(),
        "expand-xincludes": false(),
        "process-xsl-pi": false(),
        "encoding": "UTF-8",
        "omit-xml-declaration": true()
    }
)

Before we merge this PR, we'd ideally discuss the direction of parameter handling and make any necessary adjustments. Thoughts?

**Comment by joewiz** [comment]
Also, just to clarify for the purposes of writing tests, I seem to recall that `file:sync` is one-way - from database to disk - and will not pull changes from disk back to database.

**Comment by joewiz** [comment]
(Also on the topic of how file:sync serializes, see https://github.com/eXist-db/exist/issues/2394.)

**Comment by line-o** [comment]
@joewiz Yes, you are right file:sync only allows to sync to disk not from it.
Re: your proposed change to the function signature:

```xquery
file:sync(
    $source-collection, 
    $destination-directory, 
    map { "dateTime": xs:dateTime("2021-01-12T11:47:08.12-05:00"), ... }
)
```

I would love to change file:sync#3 to have the options map as the third parameter. But this then means that we have an _almost_ breaking change to the API.
We could allow a xs:dateTime or a map(*) as the third parameter to file:sync#3, but this ODD.
While _I really like to do that_, maybe creating a separate file:sync#4 with explicit optional xs:dateTime as the third parameter and the options map as the fourth is the safer option.

joewiz:
@line-o Great. And I actually like your idea of the hybrid 3rd parameter. We could allow `(xs:dateTime | map(*))` at first and immediately deprecate `xs:dateTime`, flagging it as to-be-deleted in the next major version of eXist.

line-o:
@joewiz The first working version with the 3rd parameter accepting either an xs:dateTime or an options map is now in the follow-up PR #4081 
I invite @joewiz and @sten1ee to discuss progress there and would propose to close this PR as development has taken a different route.

PR #3092, opened 2019-11-04
[feature] implemented array:put

windauer:
### Description:
Implemented array:put

### Reference:
[array:put](https://www.w3.org/TR/xpath-functions-31/#func-array-put)

### Type of tests:
XQSuite Tests: exist-core/src/test/xquery/arrays/arrays.xql


duncdrum:
see https://github.com/eXist-db/documentation/blob/ed45931fe7bc829664db02c2ef075b976490366f/src/main/xar-resources/data/xquery/xquery.xml#L116

windauer:
@duncdrum I just send a PR to update the documentation, thanks for the hint! https://github.com/eXist-db/documentation/pull/408 - once it is merged I will do the same for the json-to-xml function.

PR #3093, opened 2019-11-04
[feature] implemented xquery function json-to-xml

windauer:
### Description:
Implemented fn:json-to-xml 

### Reference:
[Specification](https://www.w3.org/TR/xpath-functions-31/#func-json-to-xml)

### Type of tests:
XQSuite: exist-core/src/test/xquery/xquery3/json-to-xml.xql

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/JSON.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/JSON.java
@@ -82,12 +82,12 @@ public class JSON extends BasicFunction {
         )
     };
 
-    public final static String OPTION_DUPLICATES = "duplicates";
-    public final static String OPTION_DUPLICATES_REJECT = "reject";
-    public final static String OPTION_DUPLICATES_USE_FIRST = "use-first";
-    public final static String OPTION_DUPLICATES_USE_LAST = "use-last";
-    public final static String OPTION_LIBERAL = "liberal";
-    public final static String OPTION_UNESCAPE = "unescape";
+    public static final String OPTION_DUPLICATES = "duplicates";
+    public static final String OPTION_DUPLICATES_REJECT = "reject";
+    public static final String OPTION_DUPLICATES_USE_FIRST = "use-first";
+    public static final String OPTION_DUPLICATES_USE_LAST = "use-last";
+    public static final String OPTION_LIBERAL = "liberal";
+    public static final String OPTION_UNESCAPE = "unescape";
     public static final QName KEY = new QName("key",null);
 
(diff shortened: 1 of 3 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/JSON.java
```diff
@@ -68,6 +88,7 @@
     public final static String OPTION_DUPLICATES_USE_LAST = "use-last";
     public final static String OPTION_LIBERAL = "liberal";
     public final static String OPTION_UNESCAPE = "unescape";
+    public static final QName KEY = new QName("key",null);
```
joewiz: A minor, stylistic suggestion: perhaps it's best to follow the order used on previous lines - `final static` instead of `static final`.

adamretter: The recommended formulation is `public static final`. Regardless, it should be consistent throughout ;-)

duncdrum:
see https://github.com/eXist-db/documentation/blob/ed45931fe7bc829664db02c2ef075b976490366f/src/main/xar-resources/data/xquery/xquery.xml#L137

exist-core/src/test/xquery/xquery3/json-to-xml.xql
```diff
@@ -0,0 +1,49 @@
...
+(:~
+ : Test various F+O functions
+ :)
+module namespace jsonxml="http://exist-db.org/xquery/test/jsonxml";
+
+import module namespace test="http://exist-db.org/xquery/xqsuite" at "resource:org/exist/xquery/lib/xqsuite/xqsuite.xql";
+
+
+declare
+    %test:assertEquals("<fn:map xmlns:fn='http://www.w3.org/2005/xpath-functions'><fn:number key='x'>1</fn:number><fn:array key='y'><fn:number>3</fn:number><fn:number>4</fn:number><fn:number>5</fn:number></fn:array></fn:map>")
```
dizzzz: during development of the mongodb and couchbase drivers, I found several issues while parsing complexer structures (deeper, nested). You might want to test these as well.

joewiz: If we can get the XQuery test suite runner working under eXist 5.0.0, I think we could be confident: https://github.com/eXist-db/exist-xqts-runner/issues/7.

windauer:
updated the documentation ( https://github.com/eXist-db/documentation/pull/409 ) and followed the stylistic suggestion from @joewiz but in the order @adamretter proposed ;)

PR #3107, opened 2019-11-13
Lucene:  add support for minimum-must-match on BooleanQuery 

### LINKED ISSUE #392: Patch: add support for minimum-must-match on BooleanQuery
Below is a small patch for your consideration to allow specification of the minimumNumberShouldMatch for BooleanQuery(s) in XMLToQuery.

The feature is not available in the classic Lucene query syntax. It is made available in the SOLR eDisMax query parser.

The feature allows to cut-off needless disjunctive searches.

It used like:

``` xquery
    let $min := if (count($terms) le 3) then $terms else (count($terms) + 1) idiv 2
    let $query := 
        <bool min="{$min}">{
            for $term in $terms
            return
                <term>{$term}</term>
            }</bool>
```

Here is the patch since I don't see how to attach the patch file to the issue:

```
### Eclipse Workspace Patch 1.0
#P eXist-2.1dev
diff --git extensions/indexes/lucene/src/org/exist/indexing/lucene/XMLToQuery.java extensions/indexes/lucene/src/org/exist/indexing/lucene/XMLToQuery.java
index 7aadf32..f455e71 100644
--- extensions/indexes/lucene/src/org/exist/indexing/lucene/XMLToQuery.java
+++ extensions/indexes/lucene/src/org/exist/indexing/lucene/XMLToQuery.java
@@ -344,6 +344,17 @@

     private Query booleanQuery(String field, Element node, Analyzer analyzer, Properties options) throws XPathException {
         BooleanQuery query = new BooleanQuery();
+
+       String minOpt = node.getAttribute("min");
+       if (minOpt != null) {
+           try {
+               int minMust = Integer.parseInt(minOpt);
+               query.setMinimumNumberShouldMatch(minMust);
+           } catch (Exception ex) {
+
+           }
+       }
+
         Node child = node.getFirstChild();
         while (child != null) {
             if (child.getNodeType() == Node.ELEMENT_NODE) {
```

dizzzz:
add support for minimum-must-match on BooleanQuery
added tests
fixes #392

can one help with documentation?

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/XMLToQuery.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/XMLToQuery.java
@@ -338,6 +338,18 @@ public class XMLToQuery {
 
     private Query booleanQuery(String field, Element node, Analyzer analyzer, QueryOptions options) throws XPathException {
         BooleanQuery query = new BooleanQuery();
+
+        // Specifies a minimum number of the optional BooleanClauses which must be satisfied.
+        String minOpt = node.getAttribute("min");
+        if (minOpt != null) {
+            try {
+                int minMust = Integer.parseInt(minOpt);
+                query.setMinimumNumberShouldMatch(minMust);
+            } catch (NumberFormatException ex) {
+                // ignore
+            }
+        }
+
         Node child = node.getFirstChild();
         while (child != null) {
             if (child.getNodeType() == Node.ELEMENT_NODE) {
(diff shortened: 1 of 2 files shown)


duncdrum:
docs should go to http://www.exist-db.org/exist/apps/doc/lucene.xml?field=all&id=D3.15.71.13#describe

dizzzz:
hold a moment... occur="optional" is not correct... it must be "should" .... RTFM :-)

duncdrum:
Hmm not sure what to make of the MacOS deadlock. It happened at least twice in a row.

duncdrum:
@dizzzz maybe you could rebase?

dizzzz:
@dependabot-bot rebase

duncdrum:
And again, same spot. @adamretter @wolfgangmm ant ideas why lucene  would go on  strike  on MacOS only here?

dizzzz:
strange

PR #3137, opened 2019-12-03
(4.x.x) Switch to script drive before calling subsequent commands

reinhapa:
### Description:
Switches to the same drive as the script location in case of calling a script from an other drive. This fixes the command file not found error due to try to invoke subsequent calls on the wrong drive.

### DIFF
--- a/tools/yajsw/bin/installService.bat
+++ b/tools/yajsw/bin/installService.bat
@@ -1,4 +1,4 @@
-cd %~dp0
+cd /D %~dp0
 call setenv.bat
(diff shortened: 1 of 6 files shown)


tools/yajsw/bin/installService.bat
```diff
@@ -1,4 +1,4 @@
-cd %~dp0
+cd /D %~dp0
```
dizzzz: today I learned....

adamretter:
@reinhapa Looks good. You might want to also send this upstream - https://sourceforge.net/projects/yajsw/

Issue #1960, opened 2018-06-19
wrapping nodes using for-each after an arrow-operator


line-o:
### What is the problem

Wrapping nodes in a function used in `fn:for-each` after an Arrow Operator incorrectly outputs one element per item in the source sequence.

```xquery
let $xml := <root><i/><i/><i/></root>
return
    $xml/i => for-each(function ($i) { <wrap>{$i}</wrap> })
```

returns

```xml
<wrap><i/><i/><i/></wrap>
<wrap><i/><i/><i/></wrap>
<wrap><i/><i/><i/></wrap>
```

### What did you expect

```xml
<wrap><i/></wrap>
<wrap><i/></wrap>
<wrap><i/></wrap>
```

### Describe how to reproduce or add a test

2 tests:
- *`ao:wrap-atomic-sequence`* does work
- *`ao:wrap-element-sequence`* outputs 9 instead of 3

```xquery
xquery version "3.1";

module namespace ao="http://exist-db.org/xquery/test/arrowop";
declare namespace test="http://exist-db.org/xquery/xqsuite";

declare
	%private
function ao:wrap ($item as item()) as item() { 
    <wrap>{$item}</wrap>
};

declare
	%private
function ao:wrap-explicit-type-conversion ($item as item()) as item() { 
    <wrap>{xs:string($item)}</wrap>
};

declare
	%private 
function ao:first-element-text ($s as node()*) as text() {
    $s[1]/text()
};

declare
    %test:assertEquals("1")
function ao:wrap-atomic-sequence () {
    (1, 2, 3) => for-each(ao:wrap#1) => ao:first-element-text()
};

declare
    %test:assertEquals("1")
function ao:wrap-atomic-sequence-with-explicit-type-conversion () {
    (1, 2, 3) => for-each(ao:wrap-with-explicit-type-conversion#1) => ao:first-element-text()
};

declare
	%private 
function ao:get-i-elements ($i as item()) { $i//i };

declare
    %test:assertEquals(3)
function ao:wrap-element-sequence () {
    let $xml := <root><i/><i/><i/></root>
    return $xml/node()
        => for-each(ao:wrap#1)
        => for-each(ao:get-i-elements#1)
        => count()
};
```

### Context information

- Tested on 4.1.0 and 5.0.0-RC1 (8339007319a720c97733c1c35b30d6e20ff41c9d) 
- java version "1.8.0_152"
- Operating system: MacOs (64 bit)
- both are clones from GitHub
[...]

PR #3163, opened 2019-12-16
[bugfix] root node should use supplied context if present

adamretter:
Gets us another 4 % in the XQTS 3.1

PR #3167, opened 2019-12-17
Tests for context item in functions

adamretter:
These tests show that eXist-db is incorrectly setting the context item inside both static and dynamic function calls.

According to https://www.w3.org/TR/xquery-31/#id-eval-function-call when evaluating the body of a function:
> The focus (context item, context position, and context size) is absent

Unfortunately these tests show otherwise :-(

The issue is predominantly with `RootNode.java` which always establishes a context item, i.e. it is unaware of when it is inside a `UserDefinedFunction.java`.

I am not sure of the best way to fix this? I can see two options at the moment:
1) we make `RootNode.java` aware that it is being called inside a function, and have it always throw `XPDY0002` in that case.
2) we modify `XQueryContext` so that the "focus (context item, context position, and context size)" is stack based and gets pushed down in `UserDefinedFunction` before the body is eval'd, so that it is unavailable to the RootNode.

@wolfgangmm thoughts please? it would be good to discuss it. Some of your XQSuite tests for the Range Index (`extensions/indexes/range/src/test/xquery/range/range.xql`) are relying on this incorrect behaviour, but I think they could be rewritten to start with `fn:collection(...)`.

In addition in `UserDefinedFunction`, the function body is evaluated like so:
```java
result = body.eval(contextSequence, contextItem);
```

I suspect that should also be adjusted to:
```java
result = body.eval(null, null);
```

### DIFF
--- a/exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
+++ b/exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
@@ -1,0 +1,90 @@
+package org.exist.xquery;
+
+import com.evolvedbinary.j8fu.Either;
+import org.exist.EXistException;
+import org.exist.security.PermissionDeniedException;
+import org.exist.storage.BrokerPool;
+import org.exist.storage.DBBroker;
+import org.exist.test.ExistEmbeddedServer;
+import org.exist.xquery.value.Sequence;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import static com.evolvedbinary.j8fu.Either.Left;
+import static com.evolvedbinary.j8fu.Either.Right;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class ContextItemTests {
+
+    @ClassRule
+    public static final ExistEmbeddedServer server = new ExistEmbeddedServer(true, true);
+
+    @Test
+    public void staticFunctionNoContextItem() throws EXistException, PermissionDeniedException {
+        final String query =
+                "declare function local:x() {" +
+                "  //x" +
+                "};" +
+                "local:x()";
+
+        final Either<XPathException, Sequence> result = executeQuery(query);
+        assertXQueryError(ErrorCodes.XPDY0002, result);
+    }
+
+    @Test
+    public void staticFunctionNoInheritContextItem() throws EXistException, PermissionDeniedException {
+        final String query =
+                "declare function local:x() {" +
+                "  //x" +
+                "};" +
+                "<doc><x>1</x><x>2</x></doc>/local:x()";
+
+        final Either<XPathException, Sequence> result = executeQuery(query);
+        assertXQueryError(ErrorCodes.XPDY0002, result);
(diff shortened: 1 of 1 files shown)


duncdrum:
Phew this is a big one. Because of the amount of potential breakage i d put this on the 6.0.0 milestone (which might now happen rather soon)

adamretter:
@duncdrum Do you see a potential for much breakage outside of broken XQSuite tests? I think it is probably just XQTS tests that might break and in the exist codebase it seems to be just those range index ones.

duncdrum:
I think that lots of folks might have xqsuite tests involving range indexes. Also god knows how many local functions rely on the faulty behavior. 

Does not scream 5.1.2 to me.

adamretter:
@duncdrum tests on range indexes are fine.

It's the tests in exist for range indexes which start without an explicit context item that are the problem, e.g. from `extensions/indexes/range/src/test/xquery/range/range.xql`:

```
declare
    %test:args("üss")
    %test:assertEquals("Rüsselsheim")
    %test:args("ta M")
    %test:assertEquals("Almweide")
function rt:contains-string($name as xs:string) {
    //address[range:contains(name, $name)]/city/text()
};
```

The above is invalid XQuery as it starts `//` inside a function with no context item.

dizzzz:
tests are failing...

adamretter:
@dizzzz Exactly my point!

adamretter:
I am still waiting on comments from @wolfgangmm

joewiz:
@adamretter Is this still something you're hoping to include in 5.3.0?

adamretter:
@joewiz I'm still waiting for input on it

adamretter:
In todays community call, @wolfgangmm expressed a preference to fixing it with approach 1:

> we make RootNode.java aware that it is being called inside a function, and have it always throw XPDY0002 in that case.

line-o:
As @adamretter pointed out there are lots of positive tests present already.

line-o:
@adamretter setting both contextItem and contextSequence to null was also done for HOFs to fix https://github.com/eXist-db/exist/issues/1960

line-o:
RootNode.java does not need to know where it is called. As long as every Function and UserDefinedFunction call nulls the context.

adamretter:
> `RootNode.java` does not need to know where it is called. As long as every `Function` and `UserDefinedFunction` call nulls the context.

@line-o Hmm! I am not sure it is that simple... Wouldn't that possibly break things like `$sequence[my-function(.)]` and also `$sequence ! my-function(.)`?

joewiz:
@adamretter I see we have Wolfgang's agreement on the approach and Dannes' approval, but there are some remaining CI failures. Can this be merged, or is additional work needed to get passing CI results? I ask because I recall that this PR needs to be resolved in order for https://github.com/eXist-db/exist/pull/3163.

adamretter:
@joewiz I believe @line-o took over the context item stuff in #3163

line-o:
this PR is superseded by #3605

PR #3178, opened 2019-12-30
Misc code improvements

dizzzz:
End-of-year Code cleanup, mechanical refactoring mostly.
Happy 2020 !

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/NativeBroker.java
+++ b/exist-core/src/main/java/org/exist/storage/NativeBroker.java
@@ -175,8 +175,7 @@ public class NativeBroker extends DBBroker {
 
     private final NodeProcessor nodeProcessor = new NodeProcessor();
 
-    private final IEmbeddedXMLStreamReader streamReader = null;
-    private IEmbeddedXMLStreamReader streamReaderNG = null;
+    private IEmbeddedXMLStreamReader streamReader;
 
     private final LockManager lockManager;
     private final Optional<JournalManager> logManager;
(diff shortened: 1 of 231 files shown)


dizzzz:
@adamretter ? @wolfgangmm ?

adamretter:
@dizzzz I am slowly working my way through it... the change-set is huge ;-)

exist-core/src/main/java/org/exist/xupdate/Modification.java
```diff
@@ -206,12 +205,12 @@ protected void declareVariables(XQueryContext context) throws XPathException {
...
-			context.declareNamespace(
-				entry.getKey(),
-				entry.getValue());
-		}
+        for (Map.Entry<String, String> stringStringEntry : namespaces.entrySet()) {
+            entry = stringStringEntry;
+            context.declareNamespace(
+                    entry.getKey(),
+                    entry.getValue());
+        }
```
adamretter: I think this needs tidying up. `entry` is redundant, and then `stringStringEntry` should be renamed as `entry`.

exist-core/src/main/java/org/exist/storage/NativeBroker.java
```diff
@@ -177,9 +173,9 @@
 
     private final Runtime run = Runtime.getRuntime();
 
-    private NodeProcessor nodeProcessor = new NodeProcessor();
+    private final NodeProcessor nodeProcessor = new NodeProcessor();
 
-    private IEmbeddedXMLStreamReader streamReader = null;
+    private final IEmbeddedXMLStreamReader streamReader = null;
```
adamretter: So `streamReader` will always be `null`?

seems that this should be removed, and `streamReaderNG` renamed to `streamReader`

exist-core/src/main/java/org/exist/xupdate/Modification.java
dizzzz: ah... good catch; i will fix

dizzzz:
Thnx @adamretter  waiting for build to finish....

dizzzz:
image

dizzzz:
All green !

PR #3182, opened 2020-01-04
Multi-valued hierarchical facets

wolfgangmm:
Extend hierarchical facets to accept multiple values passed in via an array.

Defining a hierarchical facet on a node, it is possible that a node is a member of multiple hierarchies at the same time. For example, a text could fall into several hierarchical subject categories like science/math or humanities/history. Currently this is not supported: we only allow a single hierarchical facet per dimension.

Multiple hierarchies can be easily described as an XQuery array of sequences like `[('science', 'math'), ('humanities', 'history')]`. This PR extends the facet indexer to check if the result of the XQuery expression to obtain the facet value is an array. If it is, it is treated as a multi-valued hierarchical facet.

### Type of tests:

See extended xqsuite test.

### Documentation

https://github.com/eXist-db/documentation/commit/148bf8a8636bb16e55c4757b63f36d263c78f772

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
@@ -29,11 +29,15 @@ import org.exist.security.PermissionDeniedException;
 import org.exist.storage.DBBroker;
 import org.exist.util.DatabaseConfigurationException;
 import org.exist.xquery.XPathException;
+import org.exist.xquery.functions.array.ArrayType;
 import org.exist.xquery.value.Sequence;
 import org.exist.xquery.value.SequenceIterator;
+import org.exist.xquery.value.Type;
 import org.w3c.dom.Element;
 
 import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 
 /**
@@ -62,7 +66,7 @@ public class LuceneFacetConfig extends AbstractFieldConfig {
                 (hierarchicalOpt.equalsIgnoreCase("true") || hierarchicalOpt.equalsIgnoreCase("yes"));
 
         config.facetsConfig.setHierarchical(dimension, isHierarchical);
-        config.facetsConfig.setMultiValued(dimension, !isHierarchical);
+        config.facetsConfig.setMultiValued(dimension, true);
     }
 
     @Nonnull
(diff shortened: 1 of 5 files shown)


extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
```diff
@@ -92,6 +98,19 @@ protected void processResult(Sequence result, Document luceneDoc) throws XPathEx
         }
     }
 
+    private void createHierarchicalFacet(Document luceneDoc, Sequence seq) throws XPathException {
+        final List<String> paths = new ArrayList<>(seq.getItemCount());
+        for (SequenceIterator i = seq.unorderedIterator(); i.hasNext(); ) {
+            final String value = i.nextItem().getStringValue();
+            if (value.length() > 0) {
```
dizzzz: functional identical to !String#isEmpty() ; the current is faster for sure.
Is there a NPE risk?

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
```diff
@@ -92,6 +98,19 @@ protected void processResult(Sequence result, Document luceneDoc) throws XPathEx
         }
     }
 
+    private void createHierarchicalFacet(Document luceneDoc, Sequence seq) throws XPathException {
+        final List<String> paths = new ArrayList<>(seq.getItemCount());
```
adamretter: As you eventually need an array and the size of the array is known, it would be much more efficient to just use an array rather instead of an array list.

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
wolfgangmm: In line with XQuery specs, `getStringValue()` on an item should not return null.

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
wolfgangmm: Previously an array was used, but this caused an exception being thrown by lucene if `getStringValue()` returned an empty string (which is a valid value in XQuery but rejected by lucene). We thus have to build the list dynamically.

wolfgangmm:
As @tuurma just pointed out to me, this PR needs to be slightly extended to be really useful, so please do not merge yet. I'll update it today or tomorrow.

wolfgangmm:
Ok, I added one more commit to also cover the query side of things and am done now. Documentation will be coming...

tuurma:
accompanying documentation PR now in, ready to merge

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
joewiz: @adamretter Could you please mark this and your requested changes as resolved (assuming they are)?

dizzzz:
image

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
adamretter: @wolfgangmm Sure... but why not:

```java
final String[] paths = new String[seq.getItemCount()];
int i = 0;
for (final SequenceIterator i = seq.unorderedIterator(); i.hasNext(); ) {
  final String value = i.nextItem().getStringValue();
  if (!value.isEmpty() > 0) {
    paths[i++] = value;
  }
}
if (i > 0) {
  luceneDoc.add(new FacetField(dimension, Arrays.copyOf(paths, i)));
}
```

adamretter: @joewiz it's not resolved yet ;-)

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
adamretter: I think @dizzzz is right here.

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneFacetConfig.java
joewiz: @wolfgangmm Do you agree with @adamretter's suggestion? I've resolved the merge conflict, so this is the only remaining item before the PR can be merged.

wolfgangmm: We cannot have empty items in the array. i'm not sure if the suggestion considers this,  but anyway, I absolutely don't see any benefit in this formulation. It's not that this code is critical in any way.

joewiz: Excellent, thanks. We’ll move ahead as is.

PR #3185, opened 2020-01-05
Porter stemming code update

dizzzz:
- sync from author
- performance update by using native array copy


joewiz:
@dizzzz I'm curious, is porter stemming exposed as an XQuery function, or is it just a java utility?

dizzzz:
I realized after it seems not be be used by any code !?

adamretter:
@dizzzz I think this is old code that got left behind. I think it dates from way back before I did the first eXist-db <-> Lucene integration.

PR #3186, opened 2020-01-05
FastStringBuffer improvement

dizzzz:
Improved performance by using native array copy
Synced code from saxonHE
Added tests

### DIFF
--- a/exist-core/src/main/java/org/exist/util/FastStringBuffer.java
+++ b/exist-core/src/main/java/org/exist/util/FastStringBuffer.java
@@ -7,19 +7,19 @@
  * modify it under the terms of the GNU Lesser General Public License
  * as published by the Free Software Foundation; either version 2
  * of the License, or (at your option) any later version.
- *  
+ *
  * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


exist-core/src/main/java/org/exist/util/FastStringBuffer.java
```diff
@@ -471,12 +472,12 @@ public String getNormalizedString( int mode ) {
      *@param  sb    Description of the Parameter
      *@param  mode  Description of the Parameter
      *@return       The normalizedString value
-     */    
+     */
     public StringBuffer getNormalizedString(StringBuffer sb, int mode ) {
```
adamretter: Should this be changed to use `StringBuilder` rather than String buffer?

dizzzz: the classname is Fast...StringBuffer :-) but it makes sense in a way, I will check the users of the method.

dizzzz:
@adamretter reworked PR, added some additional improvements that might yield in less overhead (less code execution, stack usage, unneeded conversation to objects etc)

adamretter:
Excellent stuff!

Issue #3148, opened 2019-12-06
java.lang.ClassCastException with fn:subsequence


joewiz:
### What is the problem

The following query raises an error:

```xquery
("x", "y") => subsequence(1, 1) => matches("x")
```

The error:

```text
2019-12-06 14:08:51,839 [qtp1729081400-44] ERROR (XQueryServlet.java [process]:559) - org.exist.xquery.value.StringValue cannot be cast to org.exist.xquery.value.NodeValue 
java.lang.ClassCastException: org.exist.xquery.value.StringValue cannot be cast to org.exist.xquery.value.NodeValue
	at org.exist.xquery.value.SubSequence.isPersistentSet(SubSequence.java:377) ~[exist-core-5.1.0.jar:5.1.0]
	at org.exist.xquery.functions.fn.FunMatches.eval(FunMatches.java:322) ~[exist-core-5.1.0.jar:5.1.0]
...
```

### What did you expect

I expected the query to return `true()` instead of an error.  

The following slight variants of the query do return the expected result:

```xquery
("x", "y") => subsequence(1, 1) => string() => matches("x")
```

```xquery
("x", "y") => head() => matches("x")
```

The problem seems specific to `fn:subsequence()` and its handling of data types.

### Describe how to reproduce or add a test

```xquery
xquery version "3.1";

module namespace t="http://exist-db.org/xquery/test";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare
    %test:assertTrue
function t:test() {
    ("x", "y") => subsequence(1, 1) => matches("x")
};
```

This test returns:

```xml
<testcase name="test" class="t:test">
    <error type="java:java.lang.ClassCastException"
        message="org.exist.xquery.value.StringValue cannot be cast to org.exist.xquery.value.NodeValue"
    />
</testcase>
```

### Context information

- eXist-db version + Git Revision hash: every release of eXist since 4.5.0  
- Java version: 1.8.0_232
- Operating system: macOS 10.15.1
- 32 or 64 bit: 64 bit
- How is eXist-db installed? built develop from source, tested previous releases with DMG download
- Any custom changes in e.g. conf.xml: none

adamretter:
@joewiz would you be able to try this patch?

```java
diff --git a/exist-core/src/main/java/org/exist/xquery/value/SubSequence.java b/exist-core/src/main/java/org/exist/xquery/value/SubSequence.java
index 03183592aa..acc43c1c78 100644
[...]

PR #3142, opened 2019-12-05
Lucene index fixes

wolfgangmm:
In specific cases a wrong lucene index config (potentially belonging to a different collection) was used when storing a document into a collection which did not define a lucene index at all. We need to make sure the config is cleared and set to an empty default before the worker is reused for storing another document. The same applies to the range index.

PR #3199, opened 2020-01-15
Fix cast exception in fn:subsequence

adamretter:
Closes https://github.com/eXist-db/exist/issues/3148


joewiz:
Looks like Appveyor's Windows/jdk11 build timed out here, just as it did in https://github.com/eXist-db/exist/pull/3142#issuecomment-574676125...

Issue #3193, opened 2020-01-13
[BUG] stats.dbx


pwin:
**Describe the bug**
An error being reported when starting exist-db 5.1.1 on Windows 10 using Azul Java v11
```
13 Jan 2020 12:55:22,350 [main] INFO  (JettyStart.java [run]:200) - Configuring eXist from C:\exist-db\exist-db.5.1.1\db\etc\conf.xml
org.exist.storage.btree.DBException: Error while writing C:\exist-db\exist-db.5.1.1\data\stats.dbx: C:\exist-db\exist-db.5.1.1\data\stats.dbx
        at org.exist.storage.statistics.IndexStatistics.sync(IndexStatistics.java:114)
        at org.exist.indexing.IndexManager.sync(IndexManager.java:264)
        at org.exist.storage.NativeBroker.sync(NativeBroker.java:3710)
        at org.exist.storage.BrokerPool.sync(BrokerPool.java:1457)
        at org.exist.storage.BrokerPool._initialize(BrokerPool.java:655)
        at org.exist.storage.BrokerPool.initialize(BrokerPool.java:438)
        at org.exist.storage.BrokerPools.configure(BrokerPools.java:174)
        at org.exist.storage.BrokerPools.configure(BrokerPools.java:109)
        at org.exist.jetty.JettyStart.run(JettyStart.java:205)
        at org.exist.jetty.JettyStart.main(JettyStart.java:95)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.exist.start.Main.invokeMain(Main.java:118)
        at org.exist.start.Main.runEx(Main.java:248)
        at org.exist.start.Main.run(Main.java:123)
        at org.exist.start.Main.main(Main.java:60)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #3201, opened 2020-01-15
Make sure to create the stats index before writing if it does not exist

adamretter:
Closes https://github.com/eXist-db/exist/issues/3193


dizzzz:
shouldn't the conf.xml  be generated?

adamretter:
@dizzzz erm... Nope the conf.xml should be static

PR #3208, opened 2020-01-18
Option to make Document locks hierarchical

adamretter:
This adds a new option to `conf.xml`:

```xml
<document use-path-locks="false"/>
```

When this is enabled, document and collection locks participate in the same locking hierarchy. This means that it becomes impossible for deadlocks to occur between documents and collections.

This should likely be enabled by default in future.

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/lock/LockManager.java
+++ b/exist-core/src/main/java/org/exist/storage/lock/LockManager.java
@@ -31,7 +31,6 @@ import uk.ac.ic.doc.slurp.multilock.MultiLock;
 
 import java.util.Arrays;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
 
 /**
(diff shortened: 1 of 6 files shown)


exist-core/src/main/java/org/exist/storage/lock/LockManager.java
```diff
@@ -457,25 +488,33 @@ DocumentLock getDocumentLock(final String documentPath) {
...
+            final long groupId = System.nanoTime();
+            final String path = documentPath.toString();
 
-        final DocumentLock lock = getDocumentLock(path);
-        try {
+            final MultiLock lock = getDocumentLock(path);
             lockTable.attempt(groupId, path, LockType.DOCUMENT, Lock.LockMode.READ_LOCK);
 
-            lock.readLock().lockInterruptibly();
+            if (lock(lock, Lock.LockMode.READ_LOCK)) {
```
dizzzz: lock lock lock :-) not so readable....

adamretter:
@duncdrum docs - https://github.com/eXist-db/documentation/pull/426

adamretter:
@duncdrum can we get this one merged now?

PR #3223, opened 2020-01-23
Pass args to release plugin

adamretter:
Small fix to release process


dizzzz:
@duncdrum i think the first arguments are used by maven, the second one are parameters to a specific module.... #guess

adamretter:
@dizzzz That's a Bingo!

duncdrum:
I still don’t get it, why do we pass the same arguments to haben and then again to a plug-in. What does the release plugin do with docker true?

PR #3228, opened 2020-01-25
[fix] map:remove sequence of keys

### LINKED ISSUE #1656: Align map:remove with 3.1 F&O spec to accept zero-or-more keys
### What is the problem

eXist's implementation of `map:remove` allows only a single key:

> `map:remove($map as map, $key as xs:string)`

### What did you expect

The XQuery 3.1 F&O spec says that `map:remove` can take 0 or more keys:

> `map:remove($map as map(*), $keys as xs:anyAtomicType*)`

See https://www.w3.org/TR/xpath-functions-31/#func-map-remove.

Besides the cardinality difference, note that keys can be of `xs:anyAtomicType`, not just `xs:string`.

This example is taken from the spec at the link above:

```xquery
xquery version "3.1";

let $week := map{0:"Sonntag", 1:"Montag", 2:"Dienstag",
       3:"Mittwoch", 4:"Donnerstag", 5:"Freitag", 6:"Samstag"}
return
    map:remove($week, (0, 6 to 7)) 
```

BaseX and Saxon return the expected result: 

```xquery
map{1:"Montag",2:"Dienstag",3:"Mittwoch",4:"Donnerstag",5:"Freitag"}
```

eXist returns the following error:

> err:XPTY0004 checking function parameter 2 in call map:remove($week, untyped-value-check[xs:string, ( 0, 6 to 7 )]): XPTY0004: The actual cardinality for parameter 2 does not match the cardinality declared in the function's signature: map:remove($map as map, $key as xs:string) map. Expected cardinality: exactly one, got 3. 

### Context information

- eXist-db version + Git Revision hash: eXist 3.7.0-SNAPSHOT+9091c3032
- Java version: Oracle JDK 1.8.0_152
- Operating system: macOS 10.13.1
- 32 or 64 bit: 64 bit

line-o:
### Description:

AbstractMapType.remove and all overrides now expect an array of
AtomicValues.

MapFunction.remove was adapted to take that into account.

### Reference:

fixes #1656 

### Type of tests:

Add tests to XQsuite to ensure sequences of keys can be deleted and
the changes do not break behaviour for non existing keys.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapFunction.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapFunction.java
@@ -145,9 +145,16 @@ public class MapFunction extends BasicFunction {
         return null;
     }
 
-    private Sequence remove(final Sequence[] args) {
+    private Sequence remove(final Sequence[] args) throws XPathException {
         final AbstractMapType map = (AbstractMapType) args[0].itemAt(0);
-        return map.remove((AtomicValue) args[1].itemAt(0));
+
+        final int length = args[1].getItemCount();
+        final AtomicValue[] keys = new AtomicValue[length];
+
+        for (int i = 0; i < length; i++) {
+            keys[i] = (AtomicValue) args[1].itemAt(i);
+        }
+        return map.remove(keys);
     }
 
     private Sequence keys(final Sequence[] args) {
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/SingleKeyMapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/SingleKeyMapType.java
@@ -1,5 +1,6 @@
 package org.exist.xquery.functions.map;
 
+import com.github.krukow.clj_lang.IPersistentMap;
 import org.exist.xquery.Constants;
 import org.exist.xquery.XPathException;
 import org.exist.xquery.XQueryContext;
@@ -59,12 +60,18 @@ public class SingleKeyMapType extends AbstractMapType {
     }
 
     @Override
-    public AbstractMapType remove(AtomicValue key) {
-        try {
+    public AbstractMapType remove(final AtomicValue[] keysAtomicValues) throws XPathException {
+        for (final AtomicValue key: keysAtomicValues) {
+            if (key != this.key) {
+                continue;
+            }
+            this.key = null;
+            value = null;
             return new MapType(context);
-        } catch (final XPathException e) {
-            return null;
         }
+        final MapType map = new MapType(context);
+        map.add(this);
+        return map;
     }
 
     @Override
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/AbstractMapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/AbstractMapType.java
@@ -54,7 +54,7 @@ public abstract class AbstractMapType extends FunctionReference
 
     public abstract Sequence keys();
 
-    public abstract AbstractMapType remove(AtomicValue key);
+    public abstract AbstractMapType remove(AtomicValue[] keysAtomicValues) throws XPathException;
 
     public abstract int getKeyType();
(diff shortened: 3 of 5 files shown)


exist-core/src/main/java/org/exist/xquery/functions/map/MapFunction.java
```diff
@@ -145,9 +145,13 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
         return null;
     }
 
-    private Sequence remove(final Sequence[] args) {
+    private Sequence remove(final Sequence[] args) throws XPathException {
         final AbstractMapType map = (AbstractMapType) args[0].itemAt(0);
-        return map.remove((AtomicValue) args[1].itemAt(0));
+        AtomicValue[] keys = new AtomicValue[args[1].getItemCount()];
```
dizzzz: Could be replaced with nice lambda/collector expression?

line-o: @dizzzz Do you have a link to an example?

line-o: I had a quick look around and I cannot see how this could be applied here.

exist-core/src/main/java/org/exist/xquery/functions/map/SingleKeyMapType.java
```diff
@@ -59,12 +60,16 @@ public int size() {
     }
 
     @Override
-    public AbstractMapType remove(AtomicValue key) {
-        try {
+    public AbstractMapType remove(AtomicValue[] keysAtomicValues) throws XPathException {
+        for (final AtomicValue key: keysAtomicValues) {
```
dizzzz: I do not understand the logic here :-)

line-o: For some reason Maps with just one key have their own java representation.
`let $a := map { 'a': 1 }` 
this implementation ensures that `map:remove($a, ('z','zz','zzz'))`, `map:remove($a, ('z','a'))` and `map:remove($a, 'a')` behave as expected.

exist-core/src/main/java/org/exist/xquery/functions/map/MapFunction.java
```diff
@@ -145,9 +145,16 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
         return null;
     }
 
-    private Sequence remove(final Sequence[] args) {
+    private Sequence remove(final Sequence[] args) throws XPathException {
         final AbstractMapType map = (AbstractMapType) args[0].itemAt(0);
-        return map.remove((AtomicValue) args[1].itemAt(0));
+
+        final int length = args[1].getItemCount();
+        AtomicValue[] keys = new AtomicValue[length];
```
adamretter: Could be final.

exist-core/src/main/java/org/exist/xquery/functions/map/SingleKeyMapType.java
```diff
@@ -59,12 +60,16 @@ public int size() {
     }
 
     @Override
-    public AbstractMapType remove(AtomicValue key) {
-        try {
+    public AbstractMapType remove(AtomicValue[] keysAtomicValues) throws XPathException {
```
adamretter: Parameter could be final

exist-core/src/main/java/org/exist/xquery/functions/map/SingleKeyMapType.java
```diff
@@ -59,12 +60,16 @@ public int size() {
...
+        for (final AtomicValue key: keysAtomicValues) {
+            if (key != this.key) { continue; }
```
adamretter: Multi-line.

joewiz:
Excellent! Thanks @line-o, and thanks, @dizzzz and @adamretter!

PR #3159, opened 2019-12-13
(4.x.x) Add a test to assert the integrity of installer/jobs.xml

ohumbel:
This PR adds a long promised test which applies to the `develop-4.x.x` branch only.

See discussion on https://github.com/eXist-db/exist/pull/2896.

PR #3239, opened 2020-02-02
WIP: Add a test to assert the integrity of exist-installer/src/main/izpack/jobs.xml

ohumbel:
PR #3159 should be ported to `develop`

### DIFF
--- a/exist-core/src/test/java/org/exist/installer/JobsXmlConfigurationTest.java
+++ b/exist-core/src/test/java/org/exist/installer/JobsXmlConfigurationTest.java
@@ -1,0 +1,9 @@
+import org.junit.Test;
+
+public class JobsXmlConfigurationTest {
+
+    @Test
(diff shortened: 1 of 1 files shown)


ohumbel:
The situation has changed from `develop-4.x.x` to `develop`. The entries in `jobs.xml` used to refer to existing files in the checked out tree. On the current `develop` branch, the entries refer to folders and files which - hopefully - will be assembled later to those places. In other words, `$INSTALL_PATH` used to be the root of the project. But now, at the time of the test execution, it is not known any more. After `mvn package`, the latest `exist-distribution-5.3.0-SNAPSHOT-dir` would be the `$INSTALL_PATH`.

So is it worth analyzing paths which are not present?

adamretter:
@ohumbel probably not. Thanks I will merge your other PR.

ohumbel:
@adamretter Good decision IMHO - thanks for reviewing!

PR #3258, opened 2020-03-01
Resolve XSLT Modules from EXPath Packages

adamretter:
When XSLT Modules are stored in EXPath Packages, they were not previously usable by eXist-db's XSLT processor (Saxon). When using `xsl:import` or `xsl:include` they are now resolved correctly from the EXPath Pkg repository.

@duncdrum I am not sure which issues are involved, but I think this should solve your EXPath Pkg XSLT woes.


exist-core/src/main/java/org/exist/validation/resolver/eXistXMLCatalogResolver.java
```diff
@@ -49,6 +49,10 @@
  */
 public class eXistXMLCatalogResolver extends XMLCatalogResolver {
 
+    //TODO(AR) refactor to org.apache.xml.resolver.tools.CatalogResolver which is reused within here for Xerces
```
dizzzz: I'd suggest to have a look at Norman's recent work on Catalog resolvers and extend/contribute to that initiative.

Issue #3231, opened 2020-01-28
Hangup in getting permissions in an AVT?


xatapult:
I run the following XQuery script (I minimized it as much as possible). eXist eats my CPU up to 100%. I have to force shutdown the database.

```xquery
xquery version "3.1" encoding "UTF-8";

declare option exist:serialize "method=xml media-type=text/xml indent=yes";

<TEST>
{
  let $path as xs:string :='/db/TEST' 
  return
    <result>
      <collection name="{$path}" created="{xmldb:created($path)}" 
        owner-name="{sm:get-permissions(xs:anyURI($path))/sm:permission/@owner}"/>
    </result>       
}    
</TEST>
```

The cause is in the AVT for the `owner-name` attribute. But when I put that expression outside an AVT it simply works.

**Expected behavior**
The requested owner name in the attribute. Or: some error message about what I did wrong. But not a hangup

**To Reproduce**
Create a `/db/TEST` collection. Run the script above as admin.

**Context (please always complete the following information):**
 - OS: Windows 10
 - eXist-db version: 5.2.0
 - Java Version 1.8.0_202-b08

**Additional context**
- How is eXist-db installed? jar
- Any custom changes in e.g. `conf.xml`? no

xatapult:
I tried doing this through REST and and through oXygen, both the same results

adamretter:
@xatapult Thanks Erik, I can see and reproduce the issue.

It's quite impressive! I have managed to reduce it to this:

```xquery
<outer>
	<inner perm="{sm:get-permissions(xs:anyURI("/db"))/sm:permission/@owner}"/>
</outer>
```

I suspect this is a regression in 5.2.0 over 5.1.1.

xatapult:
I have reasons to suspect it was in 4.7 already...

joewiz:
Sorry, what's an AVT?

xatapult:
ATtribute Value Template, informally speaking the {...} to set an attribute's value with an XPath expression

joewiz:
Ah, I see - I found the term [defined](https://www.w3.org/TR/xslt-30/#attribute-value-templates) in the XSLT spec. How strange that it's significant for this bug that the expression is [enclosed](https://www.w3.org/TR/xquery-31/#id-enclosed-expr)!

adamretter:
I have a fix which involved not reusing the XQueryContext's Memtree builder:

```
[...]

PR #3262, opened 2020-03-02
Make sure the Memtree in the query context is correctly used

adamretter:
Closes https://github.com/eXist-db/exist/issues/3231

This prevents various nested Memtree operations from causing a query to hang.


joewiz:
@adamretter Looks like this fixes many more cases where this would've cropped up than were originally reported. Excellent!

Issue #1849, opened 2018-05-01
Bugs in integer subtype checks and casting


joewiz:
### What is the problem

When performing xs:positiveInteger and xs:negativeInteger type checks on values, the checks miss errors. 

### What did you expect

I expected type checks to be enforced.

### Describe how to reproduce or add a test

**Test 1**

Given following query:

```xquery 
-1 cast as xs:positiveInteger
```

- eXist returns `-1`.
- Saxon returns:
    > Value -1 cannot be converted to integer subtype xs:positiveInteger
- BaseX returns:
    > [FORG0001] Cannot convert xs:integer to xs:positiveInteger: -1.

**Test 2**

Given the following query:

```xquery
-1 instance of xs:positiveInteger
```

eXist correctly returns `false()` (as do Saxon and BaseX).

**Test 3**

Given the following query:

```xquery
xquery version "3.1";

declare function local:test($input as xs:positiveInteger) {
    $input
};

local:test(-1)
```

- eXist returns `-1`. 
- Saxon returns:
    > Required item type of first argument of local:test() is xs:positiveInteger; supplied value (-1) has item type xs:integer
- BaseX returns: 
    > [XPTY0004] Cannot convert xs:integer to xs:positiveInteger: $input := -1. 

**Test 4**

The opposite of test 1 is also wrong in eXist but correct in Saxon and BaseX.

```xquery
1 cast as xs:negativeInteger
```

- eXist returns `1`. 

**Test 5**

The opposite of test 2 is also correct in eXist, Saxon, and BaseX.

```xquery
1 instance of xs:negativeInteger
```

- All return `false()`.

**Test 6**

The opposite of test 3 is also wrong in eXist but correct in Saxon and BaseX.

```xquery
xquery version "3.1";

declare function local:test($input as xs:negativeInteger) {
    $input
};

local:test(1)
```

- eXist returns `1`.

### Context information

- eXist-db version + Git Revision hash: eXist 4.2.0-SNAPSHOT+201804241640 develop 30c5c331d
- Java version: 1.8.0_172-b11
- Operating system: macOS 10.13.4
- 32 or 64 bit: 64 bit
- Any custom changes in e.g. conf.xml: none

duncdrum:
*ouch* especially test6

joewiz:
[...]

PR #3266, opened 2020-03-03
Fix type conversion of xs:positiveInteger and xs:negativeInteger types 

adamretter:
Closes https://github.com/eXist-db/exist/issues/1849


joewiz:
@adamretter Thank you! I bet this helps with XQTS conformance too.

Issue #3024, opened 2019-09-23
Restoring a large backup can report no progress for some time.


adamretter:
As reported to me by Vincent Lizzi... When restoring a very large backup using XML-RPC remote API, there can be a very long pause before any progress at all is reported. It appears to the user that either nothing is happening or the system has crashed.

* When restoring a backup, we first create a Zip which we then copy up to the server before initiating the restore process. The creation of this zip and copying it to the server needs to be included in the restore progress feedback.

PR #3271, opened 2020-03-04
Fix reporting of Restored Documents when restoring a backup

adamretter:
Closes https://github.com/eXist-db/exist/issues/3024

* Allow the user to see the result of the restore process before dismissing the restore dialog.
* Improve/correct percentage display in Restore Dialog.
* Make sure to report Zip and Transfer process when restoring backups to a remote server, otherwise to the user the restore process can appear to be doing nothing.
* Get the correct total of files to be restored
* When skipping packages during restore (due to newer version), skipped documents must be counted too (for progress)


joewiz:
I just restarted the one failed CI build in AppVeyor.

adamretter:
@joewiz thanks all is green now :-)

Issue #3220, opened 2020-01-22
unparsed-text() for user guest seems to think it must read from the file system


xatapult:
My XQuery script performs a call to `unparsed-text()` and asks for a file *in* the database that has world read permissions. E.g. 

```
unparsed-text('/db/apps/eXide/build.js')
```

What I get is: `err:FOUT1170 non-dba user not allowed to read from file system`

(Anyway, why is the error code something with FOUT, the Dutch word for error...)

**Expected behavior**
The unparsed text of the requested file 

**To Reproduce**
* run eXide as guest user. 
* Enter ` unparsed-text('/db/apps/eXide/build.js')`  as XQuery script
* Run it: error
* Login as amin
* Run it again: Ok

**Context (please always complete the following information):**
 - OS: WIndows 10
 - eXist-db version: 5.1.1 (latest version)
 - Java Version "1.8.0_241"

**Additional context**
- How is eXist-db installed? JAR installed
- Any custom changes in e.g. `conf.xml`? No

joewiz:
I can confirm the report still affects 5.2.0-develop (current HEAD).

adamretter:
@xatapult @joewiz hmm we seem to have a test case which covers this:

```xquery
declare
    %test:assertEquals("Hello&#10;world!&#xD;Hello&#xD;&#xA;world!")
function upt:unparsed-text-from-db() {
    unparsed-text("/db/test-unparsed-text/test.txt")
};
```
See https://github.com/eXist-db/exist/blob/develop/exist-core/src/test/xquery/unparsed-text.xql#L56

joewiz:
@adamretter If the tests run as dba, they wouldn't encounter the bug.

dariok:
It seems this one goes back a bit:

I tried reading a world-readable file from the DB using `unparsed-text()` on eXist 4.7.1 and I, too, get the `FOUT1170`.
Trying to open a URL (`http://exist.ulb.tu-darmstadt.de:8080/exist/restxq/edoc/collection/arndt/nav.xml`) using `unparsed-text()` without being logged in also throws `FOUT1170`. The same on 5.1.0

Don't know whether it's related: after logging in as admin, `unparsed-text()` on 4.7.1 throws `err:FOUT1170 Server returned HTTP response code: 500 for URL: http://exist.ulb.tu-darmstadt.de:8080/exist/restxq/edoc/collection/arndt/nav.xml`
(`doc()` says the same but returns code `FODC0005`)

[...]

PR #3285, opened 2020-03-09
With fn:unparsed-text#1 non-dba users should only be restricted from filesystem

adamretter:
Closes https://github.com/eXist-db/exist/issues/3220


joewiz:
@adamretter Thank you!

PR #3290, opened 2020-03-09
fn:parse-xml-fragment should return a document-node()

adamretter:
Closes https://github.com/eXist-db/exist/issues/2181


joewiz:
@adamretter Excellent! Many thanks!

PR #3297, opened 2020-03-11
Ensure Inspect Module functions throws errors appropriately

adamretter:
Closes https://github.com/eXist-db/exist/pull/1285


adamretter:
@dizzzz Possibly. I wasn't really sure of an easy way to generate such a test case. Also as XQSuite uses all of these Inspect functions, I reasoned that there was already good test coverage.

PR #3313, opened 2020-03-19
Make fn:collection signature compliant with the XQuery 3.1 F&O spec

adamretter:
Brings the signature of fn:collection inline with the specifications.

I guess this could be regarded as a breaking API change, so might require a major release.


joewiz:
Sorry, it's a little hard for me to tell from the diff what changed about the signature. Could you summarize the before & after? Thanks!

adamretter:
@joewiz Sure...

Previously eXist-db had:
```
fn:collection($collection-uris as xs:string*) as item*
```

This PR corrects it to match the XQuery 3.1 F+O spec:
```
fn:collection() as item*
fn:collection($arg as xs:string?) as item*
```

exist-core/src/main/java/org/exist/xquery/functions/xmldb/FunXCollection.java
```diff
@@ -21,37 +21,48 @@
...
 import org.exist.xquery.value.Type;
 
+import java.net.URI;
+
+import static org.exist.xquery.FunctionDSL.*;
+import static org.exist.xquery.functions.xmldb.XMLDBModule.functionSignature;
+
 /**
- * @author wolf
- * @author ljo
```
wolfgangmm: I'd say we either remove author tags altogether or correct them where found. Just removing is against the rules of conduct.

adamretter: I removed them, because this is not the same file. I wrote this from scratch. So it is not a derivative of the original work, so it is fine.

adamretter: However, that said, if you want to remove all the `@author` declarations from source files in eXist-db then I would support that, as I don't generally regard them as a good thing.

line-o: I am in favour of no authors, but I also think this should be discussed briefly in the next call on Monday so that more people have a chance to react.

joewiz:
@adamretter Thank you for the explanation. That's very clear.

joewiz:
We now have a conflict.

exist-core/src/main/java/org/exist/xquery/functions/xmldb/FunXCollection.java
line-o: No decision was made on this issue in the previous two calls.

joewiz: As agreed in today's Community Call, contributors will henceforth remove author tags when they touch a file in a PR. Please remember to remove author tags in your PRs.

dizzzz (approved):

duncdrum:
@adamretter can you rebase this please

adamretter:
This one is now rebased and hopefully ready to merge.

exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
```diff
@@ -41,35 +40,33 @@
...
 import java.net.URISyntaxException;
-import java.util.ArrayList;
 import java.util.Iterator;
-import java.util.List;
+
+import static org.exist.xquery.FunctionDSL.*;
 
 /**
- * @author wolf
+ * @author <a href="mailto:adam@evolvedbinary.com">Adam Retter</a>
```
line-o, line 48: The author is still in here, from the discussion in this thread they should all be removed.

exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
```diff
@@ -41,35 +40,33 @@
...
+    static final FunctionSignature[] FS_COLLECTION = functionSignatures(
+            new QName(FS_COLLECTION_NAME, Function.BUILTIN_FUNCTION_NS),
+            "Returns the documents contained in the Collections specified in the input sequence. "
+                    + XMLDBModule.COLLECTION_URI + " Documents contained in sub-collections are also included. "
+                    + "If no value is supplied, the statically know documents are used; for the REST Server this could be the collection in the URI path.",
+            returnsOptMany(Type.ITEM, "The items indicated by the Collection URI"),
+            arities(
+                    arity(),
+                    arity(
+                            optParam("arg", Type.STRING,"The Collection URI")
```
line-o, line 62: Then name of the argument was changed from a descriptive name `collection-uri` to a generic one `arg`.
It is the name that is used in XQ31 functions and operators spec, but since
the upcoming XQ40 functions and operators spec now calls it `uri` I would like to see us use a more descriptive term and maybe settle for `uri`. Source https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-collection

exist-core/src/main/java/org/exist/xquery/functions/xmldb/FunXCollection.java
```diff
@@ -21,37 +21,48 @@
...
+ * @author <a href="mailto:adam@evolvedbinary.com">Adam Retter</a>
```
line-o, line 37: same here

exist-core/src/main/java/org/exist/xquery/functions/xmldb/FunXCollection.java
```diff
@@ -21,37 +21,48 @@
...
-		new FunctionParameterSequenceType("collection-uris", Type.STRING, Cardinality.ONE_OR_MORE, "The collection URIs")},
-            new FunctionReturnSequenceType(Type.NODE, Cardinality.ZERO_OR_MORE, "the document nodes from the specified collections excluding sub-collections"));
-
-	public FunXCollection(XQueryContext context) {
-		super(context, signature, false);
+
+	private static final String FS_XCOLLECTION_NAME = "xcollection";
+	static final FunctionSignature FS_XCOLLECTION = functionSignature(
+			FS_XCOLLECTION_NAME,
+			"Returns the document nodes in the collections $collection-uris " +
```
line-o, line 44: mismatch of argument name in description and signature
`collection-uris` <> `arg`
This function is not defined in the XQ31 spec but part of the xmldb namespace.

exist-core/src/main/java/org/exist/xquery/functions/xmldb/FunXCollection.java
```diff
@@ -21,37 +21,48 @@
...
+					"non-recursively, i.e. does not include document nodes found in " +
+					"sub-collections. " +
+					"C.f. fn:collection().",
+			returnsOptMany(Type.ITEM, "The items from the specified collections excluding sub-collections"),
+			params(
+				manyParam("args", Type.STRING,"The Collection URI")
```
line-o, line 50: The very generic argument name `args` should either stay `collection-uris` or use `uris` in line with the XQ40 spec.

line-o (changes requested):
Argument names should be as descriptive as possible.
Using different argument names in function description and signature is confusing to developers.  
The author discussion in this thread is not addressed.

exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
adamretter: eXist-db is not yet doing anything related to XQ40.
eXist-db targets the XQuery 3.1 currently.
The goal of this PR (as per its title) is to make this `fn:collection` compliant with the XQuery 3.1 specification, the XQuery F&O 3.1 spec names this parameter `arg` and so if we are to follow the spec, then we must do the same!

exist-core/src/main/java/org/exist/xquery/functions/xmldb/FunXCollection.java
adamretter: eXist-db targets XQuery 3.1 not 4.0 (yet).
The official 3.1 spec compliant function `fn:collection` uses the term `arg` as this is a pluralised version of that function, using `args` here makes sense.

line-o: `xmldb:xcollection` is not part of any XQuery spec.

exist-core/src/main/java/org/exist/xquery/functions/xmldb/FunXCollection.java
adamretter: Ah! I see that I had some outstanding stashed changes that I had not pushed. I have now pushed those so that the `xmldb:xcollection` function is re-aligned with the updated `fn:collection` function.

exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
adamretter: We gain compliance with the XQuery 3.1 specification!

dizzzz:
community prefers **$collection-uris as xs:string?** as this more descriptive in the function documentation, and a step into future . For the semantics in exist there is no difference.

adamretter:
Thanks Dannes, I'm happy to go with the community consensus. Previously, I was only trying to align with the W3C spec.

I appreciate your input, as it adds weight to the argument that the community would prefer a more descriptive approach. I will happily send a follow up commit to change this soon

adamretter:
@dizzzz As requested I have appended a commit to change this.
Note - as you can only specify a single collection URI, I have renamed it to `collection-uri` as opposed to `collection-uris`. 
Does that work for you?

line-o:
Now maven-dependency checks are failing :/
Can we do something about that?

adamretter:
@line-o They are now passing.

reinhapa (approved):

Issue #780, opened 2015-09-10
U+FFFD replacement characters appear instead of SMP and SIP characters


jensopetersen:
eXist does a good job at maintaining higher-level Unicode values through transforms. However, there are exceptions. Take the following - innocent - script:

```
xquery version "3.1";

declare function local:make-lemma-version($element as element())
as element()
{
    element { node-name($element) } {
        $element/@*,
        for $child in $element/node()
        return
            if ($child instance of element(app)) 
            then $child/lem/string()
            else $child
    }
};

let $test := 
<seg>
    <app>
        <lem>a</lem>
        <rdg/>
    </app>
    &#x2a3c6;
</seg>
(:let $test := doc('/db/test.xml')/element():)
let $result := local:make-lemma-version($test)
return
    $result
```

$test contains a CJK Extension B character, U+2A3C6, which should look something like http://www.unicode.org/cgi-bin/GetUnihanData.pl?codepoint=2A3C6&useutf8=true (you may not have a font that allows you to view this character in eXide).

The result displays nicely when run over an in-memory fragment, but if you save the fragment and run it over a stored document (uncommenting the second "let $test" statement), you get two "�" U+FFFD replacement characters instead of one U+2A3C6 character. All information about the original U+2A3C6 is lost. - Many will never see a SIP character, but for me this is quite a problem.

There is a further puzzle: I have put an `a` as `<lem>` contents in the document, not because it occurs in my source, but because it does not matter what you put there. If you remove the `<lem>` contents in the document (or change `then $child/lem/string()` to `then $child/rdg/string()` or to `()` in the query), the U+2A3C6 will appear again. It will also appear again if you change `then $child/lem/string()` to `then <lem>{$child/lem/string()}</lem>`. Why?

PS: The problem not only applies to characters in the Supplementary Ideographic Plane, but also to characters in the Supplementary Multilingual Plane (including e.g. Emoticons).

duncdrum:
reproduced on 4.5.0

duncdrum:
see #798

cwittern:
Still exists in 5.2.0. :-(

cwittern:
That is great, good to know!  Thanks a lot for fixing this!

Issue #787, opened 2015-09-13
Lucene index drops and ngram index chops up SMP and SIP characters


jensopetersen:
The new range index appears to work without problems with high-range characters.

The ngram index chops up such characters in two, but due to the way ngram matches occur, this does not lead to problems with ordinary searches, only to searches using wildcards, for then two dots will represent one character.

The lucene index drops certain high-range characters, but it is not clear to me what the pattern is. The SMP characters

```
𐀀    &#x10000;
𐌑    &#x10311;
𐐥    &#x10425;
𐩨    &#x10a68;
𐰀    &#x10c00;
```

work fine, but the following do not:

```
𐄉    &#x10109;
𐆊    &#x1018a;
🕐    &#x1f550;
😀    &#x1f600;
𝌆    &#x1d306;
🔋    &#x1f50b;
```

Number 1 and 2 in the second list occur between number 1 and 2 in the first list.

With SIP characters, there is the same situation: 

```
𠀀    &#x20000;
𠄩    &#x20129;
𣜕    &#x23715;
𩨻    &#x29a3b;
𫚤    &#x2b6a4;
```

are indexed, but 

```
𬵂    &#x2cd42;
𫠷    &#x2b837;
𬹩    &#x2b9ef;
𬹩    &#x2ce69;
```

are not.

The new range index shows that this can be done. The ngram index is old, which probably explains the situation. Whether lucene-internal reasons are behind the dropping of characters is unclear to me.

jensopetersen:
With the following document, stored at /db/test

	<tests>
		<test group="1">&#x10000;</test>
		<test group="1">&#x10311;</test>
		<test group="1">&#x10425;</test>
		<test group="1">&#x10a68;</test>
		<test group="1">&#x10c00;</test>
		<test group="2">&#x10109;</test>
		<test group="2">&#x1018a;</test>
		<test group="2">&#x1f550;</test>
		<test group="2">&#x1f600;</test>
		<test group="2">&#x1d306;</test>
		<test group="2">&#x1f50b;</test>
		<test group="3">&#x20000;</test>
		<test group="3">&#x20129;</test>
		<test group="3">&#x23715;</test>
		<test group="3">&#x29a3b;</test>
		<test group="3">&#x2b6a4;</test>
		<test group="4">&#x2cd42;</test>
		<test group="4">&#x2b837;</test>
		<test group="4">&#x2b9ef;</test>
		<test group="4">&#x2ce69;</test>
	</tests>

[...]

PR #3317, opened 2020-03-20
Fix the UTF8 decoding of Node Values

adamretter:
Closes https://github.com/eXist-db/exist/issues/780

p.s. If anyone is interested in eXist-db 4.x.x then this needs back-porting to there too (by someone else)

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/dom/DOMFile.java
+++ b/exist-core/src/main/java/org/exist/storage/dom/DOMFile.java
@@ -1983,7 +1983,13 @@ public class DOMFile extends BTree implements Lockable {
                 getNodeValue(broker.getBrokerPool(), (DocumentImpl) node.getOwnerDocument(),
                         os, recordPos, true, addWhitespace);
                 final byte[] data = os.toByteArray();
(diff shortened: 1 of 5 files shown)


duncdrum:
Have you also checked #787, this might explain the test failures.

adamretter:
@duncdrum Fixed it now. #787 is unrelated to this and is solely about the NGram index.

exist-core/src/main/java/org/exist/storage/dom/DOMFile.java
```diff
@@ -1983,7 +1983,13 @@ public String getNodeValue(final DBBroker broker, final IStoredNode node, final
                 getNodeValue(broker.getBrokerPool(), (DocumentImpl) node.getOwnerDocument(),
                         os, recordPos, true, addWhitespace);
                 final byte[] data = os.toByteArray();
-                return new String(data, UTF_8);
+
+                final XMLString str = UTF8.decode(data);
```
dizzzz: where is UTF8 coming from? there is no definition nor import?

PR #3319, opened 2020-03-23
Fix EXPath XAR required eXist-db version calculation

adamretter:
If a XAR does not specify that it requires a specific version of eXist-db. Then do not enforce eXist-db version 1.4.0 to 2.2.1. That doesn't make sense, as some packages e.g. functx, don't have a dependency on eXIst-db.

@wolfgangmm just want to check this one with you...

### DIFF
--- a/exist-core/src/main/java/org/exist/repo/ClasspathHelper.java
+++ b/exist-core/src/main/java/org/exist/repo/ClasspathHelper.java
@@ -22,10 +22,12 @@ package org.exist.repo;
 
 import java.io.BufferedReader;
 import java.io.IOException;
+import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Optional;
+import java.util.Set;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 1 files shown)


joewiz:
Restarted the 1 failing Appveyor CI build

joewiz:
CI tests now pass. 

There are other EXPath Package-related components that make a similar assumption.  For example, the public-repo app [states](https://github.com/eXist-db/public-repo/blob/master/modules/list.xql#L12-L18) this:

> The default version number here is assumed when a client does not send a version parameter. It is set to 2.2.0 because this version was the last one known to work with most older packages before packages began to declare their version constraints in their package metadata. So this should stay as 2.2.0 until we (a) no longer have 2.2-era clients or (b) no longer have packages that we care to offer compatibility with 2.2.

Now that the stable release is eXist 5 and posts to the list about 2.2-era clients are few and far between, it seems reasonable to drop this assumption.

(A quick search of Dashboard and Package Manager doesn't show any mentions of 2.2, 2.2.0, or 2.2.1, so these might be the only two cases.)

adamretter:
Notes from community call - the processor check needs to disambiguate between XARs that have Java content and those that don't. When there is no processor declaration and there is Java content we should log the warning.

joewiz:
As discussed a few weeks back, @adamretter said he would work on a small improvement.

joewiz:
As discussed in today's community call:

@adamretter: This PR addresses the problem that eXist raises an error message when you try to install functx, a pure XQuery package without a version declaration. This PR removes the error message. 

@duncdrum and @line-o: What's needed in this PR: If there's java code in the package, an error message should indicate that the package is missing a version declaration and should refuse to install the package unless a valid version dependency declaration is included.

dizzzz:
What is the status? code to be added?

adamretter:
@joewiz @duncdrum @dizzzz I just pushed an additional commit. I think I have now addressed the issue as we discussed it in the community call.

Therefore, as far as I know, this is now ready to be merged.

adamretter:
@joewiz @duncdrum @dizzzz can this now be merged please?

joewiz:
Given that we have two approvals and tests pass, I will merge.

dizzzz:
We should have discussions and conclusions better organised in tickets; For in the end there was still confusion if the PR was ok or not, merely due to unclear used language. Sorry it stayed so long open, happy it is in right now.

Issue #2316, opened 2018-11-29
util:get-fragment-between is not producing correct fragment(s)


babslgam:
### What is the problem

The fragmenting process used by `util:get-fragment-between` isn't stopping at second node. 

Running `util:get-fragment-between` to get the fragment between the first and second `<pb>` in the full sample below returns:

```xml
            <pb facs="1.jpg"></pb>     
            <p>Aus dem Leben einer Kartoffel.</p>
            <pb facs="2.jpg"></pb>
       </front>
       <body>
            <pb facs="3.jpg"></pb>
            <p>Hubertus Knoll spazierte über das <pb facs="4.jpg"></pb> Feld.</p>
            <pb facs="5.jpg"></pb>
        </body>
    </text>
</TEI>
```

Maybe the problem was caused here?
https://github.com/eXist-db/exist/blob/eXist-4.4.0/src/org/exist/xquery/functions/util/GetFragmentBetween.java#L182

### What did you expect

I expected output (like in version 4.3.1):

```xml
<pb facs="1.jpg"></pb>
<p>Aus dem Leben einer Kartoffel.</p>
```

### Describe how to reproduce or add a test

Store the following as `kartoffelmann.xml`:

```xml
<TEI xmlns="http://www.tei-c.org/ns/1.0">
    <text>
        <front>
            <pb facs="1.jpg"/>
            <p>Aus dem Leben einer Kartoffel.</p>
            <pb facs="2.jpg"/>
        </front>
        <body>
            <pb facs="3.jpg"/>
            <p>Hubertus Knoll spazierte über das <pb facs="4.jpg"/> Feld.</p>
            <pb facs="5.jpg"/>
        </body>
    </text>
</TEI>
```

Run the following query:

```xquery
xquery version "3.1";

declare namespace tei = "http://www.tei-c.org/ns/1.0";

let $doc := doc("kartoffelmann.xml")
let $pbs := $doc//tei:pb
let $pb1 := $pbs[1]
let $pb2 := $pbs[2]
return
    util:get-fragment-between($pb1, $pb2, false(), true())
```

Results are as above.

### Context information

- eXist-db version:  eXist-db >= 4.4.0
- Java version:  jdk1.8.0_191
- Operating system: Windows 10
- 64 bit
- How is eXist-db installed? JAR installer
- Any custom changes in e.g. conf.xml: no

joewiz:
I can confirm @babslgam's report with the current eXist-db develop-4.x.x.  A simplified example:

[...]

PR #3328, opened 2020-03-26
Fix an issue with Java equality of DLN

adamretter:
Closes https://github.com/eXist-db/exist/issues/2316

The following returns `true` as expected:
```Java
new DLN("1.1.1").equals(new DLN("1.1.1"))
```

However, the following unexpectedly returns `false`, when one would expect it to return `true`:
```java
Optional.of(new DLN("1.1.1").equals(Optional.of(new DLN("1.1.1"))))
```

This is due to a strange definition (and implementation) of `equals` in `NodeId`; I have removed this potentially dangerous pattern so that you get expected behaviour.

Most of the test I adapted from @joewiz. I would be happy if Joe could check he agrees with them still ;-)

### DIFF
--- a/exist-core/src/test/xquery/util/get-fragment-between.xqm
+++ b/exist-core/src/test/xquery/util/get-fragment-between.xqm
@@ -50,7 +50,29 @@ function gfb:tei-fragment() {
     let $pb1 := $pbs[1]
     let $pb2 := $pbs[2]
     return
-        util:get-fragment-between($pb1,$pb2,false(),true())
+        util:get-fragment-between($pb1, $pb2, false(), false())
+};
+
+declare
+    %test:assertEquals('<TEI><text><front><pb facs="1.jpg"></pb><p>Aus dem Leben einer Kartoffel.</p></front></text></TEI>')
+function gfb:tei-fragment-root() {
+    let $doc := doc("/db/test-gfb/doc1.xml")
+    let $pbs := $doc//tei:pb
+    let $pb1 := $pbs[1]
+    let $pb2 := $pbs[2]
+    return
+        fn:replace(util:get-fragment-between($pb1, $pb2, true(), false()), "\n", "")
+};
+
+declare
+    %test:assertEquals('<TEI xmlns="http://www.tei-c.org/ns/1.0"><text><front><pb facs="1.jpg"></pb><p>Aus dem Leben einer Kartoffel.</p></front></text></TEI>')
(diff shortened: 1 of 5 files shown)


exist-core/src/test/xquery/util/get-fragment-between.xqm
```diff
@@ -0,0 +1,85 @@
...
+};
+
+declare
+    %test:tearDown
+function gfb:teardown() {
+    xmldb:remove("/db/test-gfb")
+};
+
+declare
+    %test:assertEquals('<pb facs="1.jpg"></pb><p>Aus dem Leben einer Kartoffel.</p>')
```
joewiz: The fact that the function's 4th parameter is `true()` means the resulting fragment should preserve the root namespace:

> The fourth argument, `$display-root-namespace`, is a boolean value for displaying the root node namespace. If it is set to `true()` the attribute `xmlns` in the root node of the result sequence is determined explicitly from the `$beginning-node`.

```suggestion
    %test:assertEquals('<pb xmlns="http://www.tei-c.org/ns/1.0" facs="1.jpg"></pb><p>Aus dem Leben einer Kartoffel.</p>')
```

joewiz:
Restarted the one failing Travis CI build. Still awaiting final Appveyor build.

exist-core/src/test/xquery/util/get-fragment-between.xqm
adamretter: I think the 4th parameter only has an effect when the 3rd parameter is `true`. I have committed two further tests to try and illustrate this.

Unfortunately it is very hard to know what the 'correct' behaviour of `util:get-fragment-between` is exactly mean to be as there is no documentation :-/

joewiz: Good point about the thin documentation for this function (explaining why I wasn’t sure whether the second node was preserved or discarded). 

But is it necessary to limit namespace preservation to when a root node is added? Surely users would need a way to preserve the namespace of the resulting nodes. 

My suggestion also neglected to include the namespace on the 2nd node.

adamretter: > But is it necessary to limit namespace preservation to when a root node is added? Surely users would need a way to preserve the namespace of the resulting nodes.

No... but my goal was to fix the regression, and just get it working as it was in eXist-db 4.3.1.

joewiz:
Fair enough! Thank you!!!

PR #3332, opened 2020-03-27
Group managers should be loaded from security database at startup

adamretter:
Closes https://github.com/eXist-db/exist/pull/1578

Previously you could add group managers, and save them. They were correctly persisted to the database. However, at the next database startup they were not loaded correctly and then removed from persistent storage; effectively that deleted any group managers... Ouch! Since at least 2017?!?

### DIFF
--- a/exist-core/src/main/java/org/exist/security/AbstractGroup.java
+++ b/exist-core/src/main/java/org/exist/security/AbstractGroup.java
@@ -172,6 +172,22 @@ public abstract class AbstractGroup extends AbstractPrincipal implements Compara
         managers.add(new ReferenceImpl<>(getRealm().getSecurityManager(), "getAccount", name));
     }
 
+    //this method is used by Configurator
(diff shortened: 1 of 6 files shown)


exist-core/src/main/java/org/exist/security/AbstractGroup.java
```diff
@@ -176,6 +172,22 @@ public void addManager(final String name) throws PermissionDeniedException {
         managers.add(new ReferenceImpl<>(getRealm().getSecurityManager(), "getAccount", name));
     }
 
+    //this method is used by Configurator
```
dizzzz: Where exactly?

adamretter: Yeah it wasn't easy to find and fix! It is used by `Configurator#searchForInsertMethod(Class, String)`.

PR #3335, opened 2020-03-30
[Feature] Add memory and nr of processors information to startup log.

dizzzz:
Output:

```
30 Mar 2020 19:54:38,255 [main] INFO  (JettyStart.java [run]:173) - Running with Java 1.8.0_242 [AdoptOpenJDK (OpenJDK 64-Bit Server VM) in /Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home/jre] 
30 Mar 2020 19:54:38,257 [main] INFO  (JettyStart.java [run]:180) - Maximum amount of memory for JVM: 3641 MiB 
30 Mar 2020 19:54:38,258 [main] INFO  (JettyStart.java [run]:181) - Number of processors available to JVM: 8  
```

### DIFF
--- a/exist-core/src/main/java/org/exist/jetty/JettyStart.java
+++ b/exist-core/src/main/java/org/exist/jetty/JettyStart.java
@@ -177,7 +177,7 @@ public class JettyStart extends Observable implements LifeCycle.Listener {
                 System.getProperty("java.home", "(unknown java.home)")
             );
 
-            logger.info("Maximum amount of memory for JVM: {} MiB", Runtime.getRuntime().maxMemory() / (1024 * 1024));
+            logger.info("Approximate maximum amount of memory for JVM: {}", FileUtils.humanSize(Runtime.getRuntime().maxMemory()));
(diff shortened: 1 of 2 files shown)


exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -177,6 +177,9 @@ public synchronized void run(final String[] args, final Observer observer) {
                 System.getProperty("java.home", "(unknown java.home)")
             );
 
+            logger.info("Maximum amount of memory for JVM: {} MiB", Runtime.getRuntime().maxMemory() / (1024 * 1024));
```
adamretter: Whilst `MiB` might be more accurate. Could we just use `MB` so as to cause less confusion?

Also instead of doing the calculation yourself, you might like to use `FileUtils#humanSize(long)` as that will take care of MB, GB, etc.

adamretter: LGTM otherwise :-)

dizzzz:
sorry for the typo in comment :-)

PR #3338, opened 2020-03-31
add code to disable trigger execution during expath pkg installation

chakl:
### Description:

This patch allows to install an expath package with **trigger execution suppressed** during the installation.

There can be cases where the execution of triggers during package installation is unwanted:
- if replication triggers are being used: when a package gets upgraded to a newer version, the old package data gets deleted before re-installing newer data. And replication triggers will propagate these deletions to the replicated DB. Which spoils the idea of replication.
- similar issues with triggers interacting with Amazon AWS S3
- similar issues with triggers interacting with Github/Gitlab

##### Add optional Element `skip-triggers` to file `repo.xml` of a Package

```
<meta xmlns="http://exist-db.org/xquery/repo">
    <description>An app to test trigger behavior</description>
    <author>Me</author>
    <!-- other elements ... -->

    <skip-triggers/>
</meta>
```

If the empty `skip-triggers` element is present in file `repo.xml`, then triggers shall not be executed. If the element is not present, code behavior is unchanged, so this is backwards compatible.

##### Piggyback on existing `DBBroker:isTriggersEnabled()` and `DBBroker:setTriggersEnabled()` Methods

In method `Deployment:deploy()`, method `setTriggersEnabled()` is used to temporarily disable and later re-enable trigger execution state, if the `skip-triggers` element is present in file `repo.xml`.

`setTriggersEnabled()` applies to a single broker instance only. It is used during execution of method `deploy()` in the context of a single broker during package installation, so this should be good for the purpose.

##### All Trigger Usages in Class `NativeBroker` honor the `isTriggersEnabled()` Method

All trigger instantiation and usage cases honor `isTriggersEnabled()` following this pattern:
- trigger instance initialized as `null`
- instantiate trigger only if `isTriggersEnabled() == true`
- protect all uses of potential null object with `if (trigger != null)`

#### Dragons Lurking

While not strictly relevant to the problem at hand (suppress triggers during Xar install), I suspect class `MutableCollection` may need similar treatment.

Caching effects have not been considered yet.

### Reference:

Implemented to meet customer requirements.

### Type of tests:

Works for me.

Seriously: This is a working PoC implementation, comments and criticism welcome.

### DIFF
--- a/exist-core/src/main/java/org/exist/repo/Deployment.java
+++ b/exist-core/src/main/java/org/exist/repo/Deployment.java
@@ -478,9 +478,10 @@ public class Deployment {
                 return Optional.ofNullable(targetCollection.getCollectionPath());
             }
         } catch (final XPathException e) {
+            throw new PackageException("Error found while processing repo.xml: " + e.getMessage(), e);
+	} finally {
             // reset possibly altered trigger state
             broker.setTriggersEnabled(triggerState);
-            throw new PackageException("Error found while processing repo.xml: " + e.getMessage(), e);
         }
     }
 
--- a/exist-core/src/main/java/org/exist/repo/Deployment.java
+++ b/exist-core/src/main/java/org/exist/repo/Deployment.java
@@ -352,6 +352,10 @@ public class Deployment {
         if (repoXML == null) {
             return Optional.empty();
         }
+
+        // remember current trigger state
+        final boolean triggerState = broker.isTriggersEnabled();
+
         try {
             // if there's a <setup> element, run the query it points to
             final Optional<ElementImpl> setup = findElement(repoXML, SETUP_ELEMENT);
@@ -393,9 +397,8 @@ public class Deployment {
                     targetCollection = XmldbURI.SYSTEM.append("repo/" + pkgColl);
                 }
 
-                // skip trigger execution and remember previous state
+                // skip trigger execution if demanded in repo.xml
                 final Optional<ElementImpl> skipTriggers = findElement(repoXML, SKIP_TRIGGERS_ELEMENT);
-                final boolean triggerState = broker.isTriggersEnabled();
                 if(skipTriggers.isPresent()) {
                     broker.setTriggersEnabled(false);
                 }
@@ -475,6 +478,8 @@ public class Deployment {
                 return Optional.ofNullable(targetCollection.getCollectionPath());
             }
         } catch (final XPathException e) {
+            // reset possibly altered trigger state
+            broker.setTriggersEnabled(triggerState);
             throw new PackageException("Error found while processing repo.xml: " + e.getMessage(), e);
         }
     }
--- a/exist-core/src/main/java/org/exist/repo/Deployment.java
+++ b/exist-core/src/main/java/org/exist/repo/Deployment.java
@@ -462,11 +462,6 @@ public class Deployment {
 
                 storeRepoXML(broker, transaction, repoXML, targetCollection, requestedPerms);
 
-                // restore trigger execution to previous state
-                if(skipTriggers.isPresent()) {
-                    broker.setTriggersEnabled(triggerState);
-                }
-
                 // TODO: it should be safe to clean up the file system after a package
                 // has been deployed. Might be enabled after 2.0
                 //cleanup(pkgName, repo);
--- a/exist-core/src/main/java/org/exist/repo/Deployment.java
+++ b/exist-core/src/main/java/org/exist/repo/Deployment.java
@@ -88,6 +88,7 @@ public class Deployment {
     private static final QName POST_SETUP_ELEMENT = new QName("finish", REPO_NAMESPACE);
     private static final QName TARGET_COLL_ELEMENT = new QName("target", REPO_NAMESPACE);
     private static final QName PERMISSIONS_ELEMENT = new QName("permissions", REPO_NAMESPACE);
+    private static final QName SKIP_TRIGGERS_ELEMENT = new QName("skip-triggers", REPO_NAMESPACE);
     private static final QName CLEANUP_ELEMENT = new QName("cleanup", REPO_NAMESPACE);
     private static final QName DEPLOYED_ELEMENT = new QName("deployed", REPO_NAMESPACE);
     private static final QName DEPENDENCY_ELEMENT = new QName("dependency", PKG_NAMESPACE);
@@ -392,6 +393,13 @@ public class Deployment {
                     targetCollection = XmldbURI.SYSTEM.append("repo/" + pkgColl);
                 }
 
+                // skip trigger execution and remember previous state
+                final Optional<ElementImpl> skipTriggers = findElement(repoXML, SKIP_TRIGGERS_ELEMENT);
+                final boolean triggerState = broker.isTriggersEnabled();
+                if(skipTriggers.isPresent()) {
+                    broker.setTriggersEnabled(false);
+                }
+
                 // extract the permissions (if any)
                 final Optional<ElementImpl> permissions = findElement(repoXML, PERMISSIONS_ELEMENT);
                 final Optional<RequestedPerms> requestedPerms = permissions.flatMap(elem -> {
@@ -450,6 +458,11 @@ public class Deployment {
                 }
 
                 storeRepoXML(broker, transaction, repoXML, targetCollection, requestedPerms);
+
+                // restore trigger execution to previous state
+                if(skipTriggers.isPresent()) {
+                    broker.setTriggersEnabled(triggerState);
+                }
 
                 // TODO: it should be safe to clean up the file system after a package
                 // has been deployed. Might be enabled after 2.0
--- a/exist-core/src/main/java/org/exist/storage/NativeBroker.java
+++ b/exist-core/src/main/java/org/exist/storage/NativeBroker.java
@@ -740,13 +740,15 @@ public class NativeBroker extends DBBroker {
             final CollectionCache collectionCache, final Optional<Tuple2<Permission, Long>> creationAttributes)
             throws TriggerException, ReadOnlyException, PermissionDeniedException, LockException, IOException {
 
-        final CollectionTrigger trigger;
-        if(parentCollection == null) {
-            trigger = new CollectionTriggers(this, transaction);
-        } else {
-            trigger = new CollectionTriggers(this, transaction, parentCollection);
+        CollectionTrigger trigger = null;
+        if (isTriggersEnabled()) {
+            if (parentCollection == null) {
+                trigger = new CollectionTriggers(this, transaction);
+            } else {
+                trigger = new CollectionTriggers(this, transaction, parentCollection);
+            }
+            trigger.beforeCreateCollection(this, transaction, collectionUri);
         }
-        trigger.beforeCreateCollection(this, transaction, collectionUri);
 
         final Collection collectionObj = createCollectionObject(transaction, parentCollection, collectionUri, creationAttributes);
         saveCollection(transaction, collectionObj);
@@ -758,7 +760,9 @@ public class NativeBroker extends DBBroker {
 
         collectionCache.put(collectionObj);
 
-        trigger.afterCreateCollection(this, transaction, collectionObj);
+        if (trigger != null) {
+            trigger.afterCreateCollection(this, transaction, collectionObj);
+        }
 
         return collectionObj;
     }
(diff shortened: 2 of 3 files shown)


exist-core/src/main/java/org/exist/repo/Deployment.java
```diff
@@ -392,6 +393,13 @@ private void checkProcessorVersion(final PackageLoader.Version version) throws P
                     targetCollection = XmldbURI.SYSTEM.append("repo/" + pkgColl);
                 }
 
+                // skip trigger execution and remember previous state
+                final Optional<ElementImpl> skipTriggers = findElement(repoXML, SKIP_TRIGGERS_ELEMENT);
+                final boolean triggerState = broker.isTriggersEnabled();
```
dizzzz: more compact would be `skipTriggers.ifPresent(c -> broker.setTriggersEnabled(false));`

exist-core/src/main/java/org/exist/storage/NativeBroker.java
```diff
@@ -740,13 +740,15 @@ public Collection getOrCreateCollection(final Txn transaction, final XmldbURI na
             final CollectionCache collectionCache, final Optional<Tuple2<Permission, Long>> creationAttributes)
             throws TriggerException, ReadOnlyException, PermissionDeniedException, LockException, IOException {
 
-        final CollectionTrigger trigger;
-        if(parentCollection == null) {
-            trigger = new CollectionTriggers(this, transaction);
-        } else {
-            trigger = new CollectionTriggers(this, transaction, parentCollection);
+        CollectionTrigger trigger = null;
```
dizzzz: is the removal of final of line 743 needed? also is the null initialisation needed?

exist-core/src/main/java/org/exist/storage/NativeBroker.java
```diff
@@ -758,7 +760,9 @@ public Collection getOrCreateCollection(final Txn transaction, final XmldbURI na
 
         collectionCache.put(collectionObj);
 
-        trigger.afterCreateCollection(this, transaction, collectionObj);
+        if (trigger != null) {
```
dizzzz: is a NPE really possible? whoops... nice catch

dizzzz:
nice PR, really added value;  I think there should be a test added, `extensions/modules/expathrepo/src/test/java/org/exist/repo/PackageTriggerTest.java` might be an inspiration

exist-core/src/main/java/org/exist/repo/Deployment.java
chakl: Thanks, will consider. For now, too much compaction kills debugging

exist-core/src/main/java/org/exist/collections/triggers/CollectionTriggers.java
```diff
@@ -83,72 +86,88 @@ public void configure(DBBroker broker, Txn transaction, Collection col, Map<Stri
 
     @Override
     public void beforeCreateCollection(DBBroker broker, Txn txn, XmldbURI uri) throws TriggerException {
-        for (CollectionTrigger trigger : triggers) {
-            trigger.beforeCreateCollection(broker, txn, uri);
+        if (!triggersSuspended) {
```
adamretter, line 89: These `triggersSuspended` checks should not be used here. Such checks are already present at the call site in DBBorker. This kinda breaks the responsibility pattern

exist-core/src/main/java/org/exist/collections/triggers/DocumentTriggers.java
```diff
@@ -53,7 +53,9 @@
     private SAXTrigger last = null;
     
     private final List<DocumentTrigger> triggers;
-    
+
+    private boolean skipTriggers = false;
```
adamretter, line 57: Likewise... These `triggersSuspended` checks should not be used here. Such checks are already present at the call site in DBBorker.

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
```diff
@@ -0,0 +1,126 @@
...
+
+
+    public CollectionSuspendTriggers(XQueryContext context, FunctionSignature signature) {
+        super(context, signature);
+    }
+
+    /* (non-Javadoc)
+     * @see org.exist.xquery.BasicFunction#eval(org.exist.xquery.value.Sequence[], org.exist.xquery.value.Sequence)
+     */
+    public Sequence eval(Sequence[] args, Sequence contextSequence)
```
adamretter, line 84: IMHO there are are a lot of bad things happening in this function:

1. This can be used by any user to disable all triggers for all brokers. I could issue a series of REST calls which would break anyones app!
2. This disables all triggers for a collection, and is not per-broker. 
3. Disabling triggers should be a temporary measure, as such the design of this function leads to problems with a disable-enable pattern. I would suggest it instead should be refactored as a HoF.

line-o:
@adamretter @brihaye The files of the second commit [PoC] are not part of this PR and will be moved to a separate one. We should continue the discussion there.

chakl:
@adamretter I screwed up adding changes that were meant to go into a separate PR. Removed, sorry for the confusion.

chakl:
> Since we are at it, don't beforeCreateCollection, beforeCopyCollection, beforeMoveCollection, beforeDeleteCollection, beforeUpdateDocument, beforeUpdateDocumentMetadata, beforeCopyDocument, beforeMoveDocument, beforeDeleteDocument lack a try / catch block ?

No. Please head further discussion to #3339, thanks.

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
chakl: @adamretter the code ypu are referring to was moved to #3339, let's continue discussion there.
- 1 - can you sketch out that attack scenario? while testing I saw `PermissionDenied` exceptions, so I assumed standard eXist-db access control applies;
- 2 - Yes. That is the purpose of #3339 
- 3 - Fully agree. In a perfect world we would have LISP-style `(with-triggers-disabled (code ...))` where the `with-...` clause does all the setup and teardown actions around `code`. I fear this may not work with multithreaded eXist-db and asynchrnous HTTP I/O, but I'm open for suggestions

adamretter:
>  I screwed up adding changes that were meant to go into a separate PR. Removed, sorry for the confusion.

The additional commits are still in this PR, could you remove them please?

exist-core/src/main/java/org/exist/repo/Deployment.java
```diff
@@ -462,6 +478,8 @@ private void checkProcessorVersion(final PackageLoader.Version version) throws P
                 return Optional.ofNullable(targetCollection.getCollectionPath());
             }
         } catch (final XPathException e) {
+            // reset possibly altered trigger state
+            broker.setTriggersEnabled(triggerState);
```
adamretter, line 482: This isn't strong enough... consider using a try/finally mechanism

exist-core/src/main/java/org/exist/repo/Deployment.java
```diff
@@ -392,6 +397,12 @@ private void checkProcessorVersion(final PackageLoader.Version version) throws P
                     targetCollection = XmldbURI.SYSTEM.append("repo/" + pkgColl);
                 }
 
+                // skip trigger execution if demanded in repo.xml
+                final Optional<ElementImpl> skipTriggers = findElement(repoXML, SKIP_TRIGGERS_ELEMENT);
+                if(skipTriggers.isPresent()) {
```
adamretter, line 402: Should we perhaps check for `<skipTriggers>true</skipTriggers>` seems more natural

exist-core/src/main/java/org/exist/repo/Deployment.java
chakl: You're right, thanks

exist-core/src/main/java/org/exist/repo/Deployment.java
chakl: Hm, to me the empty but present `skip-triggers` element feels more natural, but I don't mind. Whatever exist-core prefers to see.

line-o: Reading through the [repo documentation](http://exist-db.org/exist/apps/doc/repo.xml#sect-repoxml), what about
`<triggers skip="{true/false}" />` similar to the permissions setting?

chakl:
I added a test based on the test @dizzzz suggested. I'm not sure how to link it to the test run

exist-core/src/main/java/org/exist/repo/Deployment.java
chakl: @line-o yet another option, I really don't mind the representation. Shall we vote or who's the dictator?

chakl:
Ah yes, I have seen these places when looking where triggers are used. Related, but not in the scope of this PR IMHO

exist-core/src/main/java/org/exist/repo/Deployment.java
adamretter: No dictators here, just consensus.

line-o: If I have to sell my proposal: `<triggers />` is easier to extend. The attribute name `execute-on-install` is easy to grasp. `on|off` or `true|false` should not matter, but depends on the current implementation what counts as a boolean.

chakl:
@dizzzz Can you give me some guidance regarding the test for this feature? I have added a `PackageSkipTriggerTest` much like the `PackageTriggerTest` that you mentioned. As in your example with the `triggertest-1.0.0.xar` package, I install a  xar file crafted for this test (which would log trigger execution). The test then verifies that no trigger execution was logged.
How would I integrate this particular test into the test suite? Thanks

exist-core/src/main/java/org/exist/repo/Deployment.java
adamretter: @chakl @line-o I would suggest `<triggers skip-on-install="true"/>`

adamretter: Also what happens with uninstall of the app? Are triggers notified?

adamretter:
Hmm it has just occurred to me that there may actually be valid reasons for not doing this in quite this way.

What happens if I have configured a Trigger that I want to be notified when things in the db happen? Wouldn't this cause anything to do with installing/uninstalling this app (and all of it's content) to be invisible to it?

If so... then the approach taken is too granular. I would suggest instead that Triggers are given additional events - e.g. `startAppInstall` and `endAppInstall` - it is then up to each Trigger whether to ignore the events between those two markers or not. I think this should solve your use-case and not break the other peoples use cases for Triggers.

chakl:
Thanks for your input. Yes, the current state of this patch is to disable all trigger execution on the single broker instance that is installing the package, during the time period in which the package files get installed. So that broker would not fire any triggers during execution of the `Deployment:deploy()` method. Other broker threads would not be affected.

To be honest, I have not considered the scenario you describe, need to check.

And thanks for the constructive trigger event suggestions, need to think about this, not sure I like their implications :)

chakl:
Thanks for your input. Yes, the current state of this patch is to disable all trigger execution on the single broker instance that is installing the package, during the time period in which the package files get installed. So that broker would not fire any (inherited?) triggers during execution of the `Deployment:deploy()` method. Other broker threads would not be affected.

To be honest, I have not considered the scenario you describe, need to check. If it does break things it is wrong.

And thanks for the constructive trigger event suggestions, need to think about this, not sure I like the implications :)

chakl:
sorry

adamretter:
> not sure I like their implications

What are the implications that you see?

chakl:
triggers are stateless AFAICS. `startAppInstall` would impose state that needs to be kept somewhere. This state would be used by other triggers to decide "should I fire" - if I understand you correctly. Imposing state on triggers would need massive changes and add quite a lot of complexity.

adamretter:
@chakl Triggers definitely are not stateless ;-)

chakl:
Ermm? trigger instances get created in the context of methods that use them and get thrown away on method return. Where is state then?

chakl:
> Triggers definitely are not stateless ;-)

Got me a code pointer, please?

adamretter:
@chakl sure just look at some of the triggers that ship with eXist-db in the package `org.exist.collections.triggers`

chakl:
yes I did. Where is "not stateless" that you mentioned?

adamretter:
@chakl Erm... Several of those triggers have mutable class members for keeping state!?!

chakl:
ok will look into it, thanks

chakl:
closing, needs different approach

PR #3339, opened 2020-04-01
[PoC] XQuery functions to en/disable trigger execution on collections at runtime

chakl:
### Description:

NOTE: this patch is completely different from PR #3338 , which was focused on trigger suppression during package installation. I accidentally lumped them together in that PR, sorry for the confusion.

This is a PoC implementation of XQuery functions to disable and re-enable triggers on a collection at runtime. A typical use case would be an XQuery like this:

```
(: temporarily disable triggers on collection :)
util:collection-suspend-triggers("/db/mycollection"),

(: do things while triggers are disabled :)

(: re-enable triggers for this collection :)
util:collection-enable-triggers("/db/mycollection"),
```

Conceptually, these things need to be addressed:
- where to keep the "triggers-disabled" flag for a collection
- how to access that place from the context of an XQuery, in order to set/clear the flag
- how to provide the flag to the trigger execution logic

#### Where to keep the "triggers-disabled" Flag for a Collection

I have chosen to keep this flag in class `CollectionConfiguration` because
- IMHO it is the most obvious place, it is a collection attribute semantically similar to "permissions"
- we can easily enhance `*.xconf` file parsing to something like "this collection does not want triggers enabled"

An alternative (and possibly better, or more consistent) place would be class `MutableCollection`. However, passing that flag from class `CollectionConfiguration` would require significant changes to the API of various methods (probably changing `Tuple2` to `Tuple3` objects). I won't go that route without feedback.

#### How to access that Place from the Context of an XQuery, in order to set/clear the Flag

I have used `xquery/functions/util/CollectionName.java` as a template to create the XQuery functions `util:collection-suspend-triggers()` and `util:collection-enable-triggers()`. I may have left some artifacts..

The code accesses the `CollectionConfiguration` object of the collection representation provided by the XQuery API. Access to this object through `LockedCollection.unwrapLocked()` is brute, I'm all ears for more elegant ways. It works anyway.
Once we have the `CollectionConfiguration` object, we can use its `setSkipTriggers()` method to set the state demanded by calling XQuery function.

#### How to provide the Flag to the Trigger Execution Logic

This is a 2-step approach:
- instances of classes `CollectionTriggers` and `DocumentTriggers` are not persistent, they are throwaway objects allocated when the method executes and deallocated when the method returns. Upon instantiation a `triggersSuspended` flag is set that is derived from the collections' `CollectionConfiguration` object.
- Trigger methods in these classes honor the `triggersSuspended` flag

#### Overview of Code Changes

- `CollectionConfiguration.java` - add flag and get/set methods
- `CollectionTriggers.java` and `DocumentTriggers.java` - honor CollectionConfiguration flag on instantiation and method execution
- `UtilModule.java` and `CollectionSuspendTriggers.java` - XQuery function implementation

### Reference:

Colleagues and customers would like that functionality.

### Type of tests:

None yet.  This is PoC, I appreciate feedback before proceeding.

### DIFF
--- a/exist-core/src/main/java/org/exist/collections/CollectionConfiguration.java
+++ b/exist-core/src/main/java/org/exist/collections/CollectionConfiguration.java
@@ -90,6 +90,8 @@ public class CollectionConfiguration {
 
     private final BrokerPool pool;
 
+    private boolean skipTriggers = false;
+
     public CollectionConfiguration(final BrokerPool pool) {
         this.pool = pool;
     }
@@ -205,6 +207,14 @@ public class CollectionConfiguration {
 
     public IndexSpec getIndexConfiguration() {
         return indexSpec;
+    }
+
+    public boolean getSkipTriggers() {
+        return skipTriggers;
+    }
+
+    public void setSkipTriggers(boolean skipTriggers) {
+        this.skipTriggers = skipTriggers;
     }
 
     private void configureTrigger(final ClassLoader cl, final Element triggerElement, final XmldbURI collectionConfigurationURI, final boolean testOnly) throws CollectionConfigurationException {
--- a/exist-core/src/main/java/org/exist/collections/triggers/CollectionTriggers.java
+++ b/exist-core/src/main/java/org/exist/collections/triggers/CollectionTriggers.java
@@ -34,8 +34,10 @@ import org.exist.xmldb.XmldbURI;
  *
  */
 public class CollectionTriggers implements CollectionTrigger {
-    
+
     private final List<CollectionTrigger> triggers;
+
+    private boolean triggersSuspended = false;
 
     public CollectionTriggers(DBBroker broker, Txn transaction) throws TriggerException {
         this(broker, transaction, null, null);
@@ -50,6 +52,7 @@ public class CollectionTriggers implements CollectionTrigger {
         List<TriggerProxy<? extends CollectionTrigger>> colTriggers = null;
         if (config != null) {
             colTriggers = config.collectionTriggers();
+            triggersSuspended = config.getSkipTriggers();
         }
         
         java.util.Collection<TriggerProxy<? extends CollectionTrigger>> masterTriggers = broker.getDatabase().getCollectionTriggers();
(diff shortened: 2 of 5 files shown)


dizzzz:
please could you mark your PR as 'draft' ? Github provides two levels these days.

chakl:
@dizzzz I have no idea how to do that

joewiz:
Perhaps @dizzzz is referring to [draft pull requests](https://github.blog/2019-02-14-introducing-draft-pull-requests/), which then lets you [indicate](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-stage-of-a-pull-request) that a PR isn't yet ready for review (until it is ready), but [apparently](https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/td-p/19107) you can't switch a PR to "draft" once it's been submitted. The alternative is the "In Progress" label we use in this and many eXist-db organization repositories.

chakl:
thanks @joewiz. Actually the PR **is** ready for review (works for me). I'd like to hear opinions on doing stuff right

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
```diff
@@ -0,0 +1,126 @@
...
+
+
+    public CollectionSuspendTriggers(XQueryContext context, FunctionSignature signature) {
+        super(context, signature);
+    }
+
+    /* (non-Javadoc)
+     * @see org.exist.xquery.BasicFunction#eval(org.exist.xquery.value.Sequence[], org.exist.xquery.value.Sequence)
+     */
+    public Sequence eval(Sequence[] args, Sequence contextSequence)
```
adamretter, line 84: IMHO there are are a lot of bad things happening in this function:

1. This can be used by any user to disable all triggers for all brokers. I could issue a series of REST calls which would break anyone's app!
2. This disables all triggers for a collection, and is not per-broker. 
3. Disabling triggers should be a temporary measure, as such the design of this function leads to problems with a disable-enable pattern. I would suggest it instead should be refactored as a HoF.

exist-core/src/main/java/org/exist/collections/triggers/DocumentTriggers.java
```diff
@@ -53,7 +53,9 @@
     private SAXTrigger last = null;
     
     private final List<DocumentTrigger> triggers;
-    
+
+    private boolean skipTriggers = false;
```
adamretter, line 57: Likewise... These `triggersSuspended` checks should not be used here. Such checks are already present at the call site in DBBroker.

exist-core/src/main/java/org/exist/collections/triggers/CollectionTriggers.java
```diff
@@ -83,72 +86,88 @@ public void configure(DBBroker broker, Txn transaction, Collection col, Map<Stri
 
     @Override
     public void beforeCreateCollection(DBBroker broker, Txn txn, XmldbURI uri) throws TriggerException {
-        for (CollectionTrigger trigger : triggers) {
-            trigger.beforeCreateCollection(broker, txn, uri);
+        if (!triggersSuspended) {
```
adamretter, line 89: These `triggersSuspended` checks should not be used here. Such checks are already present at the call site in DBBroker. This kinda breaks the responsibility pattern

exist-core/src/main/java/org/exist/collections/CollectionConfiguration.java
```diff
@@ -90,6 +90,8 @@
 
     private final BrokerPool pool;
 
+    private boolean skipTriggers = false;
```
adamretter, line 93: This isn't a suitable stable place to keep this flag!

Each time the `collection.xconf` is changed your setting will be lost, as this class is GC'd and a new instance created from the `collection.xconf` file.

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
```diff
@@ -0,0 +1,126 @@
...
+import org.exist.xquery.Cardinality;
+import org.exist.xquery.FunctionSignature;
+import org.exist.xquery.XPathException;
+import org.exist.xquery.XQueryContext;
+import org.exist.xquery.value.*;
+import org.xmldb.api.base.Collection;
+import org.xmldb.api.base.XMLDBException;
+
+/**
+ * @author <a href="mailto:wolfgang@exist-db.org">Wolfgang Meier</a>
```
adamretter, line 46: @chakl Shouldn't this be you Olaf?

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
chakl: [copied from the other PR thread]

1 - can you sketch out that attack scenario? while testing I saw PermissionDenied exceptions, so I assumed standard eXist-db access control applies;
2 - Yes. That is the purpose of #3339
3 - Fully agree. In a perfect world we would have LISP-style (with-triggers-disabled (code ...)) where the with-... clause does all the setup and teardown actions around code. I fear this may not work with multithreaded eXist-db and asynchrnous HTTP I/O, but I'm open for suggestions

exist-core/src/main/java/org/exist/collections/triggers/DocumentTriggers.java
chakl: @adamretter I fail to see where these checks are already present. Got me a pointer?

exist-core/src/main/java/org/exist/collections/CollectionConfiguration.java
chakl: Ok, where would you place that flag? Obviously some flag would have to be kept somewhere

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
chakl: yes, copypasta

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
adamretter: 1. Submit simple XQuery script to the REST Server over HTTP that calls your function for every collection in the database. There is no checking who is calling it or if they are permitted to.

2 Ok.

3. This has nothing to do with "multithreaded eXist-db and asynchrnous HTTP I/O", we have other HoF's. No reason why this couldn't be one.

exist-core/src/main/java/org/exist/collections/triggers/DocumentTriggers.java
adamretter: See for example `MutableCollection#removeXMLResource`:

```
boolean useTriggers = broker.isTriggersEnabled();
                    if (CollectionConfiguration.DEFAULT_COLLECTION_CONFIG_FILE_URI.equals(name)) {
                        // we remove a collection.xconf configuration file: tell the configuration manager to
                        // reload the configuration.
                        useTriggers = false;
                        final CollectionConfigurationManager confMgr = broker.getBrokerPool().getConfigurationManager();
                        if (confMgr != null) {
                            confMgr.invalidate(getURI(), broker.getBrokerPool());
                        }
                    }

                    final DocumentTriggers trigger = new DocumentTriggers(broker, transaction, null, this, useTriggers ? getConfiguration(broker) : null);
```

exist-core/src/main/java/org/exist/collections/CollectionConfiguration.java
adamretter: I don't have a suggestion, I just noticed that where you have it won't work as you intended.

exist-core/src/main/java/org/exist/collections/triggers/DocumentTriggers.java
chakl: ah `MutableCollection`, already stared at that, and slapped my fingers when opening the editor :)
Will have a deeper look, thanks

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
chakl: What do you mean by "HoF"? higher-order functions? passing lambdas as arguments? function callbacks?

exist-core/src/main/java/org/exist/collections/CollectionConfiguration.java
chakl: Thanks for the heads-up

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
chakl: > 1. Submit simple XQuery script to the REST Server over HTTP that calls your function for every collection in the database. There is no checking who is calling it or if they are permitted to.

Hmm. Fix the REST service that does not check access permissions in the first place? A library function should not be expected/required to do that,

exist-core/src/main/java/org/exist/collections/triggers/DocumentTriggers.java
adamretter: Not sure I understand.

exist-core/src/main/java/org/exist/xquery/functions/util/CollectionSuspendTriggers.java
adamretter: @chakl I would welcome a better system for that, please feel free to send a PR. At the moment however, functions are responsible for imposing their own security, see for example the FileModule of functions.

chakl:
closing, wrong approach

PR #3341, opened 2020-04-03
Add Code Coverage

adamretter:
Adds support for Code Coverage reports with JaCoCo and also integrates Travis with Coveralls.io to report historical use/change of Code Coverage.

**NOTE**: There is a bug that this surfaces that we need to fix - https://github.com/jacoco/jacoco/issues/1042

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -25,4 +25,9 @@ jobs:
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
       - name: Maven Test
-        run: mvn -V -B clean verify      
+        run: mvn -V -B clean verify
+      - name: Maven Code Coverage
+        if: ${{ matrix.jdk == '8' && matrix.os == 'ubuntu-latest' }}
+        env:
+          COVERALLS_TOKEN: ${{ secrets.COVERALLS_TOKEN }}
+        run: mvn -V -B jacoco:report coveralls:report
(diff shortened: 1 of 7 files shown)


adamretter:
Waiting on: https://github.com/jacoco/jacoco/issues/1042

line-o:
From the referenced jacoco issue:

> I suspect that this is caused by something in your tests or main code that use embedded Jetty - the above command won't be stuck and corruption of JaCoCo file also does not happen when `org.exist.management.JmxRemoteTest` and `org.exist.http.urlrewrite.URLRewritingTest` are excluded from execution.

dizzzz (approved):

duncdrum:
should we maybe spread the load and run this on the javadoc workflow, instead of the test workflow?

duncdrum (approved):
the secrete not being availble before we merge this is WAI on gh-actions
the crucial line is https://github.com/adamretter/exist/runs/2361117888?check_suite_focus=true#step:6:10

we can thing about adding this to another less heavy workflow when we refactor the flows, soits not a stopper for coverage

adamretter:
> run this on the javadoc workflow, instead of the test workflow?

That won't work, coverage is generated based on tests.

duncdrum:
@adamretter based on the presence of tests in the code-base or based on the results of executing the tests?

adamretter:
@duncdrum The tests have to be executed.

joewiz:
@adamretter According to @duncdrum the Coveralls integration with eXist-db is broken and needs some configuration fixes.

Issue #3346, opened 2020-04-08
[BUG] format-number throws IndexOutofBoundsException


line-o:
**Describe the bug**

Calling format-number with a certain picture string throws an java.lang.IndexOutOfBoundsException

`format-number($i, "#.###,##")`

**Expected behavior**

Either the appropriate XQuery-error is thrown or the number to be formatted according to the picture string.

**To Reproduce**

```Xquery
xquery version "3.1";

module namespace fd="http://exist-db.org/xquery/test/format-numbers";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare
    %test:args("12345.6")
    %test:args("#.###,##")
    %test:assertEquals("12.345,6")
function fd:simple-number-1($number as numeric, $picture) {
    format-number($number, $picture)
};
```

**Context (please always complete the following information):**
 - OS: macOS 10.15.4
 - eXist-db version: 5.3.0-SNAPSHOT
 - Java Version Java 1.8.0_152

**Additional context**
- How is eXist-db installed? built from commit: 093c4c0726e02868acadbce9583ec5809f462b33
- Any custom changes in e.g. `conf.xml`? no

**Stack Trace**

```
2020-04-08 14:54:48,995 [qtp1126112943-134] ERROR (XQueryServlet.java [process]:559) - null
java.lang.IndexOutOfBoundsException: null
	at org.exist.util.CodePointString.insert(CodePointString.java:293) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.functions.fn.FnFormatNumbers.format(FnFormatNumbers.java:633) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.functions.fn.FnFormatNumbers.eval(FnFormatNumbers.java:109) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.BasicFunction.eval(BasicFunction.java:74) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.InternalFunctionCall.eval(InternalFunctionCall.java:41) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
[...]

PR #3348, opened 2020-04-08
Fix bounds issues with group separators in fn:format-number

adamretter:
Closes https://github.com/eXist-db/exist/issues/3346

### DIFF
--- a/exist-core/src/test/xquery/numbers/format-numbers.xql
+++ b/exist-core/src/test/xquery/numbers/format-numbers.xql
@@ -246,6 +246,14 @@ function fd:simple-number-2($number, $picture) {
 };
 
 declare
+    %test:args(12345.6)
+    %test:args("#.###,##")
+    %test:assertEquals("12345.6")
+function fd:simple-number-3($number as numeric, $picture) {
+    format-number($number, $picture)
+};
+
+declare
     %test:args(14)
     %test:args("00%")
     %test:assertEquals("1400%")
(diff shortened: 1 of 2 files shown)


exist-core/src/test/xquery/numbers/format-numbers.xql
```diff
@@ -245,6 +245,14 @@ function fd:simple-number-2($number, $picture) {
     format-number($number, $picture)
 };
 
+declare
+    %test:args(12345.6)
+    %test:args("#.###,##")
+    %test:assertEquals("12345.6")
```
line-o, line 251: I just checked: Saxon 10 (HE) does output `12345.6` as well. I still think this is not correct because the spec does not say the grouping character has to be `,` and the decimal delimiter is `.`. Isn't it the idea of a picture string to change these defaults?

line-o: These are just the defaults as far as I understood.
Similar to the example in the spec: `format-number(1234.5678, '#ʹ##0·00', 'ch')`
If `format-number(1234.5678, '#,###.##', 'de')` is supported (Saxon 10 HE does not) I stop complaining.

adamretter: If you don't specify a specific default format, there is a *default decimal format* which is defined in the XQuery spec. Take a look at that.

line-o: @adamretter you are right. The relevant portion in the spec is statically known decimal formats: https://www.w3.org/TR/2017/REC-xquery-31-20170321/#dt-static-decimal-formats
Therefore the defaults '.' and ',' cannot be changed without providing the third parameter.

line-o:
What is it with this travis build? I restarted it once but it keeps failing.

adamretter:
@line-o Travis-CI uses Google's mirror of Maven Central. Unfortunately I uploaded some bad artifacts to Maven Central, which I was able to have repaired. I have a ticket open with Google to repair the artifacts in their mirror - should be fixed in the next 6 hours hopefully.

line-o:
@dizzzz do we really need @ljo or @wolfgangmm on this?

dizzzz:
Probably not but for me the code is a bit too complex :)

line-o:
@dizzzz while the code is really not easy to understand, the scope is limited to a xquery function which was added a) fairly recently b) by @adamretter himself :)

Issue #1964, opened 2018-06-23
fn:collection returns binary resources as document-node


joewiz:
### What is the problem

The `fn:collection` function returns binary resources as `document-node`.

### What did you expect

I expected that `fn:collection` would either (1) exclude binary resources from its results (as the XQuery 1.0-3.0 specs suggest) or (2) return them with the correct datatype of `xs:base64Binary` (as the XQuery 3.1 spec suggests). 

For context, see https://www.w3.org/TR/xpath-functions/#func-collection, which lists the following function signature for `fn:collection`:

> `fn:collection($arg as xs:string?) as item()*`

Also, see this note:

> In earlier releases, the primary use for the `fn:collection` function was to retrieve a collection of XML documents, perhaps held as lexical XML in operating system filestore, or perhaps held in an XML database. In this release the concept has been generalised to allow other resources to be retrieved: for example JSON documents might be returned as arrays or maps, non-XML text files might be returned as strings, and binary files might be returned as instances of `xs:base64Binary`...

In XQuery 3.0 and earlier, the return type for `fn:collection` was `node()*`.

Thus, if eXist conforms to XQuery 1.0-3.0, `fn:collection` **should not** include binary resources at all in the results. 

However, if eXist is following XQuery 3.1, `fn:collection` **could** return binary resources as `xs:base64Binary`. 

I think we should adopt the XQuery 3.1 behavior.

### Describe how to reproduce or add a test

A simple test:

```xquery
xquery version "3.1";

(: pick a collection that contains only binary resources, like dashboard's module collection:)
let $collection-name := "/db/apps/dashboard/modules"
let $collection := collection($collection-name)
let $binary-docs := 
    for $resource-name in xmldb:get-child-resources($collection-name) 
    return
        util:binary-doc($collection-name || "/" || $resource-name)
return
    (
        every $bin in $collection satisfies $bin instance of xs:base64Binary,
        every $bin in $collection satisfies $bin instance of document-node(),
        every $bin in $binary-docs satisfies $bin instance of xs:base64Binary,
        every $bin in $binary-docs satisfies $bin instance of document-node()
    )
```

The results, `false, true, true, false` shows that eXist's `fn:collection` function is returning binary resources as `document-node`, not as `xs:base64Binary`. On the other hand, using `util:binary-doc` to retrieve the same resources shows that `util:binary-doc` correctly returns these resources as `xs:base64Binary`, not as `document-node`.

An XQSuite demonstrating this:

```xquery
[...]

PR #3349, opened 2020-04-08
fn:collection should only return XML documents

adamretter:
Closes https://github.com/eXist-db/exist/issues/1964

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
@@ -25,14 +25,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import org.exist.collections.Collection;
-import org.exist.dom.persistent.DefaultDocumentSet;
-import org.exist.dom.persistent.DocumentImpl;
-import org.exist.dom.persistent.DocumentSet;
-import org.exist.dom.persistent.MutableDocumentSet;
-import org.exist.dom.persistent.NewArrayNodeSet;
-import org.exist.dom.persistent.NodeHandle;
-import org.exist.dom.persistent.NodeProxy;
-import org.exist.dom.persistent.NodeSet;
+import org.exist.dom.persistent.*;
 import org.exist.dom.QName;
 import org.exist.numbering.NodeId;
 import org.exist.security.PermissionDeniedException;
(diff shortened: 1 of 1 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
```diff
@@ -215,22 +208,26 @@ private URI asUri(final String path) throws XPathException {
...
-                    dlock = lockManager.acquireDocumentReadLock(doc.getURI());
-                }
-                result.add(new NodeProxy(doc));
-            } catch (final LockException e) {
-                throw new XPathException(this, ErrorCodes.FODC0002, e);
-            } finally {
-                if (dlock != null) {
-                    dlock.close();
+            // filter out binary documents, fn:collection should only return XML documents
+            if (doc.getResourceType() == DocumentImpl.XML_FILE) {
```
line-o, line 217: Since no code will be executed if this check fails I would reverse this check and return immediately.
`if (doc.getResourceType() == DocumentImpl.XML_FILE) { return; }`
The benefit is more readable, less indented code. Less context to keep in mind.

adamretter: I see your point, the flip-side though is that it will increase the cyclic complexity. Also I find `continue` to be a non-obvious thing

line-o: I personally am used to write and read return-early style code. I also doubt that cyclomatic complexity increases. On the other hand this is not important enough to block this PR.

line-o: I had a second look and must admit the complexity would raise by 1 because of the extra control-flow element.

mathias-goebel:
According to Semantic Versioning: Does this imply to update the major version number?
see [semver 2.0.0 rule 8](https://semver.org/#spec-item-8)

I know that I have some code somewhere utilizing the former behaviour of fn:collection#1 and this will be a **breaking change** to these packages. And may be I am not the only one.

line-o:
@mathias-goebel this is a topic of debate wether a change to the specification compliant behaviour is a breaking change.
I have argued in your favour before.
Could you describe your use case that involves non-XML files returned by `fn:collection()`?

mathias-goebel:
I have to admit that this was coded against the behavior of the processor, and not according to the specs. But we are using code like `collection()/base-uri()` to locate binary resources in the db (or any sub-collection). This code works without using a recursive function of `xmldb:get-child-(resources|collections)`. So we are reducing the number of function calls.

### Example 1
For example [here](https://gitlab.gwdg.de/SADE/SADE/-/blob/develop/modules/callgraph/inspect2graphml.xqm#L140) module imports and function calls are analyzed.

### Example 2
In addition to the XQSuite output I calculate a test coverage by simply checking function tested for and functions available. It also uses this method for getting the binary resources, see [here](https://gitlab.gwdg.de/SADE/SADE/-/blob/develop/post-install.xq#L19).

adamretter:
>  collection()/base-uri() to locate binary resources

This is much more expensive that using xmldb:get-child-*

line-o:
@adamretter I would think so too but it is a separate issue. The interesting questions here:
- Do we consider this a breaking change? You can make that argument
- If yes, do we revert that change in behaviour for a 5.3.0 release?

joewiz:
@line-o We discussed this at the time and agreed this PR constituted a bugfix. That is, eXist's implementation of the `collection()` function did not comply with the spec's function signature. Fixing bugs will almost always cause behavior that depended on the bug's presence to break. That's how we decided this PR should be treated as a bugfix for the current major version of eXist.

@mathias-goebel Are you requesting that this PR be reverted since it constitutes a burden to adjust your codebase to conform to the XQuery specification's function signature for the `collection()` function? In other words, the presence of this PR prevents you from updating to 5.3.0-SNAPSHOT (and a hypothetical minor/patch release based on it)?

joewiz:
@mathias-goebel I've opened an issue to track discussion of this and put it on the agenda for the next Community Call.

Issue #3337, opened 2020-03-30
[BUG] util:get-resource-by-absolute-id does not work


line-o:
**Describe the bug**

`util:get-resource-by-absolute-id` does return an empty sequence when given a valid absolute resource id.

**Expected behavior**

```xquery
util:get-resource-by-absolute-id(
    util:absolute-resource-id('/db/some/file'))
```

to return '/db/some/file' when this file exists in the database.

**To Reproduce**

```Xquery
xquery version "3.1";

module namespace abs="http://exist-db.org/xquery/absolute-resource-id";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare variable $abs:XML := document { <root/> };

declare
    %test:setUp
function abs:setup() {
    let $testCol := xmldb:create-collection("/db", "test")
    return xmldb:store("/db/test", "test.xml", $abs:XML)
};

declare
    %test:tearDown
function abs:tearDown() {
    xmldb:remove("/db/test")
};

declare
    %test:assertEquals('/db/test/test.xml')
function abs:test() {
    util:get-resource-by-absolute-id(
        util:absolute-resource-id('/db/test/test.xml'))
};
```

**Context (please always complete the following information):**
 - OS: OSX 10.15.4
 - eXist-db version: 5.3.0-SNAPSHOT 093c4c0726e02868acadbce9583ec5809f462b33
 - Java Version 1.8.0_152-b16

**Additional context**
- How is eXist-db installed? built from source 
- Any custom changes in e.g. `conf.xml`? none

Issue #3031, opened 2019-09-28
util:get-resource-by-absolute-id() is broken in 5.0.0


03b8:
Hi there. I noticed that the `util:get-resource-by-absolute-id()` doesn't return anything when provided with a valid id parameter in eXist 5.0.0. I compared the behavior of the function in 4.7.1 and 5.0.0 by using the two corresponding Docker images from Docker Hub and a short XQuery snippet:
```
let $doc := doc("/db/apps/eXide/build.xml")
(: $doc can also be any other document of course :)
let $abs-id := util:absolute-resource-id($doc)
return $doc = util:get-resource-by-absolute-id($abs-id)
```
This snippet:
* returns `true()` on eXist 4.7.1 from Docker Hub `existdb/existdb:4.7.1` (image id `1937aedf933b`)
* returns `false()` on eXist 5.0.0 from Docker Hub `existdb/existdb:5.0.0` (image id `72d679c573f1`)

My OS is Debian 9. Please let me know if there's other info you need.

duncdrum:
Yup i can reproduce this. @adamretter are there changes with respect to resource-ids as part of the lock table?

adamretter:
@duncdrum Nothing intentional that I am aware of.

line-o:
#3337 is a duplicate will add my test case here and then close mine.

PR #3350, opened 2020-04-08
Fix util:get-resource-by-absolute-id

adamretter:
Fix an issue with retrieving a resource by id from XQuery.

Closes https://github.com/eXist-db/exist/issues/3031
Closes https://github.com/eXist-db/exist/issues/3337


exist-core/src/main/java/org/exist/xquery/functions/util/DocumentNameOrId.java
```diff
@@ -106,98 +104,103 @@ public DocumentNameOrId(final XQueryContext context, final FunctionSignature sig
...
-                    } else {
-                        return new NodeProxy(doc);
-                    }
+    private Sequence getResourceByAbsoluteId(final IntegerValue ivAbsoluteId)
+            throws XPathException, PermissionDeniedException, LockException, IOException {
+        final BigInteger absoluteId = ivAbsoluteId.toJavaObject(BigInteger.class);
+        final Tuple3<Integer, Integer, Byte> decoded = decodeAbsoluteResourceId(absoluteId);
+
+        final DocumentImpl doc = context.getBroker().getResourceById(decoded._1, decoded._3, decoded._2);
+        if (doc != null) {
```
line-o, line 174: Could this be rewritten as:
```java
        if (doc == null) { return Sequence.EMPTY_SEQUENCE; }

        try (final ManagedDocumentLock docLock = context.getBroker().getBrokerPool().getLockManager().acquireDocumentReadLock(doc.getURI())) {
            if (doc instanceof BinaryDocument) {
                final BinaryDocument bin = (BinaryDocument) doc;
                final InputStream is = context.getBroker().getBinaryResource(bin);
                return Base64BinaryDocument.getInstance(context, is);
            }
            return new NodeProxy(doc);
        }

```

exist-core/src/main/java/org/exist/xquery/functions/util/DocumentNameOrId.java
```diff
@@ -1,215 +1,206 @@
...
-            absoluteId = absoluteId.shiftLeft(1);
-            absoluteId = absoluteId.or(BigInteger.valueOf(doc.getResourceType() & 1));
-            result = new IntegerValue(absoluteId, Type.INTEGER);
-        } else {
-            result = Sequence.EMPTY_SEQUENCE;
+    private @Nullable <T> T getFromDocument(final XmldbURI docUri, final Function<DocumentImpl, T> docFn) throws PermissionDeniedException {
+        try (final LockedDocument lockedDoc = context.getBroker().getXMLResource(docUri, LockMode.READ_LOCK)) {
+            if (lockedDoc == null) {
+                return null;
+            } else {
```
line-o, line 158: Is else necessary here? Since the if-statement will always return when lockedDoc is null.

adamretter: It's not necessary, but it's a coding style that I follow. When in more complex or nested code, always express the `else` path to make the program flow easier to follow.

exist-core/src/main/java/org/exist/xquery/functions/util/DocumentNameOrId.java
adamretter: Yes it could, but that is not clearer to me.

We seem to spend a lot of time lately discussing code-style. Perhaps if this is a concern we should just adopted a standard like: https://google.github.io/styleguide/javaguide.html. Otherwise we are just arguing subjectively about personal taste.

line-o: Good topic for Monday 👍

exist-core/src/main/java/org/exist/xquery/functions/util/DocumentNameOrId.java
line-o: We are looking at a single statement. 🕶️

Issue #1665, opened 2017-12-27
Missing XQST0034 check


adamretter:
```xquery
declare function local:thing1() {
    <a/>
};

declare function local:thing2() {
    <thing2/>
};

declare function local:thing1() {
    <b/>
};


local:thing1()
```

This should result in `XQST0034`, however eXist compiles it just fine. It then **somehow** chooses which `thing1` to run :-(

duncdrum:
see #1174 for a related problem

Issue #1174, opened 2016-11-07
Multiple Modules with the same Target Namespace


gspringtech:
This is not handled by eXist. Only functions in the first location are recognised in the query.
Given two modules with the same namespace and two functions (test1, test2):

    import module namespace t = "http://example.com/test" at "test1.xqm", "test2.xqm";
    (t:test1(), t:test2())

duncdrum:
@gspringtech could you point us to the xquery specs section that say this should be supported?

@joewiz thoughts?

gspringtech:
@duncdrum 
https://www.w3.org/TR/xquery-31/#id-module-handling-multiple-same

dizzzz:
@gspringtech thnx; clear!

joewiz:
#3354 was orthogonal to this issue. This issue points out that eXist does not support multiple location hints following the `at` keyword in a module import.

joewiz:
Take, for example, this main module:

**main.xq**
```xquery
xquery version "3.1";

import module namespace m="http://www.w3.org/TestModules/m30/test" at "test1.xq", "test2.xq"; 

$m:one + $m:two
```

And the following two library modules:

**test1.xq**

```xquery
xquery version "3.1";

module namespace t1="http://www.w3.org/TestModules/m30/test";

declare variable $t1:one := 1;
```

**test2.xq**

```xquery
xquery version "3.1";

module namespace t2="http://www.w3.org/TestModules/m30/test";

declare variable $t2:two := 2;
```

eXist will throw an exception when evaluating main.xq:

> Cannot compile xquery: err:XPDY0002 variable '$m:two' is not set. 

Thus, as @gspringtech originally stated, eXist is only importing the first library module, not the second.

This is using eXist 5.3.0-SNAPSHOT 97e80442aa438558b3dfb107ad16b354437b4582 20201021052237.

PR #3354, opened 2020-04-09
Prohibit duplicate function declarations

adamretter:
Prohibit functions with the same name and arity as per the XQuery spec.

Closes https://github.com/eXist-db/exist/issues/1665
Closes https://github.com/eXist-db/exist/issues/1174

### DIFF
--- a/exist-core/src/test/xquery/xquery3/arrowop.xql
+++ b/exist-core/src/test/xquery/xquery3/arrowop.xql
@@ -237,14 +237,14 @@ function ao:fold-left-with-contextitem () {
 
 declare
     %test:assertEquals('dcba')
-function ao:fold-right-with-contextitem () {
+function ao:fold-right-with-contextitem-1 () {
     (<a/>,<b/>,<c/>,<d/>)
         => fold-right('', function ($a, $r) { $r || xs:string(node-name($a)) })
 };
 
 declare
     %test:assertEquals('dcba')
-function ao:fold-right-with-contextitem () {
+function ao:fold-right-with-contextitem-2 () {
     ('a','b','c','d')
         => fold-right('', function ($a, $r) { $r || $a })
 };
--- a/exist-core/src/main/java/org/exist/dom/QName.java
+++ b/exist-core/src/main/java/org/exist/dom/QName.java
@@ -246,7 +246,9 @@ public class QName implements Comparable<QName> {
 
     @Override
     public int hashCode() {
-        return namespaceURI.hashCode() ^ localPart.hashCode();
+        int result = namespaceURI.hashCode();
+        result = 31 * result + localPart.hashCode();
+        return result;
     }
 
     public javax.xml.namespace.QName toJavaQName() {
--- a/exist-core/src/main/java/org/exist/interpreter/Context.java
+++ b/exist-core/src/main/java/org/exist/interpreter/Context.java
@@ -443,11 +443,9 @@ public interface Context {
      * @param argCount the function arity
      *
      * @return the resolved function, or null
-     *
-     * @throws XPathException if an error occurs whilst resolving the function
      */
     @Nullable
-    UserDefinedFunction resolveFunction(QName name, int argCount) throws XPathException;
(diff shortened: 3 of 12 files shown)


exist-core/src/test/xquery/xquery3/arrowop.xql
```diff
@@ -237,14 +237,14 @@ function ao:fold-left-with-contextitem () {
...
-function ao:fold-right-with-contextitem () {
+function ao:fold-right-with-contextitem-1 () {
     (<a/>,<b/>,<c/>,<d/>)
         => fold-right('', function ($a, $r) { $r || xs:string(node-name($a)) })
 };
 
 declare
     %test:assertEquals('dcba')
-function ao:fold-right-with-contextitem () {
+function ao:fold-right-with-contextitem-2 () {
```
line-o, line 247: I see what you found here :)

line-o:
This is a very nice bugfix. I am cautious to approve it only because I cannot know implications of:

- Object2ObjectRBTreeMap
- Object2ObjectAVLMap
- when combining hashes the first part is multiplied by 31

I hope that @adamretter you could either explain why you chose those or somebody who understands approves.

adamretter:
@line-o to explain:

> I am cautious to approve it only because I cannot know implications of:
>
> * Object2ObjectRBTreeMap
> * Object2ObjectAVLMap

These are alternatives to Java's TreeMap. They should offer better performance and lower memory-use. I chose them based on their performance characteristics and our needs. You can read more about the performance trade-offs here - http://fastutil.di.unimi.it/docs/#Performance

> *when combining hashes the first part is multiplied by 31

This is kinda standard stuff for hashes... 31 is a prime number ;-)

line-o:
Thanks for the explanation. I found this magic prime in a long discussion over at stackoverflow.
I would love to hear why the o2o by-reference binary tree is used in one place and the AVL one in another, but I will use this as a CS exercise for later.

line-o:
When I have time and the tests pass locally I might merge before Travis is fixed... whatever comes first.

line-o:
So just for reference:

- RBTree (red-black tree) is good for Maps with frequent adds and deletions
- AVLTree is better for frequent searches and less frequent additions and deletions

So `declaredFunctions` seems to be a perfect fit for an AVLTree ... things starting to make sense

line-o:
@adamretter are additions and deletions so frequent for global variables?

line-o:
@adamretter I would like to know why you chose RBTree over AVLTree for global variables.

adamretter:
@line-o From the fastutil doc:

> So on small trees red-black trees could be faster, but on very large sets AVL trees will shine

I figured that there would be very few global variables generally.

line-o:
Thanks for the clarification. Without data we can only guess. 
It is very likely, that there will be less global variables than functions declared. It would be helpful to know where the boundary is where AVLTrees start to be faster. If we are talking millions, then RBTree is better for functions, too.

line-o:
Both trees will never exceed 10K (another informed guess)

adamretter:
I doubt they would exceed 1K

On Sun, 12 Apr 2020 at 16:14, Juri Leino <notifications@github.com> wrote:
>
> Both trees will never exceed 10K (another informed guess)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.



-- 
Adam Retter

skype: adam.retter
tweet: adamretter
http://www.adamretter.org.uk

line-o:
Is 1K very large? Large enough to benefit from AVLTree?

adamretter:
@line-o No. But it is larger than what I thought would be small for the RBTree. Maybe better to discuss by chat if you still have questions

line-o:
The implementation of Java HashMap uses RBTrees since ... wait for it ... version 8 
source: http://coding-geek.com/how-does-a-hashmap-work-in-java/#JAVA_8_improvements

adamretter:
@line-o That may be the case, but they are still more memory hungry than fastutil

Issue #2798, opened 2019-06-10
[BUG] string-length() after simple map operator


djbpitt:
Thank you for reporting your issue!

To be able to better understand you problem or suggestion, please add as much information as possible to this ticket. When there is no or just little information, it will be too difficult to work on it (and hence we will not be able to solve it)

Please fill in the following sections:

### What is the problem
> Describe exactly what you see (e.g. an output of an XQuery) 

In eXide, `1 ! string-length(.)` incorrectly returns `1`. It should raise a type error. 

### What did you expect
> Describe what you expected to happen. Add for example a reference to a [specification](https://www.w3.org/TR/xquery-3/).

`1 ! string-length()` (no dot inside the parentheses) correctly returns `1`. The reason different behavior with and without the dot is expected is explained at <https://www.biglist.com/lists/lists.mulberrytech.com/xsl-list/archives/201906/msg00021.html>. Saxon correctly implements the different behavior. And eXist correctly raises a type error with `string-length(1)`.

### Describe how to reproduce or add a test
> Describe how we can can reproduce the problem.

Type `1 ! string-length(.)` into eXide and press the Eval button.

### Context information
Please always add the following information
- eXist-db version + Git Revision hash e.g. eXist-db 3.0 /  acd0c14 : exist-db 4.4
- Java version (e.g. Java8u121) : Java 8
- Operating system (Windows 7, Linux, MacOs) : Centos 6
- 32 or 64 bit : 64 bit
- How is eXist-db installed? (JAR installer, DMG, WAR file, clone from GitHub) : jar
- Any custom changes in e.g. conf.xml : no

adamretter:
Unfortunately this is related to eXist-db's bogus handling of `.` (i.e. the context item), see - https://github.com/eXist-db/exist/issues/2308#issuecomment-443990266

line-o:
exist 5.3.0 is affected

PR #3355, opened 2020-04-09
Context Item should not be atomized when used as argument for fn:string-length

adamretter:
Closes https://github.com/eXist-db/exist/issues/2798

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/Function.java
+++ b/exist-core/src/main/java/org/exist/xquery/Function.java
@@ -317,14 +317,13 @@ public abstract class Function extends PathExpr {
 
         if (!cardinalityMatches) {
             cardinalityMatches = argument.getCardinality().isSubCardinalityOrEqualOf(argType.getCardinality());
-            if (!cardinalityMatches) {
-                if (argument.getCardinality() == Cardinality.EMPTY_SEQUENCE
-                        && argType.getCardinality().atLeastOne()) {
-                    throw new XPathException(this,
-                            ErrorCodes.XPTY0004,
-                            Messages.getMessage(Error.FUNC_EMPTY_SEQ_DISALLOWED,
-                                    argPosition, ExpressionDumper.dump(argument)));
-                }
+        }
+
+        if (!cardinalityMatches) {
+            if (argument.getCardinality() == Cardinality.EMPTY_SEQUENCE && argType.getCardinality().atLeastOne()) {
+                throw new XPathException(this,  ErrorCodes.XPTY0004,
+                        Messages.getMessage(Error.FUNC_EMPTY_SEQ_DISALLOWED,
+                                argPosition, ExpressionDumper.dump(argument)));
             }
         }
 
--- a/exist-core/src/main/java/org/exist/xquery/Function.java
+++ b/exist-core/src/main/java/org/exist/xquery/Function.java
@@ -24,6 +24,7 @@ package org.exist.xquery;
 import java.lang.reflect.Constructor;
 import java.util.List;
 
+import com.evolvedbinary.j8fu.tuple.Tuple2;
 import org.exist.dom.QName;
 import org.exist.xquery.parser.XQueryAST;
 import org.exist.xquery.util.Error;
@@ -33,6 +34,10 @@ import org.exist.xquery.value.Item;
 import org.exist.xquery.value.Sequence;
 import org.exist.xquery.value.SequenceType;
 import org.exist.xquery.value.Type;
+
+import javax.annotation.Nullable;
+
+import static com.evolvedbinary.j8fu.tuple.Tuple.Tuple;
 
 /**
  * Abstract base class for all built-in and user-defined functions.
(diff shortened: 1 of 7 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/FunStrLength.java
```diff
@@ -38,64 +32,90 @@
...
-        return result;            
-	}
+            if (contextSequence != null) {
+                context.getProfiler().message(this, Profiler.START_SEQUENCES, "CONTEXT SEQUENCE", contextSequence);
+            }
+            if (contextItem != null) {
+                context.getProfiler().message(this, Profiler.START_SEQUENCES, "CONTEXT ITEM", contextItem.toSequence());
+            }
+        }
+
```
line-o, line 96: surplus blank line

exist-core/src/main/java/org/exist/xquery/functions/fn/FunStrLength.java
```diff
@@ -38,64 +32,90 @@
...
+
+        if (contextItem != null) {
+            contextSequence = contextItem.toSequence();
+        }
+
```
line-o, line 101: surplus blank line

exist-core/src/main/java/org/exist/xquery/functions/fn/FunStrLength.java
adamretter: Seriously... Is that really a problem? For me - It helps to break up the code.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunStrLength.java
adamretter: As commented above.

exist-core/src/main/java/org/exist/xquery/Function.java
```diff
@@ -216,147 +223,191 @@ public void setArguments(final List<Expression> arguments) throws XPathException
...
-        return expr;
+        return argument;
+    }
+
+    protected boolean checkArgumentTypeCardinality(final Expression argument, @Nullable final SequenceType argType,
+            final int argPosition) throws XPathException {
+        boolean cardinalityMatches = argument instanceof VariableReference
+                || argType.getCardinality() == Cardinality.ZERO_OR_MORE;
+
+        if (!cardinalityMatches) {
```
line-o, line 318: Can this be rewritten as
```
if (!cardinalityMatches) {
    ....
}
if (!cardinalityMatches) {
    ....
}
```
?

exist-core/src/main/java/org/exist/xquery/functions/fn/FunSubSequence.java
```diff
@@ -65,21 +65,32 @@ public FunSubSequence(final XQueryContext context, final FunctionSignature signa
...
+
+            // call analyze for each argument
             final AnalyzeContextInfo argContextInfo = new AnalyzeContextInfo(contextInfo);
-            getArgument(i).analyze(argContextInfo);
+            arg.analyze(argContextInfo);
             if (i == 0) {
                 contextInfo.setStaticReturnType(argContextInfo.getStaticReturnType());
             }
+
+            if (!argumentsChecked) {
```
line-o, line 84: This block appear three times in the changed files. Would it be possible to extract it into a method, function that can be re-used?

exist-core/src/main/java/org/exist/xquery/functions/fn/FunStrLength.java
line-o: Well, if you ask me I would auto format any file in a pre-commit hook and would never have to care about these. It is very tedious.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunStrLength.java
line-o: If this was left intentionally _and_ there is no code style / linter / other dev having a problem with it, leave it. It is not a problem per se. My first thought was: Oh, this must have been missed.

joewiz:
All 4 of Travis's linux builds failed...

exist-core/src/main/java/org/exist/xquery/functions/fn/FunSubSequence.java
adamretter: It doesn't make sense as these are function specific (the override the default in `Function`), and actually I think it only appears twice the same, the other instance is slightly different.

exist-core/src/main/java/org/exist/xquery/Function.java
adamretter: Good spot. I have fixed that.

exist-core/src/main/java/org/exist/xquery/Function.java
```diff
@@ -317,14 +317,13 @@ protected boolean checkArgumentTypeCardinality(final Expression argument, @Nulla
...
-                if (argument.getCardinality() == Cardinality.EMPTY_SEQUENCE
-                        && argType.getCardinality().atLeastOne()) {
-                    throw new XPathException(this,
-                            ErrorCodes.XPTY0004,
-                            Messages.getMessage(Error.FUNC_EMPTY_SEQ_DISALLOWED,
-                                    argPosition, ExpressionDumper.dump(argument)));
-                }
+        }
+
+        if (!cardinalityMatches) {
```
line-o, line 322: cool, ... 😜 I would now combine the nested if-statements ... but that can also be addressed separately.

line-o:
I would like to change the title of this PR to
> Context Item should not be atomized when used as function argument

adamretter:
@line-o Why? This only applies to `fn:string-length`!

line-o:
Maybe I am mislead because of the added tests with number being called after the simple map operator

adamretter:
@line-o Simple Map Operator allows easy manipulation of the Context Item ;-)

line-o:
@adamretter The implementations of Function, FunHigherOrderFun, FunSubsequence _and_ FunStrLength changed. There is also a subtle but neglectable change in PathExpr.
None of this is obvious reading the title of this PR nor it is mentioned in the description in any way.
I just went through the commit messages and even there I can find no mentions of the new methods added to the Function class.
It is very helpful for all maintainers to get a grasp of concepts and ideas hidden in the code.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
```diff
@@ -102,13 +104,14 @@ public FunHigherOrderFun(XQueryContext context, FunctionSignature signature) {
 	}
 
     @Override
-    protected void checkArguments() throws XPathException {
+	protected void checkArgument(final Expression arg, @Nullable final SequenceType argType,
+			final AnalyzeContextInfo argContextInfo, final int argPosition) throws XPathException {
         // hack: order of parameters for filter and other functions has changed
```
line-o, line 109: Is this comment still valid?

adamretter:
@line-o checkArguments is only needed in very exceptional circumstances, I think it is 100% unlikely that anyone ever overrode this in any of their own custom functions written in Java. It isn't really considered a public API.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
adamretter: No idea. I didn't add the comment or implement the filter stuff - you would have to ask the author of that stuff.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunStrLength.java
```diff
@@ -38,64 +32,90 @@
...
+                See https://github.com/eXist-db/exist/issues/2798
+             */
+            return Tuple(argument, returnType);
+        } else {
+            return super.strictCheckArgumentType(argument, argType, argContextInfo, argPosition, returnType);
+        }
+    }
+
+    @Override
+    public Sequence eval(Sequence contextSequence, final Item contextItem) throws XPathException {
```
dizzzz, line 85: contextSequence can be final?

exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
line-o: But it _clearly_ does not comment what you added and that is the entire method body. Please remove it as it will only confuse future generations.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
```diff
@@ -102,13 +104,14 @@ public FunHigherOrderFun(XQueryContext context, FunctionSignature signature) {
...
         // in final XQ3 spec. This would cause some core apps (dashboard) to stop
         // working. We thus switch parameters dynamically until all users can be expected to
         // have updated to 2.2.
         if (!isCalledAs("filter")) {
```
line-o, line 113: Can this be removed?

line-o: this was added in 2014 -- 0ecdf809530818ae1f8c74d5e7cb1ac80fa70d47
We should test, if this override can be removed entirely.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunStrLength.java
```diff
@@ -38,64 +32,90 @@
...
+        }
+
+
+        if (contextItem != null) {
+            contextSequence = contextItem.toSequence();
+        }
+
+
+        if (getSignature().getArgumentCount() == 1) {
+            contextSequence = getArgument(0).eval(contextSequence);
```
line-o, line 104: @dizzz no contextSequence may mutate

PR #3361, opened 2020-04-11
[fix] remove deprecated fn:filter syntax

### LINKED ISSUE #3370: [BUG] expathrepo uses fn:filter with flipped arguments
After #3361 the below code will throw

```xquery
    exists(filter(function($pkg) { $pkg = $name }, repo:list()))
```

https://github.com/eXist-db/exist/blob/961666edaa28b9c2430749e1a54f95f1a3febe87/extensions/modules/expathrepo/src/main/resources/org/exist/xquery/modules/expathrepo/repair.xql#L67

and 

```xquery
    filter(function($name) {
        $name = ("expath-pkg.xml", "repo.xml", "exist.xml") or starts-with($name, "icon")
    }, xmldb:get-child-resources($collection))
```

https://github.com/eXist-db/exist/blob/961666edaa28b9c2430749e1a54f95f1a3febe87/extensions/modules/expathrepo/src/main/resources/org/exist/xquery/modules/expathrepo/repair.xql#L79

line-o:
### Description:

Six years ago the spec for fn:filter changed.
As it was already used in some applications like
dashboard, existdb allowed both ways to call
fn:filter.
This commit removes the special handling, where
the parameters are flipped. This will now throw
an XPTY0004 for wrong parameter type.

fixes #3370

### Reference:

Our own docs do not even mention that you could flip the arguments for fn:filter

https://exist-db.org/exist/apps/fundocs/view.html?uri=http://www.w3.org/2005/xpath-functions&location=java:org.exist.xquery.functions.fn.FnModule&details=true

### Type of tests:

One test was adapted to check for the expected error.

### DIFF
--- a/exist-core/src/test/xquery/xquery3/higher-order.xml
+++ b/exist-core/src/test/xquery/xquery3/higher-order.xml
@@ -476,11 +476,11 @@ return
     </test>
     <test output="text">
         <task>fn:filter function (wrong argument order: deprecated)</task>
-        <code>fn:filter(function($a) {$a mod 2 = 0}, 1 to 10)</code>
-        <expected>2 4 6 8 10</expected>
+        <code>fn:filter(function($a) {$a mod 2 = 0}, (1 to 10))</code>
+        <error>XPTY0004</error>
     </test>
     <test output="text">
-        <task>fn:filter function (correct argument order)</task>
+        <task>fn:filter function</task>
         <code>fn:filter(1 to 10, function($a) {$a mod 2 = 0})</code>
         <expected>2 4 6 8 10</expected>
     </test>
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
@@ -104,18 +104,6 @@ public class FunHigherOrderFun extends BasicFunction {
 	}
 
     @Override
-	protected void checkArgument(final Expression arg, @Nullable final SequenceType argType,
-			final AnalyzeContextInfo argContextInfo, final int argPosition) throws XPathException {
-        // hack: order of parameters for filter and other functions has changed
-        // in final XQ3 spec. This would cause some core apps (dashboard) to stop
-        // working. We thus switch parameters dynamically until all users can be expected to
-        // have updated to 2.2.
-        if (!isCalledAs("filter")) {
-            super.checkArgument(arg, argType, argContextInfo, argPosition);
-        }
-    }
-
-    @Override
 	public void analyze(AnalyzeContextInfo contextInfo) throws XPathException {
 		cachedContextInfo = new AnalyzeContextInfo(contextInfo);
         super.analyze(cachedContextInfo);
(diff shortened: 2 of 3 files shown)


line-o:
I just checked: The current function signature for fn:filter does not allow the flipped order of parameters.
That means this does not deprecate anything but merely cleans up the code and can be safely merged into the upcoming release.

exist-core/src/test/xquery/xquery3/higher-order.xml
```diff
@@ -475,12 +475,7 @@ return
         <expected>23 29</expected>
     </test>
     <test output="text">
-        <task>fn:filter function (wrong argument order: deprecated)</task>
-        <code>fn:filter(function($a) {$a mod 2 = 0}, 1 to 10)</code>
-        <expected>2 4 6 8 10</expected>
-    </test>
-    <test output="text">
-        <task>fn:filter function (correct argument order)</task>
+        <task>fn:filter function</task>
```
adamretter, line 478: Maybe don't remove the test - but change it so that it is expected to fail. This proves then that your changes are complete :-)

line-o: OK, will do.

line-o:
I was able to find to instances where the deprecated signature of fn:filter is still in use.
I opened to issues to keep track of them. This PR should likely wait for those to be addressed.
Search query used to find them: `org:eXist-db language:XQuery filter`

joewiz:
@line-o Great job hunting the cases of the incorrect code in the eXist-db organization. I’m a little unclear about your suggestion to hold the PR but still consider this PR for 5.2.1. If these other repos need to be fixed, then wouldn’t other similar code from outside the org break if a 5.2.0 system upgrades to a new 5.x release containing this PR?

line-o:
@joewiz yes they would break, if they call filter in a way that was never standards compliant.
Calling `filter` with its arguments flipped was formally deprecated. The test clearly states that.
I see this as a similar case as what we decided on the deprecated map-literal syntax with `:=`.

Also the fact, that I had to update my test suite for xbow, because duplicate functions now throw could be seen as a breaking change. We may need to revert https://github.com/eXist-db/exist/pull/3354, if we want to be that strict.

joewiz:
@line-o Thanks for that info. One correction: People could have and, indeed, did write code according to the *then* draft of the spec—so it's not really accurate to say that they the code was *never* standards compliant. But this PR is valuable and we want it to get into the eXist source base. But the potential for broken functional code means we handle this by (a) deprecating it in the current major version and (b) removing it in the next major version. So this PR has to wait for a major release. This has been the guiding principle and practice (though no one claims we've been perfect) since we adopted semver with eXist 3.1.0 (see https://exist-db.org/exist/apps/wiki/blogs/eXist/eXist310). 

As to #3354, I see your point. Rather than continuing discussion in this PR, I've added this item to the next community call's agenda.

joewiz:
To illustrate this principle/practice, recall how we approached the removal of the draft map constructor syntax. Issue: https://github.com/eXist-db/exist/issues/2887. PR: https://github.com/eXist-db/exist/pull/2890.

joewiz:
I've restarted the failed Travis checks.

dizzzz:
All are ok !

duncdrum:
@line-o can you rebase this please

Issue #1176, opened 2016-11-10
Comparison bug with large xs:integer values


median1st:
The FLWOR expression fails to order input tuples of xs:integer type and a large value. The output tuples are ordered by just a certain portion of the number digits taken from left.
Steps to reproduce - run this XQuery fragment:
```
let $data := (
310000000000920000300,
310000000000930001800,
310000000000930003900,
310000000000970000400,
310000000000930002800,
310000000000920000200,
310000000000930001700,
310000000000930004900,
310000000000930003800,
310000000000970000300
    )
for $i in $data
order by $i 
return $i
```
The result run on version 2.2:

> 310000000000920000300
> 310000000000920000200
> 310000000000930001800
> 310000000000930003900
> 310000000000930002800
> 310000000000930001700
> 310000000000930004900
> 310000000000930003800
> 310000000000970000400
> 310000000000970000300

When the query is changed to:
```
let $data := (
310000000000920000300,
310000000000930001800,
310000000000930003900,
310000000000970000400,
310000000000930002800,
310000000000920000200,
310000000000930001700,
310000000000930004900,
310000000000930003800,
310000000000970000300
    )
for $i in $data
(:  order by $i :)
order by $i div 10000000, $i mod 10000000
return $i
```
the result is correct:
> 310000000000920000200
> 310000000000920000300
> 310000000000930001700
> 310000000000930001800
> 310000000000930002800
> 310000000000930003800
> 310000000000930003900
> 310000000000930004900
> 310000000000970000300
> 310000000000970000400

joewiz:
I get the same results for both queries on 3.0RC2 (latest from the develop branch).

median1st:
That's great, but it does not solve my problem. I'm actually implementing a serious application using eXist, so I'm using the version 2.2 that has been tagged as "production stable". The 3.0RC2 is not yet considered stable for production.
The mentioned application should actually prove to the management that XMLDB is the right infrastructure and service for the data. If that's shown, the company would buy the support. But so far I'm stumbling over bugs in a stable release (this one is the less serious as it can be worked around).

joewiz:
FYI, I'm just a user. I didn't claim to be offering to solve your problem - just adding context to advance your report. Specifically, I was saying that the problem *still affects* 3.0RC2, the version currently in development. If you are thinking about commercial support from existsolutions.com, you should contact them via http://existsolutions.com/contact.xml, not via this issue tracker.

median1st:
[...]

Issue #3330, opened 2020-03-27
[BUG] exist allows downcasts of xs:integer


line-o:
**Describe the bug**

It is not spec compliant to allow 
```xquery
function(xs:long)
function(xs:int)
function(xs:short)
function(xs:byte)
```
to be called with a literal number. 
According to the discussion in this [SO-question]( https://stackoverflow.com/questions/60877203/why-can-a-numeric-literal-never-be-of-type-xspositiveinteger-in-xquery) this is not spec-compliant.

The relevant portion of Michael Kay's answer:

> Then §3.1.1 in the XQuery spec talks about numeric literals:
>> The value of a numeric literal containing no "." and no e or E character is an atomic value of type xs:integer.

> §3.18.1 gives the rules for the "instance of" operator:
>> The boolean operator instance of returns true if the value of its first operand matches the SequenceType in its second operand, according to the rules for SequenceType matching;

> and §2.5.5.2 gives the relevant rule for SequenceType matching:
>> An ItemType consisting simply of an EQName is interpreted as an AtomicOrUnionType. The expected type AtomicOrUnionType matches an atomic value whose actual type is AT if derives-from( AT, AtomicOrUnionType ) is true.

**Expected behavior**

Calling a function with one the above type-signatures with a literal number like `1` should raise an error (`err:XPTY0004`, "cannot cast xs:integer to T"). 

**To Reproduce**

```Xquery
xquery version "3.1";

module namespace t="http://exist-db.org/xquery/test-downcast";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare function t:long ($n as xs:long) { $n };
declare function t:int ($n as xs:int) { $n };
declare function t:short ($n as xs:short) { $n };
declare function t:byte ($n as xs:byte) { $n };

declare variable t:fun-seq := (t:long#1, t:int#1, t:short#1, t:byte#1)

declare
    %test:param(1)
    %test:param(2)
    %test:param(3)
    %test:param(4)
    %test:assertError('XPTY0004')
function t:test($index) { t:fun-seq[$index](1) };
```

**Context (please always complete the following information):**
 - OS: macOS 10.15.4
 - eXist-db version: 5.3.0-SNAPSHOT
 - Java Version 1.8.0_152-b16

**Additional context**
- How is eXist-db installed? from git 093c4c0726e02868acadbce9583ec5809f462b33
- no custom changes

Issue #1715, opened 2018-02-04
Add support for the xs:dateTimeStamp datatype


joewiz:
### What is the problem

eXist supports all datatypes except one: the `xs:dateTimeStamp` datatype. From the XML Schema Datatypes spec:

> all values of dateTimeStamp are required to have explicit time zone offsets and the datatype is totally ordered.

So the following query should return `true()`:

```xquery
exists(fn:timezone-from-dateTime(xs:dateTimeStamp("2011-07-28T12:34:56-08:00")))
```

Instead, eXist returns the following error:

> Cannot compile xquery: exerr:ERROR Type: xs:dateTimeStamp is not defined

Similarly, eXist should complain about the missing explicit timezone offset in `xs:dateTimeStamp("2011-07-28T12:34:56")` but instead raises the error above.

### What did you expect

I expected eXist to support all datatypes defined at https://www.w3.org/TR/xpath-functions-31/#atomic-type-hierarchy and https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/datatypes.html#dateTimeStamp.

### Describe how to reproduce or add a test

See https://github.com/w3c/qt3tests/blob/master/xs/dateTimeStamp.xml.  This was the source of the sample query above.

### Context information

- eXist-db version + Git Revision hash: 3.7.0-SNAPSHOT develop
- Java version: 1.8.0_152
- Operating system: macOS 10.13.3
- 32 or 64 bit: 64 bit
- Any custom changes in e.g. conf.xml: none

joewiz:
There's actually 3 more unimplemented datatype in eXist: [`xs:NMTOKENS`](https://www.w3.org/TR/xmlschema-2/#NMTOKENS), [`xs:IDREFS`](https://www.w3.org/TR/xmlschema-2/#IDREFS), and [`xs:ENTITIES`](https://www.w3.org/TR/xmlschema-2/#ENTITIES) - which [the XQuery spec](https://www.w3.org/TR/xpath-functions-31/#constructor-functions-for-xsd-list-types) calls "the three built-in list types defined in XML Schema Part 2: Datatypes Second Edition." 

I came across this after reading https://github.com/BaseXdb/basex/issues/1691. I performed Christian's test in eXide and got this error:

> Cannot compile xquery: exerr:ERROR Type: xs:NMTOKENS is not defined

This led me to look up the types above.

adamretter:
[...]

PR #3363, opened 2020-04-13
Fix issues with types and comparisons

adamretter:
Closes https://github.com/eXist-db/exist/issues/1176

Cleans up value and general comparisons to bring them inline with the XQuery spec. 

Some highlights:

1. eXist-db's Type Hierarchy had a bit of a nasty hack around numeric types because it was unable to represent union types. Previously eXist-db modelled `xs:decimal`, `xs:float`, and `xs:double`, as sub-types of the union type `xs:numeric` which was incorrect. The XDM specifies that they are sub-types of `xs:anyAtomicType`, and that `xs:numeric` is a union type whose members are the set of `xs:decimal`, `xs:float`, and `xs:double`. This PR adds support for modelling Union Types from the XDM and corrects the type hierarchy in eXist-db. See - https://www.w3.org/TR/xpath-datamodel-31/#types-hierarchy

2. When performing Value and General comparisons, eXist-db's type coercion was incorrect. The XQuery spec specifies the following algorithms: [Value Comparisons](https://www.w3.org/TR/xquery-31/#id-value-comparisons), and [General Comparisons](https://www.w3.org/TR/xquery-31/#id-general-comparisons). This PR replaces the existing type coercion in eXist-db with a new version which is inline with the XQuery specification.

3. Comparisons between the values backing numbers in eXist-db were always down-cast to a Java `double` value. This meant a loss of precision when comparing larger numbers. This PR replaces that with an up-cast and lambda pass method; This accommodates the smallest common Java type, thus ensuring no truncation/rounding of numeric data.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
+++ b/exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
@@ -1004,7 +1004,7 @@ public class GeneralComparison extends BinaryOp implements Optimizable, IndexUse
                     i. If T is a numeric type or is derived from a numeric type,
                     then V is cast to xs:double.
                  */
-                if (Type.subTypeOf(otherType, Type.NUMBER)) {
+                if (Type.subTypeOfUnion(otherType, Type.NUMBER)) {
                     return value.convertTo(Type.DOUBLE);
                 }
 
--- a/exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
+++ b/exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
@@ -57,6 +57,7 @@ import java.util.List;
  *
  * @author  wolf
  * @author  andrzej@chaeron.com
+ * @author <a href="mailto:adam@evolvedbinary.com">Adam Retter</a>
  */
 public class GeneralComparison extends BinaryOp implements Optimizable, IndexUseReporter
 {
@@ -984,7 +985,7 @@ public class GeneralComparison extends BinaryOp implements Optimizable, IndexUse
         }
     }
 
-    private AtomicValue convertForCompareAtomic(final AtomicValue value, final int thisType, final int otherType)
+    private AtomicValue convertForGeneralComparison(final AtomicValue value, final int thisType, final int otherType)
             throws XPathException {
         if (thisType == Type.UNTYPED_ATOMIC) {
 
@@ -1040,6 +1041,48 @@ public class GeneralComparison extends BinaryOp implements Optimizable, IndexUse
         return value;
     }
 
+    private AtomicValue convertForValueComparison(final AtomicValue value, final int thisType, final int otherType)
+            throws XPathException {
+        /*
+            if the two operands are instances of different primitive types then:
+         */
+        if (Type.primitiveTypeOf(thisType) != Type.primitiveTypeOf(otherType)) {
+
+            /*
+                a. If each operand is an instance of one of the types xs:string or xs:anyURI,
+                   then both operands are cast to type xs:string.
+             */
+            if ((Type.subTypeOf(thisType, Type.STRING) || thisType == Type.ANY_URI)
+                    && (Type.subTypeOf(otherType, Type.STRING) || otherType ==Type.ANY_URI)) {
+                return value.convertTo(Type.STRING);
+            }
+
+            /*
+                b. If each operand is an instance of one of the types xs:decimal or xs:float,
+                then both operands are cast to type xs:float.
+             */
+            if ((Type.subTypeOf(thisType, Type.DECIMAL) || thisType == Type.FLOAT)
+                    && (Type.subTypeOf(otherType, Type.DECIMAL) || otherType == Type.FLOAT)) {
+                return value.convertTo(Type.FLOAT);
+            }
+
+            /*
+             *  c. If each operand is an instance of one of the types xs:decimal, xs:float, or xs:double,
+             *     then both operands are cast to type xs:double.
+             */
+            if ((Type.subTypeOf(thisType, Type.DECIMAL) || thisType == Type.FLOAT || thisType == Type.DOUBLE)
+                    && (Type.subTypeOf(otherType, Type.DECIMAL) || otherType == Type.FLOAT || otherType == Type.DOUBLE)) {
+                return value.convertTo(Type.DOUBLE);
+            }
+
+            /*
+             *  d. Otherwise, a type error is raised [err:XPTY0004].
+             */
+            throw new XPathException(ErrorCodes.XPTY0004, "Incompatible primitive types");
+        }
+
+        return value;
+    }
 
     /**
      * Cast the atomic operands into a comparable type and compare them.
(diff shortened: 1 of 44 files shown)


joewiz:
@adamretter Wow, thanks for diving into that challenge. I'm curious, does XQTS give you any indication about improvements (or even possibly regressions, I suppose) that result from this PR?

adamretter:
> does XQTS give you any indication about improvements (or even possibly regressions, I suppose) that result from this PR?

@joewiz actually we get a small regression in XQTS, but I think that is as a product of other bugs in eXist-db, whereby we used to pass tests that we shouldn't of.  Before this PR we had `78.92%`, after this PR we have `78.49%`. So we will have to chase those other bugs in future PRs ;-)

exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
```diff
@@ -1040,6 +1041,48 @@ private AtomicValue convertForCompareAtomic(final AtomicValue value, final int t
...
+            if ((Type.subTypeOf(thisType, Type.DECIMAL) || thisType == Type.FLOAT)
+                    && (Type.subTypeOf(otherType, Type.DECIMAL) || otherType == Type.FLOAT)) {
+                return value.convertTo(Type.FLOAT);
+            }
+
+            /*
+             *  c. If each operand is an instance of one of the types xs:decimal, xs:float, or xs:double,
+             *     then both operands are cast to type xs:double.
+             */
+            if ((Type.subTypeOf(thisType, Type.DECIMAL) || thisType == Type.FLOAT || thisType == Type.DOUBLE)
```
line-o, line 1073: Note to self: Test for behaviour with decimal subtypes and also subtypes of xs:integer

line-o:
retest #3330

wolfgangmm:
With this PR applied, I see the following NPE in the logs:

```
	at org.exist.xquery.DynamicTypeCheck.check(DynamicTypeCheck.java:84) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.DynamicTypeCheck.eval(DynamicTypeCheck.java:67) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.FunctionCall.eval(FunctionCall.java:205) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
```

adamretter:
@wolfgangmm that's not great... But you need to tell me how to reproduce it for me to help!

wolfgangmm:
@adamretter sure, this was just a notice. Debugging takes a while. The error does not pop up immediately, otherwise it would be simple.

dizzzz:
image

almost !!!!

dizzzz:
@adamretter the Java13 issue seems to be real

wolfgangmm:
The NPE I reported does not seem related to this PR. The issue also appears on older eXist versions, but only under very particular circumstances. I'll submit a separate fix.

joewiz:
@adamretter This PR doesn't address https://github.com/eXist-db/exist/issues/1715, does it?

Also, does anyone have insights into the failing Travis openjdk13 test?

adamretter:
@joewiz 

> This PR doesn't address #1715, does it?

No.

> Also, does anyone have insights into the failing Travis openjdk13 test?

Yes, I have a reproducible test case here that I am working with.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
```diff
@@ -39,7 +41,7 @@
     public static final DoubleValue NEGATIVE_INFINITY = new DoubleValue(Double.NEGATIVE_INFINITY);
     public static final DoubleValue NaN = new DoubleValue(Double.NaN);
 
-    private final double value;
+    final double value;
```
dizzzz, line 44: package protected on purpose?

exist-core/src/main/java/org/exist/xquery/value/Type.java
```diff
@@ -105,95 +105,136 @@
...
 
-        //THIS type system is broken - some of the below should be sub-types of ANY_SIMPLE_TYPE
-        //and some should not!
+        // ITEM sub-types
         defineSubType(ITEM, ATOMIC);
-        defineSubType(ATOMIC, STRING);
-        defineSubType(ATOMIC, BOOLEAN);
-        defineSubType(ATOMIC, QNAME);
+        defineSubType(ITEM, FUNCTION_REFERENCE);
+        //defineSubType(ITEM, NODE);                // TODO(AR) this appears in the XDM 3.1, but uncommenting this breaks a lot of stuff in eXist-db
```
line-o, line 135: Setting NODE to a different value (I tried 7, the first available number not yet taken) allows this definition.
https://github.com/eXist-db/exist/pull/3363/files#diff-14a14d3b3adcc1ebc87fe76d1e77c862R43
At least all Xquery3 Tests pass.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
adamretter: I think it's needed in `#createComparisonWith(NumericValue)`, did you see otherwise?

exist-core/src/main/java/org/exist/xquery/value/Type.java
adamretter: Hmmm... okay that's interesting. I would have thought that changing its value should have no effect! Thanks Juri, I will take a look at this...

line-o: It was the only negative value, therefore I gave it try.

line-o: If this is turns out to be feasible I would suggest to change NO_VALUE back to -1.

adamretter: Sure. It should not matter if it is positive or negative as long as it is a valid java int.

exist-core/src/main/java/org/exist/xquery/value/Type.java
```diff
@@ -105,95 +105,136 @@
     public final static int MAP = 102;
     public final static int ARRAY = 103;
     private final static Logger LOG = LogManager.getLogger(Type.class);
+
+    private static int NO_SUCH_VALUE = -99;
```
line-o, line 109: If Type.NODE  value is changed to a positive value this should be set to -1 again.

exist-core/src/main/java/org/exist/xquery/value/Type.java
line-o: Well with it being negative nothing works... the test suites cannot even be started. With it being zero an exception is thrown that zero is not allowed. When set to a positive value the DB runs and can be used normally and core tests pass.

line-o:
#3390 is not fixed by this PR

joewiz:
@line-o Not sure I follow. Are you’re saying someone claims this PR fixed that issue?

line-o:
@joewiz you were asked by @adamretter to test it with this PR applied. I did.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
dizzzz: just checking

exist-core/src/main/java/org/exist/xquery/value/Type.java
dizzzz: @adamretter ?

joewiz:
@line-o Ah, I see, yes, he asked if it helped. You must not've seen [my note in Slack](https://exist-db.slack.com/archives/CG2MRUZ35/p1587915341416200) that I'd tested it (we should make better use of threads so overlapping conversations aren't so hard to follow), or the Context > eXist-db Version section of my report for #3390 stating that I had tested my report against that branch.

line-o:
@joewiz yes, I noticed that after your question (re-reading if I missed something). But I think it is good info to have here, since it is mentioned.

dizzzz:
so...... what's up?

joewiz:
On today's community call @adamretter said he would like to try @line-o's suggestion.

exist-core/src/main/java/org/exist/xquery/value/Type.java
adamretter: @line-o Okay I checked out what you are mentioned.

I can indeed see the issue that you mention, it is because the Type is used in some places as an index into an array, and that can't be a negative value.

Unfortunately we cannot fix such a thing without changing the format of some `.dbx` files, as the value of the Type is encoded into key/values that are stored in the on-disk files. I haven't checked them all, but the first and most obvious is the `NativeValueIndex`. 

So if we wanted to fix that it would have to happen in a major release change and we would need to bump the file-version numbers of the relevant on-disk file formats.

I think if we get this one merged, we can open a second PR to do that for a 6.0.0 release or whatever

adamretter:
@joewiz @dizzzz I checked our the suggestion of @line-o and it works indeed, but we can't do it without changing the on-disk file format of some of the .dbx files - best that we leave such a further change to a different PR for a major release.

dizzzz:
does @line-o need to 'ok' his remarks, or I am I good to go?

adamretter:
@dizzzz IMHO you are good to go

PR #3364, opened 2020-04-13
[fix] java exception on higher order functions

### LINKED ISSUE #3358: [BUG] Java exception when passing a function with wrong arity to higher-order functions
**Describe the bug**

Passing a function with a wrong arity (not one) as the second parameter to a higher order function such as `fn:for-each` or `fn:filter` will throw a java exception.

```
ERROR (XQueryServlet.java [process]:559) - null
java.lang.ArrayIndexOutOfBoundsException: null
```

`fold-left` and `fold-right` not tested yet but are likely also affected, judging from the implementation.

**Expected behavior**

XQuery exception XPTY0004 - (argument) type mismatch

Saxon 10 HE returns `XPTY0004  The supplied function (function($m, $n){ $m }) has 2 arguments - expected 1`

**To Reproduce**

```Xquery
xquery version "3.1";

declare
    %test:assertError("XPTY0004")
function t:for-each-arity-0() {
    for-each(1, function () { 1 })
};

declare
    %test:assertError("XPTY0004")
function t:for-each-arity-2() {
    for-each(1, function ($a, $b) { 1 })
};

declare
    %test:assertError("XPTY0004")
function t:filter-arity-0() {
    filter(1, function () { 1 })
};

declare
    %test:assertError("XPTY0004")
function t:filter-arity-2() {
    filter(1, function ($a, $b) { 1 })
};

```

 - OS: 10.15.4
 - eXist-db version: 5.3.0-SNAPSHOT 22b3c2776b8ef30b55a192b7e33448a05565d4cf
 - Java Version 1.8.0_152 64-Bit

**Additional context**

line-o:
### Description:

Fixes #3358 ArrayOutOfBoundException when a function reference supplied to a higher order
function has the wrong arity. 

This will now throw error XPTY0004.

New method `funcRefHasDifferentArity` added to class `FunHigherOrderFun` for readability and code re-use.

### Reference:

https://www.w3.org/TR/xpath-functions/#func-filter

Error messages are in line with Saxon 10

### Type of tests:

A new XQSuite for higher order functions is added.
Old tests left in place for now.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
@@ -120,14 +120,19 @@ public class FunHigherOrderFun extends BasicFunction {
 		cachedContextInfo = new AnalyzeContextInfo(contextInfo);
         super.analyze(cachedContextInfo);
 	}
-	
+
 	@Override
-	public Sequence eval(Sequence[] args, Sequence contextSequence)
+	public Sequence eval(final Sequence[] args, final Sequence contextSequence)
 			throws XPathException {
         Sequence result = new ValueSequence();
+
         if (isCalledAs("for-each")) {
             try (final FunctionReference ref = (FunctionReference) args[1].itemAt(0)) {
 				ref.analyze(cachedContextInfo);
+				if (funcRefHasDifferentArity(ref, 1)) {
+					throw new XPathException(this, ErrorCodes.XPTY0004,
+							"The supplied function (" + ref.getStringValue() + ") has " + ref.getSignature().getArgumentCount() + " arguments - expected 1");
+				}
 				for (final SequenceIterator i = args[0].iterate(); i.hasNext(); ) {
 					final Item item = i.nextItem();
 					final Sequence r = ref.evalFunction(null, null, new Sequence[]{item.toSequence()});
@@ -148,9 +153,18 @@ public class FunHigherOrderFun extends BasicFunction {
 
             try (final FunctionReference ref = refParam) {
 				ref.analyze(cachedContextInfo);
+				if (funcRefHasDifferentArity(ref, 1)) {
+					throw new XPathException(this, ErrorCodes.XPTY0004,
+							"The supplied function (" + ref.getStringValue() + ") has " + ref.getSignature().getArgumentCount() + " arguments - expected 1");
+				}
+
 				for (final SequenceIterator i = seq.iterate(); i.hasNext(); ) {
 					final Item item = i.nextItem();
 					final Sequence r = ref.evalFunction(null, null, new Sequence[]{item.toSequence()});
+
+					// call to effectiveBooleanValue is not spec compliant
+					// spec: https://www.w3.org/TR/xpath-functions/#func-filter
+					// two pending tests in exist-core/src/test/xquery/xquery3/fnHigherOrderFunctions.xql
 					if (r.effectiveBooleanValue()) {
 						result.add(item);
 					}
@@ -159,6 +173,10 @@ public class FunHigherOrderFun extends BasicFunction {
         } else if (isCalledAs("fold-left")) {
             try (final FunctionReference ref = (FunctionReference) args[2].itemAt(0)) {
 				ref.analyze(cachedContextInfo);
+				if (funcRefHasDifferentArity(ref, 2)) {
+					throw new XPathException(this, ErrorCodes.XPTY0004,
+							"The supplied function (" + ref.getStringValue() + ") has " + ref.getSignature().getArgumentCount() + " arguments - expected 2");
+				}
 				final Sequence seq = args[0];
 				final Sequence zero = args[1];
 				result = foldLeft(ref, zero, seq.iterate());
@@ -166,6 +184,10 @@ public class FunHigherOrderFun extends BasicFunction {
         } else if (isCalledAs("fold-right")) {
             try (final FunctionReference ref = (FunctionReference) args[2].itemAt(0)) {
 				ref.analyze(cachedContextInfo);
+				if (funcRefHasDifferentArity(ref, 2)) {
+					throw new XPathException(this, ErrorCodes.XPTY0004,
+							"The supplied function (" + ref.getStringValue() + ") has " + ref.getSignature().getArgumentCount() + " arguments - expected 2");
+				}
 				final Sequence zero = args[1];
 				final Sequence seq = args[0];
 				if (seq instanceof ValueSequence) {
@@ -179,6 +201,10 @@ public class FunHigherOrderFun extends BasicFunction {
         } else if (isCalledAs("for-each-pair")) {
             try (final FunctionReference ref = (FunctionReference) args[2].itemAt(0)) {
 				ref.analyze(cachedContextInfo);
+				if (funcRefHasDifferentArity(ref, 2)) {
+					throw new XPathException(this, ErrorCodes.XPTY0004,
+							"The supplied function (" + ref.getStringValue() + ") has " + ref.getSignature().getArgumentCount() + " arguments - expected 2");
+				}
 				final SequenceIterator i1 = args[0].iterate();
 				final SequenceIterator i2 = args[1].iterate();
 				while (i1.hasNext() && i2.hasNext()) {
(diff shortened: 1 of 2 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
```diff
@@ -148,9 +148,18 @@ public Sequence eval(Sequence[] args, Sequence contextSequence)
 
             try (final FunctionReference ref = refParam) {
 				ref.analyze(cachedContextInfo);
+				if (!ref.getSignature().isOverloaded() && ref.getSignature().getArgumentCount() != 1) {
```
adamretter, line 151: If we want to bring it inline with the spec. Would it not be better just to revise our function signature for `fn:filter`?

line-o: Do you have a pointer how? This is the current signature for the second parameter 
```java
new FunctionParameterSequenceType("function", Type.FUNCTION_REFERENCE, Cardinality.EXACTLY_ONE, "the function to call")
```

line-o: How would one state that the provided function must have exactly one parameter of type item?

line-o: Please note: It is the arity of the user provided filter function we are discussing here.

line-o:
Why does this need to wait for 6.0.0?

line-o:
@dizzzz judge for yourself. All changes are in the PR description. This is just fixing xquery implementation bugs.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
adamretter: Ah right of course!

line-o:
A search for calls to fn:filter within the XQuery code of our org did not reveal one case where a user provided filter function returned something else than xs:boolean. Therefore it is fairly safe to state that no bundled application will break.

adamretter:
Okay, after giving this some more serious though, I think there is a better way to address the issue...

At the moment eXist-db does not do type checking on the HoF parameters which themselves are functions. Rather than just fixing fn:filter in a very ad-hoc way, we should fix this more generally. This should not be too hard to implement, we basically need to:

1. Create a sub-class of `FunctionParameterSequenceType` called something like `FunctionParameterFunctionType` which carries the extra details of the arguments accepted by the function parameter itself.

2. Override the methods `checkType` from `SequenceType` in `FunctionParameterFunctionType` to check the additional details.

line-o:
For version 6 that is definitely what we should aim for. For this fix it is not needed.

adamretter:
@line-o I think we should discuss on the community call. I am still of the opinion that this change is also a 6.0.0 change.

line-o:
If you would just say why @adamretter

adamretter:
@line-o It is as I already explained on the community call... You are changing the effective API!

At the moment the *effective* API for `fn:filter` in eXist-db is:

```
fn:filter($sequence as item()*, $f as function(*)) as item()*
```

With your change the *effective* API instead becomes:
```
fn:filter($sequence as item()*, $f as function(item()*) as xs:boolean) as item()*
```

So your change makes the current API more restrictive, this is definitely an API change. API changes are reserved in SemVer versioning for Major Releases; which would make this PR a change that needs to be scheduled for 6.0.0.

---
Also for reference the XQuery 3.1 spec for `fn:filter` API is:
```
fn:filter($seq as item()*, $f as function(item()) as xs:boolean) as item()*
```

This is different to what your code imposes, note the `item()*` (in your effective API) vs. the `item()` (in the XQ3.1 spec API). So you would need to tighten that up if the goal of the PR is to bring the code inline with the XQ3.1 spec.

line-o:
`function(item()*)` is only theoretical since `((1, 2), (3, 4))` will evaluate to `(1, 2, 3, 4)` _before_ it gets passed to `fn:filter`.

adamretter:
@line-o with the signature:

> since ((1, 2), (3, 4)) will evaluate to (1, 2, 3, 4) before it gets passed to fn:filter.

Erm... That isn't passed to `fn:filter`, it is passed to the inner function

line-o:
@adamretter `fn:filter(((1, 2), (3,4)), $fun)` is equivalent to `fn:filter((1, 2, 3, 4), $fun)` otherwise we have a much bigger issue in the implementation.

adamretter:
@line-o Okay we were talking about different `item()*` positions in the signature!

Consider this which works in eXist-db right now due to the current effective API:

```xquery 3.1
fn:filter((1 to 5), function($i) as xs:string {
	"hello"
})
```

With your change, because the effective API changes, that code will no longer work... or did I miss something?

line-o:
It should have never worked and is clearly against the spec. It is not used in any bundled application. It is fairly easy to fix by wrapping `xs:boolean()` or `exists()` around the return value.
Compare that to the dramatic changes introduced in #3363 which was added to milestone 5.2.1 by you @adamretter 
But if that is really an issue I will revert that change to keep in the fix for #3358 which is really bad.

adamretter:
> It should have never worked and is clearly against the spec.

I totally agree with you. However, we must also agree that that is the API that eXist-db presents today.

> It is not used in any bundled application. 

Indeed. However, we can't say that others haven't done something. Whilst my example is very simple, there are good use-cases where you could write more complex code in that way i.e. relying on XQuery 3.1's definition of effective boolean value.

> Compare that to the dramatic changes introduced in #3363 which was added to milestone 5.2.1 by you @adamretter

Unless I have missed something, there is no effective API change in #3363. That is why it was given the 5.2.1 milestone.

> But if that is really an issue I will revert that change to keep in the fix for #3358 which is really bad.

I have no issue with your change per-se. I think it is a good change and I welcome it.

However we have a semver versioning process, and we have been following that up to now. As such that means that an API change is a major release. It's not personal, its just the standards we follow ;-)

line-o:
breaking change reverted: moved to issue #3382

line-o:
Since all higher order functions are affected by the java exception this PR now attempts to fix all of them by adding arity checks before execution.

adamretter:
@line-o Have you given consideration to -https://github.com/eXist-db/exist/pull/3364#issuecomment-613973053 ? I didn't see any response to it.

line-o:
@adamretter I answered:
>  For version 6 that is definitely what we should aim for. For this fix it is not needed.

line-o:
Since array:for-each, array:filter, array:fold-left, array:fold-right and map:for-each are also affected, I would really like to go for your proposed general approach @adamretter 
I had a look around. So this sparked new questions:

- The name would be `FunctionParameterFunctionSequenceType` as a sequence of functions is one possible value. Do you agree?
- how to limit evaluation - or how to remove complexity - with obsessive uses 
    for example: `function(xs:string, function(function(xs:string) as node()*) as element(a)?) as document?`
    I do not know if something like this would be allowed by the spec or what actually is.
    This is obviously not for know since the function parameter types are not parsed in that way yet.
    But since we know this is where we will want to go in the future it might make sense to think about it right away.
- As these are function signatures there is some overlap with `FunctionSignature` class, that is already dealing with return values and parameter types. I wonder if this could be abstracted or interfaced for re-use

adamretter:
> The name would be FunctionParameterFunctionSequenceType

The name doesn't make much difference to me, and you are right it is probably best to stick with the existing convention even if it isn't great.

>  as a sequence of functions is one possible value. Do you agree?

No. Actually a sequence of functions is the **only** possible value for a `FunctionParameterFunctionSequenceType`


> how to limit evaluation - or how to remove complexity - with obsessive uses
for example: function(xs:string, function(function(xs:string) as node()*) as element(a)?) as document?
I do not know if something like this would be allowed by the spec 

It is absolutely allowed and not "obsessive" :-p

The definition should easily be coped with by the `FunctionParameterFunctionSequenceType`. That class allows you to define the parameters to the function and its return type; The parameters to the function are of type `FunctionParameterSequenceType`. As `FunctionParameterFunctionSequenceType` is a sub-class of `FunctionParameterSequenceType` then a parameter to a function can of course be a function itself. Nice and simple :-)

> As these are function signatures there is some overlap with FunctionSignature class, that is already dealing with return values and parameter types. I wonder if this could be abstracted or interfaced for re-use

I think they have quite different uses. I would not worry about that at least in the first iteration.

line-o:
This PR should be merged as it adds a band aid to HoFs before #3389 can be merged in to exist v6

joewiz:
@line-o At least 1 approving review is required. 

@adamretter @dizzzz @eXist-db/core Please review when possible.

adamretter:
@line-o I do not see any advantage of merging this when #3389 is available? Am I missing something?

line-o:
@adamretter This will be the non-breaking band-aid, fixing NPEs before exist 6 is released.

dizzzz:
I am confused.....

adamretter:
@dizzzz Me too!

joewiz:
As discussed in last week's call, this is ok to review/merge for 5.3.0.

PR #3365, opened 2020-04-13
[wip] rename weird test of arrowop

line-o:
~~Removes a test that is plain wrong.~~

The test is correct (see discussion). Since it is not obvious what is tested here, the test itself will be renamed to mitigate that.


adamretter:
Hmmm actually this test is definitely run... it is reported as passing which is actually more worrying than it not being run :-(

1. Output from Maven Surefire:
Screenshot 2020-04-14 at 11 23 38

2. Output from IntelliJ JUnit:
Screenshot 2020-04-14 at 11 26 12

line-o:
Yeah, that is bad news. Now we definitely have to look at all tests described in XML.

adamretter:
@line-o Yeah. Unfortunately we have a test suite that was developed and added without any tests for itself!

adamretter:
@line-o Okay we can eliminate the Java test driver around it, as the pure XQuery part still claims it passes:

```
import module namespace t = "http://exist-db.org/xquery/testing"
    at "resource:org/exist/xquery/lib/test.xq";


let $testSet := <TestSet>
    <testName>xquery3 arrow operator</testName>
    <description>
        <p>Arrow operator in XQuery 3</p>
        <author>Joe Wicentowski</author>
        <author>Dmitriy Shabanov</author>
    </description>
    <test output="text" id="arrow-op-from-31">
        <task></task>
        <code><![CDATA[
xquery version "3.0";

"hello" => string-length()

]]></code>
        <error>EXXQDY0003</error>
    </test>
</TestSet>


return
	t:run-testSet($testSet, ())
```

line-o:
@adamretter I suggest we open an issue describing this. Since the test is still bogus, this PR can still be merged independently of the underlying issue.

adamretter:
@line-o Okay a little debugging and I got to the bottom of it.

The test is correct and perfectly valid... see if you can guess why ;-)

line-o:
:( because of the xquery version perhaps?

adamretter:
@line-o Exactly. Can you close this PR now please.

PR #3368, opened 2020-04-14
Improve XQuery test naming

adamretter:
Improves the naming of tests expressed in XQuery and XML.

Also:
* XML test suites are now prefixed `xmts`
* XQuery test suites are now prefixed `xqts`


line-o:
The prefix is a neat feature. Have you thought about using `xmlts` as a prefix for the old-style testuites? Or is the equal length of both prefixes somewhat useful/ important?

line-o:
Then I would propose `xmlts`, because I will recognise it in two months :)

line-o:
Where are these prefixed names visible? I could not find them in IntelliJ nor maven test output.

Screenshot 2020-04-15 at 07 47 13

adamretter:
@line-o IntelliJ doesn't show the package names (of which the prefix is a part).

You can see the package names in the Surefire Report generated by Maven.

PR #3376, opened 2020-04-19
Switch to `empty least` as default ordering behaviour in XQuery

adamretter:
At the moment eXist-db implements `empty greatest` as the default ordering in XQuery.

The spec says it is `implementation defined`. However, Saxon and BaseX both implement `empty least` as the default. IMHO `empty least` is the most likely *expected* behaviour.

In the interest of improved interoperability for XQuery users. This PR switched eXist-db from `empty greatest` to `empty least` as the default behaviour.

This isn't technically an API change, rather it is an internal behaviour change. Therefore it is a Semver minor change. But... if anyone thinks it should be a major release feature (i.e. 6.0.0), then let's hear your arguments ;-)

### DIFF
--- a/exist-core/pom.xml
+++ b/exist-core/pom.xml
@@ -751,6 +751,7 @@
                                 <exclude>src/test/java/org/exist/xquery/value/SubSequenceRangeTest.java</exclude>
                                 <exclude>src/test/java/org/exist/xquery/value/SubSequenceTest.java</exclude>
                                 <exclude>src/test/xquery/binary-value.xqm</exclude>
+                                <exclude>src/test/xquery/order.xqm</exclude>
 
                                 <!--
                                     Derivative work licensed under dbXML 1.0 and LGPL 2.1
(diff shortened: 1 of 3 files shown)


joewiz:
Great idea. I can't see how this would *break* code (i.e., cause errors to be thrown where there weren't any before). However, the altered behavior could require action on the developer's part to keep an app performing the same as it was before this PR, e.g., to specify `empty greatest` in `order by` expressions. For this reason—the desire to avoid forcing developers to change their code when upgrading within a major version—I would suggest we conservatively shift this to 6.0.0.

PieterLamers:
Wow that's fast! Awesome, thanks Adam. Let's see if I can dig up some more strange behavior... I would not mind a swift rollout as I would expect no one to count on this specific behavior for sorting `false` and `()`. I would expect them either in the same basket or next in line, before `true`. But yes if you want to be prudent, then you need to give the developers a proper heads up. If that requires a primary digit change, then so it must be.

line-o:
I consider this a breaking change.
An app lists items in a table to its users, sorted by the current behaviour. This order will change after the PR is applied.

reinhapa (approved):

PR #3379, opened 2020-04-20
Update collection.xconf.xsd for eXist 5

joewiz:
### Description:

This PR registers elements added in the [Lucene facets & fields feature](https://exist-db.org/exist/apps/doc/lucene#facets-and-fields) released in eXist 5.  Before this PR, valid xconf files would raise schema validation errors in eXide.  The previous XSD itself raised errors in oXygen; this PR addresses those issues and makes other improvements that require XSD 1.1.

**Details:**

The xconf used in the facets & fields documentation in the previous link raised errors when validated against the previous schema, because of the new `<facet>` and `<field>` elements:

```xml
    <index xmlns:db="http://docbook.org/ns/docbook" xmlns:xs="http://www.w3.org/2001/XMLSchema">
        <lucene>
            <text qname="db:article">
                <facet dimension="keyword" expression="db:info/db:keywordset/db:keyword"/>
                <field name="title" expression="db:info/db:title"/>
                <field name="author" expression="for $au in db:info/db:author string-join(($au/db:personname/db:firstname, $au/db:personname/db:surname), ' ')"/>
            </text>
        </lucene>
    </index>
</collection>
```

The errors raised in eXide were like:

> `http://www.w3.org/TR/xml-schema-1#cvc-complex-type.2.4.a?{"http://exist-db.org/collection-config/1.0":facet}&{"http://exist-db.org/collection-config/1.0":inline, "http://exist-db.org/collection-config/1.0":ignore}`

The PR adds handling for the new `<facet>` element (and its required `@dimension` and `@expression` attributes and its optional `@expression` attribute); and the repurposed `<field>` element (and its required `@name` attribute and 5 optional attributes, `@if`, `@expression`, `@type`, `@analyzer`, and `@store`).

In the course of applying these changes, I also cleaned up some other aspects of the schema:

- Previously some attributes were defined inline, while others were defined in attribute groups and referenced by name. This difference made the schema hard to read and hard edit. I made minimal changes needed to handle attributes consistently, with named groups and `-Req` or `-Opt` suffixes to indicate whether the attribute is required or optional.
- I also removed extraneous min/maxOccurs when the values were the defaults.
- I added some assertions which required XSD 1.1.

oXygen raised errors in the previous XSD itself:

> `"http://exist-db.org/collection-config/1.0":lucene` and `"http://exist-db.org/collection-config/1.0":lucene` (or elements from their substitution group) violate "Unique Particle Attribution". During validation against this schema, ambiguity would be created for those two particles.

### Reference:

- Facets & fields docs: https://exist-db.org/exist/apps/doc/lucene#facets-and-fields

### Type of tests:

None included. I ran the xsd against 150 xconf files—all of my cloned eXist-db org repos, my projects' own xconf files, as well as these, which make full use of the facet & field features: 

- https://gitlab.existsolutions.com/tei-publisher/tei-publisher-app/blob/litelement/collection.xconf
- https://gitlab.existsolutions.com/tei-publisher/vangogh/blob/master/collection.xconf

The most common schema errors—none serious—raised when validating xconfs against the schema are:

- `<fulltext>` element should be removed. 
- `xmlns:xs="http://www.w3.org/2001/XMLSchema"` should be added to the `<index>` element when `xs:` is invoked in `@type` attributes used to indicate the datatype of an index.
- some "new" range indexes incorrectly use a `@match` attribute on `<create>`, whereas the only documented attribute for defining range index with the `<create>` element is with the `@qname` attribute (aside from child `<field>` elements).

The only side known effect is that eXide will need an update to switch from `validation:jing-report` to `validation:jaxp-report` due to the switch to XSD 1.1. It seems our version of jing can't handle XSD 1.1. That PR is in https://github.com/eXist-db/eXide/pull/252.

### DIFF
--- a/schema/collection.xconf.xsd
+++ b/schema/collection.xconf.xsd
@@ -2,7 +2,8 @@
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:dcterms="http://purl.org/dc/terms/"
     xmlns="http://exist-db.org/collection-config/1.0"
-    targetNamespace="http://exist-db.org/collection-config/1.0" elementFormDefault="qualified">
+    targetNamespace="http://exist-db.org/collection-config/1.0" elementFormDefault="qualified"
+    xmlns:vc="http://www.w3.org/2007/XMLSchema-versioning" vc:minVersion="1.1">
 
     <xs:annotation>
         <xs:documentation>Schema for eXist-db Collection Configuration files /db/system/config/db/**/collection.xconf</xs:documentation>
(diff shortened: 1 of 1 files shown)


joewiz:
@adamretter Thanks!

PR #3385, opened 2020-04-22
Fix Travis Deploy step

adamretter:
Deploy step is only run from JDK8 build env, so let's make sure that one passes!


dizzzz:
I do not really understand the change, but if t fixes an issue, I am happy :-)

line-o:
LGTM

PR #3386, opened 2020-04-22
Offer further build profiles

adamretter:
@line-o If you don't need the distribution archives, e.g. when testing/hacking locally. Then with this PR, setting `-P !build-dist-archives` can save ~40 seconds on each build (at least on my machine).

### DIFF
--- a/exist-distribution/pom.xml
+++ b/exist-distribution/pom.xml
@@ -960,16 +960,11 @@
                         <goals>
                             <goal>single</goal>
                         </goals>
-                        <configuration>
-                            <attach>false</attach>
-                            <tarLongFileMode>gnu</tarLongFileMode>
-                            <descriptors>
-                                <descriptor>src/assembly/dist-assembly-unix.xml</descriptor>
-                                <descriptor>src/assembly/dist-assembly-win.xml</descriptor>
-                            </descriptors>
-                        </configuration>
                     </execution>
                 </executions>
+                <configuration>
+                    <skipAssembly>true</skipAssembly>  <!-- Overridden by profile build-dist-archives -->
+                </configuration>
             </plugin>
 
             <plugin>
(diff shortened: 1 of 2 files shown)


line-o:
@adamretter You heard me :) Thanks!
Two additional profiles are added, which should be mentioned here for completeness:

- `concurrency-stress-tests` and 
- `micro-benchmarks`

adamretter:
> You heard me :) Thanks!
@line-o I always "hear you", we just have different priorities sometimes

line-o:
@adamretter the other two are not executed on `mvn package`, correct?
I understand I can the new flag as such: `mvn package -P !build-dist-archives`

line-o:
Could you add an example how to trigger the other two? @adamretter

adamretter:
> the other two are not executed on mvn package, correct?

That's incorrect.

line-o:
@adamretter so why were the other two moved from modules to profiles?

adamretter:
@line-o Maybe best if you have a read-up on Maven build configurations. Sonatype produce some PDFs which are not a bad starting point, you can find them via Google.

line-o:
I still see three things that can be shaved off of local builds:
```
[INFO] eXist-db IzPack Installer .......................... SUCCESS [02:00 min]
[INFO] eXist-db Core JCStress Tests ....................... SUCCESS [  9.930 s]
[INFO] eXist-db Core JMH benchmarks ....................... SUCCESS [ 10.049 s]
```

Why would the last two need to run for every build?

line-o:
for the record: `Total time:  05:36 min`

line-o:
maybe JCStress and JMH benchmarks can be skipped by `skipTests=true` or is there a skipBenchmarks?

line-o:
Those three items are unnecessary and make up almost half of the build time.

line-o:
I ended up testing a lot of different calls to `mvn package` and ended up with 
```bash
mvn -T2C clean package -DskipTests -Ddependency-check.skip=true -Ddocker=false -P \!build-dist-archives,\!mac-dmg-on-mac,\!codesign-mac-dmg,\!mac-dmg-on-unix,\!installer,\!concurrency-stress-tests,\!micro-benchmarks
```

I would suggest to create an alias for that as it is impossible to type and remember. But at least this saves a lot of precious time.

Issue #1917, opened 2018-06-07
Cardinality of return types not enforced


joewiz:
### What is the problem

eXist fails to enforce the cardinality of function return types, except in the case of the arrow operator.  For example, take the two following cases:

**Case 1:**
```xquery
index-of(("A", "B", "C", "B"), "B")
```

**Case 2:**
```xquery
("A", "B", "C", "B") => index-of("B")
```

The result of these cases should be identical, returning `(2, 4)` - a sequence of 2 integers. Before https://github.com/eXist-db/exist/pull/1916, eXist's return cardinality for `fn:index-of` was `ZERO_OR_ONE`, so both should have raised an error, but only Case B did. 

This observation, together with the very large number of cardinality mismatches documented at https://gist.github.com/joewiz/78b1827971a3c1c5dec8fc2d13dec8b2, suggests that eXist may be failing to enforce its own cardinality checks.

### What did you expect

I expect eXist to enforce its return type's cardinality regardless of how a function is called - either directly, via arrow operator, etc.

### Describe how to reproduce or add a test

I think that developing a test for this would require java skills that I unfortunately lack.  

(Return types are correctly enforced in XQuery:

```xquery
xquery version "3.1";

declare function local:zero-or-one-integer($x as item()*) as xs:integer? {
    $x
};

local:zero-or-one((1, 2))
```

This correctly raises the error.  So the issue reported here appears to only affect functions implemented in Java.)

### Context information

- eXist-db version + Git Revision hash: eXist 4.2.0
- Java version: 1.8.0_172-b11
- Operating system: macOS 10.13.5
- 32 or 64 bit: 64 bit
- How is eXist-db installed: clone from GitHub
- Any custom changes in e.g. conf.xml: none

joewiz:
In eXist 5.2.0, no error is raised in the `index-of` examples at the beginning of this issue. But I assume the general problem of "eXist not enforcing return types on XQuery functions implemented in Java" persists in eXist 5. I seem to recall seeing confirmation of that statement but can't find it at the moment. 

Would it be correct to advance this to the eXist 5 milestone then?

adamretter:
@joewiz eXist-db 5.2.0 is still very lax about enforcing return types.

It also doesn't cope well with compound types, e.g. `map(xs:string, map(xs:string, xs:int)`, I think I fixed the `document-node(element(x))` compound thing recently, but that might need further testing too.

ndw:
Adam suggested I append this here as a related error.

[...]

PR #3389, opened 2020-04-24
[Feat] function type checks

line-o:
### Description:

BREAKING CHANGE

Internal module functions can now enforce function parameters to adhere to a given signature.

This is added to all HoFs `fn:filter`, `fn:for-each`, `fn:fold-left`, `fn:fold-right` as well
as their Array and Map function variants.

It is written such that other function values can also be checked `array(xs:string)` or complex ones like `map(xs:integer, function(array(*)) as xs:boolean))`

Examples:

- `fn:filter` -> `function (item()) as xs:boolean`
- `array:for-each-pair` -> `function (item()*, item()*) as item()*`
- `map:for-each` -> `function (xs:anyAtomicValue, item()*) as item()*`

### Reference:

Result of discussion with @adamretter in #3364
Is related to #1917  
supersedes #3361 

### Type of tests:

All tests pass, two where adapted to match correct behaviour.


exist-core/src/main/java/org/exist/xquery/Function.java
```diff
@@ -30,10 +30,8 @@
 import org.exist.xquery.util.Error;
 import org.exist.xquery.util.ExpressionDumper;
 import org.exist.xquery.util.Messages;
-import org.exist.xquery.value.Item;
-import org.exist.xquery.value.Sequence;
-import org.exist.xquery.value.SequenceType;
-import org.exist.xquery.value.Type;
+import org.exist.xquery.value.*;
+import org.fife.ui.rsyntaxtextarea.modes.DartTokenMaker;
```
line-o, line 34: this import was added by IDEA... will remove it

exist-core/src/main/java/org/exist/xquery/Function.java
```diff
@@ -377,6 +382,8 @@ protected boolean checkArgumentTypeCardinality(final Expression argument, @Nulla
             argument = new UntypedValueCheck(context, argType.getPrimaryType(),
                     argument, new Error(Error.FUNC_PARAM_TYPE, String.valueOf(argPosition), mySignature));
             returnType = argument.returnsType();
+        } else if (Type.subTypeOf(argType.getPrimaryType(), Type.FUNCTION_REFERENCE)) {
```
line-o, line 385: this will need cleanup as well

exist-core/src/main/java/org/exist/xquery/functions/fn/FunHigherOrderFun.java
```diff
@@ -3,208 +3,200 @@
...
             new QName("for-each", Function.BUILTIN_FUNCTION_NS),
             "Applies the function item $function to every item from the sequence " +
                     "$sequence in turn, returning the concatenation of the resulting sequences in order.",
-            new SequenceType[] {
+            new SequenceType[]{
                     new FunctionParameterSequenceType("sequence", Type.ITEM, Cardinality.ZERO_OR_MORE, "the sequence on which to apply the function"),
-                    new FunctionParameterSequenceType("function", Type.FUNCTION_REFERENCE, Cardinality.EXACTLY_ONE, "the function to call")
+                    new FunctionParameterFunctionSequenceType("function", Type.FUNCTION_REFERENCE,
+                            new SequenceType[] {
+                                    new SequenceType(Type.ITEM, Cardinality.ZERO_OR_MORE)
```
line-o, line 20: Cardinality needs to change to EXACTLY_ONE

JoernT (approved):
look good to me - very clean

joewiz:
As discussed in the Community Call, this PR helps with XQuery conformance, achieves 9 more passing XQTS tests that previously failed, provides better static checking, and lays the groundwork for additional type checking of function parameters and return values.

reinhapa (approved):

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * @param attributeName     The name of the parameter in the <strong>FunctionSignature</strong>.
+     * @param primaryType       The <strong>Type</strong> of the parameter.
+     * @param parameterTypes    The <strong>parameters</strong> the function(s) must accept.
+     * @param returnType        The <strong>Type</strong> the function(s) needs to return.
+     * @param cardinality       The <strong>Cardinality</strong> of the parameter.
+     * @param description       A description of the parameter in the <strong>FunctionSignature</strong>.
+     * @see org.exist.xquery.FunctionSignature @see Type @see org.exist.xquery.Cardinality
+     */
+    public FunctionParameterFunctionSequenceType(final String attributeName, final int primaryType, final SequenceType[] parameterTypes, final SequenceType returnType, final Cardinality cardinality, final String description) {
```
adamretter, line 52: If I am not mistaken then the value supplied to `primaryType` appears to always be `Type.FUNCTION_REFERENCE`. If that is the case, then the parameter can be removed in favour of the constant.

In the same vein it might make sense to modify `FunctionParameterSequenceType` so that its public constructor cannot accept a `Type.FUNCTION_REFERENCE`; instead throwing an `IllegalArgumentException` indicating that `FunctionParameterFunctionSequenceType` should be constructed instead.

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+public class FunctionParameterFunctionSequenceType extends FunctionParameterSequenceType {
+
+    private SequenceType[] parameters = null;
+    private SequenceType returnType = new SequenceType(Type.ITEM, Cardinality.ZERO_OR_MORE);
+
+    /**
+     * mimics isOverloaded property of FunctionSignature
+     * @see org.exist.xquery.FunctionSignature
+     */
+    private int arity = -1;
```
adamretter, line 41: It is not clear to me how this relates to `isOverloaded` could you add some information? It is not possible to have `overloaded` function signatures in XQuery, this was previously just a non-compliant thing offered by eXist-db that was later removed.

Could you add some explanation of why this is needed? It looks to me that the arity of the function (as a a parameter) is simply the value of `parameterTypes.length`, and that we don't need an additional class member variable for that.

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+import org.exist.xquery.ErrorCodes;
+import org.exist.xquery.XPathException;
+import org.exist.xquery.FunctionSignature;
+
+/**
+ * This class is used to specify the name and description of an XQuery function parameter of type function.
+ */
+public class FunctionParameterFunctionSequenceType extends FunctionParameterSequenceType {
+
+    private SequenceType[] parameters = null;
```
adamretter, line 34: This is only set from the `constructor` and could instead be assigned there and declared `final` here

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+    private SequenceType returnType = new SequenceType(Type.ITEM, Cardinality.ZERO_OR_MORE);
```
adamretter, line 35: This is only set from the `constructor` and could instead be assigned there and declared `final` here

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+     * shorthand if return type is unspecified
+     * @param attributeName     The name of the parameter in the <strong>FunctionSignature</strong>.
+     * @param primaryType       The <strong>Type</strong> of the parameter.
+     * @param parameterTypes    The <strong>Types</strong> of parameters the function needs to accept.
+     * @param cardinality       The <strong>Cardinality</strong> of the parameter.
+     * @param description       A description of the parameter in the <strong>FunctionSignature</strong>.
+     * @see org.exist.xquery.FunctionSignature @see Type @see org.exist.xquery.Cardinality
+     */
+    public FunctionParameterFunctionSequenceType(final String attributeName, final int primaryType, final SequenceType[] parameterTypes, final Cardinality cardinality, final String description) {
+        super(attributeName, primaryType, cardinality, description);
```
adamretter, line 70: Perhaps consider calling `this(...)` instead.

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+        this.parameters = parameterTypes;
+        this.arity = parameterTypes.length;
+    }
+
+    /**
+     * FIXME: is this constructor needed?
+     * @param attributeName
+     */
+    public FunctionParameterFunctionSequenceType(final String attributeName) {
+        super(attributeName);
```
adamretter, line 80: Perhaps consider calling `this(...)` instead.

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+     * @see org.exist.xquery.FunctionSignature @see Type @see org.exist.xquery.Cardinality
+     */
+    public FunctionParameterFunctionSequenceType(final String attributeName, final int primaryType, final SequenceType[] parameterTypes, final Cardinality cardinality, final String description) {
+        super(attributeName, primaryType, cardinality, description);
+        this.parameters = parameterTypes;
+        this.arity = parameterTypes.length;
+    }
+
+    /**
+     * FIXME: is this constructor needed?
```
adamretter, line 76: Is it needed?

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+     * @param attributeName
+     */
+    public FunctionParameterFunctionSequenceType(final String attributeName) {
+        super(attributeName);
+    }
+
+    @Override
+    public boolean checkType(Sequence seq) throws XPathException {
```
adamretter, line 84: `seq` can be `final`

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+
+    public boolean checkType(final FunctionReference ref) throws XPathException {
+        final FunctionSignature sig = ref.getSignature();
+        // check arity of referenced function call
+        arityMatches(sig);
+
+        // check return type
+        returnTypeMatches(sig);
+
+        if (arityUnspecified()) { return true; }
```
adamretter, line 117: Formatting please.

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+
+        // check argumentTypes
+        final SequenceType[] arguments = sig.getArgumentTypes();
+        for (int i = 0; i < arguments.length; i++) {
+            final int argumentType = arguments[i].getPrimaryType();
+            final int parameterType = parameters[i].getPrimaryType();
+            // ITEM is likely unspecified return type - catch error later
```
adamretter, line 124: Can be improved by final assignment of return type in Constructor, and using either a tombstone value to indicate non-assignment, or a null Integer instead of an int.

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        return true;
+    }
+
+    private void arityMatches (final FunctionSignature sig) throws XPathException {
+        if (arityUnspecified()) { return; }
+        final int otherArity;
+
+        if (sig.isOverloaded()) {
+            // concat#3 will not return the correct argument count but the number of
```
adamretter, line 140: I am not sure why this would be dealing with `concat#3`?
`fn:concat` is a *var-args* function; the only on in XQuery actually.

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
```diff
@@ -0,0 +1,171 @@
...
+     * @see org.exist.xquery.FunctionSignature
+     */
+    private int arity = -1;
+
+    /**
+     * @param attributeName     The name of the parameter in the <strong>FunctionSignature</strong>.
+     * @param primaryType       The <strong>Type</strong> of the parameter.
+     * @param parameterTypes    The <strong>parameters</strong> the function(s) must accept.
+     * @param returnType        The <strong>Type</strong> the function(s) needs to return.
+     * @param cardinality       The <strong>Cardinality</strong> of the parameter.
```
adamretter, line 48: The `cardinality` parameter should follow the `primaryType` parameter, as they are part of the same sequence type. This is how it currently works in `FunctionParameterSequenceType` and that should be preserved here, as it makes the code for defining a `FunctionSignature` more logical to both write and read.

exist-core/src/main/java/org/exist/xquery/FunctionTypeCheck.java
```diff
@@ -0,0 +1,140 @@
...
+package org.exist.xquery;
+
+import org.exist.dom.persistent.DocumentSet;
+import org.exist.xquery.util.ExpressionDumper;
+import org.exist.xquery.value.*;
+
+/**
+ * Check a function parameter type at runtime.
+ *
+ * @author wolf
```
adamretter, line 31: Did you perhaps mean `@author line-o`?

exist-core/src/main/java/org/exist/xquery/FunctionTypeCheck.java
```diff
@@ -0,0 +1,140 @@
...
+ */
+public class FunctionTypeCheck extends AbstractExpression {
+
+    private final Expression expression;
+    private final FunctionParameterFunctionSequenceType requiredType;
+
+    public FunctionTypeCheck(XQueryContext context, final FunctionParameterFunctionSequenceType requiredType, Expression expr) {
```
adamretter, line 38: parameters can be final. This applies throughout this class...

exist-core/src/main/java/org/exist/xquery/FunctionTypeCheck.java
```diff
@@ -0,0 +1,140 @@
...
+        super(context);
+        this.requiredType = requiredType;
+        this.expression = expr;
+    }
+
+    /* (non-Javadoc)
+     * @see org.exist.xquery.Expression#analyze(org.exist.xquery.AnalyzeContextInfo)
+     */
+    public void analyze(AnalyzeContextInfo contextInfo) throws XPathException {
```
adamretter, line 47: Please remove the `/* (non-Javadoc)` (which is not used since very old versions of auto-code generation in Eclipse IDE), and replace with JavaDoc. This applies throughout this class...

exist-core/src/main/java/org/exist/xquery/Function.java
```diff
@@ -56,98 +53,67 @@
...
      */
     protected Module parentModule;
-
+    /**
+     * Flag to indicate if argument types are statically checked.
+     * This is set to true by default (meaning: no further checks needed).
+     * Method {@link #setArguments(java.util.List)} will set it to false
+     * (unless overwritten), thus enforcing a check.
+     */
+    protected boolean argumentsChecked = true;
```
adamretter, line 68: From reading the JavaDoc and the naming of the variable. It is not clear to me whether `argumentsChecked` means that: "The arguments have already been checked", or whether it means that "The arguments will need to be checked". Perhaps some rewording/renaming here would help?

adamretter: Also, is `protected` scope correct here?

adamretter (changes requested):
I think this is a great start, and I am excited to see it. I do think it requires some work yet to complete it and polish it.

Apart from my specific code comments I am a bit confused by this statement in the PR description:

> NOTE:
> An implementation decision was, to still allow developers to write
> ```
> filter(1, function ($a) { exists($a) })
> ```
> instead of having to write
> ```
> filter(1, function ($a) as xs:boolean { exists($a) })
> ```

Unless I am mistaken, that isn't a decision for an Implementation... It is actually required by the XQuery 3.1 specification, see: https://www.w3.org/TR/xquery-31/#prod-xquery31-InlineFunctionExpr

I am also not sure I understand this statement:

> This later check is currently done in the function implementation itself (see fn:filter).

I understand that every function that accepts a function (as a parameter) will need its function signature modified to declare the additional parameter/return types, but 
does that also mean that each of those also needs to be modified to perform the actual type checks of the function parameter's parameters? If so, it would seem that this checking should be moved out into something like the `DynamicCardinalityCheck` class.

duncdrum:
@line-o can you rebase this?

line-o:
I will. But there is more to do before this can go in. I need to pick it up again.

line-o:
Rebased onto current develop, all tests pass and an additional 19 XQTS tests pass.

```xml
            <testcase name="xqhof21"/>
            <testcase name="xqhof20"/>
            <testcase name="xqhof22"/>
            <testcase name="map-get-100"/>
            <testcase name="for-each-pair-903"/>
            <testcase name="for-each-pair-904"/>
            <testcase name="fn-for-each-pair-037"/>
            <testcase name="fn-random-number-generator-30"/>
            <testcase name="array-for-each-006"/>
            <testcase name="array-filter-008"/>
            <testcase name="array-filter-007"/>
            <testcase name="for-each-014"/>
            <testcase name="for-each-013"/>
            <testcase name="filter-902"/>
            <testcase name="filter-007"/>
            <testcase name="fn-filter-015"/>
            <testcase name="fn-filter-018"/>
            <testcase name="filter-901"/>
            <testcase name="filter-008"/>
            <testcase name="array-for-each-pair-308"/>
```

The one failing test case `map-contains-017` is a due to https://github.com/eXist-db/exist/issues/6122

exist-core/src/main/java/org/exist/xquery/FunctionTypeCheck.java
line-o: FunctionTypeCheck is an old class and out of scope for this PR. We can always improve on this later.

exist-core/src/main/java/org/exist/xquery/FunctionTypeCheck.java
line-o: FunctionTypeCheck is an old class and out of scope for this PR. We can always improve on this later.

exist-core/src/main/java/org/exist/xquery/Function.java
line-o: Function is an old class and out of scope for this PR. We can always improve on this later.

exist-core/src/main/java/org/exist/xquery/FunctionTypeCheck.java
line-o: This is existing code and out of scope for this PR

line-o:
once #6128 is merged I will have to rebase

exist-core/src/main/java/org/exist/xquery/value/FunctionParameterFunctionSequenceType.java
line-o: It took me quite a while to understand why that might be useful here. But better late than never.

Issue #3395, opened 2020-04-27
[BUG]: travis no longer executes deployment to dockerhub


duncdrum:
Commits to master and to develop should trigger pushes of new images to dockerhub, via the deployment step. Our last successful push is 16days ago. 

![Screenshot 2020-04-27 at 11 34 14](https://user-images.githubusercontent.com/6205362/80357292-0dacba00-887b-11ea-9e85-f09de36d4839.png)

taking the [log of the last run](https://travis-ci.com/github/eXist-db/exist/jobs/323525301#L3286), nothing seems to happen during the deployment step. 

![Screenshot 2020-04-27 at 11 43 29](https://user-images.githubusercontent.com/657222/80358297-79dbed80-887c-11ea-8a51-e6d85ca71a64.png)

**Expected behavior**
Tagged images should be pushed via travis automatically.

line-o:
@duncdrum I am correctly assuming that we would see an error, if Dockerhub credentials would not be working?

duncdrum:
correct, the credentials are working

PR #3396, opened 2020-04-27
fix deployment syntax on travis

duncdrum:
see #3395
 we ll only know if this fixes the problem, once it is merged to develop unfortunately.


dizzzz:
if it works, it works.

PR #3399, opened 2020-04-29
Add functioning invite link to eXist Cmty Slack

joewiz:
### Description:

The Slack link in our README wasn't an invite link. This should help people who haven't seen the link in exist-open join the Slack.

### Reference:

n/a

### Type of tests:

n/a


dizzzz:
if it works; what is the risk dat the magic token expires?

joewiz:
@dizzzz Luckily this link doesn't expire—as shown in the screenshot below. If we ever revoke/reissue this link, we'll just update it here. The original link, unfortunately, doesn't give people a way to request access. So the only way to effectively let people join is by sharing the full link.

> Screen Shot 2020-04-29 at 2 12 06 PM

PR #3409, opened 2020-05-12
Fix the Node/Item XDM Relationship

adamretter:
Small enhancement built on https://github.com/eXist-db/exist/pull/3363

But... it changes the on-disk format of the Native Value Index (`values.dbx`), and so requires a major release.


line-o (dismissed):
Please also revert the integer value of NO_SUCH_VALUE back to -1

adamretter:
@line-o I don't think that needs to be done.

line-o:
It makes much more sense to have it as -1, when it is the only negative value. Comparable to `['a' , 'b'].indexOf('c')`

adamretter:
@line-o That's one perspective... feel free to send a follow-up PR.

dizzzz:
What is the status of the PR?

joewiz:
@dizzzz When you have the chance could you restart the AppVeyor CI on this PR? Thanks!

dizzzz:
sorry... 
image

exist-core/src/main/java/org/exist/xquery/value/Type.java
```diff
@@ -41,7 +41,7 @@
  */
 public class Type {
 
-    public static final int NODE = -1;
+    public static final int NODE = 7;
```
reinhapa, line 44: @adamretter  What is the technical reason of changing this value?

adamretter: By keeping this value as a positive value, it means that all Type values can be stored into 16 bits, as they are then effectively uint16_t. It also means that in future for Union Types, we can do simple bit operations on them.

We could actually go one better here, and renumber some of the others so that we can fit all the types into 8 bits, i.e. values in the integer range 1 to 254

reinhapa: I see. I think it would make sense to make those changes now as we do a major release anyway. So there such a (I think) breaking change is ok.

So push that commit here too.

line-o:
NO_SUCH_VALUE was changed from -1 to -99 in 3102f46fd2f24282b7d27372a4a8c219eee7c01d
I would like the reordering of Value types to include this to be -1 again.

exist-core/src/main/java/org/exist/xquery/value/Type.java
```diff
@@ -41,83 +42,105 @@
...
-    public final static int ARRAY = 103;
     private final static Logger LOG = LogManager.getLogger(Type.class);
 
-    private static int NO_SUCH_VALUE = -99;
-
-    private final static int[] superTypes = new int[512];
-    private final static Int2ObjectOpenHashMap<String[]> typeNames = new Int2ObjectOpenHashMap<>(64, Hash.FAST_LOAD_FACTOR);
-    private final static Object2IntOpenHashMap<String> typeCodes = new Object2IntOpenHashMap<>(100, Hash.FAST_LOAD_FACTOR);
+    // Tombstone value used for FastUtil Int maps
+    private static final int NO_SUCH_VALUE = -99;
```
reinhapa, line 48: @adamretter why is this value `-99` in this case? I this a related to the FastUtil API?

reinhapa (approved):

line-o (commented):
The changes look good, from a first look. Ignore earlier comment, as it did not show the new commits at the time of writing.

exist-core/src/main/java/org/exist/xquery/value/Type.java
line-o: I just tested this. There is no technical reason to leave it at `-99` the tests all pass.

adamretter: Indeed, there is no technical reason why it is -1 or -99, technically it makes no difference. To be clear, I have never claimed otherwise. The reason it was changed was for human reasons:

1. When looking at it as an integer value it is much easier to spot.
2. When looking at it as an unsigned binary value the wrap causes it to look similar to other valid values, whereas -99 was an easier pattern to spot.

Both of these were factors as to why I changed it whilst developing and debugging this code. As it technically makes no difference, I suggest leaving it as -99 to assist the humans that will likely have to work on this code in future.

adamretter:
> I would propose to shift the current mapping of node types by one which would make the relationship quite clear as all
types that are subtypes of NODE would have a higher integer value;

I am not sure if I understand what you are suggesting? At the moment in this PR I have:

```java
    public final static int NODE = 56;
    public final static int ATTRIBUTE = 57;
    public final static int COMMENT = 58;
    public final static int DOCUMENT = 59;
    public final static int ELEMENT = 60;
    public final static int NAMESPACE = 61;
    public final static int PROCESSING_INSTRUCTION = 62;
    public final static int TEXT = 63;
```

It looks to me from the above that "all types that are subtypes of NODE" already "have a higher integer value" than NODE.

exist-core/src/main/java/org/exist/xquery/value/Type.java
line-o: Not convinced `-1` is very different from the other values as it is the only negative value.

adamretter: As the person who actually did the work on this and spent time debugging and fixing it when there were issues, I would hope that you would believe me when I am trying to tell you that I had problems spotting the `-1` amongst other values, and that switching it to `-99` made it easier to find the issues! As you did not do this work yourself, and did not have that experience personally, I find your comments perplexing!

As we agree that there is no technical advantage to changing the value, but that I have explained that there is a very real human advantage to leaving it as -99, please can you explain what your argument is for changing it to `-1`, i.e. What advantage do you see that causes you to be so insistent about this?

line-o: lets make it `Integer.MIN_VALUE` then

line-o (commented):
What do you think?

exist-core/src/main/java/org/exist/xquery/value/Type.java
adamretter: For what reason exactly?

dizzzz (approved):

reinhapa:
@adamretter seem to have some license header problem, that needs to be fixed first

line-o:
I checked the unsigned binary values of 
- -1 : 1111111111111111
- -99 : 1111111110011101
- -2147483648 : 10000000000000000000000000000000

dizzzz:
rebase/merge needed

dizzzz:
image

dizzzz:
The value is a non-discussion ; it could be -42 too :-/  
-1 might bee too obvious and could have a different meaning; we use symbols anyway.
Leave it as is? Need rebase though  / merge conflict needs to be resolved.

We'd really have this PR in

exist-core/src/main/java/org/exist/xquery/value/Type.java
```diff
@@ -683,4 +711,44 @@ public static int primitiveTypeOf(final int type) throws IllegalArgumentExceptio
...
+            case Node.DOCUMENT_NODE:
+                return Type.DOCUMENT;
+
+            // un-mappable Node types, so just return the XDM Node type
+            case Node.ENTITY_REFERENCE_NODE:
+            case Node.ENTITY_NODE:
+            case Node.DOCUMENT_TYPE_NODE:
+            case Node.DOCUMENT_FRAGMENT_NODE:
+            case Node.NOTATION_NODE:
+                return Type.NODE;
```
reinhapa, line 747: Could be simplified using the new switch as a function directly returning the value not using `return`...

reinhapa (changes requested):
Some test fail too..

adamretter:
> Some test fail too..

@reinhapa Yes, I added a whole load of new tests which show existing problems in eXist-db, I have just one case left to solve and then we should pass them all. When I do I will post updates XQTS numbers to. Stay tuned...

reinhapa:
> @reinhapa Yes, I added a whole load of new tests which show existing problems in eXist-db, I have just one case left to solve and then we should pass them all. When I do I will post updates XQTS numbers to. Stay tuned...

Looking forward to it

exist-core/src/main/java/org/exist/xquery/value/Type.java
dizzzz: java17 or java21?

dizzzz:
again, I do not really care; if for debugging 99 is easier to spot in  debugger, please leave it that way. I feel that it makes a good reason. I'd like to stop this discussion please.

dizzzz:
Needed
- rebase
- reformat
- check tests

dizzzz:
I'd like to see the PR pulled in as-is, the discussion is interesting (..) but not so important I'd say. Waste of time/energy....

adamretter:
@dizzzz I have rebased it as you requested. It think it is now ready for you to merge.

exist-core/src/main/java/org/exist/xquery/value/AtomicValue.java
```diff
@@ -398,16 +398,29 @@ public String getStringValue() {
...
+                case Type.NAME:
+                case Type.NCNAME:
+                case Type.ID:
+                case Type.IDREF:
+                case Type.ENTITY:
+                    return new StringValue(getExpression(), "", requiredType);
+                case Type.ANY_URI:
+                    return AnyURIValue.EMPTY_URI;
+                case Type.BOOLEAN:
+                    return BooleanValue.FALSE;
```
reinhapa, line 419: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/AtomicValue.java
```diff
@@ -453,8 +466,9 @@ public AtomicValue convertTo(final int requiredType) throws XPathException {
                 //return new GMonthDayValue(value);
                 //case Type.UNTYPED_ATOMIC :
                 //return new UntypedAtomicValue(getStringValue());
-                default -> throw new XPathException(getExpression(), "cannot convert empty value to " + requiredType);
-            };
+                default:
+                    throw new XPathException(getExpression(), "cannot convert empty value to " + requiredType);
+            }
```
reinhapa, line 471: Use shorter switch style if possible

exist-core/src/main/java/org/exist/xquery/value/BooleanValue.java
```diff
@@ -69,21 +69,29 @@ public String getStringValue() throws XPathException {
...
+            case Type.DOUBLE:
+                return new DoubleValue(getExpression(), value ? 1 : 0);
+            case Type.STRING:
+                return new StringValue(getExpression(), getStringValue());
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.XPTY0004,
+                        "cannot convert 'xs:boolean(" + value + ")' to " + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 94: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/DateTimeValue.java
```diff
@@ -145,23 +145,37 @@ public int getType() {
...
+                return new GMonthValue(getExpression(), calendar);
+            case Type.STRING:
+                return new StringValue(getExpression(), getStringValue());
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "Type error: cannot cast xs:dateTime to "
+                                + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 178: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/DateValue.java
```diff
@@ -95,27 +95,37 @@ public int getType() {
...
             }
-            default -> throw new XPathException(getExpression(), ErrorCodes.FORG0001, "can not convert " +
-                    Type.getTypeName(getType()) + "('" + getStringValue() + "') to " +
-                    Type.getTypeName(requiredType));
-        };
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001, "can not convert " +
+                        Type.getTypeName(getType()) + "('" + getStringValue() + "') to " +
+                        Type.getTypeName(requiredType));
+        }
```
reinhapa, line 128: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/AnyURIValue.java
```diff
@@ -271,18 +271,21 @@ public boolean effectiveBooleanValue() throws XPathException {
...
+                return this;
+            case Type.STRING:
+                return new StringValue(getExpression(), uri);
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "Type error: cannot cast xs:anyURI to "
+                                + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 288: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/DayTimeDurationValue.java
```diff
@@ -172,35 +172,39 @@ public String getStringValue() {
...
             }
-            default -> throw new XPathException(getExpression(), ErrorCodes.XPTY0004, "cannot cast '" +
-                    Type.getTypeName(this.getItemType()) + "(\"" + getStringValue() + "\")' to " +
-                    Type.getTypeName(requiredType));
-        };
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.XPTY0004, "cannot cast '" +
+                        Type.getTypeName(this.getItemType()) + "(\"" + getStringValue() + "\")' to " +
+                        Type.getTypeName(requiredType));
+        }
```
reinhapa, line 207: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/DecimalValue.java
```diff
@@ -227,28 +227,47 @@ public boolean hasFractionalPart() {
...
+                return value.signum() == 0 ? BooleanValue.FALSE : BooleanValue.TRUE;
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "cannot convert  '"
+                                + Type.getTypeName(this.getType())
+                                + " ("
+                                + value
+                                + ")' into "
+                                + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 270: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/GDayValue.java
```diff
@@ -76,15 +76,22 @@ private static XMLGregorianCalendar stripCalendar(XMLGregorianCalendar calendar)
...
+                return this;
+            case Type.STRING:
+                return new StringValue(getExpression(), getStringValue());
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "Type error: cannot cast xs:time to "
+                                + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 94: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/GMonthDayValue.java
```diff
@@ -73,15 +73,22 @@ private static XMLGregorianCalendar stripCalendar(XMLGregorianCalendar calendar)
...
+                return this;
+            case Type.STRING:
+                return new StringValue(getExpression(), getStringValue());
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "Type error: cannot cast xs:time to "
+                                + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 91: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/GMonthValue.java
```diff
@@ -110,15 +110,22 @@ private static String fixTimezone(String value) {
...
+                return this;
+            case Type.STRING:
+                return new StringValue(getExpression(), getStringValue());
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "Type error: cannot cast xs:gMonth to "
+                                + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 128: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/GYearMonthValue.java
```diff
@@ -75,15 +75,22 @@ private static XMLGregorianCalendar stripCalendar(XMLGregorianCalendar calendar)
...
+            case Type.ITEM:
+                return new GYearMonthValue(getExpression(), this.calendar);
+            case Type.STRING:
+                return new StringValue(getExpression(), getStringValue());
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "Type error: cannot cast xs:time to "
+                                + Type.getTypeName(requiredType));
```
reinhapa, line 92: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/GYearValue.java
```diff
@@ -74,15 +74,22 @@ private static XMLGregorianCalendar stripCalendar(XMLGregorianCalendar calendar)
...
+                return this;
+            case Type.STRING:
+                return new StringValue(getExpression(), getStringValue());
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "Type error: cannot cast xs:time to "
+                                + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 92: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/QNameValue.java
```diff
@@ -124,17 +124,21 @@ private String computeStringValue() {
...
+            case Type.QNAME:
+                return this;
+            case Type.STRING:
+                return new StringValue(getExpression(), getStringValue());
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "A QName cannot be converted to " + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 141: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/TimeValue.java
```diff
@@ -94,17 +94,24 @@ public int getType() {
...
-        };
+            case Type.STRING:
+                return new StringValue(getExpression(), getStringValue());
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.FORG0001,
+                        "Type error: cannot cast xs:time to "
+                                + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 114: Restore previous shorter switch version

exist-core/src/main/java/org/exist/xquery/value/YearMonthDurationValue.java
```diff
@@ -109,28 +109,36 @@ public String getStringValue() {
...
-                    "cannot cast 'xs:yearMonthDuration(\"" + getStringValue() +
-                            "\")' to " + Type.getTypeName(requiredType));
-        };
+            case Type.UNTYPED_ATOMIC:
+                return new UntypedAtomicValue(getExpression(), getStringValue());
+            default:
+                throw new XPathException(getExpression(), ErrorCodes.XPTY0004,
+                        "cannot cast 'xs:yearMonthDuration(\"" + getStringValue() +
+                                "\")' to " + Type.getTypeName(requiredType));
+        }
```
reinhapa, line 141: Restore previous shorter switch version

reinhapa (changes requested):
1. The tests do not pass at all
2. Why where the shorter switch expressions being replaced with the old style?

reinhapa (approved):

reinhapa:
@adamretter there still testfailures

adamretter:
@reinhapa to answer your questions:

> The tests do not pass at all

They should do now. I have split some things not related to Node/Item XDM Relationship out of this PR and will send them in a follow-up PR.

> Why where the shorter switch expressions being replaced with the old style?

This is intentional, let me try and explain...

The "shorter switch expressions" (as you described them) are not readable when there are multiple values for the same case. The reason for this is they extend horizontally across the screen and make it impossible to understand easily see the changes between two versions of the code. The "old style" (as you described them) instead is laid out vertically and is much easier to read when you are considering what has changed between two or more versions.

Also when considering changes, consider that both diff tools are line based, and so putting it all onto one line stops a diff tool from clearly showing and/or highlighting small changes. For example here is a screenshot from a 2 version diff:

Screenshot 2024-10-09 at 19 13 50

Now consider when you have a 3 version diff as is common when working with Git versions and tools like diffmerge. It becomes almost impossible to understand the changes when you have one long line in 3 different versions that are compared side by side.

The "shorter switch expressions" are not bad per-se, but they are much much worse (i.e. less readable and understandable) by tools (like git, diff, or diffmerge, or humans) which are line based when you are comparing multiple versions. The reason for that is that when there are multiple case values they are laid out horizontally all on one line and that line often extends far off the viewable area on the screen.

Does that make sense?

dizzzz (approved):
LGTM (though I had one or 2 questions)

exist-core/src/main/java/org/exist/xquery/value/Type.java
```diff
@@ -41,83 +42,105 @@
...
+    private static final int NO_SUCH_VALUE = -99;
+
+    public final static int ITEM = 1;
+
+    public final static int ANY_TYPE = 2;
+
+    public final static int ANY_SIMPLE_TYPE = 3;
+
+    /* xs:anyAtomicType and its subtypes */
+    public final static int ANY_ATOMIC_TYPE = 4;
```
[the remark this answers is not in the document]
adamretter: > Just a question: why this rename from ATOMIC ? Is this more in line with the (XDM?) spec?

@dizzzz Yes, exactly - https://www.w3.org/TR/xpath-datamodel-31/#anyAtomicType

> thinking a nit more (from the top of my head) that this might break a lot of existing (java) extension (if they are already not broken :-P )

@dizzzz I doubt many of them are using `ATOMIC` as it is a very generic type. Also this is targeting 7.x.x where we are allowed to improve things that might break things.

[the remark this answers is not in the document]
dizzzz: my mistake, sorry, I thought that more names were changed.

dizzzz:
thnx!

reinhapa:
> @reinhapa to answer your questions:
> This is intentional, let me try and explain...

...

> The "shorter switch expressions" are not bad per-se, but they are much much worse (i.e. less readable and understandable) by tools (like git, diff, or diffmerge, or humans) which are line based when you are comparing multiple versions. The reason for that is that when there are multiple case values they are laid out horizontally all on one line and that line often extends far off the viewable area on the screen.
> 
> Does that make sense?

I see your point of the problem when comparing. But this could be circumvented by still putting each value on on a seperate line but leave them on the same `case` label if realy needed but gaining the security of no fall-thru errors. Besides that it would be cool to have those types as Enum anyway in the future as a missing value would be detected already at compile time.

adamretter:
> would be cool to have those types as Enum anyway

Yes it would, I already did that in another branch on another project ;-)

PR #3423, opened 2020-05-21
[feature] support match highlighting on ft:field hits. Closes #3210

### LINKED ISSUE #3210: [feature] add match highlighting to ft:field hits

The field feature in eXist 5 allows full text search within the field, and when hits are found, the field can be returned. Yet there is no facility for the highlighting of matches in fields. 

This feature request flows from the motivation for the index defined in https://github.com/eXist-db/exist/blob/develop/extensions/indexes/lucene/src/test/xquery/lucene/facets.xql#L187-L192: some divs are English, some are German, and we can use fields to perform language-specific and language-sensitive searches. Naturally, a search interface would display a KWIC summary of the hits. However, we do not have any facility for applying match highlighting to the results of field searches.

I would expect a function like `util:expand` to highlight matches in full text hits in a field. However, `ft:field` returns a string, and `util:expand` operates on XML nodes stored in the database. Thus, `util:expand` may not be suitable. Instead, perhaps it'll be necessary to create a new `ft:highlight-field-matches` function? Since it operates on strings, perhaps it could return results in an `<exist:field>` element, with matches wrapped in an `<exist:match>` element, like:

```xquery
declare
    %test:args("english:bir*")
    %test:assertEquals("<exist:field>And the <exist:match>birds</exist:match> are singing</exist:field>")
function facet:highlight-field-matches($query as xs:string) {
    doc("/db/lucenetest/multi-lang.xml")//div[ft:query(., $query, map {"fields": "english"})]
    => ft:highlight-field-matches("english")
};
```

I tried `util:expand`, but since it takes a node and `ft:field` returns a string, it throws an error.

**Additional context**

I imagine that search applications with diverse data (i.e., different XML vocabularies or different structures) may make growing use of `ft:field` to allow corpus-wide searches across collections. For example, a corpus may contain both TEI and DocBook collections, each of which would require distinct indexes (one for `tei:TEI` and one for `dbk:article`), but which could feed into a common field containing the contents to be made full text searchable. This is already possible, but when displaying search results, it is common to display KWIC summaries.

wolfgangmm:
# Description:

Implements match highlighting in query results obtained from lucene fields as proposed by @joewiz in #3210. Because fields are constructed at indexing time, the usual approach for match highlighting with `util:expand` cannot be applied. We also need to make sure that the same analyzer is applied as when the field was constructed. The PR adds one dedicated function.

### Reference:

#3210

### Type of tests:

Test case using German language analyzer included.

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
@@ -565,6 +565,23 @@ public class LuceneIndexWorker implements OrderedValuesIndex, QNamedKeysIndex {
         }
     }
 
+    /**
+     * Calls {@link LuceneUtil#extractTerms(Query, Map, IndexReader, boolean)}  to extract
+     * the terms which would be matched by the given query.
+     *
+     * @param query to extract terms for
+     * @return the map returned by {@link LuceneUtil#extractTerms(Query, Map, IndexReader, boolean)}
+     * @throws IOException in case of Lucene IO error
+     */
+    public Map<Object, Query> getTerms(final Query query) throws IOException {
+        final Map<Object, Query> termMap = new TreeMap<>();
+        index.withReader(reader -> {
+            LuceneUtil.extractTerms(query, termMap, reader, false);
+            return termMap;
+        });
+        return termMap;
+    }
+
     public NodeSet queryField(XQueryContext context, int contextId, DocumentSet docs, NodeSet contextSet,
             String field, String queryString, int axis, QueryOptions options)
             throws IOException, XPathException {
@@ -1091,7 +1108,7 @@ public class LuceneIndexWorker implements OrderedValuesIndex, QNamedKeysIndex {
      *
      * @return the lucene config or null
      */
-    @Nullable protected LuceneConfig getLuceneConfig(DBBroker broker, DocumentSet docs) {
+    public LuceneConfig getLuceneConfig(DBBroker broker, DocumentSet docs) {
         for (Iterator<Collection> i = docs.getCollectionIterator(); i.hasNext(); ) {
             Collection collection = i.next();
             IndexSpec idxConf = collection.getIndexConfiguration(broker);
--- a/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Field.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Field.java
@@ -20,13 +20,29 @@
 
 package org.exist.xquery.modules.lucene;
 
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.analysis.TokenStream;
+import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
+import org.apache.lucene.analysis.tokenattributes.OffsetAttribute;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.search.PhraseQuery;
+import org.apache.lucene.search.Query;
+import org.exist.Namespaces;
 import org.exist.dom.QName;
+import org.exist.dom.memtree.InMemoryNodeSet;
+import org.exist.dom.memtree.MemTreeBuilder;
 import org.exist.dom.persistent.Match;
 import org.exist.dom.persistent.NodeProxy;
-import org.exist.indexing.lucene.LuceneIndex;
-import org.exist.indexing.lucene.LuceneMatch;
+import org.exist.indexing.lucene.*;
+import org.exist.storage.NodePath;
 import org.exist.xquery.*;
 import org.exist.xquery.value.*;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.io.Reader;
+import java.io.StringReader;
+import java.util.Map;
 
 public class Field extends BasicFunction {
 
@@ -61,6 +77,21 @@ public class Field extends BasicFunction {
                     },
                     new FunctionReturnSequenceType(Type.ITEM, Cardinality.ZERO_OR_MORE,
                             "Sequence corresponding to the values of the field attached, cast to the desired target type")
+            ),
+            new FunctionSignature(
+                    new QName("highlight-field-matches", LuceneModule.NAMESPACE_URI, LuceneModule.PREFIX),
+                    "Highlights matches for the last executed lucene query within the value of a field " +
+                    "attached to a particular node obtained via a full text search. Only fields listed in the 'fields' option of ft:query will be " +
+                    "available to highlighting.",
+                    new SequenceType[] {
+                            new FunctionParameterSequenceType("node", Type.NODE, Cardinality.EXACTLY_ONE,
+                                    "the context node to check for attached fields which should be highlighted"),
+                            new FunctionParameterSequenceType("field", Type.STRING, Cardinality.EXACTLY_ONE,
+                                    "name of the field to highlight")
+                    },
+                    new FunctionReturnSequenceType(Type.ELEMENT, Cardinality.ZERO_OR_ONE,
+                            "An exist:field containing the content of the requested field with all query " +
+                                    "matches enclosed in an exist:match")
             )
     };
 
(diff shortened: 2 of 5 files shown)


joewiz:
@wolfgangmm A huge thank you! In my testing this function not only performs the required task; it does so 2-300x faster than util:expand. 

In the following table and corresponding chart, I show the time that each function took when provided growing numbers of hits from a ft:query call. (Specifically, I performed a search of the term "soviet" on 5 volumes from the FRUS TEI corpus, a query which returned 126 TEI `<div>` elements. I measured the time it took the functions to process 10, 20, 50, and 100 of these hits.) Effectively, util:expand took .2-.3s per hit, whereas ft:highlight-field-matches took .001s per hit. Thus, the 2-300x figure mentioned above.

hits | util:expand | ft:highlight-field-matches
-- | -- | --
10 | 3.021s | 0.015s
20 | 5.632s | 0.039s
50 | 12.51s | 0.069s
100 | 22.961s | 0.144s

> Screen Shot 2020-05-21 at 5 54 39 PM

I observed some minor differences in the output of kwic:summarize on these hits, but I think this is a side-effect of the whitespace handling in the original lucene full text (which knows about nodes) vs. the lucene fields (which only knows about strings)—nothing that can't be overcome. This isn't a comment on this PR, but more of an issue that users might encounter if they migrate from the original lucene full text to the fields.

adamretter:
@wolfgangmm Looks good. Could you send a PR for the documentation too please and then link that here?

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
```diff
@@ -565,6 +565,23 @@ public void compute(DirectoryTaxonomyReader reader, FacetsConfig config, FacetsC
...
+     * @return the map returned by {@link LuceneUtil#extractTerms(Query, Map, IndexReader, boolean)}
+     * @throws IOException in case of Lucene IO error
+     */
+    public Map<Object, Query> getTerms(final Query query) throws IOException {
+        final Map<Object, Query> termMap = new TreeMap<>();
+        index.withReader(reader -> {
+            LuceneUtil.extractTerms(query, termMap, reader, false);
+            return termMap;
+        });
+        return termMap;
```
dizzzz, line 582: so this effectively only returns an empty TreeMap no?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Field.java
```diff
@@ -84,13 +115,134 @@ public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathExce
...
+                                    int t = 1;
+                                    OffsetAttribute offset = stream.getAttribute(OffsetAttribute.class);
+                                    final int startOffset = offset.startOffset();
+                                    int endOffset = offset.endOffset();
+                                    while (stream.incrementToken() && t < phraseTerms.length) {
+                                        token = stream.getAttribute(CharTermAttribute.class).toString();
+                                        if (token.equals(phraseTerms[t].text())) {
+                                            offset = stream.getAttribute(OffsetAttribute.class);
+                                            endOffset = offset.endOffset();
+                                            if (++t == phraseTerms.length) {
```
dizzzz, line 190: ++t on purpose? for readability it is better to have it as a separate expression no?

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
line-o: I am not that familiar with the `->` notation in Java. It is likely that termMap is mutated in `LuceneUtil.extractTerms` and that resulting map is then returned.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Field.java
line-o: I agree, it is definitely more readable to have it in a separate expression. The subtle difference between `++t` and `t++` is easy to miss.

line-o:
@joewiz could you give an example of the differences in `kwic:summarize` output? That should definitely be mentioned in the documentation.

joewiz:
@line-o Here's an example. Given a plain xconf like:

```xml
<text qname="tei:div" index="no">
    <field name="fulltext"/>
</text>
```

Sending the fulltext results to `kwic:summarize` returns the following results for each method:

```xml
<comparison>
    <hits method="util:expand">
        <p n="1">
            <span class="previous"/>
            <span class="hi">Soviet</span>
            <span class="following"> Union to the Department of StateSource: Department of State,
                Ce...</span>
        </p>
        <p n="2">
            <span class="previous">...mber 6, the Embassy reported on </span>
            <span class="hi">Soviet</span>
            <span class="following"> media reaction to the U.S. P...</span>
        </p>
        <p n="3">
            <span class="previous">...iya, Moskovskaya Pravda, and on </span>
            <span class="hi">Soviet</span>
            <span class="following"> radio and television. (Department of State, Central ...</span>
        </p>
        <p n="4">
            <span class="previous">...: After initial surprise over the Reagan landslide victory, </span>
            <span class="hi">Soviet</span>
            <span class="following"> officials are venting their frus...</span>
        </p>
    </hits>
    <hits method="ft:expand-field-matches">
        <p n="1">
            <span class="previous"/>
            <span class="hi">Soviet</span>
            <span class="following"> Union to the Department of State Source: Department of State,
                C...</span>
        </p>
        <p n="2">
            <span class="previous">...mber 6, the Embassy reported on </span>
            <span class="hi">Soviet</span>
            <span class="following"> media reaction to the U.S. P...</span>
        </p>
        <p n="3">
            <span class="previous">...a , Moskovskaya Pravda , and on </span>
            <span class="hi">Soviet</span>
            <span class="following"> radio and television. (Department of State, Central ...</span>
        </p>
        <p n="4">
            <span class="previous">... After initial surprise over the Reagan landslide victory, </span>
            <span class="hi">Soviet</span>
            <span class="following"> officials are venting their frus...</span>
        </p>
    </hits>
</comparison>
```

The differences are quite subtle and could probably be overcome by providing an `@expression` on the field definition to "normalize" whitespace on the source nodes before they're indexed. The results probably also depend upon conf.xml settings like `indexer/@suppress-whitespace` and `@preserve-whitespace-mixed-content` - but in my example I'm using the default settings, "none" and "no".

Here are the corresponding nodes in the source TEI (apologies for the indentation, normally I'd format & indent these before posting them, but I figured it would be best to show the original indentation):

```xml
<!-- n="1" -->
                    <head>1. Telegram From the Embassy in the Soviet Union to the Department of
                            State<note n="1" xml:id="d1fn1" type="source">Source: Department of
                            State, Central Foreign Policy File, D800532–1105. Confidential;
                            Immediate. Sent for information to Leningrad, Beijing, Belgrade, Berlin,
                            Bonn, Bucharest, Budapest, London, Munich, New Delhi, Paris, Prague,
                            Sofia, Tokyo, Warsaw, and the U.S. Mission to <gloss target="#t_NATO_1"
                                >NATO</gloss>.</note>
                    </head>

<!-- n="2", n="3" -->

                    <p>17743. Subject: (C) Initial Moscow Views on the U.S. Election. Ref: Moscow
                            17719<note n="2" xml:id="d1fn2">In telegram 17719 from Moscow, November
                            6, the Embassy reported on Soviet media reaction to the U.S.
                            Presidential election as reported by <gloss target="#t_TASS_1"
                                >TASS</gloss>, <hi rend="italic">Pravda</hi>, <hi rend="italic"
                                >Sovetskaya Rossiya</hi>, <hi rend="italic">Moskovskaya Pravda</hi>,
                            and on Soviet radio and television. (Department of State, Central
                            Foreign Policy File, D800531–0499)</note> (NOTAL).</p>

<!-- n="4" -->

                    <p>2. Summary: After initial surprise over the <persName corresp="#p_RRW_1"
                            >Reagan</persName> landslide victory, Soviet officials are venting their
                        frustrations over Soviet relations with the <persName corresp="#p_CJ_1"
                            >Carter</persName> administration and intensely watching personalities
                        in the new administration for clues to future behavior. They are saying that
                        a workable relationship can be developed but they emphasize the importance
                        of negotiating at least a modified version of <gloss target="#t_SALT_1"
                            >SALT</gloss> II. They hope for early contact with members of the
                            <persName corresp="#p_RRW_1">Reagan</persName> foreign policy team as
                        part of the process in deciding how to deal with it. Our refusenik and
                        dissident contacts are unhappy over the election’s outcome, fearing the new
                        administration will lessen the U.S. commitment to human rights. End
                        summary.</p>

```

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
adamretter: Indeed, termMap may be mutated, but the formulation is pretty confusing. Perhaps instead consider:

```java
        return index.withReader(reader -> {
            final Map<Object, Query> termMap = new TreeMap<>();
            LuceneUtil.extractTerms(query, termMap, reader, false);
            return termMap;
        });
```

line-o: much better @adamretter

line-o:
@joewiz Thank you much clearer now. So, the difference boils down to one or more characters being added or missing in the summaries before and after elements. That's worth a mention, I would say.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Field.java
dizzzz: for me it is.

joewiz:
@wolfgangmm I was just about to merge this, now that the accompanying documentation PR has been merged, but I see there are a couple of unresolved suggestions above. Please see:

- https://github.com/eXist-db/exist/pull/3423#pullrequestreview-416996229
- https://github.com/eXist-db/exist/pull/3423#discussion_r429384801

dizzzz:
@wolfgangmm  ?

wolfgangmm:
@dizzzz sorry, I have a huge workshop coming next Monday. Just finishing preparations and then I'll look into this PR again ;-)

PR #3428, opened 2020-05-22
Repair module prefixes. Fixes #3424

### LINKED ISSUE #3424: [BUG] util:registered-functions returns QNames with null namespace prefixes
**Describe the bug**

For a small number of eXist extension functions, `util:registered-functions` returns results with a null namespace prefix. 

This causes incorrect the namespace prefix for affected functions to be omitted in eXist's function documentation, and prevents IDEs such as oXygen from accessing the affected functions when users request function autocompletion/documentation based on a function's prefix. (See screenshots below.)

Affected functions:

- `http://exist-db.org/xquery/xmldb`
  - copy-collection
  - copy-resource
- `http://exist-db.org/xquery/inspection`
  - inspect-function
  - inspect-module
  - inspect-module-uri
- `http://exist-db.org/xquery/system`
  - restore
- `http://exist-db.org/xquery/sql`
  - execute
- `http://exist-db.org/xquery/util`
  - absolute-resource-id
  - binary-doc
  - binary-doc-available
  - binary-doc-content-digest
  - document-id
  - document-name
  - eval
  - eval-and-serialize
  - eval-inline
  - eval-with-context
  - get-resource-by-absolute-id
  - is-binary-doc
  - qname-index-lookup
- `http://exist-db.org/xquery/compression`
  - deflate
  - gzip
  - inflate
  - tar
  - ungzip
  - zip

**Expected behavior**

Every function returned by `util:registered-functions` should have the namespace prefix of its parent module—and should never return a null namespace prefix.

**To Reproduce**

The query below lists which functions are affected.

```xquery
xquery version "3.1";

for $namespace-uri in util:registered-modules()
let $functions := try { util:registered-functions($namespace-uri) } catch * { <error>{$err:code} raised when getting registered functions for module {$namespace-uri}</error> }
return
    if ($functions instance of element(error)) then
        ()
    else
        let $prefix-analysis := 
            for $function in $functions
            group by $prefix := fn:prefix-from-QName($function)
            return
                map { 
                    "prefix": $prefix,
                    "functions": 
                        array { $function ! fn:local-name-from-QName(.) }
                }
        where count($prefix-analysis) gt 1
        order by $namespace-uri
        return
            map { 
                "module": $namespace-uri,
                "prefix-analysis": array { $prefix-analysis }
            }
```

Note, too, that running `inspect:inspect-functions` on the same modules returns results with a similar discrepancy - `base64-encode` has a prefix, whereas `binary-doc` lacks one:

```xml
<function name="util:base64-encode" module="http://exist-db.org/xquery/util"/>
<function name="binary-doc" module="http://exist-db.org/xquery/util"/>
```

**Screenshots**

oXygen's function documentation omits util:binary-doc:

> Screenshot 2020-05-14 at 08 57 22

eXist's function documentation omits the prefix for util:binary-doc:

> Screen Shot 2020-05-14 at 11 30 30 AM

 - OS: macOS 10.13.6
 - eXist-db version: 5.2.0, 5.3.0-SNAPSHOT
 - Java Version: 1.8.0_242 (Zulu 8.44.0.11-CA-macosx) (build 1.8.0_242-b20)

**Additional context**
- How is eXist-db installed? DMG, source
- Any custom changes in e.g. `conf.xml`? no

dizzzz:
Fixes #3424

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/inspect/InspectionModule.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/inspect/InspectionModule.java
@@ -78,10 +78,10 @@ public class InspectionModule extends AbstractInternalModule {
     }
 
     static FunctionSignature functionSignature(final String name, final String description, final FunctionReturnSequenceType returnType, final FunctionParameterSequenceType... paramTypes) {
-        return FunctionDSL.functionSignature(new QName(name, NAMESPACE_URI), description, returnType, paramTypes);
+        return FunctionDSL.functionSignature(new QName(name, NAMESPACE_URI, PREFIX), description, returnType, paramTypes);
     }
 
     static FunctionSignature[] functionSignatures(final String name, final String description, final FunctionReturnSequenceType returnType, final FunctionParameterSequenceType[][] variableParamTypes) {
-        return FunctionDSL.functionSignatures(new QName(name, NAMESPACE_URI), description, returnType, variableParamTypes);
+        return FunctionDSL.functionSignatures(new QName(name, NAMESPACE_URI, PREFIX), description, returnType, variableParamTypes);
     }
 }
--- a/exist-core/src/main/java/org/exist/xquery/functions/system/SystemModule.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/system/SystemModule.java
@@ -106,10 +106,10 @@ public class SystemModule extends AbstractInternalModule {
     }
 
 	static FunctionSignature functionSignature(final String name, final String description, final FunctionReturnSequenceType returnType, final FunctionParameterSequenceType... paramTypes) {
-		return FunctionDSL.functionSignature(new QName(name, NAMESPACE_URI), description, returnType, paramTypes);
+		return FunctionDSL.functionSignature(new QName(name, NAMESPACE_URI, PREFIX), description, returnType, paramTypes);
 	}
 
 	static FunctionSignature[] functionSignatures(final String name, final String description, final FunctionReturnSequenceType returnType, final FunctionParameterSequenceType[][] variableParamTypes) {
-		return FunctionDSL.functionSignatures(new QName(name, NAMESPACE_URI), description, returnType, variableParamTypes);
+		return FunctionDSL.functionSignatures(new QName(name, NAMESPACE_URI, PREFIX), description, returnType, variableParamTypes);
 	}
 }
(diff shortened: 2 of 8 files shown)


joewiz:
@dizzzz Thank you for digging into this one!

dizzzz:
Added simple testcase ; also searched for other potential issues (not found)

joewiz:
@dizzzz Here's oXygen's formatted version of the test function:

```xquery
(: When there is an issue, the issue is printed as part of the error messages:)
declare
%test:assertEquals("END")
function insp:check_if_prefix_is_present() {
    (
    for $namespace-uri in util:registered-modules()
    let $functions := try {
        util:registered-functions($namespace-uri)
    } catch * {
        <error>{$err:code} raised when getting registered functions for module {$namespace-uri}</error>
    }
    for $fie in $functions
    return
        if (contains($fie, ":") or contains($namespace-uri, "w3.org"))
        then
            ()
        else
            $fie || " in " || $namespace-uri
    , "END")
};
```

line-o:
my personal preference would be
```xquery
(: When there is an issue, the issue is printed as part of the error messages:)
declare
    %test:assertEquals("END")
function insp:check_if_prefix_is_present() {
    for $namespace-uri in util:registered-modules()
    let $functions :=
        try {
            util:registered-functions($namespace-uri)
        }
        catch * {
            <error>{$err:code} raised when getting registered functions for module {$namespace-uri}</error>
        }
    for $fie in $functions
    return
        if (contains($fie,":") or contains($namespace-uri, "w3.org"))
        then ()
        else $fie || " in " || $namespace-uri
    ,
    "END"
};
```

adamretter:
@dizzzz So I understand the change, but I don't understand why it is necessary. AFAIK prefixes are completely ephemeral and don't make up part of a function name. For example, you can have the same function module bound to different namespace prefixes where needed.

joewiz:
@adamretter Can you envision a different solution to the problems I described in #3424?

joewiz:
It's good to see the Travis CI openjdk14 build passing!

dizzzz:
I have no idea, but at least all functions work the same right now.

dizzzz:
I had the same issues with eXide where functions could not be resolved by typing the prefix: ; For me, I think I fixed a regression.

adamretter:
> Can you envision a different solution to the problems I described in #3424?

I can think of a few, but all of them require a lot of work. I am just not sure that `util:registered-modules` is designed in the best way, or the module system itself. Consider that in the module's configuration in conf.xml you don't specify the prefix, rather it is hard-coded into the Java code of the module. Also that prefix is also only the default, you will still find lots of old eXist code that uses `xdb` and not `xmldb` as the prefix for the `http://exist-db.org/xquery/xmldb` module... So when a user searches xdb:func they probably won't find it.

However, all that being said this PR is a simple fix which reinstates the previous behaviour. Thanks @dizzzz :-).
I might not like the previous behaviour, but I can send a bigger PR later if I want to fix/improve that too ;-)

adamretter:
> It's good to see the Travis CI openjdk14 build passing!

It's kinda intermittent ;-)

dizzzz:
I can indeed imagine that the other way is a lot of work, potentially affecting applications 
.

Thnx!

Issue #3450, opened 2020-06-16
[BUG] missing location from error messages


joewiz:
**Describe the bug**

I have discovered several cases in which eXist fails to include the location of errors in queries.

**Expected behavior**

When reporting errors, eXist should always report the line number and position (or column and row) of errors.

For a clue about what might be needed to fix this, see several similar fixes: https://github.com/eXist-db/exist/search?q=error+location&unscoped_q=error+location&type=Commits.

**To Reproduce**

The error messages raised by the conditions in the conditional expressions below fail to include the location of the error.

```xquery
if (map { "foo": "bar" } eq 1) then "yay" else "boo"
```

The error raised:

> err:FOTY0013 A function item other than an array cannot be atomized

Similarly:

```xquery
if ("a" eq 1) then "yay" else "boo"
```

The error raised: 

> err:XPTY0004 can not compare xs:string('a') with xs:integer('1')

I would expect the error message to include the location of the error—the condition.

Saxon reports the error at line 3 column 6 (one character to the right of the open parens of the condition). BaseX reports the error at line 3 column 30 (or 13 in the 2nd query)—(the close parens of the condition). eXist should be able to report column and line numbers of errors like this and others that it raises.

**Context (please always complete the following information):**
 - OS: macOS 10.15.5
 - eXist-db version: 5.2.0, 5.3.0-SNAPSHOT
 - Java Version: JDK 1.8.0_252 (Zulu 8.46.0.19-CA-macosx) (build 1.8.0_252-b14)

**Additional context**
- How is eXist-db installed? 5.2.0 via DMG installer, 5.3.0-SNAPSHOT built from source 
- Any custom changes in e.g. `conf.xml`? None.

PR #3452, opened 2020-06-17
bugfix: add missing location from error messages

dizzzz:
Simpe fix for #3450

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ConditionalExpression.java
+++ b/exist-core/src/main/java/org/exist/xquery/ConditionalExpression.java
@@ -92,8 +92,9 @@ public class ConditionalExpression extends AbstractExpression implements Rewrita
      */
     public Sequence eval(Sequence contextSequence, Item contextItem) throws XPathException {
         context.expressionStart(this);
-        final Sequence testSeq = testExpr.eval(contextSequence, contextItem);
+
         try {
+            final Sequence testSeq = testExpr.eval(contextSequence, contextItem);
             if (testSeq.effectiveBooleanValue()) {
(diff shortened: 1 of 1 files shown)


exist-core/src/main/java/org/exist/xquery/ConditionalExpression.java
```diff
@@ -92,8 +92,9 @@ public void analyze(AnalyzeContextInfo contextInfo) throws XPathException {
      */
     public Sequence eval(Sequence contextSequence, Item contextItem) throws XPathException {
         context.expressionStart(this);
-        final Sequence testSeq = testExpr.eval(contextSequence, contextItem);
+
         try {
+            final Sequence testSeq = testExpr.eval(contextSequence, contextItem);
```
adamretter, line 97: This assumes that the `testExpr` is on the same line as the conditional expression. The line/column number will be close but not necessarily accurate. I wonder if it is better to have no line number, or to have a line number which is in the vicinity of the error?

Although... I note that this is already problematic below for the `thenExpr` and `elseExpr`.

joewiz:
@dizzzz Thank you! I can confirm that with your PR, the error raised now points to the line and column where the conditional expression—the `if` clause—begins:

> err:XPTY0004 can not compare xs:string('a') with xs:integer('1') [at line 3, column 1]

exist-core/src/main/java/org/exist/xquery/ConditionalExpression.java
dizzzz: yes i realised after closing the laptop yesterday that the solution is just too simple :-)

Issue #3434, opened 2020-05-27
[BUG] Command line scripts for backup/restore do not support SSL


zed-g:
**Describe the bug**
Using the shell scripts provided by eXist (`backup.sh`/`backup.bat`) it is not possible to backup or restore from/to an eXist server with SSL, as the flag `-ossl-enable` is ignored and an error (`org.xmldb.api.base.XMLDBException: HTTP server returned unexpected status: Bad Request`) is returned. @adamretter wrote on slack:

> I took a look at the code. It doesn't look like it will set the SSL property from the Backup/Restore entry/points whereas it does for the client

**Expected behavior**
I expected to use a script like this:

```shell
backup --restore local\path\__contents__.xml --user admin --password foobar -ouri=xmldb:exist://exist.host.de/exist/xmlrpc -ossl-enable=true
```
**To Reproduce**

Try to connect to a remote eXist server with SSL using the provided backup/restore scripts (see shell example above).

**Context (please always complete the following information):**
 - OS: Windows 7
 - eXist-db version: 4.7.1
 - Java Version: JRE build 1.8.0_241-b07

zed-g:
If you give us some pointers on what to do, we could _maybe_ provide a fix. I'm interested in seeing this fixed, since this bug breaks part of our workflow.

adamretter:
@zed-g if you look in the `exist-core` module at how SSL is enabled in `org.exist.client.InteractiveClient` and then apply the same kind of idea to `org.exist.backup.Main`

MisterZurk0n:
I'm going to work on this issue and will try to fix it

MisterZurk0n:
So I think I've found a solution to the problem. As you said, I have oriented myself to the client class and set the SSL Property in the Main.class explicitly in the `process()` function:

     Optional.ofNullable(arguments.get(optionArg)).ifPresent(options -> {
            options.forEach(properties::setProperty);

            if(options.containsKey(SSL_ENABLE)){
                properties.setProperty(SSL_ENABLE, options.get(SSL_ENABLE));
            }
        });

And then I set the Property for the `database`:
`database.setProperty(SSL_ENABLE, properties.getProperty(SSL_ENABLE));`

This worked for me and the command mentioned in the description threw no Errors.

adamretter:
@MisterZurk0n Can you send a PR so we can take a look?

MisterZurk0n:
[...]

PR #3453, opened 2020-06-17
Fix: enabling SSL via commandline arguments, when doing a backup/restore

MisterZurk0n:
A small fix for the Backup.Main, which enables the user to connect to a remote eXist server via SSL

Closes https://github.com/eXist-db/exist/issues/3434

### DIFF
--- a/exist-core/src/main/java/org/exist/backup/Main.java
+++ b/exist-core/src/main/java/org/exist/backup/Main.java
@@ -63,6 +63,8 @@ public class Main {
     private static final String CREATE_DATABASE_PROP = "create-database";
     private static final String BACKUP_DIR_PROP = "backup-dir";
 
+    public static final String SSL_ENABLE = "ssl-enable";
+
     private static final String DEFAULT_USER = "admin";
     private static final String DEFAULT_PASSWORD = "";
     private static final String DEFAULT_URI = "xmldb:exist://";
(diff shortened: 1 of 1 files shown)


adamretter:
Hi @MisterZurk0n thanks for your PR :-)

I made a small additional commit...

Previously, you had it set to take the SSL setting as part of the *options* argument i.e. `-ossl-enable=true`.

Instead, I modified it to have its own argument in the same was as the Java Admin Client. You can now use either `--use-ssl` or the shorter `-S` to enable SSL.

I tested this by running the backup over SSL using the following command:

```
$ bin/backup.sh --use-ssl -u admin -ouri=xmldb:exist://localhost:8443/exist/xmlrpc -b /db
```

NOTE: You still have to set the correct port, otherwise it will attempt SSL to port 8080.

If you are happy with this, let me know and I will get it merged.

MisterZurk0n:
That is way better than my approach, you can merge this. I was too focused on the commandline example which was given in the issue description. Thank you for your help!

adamretter:
@dizzzz I just rebased. Assuming that CI passes (excl. Java 14), then we should be good to go :-)

joewiz:
Oops, unless this flag was already in the documentation, the PR should probably have been labelled as "needs documentation" and held until documentation was submitted - but luckily a user spotted this and volunteered to submit it! https://github.com/eXist-db/exist/issues/3434#issuecomment-650168949

PR #3457, opened 2020-06-21
Revert PR #3455 maven-surefire-plugin-3.0.0-M5

adamretter:
There seems to be an issue with Maven Surefire Plugin 3.0.0-M5 when running parallel builds.


joewiz:
@adamretter Thank you! With this PR, we're back to just JDK14 failing on Travis.

PR #3467, opened 2020-06-26
add tests for util:expand on queries using Lucene fields

tuurma:
currently `util:expand` when used for queries which include criteria for Lucene fields produce superfluous matches (for full text fragments matching criteria specified for fields)

This PR extends the `ft-match` test set for Lucene for cases including fields.

Note: should not be merged until current util:expand behaviour is fixed.


dizzzz:
image

tuurma:
Hi @dizzzz, yes, failure is an expected result at this stage, because current behaviour of util:expand is, in fact, incorrect. Should I mark up the test or the PR in some way to make it clearer?

dizzzz:
Please convert the PR to a draft PR... then it is clear that it is not yet finished and ready to go.

line-o:
@tuurma Is this PR now obsolete since field matches are handled by a different function?

duncdrum:
@tuurma can you confirm what should happen with this PR?

PR #3469, opened 2020-06-28
Get location from predicate evaluation and add to exception

### LINKED ISSUE #3462: [BUG] missing location from error messages
**Describe the bug**

I have discovered two more cases in which eXist fails to include the location of errors in queries.

**Expected behavior**

When reporting errors, eXist should always report the line number and position (or column and row) of errors.

See also https://github.com/eXist-db/exist/issues/3450.

**To Reproduce**

The error messages raised by the conditions in the conditional expressions below fail to include the location of the error.

```xquery
map {
    "foo": ["bar", "baz"]
}[?foo eq "bar"]
```

eXist's error lacks location info:

> err:XPTY0004 Expected single atomic value but found array with length 2

BaseX returns the location:

> Stopped at /Users/joe/workspace/XQuery-for-Humanists/Test/file, 3/16:
> [XPTY0004] Item expected, sequence found: ("bar", "baz").

Saxon returns the location at line 3 column 4:

> A sequence of more than one item is not allowed as the first operand of 'eq' (["bar", "baz", ])

Similarly:

```xquery
map {
    "foo": 
        map { 
            "bar": "baz"
        }
}[?foo eq "bar"]
```

eXist's error lacks location info:

> err:FOTY0013 A function item other than an array cannot be atomized

BaseX returns:

> Stopped at /Users/joe/workspace/XQuery-for-Humanists/Test/file, 6/16:
> [FOTY0013] Items of type map(xs:string, xs:string?) cannot be atomized.

Saxon returns an error at line 6 column 4:

> An atomic value is required for the first operand of 'eq', but the supplied type is a map type, which cannot be atomized

eXist should be able to report column and line numbers of errors.

 - OS: macOS 10.13.6
 - Java Version: JDK 11.0.7+10 (Liberica 2020-04-14 LTS)

**Additional context**
- How is eXist-db installed? 5.3.0-SNAPSHOT built from source 
- Any custom changes in e.g. `conf.xml`? None.

dizzzz:
Isolate location from predicate and add to exception
Closes #3462

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/FilteredExpression.java
+++ b/exist-core/src/main/java/org/exist/xquery/FilteredExpression.java
@@ -144,10 +144,21 @@ public class FilteredExpression extends AbstractExpression {
     }
 
     private Sequence processPredicate(Sequence contextSequence, Sequence seq) throws XPathException {
-        for (final Predicate pred : predicates) {
-            seq = pred.evalPredicate(contextSequence, seq, Constants.DESCENDANT_SELF_AXIS);
-            //subsequent predicates operate on the result of the previous one
-            contextSequence = null;
+
+        int line=-1;
+        int column=-1;
+        try { // Keep try-catch out of loop
+            for (final Predicate pred : predicates) {
+                line = pred.getLine();
+                column = pred.getColumn();
+                seq = pred.evalPredicate(contextSequence, seq, Constants.DESCENDANT_SELF_AXIS);
+                //subsequent predicates operate on the result of the previous one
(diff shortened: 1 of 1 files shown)


dizzzz:
Hmmm yes I have been trying to dive into that, but deeper in the code there is no try-catch used at all; and on crucial locations there is no location info :-/

joewiz:
Travis failed not only on the expected OpenJDK 14 run, but also on the macOS run. Just restarted that one.

dizzzz:
image
 all green

PR #3492, opened 2020-07-28
Small cleanup to fn:matches code

adamretter:
No functional changes.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunMatches.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunMatches.java
@@ -50,70 +50,70 @@ import static org.exist.xquery.regex.RegexUtil.*;
 
 /**
  * Implements the fn:matches() function.
- * 
+ * <p>
  * Based on the java.util.regex package for regular expression support.
- * 
+ *
  * @author <a href="mailto:wolfgang@exist-db.org">Wolfgang Meier</a>
  */
 public class FunMatches extends Function implements Optimizable, IndexUseReporter {
 
-	protected static final String FUNCTION_DESCRIPTION_1_PARAM =
-		"The function returns true if $input matches the regular expression " +
-		"supplied as $pattern, if present; otherwise, it returns false.\n\n";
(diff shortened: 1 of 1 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/FunMatches.java
```diff
@@ -46,92 +44,85 @@
...
 import java.util.regex.PatternSyntaxException;
 
+import static org.exist.xquery.FunctionDSL.*;
+import static org.exist.xquery.functions.fn.FnModule.functionSignatures;
 import static org.exist.xquery.regex.RegexUtil.*;
 
 /**
  * Implements the fn:matches() function.
- * 
+ * <p>
```
line-o, line 53: This new paragraph is never closed. Is that a problem?

adamretter: @line-o I think in Javadoc that is considered ok (nope... I don't know why either?!?)

line-o: Well it is not necessary in HTML to close some elements (p, li, th, td, ...). That is why I was asking, as it could be OK.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunMatches.java
```diff
@@ -219,19 +216,15 @@ public NodeSet preSelect(Sequence contextSequence, boolean useContext) throws XP
         final boolean caseSensitive = !hasCaseInsensitive(flags);
 
         final String pattern;
-		if(isCalledAs("matches-regex")) {
```
line-o, line 222: Why is `isCalledAs("matches-regex")`  removed from the code?
Is it no longer necessary, was it previously deprecated?

adamretter: The function definition was removed sometime ago, but it seems that we forgot to remove some of the implementation.

Issue #1655, opened 2017-12-10
AST "unexpected end of subtree" error with map postfix lookup


joewiz:
### What is the problem

When filtering a map entry with a postfix lookup, an error is thrown if the postfix lookup expression does not come first in the predicate.

### What did you expect

I expected a postfix expression to be valid regardless of its position in a predicate.

### Describe how to reproduce or add a test

The following query triggers the error:

```xquery
xquery version "3.1";

let $map := map { "name": "mike" }
return
    $map["mike" eq ?name]
```

BaseX and Saxon both return the expected result without error:

```xquery
map { "name": "mike" }
```

eXist returns the following error:

`Cannot compile xquery: exerr:ERROR <AST>:0:0: unexpected end of subtree`

eXist is able to complete the query if the postfix lookup is performed first in the predicate:

```xquery
xquery version "3.1";

let $map := map { "name": "mike" }
return
    $map[?name eq "mike"]
```

### Context information

- eXist-db version + Git Revision hash: eXist 3.7.0-SNAPSHOT+9091c3032
- Java version: Oracle JDK 1.8.0_152
- Operating system: macOS 10.13.1
- 32 or 64 bit: 64 bit
- Any custom changes in e.g. conf.xml: none

joewiz:
This bug is still present in eXist 5.0.0.

line-o:
The same error is thrown with this `(map {"f": 1}, map {"f": 2}) ! ?f`

PR #3496, opened 2020-08-04
allow function types in context item

### LINKED ISSUE #3240: [BUG] eXist throws errors if context item (.) is a map, array, or function
**Describe the bug**

According to [the spec](https://www.w3.org/TR/xquery-31/#dt-context-item):

> The **context item** is the item currently being processed. [Definition: When the context item is a node, it can also be referred to as the **context node**.] The context item is returned by an expression consisting of a single dot (`.`). 

However, eXist throws an error when `.` is a map, array, or function—valid types of items. 

**Expected behavior**

eXist should process maps, arrays, and function items when using the context item (`.`) and should not throw an error. 

For example, the following code should return the 1st map and does so in BaseX and Saxon:

```xquery
(map { "id": 1 }, map { "id": 2, "foo": 3 })[map:size(.) eq 1]
```

Instead, eXist throws the following error:

> err:XPTY0004 checking function parameter 1 in call map:size(self::node()): cannot convert map(*)('') to a node set

Similarly, this query should return the 1st array and does so in BaseX and Saxon:

```xquery
([1, 2], ["a", "b", "c"])[array:size(.) eq 2]
```

Instead, eXist throws the following error:

> err:XPTY0004 checking function parameter 1 in call array:size(self::node()): cannot convert array(*)('') to a node set

Similarly, the following query should return `string-join#1` and does so in BaseX and Saxon:

```xquery
(current-date#0, string-join#1)[function-arity(.) eq 1]
```

Instead, eXist throws the following error:

> err:XPTY0004 checking function parameter 1 in call function-arity(self::node()): cannot convert function(*)('') to a node set

Other issues have touched on this, but not as directly as reported here:

- https://github.com/eXist-db/exist/issues/2205
- https://github.com/eXist-db/exist/issues/2364
- https://github.com/eXist-db/exist/issues/2308

@adamretter investigated how eXist treats `.` here: https://github.com/eXist-db/exist/issues/2308#issuecomment-443990266.

**To Reproduce**

The queries above, as well as 2 other variants, are provided here as a self-contained XQSuite test:

```xquery
xquery version "3.1";

module namespace t="http://exist-db.org/xquery/test/context-item";

declare
    %test:assertTrue
function t:map-function() {
    (map { "id": 1 }, map { "id": 2, "foo": 3 })[map:size(.) eq 1]?id eq 1
};

declare
    %test:assertTrue
function t:map-lookup() {
    (map { "id": 1 }, map { "id": 2, "foo": 3 })[.?id eq 2]?id eq 2
};

declare
    %test:assertTrue
function t:array-function() {
    ([1, 2], ["a", "b", "c"])[array:size(.) eq 2]?1 eq 1
};

declare
    %test:assertTrue
function t:named-function-function() {
    (current-date#0, string-join#1)[function-arity(.) eq 1] => exists()
};

declare
    %test:assertTrue
function t:anonymous-function-function() {
    (function() { "foo" }, function($x) { "bar" })[function-arity(.) eq 1] => exists()
};
```

The test returns the following results:

```xml
<testsuites>
    <testsuite package="http://exist-db.org/xquery/test/context-item"
        timestamp="2020-02-04T11:08:49.657-05:00" tests="5" failures="0" errors="5" pending="0"
        time="PT0.05S">
        <testcase name="anonymous-function-function" class="t:anonymous-function-function">
            <error type="err:XPTY0004"
                message="It is a type error if, during the static analysis phase, an expression is found to have a static type that is not appropriate for the context in which the expression occurs, or during the dynamic evaluation phase, the dynamic type of a value does not match a required type as specified by the matching rules in 2.5.4 SequenceType Matching. checking function parameter 1 in call function-arity(self::node()): cannot convert function(*)('') to a node set"
            />
        </testcase>
        <testcase name="array-function" class="t:array-function">
            <error type="err:XPTY0004"
                message="It is a type error if, during the static analysis phase, an expression is found to have a static type that is not appropriate for the context in which the expression occurs, or during the dynamic evaluation phase, the dynamic type of a value does not match a required type as specified by the matching rules in 2.5.4 SequenceType Matching. checking function parameter 1 in call array:size(self::node()): cannot convert array(*)('') to a node set"
            />
        </testcase>
        <testcase name="map-function" class="t:map-function">
            <error type="err:XPTY0004"
                message="It is a type error if, during the static analysis phase, an expression is found to have a static type that is not appropriate for the context in which the expression occurs, or during the dynamic evaluation phase, the dynamic type of a value does not match a required type as specified by the matching rules in 2.5.4 SequenceType Matching. checking function parameter 1 in call map:size(self::node()): cannot convert map(*)('') to a node set"
            />
        </testcase>
        <testcase name="map-lookup" class="t:map-lookup">
            <error type="java:org.exist.xquery.XPathException"
                message="exerr:ERROR cannot convert map(*)('') to a node set [at line 15, column 22, source: /db/context-item-tests.xql]&#xA;In function:&#xA;&#x9;t:map-lookup() [6:5:/db/context-item-tests.xql]&#xA;&#x9;test:apply(function(*), item()*) [560:9:jar:file:/Applications/eXist-db.app/Contents/Java/exist-core-5.2.0.jar!/org/exist/xquery/lib/xqsuite/xqsuite.xql]&#xA;&#x9;test:apply(function(*), element(), item()*) [467:9:jar:file:/Applications/eXist-db.app/Contents/Java/exist-core-5.2.0.jar!/org/exist/xquery/lib/xqsuite/xqsuite.xql]&#xA;&#x9;test:call-test(function(*), element(), element()*) [279:32:jar:file:/Applications/eXist-db.app/Contents/Java/exist-core-5.2.0.jar!/org/exist/xquery/lib/xqsuite/xqsuite.xql]. cannot convert map(*)('') to a node set"
            />
        </testcase>
        <testcase name="named-function-function" class="t:named-function-function">
            <error type="err:XPTY0004"
                message="It is a type error if, during the static analysis phase, an expression is found to have a static type that is not appropriate for the context in which the expression occurs, or during the dynamic evaluation phase, the dynamic type of a value does not match a required type as specified by the matching rules in 2.5.4 SequenceType Matching. checking function parameter 1 in call function-arity(self::node()): cannot convert function(*)('') to a node set"
            />
        </testcase>
    </testsuite>
</testsuites>
```

 - OS: macOS 10.15.2
 - eXist-db version: 5.2.0
 - Java Version: 1.8.0_232 (Zulu 8.42.0.23-CA-macosx)

**Additional context**

line-o:
## Description:

Enable function items (functions, arrays, maps) to be used as values of the context item `.`

#### in predicates: 

- [x] `([1, 2], [2, 4])[array:get(., 1) > 1]` returns `[2, 4]` 
- [x] `([1, 2], [2, 4])[(.)?1 > 1]` returns `[2, 4]` 
- [x] `([1, 2], [2, 4])[(.)(1) > 1]` returns `[2, 4]` 
- [x]  `([1, 2], [2, 4])[.?1 > 1]` returns `[2, 4]` 
- [x]  `([1, 2], [2, 4])[.(1) > 1]` returns `[2, 4]` 
- [x] `(map{"a":2}, map{"a":1})[map:get(., "a") > 1]` returns `map{"a":2}` 
- [x] `(map{"a":2}, map{"a":1})[(.)?a > 1]` returns `map{"a":2}` 
- [x] `(map{"a":2}, map{"a":1})[(.)("a") > 1]` returns `map{"a":2}` 
- [x]  `(map{"a":2}, map{"a":1})[.?a > 1]` returns `map{"a":2}` 
- [x] `(map{"a":2}, map{"a":1})[.("a") > 1]` returns `map{"a":2}` 
- [x] `(sum#2, function(){})[function-arity(.) > 1]` returns `sum#2` 
- [x] `(function(){2}, function(){1})[(.)() > 1]` returns `2` 
- [x] `(function(){2}, function(){1})[.() > 1]` returns `2` 

#### after the simple map operator

The map expression status applies to arrays as well

- [x] `(map { 'key': 'value' }, map { 'key': 'other' }) ! map:get(., 'key')` returns `('value', 'other')`
- [x] `(map { 'key': 'value' }, map { 'key': 'other' }) ! (.)?key` returns `('value', 'other')`
- [x] `(map { 'key': 'value' }, map { 'key': 'other' }) ! (.)("key")` returns `('value', 'other')`
- [x] `(map { 'key': 'value' }, map { 'key': 'other' }) ! .("key")` returns `('value', 'other')`
- [x] `(map { 'key': 'value' }, map { 'key': 'other' }) ! (.?key)` returns `('value', 'other')`
- [x] `(map { 'key': 'value' }, map { 'key': 'other' }) ! .?key` returns `('value', 'other')`
- [x] `(function(){2}, function(){1}) ! (.)()` returns `(2,1)` 
- [x] `(function(){2}, function(){1}) ! .()` returns `(2,1)` 
- [x] `(function(){2}, function(){1}) ! (.())` returns `(2,1)` 

Unary lookup statements are not part of this PR but tracked in #1655

- `(map{"a":2}, map{"a":1})[?a > 1]` returns `map{"a":2}`
- `([1, 2], [2, 4])[?1 > 1]` returns `[2, 4]`
- `(map { 'key': 'value' }, map { 'key': 'other' }) ! ?key` returns `('value', 'other')`

### Reference:

fixes #3240 

### Type of tests:

XQsuite tests added

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ContextItemExpression.java
+++ b/exist-core/src/main/java/org/exist/xquery/ContextItemExpression.java
@@ -1,0 +1,80 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery;
+
+import org.exist.xquery.util.ExpressionDumper;
+import org.exist.xquery.value.Item;
+import org.exist.xquery.value.Sequence;
+import org.exist.xquery.value.Type;
+import static org.exist.xquery.ErrorCodes.XPDY0002;
+
+public class ContextItemExpression extends LocationStep {
+
+    private int returnType = Type.ITEM;
+
+    public ContextItemExpression(final XQueryContext context) {
+        // TODO: create class AnyItemTest (one private implementation found in saxon)
+        super(context, Constants.SELF_AXIS, new AnyNodeTest());
+    }
+
+    @Override
+    public void analyze(AnalyzeContextInfo contextInfo) throws XPathException {
+        contextInfo.addFlag(DOT_TEST);
+        // set return type to static type to allow for index use in optimization step
+        returnType = contextInfo.getStaticType();
+
+        super.analyze(contextInfo);
+    }
+
+    @Override
+    public Sequence eval(final Sequence contextSequence, final Item contextItem) throws XPathException {
+        if (contextSequence == null && contextItem == null) {
+            throw new XPathException(this, XPDY0002, "Nothing to select");
+        }
+
+        // function types in context item are returned unchanged
+        if (contextItem != null &&
+                Type.subTypeOf(contextItem.getType(), Type.FUNCTION_REFERENCE)) {
+            return contextItem.toSequence();
+        }
+
+        // function types in context sequence are returned unchanged
+        if (contextSequence != null &&
+                Type.subTypeOf(contextSequence.getItemType(), Type.FUNCTION_REFERENCE)) {
+            return contextSequence;
+        }
+
+        // handle everything else in super class
+        return super.eval(contextSequence, contextItem);
+    }
+
+    @Override
+    public int returnsType() {
+        return returnType;
+    }
+
+    @Override
+    public void dump(final ExpressionDumper dumper) {
+        dumper.display(".");
+    }
+
+}
--- a/exist-core/src/test/xquery/xquery3/bang.xql
+++ b/exist-core/src/test/xquery/xquery3/bang.xql
@@ -19,7 +19,7 @@
  : License along with this library; if not, write to the Free Software
  : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  :)
-xquery version "3.0";
+xquery version "3.1";
 
 module namespace bang="http://exist-db.org/xquery/test/bang";
 
@@ -202,3 +202,90 @@ declare
 function bang:element-left-number-attribute-right() {
     <emp id='1'/> ! number(@id)
 };
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:map-get () {
+    (map {'a':1}, map {'a':2}, map {'a':3}) ! map:get(., 'a')
+};
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:array-get () {
+    ([1], [2], [3]) ! array:get(., 1)
+};
+
+declare
+     %test:assertEquals(1,2,3)
+function bang:map-lookup-context-item () {
+    (map {'a':1}, map {'a':2}, map {'a':3}) ! .?a
+};
+
+declare
+     %test:assertEquals(1,2,3)
+function bang:map-lookup-parenthesized () {
+    (map {'a':1}, map {'a':2}, map {'a':3}) ! (.?a)
+};
+
+declare
+     %test:assertEquals(1,2,3)
+function bang:map-lookup-parenthesized-context-item () {
+    (map {'a':1}, map {'a':2}, map {'a':3}) ! (.)?a
+};
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:array-lookup-standard () {
+    ([1], [2], [3]) ! .?1
+};
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:array-lookup-parenthesized () {
+    ([1], [2], [3]) ! (.?1)
+};
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:array-lookup-parenthesized-context-item () {
+    ([1], [2], [3]) ! (.)?1
+};
+
+declare
+    %test:assertEquals(1,1,1,1,1)
+function bang:mixed-function-types-call () {
+    let $id := function ($a) { $a }
+    return (function ($a) {1}, $id, sum#1, [1], map{1:1}) ! .(1)
+};
+
+declare
+    %test:assertEquals(1,1,1,1,1)
+function bang:mixed-function-types-call-parenthesized () {
+    let $id := function ($a) { $a }
+    return (function ($a) {1}, $id, sum#1, [1], map{1:1}) ! (.(1))
+};
+
+declare
+    %test:assertEquals(1,1,1,1,1)
+function bang:mixed-function-types-call-parenthesized-context-item () {
+    let $id := function ($a) { $a }
+    return (function ($a) {1}, $id, sum#1, [1], map{1:1}) ! (.)(1)
+};
+
+(:~
+ : parse error needs to be fixed first
+ : https://github.com/eXist-db/exist/issues/1655
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:array-lookup-implicit-context () {
+    ([1], [2], [3]) ! ?1
+};
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:map-lookup-implicit-context () {
+    (map {'a':1}, map {'a':2}, map {'a':3}) ! ?a
+};
+
+ :)
--- a/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
+++ b/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
@@ -283,7 +283,8 @@ versionDecl throws XPathException
         { #versionDecl = #(#[VERSION_DECL, v.getText()], enc); }
 	;
 
-setter:
+setter
+:
 	(
 		( "declare" "default" ) =>
 		"declare"! "default"!
@@ -321,12 +322,12 @@ setter:
 	;
 
 preserveMode
-	:
+:
 	( "preserve" | "no-preserve" )
 	;
 
 inheritMode
-	:
+:
 	( "inherit" | "no-inherit" )
 	;
 
@@ -406,7 +407,7 @@ schemaImport throws XPathException
 
 schemaPrefix
 { String prefix = null; }
-	:
+:
 	"namespace"! prefix=ncnameOrKeyword EQ!
 	{ #schemaPrefix = #[NCNAME, prefix]; }
 	|
@@ -416,15 +417,17 @@ schemaPrefix
 annotateDecl! throws XPathException
 :
 	decl:"declare"! ann:annotations!
-	(	("function")=> f:functionDecl[#ann] { #annotateDecl = #f; }
-	| 	("variable")=> v:varDecl[#decl, #ann] { #annotateDecl = #v; }
+	(
+	    ("function") => f:functionDecl[#ann] { #annotateDecl = #f; }
+	    |
+	    ("variable") => v:varDecl[#decl, #ann] { #annotateDecl = #v; }
 	)
-;
+    ;
 
 contextItemDeclUp! throws XPathException
 :
 	decl:"declare"! c:contextItemDecl[#decl] { #contextItemDeclUp = #c; }
-;
+    ;
 
 contextItemDecl [XQueryAST decl] throws XPathException
 :
@@ -442,8 +445,9 @@ contextItemDecl [XQueryAST decl] throws XPathException
 
 
 annotations
-:       (annotation)*
-;
+:
+    (annotation)*
+    ;
 
 annotation
 {
@@ -452,14 +456,13 @@ annotation
 :
 	MOD! name=eqName! (LPAREN! literal (COMMA! literal)* RPAREN!)?
         { #annotation= #(#[ANNOT_DECL, name], #annotation); }
-;
+    ;
 
 eqName returns [String name]
-{
-	name= null;
-}
-:	( name=qName | name=uriQualifiedName )
-;
+{ name= null; }
+:
+    ( name=qName | name=uriQualifiedName )
+    ;
 
 uriQualifiedName returns [String name]
 {
@@ -470,7 +473,7 @@ uriQualifiedName returns [String name]
     ( uri=bracedUriLiteral name=ncnameOrKeyword )
     // convert to Clark notation
     { name = "{" + uri + "}" + name; }
-;
+    ;
 
 bracedUriLiteral returns [String uri]
 {
@@ -479,12 +482,12 @@ bracedUriLiteral returns [String uri]
 :
     lit:BRACED_URI_LITERAL
     { uri = lit.getText(); }
-;
+    ;
 
 functionDeclUp! throws XPathException
 :
 	"declare"! f:functionDecl[null] { #functionDeclUp = #f; }
-;
+    ;
 
 functionDecl [XQueryAST ann] throws XPathException
 { String name= null; }
@@ -546,8 +549,10 @@ uri throws XPathException
 	STRING_LITERAL
 	;
 
-typeDeclaration throws XPathException:
-	"as"^ sequenceType ;
+typeDeclaration throws XPathException
+:
+	"as"^ sequenceType
+	;
 
 // === Types ===
 
@@ -560,7 +565,7 @@ sequenceType throws XPathException
 
 occurrenceIndicator
 :
-        ( QUESTION | STAR | PLUS )
+    ( QUESTION | STAR | PLUS )
 	;
 
 itemType throws XPathException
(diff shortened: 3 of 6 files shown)


exist-core/src/test/xquery/filter-function-items-in-context.xql
```diff
@@ -0,0 +1,69 @@
...
+function filter-function-items-in-context:anonymous-function-arity () {
+    (
+        function() { "foo" },
+        function($x) { "bar" }
+    )
+    [function-arity(.) eq 1] => count()
+};
+
+declare
+    %test:assertEquals(4)
```
line-o, line 60: this should be 6...

exist-core/src/test/xquery/xquery3/bang.xql
```diff
@@ -181,3 +181,46 @@ declare
...
+    %test:assertEquals(1,2,3)
+function bang:array-get () {
+    ([1], [2], [3]) ! array:get(., 1)
+};
+
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:map-lookup () {
+    (map {'a':1}, map {'a':2}, map {'a':3}) ! .?a
```
line-o, line 201: this still fails

exist-core/src/test/xquery/xquery3/bang.xql
```diff
@@ -181,3 +181,46 @@ declare
...
+};
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:array-lookup () {
+    ([1], [2], [3]) ! .?1
```
line-o, line 207: this test also fails

exist-core/src/test/xquery/filter-function-items-in-context.xql
```diff
@@ -0,0 +1,69 @@
...
+function filter-function-items-in-context:mixed-sequence () {
+    (
+        $filter-function-items-in-context:arrs,
+        $filter-function-items-in-context:maps,
+        string-join#1,
+        function($x) { $x }
+    )
+    [. instance of function(*)] => count()
+};
```
line-o, line 69: `([1,2],[1,2,3])[.?2 eq 2]` will be added, but will fail ATM

line-o:
Regression: `1234 ! string-length(.)` returns 4 instead of a type error.

line-o:
Looking at the test results, there is more to do: 
- xquery.dates.DateTests need to be fixed: `//date[. gt xs:dateTime('1960-03-19T19:03:59.782+01:00')]` cannot work because of incompatible types
- RangeIndex optimisations are either broken or tests need to be fixed
- org.exist.xquery.XPathQueryTest.predicates2[local] and [remote] are broken as well because they expect to get a LocationStep

exist-core/src/test/xquery/filter-function-items-in-context.xql
joewiz: How about `([1,2],[1,2,3])[?2 eq 2]`? I'm curious about how `.?` differs from `?` - is the former redundant?

exist-core/src/test/xquery/xquery3/bang.xql
joewiz: Similarly to my comment above, would `?a` work, or is `.?` necessary to select the context item?

joewiz:
@line-o It's exciting to see this in progress!

exist-core/src/test/xquery/filter-function-items-in-context.xql
line-o: Of course these need to be added as well. But since the unary lookup could be seen as a different construct and is also a different path in the parser, I do not know if it should be part of this specific test suite.

exist-core/src/test/xquery/xquery3/bang.xql
```diff
@@ -181,3 +181,46 @@ declare
...
+declare
+    %test:assertEquals(1,2,3)
+function bang:array-lookup-implicit-context () {
+    ([1], [2], [3]) ! ?1
+};
+
+declare
+    %test:assertEquals(1,2,3)
+function bang:map-lookup-implicit-context () {
+    (map {'a':1}, map {'a':2}, map {'a':3}) ! ?a
```
line-o, line 223: @joewiz Unary lookup (`?a`) should/must work but is technically a separate issue.

exist-core/src/test/xquery/xquery3/bang.xql
line-o: This is a separate issue #1655 (see my comment below)

exist-core/src/test/xquery/xquery3/bang.xql
line-o: Looking at my comment the test will be renamed to `bang:map-unary-lookup`

line-o:
I would propose to narrow the scope of this PR and the issue #3240. As this PR already enhances the capabilities of XQuery developers by enabling function items to be in context (the original issue title).
Unary lookups (`?key`) already have a separate issue #1655 and we could also open one for lookups with context item (`.?key`)

joewiz:
@line-o If your proposals help group the problems together better according to the affected regions of the spec/parser/source code, then that's fine with me. So whether it's via 1 or multiple issues/PRs, I would like to see both `.?a` and `?a` work.

line-o:
ContextItemExpression needs to extend LocationStep otherwise a lot is broken: predicates, index optimizations to name the most important ones.

joewiz:
We can restore the request for reviewer when the PR is no longer WIP.

exist-core/src/test/xquery/xquery3/bang.xql
line-o: Currently, one would need to write the above expression as
```xquery
(map {'a':1}, map {'a':2}, map {'a':3}) ! (.)?a
```
Working to resolve this as we are getting closer and closer to a stable solution

exist-core/src/test/xquery/xquery3/bang.xql
line-o: This works... getting there
```xquery
([1], [2], [3]) ! (.)?1
```

line-o:
ContextItemExpression now extends LocationStep but the optimisation and index use seems to be broken still.

extensions/indexes/lucene/src/test/java/org/exist/indexing/lucene/LuceneIndexTest.java
```diff
@@ -363,13 +363,13 @@ public void moreElaborateQueries() throws EXistException, CollectionConfiguratio
             };
 
             String[]  queryTemplates = {
-                    "//.[ft:query(title, '%s')]",
```
line-o, line 366: `//.` is a valid XPath expression after all. This means there is a regression that needs fixing.

line-o:
(old) Range and Lucene-index use and optimisations are fixed.

line-o:
Current status before this is done:
The context item expression does not seem to evaluate correctly. It always needs an extra parenthesis around the expression it is in than it should be:
Examples

|| expected || actual ||
| `//.[1]` | `(//.)[1]` |
| `[1] ! .?1 ` | `[1] ! (.)?1` |
| `[1][.?1>0] ` | `[1][(.?1) > 0]` |

line-o:
@joewiz @wolfgangmm @adamretter What do you think, is this good enough as is to be merged?
`[1] ! .?1` and `[1][.?1 > 0]` are not parsed correctly, which is technically a different issue.

joewiz:
@line-o I think the handful of  `.?` tests could be split into a separate issue so that your fixes for the much broader swath of problems here could be merged. (I still do not know of a case when `?` could not be a perfect substitute for `.?` - though there may well be such cases.) It certainly isn't invalid XQuery, so we should have an issue to track it, but I see no reason to make a clean break and review/merge this PR as is, if it's ready.

My only question: There are still other non-`.?` issues that aren't checked in the original post above. Could you check/confirm their status?

line-o:
@joewiz do you mean function calls on context item?
Example: 
valid but throws: `(function () {}, fn:trace#0) ! .()`
would still need to be written as `(function () {}, fn:trace#0) ! (.)()`

line-o:
I just retested the query above
```query
(function ($a) {$a}, sum#1, [1], map{1:1}) ! .(1)
```
and also
```query
(function ($a) {$a}, sum#1, [1], map{1:1}) ! (.(1))
```
does _NOT_ throw it returns the elements of the original sequence (!)
`(anonymous-function)#1, fn:sum#1, [1], map {1:1}`

I would propose not to merge this, before this is not solved as this is impossible to understand for a developer and leads to hard to debug errors.

line-o:
@wolfgangmm @adamretter do you have an idea/pointer where to look for this?

line-o:
I will also add all three as tests to the xqsuite

joewiz:
@line-o By the "other non-`.?` issues that aren't checked in the original post above", I meant these items whose checkboxes remain unchecked:

- `([1, 2], [2, 4])[?1 > 1]` returns `[2, 4]`
- `([1, 2], [2, 4])[.(1) > 1]` returns `[2, 4]`
- `(map{"a":2}, map{"a":1})[?a > 1]` returns `map{"a":2}`
- `(map{"a":2}, map{"a":1})[.("a") > 1]` returns `map{"a":2}`
- `(function(){2}, function(){1})[.() > 1]` returns `2`
- `(map { 'key': 'value' }, map { 'key': 'other' })` ! `.("key") returns ('value', 'other')`
- `(map { 'key': 'value' }, map { 'key': 'other' })` ! `?key returns ('value', 'other')`
- `(function(){2}, function(){1}) ! .()` returns `(2,1)`
- `(function(){2}, function(){1}) ! (.())` returns `(2,1)`

Since they were unchecked, I assumed the PR doen't yet fix these - but I wanted to ask and confirm, since I had the impression from your commennt today (https://github.com/eXist-db/exist/pull/3496#issuecomment-726192937) that all but the cases containing `.?` were now fixed by this PR? Or perhaps I misunderstand?

line-o:
@joewiz OK, I confirm that only those expressions whose checkbox are checked are working as expected.
My last commit now includes those after simple operator (because those MUST work as they do not throw but return wrong values).

line-o:
I will strike the ones that are covered by #1655

line-o:
side note: Saxon 10HE allows 
```xquery
(function (\$a) {\$a}, sum#1, [1], map{1:1}) ! (1)
```

exist-core/src/test/xquery/xquery3/bang.xql
line-o: done

exist-core/src/test/xquery/filter-function-items-in-context.xql
line-o: done

adamretter:
@line-o Exciting stuff. Can you post the XQTS results for before and after in the description of the PR please?

line-o:
@adamretter I would love to. I was not able to run XQTS tests myself to this day. Can I run the complete testsuite at once?

line-o:
@adamretter how hard would it be to get the XQTS run as an optional GitHub action here?

adamretter:
@line-o Okay, can you open an issue on the XQTS repo then, report what you tried and the steps and the errors that you had...

line-o:
@adamretter see https://github.com/eXist-db/exist-xqts-runner/issues/13

line-o:
With this PR applied the XQTS test success percentage jumps from 79.99% to a 80.08% ... a whopping 0.09%!
(+ some, see https://github.com/eXist-db/exist-xqts-runner/issues/15)

dizzzz (approved):

exist-core/src/main/java/org/exist/xquery/ContextItemExpression.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.exist.xquery.util.ExpressionDumper;
+import org.exist.xquery.value.Item;
+import org.exist.xquery.value.Sequence;
+import org.exist.xquery.value.Type;
+import static org.exist.xquery.ErrorCodes.XPDY0002;
+
+public class ContextItemExpression extends LocationStep {
+
+    int returnType = Type.ITEM;
```
adamretter, line 32: This should be `private`

exist-core/src/test/xquery/xquery3/bang.xql
```diff
@@ -163,11 +163,16 @@ function bang:constructor() {
...
+(:~
+ : implicit context cannot be tested in a xquery testsuite
+ : because it will raise a static error now
 declare
+    %test:pending
     %test:assertTrue
 function bang:implicit-context() {
     count(//* ! local-name(.))
 };
+ :)
```
adamretter, line 175: Does this need to be commented out - it seems to be marked as pending already.

exist-core/src/test/xquery/xquery3/bang.xql
```diff
@@ -202,3 +207,104 @@ declare
...
+};
+
+declare
+     %test:assertEquals(1,2,3)
+function bang:map-lookup-parenthesized-context-item () {
+    (map {'a':1}, map {'a':2}, map {'a':3}) ! (.)?a
+};
+
+(:~
+ : TODO err:XPTY0004
```
adamretter, line 242: Remove the `TODO`

adamretter (changes requested):
Just some tiny tiny changes.

line-o:
this is ready to be merged

exist-core/src/test/xquery/xquery3/bang.xql
```diff
@@ -202,3 +202,90 @@ declare
...
+    let $id := function ($a) { $a }
+    return (function ($a) {1}, $id, sum#1, [1], map{1:1}) ! (.)(1)
+};
+
+(:~
+ : parse error needs to be fixed first
+ : https://github.com/eXist-db/exist/issues/1655
+
+declare
+    %test:assertEquals(1,2,3)
```
adamretter, line 280: I think we discussed on the call - uncommenting these and making them as pending instead?

line-o: No, these cannot be uncommented. Due to parser errors they entire module will not be executed.

adamretter: Ah! Okay :-)

adamretter:
Thanks Juri this looks in good shape.

Issue #3497, opened 2020-08-05
[BUG] ()/fn:base-uri() incorrectly raises XPDY0002


adamretter:
The following query, incorrectly raises `XPDY0002` - `Context sequence is empty and no argument specified`

```xquery
()/fn:base-uri()
```

The XQuery 3.1 F+O spec states:

```
The zero-argument version of the function returns the base URI of the context node: it is equivalent to calling fn:base-uri(.).

The single-argument version of the function behaves as follows:

1. If $arg is the empty sequence, the function returns the empty sequence.
2. Otherwise, the function returns the value of the dm:base-uri accessor applied to the node $arg. This accessor is defined, for each kind of node, in the XDM specification (See Section 5.2 base-uri Accessor DM31).
```

In this instance clause (1) should apply.

PR #3498, opened 2020-08-05
fn:base-uri should not raise XPDY0002 when the context item is empty

adamretter:
Closes https://github.com/eXist-db/exist/issues/3497

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunBaseURI.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunBaseURI.java
@@ -115,9 +115,12 @@ public class FunBaseURI extends BasicFunction {
             }
         } else {
             if (args.length == 0) {
-                if (contextSequence == null || contextSequence.isEmpty()) {
+                if (contextSequence == null) {
                     throw new XPathException(this, ErrorCodes.XPDY0002,
-                        "Context sequence is empty and no argument specified");
+                        "The context item is absent");
+                }
+                if (contextSequence.isEmpty()) {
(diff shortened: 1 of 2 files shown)


adamretter:
@line-o The existing tests for `fn:base-uri` were already in Java, so I continued that way for continuity. Also I still don't trust XQSuite entirely.

Many of the Java tests assert behaviour rather than a single thing, so I don't think my approach is wrong nessecarily.

dizzzz:
Ok for me

joewiz:
@line-o I see your agenda item from Monday's call:

> I would suggest to stick to XQSuite tests wherever possible because they are more accessible to developers unfamiliar with java.

I don't see any notes detailing whether this was discussed, but it seems to be a good suggestion going forward - which I would articulate as: use xqsuite as a default, with the caveats that of course there may be compelling reasons to do otherwise, such as continuity/grouping with other similar tests (so as not to split substantively similar tests across different locations), or to test functionality that xqsuite cannot handle. 

Since we already have working tests here that meet this principle and have secured approval, I will merge this PR.

adamretter:
@joewiz I actually really strongly disagree. As an eXist-db developer, XQSuite tests are horrible to debug, the ones written from Java are much easier to work with. Also I don't trust that XQSuite doesn't still have some false-positives etc. If our XQuery implementation were more spec compliant, I might begin to trust XQSuite further.

joewiz:
@adamretter Re: test suites, let's discuss this at a community call. Since there were no notes, I presume it wasn't discussed, so the agenda item will carry forward to next week's call.

PR #3527, opened 2020-09-14
Add a caller attribute to the output of system:get-running-xqueries#0

adamretter:
Allows the user to identify the query that asked for the list of running queries.


joewiz:
@adamretter The motivation for this PR was the thread with @xatapult, correct? https://exist-db.slack.com/archives/CG2MRUZ35/p1599633218049700

dizzzz (approved):

adamretter:
@joewiz Yes, that is correct.

PR #3528, opened 2020-09-14
Fix an issue in the query history in the Java Admin Client

adamretter:
Previously when selecting a query from the history to restore to the query dialog, the query was incorrectly acquired.


dizzzz (approved):
Great; i had this on my list....

Issue #3039, opened 2019-10-06
Bug in util:eval-with-context's handling of / when context item has been declared


joewiz:
### What is the problem

Since #2155, the use of `/` when a context item has been declared in the prolog has worked correctly in eXist. 

However, when using [`util:eval-with-context`](http://exist-db.org/exist/apps/fundocs/index.html?q=eval-with-context) to evaluate an expression and override the default context, expressions that select `/` are not evaluated relative to the supplied context. Instead, `/` selects all database contents—proving that the context has not been properly set.

### What did you expect

I expected `util:eval-with-context` to correctly handle expressions against `/`.

For example, the following expression should return 1 but instead returns the count of documents in the database:

```xquery
util:eval-with-context("count(/)", (), false(), document { <x/> })
```

In contrast, setting the context item via prolog declaration works correctly. The following query returns `1`, not the count of documents in the database.

```xquery
xquery version "3.1";

declare context item := document { <x/> };

count(/)
```

Allowing users to set a context is precisely the purpose of `util:eval-with-context`. 

### Describe how to reproduce or add a test

The following XQSuite test raises an error:

```xquery
xquery version "3.1";

module namespace ewc="http://exist-db.org/xquery/eval-with-context";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare
    %test:assertEquals(1)
function ewc:root-node() {
    util:eval-with-context('count(/)', (), false(), document { <x/> })
};
```

The error:

```xml
<testcase name="root-node" class="ewc:root-node">
    <failure message="assertEquals failed." type="failure-error-code-1">1</failure>
    <output>376</output>
</testcase>
```

### Context information

- eXist-db version + Git Revision hash: 5.0.0 (release)
- Java version: OpenJDK 12.0.2+10
- Operating system: macOS 10.13.6
- 32 or 64 bit: 64 bit
- How is eXist-db installed: Mac DMG app
- Any custom changes in e.g. conf.xml: none

joewiz:
The motivation for my report is similar to that of #2155. This time I'm reviewing final proofs of the manuscript of my forthcoming XQuery book, and I'm trying to test that all code samples work as expected. To do so, I've pasted them all into a single query as an array of maps; a separate array centrally defines the context items. (This is a pattern used in XQTS.) I'd tried to use `eval-with-context` to supply the context item for these code samples, and found unexpected errors. If there's a fix, I'd be grateful, as it would help with my proofing of the book before publication!

[...]

PR #3529, opened 2020-09-14
Make sure the correct context sequence is used with util:eval-with-context

adamretter:
Closes https://github.com/eXist-db/exist/issues/3039

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/RootNode.java
+++ b/exist-core/src/main/java/org/exist/xquery/RootNode.java
@@ -98,7 +98,30 @@ public class RootNode extends Step {
         // get statically known documents from the context
         DocumentSet ds = context.getStaticallyKnownDocuments();
         if (ds == null || ds.getDocumentCount() == 0) {return Sequence.EMPTY_SEQUENCE;}
-        
+
+        // fix for util:eval-with-context
+        if (contextSequence != null) {
+            if (!contextSequence.isEmpty()) {
+                final Item item = contextSequence.itemAt(0);
+                // context item must be a node
+                if (!Type.subTypeOf(item.getType(), Type.NODE)) {
+                    throw new XPathException(this, ErrorCodes.XPTY0020, "Context item is not a node");
+                }
+                final NodeValue node = (NodeValue)item;
+                // return fn:root(self::node()) treat as document-node()
+                if (node.getImplementationType() == NodeValue.PERSISTENT_NODE) {
+                    return new NodeProxy(((NodeProxy)item).getOwnerDocument());
+                } else {
+                    if (node.getType() == Type.DOCUMENT) {
+                        return node;
+                    }
+                    return (org.exist.dom.memtree.DocumentImpl) node.getOwnerDocument();
+                }
+            } else {
+                return Sequence.EMPTY_SEQUENCE;
+            }
+        }
+
 //        // if the expression occurs in a nested context, we might have cached the
 //        // document set
 //        // TODO: disabled cache for now as it may cause concurrency issues
--- a/exist-core/src/test/java/org/exist/storage/ShutdownTest.java
+++ b/exist-core/src/test/java/org/exist/storage/ShutdownTest.java
@@ -90,7 +90,7 @@ public class ShutdownTest {
 
                 final XQuery xquery = pool.getXQueryService();
                 assertNotNull(xquery);
-                final Sequence result = xquery.execute(broker, "//SPEECH[contains(LINE, 'love')]", Sequence.EMPTY_SEQUENCE);
+                final Sequence result = xquery.execute(broker, "//SPEECH[contains(LINE, 'love')]", null);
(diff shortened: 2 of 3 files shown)


dizzzz:
Please could we have Joe's test included as well?

```xquery
xquery version "3.1";

module namespace ewc="http://exist-db.org/xquery/eval-with-context";

declare
    %test:assertEquals(1)
function ewc:root-node() {
    util:eval-with-context('count(/)', (), false(), document { <x/> })
};
```

dizzzz (changes requested):
Add joe's test case

dizzzz:
A lot of travis tests fail, in the range index tests for example.

adamretter:
@dizzzz Indeed. Those range index tests are not valid

line-o:
What is wrong with the range index tests @adamretter ?
`xqts.org.exist-db.xquery.range.test.contains-string` for instance

joewiz:
@adamretter Both Travis and AppVeyor failed. Should these be restarted, or is there a known problem with the PR?

adamretter:
@joewiz From the Travis log it looks like there are a lot of failing tests related to Range Indexes now.

joewiz:
Tests may be fixed by #3605; rebase and test after it's merged.

joewiz:
@adamretter Do you think it would be possible to relax the 4th parameter of `util:eval-with-context` to be `$context as item()?` instead of `node()?`? This would allow the counterpart of the following query:

```xquery
xquery version "3.1";

declare context item := [1, 2, 3];

array:size(.)
```

... which returns `3`, to be accomplished dynamically using `util:eval-with-context()`:

```xquery
xquery version "3.1";

util:eval-with-context(array:size(.), (), false(), [1, 2, 3])
```

... which currently raises the error:

> err:XPDY0002 Nothing to select [at line 3, column 35]

Line 3 column 35 resolves to the `.` inside of `array:size(.)`.

This change would bring parity between this function and the [context item declaration](https://www.w3.org/TR/xquery-31/#id-context-item-declarations). It would open up the possibility of setting the context dynamically in xqsuite tests. For example, at the moment it's not possible to adapt my suggested test for @dizzzz's PR https://github.com/eXist-db/exist/pull/3644#issuecomment-736877075 to xqsuite due to the `node()` restriction. Relaxing it and allowing `item()` would allow Dannes to add this test to his PR.

adamretter:
> Do you think it would be possible to relax the 4th parameter of util:eval-with-context to be $context as item()? instead of node()?

@joewiz The 4th parameter is `$eval-context-item` not `$context`. Also, the `$eval-context-item` parameter is already an `item()?`. Did you mean the 4th parameter?

> `util:eval-with-context('array:size(.)', (), false(), [1, 2, 3])`

That query currently evaluates to `1`. Likely because the context item has to be a Sequence, and so we end up with `([1,2,3])`.

joewiz:
@adamretter I'm terribly sorry for the muddled question. Let me start over, because I agree that what I wrote above was quite unclear. Here is my observation:

Building a fresh checkout of the PR's branch, the following sample queries all return the error `err:XPDY0002 Nothing to select`—with the location of the error pointing to the `.` in the expression supplied in the 1st parameter. This error message originates from https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/xquery/ContextItemExpression.java#L50-L52, which suggests that `util:eval-with-context()` is not setting a context item (or a context sequence), despite the 4th parameter in all cases clearly being a valid item that conforms to the function signature's type declaration, `item()?`. 

- `util:eval-with-context(., (), false(), <x/>)`
- `util:eval-with-context(., (), false(), document { <x/> })`
- `util:eval-with-context(count(.), (), false(), (1, 2, 3))`
- `util:eval-with-context('array:size(.)', (), false(), [1, 2, 3])`
- `util:eval-with-context(.?foo, (), false(), map { "foo": "bar" })`
- `util:eval-with-context(., (), false(), ())`

According to my understanding, instead of raising this error, these expressions should produce the identical result as if the context item declaration was used to set the context item. For example, adapting the 1st example above:

```xquery
xquery version "3.1";

declare context item := <x/>; 

.
```

This query returns the expected result, `<x/>`. But `util:eval-with-context(., (), false(), <x/>)` raises the `Nothing to select` error.

joewiz:
Oh my, I just realized my error. I was supplying an expression in the 1st parameter instead of a string! 

Sorry for the noise @adamretter!

adamretter:
@joewiz Okay... so all is okay now?

joewiz (approved):
@adamretter Yes!

joewiz:
@adamretter I see this PR is marked "Draft" - are there still other changes you plan? I see that CI tests all pass.

adamretter:
@joewiz It was only draft whilst I was waiting for other PRs to be merged first. So yes, I think it is now good to go.

dizzzz (approved):

Issue #846, opened 2015-12-03
Incomplete regex character class with XQuery analyze-string()


ptrckb:
It seems that the shorthand character class '\w' does not include some (any?) accented letter in patterns for `analyze-string()`.

For example:

``` xquery
analyze-string('é', '\w')
```

return:

``` xml
<fn:analyze-string-result xmlns:fn="http://www.w3.org/2005/xpath-functions">
<fn:non-match>é</fn:non-match>
</fn:analyze-string-result>
```

Still the same shorthand character class does include them in patterns for `matches()`.

``` xquery
matches('é', '\w')        
```

return

``` xquery
true
```

Tested with 2.2 and develop (395c00091059929adcef6f537c662e60f0a4e3a8)

dizzzz:
I might be wrong but the xpath regexp is different from the Java one. Maybe the wrong pattern || engine is used here....

joewiz:
@ptrckb Somewhere between 2015 and today (i.e., eXist 5.3.0-SNAPSHOT develop branch), eXist began to return the correct results from `fn:analyze-string`:

```xml
<fn:analyze-string-result xmlns:fn="http://www.w3.org/2005/xpath-functions">
    <fn:match>é</fn:match>
</fn:analyze-string-result>
```

So, happily, this issue can be closed!

Issue #59, opened 2013-10-08
matches(): non-compliant to the spec and buggy


rvdb:
eXist's implementation of matches(xs:string*, xs:string) differs w.r.t. the official matches(xs:string?, xs:string) function. As the cardinality for the first string parameter suggests, eXist accepts multiple strings as first argument for matches(), while this throws an error against the standard XPath function. This can be demonstrated by running following query with Saxon and eXist (tested with eXist-2.1):

```xquery
  let $data :=
    <data>
      <entry n="1">
        <val>abc</val>
      </entry>
      <entry n="2">
        <val>xyz</val>
        <val>bcd</val>
      </entry>
      <entry n="3">
        <val>123</val>
        <val>haha</val>
      </entry>
      <entry n="4">
        <val>haha</val>
        <val>123</val>
      </entry>
  </data>
  return $data//entry[matches(val,'a')]
```

While Saxon throws an error ("XPTY0004: A sequence of more than one item is not allowed as the first argument of matches()"), eXist lets this pass. Yet, there seems to be a bug when the first argument of matches is a sequence of strings. Above query returns following results:

```xml
  <entry n="1">
    <val>abc</val>
  </entry>
  <entry n="4">
    <val>haha</val>
    <val>123</val>
  </entry>
```

While both entries 3 and 4 should be returned, 3 is omitted from the results. This illustrates that if the first argument to eXist's matches() function is a sequence, only the first item is evaluated.

This bug aside, the non-compliant behaviour is documented in the eXist function documentation at http://demo.exist-db.org/exist/apps/fundocs/view.html?uri=http://www.w3.org/2005/xpath-functions&location=java:org.exist.xquery.functions.fn.FnModule#matches.2. Still, I'm curious about what motivated this discrepancy from the official spec, and the decision to keep the eXist matches() function unchanged in the http://www.w3.org/2005/xpath-functions namespace? IMO, this makes it too easy to write non-compliant XQuery code.

adamretter:
This issue https://github.com/eXist-db/exist/issues/581 would also have an impact here.

adamretter:
It is unlikely we are going to fix this as it is a useful optimisation. However, before we close this ticket we should add to the documentation that `fn:matches`, `fn:contains`, `fn:starts-with` and `fn:ends-with` are non-standard and update the descriptions of the function signatures to mention this also.

duncdrum:
ok 2013… *yay* for docs

joewiz:
[...]

PR #3530, opened 2020-09-14
fn:replace and fn:analyze-string are now implemented with Saxon

adamretter:
In an attempt to fix some bugs with eXist-db's `fn:replace` and `fn:analyze-string`, these new versions just wrap the Saxon equivalents.

If people like this, we could also extend it to `fn:matches`.

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/BrokerPool.java
+++ b/exist-core/src/main/java/org/exist/storage/BrokerPool.java
@@ -23,6 +23,7 @@ package org.exist.storage;
 
 import com.evolvedbinary.j8fu.fsm.AtomicFSM;
 import com.evolvedbinary.j8fu.fsm.FSM;
+import com.evolvedbinary.j8fu.lazy.AtomicLazyVal;
 import net.jcip.annotations.GuardedBy;
 import net.jcip.annotations.ThreadSafe;
 import org.apache.logging.log4j.LogManager;
@@ -86,6 +87,7 @@ import java.util.*;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
@@ -376,6 +378,13 @@ public class BrokerPool extends BrokerPools implements BrokerPoolConstants, Data
     private Optional<ExistRepository> expathRepo = Optional.empty();
 
     private StartupTriggersManager startupTriggersManager;
+
+    /**
+     * Configuration for Saxon.
+     *
+     * One instance per-database, lazily initialised.
+     */
+    private AtomicLazyVal<net.sf.saxon.Configuration> saxonConfig = new AtomicLazyVal<>(net.sf.saxon.Configuration::newConfiguration);
 
     /**
      * Creates and configures the database instance.
@@ -1919,6 +1928,10 @@ public class BrokerPool extends BrokerPools implements BrokerPoolConstants, Data
 
     public PluginsManager getPluginsManager() {
         return pluginManager;
+    }
+
+    public net.sf.saxon.Configuration getSaxonConfiguration() {
+        return saxonConfig.get();
     }
 
     /**
--- a/exist-core/src/main/java/org/exist/storage/RegexMatcher.java
+++ b/exist-core/src/main/java/org/exist/storage/RegexMatcher.java
@@ -26,26 +26,24 @@ import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
 import org.exist.EXistException;
+import org.exist.util.PatternFactory;
 
 /**
  * A {@link org.exist.storage.TermMatcher} that matches index entries against a
  * regular expression. Used by {@link org.exist.storage.NativeValueIndex}.
- * 
- * @author wolf
- *
  */
 class RegexMatcher implements TermMatcher {
 	
 	private Matcher matcher;
-    private boolean matchAll = false;
+    private boolean matchAll;
 
-    public RegexMatcher(String expr, int flags) throws EXistException {
+    public RegexMatcher(final String expr, final int flags) throws EXistException {
         this(expr, flags, false);
     }
-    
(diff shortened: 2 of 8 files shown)


joewiz:
@adamretter Are the Saxon-specific flags now supported as well? (See https://www.saxonica.com/documentation9.5/functions/fn/matches.html.)

> Saxon implements some extensions to the flags argument. The value of this argument contains the standard flags defined in the spec (`i`, `m`, `s`, `x`, `q`) followed optionally by a semicolon and then Saxon-specific flags. The Saxon-specific flags are:
> 
> - `j` - Use the standard Java regular expression engine, allowing native regex syntax
> - `n` - Use the standard .NET regular expression engine, allowing native regex syntax
> - `g` - Switch on diagnostics
> - `k` - Treat unknown Unicode block names as matching any character (this is the default for XSD 1.1)
> - `K` - Treat unknown Unicode block names as an error (this is the default for XSD 1.0 and for XPath/XSLT/XQuery)

See discussion of this in Saxon and BaseX at https://github.com/eXist-db/exist/issues/846.

joewiz:
+1 to the idea of extending your approach to `fn:matches`.

Do you by chance show any change in XQTS performance as a result of this PR?

adamretter:
> Are the Saxon-specific flags now supported as well?

@joewiz yes I think so, but obviously not the .NET one ;-)

This work was actually done in March, but I forgot to backport it to eXist-db...

I haven't run the XQTS for this against eXist-db. I would welcome someone else trying it and posting results though...

dizzzz (changes requested):
PR is not complete?


image

adamretter:
@dizzzz Ah I think I screwed up the rebase a little... Should be fixed now I hope.

dizzzz (approved):
The drawback I see is that this is effectively a tighter integration with SAXON; here the external API can changed between versions...

(note: I tried to update to saxon10 without luck yet)

adamretter:
@dizzzz That's correct. eXist-db would become more dependant on Saxon internal APIs. In FusionDB where we implemented fn:transform we already have such a dependency, perhaps eXist-db will in future too?

dizzzz:
I very agree :-)

dizzzz (approved):

ljo (approved):
With the fixed rebase it looks good. The tighter dependency is of course something we need to note for coming strategic decisions.

wolfgangmm (approved):
Do we know which bugs this fixes? If so, it might be good to at least list them in the release notes, so users know in case they have relied on buggy behaviour. As far as I can see, `analyze-string` is used quite a lot.

joewiz:
@adamretter I can confirm that with this PR, the [Saxon-specific `j` flag](https://www.saxonica.com/documentation9.5/functions/fn/matches.html) is indeed now supported. (I didn't really know how to test the other Saxon-specific flags; I just checked `j`.) For example, users could now use this approach to access the handy [`\b` regex metacharacter](https://www.regular-expressions.info/wordboundaries.html) for word boundaries, which isn't supported in standard XPath regex:

```xquery
analyze-string("foo bar", "\b", ";j")
```

Your idea to extend this PR to include `fn:matches` makes a lot of sense, since users might expect that from the example above, the following one should also work:

```xquery
matches("foo bar", "\b", ";j")
```

But, naturally, this variant currently fails with the following error:

> err:FORX0001 Invalid regular expression flag: ; 

Are you still open to extending this PR to include `fn:matches`?

joewiz:
@adamretter I tried running XQTS but got this compilation error:

```

**Comment by dizzzz** [comment]
`FastByteArrayInputStream` has been replaced by the apache.org foundation alternative.

**Comment by dizzzz** [comment]
hopefully fixed with https://github.com/eXist-db/exist-xqts-runner/pull/10

**Comment by joewiz** [comment]
Thanks to @adamretter and @dizzzz, I have now been able to run XQTS on this PR's branch and on the develop branch. There were no differences in `fn:replace` tests, but there was a slight difference in `fn:analyze-string`:

**eXist 5.3.0-SNAPSHOT develop 8ff4383d6d19fe7a525b49702fa4aa515e4ef5aa** - 7 failures

> Screen Shot 2020-10-27 at 4 43 13 PM


**this PR** - 6 failures

> Screen Shot 2020-10-27 at 4 13 41 PM

Analyzing the test results, both branches fail identically on the same 6 tests (analyzeString-010, 009, 011, 015, 014, and 903), but this PR passes on one more test that the develop branch fails on: [analyzeString-017](https://github.com/w3c/qt3tests/blob/master/fn/analyze-string.xml#L210-L218). 

This test performs the expression:

```xquery
analyze-string("banana", "(b)(x?)")
```

eXist-develop returns the following incorrect result - with one nested group element instead of two sibling group elements:

```xml
<fn:analyze-string-result xmlns:fn="http://www.w3.org/2005/xpath-functions">
    <fn:match>
        <fn:group nr="1">b<fn:group nr="2"/></fn:group>
    </fn:match>
    <fn:non-match>anana</fn:non-match>
</fn:analyze-string-result>
```

This PR returns the following correct result:

```xml
<fn:analyze-string-result xmlns:fn="http://www.w3.org/2005/xpath-functions">
    <fn:match>
        <fn:group nr="1">b</fn:group>
        <fn:group nr="2"/>
    </fn:match>
    <fn:non-match>anana</fn:non-match>
</fn:analyze-string-result>
```

All in all, I would say this PR is net win: no regressions, improved XQTS results, and a new hook for accessing Java native regex.

It's possible some code depends on the buggy behavior, but I think if we state in the release notes that this PR results in eXist passing [analyzeString-017](https://github.com/w3c/qt3tests/blob/master/fn/analyze-string.xml#L210-L218), then developers upgrading their apps would be able to check their code accordingly.

As thrilled as I'd be to see this PR extended to `fn:matches`, I remembered that [our stance on making this function spec compliant](https://github.com/eXist-db/exist/issues/59#issuecomment-521397805) was that it would require a major release. One option to us, then, would be to (1) issue a deprecation notice now in this PR (bound for the next patch or minor release), that eXist's buggy handling of regex would go away in an upcoming major release. Then, if you were open to it, (2) open a PR tagged as 6.0.0, extending this PR to `fn:matches`. How does that sound?

dizzzz:
For wolf's request regarding the release notes: should file a ticket for this? how to make sure that we document it?
Happy to pull-in  the PR.....

dizzzz (approved):

PR #3535, opened 2020-09-20
Remove document metadata indirection

adamretter:
I back-ported this improvement from FusionDB.

I previously removed the lazy-loading do Document metadata in eXist-db due to concurrency issues. However, the indirection between a persistent document and its metadata was left in-place.

This PR removes the indirection. This removes one object per Document object, resulting in a memory reduction for DocuemntImpl. It also reduces the function call indirection involved in looking up document metadata, resulting in a small-performance increase.


dizzzz:
I will have a look, as it is a larger PR i need some time

dizzzz:
in macOS, a deadlock is seems te be detected....

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   CollectionLocksTest.stress_noDeadlock_writeWrite_subtree:554->stress_noDeadlock:690 Stress test likely showed a thread deadlock
```

dizzzz (approved):

dizzzz:
@adamretter looks good
@line-o @joewiz @wolfgangmm  for this change I need to update the replication extension.

adamretter:
@dizzzz regarding:
> Stress test likely showed a thread deadlock

macOS on Travis can be quite slow at times. Those tests have a timeout in them, whereby if the test does not complete before a timeout, then it assumes a deadlock has occurred and reports as such. This sometimes pops up on Travis form time to time, re-running the tests usually causes it to go away - you can now see all tests are passing ;-)

dizzzz:
I restarted it already :/

adamretter:
@dizzzz No problem - all is passing OK :-)

wolfgangmm:
Needs wrapper `getMetadata` to go into 5. This method should be deprecated and then removed in 6.

adamretter:
@wolfgangmm @dizzzz I just pushed a commit which makes this API compatible with the eXist-db 5.x.x series whilst not sacrificing memory use.

wolfgangmm (approved):
Looks ok to me now. @dizzzz do you want to check if your replication extension works with the  added wrapper?

dizzzz:
On my way....

dizzzz:
Looks good!

dizzzz (approved):

adamretter:
@dizzzz @wolfgangmm thanks for the review and merge

dizzzz:
As always

PR #3544, opened 2020-09-28
add util:base64-encode-url-safe

line-o:
### Description:

There is a number of use cases (data-url, JWT) where the data must be encoded as base64 url-safe.
This PR adds a new utility function adding this functionality.

Compare the output of 

```
util:base64-encode(".ÿd"),
util:base64-encode-url-safe(".ÿd")
```

Additionally, this PR deprecates `util:base64-encode#2`. The trim-parameter serves no purpose since the output cannot be chunked because of the encoder that is used internally.

The class is refactored for readability and the formatting is fixed.

### Reference:

[RFC4648](https://tools.ietf.org/html/rfc4648)

### Type of tests:

Java tests were added to the tests that were already existing, ensuring that the functionality will persist even when implementation changes in the future.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
@@ -42,82 +42,80 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 /**
  * Base64 String conversion functions.
  *
- * @author  Andrzej Taramina (andrzej@chaeron.com)
+ * @author Andrzej Taramina (andrzej@chaeron.com)
  */
 
-public class Base64Functions extends BasicFunction
-{
-    protected static final Logger           logger       = LogManager.getLogger( Base64Functions.class );
+public class Base64Functions extends BasicFunction {
+    protected static final Logger logger = LogManager.getLogger(Base64Functions.class);
 
-   public final static FunctionSignature[] signatures = {
-        new FunctionSignature(
-            new QName( "base64-encode", UtilModule.NAMESPACE_URI, UtilModule.PREFIX ),
-            "Encodes the given string as Base64",
-            new SequenceType[] {
-                new FunctionParameterSequenceType( "string", Type.STRING, Cardinality.ZERO_OR_ONE, "The string to be Base64 encoded" )
-            },
-            new FunctionReturnSequenceType( Type.STRING, Cardinality.ZERO_OR_ONE, "the Base64 encoded output, with trailing newlines trimmed" )
-        ),
-		
-		 new FunctionSignature(
-            new QName( "base64-encode", UtilModule.NAMESPACE_URI, UtilModule.PREFIX ),
-            "Encodes the given string as Base64",
-            new SequenceType[] {
-                new FunctionParameterSequenceType( "string", Type.STRING, Cardinality.ZERO_OR_ONE, "The string to be Base64 encoded" ),
-				new FunctionParameterSequenceType( "trim", Type.BOOLEAN, Cardinality.EXACTLY_ONE, "Trim trailing newlines?" )
-            },
-            new FunctionReturnSequenceType( Type.STRING, Cardinality.ZERO_OR_ONE, "the Base64 encoded output" ),
-			"This function is deprecated. The output does not need to be trimmed, please use util:base64-encode#1 instead."
-        ),
-		
-         new FunctionSignature(
-            new QName( "base64-decode", UtilModule.NAMESPACE_URI, UtilModule.PREFIX ),
-            "Decode the given Base64 encoded string back to clear text",
-            new SequenceType[] {
-                new FunctionParameterSequenceType( "string", Type.STRING, Cardinality.ZERO_OR_ONE, "The Base64 string to be decoded" )
-            },
-            new FunctionReturnSequenceType( Type.STRING, Cardinality.ZERO_OR_ONE, "the decoded output" )
-        ),
+    public final static FunctionSignature[] signatures = {
+            new FunctionSignature(
+                    new QName("base64-encode", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
+                    "Encodes the given string as Base64",
+                    new SequenceType[]{
+                            new FunctionParameterSequenceType("string", Type.STRING, Cardinality.ZERO_OR_ONE, "The string to be Base64 encoded")
+                    },
+                    new FunctionReturnSequenceType(Type.STRING, Cardinality.ZERO_OR_ONE, "the Base64 encoded output, with trailing newlines trimmed")
+            ),
 
-		new FunctionSignature(
-			new QName( "base64-encode-url-safe", UtilModule.NAMESPACE_URI, UtilModule.PREFIX ),
-			"Encodes the given string as Base64 (url-safe)",
-			new SequenceType[] {
-					new FunctionParameterSequenceType( "string", Type.STRING, Cardinality.ZERO_OR_ONE, "The string to be Base64 encoded (url-safe)" )
-			},
-			new FunctionReturnSequenceType( Type.STRING, Cardinality.ZERO_OR_ONE, "the Base64, url-safe encoded output without padding" )
-		)
+            new FunctionSignature(
+                    new QName("base64-encode", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
+                    "Encodes the given string as Base64",
+                    new SequenceType[]{
+                            new FunctionParameterSequenceType("string", Type.STRING, Cardinality.ZERO_OR_ONE, "The string to be Base64 encoded"),
+                            new FunctionParameterSequenceType("trim", Type.BOOLEAN, Cardinality.EXACTLY_ONE, "Trim trailing newlines?")
+                    },
+                    new FunctionReturnSequenceType(Type.STRING, Cardinality.ZERO_OR_ONE, "the Base64 encoded output"),
+                    "This function is deprecated. The output does not need to be trimmed, please use util:base64-encode#1 instead."
+            ),
 
-	};
-	
+            new FunctionSignature(
+                    new QName("base64-decode", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
+                    "Decode the given Base64 encoded string back to clear text",
+                    new SequenceType[]{
+                            new FunctionParameterSequenceType("string", Type.STRING, Cardinality.ZERO_OR_ONE, "The Base64 string to be decoded")
+                    },
+                    new FunctionReturnSequenceType(Type.STRING, Cardinality.ZERO_OR_ONE, "the decoded output")
+            ),
 
-    public Base64Functions( XQueryContext context, FunctionSignature signature )
-    {
-        super( context, signature );
+            new FunctionSignature(
+                    new QName("base64-encode-url-safe", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
+                    "Encodes the given string as Base64 (url-safe)",
+                    new SequenceType[]{
+                            new FunctionParameterSequenceType("string", Type.STRING, Cardinality.ZERO_OR_ONE, "The string to be Base64 encoded (url-safe)")
+                    },
+                    new FunctionReturnSequenceType(Type.STRING, Cardinality.ZERO_OR_ONE, "the Base64, url-safe encoded output without padding")
+            )
+
+    };
+
+
+    public Base64Functions(XQueryContext context, FunctionSignature signature) {
+        super(context, signature);
     }
-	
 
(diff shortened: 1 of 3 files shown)


exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
-
-    public Sequence eval( Sequence[] args, Sequence contextSequence ) throws XPathException
-    {
-		Sequence value	= Sequence.EMPTY_SEQUENCE;
-		boolean  trim	= true;
-		
-        if( !args[0].isEmpty() ) {       
-			final String str = args[0].getStringValue();
-			
-			if( args.length == 2 ) {
```
dizzzz, line 97: can this be removed without side effects?

exist-core/src/test/java/org/exist/xquery/functions/util/Base64FunctionsTest.java
```diff
@@ -66,5 +66,28 @@ public void testBase64EncodeDecode() throws XMLDBException {
         assertEquals("This is a test!", r);
     }
 
+    @Test
+    public void testBase64EncodeUrlSafeNoSpecial() throws XMLDBException {
```
dizzzz, line 70: should we have a test showing that the old 2 param function still works as it did previously?

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
line-o: It was not removed:
`if (args[0].isEmpty()) { return Sequence.EMPTY_SEQUENCE; }`
is the same condition just flipped.

exist-core/src/test/java/org/exist/xquery/functions/util/Base64FunctionsTest.java
line-o: Well, I will add it. I thought the two parameter function had no tests and is deprecated now what is the use. But I see why you would want that.

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
-            "Decode the given Base64 encoded string back to clear text",
-            new SequenceType[] {
-                new FunctionParameterSequenceType( "string", Type.STRING, Cardinality.ZERO_OR_ONE, "The Base64 string to be decoded" )
-            },
-            new FunctionReturnSequenceType( Type.STRING, Cardinality.ZERO_OR_ONE, "the decoded output" )
-        )
+public class Base64Functions extends BasicFunction {
+    protected static final Logger logger = LogManager.getLogger(Base64Functions.class);
+
+    public final static FunctionSignature signatures[] = {
```
adamretter, line 51: If you ware modifying/cleaning up the function definitions, may I suggest that it might be better to write them using the FunctionDSL.

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
+
+
+    public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathException {
+        if (args[0].isEmpty()) {
+            return Sequence.EMPTY_SEQUENCE;
+        }
+
+        final String str = args[0].getStringValue();
+
+        if (isCalledAs("base64-encode")) {
```
adamretter, line 105: This would be better as a constant

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
+            String b64Str = Base64.encodeBase64String(str.getBytes(UTF_8));
```
adamretter, line 106: this should be final

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
+        }
+
+        if (isCalledAs("base64-encode-url-safe")) {
+            String b64Str = Base64.encodeBase64URLSafeString(str.getBytes(UTF_8));
+            return new StringValue(b64Str);
+        }
+
+        // isCalledAs("base64-decode")
+        // Base64.decodeBase64 can handle standard and url-safe base64 encoded data
+        final byte[] data = Base64.decodeBase64(str);
```
adamretter, line 117: This will use a lenient decoding policy, see `org.apache.commons.codec.CodecPolicy`. This could be a good time to improve this so that it uses the strict policy.

line-o: Well, that is something I would like to leave for somebody else to do, as I do not know what it does or does not.

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
-            },
-            new FunctionReturnSequenceType( Type.STRING, Cardinality.ZERO_OR_ONE, "the decoded output" )
-        )
+public class Base64Functions extends BasicFunction {
+    protected static final Logger logger = LogManager.getLogger(Base64Functions.class);
+
+    public final static FunctionSignature signatures[] = {
+            new FunctionSignature(
+                    new QName("base64-encode", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
+                    "Encodes the given string as Base64",
```
adamretter, line 54: May I suggest improving the description to: `Encodes the given string as Base64 as defined by RFC 2045 §6.8 Base64 Content-Transfer-Encoding`.

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
+                    new SequenceType[]{
+                            new FunctionParameterSequenceType("string", Type.STRING, Cardinality.ZERO_OR_ONE, "The string to be Base64 encoded")
+                    },
+                    new FunctionReturnSequenceType(Type.STRING, Cardinality.ZERO_OR_ONE, "the Base64 encoded output, with trailing newlines trimmed")
+            ),
+
+            new FunctionSignature(
+                    new QName("base64-encode", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
+                    "Encodes the given string as Base64",
```
adamretter, line 63: `Encodes the given string into the Base64 alphabet as defined by RFC 2045 §6.8 Base64 Content-Transfer-Encoding.`

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
+                    "Decode the given Base64 encoded string back to clear text",
+                    new SequenceType[]{
+                            new FunctionParameterSequenceType("string", Type.STRING, Cardinality.ZERO_OR_ONE, "The Base64 string to be decoded")
+                    },
+                    new FunctionReturnSequenceType(Type.STRING, Cardinality.ZERO_OR_ONE, "the decoded output")
+            ),
+
+            new FunctionSignature(
+                    new QName("base64-encode-url-safe", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
+                    "Encodes the given string as Base64, url-safe. No padding and use - and _ instead of + and /.",
```
adamretter, line 83: The URL Safe alphabet used by Apache Commons Codec is only one of several possibilities (https://en.wikipedia.org/wiki/Base64#Variants_summary_table), with each being incompatible with each other. Are we certain this URL safe alphabet is the "defacto standard" which is also used for JWT etc?

Looking into the Commons Codec code, it's URL Safe variant appears to be compliant with RFC 4648. Do you agree? If so we should likely write the description to be more accurate like:

```
Encodes the given string into a Base64 alphabet as defined by RFC 4648 §5 Base 64 Encoding with URL and Filename Safe Alphabet.
```

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
+            )
+
     };
-	
 
-    public Base64Functions( XQueryContext context, FunctionSignature signature )
-    {
-        super( context, signature );
+
+    public Base64Functions(XQueryContext context, FunctionSignature signature) {
```
adamretter, line 93: final arguments please.

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
```diff
@@ -42,76 +42,80 @@
...
-
-	        	final byte[] data = Base64.decodeBase64(str);
-				value = new StringValue(new String(data, UTF_8));
-	        }
-		}
-		
-		return( value );
+
+
+    public Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathException {
```
adamretter, line 98: final arguments please

adamretter (changes requested):
In genera, it looks good. Just a few minor suggestions for improvement

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
adamretter: Lenient means that any invalid characters in the data you want to decode will be ignored. Strict means that it will raise an exception. We should not say that it is okay to decode junk ;-)

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
line-o: I do not see the need to do that.

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
line-o: Please explain further what should be a constant here.

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
line-o: I see that as a breaking change and will not change the current behaviour.

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
line-o: Well, I referenced RFC 4648 - this IS the de-facto standard.

line-o: Explaining what it actually does in contrast to the default behaviour helps me - as a developer more than some abstract reference to an RFC.

exist-core/src/test/java/org/exist/xquery/functions/util/Base64FunctionsTest.java
line-o: test added

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
line-o: Do you agree @dizzzz ?

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
line-o: I added: `(sRFC 2045 §6.8)` to the description

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
line-o: No benefit improving the description of deprecated functions

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
line-o: `(see RFC 4648 §5)` was added to description

exist-core/src/main/java/org/exist/xquery/functions/util/Base64Functions.java
dizzzz: "base64-encode"

dizzzz (approved):

line-o:
Could you someone restart the appveyor build, please.

dizzzz:
restarted

dizzzz:
LGTM but 
image

dizzzz (approved):

joewiz:
@dizzzz During this week's Community Call @line-o stated that he had addressed the requests in his follow-up commits. Now that we have an approval of the PR, I will clear the request for changes and merge.

Issue #3375, opened 2020-04-18
[feature] Make eXist-db Java13 compliant 


dizzzz:
During running unit tests, at least one test fails, only with Java13:

```
[WARNING] Tests run: 156, Failures: 0, Errors: 0, Skipped: 2, Time elapsed: 3.288 s - in xquery.arrays.ArrayTests
1466[WARNING] Tests run: 458, Failures: 0, Errors: 0, Skipped: 3, Time elapsed: 34.446 s - in xquery.xquery3.XQuery3Tests
1467[INFO] Tests run: 58, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 33.929 s - in xquery.optimizer.OptimizerTests
1468[INFO] 
1469[INFO] Results:
1470[INFO] 
1471[ERROR] Errors: 
1472[ERROR]   VariableByteStreamTest.copyTo:82 » ArrayIndexOutOfBounds Index 3106157 out of ...
1473[INFO] 
1474[ERROR] Tests run: 4673, Failures: 0, Errors: 1, Skipped: 102
1475[INFO] 
```

adamretter:
Just to let you know @dizzzz I am digging into this. It's a tricksy one...

dizzzz:
@adamretter the issue is not visible in all PRs....

adamretter:
@dizzzz Indeed. The issue I am tracing appears to be non-deterministic on JDK 13 :-/ I am still not sure I believe it!

reinhapa:
@adamretter do we have something like code coverage in the build? We got something similar in our build when using jacoco...

adamretter:
@reinhapa No, there is no code coverage in that build or any other form of bytecode instrumentation that I am aware of.

I do have a branch where I introduce JaCoCo but that is separate... and that has other problems... https://github.com/eXist-db/exist/pull/3341

adamretter:
I finally managed to isolate the issue to an independent stripped down test case - https://github.com/adamretter/vbe-test

Still it's a mystery right now that I trying to understand, but much less code to look through now at least.

reinhapa:
@adamretter I wil try to take a look into this

adamretter:
Looks like a JIT issue with HotSpot JDKs, versions 12, 13, 14, and 15 are affected - https://github.com/adamretter/vbe-test#testing-results

adamretter:
[...]

PR #3545, opened 2020-09-29
Block startup for Java 12-15 to protect from data corruption

dizzzz:
Add warning and delay to indicate that openJDK12 and newer should not be used with eXist-db, for now.

Example output:
```
*****************************************************
eXist-db should only be started with Java8 or Java11!
OpenJDK versions 12 and newer contain at this moment a
severe bug in the JIT compiler that will cause data loss.
The problem has been reported and will be addressed in
future versions of the OpenJDK.
Detected version is 15.
*****************************************************
```


joewiz:
Great idea, Dannes! How about this version, with minor edits:

> Warning: eXist-db should only be used with Java 8 or 11! OpenJDK versions 12 and higher currently suffer from a critical bug in the JIT compiler that will cause data loss in eXist-db. The problem has been reported to the OpenJDK community. 
> 
> For more information, see:
>
> * https://bugs.openjdk.java.net/browse/JDK-8253191
> * https://github.com/eXist-db/exist/issues/3375
>
> The detected version of Java on your system is " + jreVersion + ". eXist-db will continue its startup sequence after a short delay, but until this OpenJDK bug is resolved, we urge you to switch to Java 8 or 11.

joewiz (approved):

dizzzz:
thnx, good to have native speakers onboard :-)

dizzzz:
updated to

image

dizzzz:
I decided not to abort the startup of the database, although there are probably good arguments to do so.

adamretter:
Couple of thoughts:

```
eXist-db should only be started with Java8 or Java11!
```

should read something like:
```
eXist-db should only be used with Java versions 8 through 11!
```

As Java 9, and 10 are also OK.

---

> I decided not to abort the startup of the database, although there are probably good arguments to do so.

You probably should do so, as otherwise corruption is inevitable. The worst scenario would be someone who upgrades to a newer eXist-db/Java and corrupts their otherwise good database files.
...Also if you can return an appropriate exit code if possible that would be great.

adamretter (commented):
I think you might need to place this code into some sort of common place, and call it from there.

Unfortunately there are multiple ways that eXist-db can be started, and several of them don't involve going via this start/Main class.

dizzzz:
I know that exist can be started from may ways, this part is the most common one I guess. I'd like to see that ppl use LTS versions only, java9/10 are really EOL for ages right now. A little push does not hurt.

dizzzz:
with a hard stop, I do agree that java9/10 should not be blocked

dizzzz:
@adamretter what locations would you indicate as 'entry' for starting eXist-db ? I think that only the common ones would be sufficient no?

dizzzz:
new attempt, now with not starting eXist-db

image

dizzzz:
for me, now it stops as it was intended as a simple change.

dizzzz:
Anyone ! @wolfgangmm @adamretter @joewiz  @ljo ?

dizzzz:
mac startup: via org.exist.start.Main:
image

dizzzz:
Deamon code via main.runEx():
image

dizzzz:
I updated the PR in such a way that both routes mentioned above + normal scripts will touch the OpenJDK version check.

adamretter:
@dizzzz As requested on the call, I took a look. So there are a couple of places where I think this code also needs to be called:

1. `InteractiveClient#main` - as the db can be started in embedded mode there. The warning/error should probably only be shown when launching in embedded mode, as the JDK issue doesn't effect the remote client.
2. `JettyStart#run(String[] args, Observer)` - this can also launch the db in embedded mode. It is used for tests, the exist-db-xqts-runner and various other apps which embed eXist-db.
3. ExistEmbeddedServer and ExistXmldbEmbeddedServer - (same reasons as (2)).
4. AbstractXMLDBTask
5. ExportMain and ExportGUI - as these can start the database in embedded mode.

It might also be worth doing some testing with the launcher. As I am not sure if it will show the error properly when your warning stops eXist-db from starting. If that is the case you would need to integrate your warning into LauncherWrapper/Launcher main but show it in a dialog message.

joewiz:
Correct me if I'm wrong, @dizzzz, but as discussed on today's Community Call, Dannes will revise this to check java is either (a) between 8 and 11, OR (b) 16+. And will remove exception in favor of just logging the warning. JW and PR think this is reasonable and will be an improvement, even if the warning isn't yet shown in all scenarios.

adamretter:
> And will remove exception in favor of just logging the warning.

IMHO we should keep the exception. I have seen that few users often check the logs lately, and will likely ignore warnings. This bug leads to certain database corruption and so we should try and help users not loose data here.

dizzzz:
I agree with @adamretter but as the PR is hanging for 7 months , I really 'd like to make a small step as I am very very concerned on the instability.

@wolfgangmm  offered to update the PR to make sure more startup scenarios would be covered.

dizzzz (commented):
lgtm

adamretter:
I pushed some commits to this branch. I think it should be in good shape now

wolfgangmm (approved):
Looks good to me :-)

PR #3550, opened 2020-10-02
[hotfix] add location info to element content errors

### LINKED ISSUE #3474: [BUG] missing location info in errors for functions inside element constructors (err:XQTY0105)
**Describe the bug**

In eXist, errors involving functions inside element constructors like `element foo { map { "x": "y" } }` lack location information.

**Expected behavior**

I expected the errors to report the location of the error in the query, as they do in BaseX and Saxon. (Saxon has the best error description of the 3 in this case.)

**To Reproduce**

```xquery
xquery version "3.1";

element foo { map { "x": "y" } }
```

eXist error:

> err:XQTY0105 Enclosed expression contains function item

BaseX returns the error at the start of the enclosed content:

> Stopped at /Users/joe/Downloads/file, 3/13:
> 
> [XQTY0105] Invalid content: map { "x": "y" }.

Saxon reports the error as appearing at the first character of the map constructor, with a very clear error description (if missing the error code):

> Cannot add a map as a child of a constructed element
> 
> Start location: 3:16

 - OS: macOS 10.13.6
 - Java Version: JDK 11.0.7+10 (Liberica 2020-04-14 LTS)

**Additional context**
- How is eXist-db installed? 5.3.0-SNAPSHOT built from source 
- Any custom changes in e.g. `conf.xml`? None.

line-o:
### Description:

The queries below are all invalid because the element content contains expressions which will evaluate to function types. The error is thrown a) at run-time and b) without location information (line and column number).

```xquery
element foo { sum(?) },
element foo { map{} },
```

This PR fixes the XQuery parser so that location information is given when an error is raised and also checks each expression if the return type is known to be a function type at compilation time.

- Add `path` to XPathException thrown in EnclosedExpr (for dynamic errors, works only if location is preserved during compilation)
- Add method `addPathIfNotFunction` to SequenceConstructor to add a static type check for each path before it is added. Throws if a function type (map, array, function, partial ...) is found.
- fix `compElemConstructor` rule in XQuery.g by using existing rule `compConstructorValue`. This adds location info to element content sub-expressions
- use new SequenceConstructor method in XQueryTree.g to add paths to EnclosedExpr content and thus statically check return types of subexpressions

There is an exception when it comes to function types: Arrays are serialised when they appear in element content:

```xquery
element foo { [1] }
(: this must evaluate to <test>1</test> :)
```

### Reference:

Fixes #3474 

[Element Content Rules](https://www.w3.org/TR/xquery-31/#id-content)
As @adamretter pointed out 1.e(i) and 1.e(ii) define the behaviour: arrays are flattened but any other function type throws. 

When the function is returned at runtime Saxon10-HE throws a different error code [XTDE0450](https://www.w3.org/TR/xslt-30/#err-XTDE0450).
We should think about mimicking this behaviour or if we stick to XQTY0105 for static and dynamic errors.

### Type of tests:

adds jUnit testsuite and modifies one XQsuite test

### DIFF
--- a/exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
+++ b/exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
@@ -42,35 +42,35 @@ public class FunctionTypeInElementContent {
     @Test
     public void arrayLiteral() throws XMLDBException {
         final String query = "element test { [] }";
-        final String error = "err:XQTY0105 Function types are not allowed in element content. Got array(*) [at line 1, column 14, source: element test { [] }]";
-        assertCompilationError(query, error);
+        assertCompilationSuccess(query);
     }
+
+    @Test
+    public void arrayConstructor() throws XMLDBException {
+        final String query = "element test { array { () } }";
+        assertCompilationSuccess(query);
+    }
+
 
     @Test
     public void partialBuiltIn() throws XMLDBException {
         final String query = "element test { sum(?) }";
-        final String error = "err:XQTY0105 Function types are not allowed in element content. Got function(*) [at line 1, column 14, source: element test { sum(?) }]";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got function(*) [at line 1, column 16, source: element test { sum(?) }]";
         assertCompilationError(query, error);
     }
 
     @Test
     public void userDefinedFunction() throws XMLDBException {
         final String query = "element test { function () { () } }";
-        final String error = "err:XQTY0105 Function types are not allowed in element content. Got function(*) [at line 1, column 14, source: element test { function () { () } }]";
-        assertCompilationError(query, error);
-    }
-
-    @Test
-    public void arrayConstructor() throws XMLDBException {
-        final String query = "element test { array { () } }";
-        final String error = "err:XQTY0105 Function types are not allowed in element content. Got array(*) [at line 1, column 14, source: element test { array { () } }]";
+        // TODO: user defined function has its location offset to a weird location
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got function(*) [at line 1, column 25, source: element test { function () { () } }]";
         assertCompilationError(query, error);
     }
 
     @Test
     public void mapConstructor() throws XMLDBException {
         final String query = "element test { map {} }";
-        final String error = "err:XQTY0105 Function types are not allowed in element content. Got map(*) [at line 1, column 14, source: element test { map {} }]";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got map(*) [at line 1, column 16, source: element test { map {} }]";
         assertCompilationError(query, error);
     }
 
@@ -106,5 +106,11 @@ public class FunctionTypeInElementContent {
         } catch (XMLDBException ex) {
             assertEquals( error, ex.getMessage() );
         }
+    }
+    private void assertCompilationSuccess(final String query) throws XMLDBException {
+        final XQueryService service = (XQueryService)existEmbeddedServer.getRoot().getService("XQueryService", "1.0");
+
+        service.compile(query);
+        assertTrue( true );
     }
 }
--- a/exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
+++ b/exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
@@ -1,0 +1,110 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery;
+
+import org.exist.test.ExistXmldbEmbeddedServer;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.xmldb.api.base.XMLDBException;
+import org.xmldb.api.modules.XQueryService;
+
+import static org.junit.Assert.*;
+
+/**
+ * Ensure function types returned in element content throws at compile time and has location information
+ * https://github.com/eXist-db/exist/issues/3474
+ *
+ * @author <a href="mailto:juri@existsolutions.com">Juri Leino</a>
+ */
+public class FunctionTypeInElementContent {
+    @ClassRule
+    public static final ExistXmldbEmbeddedServer existEmbeddedServer = new ExistXmldbEmbeddedServer(true, true, true);
+
+    @Test
+    public void arrayLiteral() throws XMLDBException {
+        final String query = "element test { [] }";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got array(*) [at line 1, column 14, source: element test { [] }]";
+        assertCompilationError(query, error);
+    }
+
+    @Test
+    public void partialBuiltIn() throws XMLDBException {
+        final String query = "element test { sum(?) }";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got function(*) [at line 1, column 14, source: element test { sum(?) }]";
+        assertCompilationError(query, error);
+    }
+
+    @Test
+    public void userDefinedFunction() throws XMLDBException {
+        final String query = "element test { function () { () } }";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got function(*) [at line 1, column 14, source: element test { function () { () } }]";
+        assertCompilationError(query, error);
+    }
+
+    @Test
+    public void arrayConstructor() throws XMLDBException {
+        final String query = "element test { array { () } }";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got array(*) [at line 1, column 14, source: element test { array { () } }]";
+        assertCompilationError(query, error);
+    }
+
+    @Test
+    public void mapConstructor() throws XMLDBException {
+        final String query = "element test { map {} }";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got map(*) [at line 1, column 14, source: element test { map {} }]";
+        assertCompilationError(query, error);
+    }
+
+//  -- no error is thrown at compile time
+//    @Test
+//    public void mapConstructorInSequence() throws XMLDBException {
+//        final String query = "element test {\n" +
+//                "    \"a\",\n" +
+//                "    map {}\n" +
+//                "}";
+//        final String error = "err:XQTY0105 Function types are not allowed in element content. Got map(*) [at line 1, column 14, source: element test { map {} }]";
+//        assertCompilationError(query, error);
+//    }
+
+
+//  -- no error is thrown at compile time nor run time
+//    @Test
+//    public void arrayInSequence() throws XMLDBException {
+//        final String query = "element test {\n" +
+//                "    \"a\",\n" +
+//                "    []\n" +
+//                "}";
+//        final String error = "err:XQTY0105 Function types are not allowed in element content. Got map(*) [at line 1, column 14, source: element test { map {} }]";
+//        assertCompilationError(query, error);
+//    }
+
+    private void assertCompilationError(final String query, final String error) throws XMLDBException {
+        final XQueryService service = (XQueryService)existEmbeddedServer.getRoot().getService("XQueryService", "1.0");
+
+        try {
+            service.compile(query);
+            fail("no XMLDBException was thrown during compilation.");
+        } catch (XMLDBException ex) {
+            assertEquals( error, ex.getMessage() );
+        }
+    }
+}
--- a/exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
+++ b/exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
@@ -23,6 +23,7 @@ package org.exist.xquery;
 
 import org.exist.test.ExistXmldbEmbeddedServer;
 import org.junit.ClassRule;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.xmldb.api.base.XMLDBException;
 import org.xmldb.api.modules.XQueryService;
@@ -45,12 +46,20 @@ public class FunctionTypeInElementContent {
         assertCompilationSuccess(query);
     }
 
+    /**
+     * TODO: remove empty sequence after https://github.com/eXist-db/exist/issues/3472 is fixed
+     */
     @Test
     public void arrayConstructor() throws XMLDBException {
         final String query = "element test { array { () } }";
         assertCompilationSuccess(query);
     }
 
+    @Test
+    public void sequenceOfItems() throws XMLDBException {
+        final String query = "element test { (1, map {})[1] }";
+        assertCompilationSuccess(query);
+    }
 
     @Test
     public void partialBuiltIn() throws XMLDBException {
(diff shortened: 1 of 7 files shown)


line-o:
Tests failed so there might be unforeseen side-effects 😿 

but this is pure gold ⛳ : 

```
(: https://github.com/eXist-db/exist/issues/3474 :)
declare
    %test:pending("to be addressed in another PR")
    %test:assertEquals(333)
function et:issue3474() {
    try {
        element foo { map { "x": "y" } }
    } catch * {
        $err:line-number
    }
};
```

fails at compile time with 

> org.exist.xquery.XPathException: err:XQTY0105 Error while loading module /home/travis/build/line-o/exist/exist-core/src/test/xquery/errors.xql: Function types are not allowed in element content. Got map(*) [at line 334, column 21]

and 334 is correct, I checked

line-o:
So the issue here is:
This was possible but now fails with an AST error
```xquery
element a {}
```
While this parses fine and has all the locations in place
```xquery
element a { () }
```

Which is interestingly similar to functions

```xquery
function () {}
```
versus
```xquery
function () { () }
```

line-o:
### Discussion 

#### Arrays in element content

```xquery
element test { [1, 2] }
(: evaluates to <test>1 2</test> in current version of exist :)
```
Arrays seem to be fine in element content and get flattened. I have to check the other processors and the spec, if that is the way It is supposed to work.

#### Element content sub-expressions

```xquery
element test {
 "a",
  map {}
}
(: throws at runtime with line 1 and column 14 :)
```

The location is now tied to the opening curly brace. While this is an improvement it is definitely not ideal. Will try to get the subexpression location in as well. Especially helpful when element content constructors is tens of lines long.
On top of that, if the function type is part of a sequence, the error is not thrown at compile time. I think both issues are somewhat related to the way the element content expression is parsed.

line-o:
Saxon does serialise arrays, so that needs to be allowed.

plus these two edge cases:

```
➜ java -cp saxon-he-10.0.jar net.sf.saxon.Query -qs:"element test { ('a', map {})[1] }"

<?xml version="1.0" encoding="UTF-8"?><test>a</test>%
~/projects/saxon-he-10
```
```
➜ java -cp saxon-he-10.0.jar net.sf.saxon.Query -qs:"element test { ('a', map {})[2] }"

Error on line 1 column 17 of file:/Users/jll/projects/saxon-he-10/:
  XTDE0450  Cannot add a map to an XDM node tree
<?xml version="1.0" encoding="UTF-8"?>Query failed with dynamic error: Cannot add a map to an XDM node tree
```

joewiz:
@line-o This remains a WIP, correct?

line-o:
@joewiz The major issue - allow arrays in element content, is fixed.
We could
a) merge this PR and create a new issue for the tests that are currently pending. Specifically, for sub-expressions the location is still missing (for `element a { "a", map {} }` the location is still swallowed)
b) keep this PR open until sub-expressions do report the location as well

joewiz:
@line-o I could see the wisdom in both options, but option B seems better, since it would get the major fix into the develop branch and make its inclusion certain for the next release.

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * Ensure function types returned in element content throws at compile time and has location information
+ * https://github.com/eXist-db/exist/issues/3474
+ *
+ * @author <a href="mailto:juri@existsolutions.com">Juri Leino</a>
+ */
+public class FunctionTypeInElementContent {
```
dizzzz, line 39: By convention, shouldn't the name end with ..Test?

dizzzz:
Nice tests, I asked @ljo to review... as i am no expert on these .g files

joewiz:
@line-o Sorry, I think I misspoke - option A seems better for the reason I stated, rather than option B, since Option A would get the major fix into the develop branch and make its inclusion certain for the next release. Do you agree that Option A is the route to go? If so, I think removing the WIP will indicate to reviewers that you believe the PR is ready for review and hopefully approval?

line-o:
@joewiz Hah, good that I was doing some housekeeping! Yes, I am all for it.
I think we should already open the follow up issue.

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
adamretter: This need to renamed as @dizzzz indicated

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+        final String query = "element test { [] }";
+        assertCompilationSuccess(query);
+    }
+
+    /**
+     * TODO: remove empty sequence after https://github.com/eXist-db/exist/issues/3472 is fixed
+     */
+    @Test
+    public void arrayConstructor() throws XMLDBException {
+        final String query = "element test { array { () } }";
```
adamretter, line 54: Why not just use the other constructor for now, e.g. `[]`

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+    public static final ExistXmldbEmbeddedServer existEmbeddedServer = new ExistXmldbEmbeddedServer(true, true, true);
+
+    @Test
+    public void arrayLiteral() throws XMLDBException {
+        final String query = "element test { [] }";
+        assertCompilationSuccess(query);
+    }
+
+    /**
+     * TODO: remove empty sequence after https://github.com/eXist-db/exist/issues/3472 is fixed
```
adamretter, line 50: TODO comments in Java should not be javadoc syntax.

I would suggest instead to use the syntax:
```java
//TODO(JL) remove empty sequence after https://github.com/eXist-db/exist/issues/3472 is fixed
```

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Test
+    public void partialBuiltIn() throws XMLDBException {
+        final String query = "element test { sum(?) }";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got function(*) [at line 1, column 16, source: element test { sum(?) }]";
+        assertCompilationError(query, error);
+    }
+
+    /**
+     * TODO: Investigate does still throw without location info
```
adamretter, line 72: TODO comments should not be javadoc syntax.

What is the impact of this test which is ignored? Do you still mean to fix it in this PR?

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+        final String query = "element test { sum#0 }";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got function(*) [at line 1, column 16, source: element test { sum#0 }]";
+        assertCompilationError(query, error);
+    }
+
+    /**
+     * Does not throw at compile time
+     */
+    @Ignore
+    @Test
```
adamretter, line 86: What is the impact of this test which is ignored? Do you still mean to fix it in this PR?

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+    public void functionVariable() throws XMLDBException {
+        final String query = "let $f := function () { () } return element test { $f }";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got function(*) [at line 1, column 16, source: element test { sum(?) }]";
+        assertCompilationError(query, error);
+    }
+
+    /**
+     * TODO: remove empty sequence after https://github.com/eXist-db/exist/issues/3551 is fixed
```
adamretter, line 94: TODO comments should not be javadoc syntax.

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    @Test
+    public void userDefinedFunction() throws XMLDBException {
+        final String query = "element test { function () { () } }";
+        // TODO: user defined function has its location offset to a weird location
```
adamretter, line 99: What is the status/imapct of this TODO? do you mean to fix it in this PR?

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+        final String query = "element test { (map {}) }";
+        final String error = "An exception occurred during query execution: err:XQTY0105 Function types are not allowed in element content. Got map(*) [source: element foo { (map{}) }]";
+        assertCompilationError(query, error);
+    }
+
+    /**
+     * there is an edge case which would evaluate to empty sequence
+     * but should arguably still throw
+     * Does still throw without location info
+     * TODO: add (sub-expression) location
```
adamretter, line 126: What is the status/imapct of this TODO? do you mean to fix it in this PR?

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+    @Test
+    public void sequenceOfMapsEdgeCase() throws XMLDBException {
+        final String query = "element test { (map {})[2] }";
+        final String error = "err:XQTY0105 Function types are not allowed in element content. Got map(*) [source: element test { (map {})[2] }]";
+        assertCompilationError(query, error);
+    }
+
+    /**
+     * -- no error is thrown at compile time
+     * TODO: add (sub-expression) location
```
adamretter, line 137: What is the status/imapct of this TODO? do you mean to fix it in this PR?

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+        final String query = "element test { [map {}] }";
+        final String error = "An exception occurred during query execution: err:XQTY0105 Function types are not allowed in element content. Got map(*) [source: element test { [map{}] }]";
+        assertCompilationError(query, error);
+    };
+
+    /**
+     * This could throw at compile time, but does not
+     * TODO: add (sub-expression) location
+     */
+    @Ignore
```
adamretter, line 150: What is the status/impact of this ignored test and TODO? do you mean to fix it in this PR?

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
```diff
@@ -0,0 +1,174 @@
...
+            fail("no XMLDBException was thrown during compilation.");
+        } catch (XMLDBException ex) {
+            assertEquals( error, ex.getMessage() );
+        }
+    }
+    private void assertCompilationSuccess(final String query) throws XMLDBException {
+        final XQueryService service = (XQueryService)existEmbeddedServer.getRoot().getService("XQueryService", "1.0");
+
+        service.compile(query);
+        assertTrue( true );
```
adamretter, line 172: This line does nothing! You probably wanted to check that the result of `service.compile` is non-null instead.

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
adamretter: Also looks like https://github.com/eXist-db/exist/issues/3551 has now been merged. So I think this needs updating?

exist-core/src/test/xquery/errors.xql
```diff
@@ -325,14 +325,50 @@ function et:issue3473c() {
...
+        exists($err:line-number) and
+        $err:line-number > 0 and
+        exists($err:column-number) and
+        $err:column-number > 0
+    }
+};
+
+declare
+    %test:assertTrue
+    %test:pending("location info still missing")
```
adamretter, line 349: Should this still be pending?

exist-core/src/test/xquery/errors.xql
```diff
@@ -325,14 +325,50 @@ function et:issue3473c() {
...
+ : function type item returned in element content in sequence at runtime
+ : related to https://github.com/eXist-db/exist/issues/3474
+ :)
 declare
-    %test:pending("to be addressed in another PR")
-    %test:assertEquals(333)
-function et:issue3474() {
+    %test:assertTrue
+    %test:pending("location info still missing")
+function et:subexpression-in-enclosed-expression-evaluates-to-map() {
```
adamretter, line 335: Should this still be pending?

adamretter (changes requested):
Some small changes please

adamretter:
@line-o When you say:

> and 334 is correct, I checked

That is likely because XQSuite uses `util:eval`, I guess we should modify XQSuite to use the new `pass` parameter I provided for `util:eval` so that you get the correct line numbers of the query being evaluated, not the entire outer and inner query combined.

adamretter:
> We should think about mimicking this behaviour or if we stick to XQTY0105 for static and dynamic errors.

I just checked and BaseX returns `XQTY0105` here too. As I interpret the spec, we should be okay with `XQTY0105`

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
line-o: I will add a test with a array literal constructor, good point.

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
line-o: No, it is out of scope for this PR as discussed here.

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
line-o: see above

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
line-o: see above

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
line-o: see above

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
line-o: It asserts that no compile error was thrown. I can change that to a non-null check, if needed. What is the method name?
`assertNonNull(compilationResult)`?

exist-core/src/test/xquery/errors.xql
line-o: yes

exist-core/src/test/xquery/errors.xql
line-o: yes

line-o:
@adamretter the wrong and or missing location info in those tests has nothing to do with `util:eval`.

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
line-o: OK

line-o:
@adamretter can you show me, where util:eval is used to call the XQsuite tests? I would expect the module to be imported and the exported functions to be called one by one.

adamretter:
@line-o I see that it is only used in `test:assertXPath` now, however it should probably be fixed there to `pass` the location info.

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
adamretter: It doesn't assert that there was no error... an exception not being thrown and returning void is enough for that ;-)

You can use `assertNotNull`

exist-core/src/test/java/org/exist/xquery/FunctionTypeInElementContent.java
line-o: There was one already! Both cases need to be tested.

line-o:
#3623 is the follow up issue tracking the cases not covered here.

joewiz:
@line-o xqsuite's use of util:eval can be found here:

https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql#L883

By setting the 4th parameter to `true()`, a query like:

```xquery
util:eval("1,
2,
<3>", false(), (), true())
```

will report the error on line 3 (where the evaluated query fails) instead of on line 1 (where the query calls util:eval).

line-o:
waits on #3605 to be merged

line-o:
Rebased on current HEAD of develop. Java tests adapted to extend XQueryCompilationTest and use XQueryAssertions.
Should be good to go now.

line-o:
@adamretter you might want to approve this PR now, or remove your request for changes perhaps?

adamretter (approved):
This is now in good shape.

PR #3552, opened 2020-10-05
[fix] allow functions with empty body

### LINKED ISSUE #3551: [BUG] allow empty function body in xquery
**Describe the bug**

XQuery allows a function body to be empty. This is equivalent of a function body returning an empty sequence.

Evaluating the following line

```xquery
function () {}
```
yields a compilation error: `Syntax error within user defined function: unexpected token: }` (also: location information is missing)

```xquery
(function () {})()
```

Parses but throws when executed with `Cannot compile xquery: err:XPST0003 Syntax error within user defined function: unexpected token: } [at line 1, column 15]`

**Expected behavior**

The above to evaluate to a function with arity 0 returning an empty sequence.

Saxon10-HE handles this correctly:

```
➜ java -cp saxon-he-10.0.jar net.sf.saxon.Query -qs:"(function () {})()"
<?xml version="1.0" encoding="UTF-8"?>%
```

**To Reproduce**

```Xquery
xquery version "3.1";

declare
    %test:assertTrue
function t:test() {
  let $f := function () { () }
  return $f instance of function(*)
};

(: this will not compile :)
declare
    %test:assertTrue
function t:test() {
  let $f := function () {}
  return $f instance of function(*)
};

(: this will not run :)
declare
    %test:assertFalse
function t:test() {
  exists((function () {})())
};
```

 - OS: 10.15.1
 - eXist-db version: 5.3.0-SNAPSHOT commit: 587dde5f72791201512956125d4d8d0f6cc0a319
 - Java Version AdoptOpen JDK 8

**Additional context**

line-o:
fixes #3551

- modify grammar rule functionBody to handle an empty function body
- add XQSuite tests to ensure this works for
  - declared functions
  - declared variables of type function(*)
  - inline declarations
  - function expressions

### DIFF
--- a/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
+++ b/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
@@ -511,8 +511,15 @@ functionDecl [XQueryAST ann] throws XPathException
 	}
 	;
 
-functionBody throws XPathException:
-	LCURLY^ e:expr RCURLY! ;
+// add virtual parenthesizedExpr for empty function bodys
+// this adds an empty sequence
+// fixes #3551
+functionBody throws XPathException
+:
+    ( LCURLY RCURLY ) => l:LCURLY^ RCURLY!
+    { #functionBody= #(#l, #(#[PARENTHESIZED, "Parenthesized"], null)); }
+    | LCURLY^ expr RCURLY!
+    ;
 
 returnType throws XPathException:
 	"as"^ sequenceType ;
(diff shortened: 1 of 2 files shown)


dizzzz (approved):

joewiz:
Could someone restart the failed Appveyor CI check? (I used to have permission when we used Adam's or Duncan's account, but ever since the last change, I no longer do... If whoever does have permission could grant it to me again, I would be able to restart CI checks like this when I notice them.)

line-o:
I can't, just checked.

adamretter:
@joewiz @line-o I don't have permission either.

dizzzz:
restarted

PR #3567, opened 2020-10-17
Proposed fix for NPE with empty case expression. Fixes 3560

### LINKED ISSUE #3560: [BUG] NPE on case (in a switch statement) resolving to empty-sequence()

**Describe the bug**
NPE on `case` (in a `switch` statement) resolving to `empty-sequence()`

**Expected behavior**
No NPE, no error, same behavior as with `case ()`; 

**To Reproduce**

    let $a := 'a'
    let $test := ('a','b','c')
    let $test-two := ('d','e')
    return switch( $a)
      case $test[. eq $a] return 'one'
      case $test-two[. eq $a] return 'two'
      default return ()

returns correct result 'one'; with 
`let $a := 'd'`
one will get an NPE presumably because the parser does not like 
`$test[. eq $a]` resolving to an empty sequence.

note that `case () return 'one'` does not throw an error of any kind (although it does not match with $a being empty either).

 - OS: windows/linux
 - eXist-db version: 5.3-snapshot, 4.7.1
 - Java Version [e.g. Java8u261]

dizzzz:
Proposed fix for NPE with empty case expression.
Fixes #3560

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/SwitchExpression.java
+++ b/exist-core/src/main/java/org/exist/xquery/SwitchExpression.java
@@ -104,7 +104,7 @@ public class SwitchExpression extends AbstractExpression {
 	                if (caseSeq.hasMany()) {
 	                    throw new XPathException(this, ErrorCodes.XPTY0004, "Cardinality error in switch case operand ", caseSeq);
 	                }
-                    final AtomicValue caseVal = caseSeq.itemAt(0).atomize();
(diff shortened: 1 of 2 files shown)


wolfgangmm (approved):
Seems reasonable to me. For sure it's possible that a case clause evaluates to the empty sequence, so this should be handled.

adamretter (approved):
@dizzzz Looks good to me :-)

PR #3570, opened 2020-10-18
Fix QName comparison Namepool.

dizzzz:
Fix where QName comparison in Namepool was not taking the actual namespace into account

Found during code inspection with Intellij.


adamretter (approved):
Nice catch!

PR #3580, opened 2020-10-25
Feature/more improvements

dizzzz:
Another round of sunday-eve code improvements. Grouped per theme.

### DIFF
--- a/exist-core/src/main/java/org/exist/protocolhandler/xmldb/XmldbURL.java
+++ b/exist-core/src/main/java/org/exist/protocolhandler/xmldb/XmldbURL.java
@@ -280,7 +280,7 @@ public class XmldbURL {
      */
     public String getHost(){
         final String hostname=myUrl.getHost();
-        if("".equals(hostname)){
+        if(hostname != null && hostname.isEmpty()){
             return null;
         } else {
             return hostname;
--- a/exist-core/src/main/java/org/exist/xmldb/FullXmldbURI.java
+++ b/exist-core/src/main/java/org/exist/xmldb/FullXmldbURI.java
@@ -188,7 +188,7 @@ public class FullXmldbURI extends XmldbURI {
                     context = context.substring(0, context.length() - 1);
                 }
             }
-            this.context = "".equals(context) ? null : context;
+            this.context = context != null && context.isEmpty() ? null : context;
             recomputeURI();
 
         } catch (final URISyntaxException e) {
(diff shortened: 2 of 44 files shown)


exist-core/src/main/java/org/exist/protocolhandler/xmldb/XmldbURL.java
```diff
@@ -280,7 +280,7 @@ public String getProtocol(){
      */
     public String getHost(){
         final String hostname=myUrl.getHost();
-        if("".equals(hostname)){
+        if(hostname != null && hostname.isEmpty()){
```
adamretter, line 283: This isn't the same logic as before!

exist-core/src/main/java/org/exist/xmldb/FullXmldbURI.java
```diff
@@ -188,7 +188,7 @@ private void setContext(String context) {
                     context = context.substring(0, context.length() - 1);
                 }
             }
-            this.context = "".equals(context) ? null : context;
+            this.context = context != null && context.isEmpty() ? null : context;
```
adamretter, line 191: This is not the same logic as before.

adamretter (commented):
I am not sure about this pattern of changes:

Before -
```
return groups.toArray(new String[groups.size()]);
```

After:
```
return groups.toArray(new String[0]);
```

As I understand it, after your change two arrays have two be allocated (and then the first is discarded). Whereas before your change only a single array has to be allocated, so I would think that is more efficient.

----

dizzzz:
@adamretter yes I understand: this is the rationale as provided by intellij:

"There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).

In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.

This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs)."

adamretter:
@dizzzz Hmmmm. I would be surprised if they are mistaken. However reading the code of the JDK doesn't make it seem that way at all.

dizzzz:
Hmmmmm thnx for checking. I assumed indeed that these guys know what is happening under the hood. Maybe it is a compiler optimization step ?

adamretter:
@dizzzz Indeed it could be.

PR #3588, opened 2020-10-28
[feature] Add a  parameter to util:eval to preserve the original erro…

adamretter:
This back-ports a feature to eXist-db from FusionDB. The feature adds a `$pass` parameter to `util:eval`. This is similar to the BaseX option, it preserves the line and column number of any error within the evaluated string, rather than setting it to that of the `util:eval` expression.

Closes https://github.com/eXist-db/exist/issues/3513


dizzzz (approved):

dizzzz (changes requested):
License check fails :-/

image

adamretter:
@dizzzz Should be all sorted now.

dizzzz (approved):

PR #3596, opened 2020-10-30
Implementation of array:sort

adamretter:
Back-ported from FusionDB.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/value/AtomicValueComparator.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/AtomicValueComparator.java
@@ -33,50 +33,48 @@
 package org.exist.xquery.value;
 
 import com.ibm.icu.text.Collator;
-import net.jcip.annotations.NotThreadSafe;
-import org.exist.xquery.Constants;
+import net.jcip.annotations.ThreadSafe;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.exist.http.RESTServer;
+import org.exist.xquery.XPathException;
 
 import javax.annotation.Nullable;
 import java.util.Comparator;
 
 /**
- * Comparator for comparing instances of Item
- * apart from the XQuery atomic types there are
- * two Node types in eXist org.exist.dom.persistent.*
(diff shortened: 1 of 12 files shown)


exist-core/src/main/java/org/exist/xquery/value/AtomicValueComparator.java
```diff
@@ -0,0 +1,80 @@
...
+            return -1;
+        } else if (o2 == null) {
+            return 1;
+        }
+
+        try {
+            return o1.compareTo(collator, o2);
+        } catch (final XPathException e) {
+            LOG.error(e.getMessage(), e);
+            throw new ClassCastException(e.getMessage());
```
dizzzz, line 77: ??

exist-core/src/test/xquery/arrays/arrays.xql
```diff
@@ -1009,3 +1009,27 @@ declare
 function arr:put_pos_illegal_index() {
     array:put(["a", "b", "c", "d"], 5, "x")?*
 };
+
+declare
+    %test:assertEquals(1, 3, 4, 5, 6)
+function arr:sort-int() {
```
dizzzz, line 1015: what about a test doing 'mixed' content? ints, strings, decimals

dizzzz (approved):

exist-core/src/main/java/org/exist/xquery/value/AtomicValueComparator.java
adamretter: @dizzzz What other option do we have? The API says we can throw either:

```
     * @throws NullPointerException if an argument is null and this
     *         comparator does not permit null arguments
     * @throws ClassCastException if the arguments' types prevent them from
     *         being compared by this comparator.
```

So `ClassCastException` seemed like the *most* correct option to me.

exist-core/src/test/xquery/arrays/arrays.xql
adamretter: Done.

joewiz:
@adamretter Thank you for helping eXist move toward completing eXist's coverage of the XQuery 3.1 spec!

PR #3546, opened 2020-09-30
line and column numbers in unknown-atomic-type errors

line-o:
### Description:

Type errors for types after `instance of`, `cast as`, `castable as` and similar expressions do not show line and column numbers.

add line and column for type errors after

- [x]  `cast as` 
- [x] `castable as` 
- [x] `treat as` 
- [x] `instance of` 
- [x] tests (static compilation errors need to be tested in java)
### Reference:

fixes #3518 

### Type of tests:

jUnit tests were added 

You can also evaluate any of the following expressions in eXide to see the change

```
let $x as y := 0 return $x
```

```
function ($x as y) { $x }
```

```
<x /> instance of y
```

```
<x /> cast as
y
```

```
<x /> castable as y
```

```
<x />
treat as
y
```

PR #3605, opened 2020-11-03
Fix root node handling and implicit context in functions 

line-o:
### Description:

This PR includes and builds on #3167 and #3163
In order to make absolute XPath expressions using an implicit context in function bodies impossible the XQuery parser
is made aware of being in a function body. A static error is thrown when this happens. 
It is similar to the parser's handling of attribute and string constructor contents.
According to @adamretter this will bring us another 4% towards full compliance with the XQuery 3.1 spec.

### Reference:

According to https://www.w3.org/TR/xquery-31/#id-eval-function-call when evaluating the body of a function:

The focus (context item, context position, and context size) is absent

### Type of tests:

All XQSuite tests that used implicit context expressions are adapted.
A new set of Java test classes that specifically test for the error was created by me


line-o (commented):
ContextItemTests have no license header. How to proceed @adamretter

line-o:
I rewrote the testsuite - so no license discussions necessary.

exist-core/src/test/xquery/xquery3/bang.xql
```diff
@@ -163,11 +163,16 @@ function bang:constructor() {
     $bang:works/employee[pnum = "P4"] ! <name>{@name/string()}</name>
 };
 
+(:~
+ : implicit context cannot be tested in a xquery testsuite
+ : because it will raise a static error now
```
dizzzz, line 168: is there a ticket to trace this?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Query.java
```diff
@@ -197,16 +197,17 @@ public int getOptimizeAxis() {
     }
 
     public NodeSet preSelect(Sequence contextSequence, boolean useContext) throws XPathException {
-    	if (contextSequence != null && !contextSequence.isPersistentSet())
+        // DW: contextSequence can be null
```
dizzzz, line 200: my name can be removed :-)

dizzzz (approved):
LGTM ... nice tests, but I am not a lexer guy

exist-core/src/test/xquery/xquery3/bang.xql
line-o: I do not understand. What should be traced?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Query.java
line-o: Well, it was your comment that led me to the solution :)

ljo (approved):
Looks good from what I can tell.

line-o:
Please test if 
```
function ($node) {
  $node ! //a
}
```

line-o:
saxon evaluates
```xquery
(function ($node) { $node ! //a })(document {<root><a>1</a><b><a>3</a></b></root>})
```
to `<a>1</a><a>3</a>`

line-o:
If you pass in anything other than a document Saxon throws
```
Error on line 1 column 29 of file:/Users/jll/projects/saxon-he-10/:
  XPDY0050  The root of the tree containing the context item is not a document node
     invoked by unknown caller (class net.sf.saxon.value.MemoClosure)
     invoked by unknown caller (class net.sf.saxon.functions.hof.UserFunctionReference$BoundUserFunction)
<?xml version="1.0" encoding="UTF-8"?>Query failed with dynamic error: The root of the tree containing the context item is not a document node
```

joewiz:
@line-o Same results in BaseX. Its error for the scenario without the document node is:

> `[XPDY0050] Value has no document node: <root>...</root>`

joewiz:
(But changing `//` to `.//` causes Saxon and BaseX to return `<a>1</a><a>3</a>` even without with the document node.)

line-o:
@joewiz I do not expect `function ($n) { $n ! .//a }` to have an issue. 
But `function ($n) { $n ! //a }` will likely throw with a static error with this PR applied.

joewiz:
@line-o I just checked out this PR and built eXist, and indeed your test query with `document-node` returns this error:

> `Cannot compile xquery: err:XPDY0002 Leading '//' selects nothing, ContextItem is absent in function body [at line 1, column 29]`

The stack trace from exist.log:

```
Caused by: org.exist.xquery.XPathException: err:XPDY0002 Leading '//' selects nothing, ContextItem is absent in function body [at line 1, column 29]
	at org.exist.xquery.parser.XQueryParser.pathExpr(XQueryParser.java:7840) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.valueExpr(XQueryParser.java:7657) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.unaryExpr(XQueryParser.java:7502) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.arrowExpr(XQueryParser.java:6961) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.castExpr(XQueryParser.java:6867) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.castableExpr(XQueryParser.java:6774) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.treatExpr(XQueryParser.java:6682) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.instanceofExpr(XQueryParser.java:6591) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.intersectExceptExpr(XQueryParser.java:7998) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.unionExpr(XQueryParser.java:7228) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.multiplicativeExpr(XQueryParser.java:7163) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.additiveExpr(XQueryParser.java:7114) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.rangeExpr(XQueryParser.java:7035) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.stringConcatExpr(XQueryParser.java:6994) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.comparisonExpr(XQueryParser.java:6323) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.andExpr(XQueryParser.java:6292) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.orExpr(XQueryParser.java:4806) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.exprSingle(XQueryParser.java:4479) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.expr(XQueryParser.java:456) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.functionBody(XQueryParser.java:3047) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.inlineFunctionExpr(XQueryParser.java:11523) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.functionItemExpr(XQueryParser.java:11059) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.primaryExpr(XQueryParser.java:9568) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.postfixExpr(XQueryParser.java:8553) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.stepExpr(XQueryParser.java:8515) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.relativePathExpr(XQueryParser.java:8047) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.pathExpr(XQueryParser.java:7822) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.valueExpr(XQueryParser.java:7647) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.unaryExpr(XQueryParser.java:7502) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.arrowExpr(XQueryParser.java:6961) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.castExpr(XQueryParser.java:6867) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.castableExpr(XQueryParser.java:6774) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.treatExpr(XQueryParser.java:6682) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.instanceofExpr(XQueryParser.java:6591) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.intersectExceptExpr(XQueryParser.java:7998) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.unionExpr(XQueryParser.java:7228) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.multiplicativeExpr(XQueryParser.java:7163) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.additiveExpr(XQueryParser.java:7114) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.rangeExpr(XQueryParser.java:7035) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.stringConcatExpr(XQueryParser.java:6994) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.comparisonExpr(XQueryParser.java:6323) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.andExpr(XQueryParser.java:6292) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.orExpr(XQueryParser.java:4806) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.exprSingle(XQueryParser.java:4479) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.expr(XQueryParser.java:456) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.parenthesizedExpr(XQueryParser.java:10288) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.primaryExpr(XQueryParser.java:9289) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.postfixExpr(XQueryParser.java:8553) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.stepExpr(XQueryParser.java:8515) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.relativePathExpr(XQueryParser.java:8047) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.pathExpr(XQueryParser.java:7822) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.valueExpr(XQueryParser.java:7647) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.unaryExpr(XQueryParser.java:7502) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.arrowExpr(XQueryParser.java:6961) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.castExpr(XQueryParser.java:6867) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.castableExpr(XQueryParser.java:6774) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.treatExpr(XQueryParser.java:6682) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.instanceofExpr(XQueryParser.java:6591) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.intersectExceptExpr(XQueryParser.java:7998) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.unionExpr(XQueryParser.java:7228) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.multiplicativeExpr(XQueryParser.java:7163) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.additiveExpr(XQueryParser.java:7114) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.rangeExpr(XQueryParser.java:7035) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.stringConcatExpr(XQueryParser.java:6994) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.comparisonExpr(XQueryParser.java:6323) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.andExpr(XQueryParser.java:6292) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.orExpr(XQueryParser.java:4806) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.exprSingle(XQueryParser.java:4479) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.expr(XQueryParser.java:456) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.queryBody(XQueryParser.java:1067) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.mainModule(XQueryParser.java:812) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.module(XQueryParser.java:593) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.parser.XQueryParser.xpath(XQueryParser.java:500) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.XQuery.compile(XQuery.java:110) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.XQuery.compile(XQuery.java:79) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.XQuery.compile(XQuery.java:71) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.http.servlets.XQueryServlet.process(XQueryServlet.java:448) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	... 66 more
```

exist-core/src/test/xquery/xquery3/bang.xql
adamretter: I don't think this should be pending. If it is invalid it should be removed, or if it has a purpose it should be refactored so that it passes whilst still testing the subject.

exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
```diff
@@ -1,90 +1,100 @@
...
 package org.exist.xquery;
 
-import com.evolvedbinary.j8fu.Either;
-import org.exist.EXistException;
-import org.exist.security.PermissionDeniedException;
-import org.exist.storage.BrokerPool;
-import org.exist.storage.DBBroker;
-import org.exist.test.ExistEmbeddedServer;
-import org.exist.xquery.value.Sequence;
+import org.exist.test.ExistXmldbEmbeddedServer;
```
adamretter, line 24: Why is this rewritten to use the XML:DB API when the XMLDB:API is not needed here? I think that adds an extra level of indirection and complexity which is not required. Instead you can talk directly to the database as was the case before.

exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
```diff
@@ -1,90 +1,100 @@
...
-                "};" +
-                "local:x()";
-
-        final Either<XPathException, Sequence> result = executeQuery(query);
-        assertXQueryError(ErrorCodes.XPDY0002, result);
+    public void declaredFunctionAbsoluteSlash() {
+        final String query = "declare function local:x() { /x }; local:x()";
+        final String expected =
+                "err:XPDY0002 Leading '/' selects nothing, ContextItem is absent in function body [at line 1, column 30]";
+        assertCompilationError(query, expected);
```
adamretter, line 48: Checking such complex error messages by String comparison is much more fragile than just checking for the correct Error Code. As the thing that matters here is the error code, please change this back to match the error codes.

adamretter (changes requested):
Some concerns around ContextItemTests changes... otherwise apart from some minor things at a glance it looks good. 

I still need to test some queries that Juri suggested needed checking.
Also, the XQTS must be run before and after, the results should be posted in the PR!

exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
line-o: Well, you are right it is a little more fragile to test for the complete error message than just the error code. 
This was a deliberate choice. That way the error code, the error message and the presence of the error location are tested.

exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
line-o: Because that is the code I had at hand, which by the way was merged as part of a previous PR. I am happy to rewrite the test using `org.exist.test.ExistEmbeddedServer` if that makes the tests more stable or faster or both and can still be testing for the static error message.

exist-core/src/test/xquery/xquery3/bang.xql
line-o: I will add this test to the Java test suite.

exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
adamretter: I think you likely only need to revert your change to this class, and add in the additional test that you wanted.

exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
adamretter: Okay but that mixes concerns, these tests are about the presence or not of a Content Item. The assertions here don't need to worry about the message or location info... we have other tests for location info now (in your other PR ;-))

line-o: I disagree, the tests are there for what they test. And that is
- is a static error thrown?
- is the error code the expected one
- is the location info present
- is the error message correct

exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
line-o: I would like to keep it as it is, please see my comment below.

exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
adamretter: Okay, then I can't agree to it as is :-(

line-o: `org/exist/xquery/UnknownAtomicTypeTest.java` already use this type of tests and they were approved and merged.

adamretter: I see that, but actually it could have been done better there as well. I apologise for not spotting that in the review of that PR https://github.com/eXist-db/exist/pull/3546

line-o:
3 tests were added to ensure the absolute XPath expressions are allowed after a new context is given after the simple map operator.
They succeed on a current checkout of develop but fail now.

line-o:
I hope the extended assertions are good enough for the community.

line-o:
New test helpers added and also used for `org/exist/xquery/UnknownAtomicTypeTest.java`.
All tests should now pass including those that previously failed.

line-o:
Review and constructive feedback for 20e4656cac968c18198bb8f0d5f27540b0aeed94 appreciated.

line-o:
1 Appveyor job failed with

> [ERROR] DOCKER> Unable to pull 'openjdk:8-jdk-slim' : {"message":"toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"} (Internal Server Error: 500) [{"message":"toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"} (Internal Server Error: 500)]
107

😢

line-o:
Current develop 9ad24bd93f2183f0b61d2ca10ea77fe2a922af00 is at 79.99% spec compliance (note: the actual value is likely a bit higher, see https://github.com/eXist-db/exist-xqts-runner/issues/15)

line-o:
@brihaye well spotted! Would you want to open an issue for this (or even a PR)? Would be highly appreciated. Who knows for how long this typo existed.

brihaye:
That shows a typo in : org/exist/xquery/XQueryContext.java

ContextItemDeclaration getContextItemDeclartion() 

I don't know how many dependencies there are... :-)

line-o:
Me neither :) @brihaye

line-o:
With this PR applied exist-xqts-runner reports 81.20% successful tests.

line-o:
Without this PR applied the spec runner reports 78.12%
That is an increase of 3,08%, pretty good.

line-o:
I do not know how much we can trust the time-spent output, but it also went down from 1608.112 to 1130.465 (seconds?).

joewiz:
Travis passes, but we have one failure in Appveyor. Could someone with Appveyor privileges please restart [the failed job](https://ci.appveyor.com/project/eXistdbci/exist/builds/36384017/job/js50qmf8ixecc8kp)? 

The failure:

> [ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.34.1:build (build image) on project exist-docker: Unable to pull 'openjdk:8-jdk-slim' : {"message":"toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"} (Internal Server Error: 500) -> [Help 1]

Once that passes, we'll have overall approvals from Dannes, Leif-Jöran, and Adam—as well as XQTS results showing a 3% improvement as well as evidence suggesting improved speed completing the test suite. 

Aside from the Appveyor failure, AFAICT, the only issue appears to be @adamretter's request that @line-o reduce the test to check only the error code rather than the description and location of the error:

> that mixes concerns, these tests are about the presence or not of a Context Item. The assertions here don't need to worry about the message or location info... we have other tests for location info now (in your other PR ;-))

@line-o Could you please carry out that suggestion? Then I think this will unambiguously be ready to be merged, with full consensus by the reviewers.

line-o:
I do not understand why Java tests should not check error message and location. Testing for those in XQsuite is not necessary and also not easy for compilation errors.
The vague argument "mixes concerns" does not convince me. What are the concerns for Java vs xqsuite tests.
I did address the xmlembedded server issue and also improved the code to check for the error code specifically. The error message will also never change, it is not as brittle as suggested.

line-o:
I will put this up as a topic for tomorrow to discuss.

adamretter:
> I will put this up as a topic for tomorrow to discuss.

Sure maybe it easier to understand if we can discuss on a call...

exist-core/src/test/java/org/exist/xquery/ContextItemTests.java
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @author <a href="mailto:juri@existsolutions.com">Juri Leino</a>
+ */
+public class ContextItemTests extends XQueryCompilationTest {
+
+    @Test
+    public void declaredFunctionAbsoluteSlash() throws EXistException, PermissionDeniedException {
+        final String query = "declare function local:x() { /x }; local:x()";
+        final String expected =
+                "err:XPDY0002 Leading '/' selects nothing, ContextItem is absent in function body [at line 1, column 30]";
```
line-o, line 45: contains "Leading '/' selects nothing, ContextItem is absent in function body"

exist-core/src/main/java/org/exist/test/XQueryAssertions.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public static void assertXQErrorColumn(final int expectedColumn, final XPathException exception) {
+        assertEquals("Expected column to be " + expectedColumn + ", but got " + exception.getColumn(),
+                expectedColumn, exception.getColumn());
+    }
+
+    public static void assertXQErrorMessage(final String expectedMessage, final XPathException exception) {
+        assertEquals("Expected message to be " + expectedMessage + ", but got " + exception.getMessage(),
+                expectedMessage, exception.getMessage());
```
line-o, line 60: Create a new method `exception.getMessageRaw()` that will only return only `XpathException.message`

line-o:
This is ready to be merged.

exist-core/src/main/java/org/exist/test/XQueryCompilationTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.ClassRule;
+
+import static com.evolvedbinary.j8fu.Either.Left;
+import static com.evolvedbinary.j8fu.Either.Right;
+
+/**
+ * Base class for test suites testing XQuery compilation
+ * @author <a href="mailto:juri@existsolutions.com">Juri Leino</a>
+ */
+public class XQueryCompilationTest {
```
adamretter, line 43: As this is a "Base class" it should be marked as `abstract`

exist-core/src/main/java/org/exist/test/XQueryAssertions.java
```diff
@@ -0,0 +1,90 @@
...
+        assertXQErrorLine(line, xpe);
+        assertXQErrorColumn(column, xpe);
+    }
+
+    public static void assertXQDynamicError(final ErrorCodes.ErrorCode expectedCode, final int line, final int column, final String expectedMessage, final Either<XPathException, Sequence> actual) {
+        assertXQDynamicError(expectedCode, line, column, actual);
+        assertXQErrorMessage(expectedMessage, actual.left().get());
+    }
+
+    public static void assertXQDynamicError(final ErrorCodes.ErrorCode expectedCode, final int line, final int column, final Either<XPathException, Sequence> actual) {
```
adamretter, line 57: This appears to be a copy of `assertXQStaticError`. If (as it looks to me) I`assertXQStaticError` and `assertXQDynamicError` are indeed the same, I would ask that your remove the duplicate code and rename it to just `assertXQueryError`. I can't see anything that actually asserts whether it is a dynamic or static error... or did I miss something?

line-o: Yep, the important part. One compiles the other executes the query.

exist-core/src/main/java/org/exist/test/XQueryCompilationTest.java
line-o: Thanks for mentioning it ... now.

adamretter (changes requested):
Looks much, just some duplicated utility functions that need to be collapsed.

exist-core/src/main/java/org/exist/test/XQueryAssertions.java
line-o: DynamicErrors have `Either<XPathException, Sequence>` as actual
StaticErrors have `Either<XPathException, CompiledXQuery>`

adamretter: Ah okay! I see...

Just wondering if you had considered:

1. In this code `actual.toString()` is eagerly (always) evaluated regardless:
```java
assertTrue("Expected: " + expectedCode.getErrorQName() + ", but got result: " + actual.toString(), actual.isLeft());
```
The problem here is that the `toString()` could raise an exception. For example where Sequence is a DocumentImpl which has been disposed, toString raises an exception.

2. `toString()` may not be implemented for all implementations of the `CompiledXQuery` and `Sequence` interfaces... so you may get a non-sensical error message in those cases.

You could benefit from a better formulation like:
```java
assertLeft(() -> "Expected: " + expectedCode.getErrorQName() + ", but got result: " + actual.toString(), actual);

public static <L,R> void assertLeft(final Supplier<String> message, final Either<L,R> either) {
  if (!either.isLeft()) {
    fail(message.get());
  }
}
```

line-o: If `R.toString()` might not be present, how is the above helping. Calling `message.get()` would still throw in that case, no?

line-o: Wouldn't a static failure message be a safer option. Something along the lines of
```
if (!either.isLeft()) {
  fail("Not an error")
}
```

line-o: I would settle for 
```java
    public static <L,R> void assertIsError(final ErrorCodes.ErrorCode expected, final Either<L,R> either) {
        if (!either.isLeft()) {
            fail("Expected error: " + expected.getErrorQName() + ", but no error was thrown.");
        }
    }
```

and 

```java
    public static void assertXQDynamicError(final ErrorCodes.ErrorCode expectedCode, final int line, final int column, final Either<XPathException, Sequence> actual) {
        assertIsError(expectedCode, actual);
        final XPathException xpe = actual.left().get();
        assertXQErrorLine(line, xpe);
        assertXQErrorColumn(column, xpe);
        assertXQErrorCode(expectedCode, xpe);
    }
```

line-o: @adamretter please re-review

adamretter: > If <R>.toString() might not be present, how is the above helping. Calling message.get() would still throw in that case, no?

It's not that it's not present, it's that it would raise an exception. The difference is, that when the tests pass, you won't get an exception on the toString. You are correct, that when the tests fail, even with the new formulation, you might get an exception on the toString, so you may have to diagnose the compound failure... but a failure is a failure so it's not so bothersome in that position.

line-o: wouldn't that still evaluate the other assertions if it we do not explicitly fail here? That's how understood your previous comment.

adamretter: @line-o Can you elaborate, I don't understand your question.

line-o: What happens if `assertTrue` in `assertIsError` fails? Will the following assertions run or will the test runner stop at the first failing assertion as fail would definitely do.

adamretter: Ah, I think we are talking at cross-purposes. I was saying you can get rid of `assertIsError` completely now and just use the final `assertTrue` that I posted.

line-o: That does not answer my question and would also lead to duplicate code. I do like the current solution and would keep it.

adamretter: You wrote:

> What happens if assertTrue in assertIsError fails.

As I am saying that there would be no `assertIsError` anymore, I don't see a question that I can answer. Do you have a new question or did I miss something?

line-o: > Will the following assertions run or will the test runner stop at the first failing assertion as fail would definitely do?

line-o: And if not, I would like to avoid to write `if (!either.isLeft()) { fail(message) }` twice

line-o: ping?

line-o: OK, I tested your proposed solution and it works as expected. The other assertions are not being evaluated. This was changed in the XQueryAssertions class.

line-o:
This should be good to go now.

dizzzz (approved):

dizzzz:
build fails... 
image

exist-core/src/test/java/org/exist/xquery/AbsolutePathTests.java
```diff
@@ -0,0 +1,160 @@
...
+    public void immediateLambdaWithDocumentAndDoubleSlash() throws EXistException, PermissionDeniedException {
+        final Source expected = elemSource("<result><x/><x/><x/></result>");
+
+        final String query = "let $f := function($d) { $d ! //x }\n" +
+                "let $d := document { <root><x/><x/><y><x/></y></root> }\n" +
+                "return\n" +
+                "  <result>{ $f($d) }</result>";
+        final Either<XPathException, Sequence> actual = executeQuery(query);
+
+        assertXQResultSimilar(expected, actual);
```
line-o, line 98: @adamretter Why did you choose to check for similarity here?
Does it give a better output in case it does not match or is it a looser check to get around whitespace differences or other minor differences?

exist-core/src/test/java/org/exist/xquery/AbsolutePathTests.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Creates an Document6 Source form an XML String.
+     *
+     * @param str a string representation of XML.
+     *
+     * @return a Document Source.
+     */
+    private static Source docSource(final String str) {
```
line-o, line 144: would this helper function be of value in either
- the abstract class XQueryCompilationTest
- the assertions in XQueryAssertions
- or a new class ~XMLDomMatchers or so

exist-core/src/test/java/org/exist/xquery/AbsolutePathTests.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Creates an Element Source form an XML String.
+     *
+     * @param str a string representation of XML.
+     *
+     * @return an Element Source.
+     */
+    private static Source elemSource(final String str) {
```
line-o, line 155: same here

line-o (commented):
@adamretter Thanks for back porting the Diffmatcher. A lot of tests can likely benefit from this. The string match is no match to real value comparisons.

exist-core/src/test/java/org/exist/xquery/AbsolutePathTests.java
line-o: I am asking this because I see myself searching for that code in the future and also would like to not have it copied in several test classes.

exist-core/src/test/java/org/exist/xquery/AbsolutePathTests.java
adamretter: The difference between `similar and `identical` is referenced from the JavaDoc in `DiffMatcher.java` and documented here https://www.xmlunit.org/api/java/2.6.0/org/xmlunit/builder/DiffBuilder.html#checkForSimilar-- and 

> is it a looser check to get around whitespace differences or other minor differences

Yes.

exist-core/src/test/java/org/exist/xquery/AbsolutePathTests.java
adamretter: I will move them back into `DiffMatcher` which is where they came from in FusionDB.

exist-core/src/test/java/org/exist/xquery/AbsolutePathTests.java
adamretter: See above.

line-o:
😿

adamretter:
The CI failures have revealed a latent transient dependency issue with JAXB on OpenJDK 11. I have opened a new PR https://github.com/eXist-db/exist/pull/3643 to fix the issue, once that is merged, this PR (once rebased) should start passing CI again.

joewiz:
@line-o With #3643 now merged, would you please rebase this PR?

adamretter:
@joewiz I have taken care of?

dizzzz:
craps you are sooooo fast ; I was about to start :-)

joewiz:
Thanks @adamretter ;)

The first AppVeyor environment passed, but the 2nd failed with this:

```txt
ERROR] Received HTTP 502 when trying to access: http://exist-db.org/exist/apps/public-repo/find?processor=5.3.0-SNAPSHOT&abbrev=dashboard&info=true
[ERROR] Failed to execute goal org.exist-db.maven.plugins:public-xar-repo-plugin:1.2.0:resolve (fetch-xars) on project exist-distribution: Unable to get package info -> [Help 1]
```

I tried this URL and couldn't reproduce the 502 error, so it must've been a temporary glitch. 

@dizzzz Could you restart the test? (I'd do it myself but somehow lost permissions.)

joewiz:
@adamretter Ok, tests are passing! The PR still thinks you have requested changes. Rather than having an admin override your requests, could you mark then as resolved so the PR can be merged?

adamretter (approved):
It's now good to go :-)

dizzzz (approved):

PR #3609, opened 2020-11-06
Update to Java15

dizzzz:
Upgrade to curent OpenJDK version (15) in Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ jobs:
         - BUILD_DOCKER=false
         - MAVEN_OPTS="-Ddependency-check.skip=true"
 
-    - jdk: openjdk14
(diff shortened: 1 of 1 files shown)


dizzzz:
from a colleague I heard that a nice thing of recent OpenJDKs is.... that you can actually see where -in case of a NPE- which variable is causing the problem....

adamretter (changes requested):
Do the errors that we are experiencing show on the JDK 15 version that is present on Travis? if not, maybe we should also keep JDK 14.

Either way, we need to keep/update/remove the "allow_failures" section of `.travis.yml`

dizzzz:
@adamretter thnx for input, I re-added openjdk14...

Issue #3610, opened 2020-11-08
xmldb:login raises IllegalStateException


adamretter:
Sometimes when calling `xmldb:login` Jetty raises the following IllegalStateException when eXist-db tries to store the authenticated user in the Session.

eXist-db should handle this case (whatever the root issue is) gracefully.

```
Caused by: java.lang.IllegalStateException: Not valid for write: id=node01t6o5gyqhskj1851thoquyv2214675 created=1604829866943 accessed=1604829872742 lastaccessed=1604829866943 maxInactiveMs=1800000 expiry=1604831672742
	at org.eclipse.jetty.server.session.Session.checkValidForWrite(Session.java:667) ~[jetty-server-9.4.34.v20201102.jar:9.4.34.v20201102]
	at org.eclipse.jetty.server.session.Session.setAttribute(Session.java:806) ~[jetty-server-9.4.34.v20201102.jar:9.4.34.v20201102]
	at org.exist.http.servlets.HttpSessionWrapper.setAttribute(HttpSessionWrapper.java:121) ~[exist-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.exist.xquery.functions.xmldb.XMLDBAuthenticate.lambda$0(XMLDBAuthenticate.java:193) ~[exist-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at java.util.Optional.ifPresent(Optional.java:183) ~[?:?]
	at org.exist.xquery.functions.xmldb.XMLDBAuthenticate.cacheUserInHttpSession(XMLDBAuthenticate.java:193) ~[exist-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.exist.xquery.functions.xmldb.XMLDBAuthenticate.eval(XMLDBAuthenticate.java:170) ~[exist-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
```

ndw:
This appears to be reproducible if the browser cache is cleared of cookie data before the request.

PR #3612, opened 2020-11-08
Avoid IllegalStateException in Http Session functions

adamretter:
Fixes an `IllegalStateException` which could occur when eXist-db tries to perform Http Session functions but the session has been invalidated (but not yet cleaned up by Jetty).

I back-ported these tests and fixes from FusionDB.

Closes https://github.com/eXist-db/exist/issues/3610


dizzzz (approved):
LGTM

reinhapa:
LGTM

PR #3613, opened 2020-11-08
Log a warning if a collection.xconf is binary

adamretter:
Log's a warning if an encountered Collection Configuration document (e.g. collection.xconf) is a Binary document instead of an XML Document.


exist-core/src/main/java/org/exist/dom/persistent/DocumentImpl.java
```diff
@@ -444,11 +444,6 @@ public void setReferenced(final boolean referenced) {
         isReferenced = referenced;
     }
 
-    @EnsureContainerLocked(mode=READ_LOCK)
```
dizzzz, line 447: not used?

dizzzz (approved):

PR #3621, opened 2020-11-17
Remove code which duplicates Commons IO

adamretter:
A very simple small change to remove a class which duplicates behaviour that is available in an already used dependency.


dizzzz:
I remember to implement my own counter as the commons-io classes are doing a lot more under the hood, leading to overhead.

line-o (dismissed):
LGTM

adamretter:
@dizzzz Where do you see the overhead? I would think that reading the data from the database would far eclipse any overhead in the counting.

dizzzz:
I tried te remember from my head; need to check :-)

dizzzz (approved):

PR #3622, opened 2020-11-18
Fix NPE and Deadlock when Path Locks are used for Documents

adamretter:
The use of Path Locks for Documents has always been experimental. However, some issues have been reported when they are enabled by advanced users. This PR addresses those issues.

**NOTE:** I have tested this by running the full test-suite for eXist-db with `<document use-path-locks="false"/>` and then again with `<document use-path-locks="true"/>` set in `conf.xml`. All tests pass :-)


dizzzz (approved):
never seen the Path locks.... :-/

Issue #3624, opened 2020-11-20
[BUG] NPE at org.exist.xquery.modules.range.RangeQueryRewriter.rewriteLocationStep


joewiz:
**Describe the bug**

In eXist develop HEAD (5.3.0-SNAPSHOT), a new NPE is raised when running a query. The same query did not raise an NPE in the current stable release, 5.2.0.

**Expected behavior**

The query should not raise an NPE.

**To Reproduce**

To reproduce, install https://github.com/HistoryAtState/release (here's a prebuilt xar: [release-0.5.0.xar.zip](https://github.com/eXist-db/exist/files/5576302/release-0.5.0.xar.zip) - just remove the .zip extension), and open http://localhost:8080/exist/apps/release/ebook-batch.xq. The client sees the following 500 error:

> HTTP ERROR 500 javax.servlet.ServletException: javax.servlet.ServletException: An error occurred while processing request to /exist/apps/release/ebook-batch.xq: An error occurred: null

Attached is the corresponding error from exist.log: [exist-log-npe-rangequeryrewriter.txt](https://github.com/eXist-db/exist/files/5576331/exist-log-npe-rangequeryrewriter.txt)

An excerpt:

```txt
2020-11-20 17:13:58,622 [qtp558431044-99] ERROR (EXistServlet.java [doGet]:291) - null 
java.lang.NullPointerException: null
	at org.exist.xquery.modules.range.RangeQueryRewriter.rewriteLocationStep(RangeQueryRewriter.java:109) ~[exist-index-range-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.Optimizer.visitLocationStep(Optimizer.java:83) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.LocationStep.accept(LocationStep.java:1242) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.DynamicCardinalityCheck.accept(DynamicCardinalityCheck.java:139) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
	at org.exist.xquery.DefaultExpressionVisitor.visitBuiltinFunction(DefaultExpressionVisitor.java:48) ~[exist-core-5.3.0-SNAPSHOT.jar:5.3.0-SNAPSHOT]
[...]

PR #3625, opened 2020-11-22
Proposed fix for #3624 NPE at RangeQueryRewriter.rewriteLocationStep

dizzzz:
Proposed fix for #3624 NPE at org.exist.xquery.modules.range.RangeQueryRewriter.rewriteLocationStep
Handle all operators

Help needed for tests


joewiz:
I can confirm that this fixes the issue I reported in #3624. Thank you, @dizzzz!

adamretter:
@dizzzz Interesting stuff... I have some questions ;-)

1. Previously this was limited to `fn:matches` but now it is only restricted to a Function that implements `Lookup`. Joe mentioned he did not see the problem in 5.2.0 - but this code would not have changed between 5.3.0 and 5.2.0.

2. What was the expression that was causing the issue?

3. This code can still cause an NPE:

```
Lookup func = Lookup.create(function.getContext(), operator, path);
func.setArguments(eqArgs);
```

The `Lookup.Create` function can return `null`. So the value of `func` should probably be checked/logged before it is used or not.

dizzzz:
`eq` came in as option instead of 'matches', then I realized that only one item was handled. In other cases than `matches` NULL was returned, yielding in  NPE; 

Instead of preventing the `eq` NPE, I thought it would make sense just to handle all `enum` options 👍 That should be OK as all options should accept 2 parameters,

Hmmm did I miss a potential NPE? I thought I covered it all....

adamretter:
Hmmm, shouldn't `eq` be handled just above in the `if (expression instanceof GeneralComparison) {` branch?

dizzzz:
Hmmmm  you are right on this I guess. good spot. hmmmm

dizzzz:
What shall I do? close PR?

extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/RangeQueryRewriter.java
```diff
@@ -138,13 +138,17 @@ protected static Lookup rewrite(Expression expression, NodePath path) throws XPa
...
             if (function instanceof Lookup) {
-                if (function.isCalledAs("matches")) {
-                    eqArgs.add(function.getArgument(0));
-                    eqArgs.add(function.getArgument(1));
-                    Lookup func = Lookup.create(function.getContext(), RangeIndex.Operator.MATCH, path);
-                    func.setArguments(eqArgs);
-                    return func;
+                final RangeIndex.Operator operator = RangeIndex.Operator.getByName(function.getName().getLocalPart());
+
+                if(function==null){
```
wolfgangmm, line 143: The null check is too late here. Should go further up.

extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/RangeQueryRewriter.java
```diff
@@ -138,13 +138,17 @@ protected static Lookup rewrite(Expression expression, NodePath path) throws XPa
...
+                    return null;
                 }
+                
+                eqArgs.add(function.getArgument(0));
+                eqArgs.add(function.getArgument(1));
+                Lookup func = Lookup.create(function.getContext(), operator, path);
+                func.setArguments(eqArgs);
```
wolfgangmm, line 150: Unlikely, but func could be null. Check here?

wolfgangmm (changes requested):
I debugged the original issue myself and came to the conclusion that Dannes' proposed fix is basically correct. What we have here is an expression which has already been partially rewritten (the `GeneralComparison` was replaced by `range:eq`). I'm not sure where this partial optimization was applied - probably an earlier traversal of the expression (we would need a much more reduced test case to know), but it is certainly a possible case which should not lead to an exception.

As Dannes pointed out, `RangeQueryRewriter.rewrite` only handled `matches` (stupid me!). It should handle all functions implementing `Lookup`.

extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/RangeQueryRewriter.java
dizzzz: done

extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/RangeQueryRewriter.java
dizzzz: done

dizzzz:
@wolfgangmm @adamretter ready to go

joewiz (approved):
@dizzzz I can confirm that the updated PR still fixes the issue I reported in #3624. Thank you!

wolfgangmm (approved):

joewiz:
@dizzzz Thank you so much for this fix! And thanks to @adamretter and @wolfgangmm for your reviews.

adamretter:
I just dug into this further to try and understand why this fix works. I reduced Dannes's changes to the absolute minimum to pass Joe's reported issue #3624, and was left with just this patch:

```
diff --git a/extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/RangeQueryRewriter.java b/extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/RangeQueryRewriter.java
index 04f3db1f89..f7d47facce 100644
--- a/extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/RangeQueryRewriter.java
+++ b/extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/RangeQueryRewriter.java
@@ -139,13 +139,12 @@ public class RangeQueryRewriter extends QueryRewriter {
             InternalFunctionCall fcall = (InternalFunctionCall) expression;
             Function function = fcall.getFunction();
             if (function instanceof Lookup) {
-                if (function.isCalledAs("matches")) {
                     eqArgs.add(function.getArgument(0));
                     eqArgs.add(function.getArgument(1));
-                    Lookup func = Lookup.create(function.getContext(), RangeIndex.Operator.MATCH, path);
+                    final RangeIndex.Operator operator = RangeIndex.Operator.getByName(function.getName().getLocalPart());
+                    Lookup func = Lookup.create(function.getContext(), operator, path);
                     func.setArguments(eqArgs);
                     return func;
-                }
             }
         }
         return null;
```

As @wolfgangmm hinted at previously, the statement `if (function.isCalledAs("matches")) {` meant that `RangeQueryRewriter#rewrite(Expression, NodePath)` would return `null`, if it was presented with an instance of the `Lookup` class (which is an expression that is optimized for the Range Index) that was not `range:matches#2`.

The more I think about this, the less sense the current code makes. If we already have an instance of `Lookup` then it is already optimized and so it doesn't need to be passed to `RangeQueryRewriter#rewrite` to be rewritten as a Lookup function, because it is already a Lookup function!

With that in mind I still think it likely that the current fix is addressing a symptom rather than the root cause of the problem, and that eXist-db is doing more work than necessary by replacing one optimized Lookup expression with exactly the same optimized Lookup expression. I will keep digging...

adamretter:
> If we already have an instance of Lookup then it is already optimized and so it doesn't need to be passed to RangeQueryRewriter#rewrite to be rewritten as a Lookup function, because it is already a Lookup function!

We (Evolved Binary) have provided an updated PR that takes a more efficient path in - https://github.com/eXist-db/exist/pull/4882

PR #3626, opened 2020-11-22
Small code improvements

dizzzz:
nothing special

### DIFF
--- a/exist-ant/src/main/java/org/exist/ant/XMLDBExtractTask.java
+++ b/exist-ant/src/main/java/org/exist/ant/XMLDBExtractTask.java
@@ -36,6 +36,8 @@ import org.exist.xmldb.ExtendedResource;
 
 import java.io.*;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Properties;
 
 import javax.xml.transform.OutputKeys;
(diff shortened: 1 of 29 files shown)


exist-ant/src/main/java/org/exist/ant/XMLDBExtractTask.java
```diff
@@ -273,6 +260,18 @@ private void writeXMLResource(final XMLResource res, final File dest) throws IOE
         }
     }
 
+    private Writer getWriter(XMLResource res, File dest) throws XMLDBException, FileNotFoundException {
+        final Writer writer;
+        if (dest.isDirectory()) {
+            String fname = res.getId();
+            final File file = new File(dest, fname);
+            writer = new OutputStreamWriter(new FileOutputStream(file), UTF_8);
```
adamretter, line 268: No biggie, but for Java 8 this could be:

```java
writer = Files.newBufferedWriter(file, UTF_8);
```

dizzzz: That is a nice one, I will follow-up on that. Not sure what to use for the PathOptions yet...

adamretter (approved):
:-)

Issue #3635, opened 2020-11-30
[feature] Implement all remaining XPath and XQuery 3.1 Functions & Operators


joewiz:
According to https://github.com/eXist-db/documentation/blob/master/src/main/xar-resources/data/xquery/xquery.xml#L111-L149, the current list of functions not implemented in eXist is:

- [ ] `fn:collation-key#1, #2`
- [ ] `fn:default-language#0`
- [ ] `fn:document-uri#0`
- [ ] `fn:element-with-id#1, #2`
- [ ] `fn:format-integer#2, #3`
- [ ] `fn:nilled#0`
- [ ] `fn:path#0, #1`
- [ ] `fn:round#2`
- [ ] `fn:transform#1`
- [ ] `fn:uri-collection#0, #1`
- [ ] `map:find#2`
- [ ] `map:merge#2`

yamahito:
+1 for `uri-collection()`!

joewiz:
All implemented in `develop`

PR #3644, opened 2020-12-01
Implement fn:document-uri#0

dizzzz:
See https://github.com/eXist-db/exist/issues/3635 

No idea to write test for XPTY0004 as it is not possible to pass a non-node?

image


joewiz:
@dizzzz Super, thank you!

I'm not sure if this is what you were looking for, but the following query will return an XPTY0004:

```xquery
xquery version "3.1";

declare context item := "a";

document-uri()
```

Returns:

> err:XPTY0004 Context item is not a node. [at line 5, column 1]

I ran the XQTS (HEAD) `fn-document-uri` test both develop and your PR, and both return the same 22 failures. The command I ran after mvn installing each:

> `target/scala-2.13/exist-xqts-runner-assembly-1.0.0.jar -x HEAD -ts fn-document-uri`

Here's the junit result: [TEST-fn-document-uri.xml.zip](https://github.com/eXist-db/exist/files/5625966/TEST-fn-document-uri.xml.zip). I'm not sure if the failures are caused by our implementation of this function or other eXist factors outside this function.

joewiz:
@dizzzz With your latest commits, the XQSuite results for fn-document-uri jumped up from 40% passing to 76%, thanks to 9 more tests passing, leaving only 13 failures. Here's the updated junit result: [TEST-fn-document-uri.xml.zip](https://github.com/eXist-db/exist/files/5632470/TEST-fn-document-uri.xml.zip)

dizzzz:
@joewiz  Thnx!

one of the failing tests look like below:

image


so maybe the framework does not prepare the required data in the correct way?

adamretter:
> so maybe the framework does not prepare the required data in the correct way?

It is possible... What aspect of the query context you think is not being set correctly?

exist-core/src/main/java/org/exist/xquery/functions/fn/FunDocumentURI.java
```diff
@@ -21,86 +21,105 @@
...
+    public FunDocumentURI(final XQueryContext context, final FunctionSignature signature) {
         super(context, signature);
     }
 
     /* (non-Javadoc)
      * @see org.exist.xquery.Expression#eval(org.exist.xquery.StaticContext, org.exist.dom.persistent.DocumentSet, org.exist.xquery.value.Sequence, org.exist.xquery.value.Item)
      */
-    public Sequence eval(Sequence contextSequence, Item contextItem)
-            throws XPathException {
+    public Sequence eval(final Sequence contextSequence, final Item contextItem) throws XPathException {
```
adamretter, line 66: Can we replace the `non-Javadoc` with an `@Override` please?

exist-core/src/test/java/org/exist/xquery/functions/fn/FunDocumentUriTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public static final ExistXmldbEmbeddedServer existEmbeddedServer = new ExistXmldbEmbeddedServer(true, true, true);
+
+    @Test
+    public void testFnFunDocumentUri() {
+
+        try {
+            final ResourceSet resourceSet = existEmbeddedServer.executeQuery(
+                    "declare context item := 'a'; document-uri()");
+
+        } catch (XMLDBException ex) {
```
adamretter, line 51: Some `final` here and the line below please.

exist-core/src/test/xquery/fn.xql
```diff
@@ -126,3 +126,46 @@ function fnt:tokenize-onearg($str as xs:string) {
...
+     document-uri(root(collection('/db/fn-test')//book))
+};
+
+declare
+    %test:assertEmpty
+function fnt:document-uri1_empty() {
+     document-uri(root(collection('/db/fn-test')//bookies))
+};
+
+declare
```
adamretter, line 165: I think maybe you were searching for `%test:assertTrue` ;-)

```xquery
declare
    %test:assertTrue
function fnt:fn-document-uri-32() {
     let $uri := "/db/fn-test/test.xml"
     return
         fn:contains(fn:doc($uri)/document-uri(), $uri)
};
```

adamretter (commented):
Just some thoughts...

dizzzz:
@adamretter  looking at this specific example, it looks like `$uri` is not set or pointing into a 'wrong' direction.
To be sure I added the same test to the xqsuite file, so we can validate if the implementation of the function is ok... or not.

line-o:
@dizzzz I see you have created Java tests. Now that #3605 was merged, we could make use of XQueryCompilationTest and XQueryAssertions. Would you like to go that route?

exist-core/src/test/java/org/exist/xquery/functions/fn/FunDocumentUriTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.xmldb.api.base.ResourceSet;
+import org.xmldb.api.base.XMLDBException;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Dannes Wessels
+ */
+@RunWith(ParallelRunner.class)
+public class FunDocumentUriTest {
```
line-o, line 39: Could extend `XQueryCompilationTest`

exist-core/src/test/xquery/fn.xql
```diff
@@ -126,3 +126,46 @@ function fnt:tokenize-onearg($str as xs:string) {
...
+    %test:assertEquals("TRUE")
+function fnt:fn-document-uri-32() {
```
line-o, line 167: In the description you reference fn:document-uri-33 as a possible false negative. Is this the one you added to track that or did I misread your comment?

line-o (commented):
Very cool @dizzzz 🎉

exist-core/src/test/xquery/fn.xql
line-o: OK, they likely only differ in calling upper-case (33) and lower-case (32), right?

exist-core/src/test/xquery/fn.xql
dizzzz: :-) made the change blindly :-)

exist-core/src/test/xquery/fn.xql
dizzzz: indeed....

exist-core/src/test/xquery/fn.xql
dizzzz: Hmmm yes, but I followed the original test, no further modification

dizzzz:
@line-o  if XQueryCompilationTest  provides me benefits, yes sure

joewiz:
@dizzzz I realize you've got a java version of the XPTY0004 test now, but I can confirm that the `util:eval-with-context()` approach does work; I'd previously expressed the 1st parameter incorrectly. 

```xquery
util:eval-with-context("document-uri()", (), false(), "a")
```

With your branch, it raises the error:

> err:XPTY0004 Error while evaluating expression: document-uri(). Context item is not a node. [at line 3, column 1]

Obviously, there's no need to switch to this. I just wanted to close the loop ;)

joewiz (approved):
@dizzzz LGTM, thanks! (I'm happy to run XQTS suites anytime; just let me know when/if you'd like additional runs.)

dizzzz:
@joewiz thnx for identifying the new option! I migrated accordingly to have all tests in one location.

joewiz:
The Visual Studio + JDK 11 build on AppVeyor timed out. Could someone with permissions please restart that CI run?

dizzzz:
@adamrette I copied one XQtsuite test to our tests; this test is running fine which suggests that the xq-runner does not pass the correct/sufficient data to the test?

dizzzz:
@joewiz unfortunately these time outs appear to happen more frequently

dizzzz:
All green!

line-o (approved):
great!

PR #3653, opened 2020-12-06
Fix/2445

line-o:
### Description:

The following XPaths must evaluate to an empty sequence:

- `<a/>[self::b]`
- `<a/>/@b/c`
- `<a/>/text()/b`

### Reference:

Both baseX and Saxon return an empty sequence instead of throwing an exception.

### Type of tests:

TBD


line-o:
With this PR applied all tests in `XQTS_HEAD.prod-AxisStep.static-typing` fail (-4).

line-o:
It passes two tests that previously failed `Axes043-1` and `Axes043-2`.
> Path 'descendant-or-self::node()' from a text node.

```xquery
<a>test</a>/text()/descendant-or-self::node()
```

line-o:
The thing we need to agree on:
Do we want to adhere to the tests in XQTS and throw an error for XPaths that select nothing or do we want to handle this more loosely.
Think about a query selecting a specific element inside a document. Should this query throw each time a document does not have that? One would have to check for existence first or use try catch to do that to ensure the script runs on all documents and not stop at the first encounter of a document that does not have that element.
On the flip side, such a static check could help find problems in a query.
The previous solution only applied to very easy queries, so it definitely has to change.

line-o:
## needs documentation

The community agreed today that this PR can land in a patch or minor release.

existdb will have to state at an appropriate (prominent?) location in the documentation that it does not provide the StaticTyping feature.

line-o:
Followup on todays discussion about XQTS and the staticTyping feature:
Deactivating that feature will skip 74 Tests from the total of 31520.
26 of those exist-db passed, 43 were failing and 5 errored.
We may deactivate it but leaving it in will “hurt” our score only by a measly 0.01% (the 4 testcases in AxisStep staticTyping that would fail with PR 3653 applied).

adamretter:
@line-o I just want to see if I understand correctly, does this PR cause additional test failures in XQTS (in the static typing tests)? If so I think that needs to be resolved first, as I would rather see the failure of tests that previously passed as a regression.

line-o:
@adamretter Well, in that case I am in favour of dropping the tests of a feature that eXist-db does not implement.

adamretter:
@line-o I am still not clear on your meaning. Can you answer the question I posed? This will help make sure I understand what you were saying about the XQTS tests.

line-o:
> Deactivating that feature will skip 74 Tests

As **Static Typing** is an optional feature of XQTS we could skip those tests as we do with other features that exist-db does not implement.

adamretter:
@line-o So I am not sure I understand you exactly still, but my thoughts are this:

1. eXist-db passes 35% of the Static Typing feature tests
2. There was obviously an effort in the past to add Static Typing to eXist-db, this can be seen in the code.
3. eXist-db neither support or does not support Static Typing, it passes some tests but fails others. The same is true of XQuery 3.1 as a whole.
4. The XQuery spec allows for an implementation to:
    1. Support Static Typing
    2. Not Support Static Typing
    3. Both. i.e. provide an implementation defined option to enable/disable Static Typing.

As eXist-db already supports some of Static Typing, just fixing a couple of issues doesn't remove it from eXist-db, it just diminishes what is already there, and swings the needle a little. I would suggest that instead:

1. We add an option to enable/disable Static Typing - probably a `declare option` in the prolog, and a global conf.xml option.
2. We decided whether Static Typing should be enabled or disabled by default. As it is only partially implemented so far, it should likely be disabled by default. However, changing the status-quo is likely to break many many existing applications in strange and unexpected ways, it therefore might be better to leave it enabled by default (even if it is partial support)
3. We then make the code changes to match the above.

PR #3658, opened 2020-12-06
Make import statements consistent in whole repo

dizzzz:
In essence a check on code quality metrics.

If it blocks development, we should not pull it in.


adamretter:
@dizzzz So when you say consistent, what do you mean?

Do you mean alphabetical by package? do you also mean using or not using wildcards like `*`?

dizzzz:
- all ordered in the same way
- same use of white spacing
- same way of using .* (or : not using it)
- remove unused imports


(in essence: asked Intellij to do a global check en modify on imports :-) )

PR #3670, opened 2020-12-12
[fix] basic authentication

line-o:
A common authentication strategy is the bearer token authentication.

- The client sends `Authorization: Bearer [bearer-token]` to the server
- The server processes the Authorization header contents

At the moment, exist-db throws an error with that header. 
The BasicAuthenticator class treats any Authorization header as a basic authentication attempt, regardless of the scheme (`Basic`, `Bearer`, ...).

This PR changes BasicAuthenticator to also check if the header value starts with `Basic` before attempting to authenticate using the Basic Authentication strategy.


adamretter (approved):
LGTM

dizzzz (approved):

PR #3671, opened 2020-12-13
Concept: Make servlets pluggable

dizzzz:
Currently the IPrange servlet is (more or less) hardcoded initialised during eXist-db startup.

I think that this specific servlet should be removed BUT the concept introduced in this PR might be useful eg for authentication/authorization alternatives.

Please could you share some thoughts ?

### DIFF
--- a/exist-core/src/main/java/org/exist/jetty/JettyStart.java
+++ b/exist-core/src/main/java/org/exist/jetty/JettyStart.java
@@ -21,6 +21,32 @@
  */
 package org.exist.jetty;
 
+import net.jcip.annotations.GuardedBy;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.eclipse.jetty.server.*;
+import org.eclipse.jetty.server.handler.ContextHandler;
+import org.eclipse.jetty.server.handler.HandlerWrapper;
+import org.eclipse.jetty.servlet.ServletContextHandler;
+import org.eclipse.jetty.servlet.ServletHolder;
+import org.eclipse.jetty.util.Jetty;
+import org.eclipse.jetty.util.MultiException;
+import org.eclipse.jetty.util.component.LifeCycle;
+import org.eclipse.jetty.xml.XmlConfiguration;
+import org.exist.SystemProperties;
+import org.exist.http.servlets.ExistExtensionServlet;
+import org.exist.start.Main;
+import org.exist.storage.BrokerPool;
+import org.exist.util.ConfigurationHelper;
+import org.exist.util.FileUtils;
+import org.exist.util.SingleInstanceConfiguration;
+import org.exist.validation.XmlLibraryChecker;
+import org.exist.xmldb.DatabaseImpl;
+import org.exist.xmldb.ShutdownListener;
+import org.xmldb.api.DatabaseManager;
+import org.xmldb.api.base.Database;
+
+import javax.servlet.Servlet;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.LineNumberReader;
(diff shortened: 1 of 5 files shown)


line-o (commented):
Looks good as far as I can tell.
Will this allow for packages to add additional ExistExtensionServlets?

dizzzz:
@line-o  Yes this means that it is possible to add more JAR files that contain extensions, without changing any code.

exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -298,25 +286,24 @@ public synchronized void run(final String[] args, final Observer observer) {
...
+                if (handler instanceof ServletContextHandler) {
 
-                        logger.info("'" + contextHandler.getContextPath() + suffix + "'");
-                    }
+                    ServiceLoader<ExistExtensionServlet> load = ServiceLoader.load(ExistExtensionServlet.class);
+                    load.forEach(existExtensionServlet ->
+                    {
+                        final ServletContextHandler contextHandler = (ServletContextHandler) handler;
+                        final String pathSpec = existExtensionServlet.getPathSpec();
+                        final Servlet servlet = (Servlet) existExtensionServlet;
```
adamretter, line 299: I think `ExistExtensionServlet` should like extend `Servlet` so that we can avoid this cast which could fail otherwise.

extensions/security/iprange/pom.xml
```diff
@@ -64,5 +64,28 @@
         </dependency>
 
     </dependencies>
+
+    <build>
+    <plugins>
+        <plugin>
+            <groupId>eu.somatik.serviceloader-maven-plugin</groupId>
```
adamretter, line 71: Can you tell me about this plugin and its purpose?

adamretter (commented):
Questions ;-)

extensions/security/iprange/pom.xml
dizzzz: it scans code and generates service loader config files in the META-INF directory.

image

exist-core/src/main/java/org/exist/jetty/JettyStart.java
dizzzz: lemme try again

dizzzz:
improved logging:

```
14 Dec 2020 20:14:40,897 [main] INFO  (JettyStart.java [run]:283) - Configured contexts: 
14 Dec 2020 20:14:40,898 [main] INFO  (JettyStart.java [run]:289) - /exist (eXist XML Database) 
14 Dec 2020 20:14:40,903 [main] INFO  (JettyStart.java [lambda$run$1]:299) - /exist/iprange (IPrange filter) 
14 Dec 2020 20:14:40,903 [main] INFO  (JettyStart.java [run]:289) - / (eXist-db portal) 
14 Dec 2020 20:14:40,905 [main] INFO  (JettyStart.java [lambda$run$1]:299) - //iprange (IPrange filter) 
```

extensions/security/iprange/pom.xml
adamretter: I think the Maven Plugin is unnecessary though and for the feature it provides it will slow down the build. I think it would be better just to add the file `META_INF/services/org.exist.http.servlets.ExistExtensionServlet` to the `src/resources` folder of the module manually.

adamretter (changes requested):
In general this is great, I think we could loose the additional Maven plugin though, as it doesn't really give us much.

dizzzz:
new logs:

image

dizzzz:
@adam , in the previous picture. I find adding "iprange" to  both "/exist" and "/" a bit strange.
For 5.x this is should probably not be changed.

adamretter (approved):

joewiz:
@dizzzz When you said:

> in the previous picture. I find adding "iprange" to both "/exist" and "/" a bit strange. For 5.x this is should probably not be changed.

... I gather you were implying that something could/should be changed in 6. Do you have a recommendation? Drop iprange from the standard configuration?

dizzzz:
The original user does not use it anymore; the original author is not working on it anymore too;
So I guess we have 2 options
- delete the iprange filter
- have not not deployed baby default.

Anyway I realise I should document the new possibilities.

PR #3669, opened 2020-12-10
Use fixed user environment for tests

reinhapa:
### Description:
- converts the argument line into system property variables
- defines a fixed user environment for running the tests

PR #3673, opened 2020-12-15
Test/force test location dwes

dizzzz:
chase NPEs


dizzzz:
will be delivered with #3669

Issue #3651, opened 2020-12-05
[BUG] Unicode characters beyond the BMP get distorted in conversion from node to string values


cwittern:
**Describe the bug**
Characters, in my case Chinese characters such as "𨛌" can be saved and retrieved as character content of XML documents without problems. However, when processed in XQuery where an implicit string conversion occurs, or when retrieving the text content with the XQuery text() function, the text gets distorted.

**Expected behavior**
The returned content should always be the character content of the document

**To Reproduce**
The following when pasted and run in eXide will demonstrate the problem:
===cut here ===
```xquery
xquery version "3.1";


let
 $node := <graph>𨛌</graph>
, $savedoc := xmldb:store("/db", "temp.xml", $node)
, $doc := doc("/db/temp.xml")
 return 
<res>
    <node>{$node},{$node/text()},{string-length($node)}</node>
    <doc>{$doc},{$doc/text()},{string-length($doc)}</doc>
</res>
```
=== cut here == 

The result I receive is:
```xml
<res>
    <node>
        <graph>𨛌</graph>,𨛌,1</node>
    <doc>
        <graph>𨛌</graph>,,2</doc>
</res>
```

the expected result would be

```
<res>
    <node>
        <graph>𨛌</graph>,𨛌,1</node>
    <doc>
        <graph>𨛌</graph>,𨛌,1</doc>
</res>

```
Note: the two character string is not shown in eXide, but the string-length is shown to be 2.  In actual usage in a web application, frequently the characters �� are observed, eg two characters with the Unicode codepoint U+FFFD

**Context (please always complete the following information):**
 - OS: Ubuntu 20.04
 - eXist-db version:  5.2.0
 - Java Version :  openjdk 11.0.9.1 2020-11-04

**Additional context**
- How is eXist-db installed? eXist is installed by expanding the distribution tarball
- Any custom changes in e.g. `conf.xml`?
No changes except index parameters, triggers for backup etc.

dizzzz:
Please could you attach the files, and maybe some screenshots of the documents? In de web/my mac the characters are not rendered correctly.

cwittern:
Even if you can't see the correct character, you should be able to observe that when read from the database the character is not displayed, on the other hand it has the string-length of 2, instead of 1.  Anyway, here is a screenshot from my system which might be easier to understand.  The arrow marks the problem
[...]

PR #3676, opened 2020-12-18
Update unicode.xqm

cwittern:
Added test for bug in https://github.com/eXist-db/exist/issues/3651 in explicit conversion to text using text()

### DIFF
--- a/exist-core/src/test/xquery/unicode.xqm
+++ b/exist-core/src/test/xquery/unicode.xqm
@@ -72,7 +72,7 @@ function uni:sip-dom-persistent() {
 
 declare
     %test:assertEquals(172998)
-function uni:sip-dom-persistent() {
+function uni:sip-dom-persistent-text() {
     let $stored := xmldb:store("/db/" || $uni:COLLECTION_NAME, "sip.xml", $uni:SIP)
     return
         fn:string-to-codepoints(fn:doc($stored/text())/container/smp)
--- a/exist-core/src/test/xquery/unicode.xqm
+++ b/exist-core/src/test/xquery/unicode.xqm
@@ -72,6 +72,14 @@ function uni:sip-dom-persistent() {
 
 declare
     %test:assertEquals(172998)
+function uni:sip-dom-persistent() {
+    let $stored := xmldb:store("/db/" || $uni:COLLECTION_NAME, "sip.xml", $uni:SIP)
+    return
+        fn:string-to-codepoints(fn:doc($stored/text())/container/smp)
+};
+
+declare
+    %test:assertEquals(172998)
 function uni:sip-dom-memtree-transform() {
     let $transformed := uni:transform($uni:SIP)
     return


exist-core/src/test/xquery/unicode.xqm
```diff
@@ -70,6 +70,14 @@ function uni:sip-dom-persistent() {
         fn:string-to-codepoints(fn:doc($stored)/container/smp)
 };
 
+declare
+    %test:assertEquals(172998)
+function uni:sip-dom-persistent-text() {
+    let $stored := xmldb:store("/db/" || $uni:COLLECTION_NAME, "sip.xml", $uni:SIP)
+    return
+        fn:string-to-codepoints(fn:doc($stored/text())/container/smp)
```
joewiz, line 78: The `/text()` should be removed. `$stored` returns an `xs:string`, so a `/text()` node test would raise an error.

joewiz (changes requested):

exist-core/src/test/xquery/unicode.xqm
duncdrum, line 78:
```suggestion
        fn:string-to-codepoints(fn:doc($stored)/container/smp)/text()
```

i think there is a syntax problem

wolfgangmm: The test should actually pass if you remove the `/text()`, at least I'm getting the expected codepoint when trying it in an isolated query.

joewiz: Same here - removing `text()` or moving it to the end of the path expression—i.e., `smp/text()`—allows the tests to pass:

```xml
<testsuite package="http://exist-db.org/xquery/test/unicode"
    timestamp="2021-01-04T14:34:33.964-05:00" tests="6" failures="0" errors="0" pending="0"
    time="PT0.01S">
    <testcase name="sip-dom-memtree" class="uni:sip-dom-memtree"/>
    <testcase name="sip-dom-memtree-transform" class="uni:sip-dom-memtree-transform"/>
    <testcase name="sip-dom-persistent" class="uni:sip-dom-persistent"/>
    <testcase name="sip-dom-persistent-text" class="uni:sip-dom-persistent-text"/>
    <testcase name="sip-dom-persistent-transform" class="uni:sip-dom-persistent-transform"/>
    <testcase name="sip-dom-persistent-transform-2" class="uni:sip-dom-persistent-transform-2"/>
</testsuite>
```

joewiz: Strictly speaking, the `text()` node test is not necessary, as the string value of the `<smp>` element will be atomized when it's passed to `fn:string-to-codepoints`. It's harmless, so unless this node test is critical to demonstrating a failure (and it's not, according to the results in @wolfgangmm and my tests above), it should be removed entirely.

duncdrum: yes the test doesn't seem to reproduce the original issue, I ll try to create tweak it to reproduce the code in #3651. As reported there the combination of `text()` selector on a permanent node is what is bugged (at least in some circumstances?)

joewiz:
This PR is adds duplicate of the existing `uni:sip-dom-persistent` test and doesn't demonstrate any error with eXist. @cwittern @duncdrum Is there another issue that should be demonstrated here, or can this be closed?

duncdrum:
@joewiz good to close, if we get a reproducible bug we can open a new PR with matching test

PR #3678, opened 2020-12-21
Improve backup-ui dialog flow

### LINKED ISSUE #3677: [BUG] NPE in Backup.java
reported by Alberto.

Point of discussion: should the backup "silently" continue or should we throw an eXist specific exception?

```
java.lang.NullPointerException
        at org.exist.backup.Backup.backup(Backup.java:415)
        at org.exist.backup.Backup.backup(Backup.java:304)
        at org.exist.backup.Backup.backup(Backup.java:278)
        at org.exist.backup.Main.process(Main.java:214)
        at org.exist.backup.Main.main(Main.java:448)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.exist.start.Main.invokeMain(Main.java:118)
        at org.exist.start.Main.runEx(Main.java:248)
        at org.exist.start.Main.run(Main.java:123)
        at org.exist.start.Main.main(Main.java:60)
```

image

dizzzz:
Handle NPE, fixes #3677
If user does not explicitly agree to delete backup folder/file, abort backup. (cannot find issue report)

image
image

image

image

potential open issue: position of new dialog boxes

### DIFF
--- a/exist-core/src/main/java/org/exist/backup/Backup.java
+++ b/exist-core/src/main/java/org/exist/backup/Backup.java
@@ -391,7 +391,7 @@ public class Backup {
                         default:
                             // Abort
                             dialog.dispose();
-                            JOptionPane.showMessageDialog(null,"Backup aborted", "Abort", JOptionPane.WARNING_MESSAGE);
+                            JOptionPane.showMessageDialog(null,"Backup aborted.", "Abort", JOptionPane.WARNING_MESSAGE);
                             throw new XMLDBException(ErrorCodes.INVALID_RESOURCE, msg);
                     }
 
--- a/exist-core/src/main/java/org/exist/backup/Backup.java
+++ b/exist-core/src/main/java/org/exist/backup/Backup.java
@@ -37,10 +37,7 @@ import org.exist.xquery.value.DateTimeValue;
 import org.xml.sax.SAXException;
 import org.xml.sax.helpers.AttributesImpl;
 import org.xmldb.api.DatabaseManager;
-import org.xmldb.api.base.Collection;
-import org.xmldb.api.base.Database;
-import org.xmldb.api.base.Resource;
-import org.xmldb.api.base.XMLDBException;
+import org.xmldb.api.base.*;
 import org.xmldb.api.modules.XMLResource;
 
 import javax.annotation.Nullable;
@@ -371,11 +368,33 @@ public class Backup {
                     //Skipping resources[i]
                     continue;
                 }
+
                 resource = current.getResource(resources[i]);
 
                 if (dialog != null) {
                     dialog.setResource(resources[i]);
                     dialog.setProgress(i);
+                }
+
+                // Avoid NPE
+                if (resource == null) {
+                    final String msg = "Resource " + resources[i] + " could not be found.";
+                    System.err.println(msg);
+                    Object[] options = {"Ignore", "Abort"};
+                    int n = JOptionPane.showOptionDialog(null, msg,"Backup Error",
+                            JOptionPane.YES_NO_CANCEL_OPTION, JOptionPane.QUESTION_MESSAGE, null,
+                            options, options[1]);
+                    switch(n) {
+                        case 0:
+                            // ignore one
+                            continue;
+                        default:
+                            // Abort
+                            dialog.dispose();
+                            JOptionPane.showMessageDialog(null,"Backup aborted", "Abort", JOptionPane.WARNING_MESSAGE);
+                            throw new XMLDBException(ErrorCodes.INVALID_RESOURCE, msg);
+                    }
+
                 }
 
                 final String name = resources[i];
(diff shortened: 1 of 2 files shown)


line-o (commented):
LGTM 👍 (don't know enough about the backup process to approve)

exist-core/src/main/java/org/exist/backup/Backup.java
```diff
@@ -371,13 +369,33 @@ private void backup(final Set<String> seenBlobIds, final Collection current, fin
...
+
                 resource = current.getResource(resources[i]);
 
                 if (dialog != null) {
                     dialog.setResource(resources[i]);
                     dialog.setProgress(i);
                 }
 
+                // Avoid NPE
+                if (resource == null) {
```
adamretter, line 381: So `resources` is build from `EXistCollection#listResources`.

When calling a subsequent `EXistCollection#getResource` on one of those resource names, what does a return value of `null` actually indicate? I suspect:
1. The resource has been deleted from the Collection because it was not locked during the backup process. AFAIK backup over XML:DB is best effort and not thread/concurrent safe.
2. The database has been corrupted in some way, and the resource can no longer be retrieved.

I think the warning code below should take two different paths for alerting the user depending on whether `dialog == null`. This process can run in either GUI or console mode IIRC.

exist-core/src/main/java/org/exist/backup/Backup.java
```diff
@@ -371,13 +369,33 @@ private void backup(final Set<String> seenBlobIds, final Collection current, fin
...
+                    int n = JOptionPane.showOptionDialog(null, msg,"Backup Error",
+                            JOptionPane.YES_NO_CANCEL_OPTION, JOptionPane.QUESTION_MESSAGE, null,
+                            options, options[1]);
+                    if (n == 0) {
+                        // ignore one
+                        continue;
+                    }
+
+                    // Abort
+                    dialog.dispose();
```
adamretter, line 394: This will raise an NPE if `dialog` is `null`

exist-core/src/main/java/org/exist/backup/Backup.java
```diff
@@ -371,13 +369,33 @@ private void backup(final Set<String> seenBlobIds, final Collection current, fin
...
                     dialog.setResource(resources[i]);
                     dialog.setProgress(i);
                 }
 
+                // Avoid NPE
+                if (resource == null) {
+                    final String msg = "Resource " + resources[i] + " could not be found.";
+                    System.err.println(msg);
+                    Object[] options = {"Ignore", "Abort"};
+                    int n = JOptionPane.showOptionDialog(null, msg,"Backup Error",
```
adamretter, line 385: JOptionPane won't work if the Backup is run in console mode instead of GUI mode.

exist-core/src/main/java/org/exist/backup/Backup.java
```diff
@@ -371,13 +369,33 @@ private void backup(final Set<String> seenBlobIds, final Collection current, fin
...
+                            JOptionPane.YES_NO_CANCEL_OPTION, JOptionPane.QUESTION_MESSAGE, null,
+                            options, options[1]);
+                    if (n == 0) {
```
adamretter, line 388: Better to use `if (n == JOptionPane.YES_OPTION)` ?

adamretter (changes requested):

exist-core/src/main/java/org/exist/backup/Backup.java
```diff
@@ -269,7 +267,7 @@ public void backup(final boolean guiMode, final JFrame parent) throws XMLDBExcep
 
                         try {
                             wait(20);
-                        } catch (final InterruptedException e) {
+                        } catch (final InterruptedException ignored) {
```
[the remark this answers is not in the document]
dizzzz: this has always been this way, this is a simple rename

exist-core/src/main/java/org/exist/backup/Backup.java
dizzzz: great feedback, thnx

exist-core/src/main/java/org/exist/backup/Backup.java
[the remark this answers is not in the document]
adamretter: Ah okay, perhaps you could add this to the `catch`:

```
// NOTE: must set interrupted flag
Thread.currentThread().interrupt();
e.PrintStackTrace();
throw new IOException(e.getMessage(), e);
```

wolfgangmm (approved):
Looks good to me now.

Issue #1463, opened 2017-06-23
Element constructor creates document node. It should not.


mjhallman:
```xquery
let $x := element {"root"}  {}

return $x/root() instance of document-node()
```
This returns `true` in eXist-db which is incorrect. The result of `$x/root()` should be an `element()`.
I have tested this on Saxon and BaseX which both correctly return `false`.

dizzzz:
please fill in the issue-template. It makes sure you'll add all required information, info we need to understand the exact context. thnx

adamretter:
@dizzzz I helped @mjhallman report these. Everything that is needed to reproduce is there.

adamretter:
It's the `HEAD` of the `develop` branch.

adamretter:
A little investigation has shown that the problem is actually `fn:root`.  In-memory elements in eXist-db always have an Owner Document, which they should not when directly constructed with `element x {}`, and so `fn:root` always returns the `document-node()` by mistake.

See the examples here: https://www.w3.org/TR/xpath-functions/#func-root-examples

The following XQUnits show the issue:

```xquery
declare
    %test:assertEquals("false")
function root:in-mem-element() {
    fn:root(<anything/>) instance of document-node()
};
```
The above should return `false`, but eXist-db returns `true`.

```xquery
declare
    %test:assertEquals("true")
function root:in-mem-document() {
    fn:root(document { <anything/> }) instance of document-node()
};
```

The above should (and does in eXist-db) return `true`.

adamretter:
@wolfgangmm I would welcome a discussion about this. From my perspective it requires quite some re-engineering of the in-memory DOM model, but maybe you have some clever ideas to fix this up?

wolfgangmm:
@adamretter we know where the in-memory DOM is created by a constructor, so could just set a flag.

duncdrum:
Yes someone working on tei-publisher just stumbled on `fn:root()` 
```
xquery version "3.0";
declare function local:node-kind
  ( $nodes as node()* )  as xs:string* {

 for $node in $nodes
 return
 if ($node instance of element()) then 'element'
 else if ($node instance of attribute()) then 'attribute'
 else if ($node instance of text()) then 'text'
 else if ($node instance of document-node()) then 'document-node'
 else if ($node instance of comment()) then 'comment'
 else if ($node instance of processing-instruction())
         then 'processing-instruction'
 else 'unknown'
 } ;

let $data :=
<root xml:base="bl">
    <dummy/>
[...]

PR #3690, opened 2021-01-01
[feature] add fn:path

dizzzz:
Implementation of fn:path
https://www.w3.org/TR/xpath-functions-31/#func-path 

image


dizzzz:
Blocked by https://github.com/eXist-db/exist/issues/1463

exist-core/src/main/java/org/exist/dom/memtree/DocumentImpl.java
```diff
@@ -302,6 +302,13 @@ public void replaceReferenceNode(final int nodeNum, final CharSequence ch) {
         addChars(nodeNum, ch);
     }
 
+    /**
+     * @return TRUE when the Document is constructed via document {}.
+     */
+    public boolean isExplicitlyCreated(){
+        return explicitlyCreated;
```
adamretter, line 309: Hmmm.... Should this also determine the type returned from `#getNodeType()`. It would then be cheaper to call `Node#getNodeType()` than `node instance of`

dizzzz: pooh..... lets discuss with @wolfgangmm ; I am open for improvements :-)

dizzzz: @adamretter the document check is done Max 1 per call, so it will not be too expensive.
I tried to improve but it is (AFAICsee) impossible. And I do  now want spend a 4th day on this simple function :-)

When #1463 is implemented, we can reconsider.

For me this PR is good to go.

dizzzz: checking S.O.: https://stackoverflow.com/questions/103564/the-performance-impact-of-using-instanceof-in-java 
performance is not an issue anymore in  modern JDKs.

dizzzz:
Ready for final review

dizzzz:
to test: stored document

dizzzz:
During last check, I discovered that the function does NOT WORK correctly with queries on persistent documents. There is no way now to distinguish between a stored document and a constructed element (checked with  `isDocumentNodeCreatedAsInMemDocument(node)()`.

To be able to finish this PR, #1463 needs to be fixed first,

```xml
<testcase name="comments_persistent" class="path:comments_persistent">
            <failure message="assertEquals failed." type="failure-error-code-1">/Q{}div[1]/comment()[3]</failure>
            <output>Q{http://www.w3.org/2005/xpath-functions}root()/comment()[3]</output>
</testcase>
```

line-o:
I am happy to help @dizzzz 

In the meantime I created an implementation in XQuery:

https://gist.github.com/line-o/f3683052951fb9eec28651638f779720#file-xpaths-xq

line-o:
implemented in b9ebedd9630aaafd3218ffb390de5114429d7c33

PR #3701, opened 2021-01-11
Remove legacy json utilities

### LINKED ISSUE #688: xml-to-json does not escape properly
import module namespace json = "http://www.json.org";

let $jrData := json:xml-to-json($result)

<result>
<Description>
<title>Alla turca </title>
<type>Sachbücher</type>
<date>1951</date>
<description>Sonate KV 331 mit dem 3. Satz "Alla turca".</description>
</Description>
</result>

<jrData>
{"title" : "Alla turca ", "type" : "Sachbücher", "date" : "1951",
"description" : "Sonate KV 331 mit dem 3. Satz "Alla turca"."}
</jrData>

 As one can see, the function xml-to-json ought to have escaped the double quotes in the data with a backslash.

joewiz:
### Description:

Remove json.xq and jsonp.xq - legacy json utilities

### Reference:

Closes #688. From that issue:

> @wolfgangmm: The json:xml-to-json function is a bit outdated (I wonder if we should remove the module from the next release). I would prefer to use the json serializer...
> 
> @adamretter: I agree, can we remove those old modules?
>
> @wolfgangmm: Yes, we should remove json.xq and jsonp.xq.

### Type of tests:

n/a


dizzzz (approved):

adamretter:
I think this should likely be a 6.0.0 feature

joewiz:
I've extracted the general conf.xml fixes to https://github.com/eXist-db/exist/pull/3703, which needn't wait for the next major version.

PR #3705, opened 2021-01-14
Broker pool service api

adamretter:
Improves the API of BrokerPool Service to add additional life-cycle functions for starting and stopping Broker Pool services.

1. `BrokerPoolService#startPreSystem`
Permits activities that need to happen before the database enters system (single-user) mode during startup.

2. `BrokerPoolService#stopMultiUser`
Permits activities that need to happen before the database exits multi-user mode during shutdown.

NOTE: Even though this is an internal API, I have taken care to ensure that it is backwards compatible.

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/BrokerPool.java
+++ b/exist-core/src/main/java/org/exist/storage/BrokerPool.java
@@ -521,6 +521,12 @@ public class BrokerPool extends BrokerPools implements BrokerPoolConstants, Data
 
         this.startupTriggersManager = servicesManager.register(new StartupTriggersManager());
 
+        // this is just used for unit tests
+        final BrokerPoolService testBrokerPoolService = (BrokerPoolService) conf.getProperty("exist.testBrokerPoolService");
+        if (testBrokerPoolService != null) {
+            servicesManager.register(testBrokerPoolService);
+        }
+
         //configure the registered services
         try {
             servicesManager.configureServices(conf);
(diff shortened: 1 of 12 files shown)


exist-core/src/main/java/org/exist/storage/BrokerPool.java
```diff
@@ -1547,7 +1519,7 @@ public void triggerSync(final Sync syncEvent) {
     //TOUNDERSTAND (pb) : synchronized, so... "schedules" or, rather, "executes" ?
     public void triggerSystemTask(final SystemTask task) {
         final State s = status.getCurrentState();
-        if(s == State.SHUTTING_DOWN) {
+        if(s == State.SHUTTING_DOWN_MULTI_USER_MODE || s == State.SHUTTING_DOWN_SYSTEM_MODE) {
```
dizzzz, line 1522: ` isShuttingDown()`

adamretter: I need the state to be consistent, so we just get the state once and use it here and below to check if the database is shutdown.

If I called `isShuttingDown()` it would make a second call to `status.getCurrentState` and that would not be consistent with `s`.

dizzzz: but... it has the same code?
image

dizzzz: ah sorry Now I see it

dizzzz (approved):

PR #3715, opened 2021-01-22
Remove jQuery

adamretter:
eXist-db shipped with jQuery 1.7.1, however it appeared only to be used for a simple "fade-in". This can now be replaced with modern CSS 3.

This was undertaken because jquery version 1.7.1, and all newer versions apart from latest two versions (3.5.0 and 3.5.1) have known Security Vulnerabilities - https://snyk.io/vuln/npm:jquery

At the same time, I also made sure the HTML page is both valid HTML 5 and XML, and fixed an issue with font-loading and some smaller markup problems.


dizzzz (approved):

adamretter:
Thanks @dizzzz  :-)

PR #3725, opened 2021-02-04
[test] extend tests for immutability of maps 

### LINKED ISSUE #3724: [BUG] Side effects when calling map functions
**Describe the bug**

In some cases it is possible that the value of a variable bound to a map is mutated. 

**Expected behavior**

In XQuery, variables are immutable. Once a variable is bound, its value cannot change within the scope of its FLWOR expression and binding sequence. See https://www.w3.org/TR/xquery-31/#id-binding-rules.

In the example below, the variable `$three-entry-map` is bound to a map containing three entries:

```xquery
map {
    "a": 1,
    "b": 2,
    "c": 3
}
```

The query returns this variable—this 3-entry map—and it does so in BaseX and Saxon. 

However, in eXist, the presence of a map-altering function like `map:remove()`, `map:put()`, or `map:merge()` in a subsequent `let` clause somehow mutates the `$three-entry-map` variable. As a result, in eXist, the example queries all returns the following map, with only 2 entries instead of the expected 3:

```xquery
map {
    "a": 1,
    "c": 3
}
``` 

This violates immutability. 

**To Reproduce**

The problem has been reduced to the following:

```xquery
xquery version "3.1";

let $three-entry-map := 
    map:merge((
        map {
            "a" : 1,
            "b" : 2,
            "c" : 3
        }
    ))
let $some-other-map := map:remove($three-entry-map, "b")
return 
    $three-entry-map
```

In BaseX and Saxon, this returns the expected 3-entry map. In eXist, it returns a 2-entry map.

Similarly, the presence of `map:put()` will trigger the problem in eXist:

```xquery
xquery version "3.1";

let $three-entry-map := 
    map:merge((
        map {
            "a" : 1,
            "b" : 2,
            "c" : 3
        }
    ))
let $some-other-map := map:put($three-entry-map, "b", "foo")
return 
    $three-entry-map
```

Lastly, `map:merge()` will trigger it too:

```xquery
xquery version "3.1";

let $three-entry-map := 
    map:merge((
        map {
            "a" : 1,
            "b" : 2,
            "c" : 3
        }
    ))
let $some-other-map := 
    map:merge((
        $three-entry-map, 
        map { "b": "foo" }
    ))
return 
    $three-entry-map
```

In all 3 cases, in BaseX and Saxon, this returns the expected 3-entry map. In eXist, it returns a 2-entry map.

Even this case without a `let` clause reveals the problem - returning an empty map instead of a 1-entry map:

```xquery
xquery version "3.1";

map:merge( map { "a" : 1 } ) ! (., map:put(., "a", 2))[1]
```

 - OS: linux, windows, macOS
 - eXist-db version: 5.3.0-SNAPSHOT 303ca0db511ee93bac1b50ccede357fde9658e7b
 - Java Version 1.8.0_271

line-o:
### Description:

This PR now only adds tests that extend the previous test for the immutability of maps

~~fork maps before mutating them~~

### Reference:

refs #3724

Map implementation (bifurcan) documentation of the `forked()` method
https://lacuna.io/docs/bifurcan/io/lacuna/bifurcan/IMap.html#forked--

### Type of tests:

Extended XQsuite tests in exist-core/src/test/xquery/maps/maps.xql

### DIFF
--- a/exist-core/src/test/xquery/maps/maps.xql
+++ b/exist-core/src/test/xquery/maps/maps.xql
@@ -644,93 +644,92 @@ declare
     %test:assertTrue
 function mt:immutable-put-then-put() {
     let $extended := map:put(mt:create-test-map(), $mt:test-key-two, false())
-    let $expected := $extended?($mt:test-key-one)
+    let $expected := $extended($mt:test-key-one)
     let $result := map:put($extended, $mt:test-key-one, false())
-
-    return $expected eq $extended?($mt:test-key-one)
+    return $expected eq $extended($mt:test-key-one)
 };
 
 declare
     %test:assertTrue
 function mt:immutable-put-then-remove() {
     let $extended := map:put(mt:create-test-map(), $mt:test-key-two, false())
-    let $expected := $extended?($mt:test-key-one)
+    let $expected := $extended($mt:test-key-one)
     let $result := map:remove($extended, $mt:test-key-one)
 
-    return $expected eq $extended?($mt:test-key-one)
+    return $expected eq $extended($mt:test-key-one)
 };
 
 declare
     %test:assertTrue
 function mt:immutable-put-then-merge() {
     let $extended := map:put(mt:create-test-map(), $mt:test-key-two, false())
-    let $expected := $extended?($mt:test-key-one)
+    let $expected := $extended($mt:test-key-one)
     let $result := map:merge(($extended, map { $mt:test-key-one : false() }))
 
-    return $expected eq $extended?($mt:test-key-one)
+    return $expected eq $extended($mt:test-key-one)
 };
 
 declare
     %test:assertTrue
 function mt:immutable-remove-then-put() {
     let $removed := map:remove(mt:create-test-map(), $mt:test-key-two)
-    let $expected := $removed?($mt:test-key-one)
+    let $expected := $removed($mt:test-key-one)
 
     let $result := map:put($removed, $mt:test-key-one, false())
 
-    return $expected eq $removed?($mt:test-key-one)
+    return $expected eq $removed($mt:test-key-one)
 };
 
 declare
     %test:assertTrue
 function mt:immutable-remove-then-remove() {
     let $removed := map:remove(mt:create-test-map(), $mt:test-key-two)
-    let $expected := $removed?($mt:test-key-one)
+    let $expected := $removed($mt:test-key-one)
 
     let $result := map:put($removed, $mt:test-key-one, false())
 
-    return $expected eq $removed?($mt:test-key-one)
+    return $expected eq $removed($mt:test-key-one)
 };
 
 declare
     %test:assertTrue
 function mt:immutable-remove-then-merge() {
     let $removed := map:remove(mt:create-test-map(), $mt:test-key-two)
-    let $expected := $removed?($mt:test-key-one)
+    let $expected := $removed($mt:test-key-one)
     let $result := map:merge(($removed, map { $mt:test-key-one : false() }))
 
-    return $expected eq $removed?($mt:test-key-one)
+    return $expected eq $removed($mt:test-key-one)
 };
 
 declare
     %test:assertTrue
 function mt:immutable-merge-then-put() {
     let $merged := map:merge(mt:create-test-map())
-    let $expected := $merged?($mt:test-key-one)
+    let $expected := $merged($mt:test-key-one)
 
     let $result := map:put($merged, $mt:test-key-one, false())
 
-    return $expected eq $merged?($mt:test-key-one)
+    return $expected eq $merged($mt:test-key-one)
 };
 
 declare
     %test:assertTrue
 function mt:immutable-merge-then-remove() {
     let $merged := map:merge(mt:create-test-map())
-    let $expected := $merged?($mt:test-key-one)
+    let $expected := $merged($mt:test-key-one)
 
     let $result := map:remove($merged, $mt:test-key-one)
 
-    return $expected eq $merged?($mt:test-key-one)
+    return $expected eq $merged($mt:test-key-one)
 };
 
 declare
     %test:assertTrue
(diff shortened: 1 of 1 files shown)


line-o:
So, in `SingleKeyMapType#put` a different method is used to create the copy of the map.
`return new MapType(context, map.forked(), keyType);`
I have not enough knowledge yet which one is the better in the instances changed here.

`.forked()` or `.clone()` - this will likely have effects on memory usage and maybe performance.

line-o:
I do not understand the underlying implementation of "forked", though:
```java
    default IMap<K, V> forked() {
        return this;
    }
```

line-o:
using `.forked()` also works, so we can freely choose.

line-o:
Ready to be merged from my point of view. Tests fail until the fix is applied.

dizzzz (approved):

joewiz (approved):
I can confirm that when building eXist from this PR's branch, the tests that @PieterLamers and I developed in https://github.com/eXist-db/exist/issues/3724 all return the expected results.

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
```diff
@@ -156,7 +156,7 @@ public void add(final AbstractMapType other) {
     public AbstractMapType merge(final Iterable<AbstractMapType> others) {
 
         // create a transient map
-        IMap<AtomicValue, Sequence> newMap = map.linear();
+        IMap<AtomicValue, Sequence> newMap = map.forked();
```
dizzzz, line 159: ah ok. Being not at home, I could not see the functionality behind `linear()`

line-o:
@joewiz great!

Do you think the test cases I added are covering every combination we need to check?
https://github.com/eXist-db/exist/pull/3725/files#diff-6cbf91fe2ed15b2f66e0c96ba43b21e94f3be9075de054b215a1cb0b5a88cfd7

joewiz:
@line-o Yes, I think you've got the ones in the issue covered with mt:immutable-put-after-merge, mt:immutable-remove-after-merge, and mt:immutable-merge-after-merge, and the other variations were a great idea too.

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
adamretter: I don't understand why forked is now used here, as `forked` is called below.

line-o: That is too late, as the tests clearly prove. But maybe the second call to forked can be removed or changed to `.linear()`?

line-o: The second call will likely not do much the documentation states:
"For already forked maps the call to .forked() will just return itself".

adamretter:
@line-o I just built the HEAD of Bifurcan and checked against that too. I wanted to do so, as it has some fixes that are not released yet. Unfortunately there was no change, still 6 failing tests. I will dig a little deeper...

adamretter:
@line-o I just expanded on some of your tests, and it caused me to find another bug in maps. The XQuery F+O specs says that default behaviour for `map:merge` when there are duplicates keys is to `use-first`, however eXist-db does `use-last` and some of your tests were relying on that - I will fix that too...

dizzzz:
Is there any new info on this?

dizzzz:
what is the state of the PR?  @line-o  @adamretter 
I marked it as draft to prevent accidental pulling in

line-o:
I am waiting for this to either a) be merged as it is or for a review that tells what needs to change (I expect this to be "remove the surplus calls to forked()")
What I will definitely do is fix the minor issue in one the tests.

line-o:
I can confirm that #3724 is fixed by the new release of bifurcan.

joewiz:
@line-o That's great news! Thank you for all of your efforts on the maps issue, Juri! When you have the chance, could you please approve #3886?

line-o:
Oh wait. I was under the impression that this PR was already applied!
This needs to be retested.

line-o:
I pulled the PR with bifurcan pre6 built exist and re-run the tests. It appears to not been fixed with that.

adamretter:
@line-o yes I know. I did the same over the weekend. I'm still working on it...

line-o:
I would appreciate you let me finish this here. Or at least argue why this is not the way to go instead.

line-o:
Rebase and with improved test cases.
I still see odd behaviour and fear that there is a) still issues in the map implementation or b) issues in how this library is used.

adamretter:
This is no longer neeed. The bug was indeed in Bifurcan and was fixed as https://github.com/lacuna/bifurcan/issues/35#issuecomment-841802782 and upgraded in eXist-db as https://github.com/eXist-db/exist/pull/3886

The tests from this PR have been incorporated and expanded upon in https://github.com/eXist-db/exist/pull/3738

@line-o are you happy to close this one now?

line-o:
@adamretter I actually re-incorporated your tests here and added a final improvement to them. What about we make this PR the one adding the tests and #3738 can then be solely about `map:merge#2`.
That is also nicer later when we compile the changelog.

line-o:
I will take responsibility in changing the PR description and title as well as the commit history assuming you do so for #3738 as well.

adamretter:
@line-o I don't think that makes sense, you would also need to remove all your Java changes in this PR.
As they were an early attempt to fix the Maps problem that was reported in eXist-db's code, but the real problem wasn't in eXist-db's code... it was in Bifurcan which I worked on diagnose and get it resolved at its source. Wouldn't this PR then solely end up just being tests most of which were created in my other PR?

If you want my PR split into two, with and without the change to merge-order to match the spec, I have previously already offered to do that...

line-o:
Of course I will remove the no longer needed commits.

line-o:
I would also rather squash all the back and forth in the tests into a single commit.

Issue #3650, opened 2020-12-04
User-group relations not restored w/o restart


mgraauw:
**Describe the bug**
When restoring a backup (from eXist 2.2, but most likely applies to all backups) to eXist 5.2.0, users and groups are restored, but the user-group relations only after exist restart.

Example: my backup has a group `lab`, and a user `marc`, who is a member of group `lab`.
* After clean install of 5.2.0, I have the usual users (`admin`, `guest`, **NO** `marc`).
* After restoring, I see user `marc' and group `lab`, but `marc` is only a member of the group `nogroup`.
    * Same results in:
        *  User Manager
        * Java Admin Client,
        * XQuery via: `sm:get-user-groups('marc')`

After stop and start of eXist, `marc` is member of group `lab`. This led me to believe the restore had failed and retry several times until I found a restart solved things. Apparently user-group relations aren't refreshed properly after restore.

**Expected behaviour**
user-group membership should be available right after restore.

**Context (please always complete the following information):**
 - OS: Win 10
 - eXist-db version: 5.2.0
 - Java Version JRE 1.8.0_211

**Additional context**
Installing with jar installer

adamretter:
I can confirm that this is an issue on versions 5.1.1 and 5.3.0-SNAPSHOT.

PR #3731, opened 2021-02-11
Group membership is not correct immediately after restoring a backup

adamretter:
Closes https://github.com/eXist-db/exist/issues/3650

It absolutely blows my mind that until today only 2 people have recently reported this but yet the problem has been around for quite some time (at least 5.1.1 if not before). I wonder if anyone else is doing backups/restores of their databases?!?

---

This PR provides a fix that **only** addresses the problem for those users that restore full or incremental backups of the entire database, i.e. starting from `/db`.

Any partial restore that starts from `/db/system` or below and includes the users and groups will still not correctly load the groups before the users. Unfortunately solving such a thing with the current approach of the restore code is impossible, and the restore system itself needs rewriting to take a different approach. As most users backup and restore full-databases this should solve it for the majority.

We may contribute a rewrite of the backup/restore system in future to fix a number of additional issues.

### DIFF
--- a/exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java
+++ b/exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java
@@ -37,6 +37,9 @@ import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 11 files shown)


dizzzz:
I typically restore a database, localmode @  commandline, so a db startup is done anyway :-/

exist-core/src/main/java/org/exist/backup/ZipArchiveBackupDescriptor.java
```diff
@@ -149,6 +155,30 @@ public BackupDescriptor getBackupDescriptor(String describedItem) {
         return bd;
     }
 
+    @Override
+    public List<BackupDescriptor> getChildBackupDescriptors() {
+        final Pattern ptnDescriptor = Pattern.compile("(" + base + "/[^/]+/" + ")" + BackupDescriptor.COLLECTION_DESCRIPTOR);
+        final Matcher mtcDescriptor = ptnDescriptor.matcher("");
+        try (final Stream<BackupDescriptor> entries = archive.stream()
```
dizzzz, line 162: Would it make sense to add some logging for this step (assuming this could consume some time)

exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java
```diff
@@ -73,6 +76,29 @@ public BackupDescriptor getChildBackupDescriptor(final String describedItem) {
         return bd;
     }
 
+    @Override
+    public List<BackupDescriptor> getChildBackupDescriptors() {
+        try {
+            try (final Stream<BackupDescriptor> entries = Files.list(descriptor.getParent())
```
dizzzz, line 82: Would it make sense to add some logging for this step (assuming this could consume some time)

dizzzz (approved):

exist-core/src/main/java/org/exist/backup/ZipArchiveBackupDescriptor.java
adamretter: I added as requested :-)

exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java
adamretter: I added as requested :-)

adamretter:
@dizzzz I added the logging as you requested. Let me know... ;-)

PR #3738, opened 2021-02-18
Switch map:merge from use-last to use-first as per XQuery spec

### LINKED ISSUE #3750: [BUG] behaviour of map:merge#1 is not standards compliant
**Describe the bug**

@adamretter found out that the handling of duplicate keys when merging maps  with `map:merge` is not according to the specification.
Exists uses `use-last` whereas the XQuery 3.1 specification states `use-first` must be used.

**Expected behavior**

Existdb's implementation of `map:merge#1` to adhere to the specification.

**To Reproduce**

```Xquery
xquery version "3.1";

declare
    %test:assertTrue
function t:test() {
    let $m1 := map{'a':1}
    let $m2 := map{'a':2}
    let $merged := map:merge(($m1, $m2))
    return $merged?a eq $m1?a
};
```

eXist-db version: 5.3.0-SNAPSHOT 
likely all implementations of map:merge are affected

adamretter:
Closes #3750

Fixes `map:merge` in eXist-db so that It implements `use-first` as the default behaviour for as per the XQuery spec. Before this change eXist-db implemented `use-last` as the default.


joewiz:
Sounds like a great plan.

Looking ahead to when `map:merge#2` is added, we can check it off from https://github.com/eXist-db/exist/issues/3635 and will need to remove it from the unsupported functions in the XQuery article at https://github.com/eXist-db/documentation/blob/master/src/main/xar-resources/data/xquery/xquery.xml#L147.

line-o:
I do consider the switch from implicit use-last to use-first a breaking change.

adamretter:
> I do consider the switch from implicit use-last to use-first a breaking change.

It's a bugfix, but I agree it is also likely a breaking change for people who have relied on the non-spec compliant behaviour that eXist-db previously implemented.

line-o:
I also would say that as long as this PR is not fixing #3724 it does _not_ supersede #3725

line-o:
I will incorporate the improved tests up until the breaking changes.

line-o:
Saxon has changed the behaviour of `map:merge` quite recently 9.7.0.7 -> 9.7.0.8  https://www.saxonica.com/documentation/index.html#!functions/map/merge

adamretter:
I opened the issue report with Bifurcan - https://github.com/lacuna/bifurcan/issues/35

adamretter:
@joewiz @line-o this is now done.
We can decide in the next week or two whether to split this in two for a 5.x.x and a 6.0.0, or just go with it for a 6.0.0.

joewiz:
This PR will be needed to make use of Mary Holstege's Aho-Corasick module https://www.balisage.net/Proceedings/vol26/html/Holstege01/BalisageVol26-Holstege01.html. Installing the `art/art_no_data.zip` version from https://mathling.com/code/ into `/db/apps/art` and running `/db/apps/art/tests/test-aho.xqy` returns this error: 

> Cannot compile xquery: err:XPST0017 error found while loading module aho: Error while loading module ../string/aho.xqy: error found while loading module util: Error while loading module ../core/utilities.xqy: Unexpectedly received 2 parameter(s) in call to function 'map:merge()'. Defined function signatures are: map:merge($maps as map(*)*) map(*) [at line 891, column 3]

dizzzz (approved):

dizzzz:
wait for 6.x

joewiz (approved):
LGTM

line-o:
This PR adds more than its title and description give away and only the switch to "use-first" by default is breaking.
I would like this PR to be broken up in two parts:

1. every non-breaking change: support of "reject" and "combine"
2. the breaking change that switches the default behaviour

Doing this would allow us to merge in the enhancements earlier (prior to exist7) giving users more time to adapt.

line-o:
the only commit that is breaking is 8d7eb001c62badf468a6fe6be4d05efe37283bc6

line-o:
#4621 now contains the implementation of options "combine" and "reject" (squashed in to one). When it is merged the last three commits here should be removed and this PR rebased on current develop.

line-o (commented):
We should take extra care that all existing XQuery code that makes use of map:merge#1 is checked. XQSuite is one example but I found many more instances.
This is to not break existing functionality that is hard to trace down for users.

PR #3751, opened 2021-02-27
Improve CI in GitHub Actions

adamretter:
* Remove SonarQloud integration that isn't working
* Separate the Build and Test steps
* Add Javadoc verification step
* Fix existing broken Javadoc
* Add JDK 15
* Update the badges in README.md


.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+    branches:
+      - develop
+      - master
+  pull_request:
+
+jobs:
+  build:
+    name: Build and Test (${{ matrix.os }} / OpenJDK ${{ matrix.jdk }})
+    strategy:
+      fail-fast: true
```
dizzzz, line 14: fail fast.... does this mean that if the Java8 build fails, java11 is not started anymore?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - develop
+      - master
```
dizzzz, line 7: master makes sense

.github/workflows/ci.yml
adamretter: The first build that fails causes the rest to be cancelled or not started. The idea is that it saves build-minutes

dizzzz: Ok, but it would make comparing eg Java8 and Java11 results a bit more right? What if Java8 is built first, which fails whilst Java11 is just OK...

dizzzz:
@adamretter are you sure the PR is ok? I ask as the checks overview is very short: 
image

.github/workflows/ci.yml
```diff
@@ -1,11 +1,12 @@
 name: CI
+
 on:
   push:
     branches:
       - develop
       - master
   pull_request:
-    types: [opened, synchronize, reopened]
+
```
line-o, line 9: It looks like types need to be listed here, otherwise pull requests are not run.

.github/workflows/ci.yml
```diff
@@ -18,26 +18,17 @@ jobs:
...
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
-      - name: Build and analyze
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-        run: mvn -B verify
+      - name: Maven Build
+        run: mvn -V -B -DskipTests=true "-Dmaven.javadoc.skip=true" install
```
line-o, line 32: Are the tickles around `-Dmaven.javadoc.skip=true` necessary?

.github/workflows/ci.yml
```diff
@@ -18,26 +18,17 @@ jobs:
...
+      - name: Maven Test
+        run: mvn -V -B verify -Ptests
```
line-o, line 34: What does this maven call do?

.github/workflows/ci.yml
adamretter: Not according to the docs:

> Note: By default, a workflow only runs when a pull_request's activity type is opened, synchronize, or reopened. To trigger workflows for more activity types, use the types keyword.

I think I have also seen this in other projects

.github/workflows/ci.yml
adamretter: The Windows platforms run in Powershell, and these are necessary to cope with that monstrosity.

.github/workflows/ci.yml
adamretter: https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#a-build-lifecycle-is-made-up-of-phases

.github/workflows/ci.yml
adamretter: Then the build will fail. It fails on the first failure. It's a trade-off between saying "any fail is a fail" and I want it to cost less to use CI, or I am willing to pay more and I want to see all results.

dizzzz (approved):

.github/workflows/ci.yml
adamretter: I have added this back in.

dizzzz:
hmmm no change ?

PR #3761, opened 2021-03-02
[fix] fn:position() off-by-one after any predicate 

### LINKED ISSUE #3737: position() function off-by-one error
Working with eXist 5.2, the position function returns what appears to be a 0-based index instead of 1-based as required:

```xquery
let $a := <a>
  <b n="a"/>
  <b n="b"/>
  <b n="c"/>
</a>
return $a/b[@n=('b', 'c')]/position()
```
returns
```
0
1
``` 
instead of the expected
```
1
2
```

**To Reproduce**
The follow xQuery test tests for the proper results.

```xquery
xquery version "3.1";

<a>
  <b n="a"/>
  <b n="b"/>
  <b n="c"/>
</a>
};

declare
    %test:assertTrue
function t:test() {
    let $r := $t:XML/a/b[@n=('b', 'c')]/position()
    return count($r) eq 2 and $r[1] eq 1 and $r[2] eq 2
};
```

 - OS: Mac OS 11.1, RHEL 7, CentOS 7
 - eXist-db version: 5.2.0
 - Java Version: 11.0.8 and others

**Additional context**
- How is eXist-db installed? JAR installer

line-o:
### Description:

```query
<a><b/></a>/b[true()]/position()
```
should return 1

With this PR applied it does.

### Reference:

fixes #3737

### Type of tests:

XQSuite tests in fn-position.xqm

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/LocationStep.java
+++ b/exist-core/src/main/java/org/exist/xquery/LocationStep.java
@@ -227,7 +227,7 @@ public class LocationStep extends Step {
             result = pred.evalPredicate(outerSequence, result, axis);
             // subsequent predicates operate on the result of the previous one
             outerSequence = null;
-            context.setContextSequencePosition(-1, null);
+            context.setContextSequencePosition(0, null);
         }
         return result;
     }
--- a/exist-core/src/test/xquery/fn-position.xqm
+++ b/exist-core/src/test/xquery/fn-position.xqm
@@ -1,0 +1,85 @@
+xquery version "3.1";
+
+(: https://github.com/eXist-db/exist/issues/3737 :)
+module namespace fn-pos="http://exist-db.org/xquery/test/fn-position";
+
+
+declare namespace test="http://exist-db.org/xquery/xqsuite";
+
+
+declare variable $fn-pos:collection := 'test-fn-pos';
+declare variable $fn-pos:doc := 'stored-nodes.xml';
+declare variable $fn-pos:nodes := document {
+    <a>
+      <b n="a"/>
+      <b n="b"/>
+      <b n="c"/>
+    </a>
+};
+
+declare
+    %test:setUp
+function fn-pos:setup () {
+    xmldb:create-collection('/db', $fn-pos:collection),
+    xmldb:store('/db/' || $fn-pos:collection, $fn-pos:doc, $fn-pos:nodes)
+};
+
+declare
+    %test:tearDown
+function fn-pos:teardown () {
+    xmldb:remove('/db/' || $fn-pos:collection, $fn-pos:doc)
+};
+
+declare
+    %test:assertEquals(1,1,2,3)
+function fn-pos:test-in-memory-dom() {
+    $fn-pos:nodes/a/position(),
+    $fn-pos:nodes//b/position()
+};
+
+declare
+    %test:assertEquals(1,1,2,3)
+function fn-pos:test-in-memory-dom-with-predicates() {
+    $fn-pos:nodes/a[true()]/position(),
+    $fn-pos:nodes//b[true()]/position()
+};
+
+declare
+    %test:assertEquals(1,1)
+function fn-pos:test-in-memory-dom-with-positional-predicates() {
+    $fn-pos:nodes/a[1]/position(),
+    $fn-pos:nodes//b[3]/position()
+};
+
+declare
+    %test:assertEquals(1,1,2,3)
+function fn-pos:test-persistent-dom() {
+    let $dom := doc('/db/' || $fn-pos:collection || '/' || $fn-pos:doc)
+
+    return (
+        $dom/a/position(),
+        $dom//b/position()
+    )
+};
+
+declare
+    %test:assertEquals(1,1,2,3)
+function fn-pos:test-persistent-dom-with-predicates() {
+    let $dom := doc('/db/' || $fn-pos:collection || '/' || $fn-pos:doc)
+
+    return (
+        $dom/a[true()]/position(),
+        $dom//b[true()]/position()
+    )
+};
+
+declare
+    %test:assertEquals(1,1)
+function fn-pos:test-persistent-dom-with-positional-predicates() {
+    let $dom := doc('/db/' || $fn-pos:collection || '/' || $fn-pos:doc)
+
+    return (
+        $dom/a[1]/position(),
+        $dom//b[3]/position()
+    )
+};
--- a/exist-core/src/test/xquery/fn-position.xqm
+++ b/exist-core/src/test/xquery/fn-position.xqm
@@ -1,3 +1,24 @@
+(:
+ : eXist-db Open Source Native XML Database
+ : Copyright (C) 2001 The eXist-db Authors
+ :
+ : info@exist-db.org
+ : http://www.exist-db.org
+ :
+ : This library is free software; you can redistribute it and/or
+ : modify it under the terms of the GNU Lesser General Public
+ : License as published by the Free Software Foundation; either
+ : version 2.1 of the License, or (at your option) any later version.
+ :
+ : This library is distributed in the hope that it will be useful,
+ : but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


exist-core/src/main/java/org/exist/xquery/LocationStep.java
```diff
@@ -227,7 +227,7 @@ private Sequence processPredicate(Sequence outerSequence, final Sequence context
             result = pred.evalPredicate(outerSequence, result, axis);
             // subsequent predicates operate on the result of the previous one
             outerSequence = null;
-            context.setContextSequencePosition(-1, null);
```
dizzzz, line 230: Ah! I figured out there was 'something' starting from -1 ....
Did you run the xqts stuff?

dizzzz (approved):
One question....

exist-core/src/main/java/org/exist/xquery/LocationStep.java
line-o: No, I haven't

line-o: Do you expect that this would bring us closer to 100%?

line-o: OK, I installed the SNAPSHOT then packaged XQTS-runner and ran that.
It hangs at
`03 Mar 2021 11:17:13,313 [XQTSRunnerSystem-akka.actor.default-dispatcher-8] INFO  (slf4j.scala [info]:159) - Serialized results for TestSet: prod-Predicate OK.`
I waited half an hour ... twice. So, no XQTS run :(

line-o: Just checked, this is with the lates changes to XQTS-runner `8e39a3dd824d05a55d931b1624b18e83491264c0`

adamretter: @line-o I would need more information about the XQTS problem to be able to help... For example, what arguments did you give to XQTS? and what was the stack trace / log when you say it produced no results. Did it hang or did it quit? Maybe open an issue against XQTS with steps to reproduce what you are seeing?

line-o: in exist-db branch fix/3737

1.    mvn clean install -DskipTests=true

in xqts-runner branch master

1. sbt package
2. target/scala-2.13/exist-xqts-runner-assembly-1.0.0.jar -l tests -x HEAD

dizzzz:
@dependabot rebase

exist-core/src/main/java/org/exist/xquery/LocationStep.java
adamretter: @line-o addressed here - https://github.com/eXist-db/exist/pull/3774

line-o:
XQTS test result with this PR applied 

-- | -- | -- | -- | -- | --
31520 | 5146 | 563 | 1256 | 81.89% | 1127.849

NOTE: the sum of tests is not the same as the one in  @adamretter 's screenshot (31557)
I am currently looking into that.

exist-core/src/main/java/org/exist/xquery/LocationStep.java
line-o: Works now 
NOTE: I had to use `sbt assemble` instead of `sbt package`

line-o:
The testsuite needed to be updated to its latest version 
(-x HEAD does not download the newest with an existing checkout of HEAD)


-- | -- | -- | -- | -- | --
31557 | 5161 | 570 | 1260 | 81.84% | 1649.272

line-o:
Judging from the screenshot in #3774 we have two new failures with this PR applied

line-o:
Weird, when running the tests against 91eb77a76132cb44191208b591a87218c6c010e3

I get 5163 failing tests (so actually two more than with this PR applied)

The two tests in question are **group-015** and **cbcl-left-outer-join-004**

line-o:
```xml
  <test-case name="group-015">
      <description>No value comparisons are available to compare the grouping keys.</description>
      <created by="Josh Spiegel" on="2012-10-02"/>
      <modified by="Michael Kay" on="2017-03-17" change="avoid assert-xml for non-XML results"/>
      <test>
          for $x in (true(), "true", xs:QName("true"))
          group by $x
          return $x
      </test>
      <result>
        <assert-permutation>true(), "true", xs:QName("true")</assert-permutation>
      </result>
   </test-case>
```

```xml
   <test-case name="cbcl-left-outer-join-004">
      <description/>
      <created by="Nick Jones" on="2010-03-24"/>      
      <dependency type="spec" value="XQ10+"/>
      <test><![CDATA[<tbody> { 
         let $rows := <tables> <table> <row> <entry>Acetazolamide</entry> <entry>Acetazolamide</entry> </row> <row> <entry>Acetazolamide sodium</entry> <entry>Acetazolamide sodium</entry> </row> </table> <table> <row> <entry>Acetylcholine chloride</entry> <entry>Acetylcholine chloride</entry> </row> </table> <table> <row> <entry>Acetylcysteine</entry> <entry>Acetylcysteine</entry> </row> <row> <entry>Acetylcysteine sodium</entry> <entry>Acetylcysteine sodium</entry> </row> </table> </tables>/table/row 
         return for $g in distinct-values($rows/entry[2][string(.)])  
                order by $g 
                return let $matches := for $row in $rows 
                                       where $g = string($row/entry[2][string(.)]) 
                                       return $row/entry[1] 
                       return <entry> { $matches/( <link> { node() } </link>, text { if (position() lt last()) then '; ' else () } ) } </entry> 
         } </tbody>]]></test>
      <result>
         <assert-xml>&lt;tbody&gt;&lt;entry&gt;&lt;link&gt;Acetazolamide&lt;/link&gt;&lt;/entry&gt;&lt;entry&gt;&lt;link&gt;Acetazolamide sodium&lt;/link&gt;&lt;/entry&gt;&lt;entry&gt;&lt;link&gt;Acetylcholine chloride&lt;/link&gt;&lt;/entry&gt;&lt;entry&gt;&lt;link&gt;Acetylcysteine&lt;/link&gt;&lt;/entry&gt;&lt;entry&gt;&lt;link&gt;Acetylcysteine sodium&lt;/link&gt;&lt;/entry&gt;&lt;/tbody&gt;</assert-xml>
      </result>
   </test-case>
```

line-o:
While the latter makes sense as it compares `last()` with `position()`, I wonder how the grouping got affected.

line-o:
Can someone ... AKA @dizzzz restart the appveyor build, pliis 🙏

line-o:
@adamretter can you please run XQTS against this PR to see if you get the same results?

dizzzz:
I see one extra issue with the PR :-(
image

develop (top) vs branch  (below) --> rebase needed?

line-o:
@dizzzz you have an older version of the q3testsuite the current one has 31557 tests. You need to either remove the current test folder works/q3testmaster or specify a different folder with `-l mycustomfolder`.
And please specify the commit you checked against.

dizzzz:
i only know about ` java -jar target/scala-2.13/exist-xqts-runner-assembly-1.0.0.jar -x HEAD` :-)

line-o:
@dizzzz First of all, thank you for checking and also restarting the build 💯 🥇

dizzzz:
I started blindly to to some tests.... not checking the ticket :-P
We need to update the documentation of the xqts suite

line-o:
Because of the doubts raised by @dizzzz I reran qt3tests against
- current develop (91eb77a76132cb44191208b591a87218c6c010e3) and
- this branch, which is rebased on top of current develop (f7f81990be64c023d62aee8d0215fb640bcfe6b0)

both result in 

-- | -- | -- | -- | -- | --
31557 | 5162 | 570 | 1260 | 81.84% | 1276.237

🤷

line-o:
I suspect that it is necessary to run `sbt clean` first

line-o:
`sbt clean` does _not_ change the outcome. 

So my conclusion is:with this PR applied the q3test result stays exactly the same.

dizzzz:
I see with patch 5161 failures, on develop 5162 so it is actually -as expected!- an improvement!

PR #3773, opened 2021-03-10
Fix storage of repo.xml during package installation

### LINKED ISSUE #3734: [BUG] repo: post-install scripts cannot access repo.xml descriptor
**Describe the bug**

The repo.xml file is not stored until after the script defined in the `<finish>` element (conventionally, "post-install.xql"), has been executed. As a result, the post-install.xql file cannot access data in repo.xml, such as its `<permissions>` element.

**Expected behavior**

According to the documentation's explanation of the `<finish>` element:

> `<finish>`: Like `<prepare>`, this element must point to an XQuery script. This will be executed *after* all data has been uploaded to the database.

(Emphasis in the original.)

Thus, the script defined in the `<finish>` element should run *after* all package assets have been installed.

**To Reproduce**

Many apps contain a stock config.xqm module, with a function, `config:repo-descriptor()`, which returns the repo.xml file. Calling this function in a post-install script fails with the error:

> Caused by: org.exist.xquery.XPathException: exerr:ERROR The actual return type does not match the sequence type declared in the function's signature: config:repo-descriptor() element(). Expected cardinality: exactly one, got 0. [at line 58, column 40, source: /Users/joe/Library/Application Support/org.exist/expathrepo/public-repo-2.0.0/post-install.xq]

The source of the problem is clear enough when viewing Deployment.java, which handles package installation:

https://github.com/eXist-db/exist/blob/303ca0db511ee93bac1b50ccede357fde9658e7b/exist-core/src/main/java/org/exist/repo/Deployment.java#L446-L454

I believe the fix should be to move line 454 to before line 446.

 - OS: macOS 11.2.1
 - eXist-db version: 5.3.0-SNAPSHOT (current develop)
 - Java Version: Liberica OpenJDK 1.8.0_28

**Additional context**
- How is eXist-db installed? Built from source

joewiz:
### Description:

Closes #3734. 

This PR simply makes sure repo.xml is stored before post-install.xql is run, as documentation indicates it should be—instead of waiting to store repo.xml after post-install.xql, which makes it impossible to read from repo.xml to determine default permissions, etc. 

Also, this PR adds to and, in some cases, corrects comments explaining the sequence of deployment—incrementally adding to our understanding of how deployment works and where it could be improved.

### Reference:

- https://github.com/eXist-db/exist/issues/3734#issuecomment-795106259
- https://github.com/eXist-db/public-repo/issues/55
- https://github.com/eXist-db/exist/commit/877a6b71c6a55ab2466fe51a8ab996ea9bd62330

### Type of tests:

I tested this PR on my fix to the issue above, and confirmed that when applying this PR, a corresponding PR for the public-repo, https://github.com/eXist-db/public-repo/pull/67, works perfectly.

### DIFF
--- a/exist-core/src/main/java/org/exist/repo/Deployment.java
+++ b/exist-core/src/main/java/org/exist/repo/Deployment.java
@@ -362,7 +362,7 @@ public class Deployment {
                 runQuery(broker, null, packageDir, setupPath.get(), pkgName, QueryPurpose.SETUP);
                 return Optional.empty();
             } else {
-                // otherwise copy all child directories to the target collection
+                // otherwise create the target collection
                 XmldbURI targetCollection = null;
                 if (userTarget != null) {
                     try {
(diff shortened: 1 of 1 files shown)


adamretter (approved):
LGTM

PR #3774, opened 2021-03-11
Fix a regression in finding the roots of the type hierarchy

adamretter:
Regression was caused by f69628d

With this the XQTS output is:
Screenshot 2021-03-11 at 13 58 23

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/value/Type.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/Type.java
@@ -579,7 +579,7 @@ public class Type {
         //Don't introduce a shortcut (starting at getSuperType(type1) here
         //type2 might be a super-type of type1
         int t;
-        for (t = type1; t != ITEM; t = getSuperType(t)) {
+        for (t = type1; t != ITEM && t != ANY_TYPE; t = getSuperType(t)) {
             //Shortcut
             if (t == type2) {
                 return t;
@@ -587,12 +587,12 @@ public class Type {
             t1.add(t);
         }
         //Starting from type2's super type : the shortcut should have done its job
-        for (t = getSuperType(type2); t != ITEM; t = getSuperType(t)) {
+        for (t = getSuperType(type2); t != ITEM && t != ANY_TYPE; t = getSuperType(t)) {
             if (t1.contains(t)) {
                 return t;
             }
         }
-        return ITEM;
+        return t;
     }
 
     /**


dizzzz (approved):

line-o:
Cool!

exist-core/src/main/java/org/exist/xquery/value/Type.java
```diff
@@ -579,20 +579,20 @@ public static int getCommonSuperType(final int type1, final int type2) {
         //Don't introduce a shortcut (starting at getSuperType(type1) here
         //type2 might be a super-type of type1
         int t;
-        for (t = type1; t != ITEM; t = getSuperType(t)) {
+        for (t = type1; t != ITEM && t != ANY_TYPE; t = getSuperType(t)) {
```
line-o, line 582: Would that be one less check for the more frequent case t being ITEM?
`!(t==ITEM || t==ANY_TYPE)`

line-o:
Weird fact:
When I run XQTS with -x HEAD it reports a slightly different sum of tests


-- | -- | -- | -- | -- | --
31520 | 5146 | 563 | 1256 | 81.89% | 1127.849

adamretter:
>  with -x HEAD

As opposed to?

line-o:
As opposed to the stats you posted with 31557 tests in total.
Do you have any idea what could be the reason?

adamretter:
@line-o I ran with -x HEAD as well. The XQTS caches the download of the XQTS tests, perhaps you have an older version (they are not in-frequently updated)

line-o:
yep -x HEAD does not download when *-HEAD is present


-- | -- | -- | -- | -- | --
31557 | 5161 | 570 | 1260 | 81.84% | 1649.272

line-o:
Which commit gave you 5159 failures? I get 
- 5161 for #3761 (f7f81990be64c023d62aee8d0215fb640bcfe6b0) and 
- 5163 for this PR applied (91eb77a76132cb44191208b591a87218c6c010e3)

adamretter:
It was this PR but I cannot tell you if that was before or after it was rebased/merged to develop.

adamretter:
I think this information would be much more useful in an XQTS issue than in a closed issue here... perhaps we could move the converstation.

line-o:
Well, I just wanted to post it on slack ... :)
Why would you want to have it on XQTS?

PR #3791, opened 2021-03-30
Fix incomplete type error reporting

### LINKED ISSUE #3628: [BUG] incomplete type error reporting
**Describe the bug**
Type error message about a mismatch between the expected element name and the supplied element does not report the element names.

**Expected behavior**
report the element names

**To Reproduce**
`let $test as element(b)? := <a/>
return $test`

will return `err:XPTY0004 Invalid type for variable $test. Expected element()?, got element()`
whereas it should return ...`Expected element(b)?, got element(a)`

 - OS: linux
 - eXist-db version: 5.3.0-SNAPSHOT dated 20200816142230 2e983c7a95fa4f8bde6b3e5a44fda59326d1de3b

dizzzz:
Fixed #3628 

the query

```xquery
let $test as element(b)? := <a c="2"><d/></a> return $test
```

returns now

```
err:XPTY0004 Invalid type for variable $test. Expected element(b)?, got element(a)
```

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/LetExpr.java
+++ b/exist-core/src/main/java/org/exist/xquery/LetExpr.java
@@ -153,12 +153,15 @@ public class LetExpr extends BindingExpression {
                                 }
                             }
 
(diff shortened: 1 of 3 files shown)


adamretter (approved):
Nice!

PR #3799, opened 2021-04-02
[bugfix] NPE on predicates with an empty sequence

### LINKED ISSUE #3797: [BUG] NPE on empty sequence in predicate
**Describe the bug**

The following XQuery leads to a NullPointerException `(1)[()]`

**Expected behavior**

If this is an error, an exception with a meaningful error message and location date should be thrown.

**To Reproduce**

evaluate the XQuery snippet above

 - eXist-db version: latest develop 4415b13d
 - Java Version recent Adopt Open JDK 8

**Additional context**

line-o:
### Description:

If the inner sequence in a predicate evaluates to an empty sequence, an empty sequence is returned.

### Reference:

fixes #3797

Saxon 10 HE and BaseX 9.5 as well as eXist-db versions > 5.3.0-SNAPSHOT all have this behaviour.

@dizzzz showed in #3800 a nicer condition with `Sequence#isEmpty` so I incorporated that.

### Type of tests:

Several tests added to positional.xql XQSuite

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/Predicate.java
+++ b/exist-core/src/main/java/org/exist/xquery/Predicate.java
@@ -194,7 +194,11 @@ public class Predicate extends PathExpr {
                         innerSeq = inner.eval(Dependency.dependsOn(inner.getDependencies(), Dependency.CONTEXT_ITEM)
                                 ? contextSequence : null);
                     }
-                    if (innerSeq.getCardinality().isSubCardinalityOrEqualOf(Cardinality.EXACTLY_ONE)) {
+
+                    // We must check for empty sequences here to avoid an NPE
+                    if (innerSeq.isEmpty()) {
+                        result = Sequence.EMPTY_SEQUENCE;
+                    } else if (innerSeq.getCardinality().isSubCardinalityOrEqualOf(Cardinality.EXACTLY_ONE)) {
                         result = selectByPosition(outerSequence, contextSequence, mode, innerSeq);
                     } else {
                         throw new XPathException(this, ErrorCodes.FORG0006, "Effective boolean value is not defined for a sequence of two or more items starting with a " + Type.getTypeName(innerSeq.itemAt(0).getType()) + " value");
--- a/exist-core/src/test/xquery/positional.xql
+++ b/exist-core/src/test/xquery/positional.xql
@@ -105,3 +105,29 @@ declare
 function pf:multiple-positions() {
     (3,4)[1,2]
 };
+
+declare
+    %test:assertError("err:FORG0006")
+function pf:implicit-position-sequence() {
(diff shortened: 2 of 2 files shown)


line-o:
XQTS test result

-- | -- | -- | -- | -- | --
31557 | 5161 | 563 | 1260 | 81.86% | 1345.101

exist-core/src/main/java/org/exist/xquery/Predicate.java
```diff
@@ -194,9 +194,13 @@ public Sequence evalPredicate(final Sequence outerSequence,
...
                                 ? contextSequence : null);
                     }
-                    if (innerSeq.getCardinality().isSubCardinalityOrEqualOf(Cardinality.EXACTLY_ONE)) {
+
+                    if (innerSeq.getCardinality().equals(Cardinality.EMPTY_SEQUENCE)) {
+                        result = Sequence.EMPTY_SEQUENCE;
+                    } else if (innerSeq.getCardinality().isSubCardinalityOrEqualOf(Cardinality.EXACTLY_ONE)) {
                         result = selectByPosition(outerSequence, contextSequence, mode, innerSeq);
-                    } else {
+                    }
```
dizzzz, line 202: formatting is not ok?

exist-core/src/main/java/org/exist/xquery/Predicate.java
```diff
@@ -194,9 +194,13 @@ public Sequence evalPredicate(final Sequence outerSequence,
                         innerSeq = inner.eval(Dependency.dependsOn(inner.getDependencies(), Dependency.CONTEXT_ITEM)
                                 ? contextSequence : null);
                     }
-                    if (innerSeq.getCardinality().isSubCardinalityOrEqualOf(Cardinality.EXACTLY_ONE)) {
+
+                    if (innerSeq.getCardinality().equals(Cardinality.EMPTY_SEQUENCE)) {
```
dizzzz, line 198: in essence I have the same check...... I think.

exist-core/src/main/java/org/exist/xquery/Predicate.java
line-o: It depends, I would write all else blocks that way. The implicit coding style of eXist-db is different, though. I will change that.

exist-core/src/main/java/org/exist/xquery/Predicate.java
line-o: Are you sure `isEmpty()` works the way we think it does? @dizzzz

line-o: I checked, it is exactly what we want :)

dizzzz (approved):

PR #3810, opened 2021-04-08
Fixes corrupted stream message with maven-surefire-plugin

reinhapa:
Signed-off-by: Patrick Reinhart <patrick@reini.net>


line-o (commented):
Thanks for your contribution. Can you please change the commit message to
`[bugfix] corrupted stream message with maven-surefire-plugin`
I would also love to have to separate commits while you are at it
`[ignore] update maven-surefire-plugin to 3.0.0-M5`

PR #3811, opened 2021-04-08
[hotfix] repair appveyor build

dizzzz:
- upgrade Ubuntu
- remove docker part

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,9 +7,9 @@ environment:
       JAVA_HOME: C:\Program Files\Java\jdk1.8.0
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       JAVA_HOME: C:\Program Files\Java\jdk11
-    - APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu1804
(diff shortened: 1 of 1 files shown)


duncdrum (approved):
The changes are harmless, but i don't know what it is we are running linux builds on appveyor for. If we want to use these runs and their result, we should run docker tests, if we don't I'd say ditch the builds from the matrix and safe us all 1h per PR waiting on sequential builds

dizzzz:
at least the build is green, that saves me/us a few clicks every time

dizzzz:
ditching is ok for me for sure, red marks are not :-)
for me I still trust the appveyor builds as GHA has red marks still

adamretter:
@dizzzz Updating to newer Ubuntu is fine by me. Just wondering why you disabled building eXist-db's Docker Image?

dizzzz:
@adamretter I disabled docker as this was the part that was giving build failures according to the logs.

duncdrum:
@dizzzz that should only be a problem on `Java 11` if there are docker related failures on Java8 that would indicate a problem if we trust appveyor more then github actions

dizzzz:
Hi, "should" is the theory but in practice java8 and Java11 failed. Hence my fix. I really checked it.

PR #3700, opened 2021-01-10
[bugfix] make fn:base-uri compliant 

dizzzz:
Fixes #3699, a few xqts testcases and dom-ts:

dom-ts:
image

PR #3813, opened 2021-04-10
[bugfix] reimplement fn:base-uri()

dizzzz:
Clean version of the former PR https://github.com/eXist-db/exist/pull/3700

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunBaseURI.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunBaseURI.java
@@ -23,8 +23,6 @@ package org.exist.xquery.functions.fn;
 
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.exist.xquery.*;
 import org.exist.xquery.value.*;
 import org.w3c.dom.Node;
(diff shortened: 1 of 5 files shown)


dizzzz:
image

dizzzz:
Commit is clean, sonar cloud only gives coverage reports (not for this PR to be addressed). Codacy ... I do not agree with that reports, so I am not willing to fix these.

IMO the PR is clean,  upto quality standards and tested. Please pull it in?

exist-core/src/main/java/org/exist/xquery/functions/fn/FunBaseURI.java
```diff
@@ -21,67 +21,55 @@
  */
 package org.exist.xquery.functions.fn;
 
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.exist.xquery.*;
```
duncdrum, line 28: @dizzzz I find it hard to disagree with avoiding unused imports here

duncdrum (changes requested):
LGTM i would like you to remove the unused imports though

exist-core/src/main/java/org/exist/xquery/functions/fn/FunBaseURI.java
dizzzz: Please could you elaborate, I find this text a bit cryptic (I really can't understand it). 

the  `import org.exist.xquery.*;` is really needed?

dizzzz: as in: there are no unused imports

PR #3814, opened 2021-04-12
Import repo logging

adamretter:
1. Log a warning if autodeploy dir is not found
2. Show Repo logs on the Docker stdout console

### DIFF
--- a/exist-core/src/main/java/org/exist/repo/AutoDeploymentTrigger.java
+++ b/exist-core/src/main/java/org/exist/repo/AutoDeploymentTrigger.java
@@ -77,7 +77,7 @@ public class AutoDeploymentTrigger implements StartupTrigger {
         }
 
         if (!Files.isReadable(autodeployDir) && Files.isDirectory(autodeployDir)) {
-            LOG.warn("Unable to read autodeploy directory: " + autodeployDir);
+            LOG.warn("Unable to read autodeploy directory: {}", autodeployDir);
             return;
         }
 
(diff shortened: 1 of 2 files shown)


duncdrum (approved):

duncdrum (commented):
if there is new info in the logs by default, could you add a test to make sure that it stays there going forward

adamretter:
@duncdrum testing for log output is not something we have ever done and would be very hard to achieve. I think that is likely a step to far...

duncdrum:
while we're not testing logging as such, we are and have bee reading the logs of the container, to ensure our modifications to the logging config are working as intended, if we add more modifications, we should make sure that the new messages are visible to the client.

adamretter:
@duncdrum I won't be adding a test for this because I don't believe the time/effort trade-off is worth the value of the test.

Do you want me to close this PR, or are you happy to proceed without the test?

exist-core/src/main/java/org/exist/repo/AutoDeploymentTrigger.java
```diff
@@ -77,6 +77,7 @@ public void execute(final DBBroker sysBroker, final Txn transaction, final Map<S
         }
 
         if (!Files.isReadable(autodeployDir) && Files.isDirectory(autodeployDir)) {
+            LOG.warn("Unable to read autodeploy directory: " + autodeployDir);
```
line-o, line 80: The proposed fix for the code smell reported by SonarCloud
`LOG.warn("Unable to read autodeploy directory: {}", autodeployDir)`

dizzzz (changes requested):
LGTM, small logging item as requested

exist-core/src/main/java/org/exist/repo/AutoDeploymentTrigger.java
dizzzz: It is consistent with style of rest of code.

adamretter:
@line-o @dizzzz done.

dizzzz (approved):

PR #3830, opened 2021-04-21
[improvement] Improve invoking logger using log4j2 and slf4j templating

dizzzz:
Change logger string concatenation by correct log4j2 templating. Can improve performance.

Mechanically changed by feature of Intellij.

image

### DIFF
--- a/exist-core/src/main/java/org/exist/SystemProperties.java
+++ b/exist-core/src/main/java/org/exist/SystemProperties.java
@@ -54,7 +54,7 @@ public class SystemProperties {
                 properties.load(is);
             }
         } catch (final IOException ioe) {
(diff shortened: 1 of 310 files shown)


line-o (commented):
There is a test that fails now because of a call to `Log.error`.

Also we might want to do something about the hundreds of warnings about illegal reflective access that are reported (`--illegal-access=warn`).

line-o:
https://github.com/eXist-db/exist/pull/3830/checks?check_run_id=2400210987#step:6:9486

dizzzz:
ah. mocking log calls. interesting :-)

dizzzz:
PR is ready for review

joewiz (approved):

PR #3831, opened 2021-04-21
Connection Pooling for SQL Module

adamretter:
This feature is backported from FusionDB to eXist-db at the request of @PieterLamers

It allows connection pools to be established via eXist-db's conf.xml for the SQL Module. The connections can then be obtained via `sql:get-connection-from-pool`.

### DIFF
--- a/extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/GetConnectionFunction.java
+++ b/extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/GetConnectionFunction.java
@@ -1,14 +1,25 @@
 /*
- * eXist-db Open Source Native XML Database
- * Copyright (C) 2001 The eXist-db Authors
+ * Copyright (C) 2014, Evolved Binary Ltd
  *
- * info@exist-db.org
- * http://www.exist-db.org
+ * This file was originally ported from FusionDB to eXist-db by
+ * Evolved Binary, for the benefit of the eXist-db Open Source community.
+ * Only the ported code as it appears in this file, at the time that
+ * it was contributed to eXist-db, was re-licensed under The GNU
+ * Lesser General Public License v2.1 only for use in eXist-db.
+ *
+ * This license grant applies only to a snapshot of the code as it
+ * appeared when ported, it does not offer or infer any rights to either
+ * updates of this source code or access to the original source code.
+ *
+ * The GNU Lesser General Public License v2.1 only license follows.
+ *
+ * ---------------------------------------------------------------------
+ *
+ * Copyright (C) 2014, Evolved Binary Ltd
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation; either
- * version 2.1 of the License, or (at your option) any later version.
+ * License as published by the Free Software Foundation; version 2.1.
  *
  * This library is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


duncdrum:
We need to update the conf.xml listing and documentation for the new feature

adamretter:
> We need to update the conf.xml listing
That one is included :-)
https://github.com/eXist-db/exist/pull/3831/files#diff-145364ddf4d436154dde2de26ccc629285e6d349ba253bf1c657c3ecd8ee890c

> and documentation for the new feature
Okay that sounds like a good idea in principle. At the moment I don't think any of the XQuery extension modules are documented in the exist-documentation (apart from the ones that come as parts of the Indexes). I am wondering where best such documentation would live? Should I have a page just for the SQL Module, or should there be a page for XQuery Extension modules from which that is linked? I guess it goes in the index somewhere here - http://www.exist-db.org/exist/apps/doc/#xq ?

duncdrum:
the extension modules appear on this page, not sure if you want to add something there or not [extensions.xml](http://www.exist-db.org/exist/apps/doc/extensions.xml?field=all&id=D3.15.29.3)

adamretter:
> the extension modules appear on this page, not sure if you want to add something there or not extensions.xml

Ah ha! That page appears to be very outdated. As it stands, I think its problematic as it has to be manually updated each time an extension is added/removed. I think it made sense in the past where not all modules were shipped with eXist-db, but since eXist-db 5.0.0 all modules are shipped in the release. Now that is the case, all modules and functions are searchable from the fundocs.

I think I could either - remove the page entirely, or repurpose it so it just contains modules that have additional documentation available.

duncdrum:
@adamretter no preference, i also didn't check with the latest version of these pages, but those are the two locations that match the feature in the PR. I would expect a minor update in at least one location so we are staying up-to-date

dizzzz:
Imo delete the page

extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/GetConnectionFunction.java
```diff
@@ -45,120 +54,138 @@
...
-            logger.error("sql:get-connection() Cannot find database driver class: " + dbDriver, cnfe);
-            throw (new XPathException(this, "sql:get-connection() Cannot find database driver class: " + dbDriver, cnfe));
-        } catch (InstantiationException ie) {
-            logger.error("sql:get-connection() Cannot instantiate database driver class: " + dbDriver, ie);
-            throw (new XPathException(this, "sql:get-connection() Cannot instantiate database driver class: " + dbDriver, ie));
-        } catch (SQLException sqle) {
-            logger.error("sql:get-connection() Cannot connect to database: " + dbURL, sqle);
-            throw (new XPathException(this, "sql:get-connection() Cannot connect to database: " + dbURL, sqle));
+        } catch (final SQLException sqle) {
+            LOGGER.error("sql:get-connection() Cannot connect to database: " + dbURL, sqle);
```
dizzzz, line 166: duplicated text. otherwise please use the {} logging templating.

dizzzz (approved):
LGTM

extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/GetConnectionFunction.java
dizzzz: sync code

adamretter:
Documentation is here: https://github.com/eXist-db/documentation/pull/608

dizzzz (approved):
LGTM

adamretter:
Thanks @dizzzz :-)

PR #3844, opened 2021-04-28
Replace Bintray URLs with GitHub Releases

joewiz:
### Description:

Switch downloads/releases URLs in the README from Bintray to GitHub Releases

### Reference:

https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

### Type of tests:

n/a


joewiz:
(The [exist-versioning-release.md document](https://github.com/eXist-db/exist/blob/develop/exist-versioning-release.md) will also need to be updated, but we don't have updated procedures yet.)

dizzzz (approved):

adamretter:
> The exist-versioning-release.md document will also need to be updated, but we don't have updated procedures yet

The uploads to BinTray were automated previously. I am not sure what we can do now, but also the pom.xml build files will need changing. We should open an issue for that separately...

PR #3877, opened 2021-05-07
Lazily created caches should be bounded

adamretter:
Previously Caches for use with the XQuery Cache Module in eXist-db could be created on-demand when the cache was first operated on. When such a cache was lazily created it had no bounds on its size, meaning that it could grow without limit until consuming all memory and causing an OutOfMemoryError.

This PR:

1. Adds configurable bounds for lazily created caches via `conf.xml`
2. Adds a sensible default bounds to `conf.xml` for lazily created caches.
3. Adds an option to disable lazily created caches entirely. I would generally recommend using this in any serious application.
4. Adds further options for configuring security aspects of lazily created caches via `conf.xml`

NOTE: That before this PR, several DoS attacks against eXist-db were possible by misusing the cache either intentionally/maliciously, or unintentionally.

### DIFF
--- a/exist-distribution/src/main/config/conf.xml
+++ b/exist-distribution/src/main/config/conf.xml
@@ -989,7 +989,31 @@
             <!--
                 Extension Modules
             -->
-            <module uri="http://exist-db.org/xquery/cache" class="org.exist.xquery.modules.cache.CacheModule"/>
+            <module uri="http://exist-db.org/xquery/cache" class="org.exist.xquery.modules.cache.CacheModule">
+                <!--
+                If a named Cache does not exist it can be created lazily on-demand
+                when an operation is first performed upon it.
+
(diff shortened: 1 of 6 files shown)


dizzzz (approved):
impressive @adamretter , great work, great catch

adamretter:
@duncdrum As requested - https://github.com/eXist-db/documentation/pull/614

adamretter:
> not sure what to make of those

@duncdrum As far as I can see they are not related. The Win J11 one is related to slow performance of the CI, and the J15 Linux one is likely showing a STDOUT corruption somewhere (but is unrelated to this change).

duncdrum (approved):
YOLO it is, 
🧙 `STDOUT` shall not thwart our upcoming next releases anymore 🌚

PR #3890, opened 2021-05-19
Backport of sql:close-connection from FusionDB

adamretter:
Allows the user to explicitly close (or return to a pool) a SQL connection before the XQuery completes, this is useful when working with Connection pools.

### DIFF
--- a/extensions/modules/sql/src/test/java/org/exist/xquery/modules/sql/Util.java
+++ b/extensions/modules/sql/src/test/java/org/exist/xquery/modules/sql/Util.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright (C) 2014, Evolved Binary Ltd
+ *
+ * This file was originally ported from FusionDB to eXist-db by
+ * Evolved Binary, for the benefit of the eXist-db Open Source community.
+ * Only the ported code as it appears in this file, at the time that
+ * it was contributed to eXist-db, was re-licensed under The GNU
+ * Lesser General Public License v2.1 only for use in eXist-db.
+ *
+ * This license grant applies only to a snapshot of the code as it
+ * appeared when ported, it does not offer or infer any rights to either
+ * updates of this source code or access to the original source code.
+ *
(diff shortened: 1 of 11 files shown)


extensions/modules/sql/src/test/java/org/exist/xquery/modules/sql/Util.java
```diff
@@ -0,0 +1,95 @@
...
+        } else {
+            context = compiled.getContext();
+            context.prepareForReuse();
+        }
+
+        if (compiled == null) {
+            compiled = xqueryService.compile(broker, context, query);
+        } else {
+            compiled.getContext().updateContext(context);
+            context.getWatchDog().reset();
```
line-o, line 90: What does `context.getWatchdog().reset()` do?

line-o:
7473d654b520bbd3b6756ff45a5c78d2be8e3463 modifies a core part of existdb
Please explain what changes were made and why they are necessary.

line-o:
The overall changes look good. If you could answer my questions so that I better understand the implications I will approve.

adamretter:
> 7473d65 modifies a core part of existdb
> Please explain what changes were made and why they are necessary.

Is there a problem with changing eXist-db suddenly?

The changes are very simple. Firstly I have reduced duplicated code which appears to have been copied and pasted, secondly it was previously impossible to return a value when modifying a context map, I have now added that facility - it is needed by the SQL Module.

extensions/modules/sql/src/test/java/org/exist/xquery/modules/sql/Util.java
adamretter: It resets the XQuery Timeout Watchdog. This isn't new, I just moved the code from where it was before.

dizzzz:
Lgtm

dizzzz (approved):

adamretter:
Thanks @dizzzz :-)

PR #3891, opened 2021-05-19
Specify the COVERALLS_TOKEN in the correct way for CI

adamretter:
It's easy when you know how ;-)

I expect this to fail in the PR for the GitHub Action `(JDK 8 / ubuntu-latest) Test`, as it won't be able to access the secrets. I have tested this change in a different project, and once merged it did work there.


duncdrum (approved):
Have you made sure that our token is a repo token, and not an org token? Even if not changing that won’t be visible here, so good to proceed.

adamretter:
@duncdrum I believe so

Issue #3895, opened 2021-05-20
[BUG]


stexandev:
**Describe the bug**
`repo:get-resource` does not throw an error if the expected resource does not exist but instead just returns nothing

**Expected behavior**
[should throw "an error"](http://www.exist-db.org/exist/apps/fundocs/view.html?uri=http://exist-db.org/xquery/repo&location=java:org.exist.xquery.modules.expathrepo.ExpathPackageModule&details=true#get-resource.2), but instead [does nothing](https://github.com/eXist-db/exist/blob/5bc6a35fd6d4a91742637aefe4ad53c2324284ee/extensions/modules/expathrepo/src/main/java/org/exist/xquery/modules/expathrepo/GetResource.java#L76)

**To Reproduce**

```Xquery
xquery version "3.1";

repo:get-resource('http://exist-db.org/apps/eXide', 'repo.xml'),
repo:get-resource('http://exist-db.org/apps/eXide', 'doesnotexist.xml')
```

run this script in eXide 

**Context (please always complete the following information):**
-    existdb/existdb:5.2.0
-    Docker version 19.03.8, build afacb8b7f0
-    Ubuntu 20.04

joewiz:
The function documentation for this function clearly indicates that an error should be thrown:

> Returns:
> xs:base64Binary? : `<status result="ok"/>` if deployment was ok. Throws an error otherwise.

That said, the bit about "`<status result="ok"/>` if deployment was ok" makes no sense, since this function doesn't deploy anything. Looking at the other functions in this module, it appears that the description for the return portion of many functions was just copied and pasted. The following functions have the word-for-word description:

- `repo:deploy`
- `repo:install-and-deploy`
- `repo:install-and-deploy-from-db`
- `repo:undeploy`

That leads me to question whether the phrase, "Throws an error otherwise", was really intended to be the behavior for this function? 

If we agree that an error should be thrown, could the fix also delete the erroneous portion from the function documentation, i.e.:

> `<status result="ok"/>` if deployment was ok.

This definitely does not belong in the function documentation for `repo:get-resource`.

line-o:
fixed by #3896

PR #3896, opened 2021-05-20
[bugfix] throw error in `repo:get-resource` if file does not exist

stexandev:
### Description:
`repo:get-resource` now throws an XPathException if the resource does not exist

### Reference:
https://github.com/eXist-db/exist/issues/3895

### Type of tests:
none


line-o (approved):
Thats a lot, @stexandev!

That is clearly better than doing //nothing :)

adamretter:
As per https://github.com/eXist-db/exist/issues/3904#issuecomment-848326023 this needs to be revered until 6.0.0 as it has caused breaking API changes in other applications.

Issue #3889, opened 2021-05-19
SQL Pooling mechanism does not always return connections


xatapult:
**Describe the bug**
We (John Benjamins, Amsterdam) have been asking for the SQL connection pooling and Adam Retter implemented this recently (may 2021). Initially all looked fine, but there seems to be a blocking bug in the implementation:

When you use the pool directly from an XQuery script all is fine. So, assuming you've set up pooling correctly in the `conf.xml` this works fine and can be repeated indefinitely (or so it seems):

```
let $connection := sql:get-connection-from-pool('your-pool-name')
return
  sql:execute($connection, 'some sql statement', false())
```

But: When I put the *same* code in a function in a separate XQuery module (that is then imported by the XQuery script), I can only call that function as many times as the size of the pool. After that it raises the error:

```
exerr:ERROR sql:get-connection-from-pool() Cannot retrieve connection from pool: ...
```

So it seems that pool connections are not returned to the pool when they're created inside a module function when that function ends.

**Expected behavior**
When an XQuery script ends, all pool connections should be given back to the pool, regardless of whether they're created directly or in a module function.

**To Reproduce**
It's rather hard to reproduce in a simple way. Since I don't know what databases are available I cannot provide code (pool names and SQL statements will always be different)

* Set up pooling in conf.xml (best with a small pool size so the bug turns up quickly)
* Use this pooling in the way described above

**Context (please always complete the following information):**
 - OS: Windows 10
 - eXist-db version: 5.3.0-SNAPSHOT build 20210517111828
 - Java Version 1.8.0_281

**Additional context**
- How is eXist-db installed? JAR
- Any custom changes in e.g. `conf.xml`? Pool set like this:

```
<module uri="http://exist-db.org/xquery/sql" class="org.exist.xquery.modules.sql.SQLModule">
  <parameter name="pool.1.name" value="pool-pres"/>
  <parameter name="pool.1.properties.jdbcUrl"
    value="jdbc:sqlserver://172.16.0.11;databaseName=pres_prod;selectMethod=direct;responseBuffering=adaptive"/>
  <parameter name="pool.1.properties.dataSource.user" value="..."/>
  <parameter name="pool.1.properties.dataSource.password" value="..."/>
  <parameter name="pool.1.properties.dataSource.databaseName" value="pres_prod"/>
  <parameter name="pool.1.properties.maximumPoolSize" value="10"/>
  <parameter name="pool.1.properties.minimumIdle" value="0"/>
[...]

PR #3897, opened 2021-05-20
Cleanup all module contexts

adamretter:
Closes https://github.com/eXist-db/exist/issues/3889

Previously Library Modules were reset with the wrong context, i.e. they were provided with the Main Module's context and not the Library Module's Context.
This meant that not all resources stored in context variables were correctly cleaned up, therefore leaking resources and memory. In addition *reset* was previously called more than once for the same module which was unnecessary; therefore we should also see a small performance improvement from this fix.

Many thanks to @xatapult and @pieterlamers for their patience in reporting this and helping to track down the issue.
As a result all SQL Connections should now also be correctly and implicitly closed when a query finishes execution.


dizzzz (approved):
LGTM

PR #3840, opened 2021-04-27
[cicd] Build with Java16

dizzzz:
Lets check the build results;
Java15 is EOL and unstable for eXist-db; v16 is more logical to use.

PR #3921, opened 2021-06-08
fix(ci): add J16 and allow failures

duncdrum:
Drop J15, build and test on J16. Failures will not impact the overall result if they occur on J16 (yet). 
Build javadocs on J11 and J16 both are failing, ignore all failures of doc building, we still get the news of whats wrong.

I dropped J8 from the docs, as these haven't been deployed anywhere from what i could tell, correct me if i m wrong. 

supsersedes #3840

### DIFF
--- a/.github/workflows/ci-docs.yml
+++ b/.github/workflows/ci-docs.yml
@@ -5,7 +5,7 @@ jobs:
     name: ${{ matrix.jdk }} Javadocs 
     strategy:
       matrix: 
-        jdk: ['11', '16']
+        jdk: ['8','11', '16']
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


adamretter (dismissed):
Thanks for discussing with me @duncdrum 

As per our discussion, we must build the docs on J8 as I want to know that works before we get to "release time" and our release platform is Java 8. So if you can make that small change then I am happy to Merge

.github/workflows/ci-docs.yml
```diff
@@ -22,6 +22,10 @@ jobs:
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
       - name: Maven Javadoc
+        if: ${{ matrix.jdk == '8' }} 
         run: mvn -V -B -q -T 2C javadoc:javadoc
-        continue-on-error: true
+      - name: Maven Javadoc allow fal
```
adamretter, line 27: `fal` -> `fail`?

dizzzz (approved):
LGTM

Issue #3063, opened 2019-10-22
Apache XML-RPC CVE-2016-5002


adamretter:
There are security issues in the latest (and final) release of Apache XML-RPC libraries. The details can be found here: https://www.cvedetails.com/cve/CVE-2016-5002/

The Apache project doesn't seem to be fixing this. I learned that this it is fixed in Redhat linux distro's. Trying to find a suitable Jar with sources was quite difficult, the closest thing I found was this - http://ftp.uni-erlangen.de/mirrors/Mageia/distrib/cauldron/SRPMS/core/release/xmlrpc-3.1.3-73.mga7.src.rpm

Ideally we should be able to find a Source-RPM from a CentOS repo. From there we can look at publishing Maven artifacts... help would be appreciated.

PR #3934, opened 2021-06-18
Update Apache XML-RPC

adamretter:
Update to Apache XML-RPC with latest Security Patches from Evolved Binary.
Addresses CVE-2019-17570 and CVE-2016-5002

**Note** - This PR changes XML-RPC API (and therefore also the XML:DB remote API).

Previously eXist-db was relying on the in-built Java Serialization for communicating ACL's, unfortunately that facility is inherently insecure. This PR now adds a custom XML-RPC Serializer and Parser for ACLs to ensure this is secure; the upshot is that the wire-protocol has to change to support this. This means users of the XML-RPC and XML:DB APIs (e.g. Java Admin Client, oXygen XML Editor, etc) will need to upgrade their libraries as this is not backwards compatible.

Closes https://github.com/eXist-db/exist/issues/3063

### DIFF
--- a/exist-core/pom.xml
+++ b/exist-core/pom.xml
@@ -346,7 +346,7 @@
         </dependency>
 
         <dependency>
-            <groupId>org.apache.xmlrpc</groupId>
+            <groupId>com.evolvedbinary.thirdparty.org.apache.xmlrpc</groupId>
             <artifactId>xmlrpc-common</artifactId>
             <exclusions>
                 <exclusion> <!-- conflicts with xerces 2.12.0 dependency on xml-apis version 1.4.01 -->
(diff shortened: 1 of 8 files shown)


line-o:
@adamretter Many thanks for this contribution.
I would like to know if this might have implications to other XML-RPC clients communicating with exist (e.g. [gulp-exist](https://github.com/eXist-db/gulp-exist)).

line-o (commented):
Looks good. I will approve after a local test.

adamretter:
> I would like to know if this might have implications to other XML-RPC clients communicating with exist (e.g. gulp-exist).

I can't say. I don't know gulp-exist or what it does or how it works, the author of that should be able to tell you.

dizzzz (approved):

Issue #3932, opened 2021-06-14
[BUG] Cache module: "lazy creation disabled" error thrown when lazy creation is enabled


joewiz:
**Describe the bug**

With a fresh build of 5.3.0-SNAPSHOT, calling `cache:put("foo", "bar", "baz")` fails with this error:

> `cache:lazy-creation-disabled` There is no such named cache: `foo`, and lazy creation of the cache has been disabled.

But in conf.xml, the cache module has the default parameter:

```xml
<module uri="http://exist-db.org/xquery/cache" class="org.exist.xquery.modules.cache.CacheModule">
    <parameter name="enableLazyCreation" value="true"/>
```

From my reading of PR #3877 where the new options were added and the accompanying [documentation PR #614](https://github.com/eXist-db/documentation/pull/614), using `cache:put` to lazily create a cache when `conf.xml` has `enableLazyCreation=true` (i.e., the default) should not throw this error.

**Expected behavior**

If `conf.xml` has `enableLazyCreation=true` (i.e., the default), a call to `cache:put()` for a cache that hasn't been explicitly created should not trigger a `cache:lazy-creation-disabled` error.

**To Reproduce**

The following test assumes that `conf.xml` has `enableLazyCreation=true`. 

```xquery
xquery version "3.1";

module namespace t="http://exist-db.org/xquery/test";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare
    %test:setUp
function t:setup() {
    if (cache:names() = "foo") then
        cache:destroy("foo")
    else
        ()
};

declare
    %test:tearDown
function t:tearDown() {
    cache:destroy("foo")
};

declare
    %test:assertEmpty
function t:test() {
    try {
        cache:put("foo", "bar", "baz")
    } catch * {
        $err:code
    }
};
```

The test suite returns this result:

```xml
<testsuite package="http://exist-db.org/xquery/test" timestamp="2021-06-14T12:30:27.216-04:00"
    tests="1" failures="1" errors="0" pending="0" time="PT0.001S">
    <testcase name="test" class="t:test">
        <failure message="assertEmpty failed." type="failure-error-code-1"/>
        <output>cache:lazy-creation-disabled</output>
    </testcase>
</testsuite>
```

**Context (please always complete the following information):**
 - OS: macOS 11.4
 - eXist-db version: eXist 5.3.0-SNAPSHOT e371efd9987a9a2f4414839c7bf1dbc20107b6d1 20210604033555
[...]

PR #3935, opened 2021-06-18
Further tests for Cache Module and a small NPE avoidance fix

adamretter:
Closes https://github.com/eXist-db/exist/issues/3932

### DIFF
--- a/extensions/modules/cache/src/main/java/org/exist/xquery/modules/cache/Cache.java
+++ b/extensions/modules/cache/src/main/java/org/exist/xquery/modules/cache/Cache.java
@@ -57,7 +57,11 @@ class Cache {
 	}
 
     public Sequence put(final String key, final Sequence value) {
-	    return store.asMap().put(key, value);
+	    final Sequence previous = store.asMap().put(key, value);
+	    if (previous != null) {
+	        return previous;
+        }
+	    return Sequence.EMPTY_SEQUENCE;
     }
 
     public Sequence list(final String[] keys) throws XPathException {
--- a/extensions/modules/cache/src/test/java/org/exist/xquery/modules/cache/LazyCacheTest.java
+++ b/extensions/modules/cache/src/test/java/org/exist/xquery/modules/cache/LazyCacheTest.java
@@ -1,0 +1,78 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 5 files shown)


line-o (approved):
Looks good `cache:put` should under no circumstance return null which could lead to an NPE.
Also CI passes and more Tests are more good.
Related question: Is the "enableLazyCreation" option documented?

line-o:
The failing build on JDK11 is due to a request timeout. Should we restart the whole process?

line-o:
Static code analysis indicates that this not null check is unnecessary. Under which circumstances previous could end up being null?
https://sonarcloud.io/project/issues?id=eXist-db_exist&open=AXof8Tqp65lVRGWXCjZE&pullRequest=3935&resolved=false&types=CODE_SMELL

adamretter:
@line-o I am afraid that the static code analysis is wrong. Put on a Java Map can return null - https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#put-K-V-

line-o:
A failure in static code analysis is worrying.

adamretter:
@line-o It's not unusual. Static code analysis is often not perfect.

line-o:
When I inspect the code in question in IDEA I get the following

Screenshot 2021-06-22 at 16 02 11

That looks to me as if the map type returned by the cache implementation annotates the value with @nonnull
That might lead to the static code analysis "thinking" that there is no way a null value might be returned.
And you also mention it as being potentially return null, that means you never actually observed it returning null, correct?

adamretter:
> When I inspect the code in question in IDEA I get the following

@line-o I think your screenshot is showing that `asMap` cannot return `null`. The subsequent `put` can still return `null`.

line-o:
OK, I come to the same conclusion.
Can we annotate (or comment) this line in some form. So that, if the same line is touched again SonarCloud does not yell and we or others have the same discussion again?

adamretter:
@line-o SonarCloud has identified many false positives. I am not sure annotating them all makes sense...

dizzzz:
probably annotating is a lot of work indeed

dizzzz (approved):

dizzzz:
I'd trust to pull in... these 429's will just re-appear

line-o:
A quick search found that adding

    //NOSONAR

To any end of a line is the way to handle won't fix and false positives in SonarQube. So it is likely to work on SonarCloud as well.

line-o:
And also from what I now know I suspect it would be rather simple to create a test case that would provoke an NPE (and also prove that this is an issue).

dizzzz (approved):

line-o:
A simple

```
cache:put("this new key", 1)
```

will cause a NPE or it does not.
Or am I missing something obvious?

line-o:
```xquery
xquery version "3.1";

let $c := cache:create("a", map {})

return cache:put("a", "k", 1)
```
does not throw

line-o:
Is the lazy / non-lazy cache creation option documented?

adamretter:
> And also from what I now know I suspect it would be rather simple to create a test case that would provoke an NPE (and also prove that this is an issue).

Test are already included. If you remove the commit with the fix and re-run the tests you will see that they fail.

adamretter:
> Is the lazy / non-lazy cache creation option documented?

@line-o Yes, already documented, see - https://github.com/eXist-db/documentation/pull/614

line-o (approved):
I was able to provoke the NPE with

   [cache:put("a", "k2", 1)]?1

PR #3941, opened 2021-06-22
[hotfix] include templating in installer and tests

line-o:
apparently #3939 was not setting the new dependencies at two crucial locations

### DIFF
--- a/exist-installer/src/main/izpack/install.xml
+++ b/exist-installer/src/main/izpack/install.xml
@@ -164,13 +164,13 @@
 
         <pack name="Apps" required="false" installGroups="Group2">
             <description>Application Packages: additional applications for XQuery development, database management,
-                documentation, examples and more. You should at least install the "dashboard" and the "shared" package to
-                be able to add more applications later.
+                documentation, examples and more. You should at least install the "dashboard" and "packageservice" 
+                package to be able to add more applications later.
             </description>
         </pack>
 
-        <pack name="shared" required="true" preselected="true" parent="Apps">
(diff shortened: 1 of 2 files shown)


exist-installer/src/main/izpack/install.xml
```diff
@@ -172,7 +172,7 @@
         <pack name="shared" required="true" preselected="true" parent="Apps">
```
adamretter, line 172: The package name should be changed also here and where it is referenced, i.e. `name="shared"` is no longer correct.

exist-installer/src/main/izpack/install.xml
```diff
@@ -172,7 +172,7 @@
...
             <description>Shared Resources: jquery, dojo, ace, templating, utilities</description>
```
adamretter, line 173: Description should be updated as this is no longer correct

exist-installer/src/main/izpack/install.xml
adamretter: Also `You should at least install the "dashboard" and the "shared" package` above is no longer correct - please revise that too.

adamretter (changes requested):
Still some fixes needed.

exist-installer/src/main/izpack/install.xml
line-o: done

exist-installer/src/main/izpack/install.xml
line-o: done

line-o:
@adamretter thanks for catching some missing bits.
That also led me to find a dangling reference to the markdown package.

PR #3942, opened 2021-06-22
[feature] switched releasing via mvn from bintray to github-releases

windauer:
Replaces bintray mvn plugin with github-release-plugin by de.jutzig

### Description:

To create a draft release call: `mvn -DskipTests -Dgithub.draft=true github-release:release` in $EXIST/exist-distribution or $EXIST/exist-installer. 

The plugin is configured to be executed in the maven "deploy" phase which will be process during `mvn release:perform`.

Documentation has been added to https://github.com/eXist-db/exist/blob/develop/exist-versioning-release.md (see changed files). 

### Reference:
* https://github.com/jutzig/github-release-plugin

### Type of tests 
* none

### Known Issues
Missing tests. The plugin has been tested on other repositories and has successfully worked for eXist-db documentation (https://github.com/eXist-db/documentation/blob/master/pom.xml). **But** I have not setup a test env to check if this will really work on the first try. Running the plugin in `-Dgithub.draft=true` works perfectly fine but no artifacts are uploaded which might be due to "draft mode". My expectation is that the plugin will work fine and not fail, in case it misses to upload a resource we can always manually upload it to the Github release or better, fix the configuration of the plugin.

### DIFF
--- a/exist-distribution/pom.xml
+++ b/exist-distribution/pom.xml
@@ -1066,36 +1066,39 @@
                     </execution>
                 </executions>
             </plugin>
-
             <plugin>
-                <groupId>com.carrotgarden.maven</groupId>
-                <artifactId>bintray-maven-plugin</artifactId>
+                <groupId>de.jutzig</groupId>
+                <artifactId>github-release-plugin</artifactId>
                 <executions>
                     <execution>
-                        <id>upload-distributions-to-bintray</id>
+                        <id>github-upload</id>
                         <phase>deploy</phase>
                         <goals>
-                            <goal>upload</goal>
+                            <goal>release</goal>
                         </goals>
+                        <inherited>false</inherited>
                         <configuration>
-                            <restApiUrl>https://api.bintray.com</restApiUrl>
-                            <subject>existdb</subject>
-                            <repository>releases</repository>
-                            <bintrayPackage>exist</bintrayPackage>
-                            <targetFolder>/</targetFolder>
+                            <description>
+                                Release Notes: https://exist-db.org/exist/apps/wiki/blogs/eXist/exist${project.version}
 
-                            <packageLicenses>GNU Lesser General Public License, version 2.1</packageLicenses>
-
-                            <!-- Bintray API credentials must be in ~/.m2/settings.xml for server id -->
-                            <serverId>exist-bintray</serverId>
-
-                            <sourceFolder>${project.build.directory}</sourceFolder>
-                            <uploadRegex>.+/((${project.artifactId})|(eXist-db))-${project.version}((\.dmg)|(-unix\.tar\.bz2)|(-win\.zip))$</uploadRegex>
+                                Maven Central: https://search.maven.org/search?q=g:org.exist-db
+                            </description>
+                            <releaseName>eXist-db ${project.version}</releaseName>
+                            <tag>eXist-${project.version}</tag>
+                            <fileSets>
+                                <fileSet>
+                                    <directory>${project.build.directory}</directory>
+                                    <includes>
+                                        <include>eXist-db-${project.version}.dmg</include>
+                                        <include>${project.artifactId})-${project.version}-unix.tar.bz2</include>
+                                        <include>${project.artifactId})-${project.version}-win.zip</include>
+                                    </includes>
+                                </fileSet>
+                            </fileSets>
                         </configuration>
                     </execution>
                 </executions>
             </plugin>
-
         </plugins>
     </build>
 
--- a/exist-distribution/pom.xml
+++ b/exist-distribution/pom.xml
@@ -1090,8 +1090,8 @@
                                     <directory>${project.build.directory}</directory>
                                     <includes>
                                         <include>eXist-db-${project.version}.dmg</include>
-                                        <include>${project.artifactId})-${project.version}-unix.tar.bz2</include>
-                                        <include>${project.artifactId})-${project.version}-win.zip</include>
+                                        <include>${project.artifactId}-${project.version}-unix.tar.bz2</include>
+                                        <include>${project.artifactId}-${project.version}-win.zip</include>
                                     </includes>
                                 </fileSet>
                             </fileSets>
--- a/exist-installer/pom.xml
+++ b/exist-installer/pom.xml
@@ -135,36 +135,38 @@
                     <skip>true</skip>
                 </configuration>
             </plugin>
-
             <plugin>
-                <groupId>com.carrotgarden.maven</groupId>
-                <artifactId>bintray-maven-plugin</artifactId>
+                <groupId>de.jutzig</groupId>
+                <artifactId>github-release-plugin</artifactId>
+                <version>1.1.1</version>
                 <executions>
                     <execution>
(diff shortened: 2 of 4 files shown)


windauer:
to see if the release plugin really works as expected for eXist-db I could create a "eXist-db-test-release" by changing the configuration. That would be publicly visible till I delete it again but then we would know for sure that all files got uploaded and that the plugin does not run into problem because some artefacts are uploaded from exist-distribution and the other from exist-installer..

dizzzz (approved):

line-o:
At the community call it was consensus that if the plugin itself works when called alone the we can safely expect it to work in the release process.

adamretter:
> to see if the release plugin really works as expected for eXist-db I could create a "eXist-db-test-release" by changing the configuration

I would suggest to run the plugin standalone and check that it did upload your test artifacts to GitHub Releases, i.e. this will show that you have your `settings.xml` correct and the changes in your `pom.xml` correct. You can then remove the test artifacts from GitHub Releases.

As it stands, it looks to me like this won't work as something is missing from `settings.xml`, i.e. whilst you have the access token, the Github username appears to be missing. Is that not required?

line-o:
It looks as if the upload was successful
https://exist-db.slack.com/archives/CG4AMEF0W/p1624372648025000

exist-distribution/pom.xml
```diff
@@ -1066,36 +1066,39 @@
...
+                                Maven Central: https://search.maven.org/search?q=g:org.exist-db
+                            </description>
+                            <releaseName>eXist-db ${project.version}</releaseName>
+                            <tag>eXist-${project.version}</tag>
+                            <fileSets>
+                                <fileSet>
+                                    <directory>${project.build.directory}</directory>
+                                    <includes>
+                                        <include>eXist-db-${project.version}.dmg</include>
+                                        <include>${project.artifactId})-${project.version}-unix.tar.bz2</include>
```
adamretter, line 1093: Syntax here and below is wrong you have a superfluous ')'

exist-installer/pom.xml
```diff
@@ -135,36 +135,38 @@
                     <skip>true</skip>
                 </configuration>
             </plugin>
-
             <plugin>
-                <groupId>com.carrotgarden.maven</groupId>
-                <artifactId>bintray-maven-plugin</artifactId>
+                <groupId>de.jutzig</groupId>
+                <artifactId>github-release-plugin</artifactId>
+                <version>1.1.1</version>
```
adamretter, line 141: Version should not be here if it is defined in the parent pom.xml

exist-versioning-release.md
```diff
@@ -262,28 +261,28 @@ Central (staging), Docker images to Docker Hub, and eXist-db distributions and i
...
 
     6.1. Warning: there is a know issue in Atomic-Wiki where your release notes might suddenly disappear. In case this happens your data is not lost but stored in  /db/apps/wiki/data/blogs/eXist/.md. You can rename it or move the content to a eXistdb<VERSION>.md file and create an according eXistdb<VERSION>.atom for it. Once these two files are available the blog entry will become visible on the eXist-db homepage and it will be visible in the eXist-db blog. 
 
-7.  Visit the GitHub releases page [https://github.com/eXist-db/exist/releases](https://github.com/eXist-db/exist/releases) and create a new release, enter the tag you previously created and link the release notes from the blog and the binaries from BinTray.
+7.  Visit the GitHub releases page [https://github.com/eXist-db/exist/releases](https://github.com/eXist-db/exist/releases) and create a new release, enter the tag you previously created and link the release notes from the blog.
```
adamretter, line 283: I would not have thought you need to do this if you are using the GitHub Release plugin?

exist-versioning-release.md
```diff
@@ -119,11 +119,11 @@ You will require a system with:
 * A GPG key (for signing release artifacts)
 * A Java KeyStore with key (for signing IzPack Installer)
 * A valid Apple Developer Certificate (for signing Mac DMG)
+* A Github account and username / password or Github Personal access tokens (https://github.com/settings/tokens) with permission to publish Github releases to the eXist-db .org
 
 1. You will need login credentials for the eXist-db organisation on:
     1. Sonatype OSS staging for Maven Central - https://oss.sonatype.org/
-    2. Bintray - https://bintray.com/existdb
-    3. DockerHub - https://cloud.docker.com/orgs/existdb/
+    2. DockerHub - https://cloud.docker.com/orgs/existdb/
```
adamretter, line 126: Is it not worth mentioning that you need repo maintainer status on the eXist-db GitHub org (to replace the line about BinTray). I don't think just anyone can upload a release to GitHub, you need certain org/repo privileges I suspect.

adamretter (changes requested):
Some small fixes required.

adamretter:
> It looks as if the upload was successful

@line-o I am afraid that is a tag and not a release. Also I would avoid naming tags or releases with `-SNAPSHOT` suffix, I could imagine that could cause great pain with Maven's concept that a SNAPSHOT is a moving target.

line-o:
Those were clearly just tests.

exist-versioning-release.md
```diff
@@ -147,11 +147,10 @@ You will require a system with:
...
             </server>
 
-            <!-- eXist-db BinTray -->
+            <!-- eXist-db Github Release -->
             <server>
-                <id>exist-bintray</id>
-                <username>YOUR-USERNAME</username>
-                <password>YOUR-PASSWORD</password>
+                <id>github</id>
+                <privateKey>[Github Personal access tokens]</privateKey>
```
adamretter, line 153: What about the GitHub username? I don't think the token by itself is enough

adamretter:
> Those were clearly just tests.

Sure! I wasn't disputing that they were tests...

exist-versioning-release.md
windauer: a Github Userame is ont required but the <privateKey> is enough. Has been tested with several repos and works perfectly fine. Other way round, in case you have 2FactorAuth enabled (which we all have in the eXist-db github org) username / password will not work.

windauer:
> > to see if the release plugin really works as expected for eXist-db I could create a "eXist-db-test-release" by changing the configuration
> 
> I would suggest to run the plugin standalone and check that it did upload your test artifacts to GitHub Releases, i.e. this will show that you have your `settings.xml` correct and the changes in your `pom.xml` correct. You can then remove the test artifacts from GitHub Releases.
> 
> As it stands, it looks to me like this won't work as something is missing from `settings.xml` (or your documentation of it in the .md file), i.e. whilst you have the access token, the Github username appears to be missing. Is that not required?

like I wrote in `Known Issues` in the PR description the Plugin has been tested standalone and works fine. But for obvious reasons it was only tested in draft mode where the connection is checked but uploads are not performed, so there is a slight risk here that something wont work. Thanks for your finding of ')' cause that would have certainly not worked!

wolfgangmm:
It appears to me that all open questions were answered, so I would merge this if nobody shouts?

wolfgangmm:
Ok, no further comments since 5 hours. I'll thus assume it's good to go and we give it a try.

PR #3948, opened 2021-06-24
[hotfix] Repair version in installer build

dizzzz:
We missed a 5.3.1 version in one pom file :-/


line-o:
3 Modules were not updated properly in the prepare:release phase of the release attempt yesterday.
But this will be rolled back this morning and then we try again until this step is done automatically.
I will close this as I do not think manually patching this is the correct way forward.

line-o:
And thanks for the helping hand @dizzzz , always appreciated.

adamretter:
This should not be done this way. I will send some messages in Slack - but we need to roll this back

PR #3951, opened 2021-06-25
Only run the github-release-plugin when mvn release:perform is performed

adamretter:
This stoped the github-release-plugin trying to publish `SNAPSHOT`s to GitHub Releases. This should fix the problem with the Nightly Builds service.


windauer (approved):

line-o (approved):
I found the configuration that calls this profile in line 770 of exist-parent/pom.xml

`<releaseProfiles>exist-release</releaseProfiles>`

PR #3959, opened 2021-06-29
Update IntelliJ run configurations for 5.4.0-SNAPSHOT

adamretter:
These need to be updated to work with the new development version.


line-o:
@adamretter The current snapshot version is 5.3.1
I would assume these versions need to be in line, correct?

line-o (commented):
See my question above

adamretter:
> The current snapshot version is 5.3.1

Actually it is 5.4.0-SNAPSHOT, as I discussed with Lars and then fixed it.

dizzzz (approved):

line-o:
@adamretter I am wondering if this is OK to merge now?

adamretter:
@line-o Yes I would this so, as it won't affect a decision about whether a 5.3.1 is needed.

PR #3960, opened 2021-06-29
Make the Compatible Java Version Check more fine-grained

adamretter:
Previously only major versions were checked, however @StephanMa pointed out that the JDK issue that eXist-db suffers from is now fixed in 15.0.2.

JDK 15.0.2 should now pass the checks.

See https://github.com/eXist-db/exist/issues/3375

### DIFF
--- a/exist-start/pom.xml
+++ b/exist-start/pom.xml
@@ -45,6 +45,19 @@
       <tag>HEAD</tag>
     </scm>
 
+    <dependencies>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
     <build>
         <plugins>
             <plugin>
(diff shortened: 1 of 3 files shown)


dizzzz (commented):
From a communication perspective, this solution is IMO far too complex;
IMO most users do nut understand the last digit in versions and..
1) Java15 is EOL
2) Java15 is not LTS
I'd vote against it (though as the code is ready, probably we should just pull it in)

dizzzz (commented):
We  suggest users to use LTS version 8 / 11 ; non LTS versions should not be actively supported?

adamretter:
@dizzzz I only tackled it because @StephanMa reported it as an issue - https://github.com/eXist-db/exist/issues/3375#issuecomment-869582892

dizzzz:
yeas i understand. Dont get me wrong, the PR is highly appreciated (but stil wondering what issue is not actually solved, and what is now potentially pulled in)

dizzzz (approved):
Ok LGTM

adamretter:
@dizzzz @line-o I think the discussion about this change should happen here rather than https://github.com/eXist-db/exist/issues/3375#issuecomment-869582892

I don't really understand the concern. My feeling is that we are loosing sight of the purpose of these changes in the first place. So let me try and detail the situation as I understand it:

1. Before forbidding any Java versions (i.e. before this PR, and before the earlier PR https://github.com/eXist-db/exist/pull/3545 that added Java versions checks), you could attempt to run eXist-db on any version of Java you wanted. We suggested that people use Java 8, but this was not enforced or really documented widely. Likewise, we didn't really have a list of 'eXist-db Supported Java versions', but we did test a few different ones in CI.

2. I discovered a bug in certain Java versions - https://bugs.openjdk.java.net/browse/JDK-8253191 and myself and @reinhapa reported it to OpenJDK. This bug causes corruption in the data stream that eXist-db is writing to disk.

3. We added checks for "*Incompatible Java versions*" in PR https://github.com/eXist-db/exist/pull/3545. The purpose of this was to stop eXist-db starting if a version of Java was used which would trigger data corruption via Java bug  [JDK-8253191](https://bugs.openjdk.java.net/browse/JDK-8253191). This also reported a message: `we recommend using Java 8 or 11`.

4. This PR improves the "*Incompatible  Java versions*" check by excluding versions that we now know are now compatible with eXist-db; since our original checks, the Java bugfix has been backported from Java 16 to new releases of Java 15. Regardless, if an incompatible Java version is detected, the code will still report the message: `we recommend using Java 8 or 11`.

---
In summary, these changes have always been about blocking ("*incompatible*") versions of Java that will cause data corruption in eXist-db. With this PR the message that is reported to the user remains the same if there is a problem, but we do not block what is a perfectly good version(s) of Java.

These changes are **NOT** about only allowing Java versions that we consider "supported" by the eXist-db developers.
If we wanted to take that route instead, we can undo all these changes, and then just enforce that the user is using Java 8. I suspect we don't want that, but I am not opposed to it, and am happy to discuss...

I hope that is useful information when considering this PR.

adamretter:
@dizzzz Thank you very much Sir :-)

PR #3962, opened 2021-06-30
[doc] update versioning and release guide

### LINKED ISSUE #3845: Fix release process since migration from BinTray to GitHub Releases
The uploads to BinTray were automated as part of the release target see - https://github.com/eXist-db/exist/blob/develop/exist-versioning-release.md#preparing-a-product-release. I am not sure what we can do now, but also the `pom.xml` build files will need changing.

1. [x] Fix the release build
2. [ ] Update  https://github.com/eXist-db/exist/blob/develop/exist-versioning-release.md

### LINKED ISSUE #3914: [Master issue] Prepare eXist 5.3.0 stable release
- [x] Memory/Resource leak in XQuery Context Variables of Library Modules\* - https://github.com/eXist-db/exist/pull/3897
- [ ] Maps\* - a follow-on PR just for 5.3.0 to https://github.com/eXist-db/exist/pull/3738 (@adamretter please replace "3738" here with the new, split PR when it's ready)
- [x] Security\* - https://github.com/eXist-db/exist/issues/3505
- [x] Security - Finish and release\* https://github.com/eXist-db/templating
- [ ] Security - core+Saxon - add option to limit RFI in fn:doc etc\* - https://github.com/eXist-db/exist/issues/3927
- [ ] Security - ~~core\* - Apache XML-RPC CVE-2019-17570 / CVE-2016-5002 - https://github.com/eXist-db/exist/issues/3063 https://github.com/eXist-db/exist/pull/3934~~ cannot be merged because it introduces a breaking change
- [ ] ~~file:sync prune param\* - https://github.com/eXist-db/exist/pull/3084~~ still needs XQSuite tests
- [x] Review/merge PR updating crypto lib for compatibility with 5.3.0 - https://github.com/eXist-db/crypto-exist-java-lib/pull/37
- [x] Security: Update core apps to resolve security issues identified by dependabot
  - [x] dashboard\* - https://github.com/eXist-db/dashboard/security/dependabot
  - ~~[ ] docs\* - https://github.com/eXist-db/documentation/security/dependabot~~ - these are build warnings, not security issues with read-only running app
  - [x] eXide\* - https://github.com/eXist-db/eXide/security/dependabot
  - [x] monex\* - https://github.com/eXist-db/monex/security/dependabot
- [x] Update core apps to use [templating v1.0.0](https://github.com/eXist-db/templating):
  - [x] doc - https://github.com/eXist-db/documentation/pull/626
  - [x] fundocs - https://github.com/eXist-db/function-documentation/pull/41
  - [x] monex - https://github.com/eXist-db/monex/pull/153
- [x] Update core apps to remove all dependencies on shared-resources:
  - [x] doc - eXist-db/documentation#626
  - [x] eXide - https://github.com/eXist-db/eXide/pull/309
  - [x] fundocs - https://github.com/eXist-db/function-documentation/pull/42
  - [x] markdown - https://github.com/eXist-db/exist-markdown/pull/7
  - [x] monex - https://github.com/eXist-db/monex/pull/155
- [x] Publish the updated core apps to public-repo and GitHub Releases
  - [x] doc
  - [x] eXide
  - [x] fundocs
  - [x] markdown
  - [x] monex
- [x] Update release process for GitHub Releases\* - https://github.com/eXist-db/exist/issues/3845

Finally: 

- [x] Publish the release according to https://github.com/eXist-db/exist/blob/develop/exist-versioning-release.md#preparing-a-product-release
- [ ] Update exist-db.org with eXist 5.3.0 and updated core apps

Open Issue
- [x]  push 5.3.0 to master
- [x]  Go to GitHub and move all issues and PRs which are still open for the release milestone to the next release milestone and close the milestones 5.3.0 and 5.2.1
- [x]  merge open PR and release crypto-lib 6.0.0
- [x]  update homebrew
- [ ]  publish updated “howto release” document #3962
- [ ]  Github-release-plugin: fix link to release notes and improve layout
- [x]  Wikipedia page: Lars asked someone from outside the community to update it. This might take a moment.

Note: Tasks marked with **\*** are drawn from @adamretter's [list from Slack](https://exist-db.slack.com/archives/CG2MRUZ35/p1587999543479300?thread_ts=1587999239.476800&cid=CG2MRUZ35).

line-o:
closes #3845
closes #3914 
Follow-up on the last release the guide how-to release a new version of exist-db was reworked.

### DIFF
--- a/exist-versioning-release.md
+++ b/exist-versioning-release.md
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
 
 ## Release Process
 
-This section details concrete steps for creating and publishing product releases. Each section here assumes you are starting with a clean Git checkout of the `develop` branch from [https://github.com/eXist-db/exist.git](https://github.com/eXist-db/exist.git).
+This section details concrete steps for creating and publishing product releases. 
 
-### Initiating Semantic Versioning
+### Release preprations and prerequisites 
 
-Version 3.0.0 was released before Semantic Versioning. The following steps will initiate Semantic Versioning for the remainder of the development phase of the next release, a new minor version to be called version 3.1.0:
-
-1.  Modify `$EXIST_HOME/build.properties` to read:
-
-    ```
-    project.version = 3.1.0-SNAPSHOT
-    ```
-
-2.  Commit the changes and push to `origin` (or `upstream` if you are on a fork).
-
-### Preparing a Product Release
-
+#### Required Software 
 Once development on a new stable version is complete, the following steps will prepare the version for release. For purposes of illustration, we will assume we are preparing the stable release of version 5.3.0.
 You will require a system with:
+
 * macOS
 * JDK 8
 * Maven 3.6.0+
 * Docker
 * GnuPG
-* A GPG key (for signing release artifacts)
-* A Java KeyStore with key (for signing IzPack Installer)
-* A valid Apple Developer Certificate (for signing Mac DMG)
-* A Github account and username / password or Github Personal access tokens (https://github.com/settings/tokens) with permission to publish Github releases to the eXist-db .org
 
-1. You will need login credentials for the eXist-db organisation on:
-    1. Sonatype OSS staging for Maven Central - https://oss.sonatype.org/
-    2. DockerHub - https://cloud.docker.com/orgs/existdb/
+#### Required Keys, Credentials and Permissions
+
+##### GPG Key / Keystore and Credentials
+A private GPG key for signing release artifacts is required. To gather the required information run `gpg --list-secret-keys --keyid-format SHORT`
+
+```
+gpg --list-secret-keys --keyid-format SHORT
+/Users/username/.gnupg/pubring.kbx
+----------------------------------
+sec   rsa4096/12345678 2013-04-26 [SC]
+```
+
+This gives you the keystore you are using (here: `/Users/username/.gnupg/pubring.kbx`) and your key-id (here: `12345678` the part after rsa4096/). 
+
+**Important Note** 
+GnuPG is using an [optimized keystore since version 2.1](https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration.html) which contains public and secret keys. Former GnuPG version had seperated keystore for public (~/.gnupg/pubring.gpg) and private keys (~/.gnupg/secring.gpg). 
+
+###### Maven Keystore values prior GnuPG 2.1
+
+* existdb.release.public-keyfile=~/.gnupg/pubring.gpg
+* existdb.release.private-keyfile=~/.gnupg/secring.gpg
+
+###### Maven Keystore values for GnuPG 2.1 or newer
+
+* existdb.release.public-keyfile=~/.gnupg/pubring.kbx
+* existdb.release.private-keyfile=~/.gnupg/pubring.kbx
+                   
+To test gpg simply create a file `test.txt` and run`gpg -ab test.txt`to sign it. you should see a new file `test.txt.asc` which means GnuPG signing worked fine. 
+
+##### Java KeyStore 
+* To create an eXist-db release a Java KeyStore with key is needed to sign the IzPack Installer. In case you don't have a Java KeyStore you can create one with: 
+
+```bash
+keytool -genkey \
+  -alias some-alias \
+  -keystore keystore.p12 \
+  -storetype PKCS12 \
+  -keyalg RSA \
+  -storepass some-pass \
+  -validity 7300 \
+  -keysize 4096
+```
+
+To verify if your Java KeyStore works fine run 
+
+```bash
+keytool -list -v -keystore /path/to/keystore.p12
+```
+
+##### Apple Developer Certificate
+
+A valid Apple Developer Certificate to sign Mac DMG is required. If you don't have one you will need to join the [Apple Developer Program](https://developer.apple.com/programs/enroll/) and to create a certificate via XCode. 
+
+##### Github
+
+A Github account and username / password or Github Personal (access tokens)[https://github.com/settings/tokens] with permission to publish Github releases to the eXist-db org. 
+
+##### Sonatype OSS
+
+You will need a Sonatype OSS acount which has been authorized for the eXist-db organisation at Sonatype OSS staging for Maven Central - https://oss.sonatype.org/
+
+
+
+##### Docker 
+
+A DockerHub user with permissions to publish to https://cloud.docker.com/orgs/existdb/ is required. 
+
+
+#### Prepare Release Notes 
+
+* There are various possibilities to create the release notes.
+
+> This section is still work in progress and manual tasks are needed to generate the release notes
+
+The release notes start with an introduction summarizing the most important changes for end-users and developers since the last version was released.
+
+Three sections follow
+
+- a list of all changes and features with links to their PRs
+- a list of all fixes with links the their PRs
+- a list of all dependencies that got updated along with the new version that is now used in existdb
+
+##### gitchangelog (Python) 
+
+* https://pypi.org/project/gitchangelog/
+* configure .gitchangelog.rc in the root of eXist-db repo
+* run `gitchangelog` in the eXist-db git repo. 
+* the list of dependency updates can be prepared with
+    ```
+    gitchangelog eXist-5.3.0..HEAD | grep "Updates" | sort > dependency-updates.txt
+    ```
+    Replace **eXist-5.3.0** with the tag of the latest release.
+
+
+### Preparing a Product Release
+
+#### Maven settings.xml
+
+Required credentials and the Maven `existdb-release-signing` profile should be stored securely in the `<servers>` section on your machine in your local `~/.m2/settings.xml` file. The contents should follow the example below.
     
-    Your credentials for these should be stored securely in the `<servers`> section on your machine in your local `~/.m2/settings.xml` file, e.g.:
-    ```xml
+```xml
     <settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
           xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
@@ -150,48 +231,72 @@ You will require a system with:
             <!-- eXist-db Github Release -->
             <server>
                 <id>github</id>
-                <privateKey>[Github Personal access tokens]</privateKey>
+                <privateKey>YOUR-Github-ACCESS-TOKEN</privateKey>
             </server>
         </servers>
+        
+        <profiles>
+           <profile>
+               <id>existdb-release-signing</id>
+               <properties>                   
+                      
+                   <existdb.release.key>YOUR-GPG-KEY-ID</existdb.release.key>
+                   <existdb.release.public-keyfile>/your/part/to/.gnupg/pubring.gpg</existdb.release.public-keyfile>
+                   <existdb.release.private-keyfile>/your/part/to/.gnupg/secring.gpg</existdb.release.private-keyfile>
+                   <existdb.release.key.passphrase>YOUR-GPG-PASSWORD</existdb.release.key.passphrase>
+
+                   <existdb.release.keystore>/your/path/to/.keystore</existdb.release.keystore>
+                   <existdb.release.keystore.pass>YOUR-KEYSTORE-PASSWORD</existdb.release.keystore.pass>
+                   <existdb.release.keystore.key.alias>YOUR-KEYSTORE-ALIAS</existdb.release.keystore.key.alias>
+                   <existdb.release.keystore.key.pass>YOUR-KEYSTORE-PASSWORD</existdb.release.keystore.key.pass>
+               </properties>
+           </profile>
+        </profiles>
+        <activeProfiles>
+            <activeProfile>existdb-release-signing</activeProfile>
+        </activeProfiles>
     </settings>
+```
+
+1. Merge any outstanding PRs that have been reviewed and accepted for the milestone eXist-5.3.0.
+
+2. Create a new clone of the branch you want to release from into a new folder.
+   In the example below the `develop` branch is cloned into the folder `release`
+
+
+    ```bash
+    git clone \ 
+      -b develop \
+      git@github.com:eXist-db/exist.git \
+      release
+    ```
+    
+3. switch to that directory
+
+    ```bash
+    cd release
     ```
 
-2. You will need your GPG Key and Java KeyStore credentials for signing the release artifacts in the `<activeProfiles`> section on your machine in your local `~/.m2/settings.xml` file, e.g.:
-    ```xml
-    <profiles>
-   
-       <profile>
-           <id>existdb-release-signing</id>
-           <properties>
-               <existdb.release.key>ABC1234</existdb.release.key>
-               <existdb.release.public-keyfile>${user.home}/.gnupg/pubring.gpg</existdb.release.public-keyfile>
-               <existdb.release.private-keyfile>${user.home}/.gnupg/secring.gpg</existdb.release.private-keyfile>
-               <existdb.release.key.passphrase>your-password</existdb.release.key.passphrase>
-   
-               <existdb.release.keystore>${user.home}/your.store</existdb.release.keystore>
-               <existdb.release.keystore.pass>your-keystore-password</existdb.release.keystore.pass>
-               <existdb.release.keystore.key.alias>your-alias</existdb.release.keystore.key.alias>
-               <existdb.release.keystore.key.pass>your-key-password</existdb.release.keystore.key.pass>
-           </properties>
-       </profile>
-   
-    </profiles>
-
-
-    <activeProfiles>
-   
-           <activeProfile>existdb-release-signing</activeProfile>
-   
-    </activeProfiles>
+    You can verify that origin now points to the main repository `git@github.com:eXist-db/exist.git`
+    
+    ```bash
+    git remote -v
     ```
 
-3.  Merge any outstanding PRs that have been reviewed and accepted for the milestone eXist-5.3.0.
 
-4.  Make sure that you have the HEAD of `origin/develop` (or `upstream` if you are on a fork).
 
-5.  Prepare the release, if you wish you can do a dry-run first by specifiying `-DdryRun=true`:
-    ```
-    $ mvn -Ddocker=true -Dmac-signing=true -Dizpack-signing=true -Darguments="-Ddocker=true -Dmac-signing=true -Dizpack-signing=true" release:prepare
+4.  Prepare the release
+    **NOTE:** If this is the first time you prepare a release or you made changes to the process you should do a dry-run first by setting `-DdryRun=true`
+    With `-V` you can also check the maven and JDK version that will be used
+
+    ```bash
+    mvn \
+      -Pdocker \
+      -Pinstaller \
+      -Pconcurrency-stress-tests \
+      -Pmicro-benchmarks \
+      -Darguments="-Ddocker=true -Dmac-signing=true -Dizpack-signing=true -Dmac.codesign.identity=\"your codesigning ID\" \
+      release:prepare
     ```
     
     Maven will start the release process and prompt you for any information that it requires, for example:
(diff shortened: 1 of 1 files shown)


exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
 
 ## Release Process
 
-This section details concrete steps for creating and publishing product releases. Each section here assumes you are starting with a clean Git checkout of the `develop` branch from [https://github.com/eXist-db/exist.git](https://github.com/eXist-db/exist.git).
```
adamretter, line 96: As discussed in the calls, I believe it is important to always start a release from a clean `git checkout` of the `develop` branch. So I think this section should not be removed

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
+* GnuPG
+
+#### Required Keys, Credentials and Permissions
+
+##### GPG Key / Keystore and Credentials
+A private GPG key for signing release artifacts is required. To gather the required information run `gpg --list-secret-keys --keyid-format SHORT`
+
+```
+gpg --list-secret-keys --keyid-format SHORT
+/Users/username/.gnupg/pubring.kbx
```
adamretter, line 117: I think we need to be more generic here. On none of my machines are the files for GPG named like the ones here!

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
+----------------------------------
+sec   rsa4096/12345678 2013-04-26 [SC]
+```
+
+This gives you the keystore you are using (here: `/Users/username/.gnupg/pubring.kbx`) and your key-id (here: `12345678` the part after rsa4096/).
```
adamretter, line 122: I think `keystore` is the wrong terminology here. `keystore` is used in the Java world. For GPG I believe the correct terminology is "public key ring"

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
+gpg --list-secret-keys --keyid-format SHORT
+/Users/username/.gnupg/pubring.kbx
+----------------------------------
+sec   rsa4096/12345678 2013-04-26 [SC]
+```
+
+This gives you the keystore you are using (here: `/Users/username/.gnupg/pubring.kbx`) and your key-id (here: `12345678` the part after rsa4096/). 
+
+**Important Note** 
+GnuPG is using an [optimized keystore since version 2.1](https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration.html) which contains public and secret keys. Former GnuPG version had seperated keystore for public (~/.gnupg/pubring.gpg) and private keys (~/.gnupg/secring.gpg).
```
adamretter, line 125: This should be moved before the commands (see comment above).

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
+
+###### Maven Keystore values for GnuPG 2.1 or newer
+
+* existdb.release.public-keyfile=~/.gnupg/pubring.kbx
+* existdb.release.private-keyfile=~/.gnupg/pubring.kbx
+                   
+To test gpg simply create a file `test.txt` and run`gpg -ab test.txt`to sign it. you should see a new file `test.txt.asc` which means GnuPG signing worked fine. 
+
+##### Java KeyStore 
+* To create an eXist-db release a Java KeyStore with key is needed to sign the IzPack Installer. In case you don't have a Java KeyStore you can create one with:
```
adamretter, line 140: `you can create one with` - we should add some text, that it is important that the user realises this should be an exceptional circumstance and not the normal, because if they use a self-generated key, then there is no authenticity. Ideally the user should use a real key purchased from a code-signing vendor.

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
+* macOS
+* JDK 8
+* Maven 3.6.0+
+* Docker
+* GnuPG
+
+#### Required Keys, Credentials and Permissions
+
+##### GPG Key / Keystore and Credentials
+A private GPG key for signing release artifacts is required. To gather the required information run `gpg --list-secret-keys --keyid-format SHORT`
```
adamretter, line 113: "A private GPG key for signing release artifacts" should be changed to:

"A Public/Private PGP key pair is needed for signing release artifacts. Such a key/pair can be generated with GPG if desired. Regardless, the public key MUST have been published to the public key server network, so that it may be verified by Maven Central when the artifacts are published".

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
+To verify if your Java KeyStore works fine run 
+
+```bash
+keytool -list -v -keystore /path/to/keystore.p12
+```
 
-1.  Modify `$EXIST_HOME/build.properties` to read:
+##### Apple Developer Certificate
 
+A valid Apple Developer Certificate to sign Mac DMG is required. If you don't have one you will need to join the [Apple Developer Program](https://developer.apple.com/programs/enroll/) and to create a certificate via XCode.
```
adamretter, line 161: `to sign Mac DMG is` -> `to sign both the Mac App and Mac DMG is`

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
+
+> This section is still work in progress and manual tasks are needed to generate the release notes
+
+The release notes start with an introduction summarizing the most important changes for end-users and developers since the last version was released.
+
+Three sections follow
+
+- a list of all changes and features with links to their PRs
+- a list of all fixes with links the their PRs
+- a list of all dependencies that got updated along with the new version that is now used in existdb
```
adamretter, line 190: Maybe a note to say that this may need to be deduplicated as the same dependency may have been updated more than once between releases. It is important to list both the version from the last release, and the newest version that goes into this release.

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
+
+##### gitchangelog (Python) 
+
+* https://pypi.org/project/gitchangelog/
+* configure .gitchangelog.rc in the root of eXist-db repo
+* run `gitchangelog` in the eXist-db git repo. 
+* the list of dependency updates can be prepared with
     ```
-    project.version = 3.1.0-SNAPSHOT
+    gitchangelog eXist-5.3.0..HEAD | grep "Updates" | sort > dependency-updates.txt
```
adamretter, line 199: Above the notes say that we are doing this for release `eXist-5.3.0`, so the above should be updated to reflect that, i.e `gitchangelog eXist-5.2.0..HEAD`

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
     ```
+    Replace **eXist-5.3.0** with the tag of the latest release.
```
adamretter, line 201: Replace **eXist-5.2.0** with the tag of the previous release.

exist-versioning-release.md
```diff
@@ -93,40 +93,121 @@ That version number is also copied into the `META-INF/MANIFEST.MF` file of any J
...
-* A Java KeyStore with key (for signing IzPack Installer)
-* A valid Apple Developer Certificate (for signing Mac DMG)
-* A Github account and username / password or Github Personal access tokens (https://github.com/settings/tokens) with permission to publish Github releases to the eXist-db .org
-
-1. You will need login credentials for the eXist-db organisation on:
-    1. Sonatype OSS staging for Maven Central - https://oss.sonatype.org/
-    2. DockerHub - https://cloud.docker.com/orgs/existdb/
+#### Maven settings.xml
+
+Required credentials and the Maven `existdb-release-signing` profile should be stored securely in the `<servers>` section on your machine in your local `~/.m2/settings.xml` file. The contents should follow the example below.
```
adamretter, line 208: This needs rewording - the `existdb-release-signing` profile does NOT go in the `<servers>` section

exist-versioning-release.md
```diff
@@ -150,48 +231,72 @@ You will require a system with:
             <!-- eXist-db Github Release -->
             <server>
                 <id>github</id>
-                <privateKey>[Github Personal access tokens]</privateKey>
+                <privateKey>YOUR-Github-ACCESS-TOKEN</privateKey>
             </server>
         </servers>
+
```
adamretter, line 237: The other servers are missing from this, including:
*  `sonatype-nexus-staging`
*  `exist-db-snapshots`
* `docker.io`

exist-versioning-release.md
```diff
@@ -150,48 +231,72 @@ You will require a system with:
...
             </server>
         </servers>
+        
+        <profiles>
+           <profile>
+               <id>existdb-release-signing</id>
+               <properties>                   
+                      
+                   <existdb.release.key>YOUR-GPG-KEY-ID</existdb.release.key>
+                   <existdb.release.public-keyfile>/your/part/to/.gnupg/pubring.gpg</existdb.release.public-keyfile>
```
adamretter, line 244: `/your/part` -> `/your/path`

exist-versioning-release.md
```diff
@@ -150,48 +231,72 @@ You will require a system with:
...
+                   <existdb.release.private-keyfile>/your/part/to/.gnupg/secring.gpg</existdb.release.private-keyfile>
+                   <existdb.release.key.passphrase>YOUR-GPG-PASSWORD</existdb.release.key.passphrase>
+
+                   <existdb.release.keystore>/your/path/to/.keystore</existdb.release.keystore>
+                   <existdb.release.keystore.pass>YOUR-KEYSTORE-PASSWORD</existdb.release.keystore.pass>
+                   <existdb.release.keystore.key.alias>YOUR-KEYSTORE-ALIAS</existdb.release.keystore.key.alias>
+                   <existdb.release.keystore.key.pass>YOUR-KEYSTORE-PASSWORD</existdb.release.keystore.key.pass>
```
adamretter, line 251: `YOUR-KEYSTORE-PASSWORD` -> `YOUR-KEY-PASSWORD`

adamretter: These are not necessarily the same. They are the same **if** you are using PKCS12 format, but not all codesigning certificates use that format, legacy keystore for example uses jks format.

exist-versioning-release.md
```diff
@@ -150,48 +231,72 @@ You will require a system with:
...
+
+1. Merge any outstanding PRs that have been reviewed and accepted for the milestone eXist-5.3.0.
+
+2. Create a new clone of the branch you want to release from into a new folder.
+   In the example below the `develop` branch is cloned into the folder `release`
+
+
+    ```bash
+    git clone \ 
+      -b develop \
```
adamretter, line 269: Instead of `-b developer` consider the faster options `--single-branch --branch develop`

exist-versioning-release.md
```diff
@@ -150,48 +231,72 @@ You will require a system with:
...
+      git@github.com:eXist-db/exist.git \
```
adamretter, line 270: Should likely use the HTTPS URLs here as that is what the release process is geared up for, i.e. see the SCM section in the `pom.xml`. If you insist on sticking with SSH here then you need an earlier section stating that you need a public SSH key registered with GitHub account and exist-db org.

exist-versioning-release.md
```diff
@@ -150,48 +231,72 @@ You will require a system with:
...
-   
-    </profiles>
-
-
-    <activeProfiles>
-   
-           <activeProfile>existdb-release-signing</activeProfile>
-   
-    </activeProfiles>
+    You can verify that origin now points to the main repository `git@github.com:eXist-db/exist.git`
```
adamretter, line 280: HTTPS URLs please

exist-versioning-release.md
adamretter: Also HTTPS is more likely to work for anyone working from within a constrained/fire-walled environment

exist-versioning-release.md
```diff
@@ -150,48 +231,72 @@ You will require a system with:
...
 
-3.  Merge any outstanding PRs that have been reviewed and accepted for the milestone eXist-5.3.0.
 
-4.  Make sure that you have the HEAD of `origin/develop` (or `upstream` if you are on a fork).
 
-5.  Prepare the release, if you wish you can do a dry-run first by specifiying `-DdryRun=true`:
-    ```
-    $ mvn -Ddocker=true -Dmac-signing=true -Dizpack-signing=true -Darguments="-Ddocker=true -Dmac-signing=true -Dizpack-signing=true" release:prepare
+4.  Prepare the release
+    **NOTE:** If this is the first time you prepare a release or you made changes to the process you should do a dry-run first by setting `-DdryRun=true`
```
adamretter, line 289: This should be reworded.

The release process is complex, you SHOULD ALWAYS do a dry-run first, regardless of how many times you have done this before. This saves having to undo bad release processes that stop 1/2 way through because perhaps you upgraded/removed changed GPG or some other external dependency.

exist-versioning-release.md
```diff
@@ -150,48 +231,72 @@ You will require a system with:
...
+    With `-V` you can also check the maven and JDK version that will be used
+
+    ```bash
+    mvn \
+      -Pdocker \
+      -Pinstaller \
+      -Pconcurrency-stress-tests \
+      -Pmicro-benchmarks \
+      -Darguments="-Ddocker=true -Dmac-signing=true -Dizpack-signing=true -Dmac.codesign.identity=\"your codesigning ID\" \
```
adamretter, line 298: The `-P` args should likely also appear in `-Darguments`

adamretter: The codesign identity could be moved into `settings.xml` which would simplify this.

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
     What is the new development version for "eXist-db"? (org.exist-db:exist) 5.4.0-SNAPSHOT: :
```
adamretter, line 313: @windauer mentioned we should make it more explicit in the docs that the next version should always be a feature version. Do we want to expand the text here?

Maven by default will prompt you with the next patch version, i.e.:
```
What is the new development version for "eXist-db"? (org.exist-db:exist) 5.3.1-SNAPSHOT: :
```
So the docs should be updated here so the Maven prompt is correct. Even if you should reject its suggestion and then input the next feature version SNAPSHOT

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
     ```
 
-6.  Once the prepare process completes you can perform the release. This will upload Maven Artifacts to Maven
-Central (staging), Docker images to Docker Hub, and eXist-db distributions and installer to Github releases:
-    ```
-    $ mvn -Ddocker=true -Dmac-signing=true -Djarsigner.skip=false -Darguments="-Ddocker=true -Dmac-signing=true -Djarsigner.skip=false" release:perform
-    ```
+5. Test the release artefacts as much as possible on different Operating systems (Mac, Windows, Linux).
```
adamretter, line 316: This testing is happening at the wrong time. This is far too late in the overall process.

Once you have merged any outstanding PRs for the upcoming release, and then believe that the code is stable. You can generate a SNAPSHOT release which you can share between colleagues and test. That should be done before any release process starts. Almost always some issue is found, which should then be fixed in a new PR, and then new SNAPSHOT generated, shared and tested again.

In the past Wolfgang, Dannes, and myself, would do extensive testing before any release process was started.

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
+
+   * `$EXIST/exist-distribution/target/` 
+     * eXist-db-<version>.dmg
+     * exist-distribution-eXist-db-<version>.bz2
+     * exist-distribution-eXist-db-<version>-win.zip
+   * `$EXIST/exist-installer-<version>.jar` 
+
+#### Signing Binaries (work in progress)
+
+In order to allow end-users to run the installer as well as the application on macOSX (Catalina) the application needs to be notarized by Apple.
```
adamretter, line 326: Reword this to something like:
```
In order to allow end-users to open the DMG, and then later the eXist-db application on macOS without any security blocks, both the application and the DMG need to be signed, notarised by Apple, and then stapled.
```

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
+    
+At this step the application is ready to be notarized and the notarisation ticket must be stapled to the app once that was successful.
```
adamretter, line 328: This is happening at the wrong stage. Notarization and Stapling needs to happen as part of `mvn release:perform`. It should be performed during the `package` phase of the lifecycle. This can then be checked to make sure it succeeded during the `verify` phase of the lifecycle.

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
+      -Pconcurrency-stress-tests \
+      -Pmicro-benchmarks \
+      release:clean
+
+##### Rollback Release
+
+If you need to rollback the `mvn release:prepare` phase:
 
-7.  Update the stable branch (`master`) of eXist-db to reflect the latest release:
+`mvn release:rollback` works but
```
adamretter, line 345: I think we should write/discuss a bit more about this. I think that `mvn release:rollback` should never be used in practice...

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
+    
+1. Once the prepare process completes successfully and artifacts have been tested you can perform the release. This will upload Maven Artifacts to Maven Central (staging), Docker images to Docker Hub, and eXist-db distributions and installer to Github releases. 
+    
+    ```bash
+    mvn \
+      -Pdocker \
+      -Pinstaller \
+      -Pconcurrency-stress-tests \
+      -Pmicro-benchmarks \
+      -Darguments="-Djarsigner.skip=false -Dmac-signing=true -Dmac.codesign.identity=\"your codesign ID\"" \
```
adamretter, line 361: The `-P` args should likely be repeated in the `-Dargument`

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
+      release:perform
     ```
-    $ git push origin eXist-5.3.0:master
+
+    #### Notarizing macOSX installer (work in progress)
```
adamretter, line 365: This seems to be in the wrong place. Also see my comments above about Notarization and where it needs to happen...

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
+    
+    Now the installer in `target/checkout/exist-distribution/eXist-5.3.0.dmg` needs to be notarized and the notarisation ticket must be stapled to the installer once that was successful.
+    The notarized installer containing the notarized app must then be uploaded to the github releases page.
+
+    
+2. Update the stable branch (`master`) of eXist-db to reflect the latest release:
```
adamretter, line 371: Before this... A visual checklist of things needs to be completed. Was the output of the `release:perform` step as expected. Do the new version numbers look correct locally? do they look correct on the exist-db GitHub? Does the tag look correct, did the tag get pushed to the exist-db GitHub? Has at least one other person also checked such things and also confirmed this to you?

Also before modifying `master`, I would suggest that the oss.sonatype.org artifacts have to be released successfully. Sonatype has its own suit of checks, and if the artifacts fail those for some reason, then you have to undo the release process, fix the issue, go all the way back to the start of the release process and start again. So it is important that Sonatatype artifacts are published OK before we change master. Changing master should really be the last thing that ever happens (apart from perhaps tweeting/emailing/etc)

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
+    2.1. Warning: there is a known issue in AtomicWiki where your release notes might suddenly disappear. In case this happens your data is not lost but stored in `/db/apps/wiki/data/blogs/eXist/.md`. You can rename it or move the content to a `eXistdb<VERSION>.md` file and create an according `eXistdb<VERSION>.atom` for it. Once these two files are available the blog entry will become visible on the eXist-db homepage and it will be visible in the eXist-db blog. 
 
-7.  Visit the GitHub releases page [https://github.com/eXist-db/exist/releases](https://github.com/eXist-db/exist/releases) and create a new release, enter the tag you previously created and link the release notes from the blog.
+3.  Visit the GitHub releases page [https://github.com/eXist-db/exist/releases](https://github.com/eXist-db/exist/releases) and create a new release, enter the tag you previously created and link the release notes from the blog.
```
adamretter, line 454: Does this still need to be done? doesn't the github-release-plugin now take care of some of this?

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
 
-11. Submit a news item to XML.com - [https://www.xml.com/news/submit-news-item/](https://www.xml.com/news/submit-news-item/).
+7. Submit a news item to XML.com - [https://www.xml.com/news/submit-news-item/](https://www.xml.com/news/submit-news-item/).
+  * use tags: database, nosql, xml, exist-db, xquery, xslt         
 
-12. Update the Wikipedia page with the new version details - [https://en.wikipedia.org/wiki/EXist](https://en.wikipedia.org/wiki/EXist).
+8. Update the Wikipedia page with the new version details - [https://en.wikipedia.org/wiki/EXist](https://en.wikipedia.org/wiki/EXist).
 
-13. Go to GitHub and move all issues and PRs which are still open for the release milestone to the next release milestone. Close the release milestone.
+9. Go to GitHub and move all issues and PRs which are still open for the release milestone to the next release milestone. Close the release milestone.
```
adamretter, line 467: We need to expand on this as it caused confusion this time... if the release was a "patch" move the issues to the next logical "patch" milestone, likewise for "feature" and "minor" releases.

adamretter (changes requested):
Good start! Needs some expansion and reordering of tasks

exist-versioning-release.md
line-o: Yes, this is why we decided to make this a dedicated step. That way there is less chance to miss it.

exist-versioning-release.md
line-o: I did use `--single-branch` and yes that is much faster. Do you happen to know if the later commands will still work? Especially `git checkout -b next-master eXist-5.3.0`. If that is the case I will re-add that.
Do you also want to say that long parameter names are preferred (`-b` is `--branch`)

exist-versioning-release.md
adamretter: Thanks, I see that below now, but I do think it is better at the beginning or at least also a note at the beginning.

exist-versioning-release.md
line-o: The interesting thing is 2FA here. Also as I really never use HTTPS, I really can't tell how that will end up working. Wouldn't you then need to tell people to create an access token and use that here?

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
+    Now the installer in `target/checkout/exist-distribution/eXist-5.3.0.dmg` needs to be notarized and the notarisation ticket must be stapled to the installer once that was successful.
+    The notarized installer containing the notarized app must then be uploaded to the github releases page.
+
+    
+2. Update the stable branch (`master`) of eXist-db to reflect the latest release:
+    
+    
+    ```bash
+    git checkout -b next-master eXist-5.3.0
+    git push origin next-master:master
```
adamretter, line 376: You don't need the intermediate branch, you can push the tag (either local or remote) directly to the remote master,

exist-versioning-release.md
line-o: The above is sample output. That is also the more recent format as is layed out in a previous section.

exist-versioning-release.md
line-o: I tried that and it did not work.

exist-versioning-release.md
adamretter: I think the `next-master` bit is superfluous - I just added a comment about that.

>Do you also want to say that long parameter names are preferred

I like to see long names in tutorials because then the parameter is not opaque, i.e. the reader can pretty much understand what it means without having to find the manual page and lookup `-b` to see what it stands for.

exist-versioning-release.md
line-o: It might work but as it it also might fail I would prefer the safer option.

exist-versioning-release.md
windauer: agreed, my miss take, will fix it

exist-versioning-release.md
line-o: Good catch. Will be removed

exist-versioning-release.md
windauer: Agreed, once its above the sample I think the discussion which keyring we show in the sample is not so important anymore, or is it? Point is since GnuPG 2.1 is .kbx thats why I would vote for the .kbx sample but I don't have strong feeling about it.. @adamretter what is printed if you run the command, only the pubring.gpg or secring.gpg and pubring.gpg ?

exist-versioning-release.md
line-o: Interesting. This sounds like very important information that belongs in this guide.

exist-versioning-release.md
line-o: Not needed. I spent Saturday night running this back and forth.

exist-versioning-release.md
adamretter: We have always used HTTPS even with 2FA in the past. It's very simple, use your same GitHub username, but instead of your "password" you just paste in a PAC in its place. If you have 2FA enabled (which eXist-db do) and you are working with HTTPS, which is the default (and I know other eXist-db developers also use it) - then you will almost certainly always be working with a PAC that you previously generated (and maybe stored in your Apple KeyChain so you don't have to remember it ;-)). But sure, a bit about needing a PAC up-front might make sense - I don't know if people need that much hand-holding then, I think we can assume that developers have knowledge of how to use GitHub.

exist-versioning-release.md
adamretter: > I tried that and it did not work.

 Tried what? Unless you share the commands, and the error, I can't comment. All I can tell you is that is how it has been done for previous released...

adamretter: > It might work but as it it also might fail I would prefer the safer option.

Git is atomic, it will either succeed or fail, and it will always do so in a safe-way

exist-versioning-release.md
line-o: Yes, I agree that this should be added in greater detail.

exist-versioning-release.md
line-o: `git push origin eXist-5.3.0:master` resulted in  `eXist-5.3.0 -> master(bad ref)` on a fresh clone after building

exist-versioning-release.md
line-o: As I understand it the .app is created in `mvn release:prepare` whereas the .dmg is an artifact created in `mvn release:perform`.

exist-versioning-release.md
line-o: Sonatype-nexus-staging and docker.io are left intact and are therefore not shown in this diff.
exist-db-snapshots is not needed for a product release according to you and thus left out.

exist-versioning-release.md
line-o: Would it be enough to state at the beginning only PKCS12 keystores are considered in this guide?

exist-versioning-release.md
line-o: As noted down below the duplication is not necessary.
Would you be willing to share the snippet how to set the codes signing identity in settings.xml?

exist-versioning-release.md
line-o: Sounds very reasonable to add in this document.

exist-versioning-release.md
line-o: Yes, I would need more information on this.

exist-versioning-release.md
line-o: It was "if you wish you can do a dry-run first", you know. But it is okay to change that to a very rigid "do a dry-run first".

exist-versioning-release.md
joewiz: Building on Adam's suggestions, I've added some more context and a link:

> In order to allow users of macOS 10.13 (Catalina) and higher to launch the eXist installer and application with minimal security warnings from the operating system, Apple encourages developers to code-sign their applications with an Apple Developer ID and submit the application (and disk image, or DMG, in the case of applications shipped in a DMG) to Apple's [notarization service](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution). The notarization service scans the application and the DMG for malicious software, and returns a certificate that the developer can staple to the application and DMG before a release is distributed to users. Without preparing eXist releases this way, users will encounter security warnings that require multiple steps to work around. Thus, in the eXist release process, both the application and the disk image (DMG) needs to be signed with an Apple Developer ID, notarized by Apple, and the notarization ticket need to be stapled to each. 

(It's macOS, not macOSX.)

exist-versioning-release.md
```diff
@@ -208,101 +313,169 @@ You will require a system with:
...
 
-7.  Visit the GitHub releases page [https://github.com/eXist-db/exist/releases](https://github.com/eXist-db/exist/releases) and create a new release, enter the tag you previously created and link the release notes from the blog.
+3.  Visit the GitHub releases page [https://github.com/eXist-db/exist/releases](https://github.com/eXist-db/exist/releases) and create a new release, enter the tag you previously created and link the release notes from the blog.
 
-8.  Send an email to the `exist-open` mailing list announcing the release with a title similar to `[ANN] Release of eXist 5.3.0`, copy and paste the release notes from the blog into the email and reformat appropriately (see past emails).
+4.  Send an email to the `exist-open` mailing list announcing the release with a title similar to `[ANN] Release of eXist 5.3.0`, copy and paste the release notes from the blog into the email and reformat appropriately (see [past emails](https://sourceforge.net/p/exist/mailman/search/?mail_list=exist-open)).
```
joewiz, line 456: Could you add an item: "Send a note to the eXist-db.org Slack #community channel" with a link to the release notes? (Phew, that's a lot of outreach duties!)

Also, it would be nice if we could find some convenient way to email the rendered HTML of the release notes, rather than the raw markdown.

duncdrum:
@line-o could you take a look at the new hotness currently on `develop` and see if anything from here needs rescuing or  close if if this is fully superseded.

PR #3976, opened 2021-07-14
Use Jakarta JAXB and extracts Test versions

reinhapa:
### Description:
- Use the Jakarta JAXB implementations instead of the old ones
- Extracts version definitions for the test environment


adamretter (approved):
Excellent. Thanks @reinhapa

PR #3988, opened 2021-07-29
Improve how XSuite starts and stops the database

adamretter:
This refactors XSuite (which is a JUnit Runner) to follow the JUnit `@BeforeClass` and `@AfterClass` approach to starting and stopping the database for each test suite. It also removes the need to start the database to introspect XQSuite tests, instead it now directly uses eXist-db's XQuery parser to find the tests.

This appears to fix some issues with the test suite that were identified by @reinhapa's JUnit 5 work.

### DIFF
--- a/exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
+++ b/exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
@@ -473,7 +473,9 @@ public class XQueryURLRewrite extends HttpServlet {
 
         try (final DBBroker broker = pool.get(Optional.ofNullable(user))) {
 
-            model.getSourceInfo().source.validate(broker.getCurrentSubject(), Permission.EXECUTE);
+            if (model.getSourceInfo().source instanceof DBSource) {
+                ((DBSource) model.getSourceInfo().source).validate(Permission.EXECUTE);
+            }
 
             if (model.getSourceInfo().source.isValid(broker) != Source.Validity.VALID) {
                 urlCache.remove(url);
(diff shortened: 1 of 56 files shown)


exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
```diff
@@ -473,7 +473,9 @@ private ModelAndView getFromCache(final String url, final Subject user) throws E
 
         try (final DBBroker broker = pool.get(Optional.ofNullable(user))) {
 
-            model.getSourceInfo().source.validate(broker.getCurrentSubject(), Permission.EXECUTE);
+            if (model.getSourceInfo().source instanceof DBSource) {
+                ((DBSource) model.getSourceInfo().source).validate(Permission.EXECUTE);
+            }
```
dizzzz, line 478: do we need a logging for the 'false' scenario?

exist-core/src/main/java/org/exist/storage/XQueryPool.java
```diff
@@ -182,7 +183,9 @@ public CompiledXQuery borrowCompiledXQuery(final DBBroker broker, final Source s
         }
 
         //check execution permission
-        source.validate(broker.getCurrentSubject(), Permission.EXECUTE);
+        if (source instanceof DBSource) {
+            ((DBSource) source).validate(Permission.EXECUTE);
+        }
```
dizzzz, line 188: do we need some logging here, in  case the match fails?

dizzzz (approved):
Nice to see to have the broker reference removed; I need to check the Messaging/replication code though, as I remember I had to do some strange things there.

exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
adamretter: I don't believe so. This was done through inheritance before, the methods for every source sublass other than DBSource were just empty

exist-core/src/main/java/org/exist/storage/XQueryPool.java
adamretter: Same as above

adamretter:
> Nice to see to have the broker reference removed; 

Thanks @dizzzz . I did preserve the old API too so there should be no breakage.

Issue #2394, opened 2019-01-16
Expose eXist-specific serialization parameters to standard methods


joewiz:
### What is the problem

The XSLT and XQuery Serialization 3.1 spec allows implementations [to define their own serialization methods and parameters](https://www.w3.org/TR/xslt-xquery-serialization-31/#serparam), but eXist doesn't make [its implementation-specific serialization parameters](https://exist-db.org/exist/apps/doc/xquery#legacy-serialization) available via the standard `fn:serialize()` function or `declare option output:*` prolog option declarations. These are only available via `util:serialize()` or via `declare option exist:serialization`. 

This has led to user confusion, e.g., https://stackoverflow.com/questions/54215193/exist-db-serialize-is-expand-xincludes-no-ignored/54220023#54220023. 

Addressing this would also improve consistency and further align eXist with the XQuery 3.1 family of specs.

### What did you expect

I expected to be able to supply eXist-specific serialization parameters such as `expand-xincludes` via the official method for implementation-defined serialization features. Quoting from [the relevant portion of the spec](https://www.w3.org/TR/xslt-xquery-serialization-31/#serparam):

> Implementations MAY define additional serialization parameters, and MAY allow users to do so. For this purpose, the name of a serialization parameter is considered to be a QName; the parameters listed above are QNames whose expanded-QName has a null namespace URI, while any additional serialization parameters that are either implementation-defined or defined by the host language MUST have names that are namespace-qualified. Any such additional serialization parameters MUST NOT be in the namespace `https://www.w3.org/2010/xslt-xquery-serialization`. 

Given the namespace requirements here, we'd use the `exist:` namespace prefix to differentiate between standard parameters like `method` and eXist's implementation-defined parameters, e.g., `exist:expand-includes`. Other eXist-specific methods/parameters would include: `method=microxml` or `jsonp=functionname`, `highlight-matches=yes|no`, `process-xsl-pi=yes|no`, `add-exist-id=yes|no`, `json-ignore-whitespace-text-nodes=yes|no`.

### Describe how to reproduce or add a test

Attempts to eXist-specific serialization parameters in `fn:serialize()`, e.g., `expand-includes=no` fails. See https://stackoverflow.com/questions/54215193/exist-db-serialize-is-expand-xincludes-no-ignored/54220023#54220023.

### Context information

- eXist-db version + Git Revision hash: eXist 4.6.0-SNAPSHOT 9e3461c4e
- Java version: OpenJDK 11.0.1+13
[...]

PR #3990, opened 2021-07-30
Add eXist specific serialization options to fn:serialize

marmoure:
Implementation of eXist-specific serialization parameters 
Closes https://github.com/eXist-db/exist/issues/2394
Documentation https://github.com/eXist-db/documentation/pull/672

You can now directly apply the serialization parameter using `fn:serialize`
```xquery
fn:serialize($node, map { xs:QName("exist:expand-xincludes") : false() })
```

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
+++ b/exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
@@ -22,6 +22,7 @@
 package org.exist.xquery.util;
 
 import org.exist.Namespaces;
+import org.exist.dom.QName;
 import org.exist.numbering.NodeId;
 import org.exist.stax.ExtendedXMLStreamReader;
 import org.exist.storage.serializers.EXistOutputKeys;
@@ -74,6 +75,7 @@ public class SerializerUtils {
         USE_CHARACTER_MAPS("use-character-maps", Type.MAP, Cardinality.ZERO_OR_ONE, Sequence.EMPTY_SEQUENCE),
         VERSION(OutputKeys.VERSION, Type.STRING, Cardinality.ZERO_OR_ONE, new StringValue("1.0"));
 
+
         final String parameterName;
         final int type;
         final Cardinality cardinality;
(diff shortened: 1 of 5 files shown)


duncdrum (changes requested):
Thank you for the PR, green Ci suites on initial run don't happen that often ;-) well done. 

Can you please add a comment about the new option to the documentation of exist in that repo and link the PR to this one. 

Please have a look at the 4 code smells, i think the logic can be streamlined to fewer nested `if` statements. 

Could you also lint the `serialize.xql` test file, the empty spaces between `{` and `<` are difficult to read, the `return` keyword is in the wrong line as well. 

Am I correct that this PR closes #2394 if so please adjust the keyword in your PR message.

dizzzz (commented):
LGTM ; wondering if the serialisation params now are identical to the ones that can be specified statically in a query.

exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
```diff
@@ -283,43 +397,43 @@ private static boolean checkTypes(final ParameterConvention parameterConvention,
...
+        } else if(Type.STRING == parameterConvention.getType()) {
             // ignore "absent" i.e. empty sequence
             if(!parameterValue.isEmpty()) {
-                properties.setProperty(parameterConvention.parameterName, ((StringValue) parameterValue.itemAt(0)).getStringValue());
+                properties.setProperty(parameterConvention.getLocalParameterName(), ((StringValue) parameterValue.itemAt(0)).getStringValue());
             }
-        } else if(Type.DECIMAL == parameterConvention.type) {
-            properties.setProperty(parameterConvention.parameterName, ((DecimalValue) parameterValue.itemAt(0)).getStringValue());
-        } else if(Type.QNAME == parameterConvention.type) {
+        } else if(Type.DECIMAL == parameterConvention.getType()) {
```
duncdrum, line 415: the remaining code-smell is between this line and ~430. I agree with the tool that this section is very hard to maintain and follow. There are two options either refactor the code to be less complex, or add some inline comments so that others can follow the logic more easily. What do you think?

duncdrum (commented):
Thanks for the updates, the all look good, two outstanding details, which shouldn't take too much of your time. either refactor too reduce the complexity reported by sonatype, or if that is too difficult, please add some inline comments to help fellow contributors understand the logic. 

the new option should also be mentioned in the docs, so that users can find it. This would be a separate PR [here](https://github.com/eXist-db/documentation)

joewiz:
As discussed during yesterday's Community Call, the answer to @dizzzz's question would require the addition of some new tests. I think something like:

```xquery
xquery version "3.1";

declare option exist:enable-xincludes "yes";

<article xmlns:xi="http://www.w3.org/2001/XInclude">
    <title>My Title</title>
    <xi:include href="{$ser:collection}/test.xml"/>
</article>
```

(Option declarations can only take string literals, so whereas fn:serialize can take `map { "indent": true() }`, the prolog analog is `declare option:indent "yes";`. See https://www.w3.org/TR/xquery-31/#id-option-declaration.)

Adam mentioned that adapting such a test to xqsuite would probably require util:eval:

```xquery
declare
    %test:assertXPath("contains($result, 'comment')")
function ser:exist-expand-xinclude-true-prolog() {

let $query := ``[xquery version "3.1";

declare option exist:enable-xincludes "yes";

<article xmlns:xi="http://www.w3.org/2001/XInclude">
    <title>My Title</title>
    <xi:include href="{$ser:collection}/test.xml"/>
</article>]``)

return 
    util:eval($query)
 };
```

I haven't tested this, but hopefully it helps toward the goal!

duncdrum (approved):

duncdrum:
The docs PR is currently blocked because of https://github.com/eXist-db/dashboard/issues/199. 
Outside of that both PRs (this and the linked docs PR) are good to merge from my side, both tested locally.

duncdrum:
there are two new commits since this was marked ready for review, @marmoure is it, or should we keep this as a draft until all commits are in?

adamretter:
@marmoure there seem to be some genuine test failures identified by CI, can you check `TestEXistXMLSerialize`.

adamretter:
Well done @marmoure on your first contribution to eXist-db, and thank you.

PR #3991, opened 2021-07-30
Refactor DBBroker into an Interface

adamretter:
Previously `DBBroker` was an Abstract Class, however for as long as I can remember there has only been a single sub-class which is `NativeBroker`. Whilst eXist-db was originally designed to have different brokers (e.g. MySQL as an option), that ability has long since disappeared. The mixing of code between DBBroker and NativeBroker was confusing.
DBBroker should have been an interface and now is.

PR https://github.com/eXist-db/exist/pull/3992 should be merged first ideally.

**NOTE** - These changes are API *backwards compatible* at the source level only! They are not backwards compatible at the *compilation level*. That is to say that libraries compiled against versions of eXist-db before this change may (if they use this class) raise the error :
```java
java.lang.IncompatibleClassChangeError: Found interface org.exist.storage.DBBroker, but class was expected
```

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/DBBroker.java
+++ b/exist-core/src/main/java/org/exist/storage/DBBroker.java
@@ -22,9 +22,6 @@
 package org.exist.storage;
 
 import com.evolvedbinary.j8fu.tuple.Tuple2;
-import it.unimi.dsi.fastutil.objects.ObjectLinkedOpenHashSet;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.exist.Database;
 import org.exist.EXistException;
 import org.exist.backup.RawDataBackup;
@@ -32,7 +29,6 @@ import org.exist.collections.Collection;
 import org.exist.collections.Collection.SubCollectionEntry;
 import org.exist.collections.triggers.TriggerException;
 import org.exist.dom.persistent.*;
-import org.exist.indexing.Index;
 import org.exist.indexing.IndexController;
 import org.exist.indexing.StreamListener;
 import org.exist.indexing.StructuralIndex;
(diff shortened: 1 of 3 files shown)


exist-core/src/main/java/org/exist/storage/DBBroker.java
```diff
@@ -74,130 +69,66 @@
...
 	// Matching types
-	public final static int MATCH_EXACT 		= 0;
-	public final static int MATCH_REGEXP 		= 1;
-//	public final static int MATCH_WILDCARDS 	= 2;  // no longer used!
-	public final static int MATCH_CONTAINS 		= 3;
-	public final static int MATCH_STARTSWITH 	= 4;
-	public final static int MATCH_ENDSWITH 		= 5;
+	int MATCH_EXACT 		= 0;
+	int MATCH_REGEXP 		= 1;
+    int MATCH_WILDCARDS 	= 2;  // no longer used!
```
dizzzz, line 77: formatting is not 100% ok in file

exist-core/src/main/java/org/exist/storage/NativeBroker.java
```diff
@@ -153,6 +158,34 @@
 
     private final static DigestType BINARY_RESOURCE_DIGEST_TYPE = DigestType.BLAKE_256;
 
+    protected boolean caseSensitive = true;
```
dizzzz, line 161: please could you some java doc here about the semantics ?

dizzzz (approved):
Looks like a clean change to me; different implementation makes no sense to me anymore.

exist-core/src/main/java/org/exist/storage/NativeBroker.java
adamretter: It seems this is not actually needed in `NativeBroker` at all, it is instead related to `NativeValueIndex` and so I have now removed it from `NativeBroker`.

exist-core/src/main/java/org/exist/storage/DBBroker.java
adamretter: Tabs vs. Spaces... I just fixed it

wolfgangmm (approved):
The refactoring is good as far as I can see.

dizzzz:
has the PR been polluted by the contents of the XAR/test PR?

adamretter:
> has the PR been polluted by the contents of the XAR/test PR?

Not polluted - just better to merge the other first as I have been stacking changes ;-)

dizzzz:
Did we forget about this pr ?

adamretter:
@dizzzz it's for 6.0.0

PR #3992, opened 2021-08-02
Instead of depending on an external XAR for tests just build it

adamretter:
Previously we depended on a XAR built from https://github.com/dizzzz/existdb-trigger-test which we included in this Git repo. This approach was fragile as API changes meant that the XAR had to be re-built and re-published.

This test just include the code for the XAR in a separate module and builds it as a dependency before the tests are run. Thanks to @dizzzz for creating the code and XAR originally.


extensions/modules/expathrepo/src/test/java/org/exist/repo/PackageTriggerTest.java
```diff
@@ -50,7 +50,7 @@
 
 public class PackageTriggerTest {
 
-    static final String xarFile = "triggertest-1.1.0.xar";
+    static final String xarFile = "exist-expathrepo-trigger-test-5.4.0-SNAPSHOT.xar";
```
dizzzz, line 53: version is hardcoded... can we derive this from the environment ?

dizzzz (approved):

dizzzz (changes requested):
image

dizzzz:
image

extensions/modules/expathrepo/src/test/java/org/exist/repo/PackageTriggerTest.java
adamretter: @dizzzz Ah yes, good point - I forgot that I left that in... let me fix...

adamretter: Okay I sent a follow-up commit to fix that.

dizzzz (approved):

dizzzz (changes requested):
still irregular build issues? 
image

extensions/modules/expathrepo/src/test/java/org/exist/repo/PackageTriggerTest.java
```diff
@@ -73,7 +74,7 @@ public static void setup() throws PermissionDeniedException, IOException, Trigge
 
         // Load XAR file
         byte[] content;
-        try (InputStream resourceAsStream = PackageTriggerTest.class.getResourceAsStream(xarFile)) {
+        try (InputStream resourceAsStream = PackageTriggerTest.class.getResourceAsStream("/" + xarFile)) {
```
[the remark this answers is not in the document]
dizzzz: updated pom file, rerunning triggered tests

dizzzz (approved):

PR #4011, opened 2021-08-23
[fix] NoOp validate() if subject is unavailable

### LINKED ISSUE #4009: [BUG] NPE in SimpleACLPermission leads to server errors (status code 500)
**Describe the bug**

It is not possible to use most of the built in applications (eXide, dashboard, ...) with the current develop branch (commit 302e6b5b5565922176b5897149203d944a5c6a76).

As already mentioned here: https://github.com/eXist-db/dashboard/issues/199

The reason is an NPE that is thrown in SimpleACLPermission.validate because the user parameter is null.
I was able to trace it to a call to DBBroker.getCurrentSubject that returns null in DBSource.validate(mode).

https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/source/DBSource.java#L174

**Expected behavior**

Built in applications to work with the current develop HEAD.

**To Reproduce**

1. Build exist from source (tested with 302e6b5b5565922176b5897149203d944a5c6a76)
2. start exist
3. open dashboard

 - eXist-db version: 5.4.0-SNAPSHOT 302e6b5b5565922176b5897149203d944a5c6a76
 - Java Version Adopt OpenJDK 8

**Additional context**

reinhapa:
### Description:
Fixes #4009 by skip access check in case of `validate(int)` method called and no subject is available on the broker instance. In case of calling the method `validate(Subject, int)` without valid subject will lead to a `PermissionDeniedException`

Signed-off-by: Patrick Reinhart <patrick@reini.net>

### DIFF
--- a/exist-core/src/main/java/org/exist/source/DBSource.java
+++ b/exist-core/src/main/java/org/exist/source/DBSource.java
@@ -171,7 +171,10 @@ public class DBSource extends AbstractSource {
     @Deprecated
     public void validate(final int mode) throws PermissionDeniedException {
         //TODO(AR) This check should not even be here! Its up to the database to refuse access not requesting source
-        validate(broker.getCurrentSubject(), mode);
+        final Subject subject = broker.getCurrentSubject();
+        if (subject != null) {
+            doValidation(subject, mode);
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


dizzzz (approved):

line-o (approved):
While this is a band-aid fix, I can confirm that with this PR applied the packages installed in exist-db are working again.
I think it could be useful to have a debug-level log entry whenever the validation is skipped.

PR #4012, opened 2021-08-23
[library] upgrade to most recent saxon 9.9 version

dizzzz:
Fixes Saxon issue Bug #4684: "NullPointerException compiling a stylesheet using eXist-DB transform() function;"

for Alexander henket


line-o (approved):

dizzzz:
Unfortunately this was not detected by dependabot; We expect that this is caused because of the non-semver numbering scheme.

PR #4031, opened 2021-09-08
Separate Memtree from in-memory DOM Document

### LINKED ISSUE #1463: Element constructor creates document node. It should not.

adamretter:
This now allows eXist-db to distinguish between in-memory document nodes and other in-memory nodes.

Closes https://github.com/eXist-db/exist/issues/1463

### DIFF
--- a/exist-core/src/main/java/org/exist/backup/xquery/ListBackups.java
+++ b/exist-core/src/main/java/org/exist/backup/xquery/ListBackups.java
@@ -120,7 +120,7 @@ public class ListBackups extends BasicFunction
                 }
             }
             builder.endElement();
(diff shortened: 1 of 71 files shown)


duncdrum (changes requested):
Ci is red across the board. The referenced issue has been fixed by another PR. @adamretter  is this still relevant, or good to close? 

needs a rebase

adamretter:
This is a large piece of work that I hope to continue. The fix for #1463 was a workaround more than anything else.

PR #4046, opened 2021-09-27
[build] use latest jdk 17, debug illegal access

reinhapa:
### Description:
Uses the latest LTS JDK 17 illegal access debug function


duncdrum:
LGTM any reason this is still marked as draft?

duncdrum:
btw should we not just switch the evergreen build to run on `17` now that it is out, instead of the non-lts `16`

reinhapa:
@duncdrum the argument `--illegal-access=debug`is only available up and until JDK 16 and removed from JDK 17 and onward

Issue #4048, opened 2021-09-28
[BUG] xmldiff:compare catches an exception 


PieterLamers:
**Describe the bug**
in a number of cases `xmldiff:compare#2` returns `exerr:ERROR An exception occurred while serializing node for comparison: Caught exception during comparison` 

**Expected behavior**
I expect `false()` to be returned, or `true()`, see tests below.

**To Reproduce**

```xquery
xquery version "3.1";

module namespace t="http://exist-db.org/xquery/test";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare %test:assertTrue function t:test-with-two-identical-nodes-succeeds() {
    let $a := (<foo/>)
    let $b := (<foo/>)
    return
        xmldiff:compare($a, $b)
};
declare %test:assertFalse function t:test-with-two-different-nodes-succeeds() {
    let $a := (<foo bar=""/>)
    let $b := (<foo/>)
    return
        xmldiff:compare($a, $b)
};
declare %test:assertTrue function t:test-with-two-identical-node-sequences-fails() {
    let $a := (<foo/>,<bar/>)
    let $b := (<foo/>,<bar/>)
    return
        xmldiff:compare($a, $b)
};
declare %test:assertFalse function t:test-with-two-different-node-sequences-fails() {
    let $a := (<foo/>,<bar/>)
    let $b := (<foo/>,<meh/>)
    return
        xmldiff:compare($a, $b)
};
declare %test:assertFalse function t:test-with-sequences-of-unequal-length-fails() {
    let $a := (<foo/>,<bar/>)
    let $b := (<foo/>)
    return
        xmldiff:compare($a, $b)
};
declare %test:assertFalse function t:test-with-sequence-of-two-and-empty-sequence-succeeds() {
    let $a := (<foo/>,<bar/>)
    let $b := ()
    return
        xmldiff:compare($a, $b)
};
```
I think the reason why the pattern is the way it is (with the last test succeeding), is that there is a false thrown after the first node comparison. I think the function cannot compare more than one node, although the signature says it does: `xmldiff:compare($node-set-1 as node()*, $node-set-2 as node()*)`

exist 5.3.0, windows 10, amazon corretto 1.8.282

adamretter:
Okay I can see the issue...

1. The function signature does allow more than one node at a time to `$node-set-1` and/or `$node-set2`
2. If there is more than one node to either of the args, then the function serializes each node in-turn into a string, and concatenates them together.

[...]

PR #4049, opened 2021-09-28
Remove xmldiff XQuery module

joewiz:
### Description:

The xmldiff XQuery module contains only one function, `xmldiff:compare()`, which simply compares two sets of nodes and returns `true()` or `false()`. It doesn't actually expose any "diffs." In fact, this function is entirely redundant with the standard `fn:deep-equal()` function and has no added value. Furthermore, `xmldiff:compare()` suffers from [bugs](https://github.com/eXist-db/exist/issues/4048), and is less capable than `fn:deep-equal()`, which now performs comparisons on maps and arrays; see https://www.w3.org/TR/xpath-functions-31/#func-deep-equal. 

This PR removes the one location in our XQuery test suite code that referenced (but did not actually use) the xmldiff module, and removes the xmldiff XQuery module from the eXist source. `mvn -V clean verify` passed all tests locally.

It does not remove all traces of xmldiff, since this may be needed for other purposes at the java level(?). I'm happy to remove all traces in follow-up commits to this PR if that's desired.

If this PR is incorporated into the next major version of eXist, it would be great to deprecate the function for now in the next 5.x release. This is beyond my abilities, so could someone help with this? 

### Reference:

- xmldiff function documentation: http://exist-db.org/exist/apps/fundocs/view.html?uri=http://exist-db.org/xquery/xmldiff&location=java:org.exist.xquery.modules.xmldiff.XmlDiffModule&details=true
- fn:deep-equal function documentation: https://www.w3.org/TR/xpath-functions-31/#func-deep-equal
- Previous discussions on exist-open from Apr 2018: https://sourceforge.net/p/exist/mailman/exist-open/?viewmonth=201804

### Type of tests:

There were no tests for the xmldiff XQuery module, so none needed to be removed (or added, for that matter) for the purposes of this PR.


duncdrum (approved):
pending CI, the changes look good. This would be a 6.0.0 milestone change though.

duncdrum:
Usage in the wild: 
- https://github.com/rhdunn/xquery-intellij-plugin/blob/250f7b7b31424fe02ebdceb33243a58ae3acac32/src/plugin-existdb/main/resources/org/exist-db/xquery/xmldiff.xqy#L12 

- https://github.com/opensiddur/opensiddur/blob/be258be4027a8517e032600cd1fedbedd9aadc5c/opensiddur-server/src/api/data/original.xqm#L101
- https://github.com/newtfire/newtfire-site

joewiz:
@EfraimFeinstein Would you have any objection to the `xmldiff:compare()` function being deprecated and removed from eXist? @duncdrum noticed that you use it in https://github.com/opensiddur/opensiddur/blob/be258be4027a8517e032600cd1fedbedd9aadc5c/opensiddur-server/src/api/data/original.xqm#L101. It can simply be replaced with `fn:deep-equal()`.

@rhdunn Same for you: https://github.com/rhdunn/xquery-intellij-plugin/blob/250f7b7b31424fe02ebdceb33243a58ae3acac32/src/plugin-existdb/main/resources/org/exist-db/xquery/xmldiff.xqy#L12.

@ebeshero Same for you: https://github.com/newtfire/newtfire-site/search?q=xmldiff

PieterLamers:
As noted in #4050 the implementation `fn:deep-equal#2` is not quite up to its task performance-wise, so that should be fixed before we get rid of `xmldiff:compare#2`.

adamretter:
@joewiz In light of https://github.com/eXist-db/exist/pull/4554 should this PR be closed now?

PR #4063, opened 2021-10-15
fix resource leak due to Files.find and Files.list

lujiefsi:
Files.find and Files.list will open stream, we should close them


lujiefsi:
> I tried to repair the PR online, but I failed; I think the `Journal.java` is not correct as there Stream is consumed elsewhere. The other one is correct.
> 
> Clearly the PR has been submitted without a build performed :-(

yes, the change  of `Journal.java`  is unnecessary.

lujiefsi:
close it with https://github.com/eXist-db/exist/pull/4065

PR #4081, opened 2021-11-10
Improve file:sync

### LINKED ISSUE #2627: As a developer I would like file:sync to also delete surplus files

line-o:
### Description:

Improve `file:sync`

The third parameter can now be either an xs:dateTime (kept for backwards compatibility) or a map.

The map has three keys at the moment:

- prune `xs:boolean` flag to control if surplus files on the filesystem should be removed
- exclude `xs:string*` sequence of glob patterns with support of `*`, `?`, `**` placeholders. Files that match will not be removed from disk ~~nor synced to the filesystem~~.
- after `xs:dateTime?` synchronise only files newer than the dateTime provided

Example:

```xquery

file:sync("/db/apps/my-app", "/Users/me/projects/my-app", map {
  "after": current-dateTime() - xs:dayTimeDuration("PT1D"), (: changes in the last day :)
  "excludes": (".git", ".env"), (: leave local settings and version control untouched  :)
  "prune": true() (: remove anything not in the database :)
})

```

Example of deprecated call:

```xquery
file:sync("/db/apps/my-app", "/Users/me/projects/my-app", 
  current-dateTime() - xs:dayTimeDuration("PT1D") (: changes in the last day :)
) 
```

Whenever `file:sync` selects a file or folder for deletion it will report this in the returned file:sync element.

Example report with deletions

```xml
<file:sync xmlns:file="http://exist-db.org/xquery/file" collection="/db/apps/my-app" dir="/Users/me/projects/my-app">
    <file:delete file="/Users/me/projects/my-app/.gitignore" name=".gitignore"/>
</file:sync>
```

The above also shows that `.*` is a reasonable excludes pattern.

Fixes the Function Signature to correctly state it returns a document-node() (was xs:boolean).

The function describes itself now as such
(NOTE: I found no way to force line-breaks before the option names)

```
Synchronize a collection with a directory hierarchy.This method is only available to the DBA role.

$collection as xs:string
    Absolute path to the collection to synchronize to disk.

$targetPath as item()
    The path or URI to the target directory. Relative paths resolve against EXIST_HOME.

$dateTimeOrOptionsMap as item()?
    Options as map(*). The available settings are:
    "prune": delete any file/dir that does not correspond to a doc/collection in the DB.
    "after": only resources modified after this date will be taken into account.
    "excludes": files on the file system matching any of these patterns will be left untouched.
    (deprecated) If the third parameter is of type xs:dateTime, it is the same as setting the "after" option.

Returns: document-node()
    A report (file:sync) which files and directories were updated (file:update) or deleted (file:delete). 
````

### Reference:

closes #3084
closes #2627

This is a follow up on #3084 rebased onto current HEAD (existdb-5.4.0-SNAPSHOT)
With this PR applied we can consider #2627 implemented.

### Type of tests:

2 XSuites with 33 Tests

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
+++ b/exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
@@ -222,7 +222,7 @@ public class SerializerUtils {
         try {
             final XMLStreamReader reader = parent.getContext().getXMLStreamReader(parameters);
             while (reader.hasNext() && (reader.next() != XMLStreamReader.START_ELEMENT)) {
-                //
+                /* advance to the first starting element (root node) of the options */
             }
 
             if (!Namespaces.XSLT_XQUERY_SERIALIZATION_NS.equals(reader.getNamespaceURI())) {
@@ -295,12 +295,14 @@ public class SerializerUtils {
             final Properties properties = new Properties();
 
             for (final W3CParameterConvention w3cParameterConvention : W3CParameterConvention.values()) {
-                final Sequence parameterValue = getParameterValue(parent, entries, w3cParameterConvention, new StringValue(w3cParameterConvention.getParameterName()));
+                final Sequence parameterValue = getParameterValue(parent, entries, w3cParameterConvention,
+                        new StringValue(w3cParameterConvention.getParameterName()));
                 setPropertyForMap(properties, w3cParameterConvention, parameterValue);
             }
 
             for (final ExistParameterConvention existParameterConvention : ExistParameterConvention.values()) {
-                final Sequence parameterValue = getParameterValue(parent, entries, existParameterConvention, new QNameValue(null, existParameterConvention.getParameterName()));
+                final Sequence parameterValue = getParameterValue(parent, entries, existParameterConvention,
+                        new QNameValue(null, existParameterConvention.getParameterName()));
                 setPropertyForMap(properties, existParameterConvention, parameterValue);
             }
 
@@ -411,37 +413,44 @@ public class SerializerUtils {
             return;
         }
 
+        final String localParameterName = parameterConvention.getLocalParameterName();
+        final String value;
+
         switch (parameterConvention.getType()) {
             case Type.BOOLEAN:
-                properties.setProperty(
-                        parameterConvention.getLocalParameterName(),
-                        ((BooleanValue) parameterValue.itemAt(0)).getValue() ? "yes" : "no"
-                );
+                value = ((BooleanValue) parameterValue.itemAt(0)).getValue() ? "yes" : "no";
+                properties.setProperty(localParameterName, value);
                 break;
             case Type.STRING:
-                properties.setProperty(parameterConvention.getLocalParameterName(), ((StringValue) parameterValue.itemAt(0)).getStringValue());
+                value = ((StringValue) parameterValue.itemAt(0)).getStringValue();
+                properties.setProperty(localParameterName, value);
                 break;
             case Type.DECIMAL:
-                properties.setProperty(parameterConvention.getLocalParameterName(), ((DecimalValue) parameterValue.itemAt(0)).getStringValue());
+                value = ((DecimalValue) parameterValue.itemAt(0)).getStringValue();
+                properties.setProperty(localParameterName, value);
                 break;
             case Type.QNAME:
                 if (Cardinality._MANY.isSuperCardinalityOrEqualOf(parameterConvention.getCardinality())) {
                     final SequenceIterator iterator = parameterValue.iterate();
                     while (iterator.hasNext()) {
-                        final String existingValue = (String) properties.get(parameterConvention.getLocalParameterName());
+                        final String existingValue = (String) properties.get(localParameterName);
+                        final String nextValue = ((QNameValue) iterator.nextItem()).getQName().toURIQualifiedName();
+
                         if (existingValue == null || existingValue.isEmpty()) {
-                            properties.setProperty(parameterConvention.getLocalParameterName(), ((QNameValue) iterator.nextItem()).getQName().toURIQualifiedName());
+                            properties.setProperty(localParameterName, nextValue);
                         } else {
-                            properties.setProperty(parameterConvention.getLocalParameterName(), existingValue + " " + ((QNameValue) iterator.nextItem()).getQName().toURIQualifiedName());
+                            properties.setProperty(localParameterName, existingValue + " " + nextValue);
                         }
                     }
                 } else {
-                    properties.setProperty(parameterConvention.getLocalParameterName(), ((QNameValue) parameterValue.itemAt(0)).getQName().toURIQualifiedName());
+                    value = ((QNameValue) parameterValue.itemAt(0)).getQName().toURIQualifiedName();
+                    properties.setProperty(localParameterName, value);
                 }
                 break;
             case Type.MAP:
                 //TODO(AR) implement `use-character-maps`
-                throw new UnsupportedOperationException("Not yet implemented support for the map serialization parameter: " + parameterConvention.getLocalParameterName());
+                throw new UnsupportedOperationException(
+                        "Not yet implemented support for the map serialization parameter: " + localParameterName);
         }
     }
 
--- a/exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
+++ b/exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
@@ -451,6 +451,9 @@ public class SerializerUtils {
                 //TODO(AR) implement `use-character-maps`
                 throw new UnsupportedOperationException(
                         "Not yet implemented support for the map serialization parameter: " + localParameterName);
+            default:
+                throw new UnsupportedOperationException(
+                        "Unsupported type " + Type.getTypeName(parameterConvention.getType()) + " for parameter value: " + localParameterName);
         }
     }
 
--- a/extensions/modules/file/pom.xml
+++ b/extensions/modules/file/pom.xml
@@ -63,6 +63,12 @@
         </dependency>
 
         <dependency>
+            <groupId>com.evolvedbinary.multilock</groupId>
+            <artifactId>multilock</artifactId>
+            <version>1.0.1</version>
+        </dependency>
+
+        <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
@@ -93,6 +99,7 @@
             <artifactId>jetty-deploy</artifactId>
             <scope>test</scope>
         </dependency>
+
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-jmx</artifactId>
--- a/extensions/modules/file/pom.xml
+++ b/extensions/modules/file/pom.xml
@@ -61,11 +61,9 @@
             <groupId>com.evolvedbinary.j8fu</groupId>
             <artifactId>j8fu</artifactId>
         </dependency>
-
         <dependency>
             <groupId>com.evolvedbinary.multilock</groupId>
             <artifactId>multilock</artifactId>
-            <version>1.0.1</version>
         </dependency>
 
         <dependency>
--- a/extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
+++ b/extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
@@ -23,11 +23,16 @@ package org.exist.xquery.modules.file;
 
 import java.io.*;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
 import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.BasicFileAttributes;
 import java.nio.file.attribute.FileTime;
+import java.net.URISyntaxException;
 import java.util.*;
+import java.util.stream.Stream;
 
 import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilder;
@@ -39,6 +44,7 @@ import javax.xml.transform.sax.SAXTransformerFactory;
 import javax.xml.transform.sax.TransformerHandler;
 import javax.xml.transform.stream.StreamSource;
 
+import org.apache.tools.ant.DirectoryScanner;
 import org.exist.collections.Collection;
 import org.exist.dom.persistent.BinaryDocument;
 import org.exist.dom.persistent.DocumentImpl;
@@ -46,6 +52,7 @@ import org.exist.dom.QName;
 import org.exist.dom.memtree.MemTreeBuilder;
 import org.exist.security.PermissionDeniedException;
 import org.exist.storage.lock.Lock.LockMode;
+import org.exist.storage.lock.ManagedLock;
 import org.exist.storage.serializers.EXistOutputKeys;
 import org.exist.storage.serializers.Serializer;
 import org.exist.util.FileUtils;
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

**Comment by line-o** [comment]
There are some tests not green yet. But it is now in a state where it makes sense to discuss further development.
Main question is: should the excludes pattern be matched against resources in the database (exclude matching resources from syncing)?

**Comment by line-o** [comment]
Windows is not behaving properly. Even after syncing to a separate folder for each test case it fails to create the test subdirectory in `helper:setup-fs-extra`.

**Comment by line-o** [comment]
Completely unrelated test failure I have never seen before: https://github.com/eXist-db/exist/runs/4212276488?check_suite_focus=true#step:5:8861

**Comment by line-o** [comment]
Patterns ending in `/*` seem to be a problem in Windows. At least the three automated tests with `test/*` as an exclude pattern let to unpredictable behaviour in the Windows test-environment (see deaf149).

**Comment by line-o** [comment]
As @duncdrum, @dizzzz  and others just pointed out the weird behaviour on Windows is very likely due to path separators being different.
So on Windows machines one would need to use the appropriate patterns with backslashes instead.

**Comment by line-o** [comment]
Uh! Codacy likes it, I am floored!

File: exist-core/src/main/java/org/exist/xquery/value/Sequence.java [hunk 1]
```diff
@@ -35,6 +35,8 @@
 import javax.annotation.Nullable;
 import java.math.BigInteger;
 import java.util.Iterator;
+import java.util.List;
```
line-o, line 38: This is a leftover from the old code and will be removed

line-o:
History can be shortened to 3 commits -> https://github.com/line-o/exist/commits/clean-history
If no one objects I will force-push that in here.

joewiz:
@line-o No objection here - sounds like a great idea.

exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
```diff
@@ -411,37 +413,44 @@ private static void setPropertyForMap(final Properties properties, final Paramet
...
+                            properties.setProperty(localParameterName, existingValue + " " + nextValue);
                         }
                     }
                 } else {
-                    properties.setProperty(parameterConvention.getLocalParameterName(), ((QNameValue) parameterValue.itemAt(0)).getQName().toURIQualifiedName());
+                    value = ((QNameValue) parameterValue.itemAt(0)).getQName().toURIQualifiedName();
+                    properties.setProperty(localParameterName, value);
                 }
                 break;
             case Type.MAP:
```
dizzzz, line 450: missing default: ?

extensions/modules/file/pom.xml
```diff
@@ -62,6 +62,12 @@
             <artifactId>j8fu</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.evolvedbinary.multilock</groupId>
+            <artifactId>multilock</artifactId>
+            <version>1.0.1</version>
```
dizzzz, line 68: not in parent pom?

line-o: I can put it the dependency in a different POM if that is better.
As I am not so familiar yet with Maven please be a little more decisive. Could you also point me to the pom file you think is best?

exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
line-o: There was none before, what would be an appropriate default here? I assume a `UsupportedOperationException` as in case of `Type.Map`.

extensions/modules/file/pom.xml
adamretter: I think @dizzzz means that as this is used in another module in eXist-db, with your change there would be two modules with the same dependency. Therefore the common dependency should be declared in the appropriate part of the parent pom and only reference in the module poms.

line-o: So, the question remains: which pom.xml is the desired one?

adamretter: >  parent pom

exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
line-o: @dizzzz What do you think?

line-o:
Windows tests still timing out. Note, his PR was rebased on current develop with #4098 already merged.

exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
dizzzz: Sorry I missed it;
So yes we should be cautious in not explicitly handling situation; parallel t the Map sounds good to me.

line-o: Added the default to guard against unsupported types in ParameterConventions (these are defined in the class itself so it is rather theoretical, but regardless it will help future devs in not making this mistake.)

line-o:
rebased onto current develop HEAD

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
-            new FunctionReturnSequenceType(Type.BOOLEAN, Cardinality.EXACTLY_ONE, "true if successful, false otherwise")
-        );
-	
-	private final static Properties DEFAULT_PROPERTIES = new Properties();
-	static {
-		DEFAULT_PROPERTIES.put(OutputKeys.INDENT, "yes");
-		DEFAULT_PROPERTIES.put(OutputKeys.OMIT_XML_DECLARATION, "no");
+    public static final String PRUNE_OPT = "prune";
+    public static final String AFTER_OPT = "after";
+    public static final String EXCLUDES_OPT = "excludes";
```
dizzzz, line 80: excludes... the intro says 'exclude'

dizzzz (approved):
LGTM

dizzzz (approved):
LGTM ... do you have doc PR?

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+
+    public static final QName FileSyncElement = new QName("sync", FileModule.NAMESPACE_URI);
+    public static final QName FileUpdateElement = new QName("update", FileModule.NAMESPACE_URI);
+    public static final QName FileDeleteElement = new QName("delete", FileModule.NAMESPACE_URI);
+    public static final QName FileErrorElement = new QName("error", FileModule.NAMESPACE_URI);
+
+    // TODO(JL) Figure out which namespace all attributes should be in (possible breaking change)
```
adamretter, line 87: Is this TODO still valid or have you figured it out?
If this is something to improve in future, perhaps it would be good to open an issue and add the link to the TODO.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+    public static final QName FileCollectionAttribute = new QName("collection", FileModule.NAMESPACE_URI);
+    public static final QName FileDirAttribute = new QName("dir", FileModule.NAMESPACE_URI);
+
+    public static final QName FileAttribute = new QName("file", XMLConstants.NULL_NS_URI);
+    public static final QName NameAttribute = new QName("name", XMLConstants.NULL_NS_URI);
+    public static final QName CollectionAttribute = new QName("collection", XMLConstants.NULL_NS_URI);
+    public static final QName TypeAttribute = new QName("type", XMLConstants.NULL_NS_URI);
+    public static final QName ModifiedAttribute = new QName("modified", XMLConstants.NULL_NS_URI);
+
+    public static final FunctionSignature signature =
```
adamretter, line 97: You could use the `FunctionDSL` here to simplify this

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+    @Override
+    public Sequence eval(final Sequence[] args, final Sequence contextSequence) throws XPathException {
+
+        if (!context.getSubject().hasDbaRole()) {
+            throw new XPathException(this, "Function file:sync is only available to the DBA role");
+        }
+
+        final String collectionPath = args[0].getStringValue();
+        final String target = args[1].getStringValue();
+        final HashMap<String, Sequence> options = getOptions(args[2]);
```
adamretter, line 138: As recognised good practice, the LHS (Left Hand Side) of a variable assignment in Java should always use the lowest-common Interface rather than the Implementation of that Interface. That is to say that this should likely be:
```java
final Map<String, Sequence> options = getOptions(args[2]);
```

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+
+        return startSync(target, collectionPath, options);
+    }
+
+    private HashMap<String, Sequence> getOptions(final Sequence parameter) throws XPathException {
+        HashMap<String, Sequence> options = new HashMap<>();
```
adamretter, line 144: Use the lowest-common Interface both of the return type of the function declaration and the HashMap creation.
Also all variables where possible should be `final`.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+                throw new XPathException(this, ErrorCodes.XPTY0004,
+                        "Invalid value for option \"excludes\", expected xs:string* got " +
+                                Type.getTypeName(seq.getItemType()));
+            }
+            options.put(EXCLUDES_OPT, seq);
+
+            checkOption(optionsMap, PRUNE_OPT, Type.BOOLEAN, options);
+            checkOption(optionsMap, AFTER_OPT, Type.DATE_TIME, options);
+        } else if (parameter.itemAt(0).getType() == Type.DATE_TIME) {
+            // QUESTION(JL) LOG deprecation warning?
```
adamretter, line 169: As it is described in this the function signature doc, I don't think we need to log anything here, you can remove the comment.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+            options.put(AFTER_OPT, parameter);
+        } else {
+            throw new XPathException(this, ErrorCodes.XPTY0004,
+                    "Invalid 3rd parameter, allowed are xs:dateTime or map(*) got " + Type.getTypeName(item.getType()));
```
adamretter, line 173: "allowed are" -> "allowed parameter types are"

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+        }
+        return options;
+    }
+
+    private void checkOption(
+            final AbstractMapType optionsMap,
+            final String name,
+            final int expectedType,
+            final HashMap<String, Sequence> options
```
adamretter, line 182: For the parameter, use the lowest-common Interface, i.e. `Map` and not `HashMap`.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+    ) throws XPathException {
+        final Sequence p = optionsMap.get(new StringValue(name));
+
+        if (p.isEmpty()) {
+            return; // nothing to do, continue
+        }
+
+        if (p.hasMany() || p.getItemType() != expectedType) {
+            throw new XPathException(this, ErrorCodes.XPTY0004,
+                    "Invalid value for option \"" + name + "\", expected " +
```
adamretter, line 192: "Invalid value for" -> "Invalid value type for"

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+                            Type.getTypeName(p.itemAt(0).getType()));
+        }
+
+        options.put(name, p);
+    }
+
+    private Sequence startSync(
+            final String target,
+            final String collectionPath,
+            final HashMap<String, Sequence> options
```
adamretter, line 203: For the parameter, use the lowest-common Interface, i.e. Map and not HashMap.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+    ) throws XPathException {
+        final Date startDate = options.get(AFTER_OPT).hasOne() ? ((DateTimeValue) options.get(AFTER_OPT)).getDate() : null;
+
+        boolean prune = ((BooleanValue) options.get(PRUNE_OPT)).getValue();
```
adamretter, line 207: `prune` should be final.
This will throw an NPE if the user doesn't specify `prune` in the map - this needs to be fixed. Should default to `false` presumably?

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+
+        final List<String> excludes = new ArrayList<>(Collections.emptyList());
+        for (final SequenceIterator si = options.get(EXCLUDES_OPT).iterate(); si.hasNext(); ) {
```
adamretter, line 210: This will throw an NPE if the user doesn't specify excludes in the map.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
-	
-	private final static Properties DEFAULT_PROPERTIES = new Properties();
-	static {
-		DEFAULT_PROPERTIES.put(OutputKeys.INDENT, "yes");
-		DEFAULT_PROPERTIES.put(OutputKeys.OMIT_XML_DECLARATION, "no");
+    public static final String PRUNE_OPT = "prune";
+    public static final String AFTER_OPT = "after";
+    public static final String EXCLUDES_OPT = "excludes";
+
+    public static final QName FileSyncElement = new QName("sync", FileModule.NAMESPACE_URI);
```
adamretter, line 82: It is good practice in Java that `static final` object members should be in all caps and should be word separated by `_` characters. See above for an example.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+        try {
+            if (p.isAbsolute()) {
+                targetDir = p;
+            } else {
+                final Optional<Path> home = context.getBroker().getConfiguration().getExistHome();
+                targetDir = FileUtils.resolve(home, target);
+            }
+
+            output.startDocument();
+            output.startElement(Sync.FileSyncElement, null);
```
adamretter, line 227: The static-final object members don't need to be be prefixed `Sync.` as they are already in scope.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+                    " for collection " + collectionPath);
+            return;
+        }
+
+        if (!Files.isWritable(targetDirectory)) {
+            reportError(output, "Failed to write to output directory: " + targetDirectory.toAbsolutePath());
+            return;
+        }
+
+        List<XmldbURI> subCollections = handleCollection(collectionPath, rootTargetAbsPath, targetDirectory, startDate, prune, excludes, output);
```
adamretter, line 268: Should be `final`

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+    private List<XmldbURI> handleCollection(
+            final XmldbURI collectionPath,
+            final String rootTargetAbsPath,
+            final Path targetDirectory,
+            final Date startDate,
+            final boolean prune,
+            final List<String> excludes,
+            final MemTreeBuilder output
+    ) throws PermissionDeniedException, LockException {
+        List<XmldbURI> subCollections;
```
adamretter, line 285: Declaration should be moved to where it is needed and marked final

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+                pruneCollectionEntries(collection, rootTargetAbsPath, targetDirectory, excludes, output);
+            }
+
+            for (final Iterator<DocumentImpl> i = collection.iterator(context.getBroker()); i.hasNext(); ) {
+                final DocumentImpl doc = i.next();
+                final Path targetFile = targetDirectory.resolve(doc.getFileURI().toASCIIString());
+                saveFile(targetFile, doc, startDate, output);
+            }
+
+            subCollections = new ArrayList<>(collection.getChildCollectionCount(context.getBroker()));
```
adamretter, line 302: Declaration should be here not above

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+                        return;
+                    }
+
+                    // handle non-empty directories
+                    if (Files.isDirectory(path)) {
+                        deleteWithExcludes(rootTargetAbsPath, path, excludes, output);
+                    } else {
+                        Files.deleteIfExists(path);
+                        // reporting
+                        output.startElement(Sync.FileDeleteElement, null);
```
adamretter, line 344: The static-final object members don't need to be be prefixed `Sync.` as they are already in scope.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+                        output.addAttribute(Sync.FileAttribute, path.toAbsolutePath().toString());
+                        output.addAttribute(Sync.NameAttribute, fileName);
+                        output.endElement();
+                    }
+
+                } catch (IOException | URISyntaxException
```
adamretter, line 350: `catch` should be final

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+                        | PermissionDeniedException | LockException e) {
+                    reportError(output, e.getMessage());
+                }
+            });
+        } catch (IOException e) {
```
adamretter, line 355: `catch` should be final

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+    }
+
+    private void saveFile(final Path targetFile, final DocumentImpl doc, final Date startDate, final MemTreeBuilder output) throws LockException {
+        try (final ManagedLock<MultiLock[]> lock = context.getBroker().getBrokerPool().getLockManager().acquireDocumentReadLock(doc.getURI())) {
+            if (startDate == null || (
+                    doc.getLastModified() > startDate.getTime() && (
+                            !Files.exists(targetFile) ||
+                                    Files.getLastModifiedTime(targetFile).compareTo(FileTime.fromMillis(doc.getLastModified())) >= 0
+                    ))) {
+                output.startElement(Sync.FileUpdateElement, null);
```
adamretter, line 367: The static-final object members don't need to be be prefixed `Sync.` as they are already in scope.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -55,183 +62,356 @@
...
+            reportError(output, e.getMessage());
+        } catch (final IOException e) {
+            reportError(output, "IO error while saving file: " + targetFile.toAbsolutePath().toString());
+        }
+    }
+
+    private void saveXML(final Path targetFile, final DocumentImpl doc, final MemTreeBuilder output) throws IOException {
+        final SAXSerializer sax = (SAXSerializer) SerializerPool.getInstance().borrowObject(SAXSerializer.class);
+        try {
+            boolean isRepoXML = Files.exists(targetFile) && FileUtils.fileName(targetFile).equals("repo.xml");
```
adamretter, line 393: Can be final

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -274,28 +454,124 @@ private void processRepoDesc(final Path targetFile, final DocumentImpl doc, fina
...
+    /**
+     * We need to convert to a relative path in relation to rootTargetAbsPath,
+     * as all the exclusion patterns are relative to rootTargetAbsPath.
+     *
+     * @param rootTargetAbsPath the root target (abs)path
+     * @param path              (abs)path to check for being excluded. Should be subdir of rootTargetAbsPath
+     * @param excludes          exclude patterns (in the convention of DirectoryScanner.match)
+     * @return true if the (rel)path in question is matched by some of the exclusion patterns
+     */
+    private static boolean isExcludedPath(String rootTargetAbsPath, Path path, List<String> excludes) {
```
adamretter, line 480: Parameters should be final

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -274,28 +454,124 @@ private void processRepoDesc(final Path targetFile, final DocumentImpl doc, fina
...
+        }
+        // root folder cannot be excluded
+        // path will then also be one character shorter than rootTargetApsPath
+        // and throw when attempting to construct the relative path
+        if (rootTargetAbsPath.startsWith(path.toString())) {
+            return false;
+        }
+
+        final String absPath = path.toAbsolutePath().toString();
+        final String relPath = absPath.substring(rootTargetAbsPath.length());
```
adamretter, line 492: There are functions in Java's `Path` class for getting relative paths, maybe better to use those.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -274,28 +454,124 @@ private void processRepoDesc(final Path targetFile, final DocumentImpl doc, fina
...
+    public static boolean matchAny(final Iterable<String> patterns, final String path) {
+        for (final String pattern : patterns) {
+            if (DirectoryScanner.match(pattern, path)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static void deleteWithExcludes(String root, final Path path, List<String> excludes, final MemTreeBuilder output) throws IOException {
```
adamretter, line 512: parameters should be final

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -274,28 +454,124 @@ private void processRepoDesc(final Path targetFile, final DocumentImpl doc, fina
...
+
+        @Override
+        public FileVisitResult visitFile(final Path file, final BasicFileAttributes attrs) throws IOException {
+            if (isExcludedPath(root, file, excludes)) {
+                hasExcludedChildren = true;
+                return FileVisitResult.CONTINUE;
+            }
+            Files.deleteIfExists(file);
+
+            output.startElement(Sync.FileDeleteElement, null);
```
adamretter, line 550: I don't think these need to be prefixed `Sync.` as that is already in-scope

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -274,28 +454,124 @@ private void processRepoDesc(final Path targetFile, final DocumentImpl doc, fina
...
+            if (exc != null) {
+                throw exc;
+            }
+            // deletion would fail due to non-empty directory
+            if (hasExcludedChildren) {
+                return FileVisitResult.CONTINUE;
+            }
+            Files.deleteIfExists(dir);
+
+            output.startElement(Sync.FileDeleteElement, null);
```
adamretter, line 569: I don't think these need to be prefixed `Sync.` as that is already in-scope

extensions/modules/file/src/test/resources/util/fixtures.xqm
```diff
@@ -0,0 +1,59 @@
...
+ : Lesser General Public License for more details.
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+module namespace fixtures="http://exist-db.org/xquery/test/util/fixtures";
+import module namespace util="http://exist-db.org/xquery/util";
```
adamretter, line 25: This import is unused and can be removed

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+/db
+    /file-module-test
+        /data
+            test-data.xml
+        test-text.txt
+        test-query.xq
+        bin
+:)
+declare function helper:setup-db() as empty-sequence() {
+    (
```
adamretter, line 42: The way that this is structured will make Java debugging of such XQuery much more difficult than it needs to be. Ideally we don't want to involve anything clever (such as the Arrow Operator) or superfluous in a test (unless the test is actually for the Arrow operator). We also should avoid any unnecessary operators and function calls. As such, I would instead suggest this should be refactored to the much simpler XQuery 1.0 code:

```xquery
let $_ := (
        xmldb:create-collection("/db", $fixtures:collection-name),
        helper:create-db-resource($fixtures:collection, "test-text.txt", $fixtures:TXT),
        helper:create-db-resource($fixtures:collection, "test-query.xq", $fixtures:XQY),
        helper:create-db-resource($fixtures:collection, "bin", $fixtures:BIN),

        xmldb:create-collection($fixtures:collection, $fixtures:child-collection-name),
        helper:create-db-resource($fixtures:child-collection, "test-data.xml", $fixtures:XML)
    )
   return ()
```
This will make Java debugging of the running XQuery much simpler. It will also as an unintended side-effect speed up the execution of this code.

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+declare function helper:clear-db() {
+    xmldb:remove($fixtures:collection)
+};
+
+declare function helper:create-db-resource($collection as xs:string, $resource as xs:string, $content as item()) as empty-sequence() {
+    (
+        xmldb:store($collection, $resource, $content),
+        xmldb:touch($collection, $resource, $fixtures:mod-date)
+    )
+    => helper:to-empty-sequence()
```
adamretter, line 63: This should be refactored to `return ()` so that the test is easy to debug

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+};
+
+declare function helper:modify-db-resource($collection as xs:string, $resource as xs:string) as empty-sequence() {
+    xmldb:touch($collection, $resource, $fixtures:mod-date-2)
+    => helper:to-empty-sequence()
```
adamretter, line 68: This should be refactored to `return ()` so that the test is easy to debug

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+};
+
+declare function helper:clear-suite-fs ($suite as xs:string) as empty-sequence() {
+    helper:glue-path((
+        util:system-property("java.io.tmpdir"),
+        $suite
+    ))
+    => helper:clear-fs()
```
adamretter, line 76: This should be refactored to `return helper:clear-fs()` so that the test is easy to debug

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+};
+
+declare function helper:clear-fs ($directory as xs:string) as empty-sequence() {
+    file:delete($directory)
+    => helper:to-empty-sequence()
```
adamretter, line 81: This should be refactored to `return ()` so that the test is easy to debug

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+declare function helper:get-test-directory ($suite as xs:string) as xs:string {
+    helper:glue-path((
+        util:system-property("java.io.tmpdir"),
+        $suite,
+        util:uuid()
+    ))
+};
+
+declare function helper:glue-path ($parts as xs:string+) as xs:string {
+    string-join($parts, "/")
```
adamretter, line 93: Wrapping one function in another just for almost no gain, will make this test much harder to debug. Instead you should remove this function and inline the `string-join($parts, '/')`

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+
+(:
+ : clear FS state and simulate additional data on the file system in a specific directory
+ : @returns given directory to allow use in pipeline (chain of arrow operators)
+ :)
+declare function helper:setup-fs-extra ($directory as xs:string) as xs:string {
+    let $action1 := file:mkdirs($directory)
+    let $action2 := file:mkdirs($directory || "/test")
+    let $action3 := (
+        (: cannot use fixtures here because this will lead to consumed input streams! :)
```
adamretter, line 104: What does this comment mean? Is there a problem somewhere else?

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+    )
+    and
+    helper:assert-permutation-of(
+        $expected?fs,
+        helper:get-dir-from-sync-result($result)
+        => helper:list-files-and-directories()
+    )
+};
+
+declare function helper:assert-permutation-of($expected as xs:anyAtomicType*, $actual as xs:anyAtomicType*) {
```
adamretter, line 157: Function is missing its return type declaration

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+    return
+        if (not($test?1 or exists($test?2)))
+        then error($helper:error,
+        "Assertion failed: expected permutation of " ||
+               "(" || string-join($expected, ", ") || ")" ||
+               " but got (" || string-join($actual, ", ") || ")")
+        else true()
+};
+
+declare function helper:permutation-reducer ($result, $next) {
```
adamretter, line 173: Function is missing its return type declaration

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+    let $first-index := index-of($result?2, $next)[1]
+    return [
+        $result?1 and $first-index > 0,
+        helper:maybe-remove-item-at-index($result?2, $first-index)
+    ]
+};
+
+declare function helper:maybe-remove-item-at-index($sequence as xs:anyAtomicType*, $index as xs:integer?) {
```
adamretter, line 181: Function is missing it's return type declaration

extensions/modules/file/src/test/resources/util/helper.xqm
```diff
@@ -0,0 +1,192 @@
...
+    then subsequence($sequence, 2)
+    else if ($index > 1)
+    then (
+        subsequence($sequence, 1, $index - 1),
+        subsequence($sequence, $index + 1)
+    )
+    else $sequence (: do nothing - will be handled later :)
+};
+
+declare function helper:to-empty-sequence($_ as item()*) as empty-sequence() {};
```
adamretter, line 192: Remove this function - test code should be simple and dumb to aid in debugging. Introducing unnecessary complexity into test code is never a good idea.

extensions/modules/file/src/test/xquery/modules/file/read-binary.xqm
```diff
@@ -19,19 +19,35 @@
  : License along with this library; if not, write to the Free Software
  : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  :)
-xquery version "3.0";
+xquery version "3.1";
 
-module namespace read-binary="http://exist-db.org/testsuite/modules/file/read-binary";
+module namespace readbinary="http://exist-db.org/testsuite/modules/file/read-binary";
+import module namespace helper="http://exist-db.org/xquery/test/util/helper" at "resource:util/helper.xqm";
```
adamretter, line 25: Unless I am mistaken I don't think the `resource:` prefixes should be used here

extensions/modules/file/src/test/xquery/modules/file/read-binary.xqm
```diff
@@ -19,19 +19,35 @@
...
+import module namespace fixtures="http://exist-db.org/xquery/test/util/fixtures" at "resource:util/fixtures.xqm";
 
 import module namespace test="http://exist-db.org/xquery/xqsuite" at "resource:org/exist/xquery/lib/xqsuite/xqsuite.xql";
```
adamretter, line 28: I think you can remove the `at "resource:org/exist/xquery/lib/xqsuite/xqsuite.xql";` bit as that is an old legacy way of doing this import for the test suite.

extensions/modules/file/src/test/xquery/modules/file/read-binary.xqm
```diff
@@ -19,19 +19,35 @@
...
+    %test:tearDown
+function readbinary:tear-down() as empty-sequence() {
+    helper:clear-suite-fs($readbinary:suite)
+};
+
+declare
+    %test:assertEquals("SERVER_SECRET=123!")
+function readbinary:without-serialization() {
+    helper:get-test-directory($readbinary:suite)
+    => helper:setup-fs-extra()
```
adamretter, line 49: I would suggest removing the arrow operator here, it will make Java debugging of this code almost impossible to follow - test code should be as simple as possible rather than clever.

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
```diff
@@ -0,0 +1,285 @@
...
+    %test:tearDown
+function sync:tear-down() {
+    helper:clear-db(),
+    helper:clear-suite-fs($sync:suite)
+};
+
+declare
+    %test:assertTrue
+function sync:simple() {
+    helper:get-test-directory($sync:suite)
```
adamretter, line 48: I would suggest removing the arrow operator here, it will make Java debugging of this code almost impossible to follow - test code should be as simple as possible rather than clever.

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
```diff
@@ -0,0 +1,285 @@
...
+        "deleted": (),
+        "fs": $fixtures:ROOT-FS
+    })
+};
+
+declare
+    %test:assertTrue
+function sync:empty-options-map() {
+    helper:get-test-directory($sync:suite)
+    => helper:sync-with-options(map{})
```
adamretter, line 61: I would suggest removing the arrow operator here, it will make Java debugging of this code almost impossible to follow - test code should be as simple as possible rather than clever.

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
```diff
@@ -0,0 +1,285 @@
...
+        "deleted": (),
+        "fs": $fixtures:ROOT-FS
+    })
+};
+
+declare
+    %test:assertError
+function sync:deprecated-options() {
+    helper:get-test-directory($sync:suite)
+    => helper:sync-with-options($fixtures:mod-date)
```
adamretter, line 73: I would suggest removing the arrow operator here, it will make Java debugging of this code almost impossible to follow - test code should be as simple as possible rather than clever.

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
```diff
@@ -0,0 +1,285 @@
...
+        "deleted": (),
+        "fs": $fixtures:ROOT-FS
+    })
+};
+
+declare
+    %test:assertError("err:XPTY0004")
+function sync:bad-options-1() {
+    helper:get-test-directory($sync:suite)
+    => helper:sync-with-options(xs:date("2012-12-21"))
```
adamretter, line 85: I would suggest removing the arrow operator here, it will make Java debugging of this code almost impossible to follow - test code should be as simple as possible rather than clever.

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
```diff
@@ -0,0 +1,285 @@
...
+};
+
+declare
+    %test:assertError("err:XPTY0004")
+function sync:bad-options-2() {
+    helper:get-test-directory($sync:suite)
+    => helper:sync-with-options("2012-12-21T10:12:21")
```
adamretter, line 92: I would suggest removing the arrow operator here, it will make Java debugging of this code almost impossible to follow - test code should be as simple as possible rather than clever.

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
```diff
@@ -0,0 +1,285 @@
...
+};
+
+declare
+    %test:assertError("err:XPTY0004")
+function sync:bad-options-3() {
+    helper:get-test-directory($sync:suite)
+    => helper:sync-with-options("lizard")
```
adamretter, line 99: I would suggest removing the arrow operator here, it will make Java debugging of this code almost impossible to follow - test code should be as simple as possible rather than clever.

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
```diff
@@ -0,0 +1,285 @@
...
+};
+
+declare
+    %test:assertError("err:XPTY0004")
+function sync:bad-options-4() {
+    helper:get-test-directory($sync:suite)
+    => helper:sync-with-options("")
```
adamretter, line 106: I would suggest removing the arrow operator here, it will make Java debugging of this code almost impossible to follow - test code should be as simple as possible rather than clever.

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
```diff
@@ -0,0 +1,285 @@
...
+};
+
+(:
+ : TODO(JL) should also report %test:assertError("err:XPTY0004")
```
adamretter, line 110: Is there an issue for this - if so can you link to it in the XQDoc please?

adamretter (changes requested):
In general this looks very good.

There are a couple of places in the Java code where NPE's could be caused by optional user input that need to be addressed. Otherwise there are a few minor non-Java-esque stylistic things that should be improved.

However, the XQuery test code IMHO tries to be far too clever. It also makes heavy use of the arrow operator and there are some unnecessary function wrappers. Unfortunately in eXist-db it is very hard to debug the underlying Java code implementation of XQuery when some of the new XQuery 3.1 constructs such as the Arrow Operator are used, therefore I would recommend removing the arrow operator from these tests; I have't found a place in the code where the arrow operator is actually needed.

In general, not just for XQuery, but for any language it is much better to create very simple and dumb tests. You should avoid abstracting too much logic and/or common code into functions or clever constructs, otherwise when you have a problem in future, you will spend just as much time debugging the test code, as actually using the test code to debug the underlying problem. By all means use all the available constructs and lots of abstraction in production code if you wish, it may help you keep the code clean and the total LoC code small... but test code should not resemble production code!

line-o:
Thanks for taking the time to review the code a second time. To my surprise a lot of new changes are now requested on changes that were already introduced.
Apart from the formal changes, I feel the conceptual changes that are requested need to be discussed.

1. possible NPEs due to missing default values: 
    the tests include several scenarios that test on values not being set. Most notably the one that passes an empty map as options and one that passes an empty sequence. Both pass.
2. unnecessary wrapped functions: 
    The goal of the functions is to create simple DSL for these kind of tests. A method used throughout this and other projects to speed up and ease the sometimes tedious task of writing many different test cases and therefore get to a higher test coverage.
3. the arrow operator: 
   - This operator is never necessary. It can always be substituted by deeply nesting function calls. If that is easier to read is up for discussion. 
   - If this standard XQuery 3.1 construct is hard to debug than an issue should be opened. It should contain why that is so. The implementation should not dictate what kind of constructs could be used.

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
line-o: This is not in scope of this PR

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
line-o: Please have a look a the tests that cover this.

extensions/modules/file/src/test/resources/util/helper.xqm
line-o: This function is simple at the moment but is here to be extended to cover different path separators.

extensions/modules/file/src/test/resources/util/helper.xqm
line-o: Yes, there is an issue in exist's handling of streams. I will need some time to create an issue because it is hard to pin down.

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
line-o: There are a lot of occasions where exceptions are wrapped. A significant percentage of the tests in XQTS testing for exceptions fails because of this.
I was not able to get to the root cause yet. But I did report this to the community mid last year.

line-o (commented):
Address review

extensions/modules/file/src/test/xquery/modules/file/sync.xqm
line-o: opened #4208

dizzzz (approved):

PR #4084, opened 2021-11-18
[ignore] update GitHub Action Test

line-o:
- use setup-java@v2 with builtin caching of maven artifacts
- change call to mvn:
  - use long commandline flags
  - add update-snapshots flag
  - remove clean flag


line-o:
With adopt on Windows selected as the Java distribution both actions fail with
`CollectionLocksTest.stress_noDeadlock_writeRead_subtree:601->stress_noDeadlock:701 Stress test`
Is this a known incompatibility?
@adamretter @reinhapa

line-o:
For later reference:
https://github.com/eXist-db/exist/runs/4256501805?check_suite_focus=true#step:4:15029

.github/workflows/ci-build.yml
```diff
@@ -1,33 +0,0 @@
-name: Build
```
adamretter, line 1: Why is this deleted? I thought we discussed at the Community Call...

line-o: I understood that it is no longer needed.

line-o:
So a specific distribution is required it looks like. Which one is known to work best on all platforms?

line-o:
Judging from https://github.com/actions/setup-java we were running Zulu before, switching to Zulu for the next round.

line-o:
With this setup the tests on JDK 11 on latest Ubuntu fail. :(

.github/workflows/ci-build.yml
adamretter: @duncdrum Could you comment as you set this up previously? I came away from the Community Call with a different understanding than Juri...

line-o:
Appveyour is _broken_ for some time now. see https://ci.appveyor.com/project/eXistdbci/exist/builds/41600374/job/b3yaqxoedvjc3jrs#L27

We should either fix or remove it.

line-o:
Maven AspectJ Plugin 
- dev.aspectj (depends on AspectJ 1.9.8.M1)
- com.mojohaus.aspectj (depends on AspectJ 1.9.7)

duncdrum (changes requested):
I m happy to try without separate build jobs, now that we have a the eb cache. 

I would however prefer if all CI scripts use the same `setup-java` plugin, and java binaries, could you please update the action plugin in the remaining yamls as well (docs, deploy, sonar)

If speed is the goal here, we should also disable and remove appveyor

PR #4089, opened 2021-11-26
[bugfix] allow empty enclosed expressions

### LINKED ISSUE #4088: [BUG] eXist doesn't allow XQuery 3.1 empty enclosed expression {} in direct element constructor
**Describe the bug**

XQuery 3.1 allows an empty enclosed expression `{}`, but eXist raises an error when an empty enclosed expression is enclosed inside a direct element constructor, e.g., `<x>{}</x>`. The error raised is:

> Cannot compile xquery: err:XPST0003 unexpected token: } (while expecting closing tag for element constructor: x) 

In contrast, no error is raised with the equivalent, expressed with a computed element constructor: `element x {}`.

**Expected behavior**

According to the definition of Enclosed Expressions in the XQuery 3.1 spec (https://www.w3.org/TR/xquery-31/#id-enclosed-expr), the EBNF form clearly labels the inner expression as optional:

```
EnclosedExpr ::= "{" Expr? "}"
```

... and:

> An enclosed expression is an instance of the `EnclosedExpr` production, which allows an optional expression within curly braces.] [Definition: In an enclosed expression, the optional expression enclosed in curly braces is called the content expression.] If the content expression is not provided explicitly, the content expression is `()`.

Saxon and BaseX both accept `<x>{}</x>`. 

**To reproduce**

```xquery
xquery version "3.1";

declare
    %test:assertTrue
function t:test() {
    try { util:eval('exists(<x>{}</x>)') } catch * { $err:code }
};
```

(Without the try-catch block, eXist won't even compile the test.)

This test returns the following error:

```xml
<testsuite package="http://exist-db.org/xquery/test" timestamp="2021-11-25T20:26:49.656-05:00" tests="1" failures="0" errors="1" pending="0" time="PT0.002S">
    <testcase name="test" class="t:test">
        <error type="err:FORG0006" message="Invalid argument type. value of type xs:QName has no boolean value."/>
    </testcase>
</testsuite>
```

 - OS: macOS 12.0.1
 - eXist-db version: 5.3.0 release
 - Java Version: OpenJDK 1.8.0_312 (liberica-jdk8-full 1.8.0_312-b07)

**Additional context**
- How is eXist-db installed? DMG jar

line-o:
### Description:

Allow empty expressions in both direct element and attribute constructors.

Example:

```xquery
<el
  at="{ (: empty attribute constructor :) }">
  { (: empty element constructor :) }
</el>
```

### Reference:

fixes #4088

### Type of tests:

XQSuite tests included

### DIFF
--- a/exist-core/src/test/xquery/xquery3/empty-direct-constructor.xqm
+++ b/exist-core/src/test/xquery/xquery3/empty-direct-constructor.xqm
@@ -1,0 +1,55 @@
+(:
+ : eXist-db Open Source Native XML Database
+ : Copyright (C) 2001 The eXist-db Authors
+ :
+ : info@exist-db.org
+ : http://www.exist-db.org
+ :
+ : This library is free software; you can redistribute it and/or
+ : modify it under the terms of the GNU Lesser General Public
+ : License as published by the Free Software Foundation; either
+ : version 2.1 of the License, or (at your option) any later version.
+ :
+ : This library is distributed in the hope that it will be useful,
+ : but WITHOUT ANY WARRANTY; without even the implied warranty of
+ : MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ : Lesser General Public License for more details.
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+module namespace ee="http://exist-db.org/xquery/test/enclosed-expressions";
+
+declare namespace test="http://exist-db.org/xquery/xqsuite";
+
+(:~ Enclosed expressions in element content may be empty in 3.1. Tests borrowed from QT3.
+ :
+ : @see https://github.com/w3c/qt3tests/blob/44252d2496286b0bd0ca1e3b3622f369f0890557/prod/DirElemContent.xml#L1173-L1193
+:)
+
+declare
+    %test:assertEquals("<elem>contentcontent</elem>")
+function ee:K2-DirectConElemContent-26a() {
+    <elem>content{}content</elem>
+};
+
+declare
+    %test:assertEquals("<elem>contentcontent</elem>")
+function ee:K2-DirectConElemContent-26b() {
+    <elem>content{(:comment:)}content</elem>
+};
+
+declare
+    %test:assertEquals('<elem attribute="contentcontent"/>')
+function ee:direct-attribute-constructor-empty() {
+    <elem attribute="content{}content" />
+};
+
+declare
+    %test:assertEquals('<elem attribute="contentcontent"/>')
+function ee:direct-attribute-constructor-comment() {
+    <elem attribute="content{(:comment:)}content" />
+};
--- a/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
+++ b/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
@@ -1975,6 +1975,9 @@ cdataSection : XML_CDATA;
 
 enclosedExpr throws XPathException
 :
+    ( LCURLY RCURLY ) => LCURLY! RCURLY!
+    { lexer.inElementContent= true; }
+    |
 	LCURLY^
 	{
 		globalStack.push(elementStack);
@@ -1990,6 +1993,9 @@ enclosedExpr throws XPathException
 
 attributeEnclosedExpr throws XPathException
 :
+    ( LCURLY RCURLY ) => LCURLY! RCURLY!
+    { lexer.inAttributeContent= true; }
+    |
 	LCURLY^
 	{
 		lexer.inAttributeContent= false;


adamretter:
As I understand it this is only valid in XQuery 3.1.

Can you add a check so that it is only used when the declared version is XQuery 3.1 please?

line-o:
Do you have a good idea on how to achieve that?

adamretter:
> Do you have a good idea on how to achieve that?

Off the top of my head... nope!

line-o:
Are there real world issues in allowing empty direct element constructors in earlier versions of XQuery?
From a quick glance this will be very hard to implement.

line-o:
For arrays the check is performed very late in the `eval` method of ArrayConstructor.

joewiz:
Thank you, @line-o! I tested this PR with [the XQSuite tests I adapted from the XQTS](https://github.com/eXist-db/exist/issues/4088#issuecomment-979657181), and eXist raises errors when compiling the test library module.

> /db/apps/eXide/modules/run-test.xql err:XPST0003 Error while loading module xmldb:exist:///db/test-empty-enclosed-expression.xq: No closing end tag found for element constructor: elem [at line 32, column 6]

Perhaps that test suite could be the basis of tests for this PR?

line-o:
@joewiz Yes, that is obviously the right thing to do. I just missed that :)

joewiz:
@adamretter @dizzzz For reference, BaseX and Saxon happily accept empty enclosed expressions, even when the XQuery version is explicitly 1.0 or 3.0, as in the following query:

```xquery
xquery version "1.0";

<elem>content{}content</elem>
```

I should note that other than flagging explicitly additive syntax (e.g., simple map operators, maps, arrays, etc.), eXist hasn't been particularly strict about flagging newer language constructions when earlier versions of XQuery are set in a module. For example, eXist happily accepts the `group by` clause in XQuery 1.0 modules and map constructors in XQuery 3.0 modules. 

Indeed, the `getXQueryVersion` method is only called [in these 14 source files](https://github.com/eXist-db/exist/search?q=getXQueryVersion) across the entire eXist code base:

1. ArrayConstructor.java
1. ArrayFunction.java
1. ArrowOperator.java
1. ConcatExpr.java
1. FunHeadTail.java
1. FunOnFunctions.java
1. FunSort.java
1. JSON.java
1. LoadXQueryModule.java
1. MapExpr.java
1. StringConstructor.java
1. SwitchExpression.java
1. TryCatchExpression.java
1. XQueryContext.java

Perhaps a comprehensive audit should be done, with recommendations for which language features need flagging at which levels of analysis. If methods aren't available for flagging language-specific features at particular phases, then they could be added and implemented comprehensively. 

In the meantime, though, this PR strikes me as a minor concession to enable proper XQuery 3.1 support for empty enclosed expressions. 

This may be my bias, but I'd err on the side of achieving full XQuery 3.1 compliance for eXist (including `count` clause and tumbling/sliding windows) before ensuring these 3.1-specific features are properly shielded from pre-3.1 code.

adamretter:
@joewiz I think an audit in a GitHub issue is a great idea.

I am happy to aim at 3.1 compliance rather that 1.0 or 3.0... on the understanding that doing so doesn't break backwards-compatibility with XQuery 1.0 or 3.0 code - as there is much more 1.0 code in the wild than there is 3.0 or 3.1.

exist-core/src/test/xquery/xquery3/empty-direct-constructor.xqm
```diff
@@ -0,0 +1,46 @@
...
+
+(:~ Enclosed expressions in element content may be empty in 3.1. Tests borrowed from QT3.
+ :
+ : @see https://github.com/w3c/qt3tests/blob/44252d2496286b0bd0ca1e3b3622f369f0890557/prod/DirElemContent.xml#L1173-L1193
+:)
+
+declare
+    %test:assertEquals("<elem>contentcontent</elem>")
+function ee:K2-DirectConElemContent-26a() {
+    <elem>content{()}content</elem>
```
joewiz, line 15: To align with [K2-DirectConElemContent-26a](https://github.com/w3c/qt3tests/blob/44252d2496286b0bd0ca1e3b3622f369f0890557/prod/DirElemContent.xml#L1173-L1182), the empty sequence inside the enclosed expression should be removed.

line-o: Changed it. Nothing propagates better than errors ;)

line-o:
What a pity. The test cases failing for textual content following the empty enclosed expression. 
Without problems I can evaluate `<a>{}<a />` and `<a>{(:comment:)}<a/>` and even `<a>a{}</a>`... , but `<a>{}.</a>` fails to compile.

joewiz:
@line-o Tricky! I've tried several combinations based on your function example and other LCURLY RCURLY examples like https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g#L1683-L1684, but after recompiling the same error is thrown on `<elem>content{}content</elem>`. I wish I understood the parser syntax more - I'm just throwing spaghetti at the wall, and nothing's sticking.

line-o:
The empty code block swallows what's coming later. It is too greedy, it seems.
There is still several options to try.

joewiz (approved):
Having tested this PR, eXist correctly parses empty enclosed expressions in element and attribute content. The XQSuite tests pass. Thank you, @line-o!

joewiz:
The only 2 test failures do not appear to be related to this PR:

- (JDK 11 / windows-latest) Test: org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M5:test
- (JDK 8 / macOS-latest) Test: org.owasp:dependency-check-maven:6.5.0:check

line-o:
It turned out that the lexer setting got lost for some reason. And we did not need to add a virtual AST node but it's possible to just skip the block completely.

line-o:
Squashed commits

PR #4020, opened 2021-08-30
[feature] use of Adoptiums 'temurin' JDK

reinhapa:
### Description:
Uses the Eclipse Foundation Adoptium JDK implementation for the build pipeline.

PR #4095, opened 2021-11-29
[upgrade] switch to aspectj version 1.9.8-M1

reinhapa:
### Description:

Update the aspectj library to version 1.9.8-M1

Fix compatibility by providing Typehints in JnlpServlet.java

### DIFF
--- a/exist-core/src/main/java/org/exist/webstart/JnlpServlet.java
+++ b/exist-core/src/main/java/org/exist/webstart/JnlpServlet.java
@@ -60,9 +60,9 @@ public class JnlpServlet extends HttpServlet {
 
         final Optional<Path> libDir = or(
             Optional.ofNullable(System.getProperty("app.repo")).map(Paths::get).filter(Files::exists),
-            (Supplier<Optional<Path>>) () -> or ( // using explicit type declaration in order to circumvent AspectJ compiler bug
+            (Supplier<Optional<Path>>) () -> or ( // using explicit type declaration in order to circumvent AspectJ compiler bug, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=553623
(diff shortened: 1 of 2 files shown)


reinhapa:
This PR is a precondition for #4020

exist-core/src/main/java/org/exist/webstart/JnlpServlet.java
```diff
@@ -59,9 +60,9 @@ public void init() throws ServletException {
 
         final Optional<Path> libDir = or(
             Optional.ofNullable(System.getProperty("app.repo")).map(Paths::get).filter(Files::exists),
-            () -> or (
+            (Supplier<Optional<Path>>) () -> or (
```
line-o, line 63: Please add an explanatory comment why we need that here as discussed in the community call.

reinhapa: @line-o  Do we have some special prefix in order to find such place for later clean up in case this problem gets fixed?

line-o:
@reinhapa please rebase onto current develop for the increased timeouts for windows to take effect now that #4098 is merged.

adamretter:
@line-o I already rebased it

line-o:
Well, then the fix is not working it seems as both windows test runs report a deadlock
- https://github.com/eXist-db/exist/runs/4365450511?check_suite_focus=true#step:5:8699
- https://github.com/eXist-db/exist/runs/4365450560?check_suite_focus=true#step:5:11336

reinhapa:
Hmm, I do not know what a good time out would be then 🙁

dizzzz (approved):

line-o (approved):
Tests on windows passed after restarting them.

PR #4157, opened 2022-01-03
Introduce new Store Document API

adamretter:
This enhancement has been back-ported from FusionDB. It introduces an additional new API for storing documents which is much simpler and easier to use that what was already present.

Previously, storing an XML document involved a more complex two step procedure of calling:
1. `Collection#validateXMLResource`
2. `Collection#store`

whilst storing a Binary document involved calling:
1. `Collection#validateBinaryResource`
2. `Collection#addBinary`

This new API means that for both XML and Binary documents you can just call:
```
DBBroker#storeDocument
```

I have then applied this to the code-base, which has simplified and removed a lot of redundant code. I have left the existing APIs in-place and marked them as deprecated so that they can be removed in a future major release; these changes are 100% backwards compatible at present.

Moving the API call to `DBBroker` also has the advantage of allowing us to more easily refactor how Collections are implemented in future.

### DIFF
--- a/exist-core/src/main/java/org/exist/backup/restore/RestoreHandler.java
+++ b/exist-core/src/main/java/org/exist/backup/restore/RestoreHandler.java
@@ -27,7 +27,6 @@ import org.exist.Namespaces;
 import org.exist.backup.BackupDescriptor;
 import org.exist.backup.restore.listener.RestoreListener;
 import org.exist.collections.Collection;
-import org.exist.collections.IndexInfo;
 import org.exist.collections.triggers.TriggerException;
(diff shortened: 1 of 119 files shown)


dizzzz:
Ok i will review; as it is a huge(r ) one; it will take some time;

adamretter:
Thanks @dizzzz apart from the changes in MutableCollection, most of the other changes were done via find-and-replace, so there is a lot of duplication

dizzzz (approved):
Nice piece of work!

adamretter:
@dizzzz thanks :-)

PR #4158, opened 2022-01-03
Remove unused PluginsManager

adamretter:
This removes a system called the Plugins Manager which was added to eXist-db some years ago. It was never implemented properly, i.e. its abstractions were at the wrong level of integration. It was also never documented, therefore it is unlikely that it was every used apart from perhaps by its author. It should be removed as legacy cruft, as it prevents us from more easily tidying up and improving other parts of the core.

As this is an API removal - This needs to be merged in a Major Release.


dizzzz (approved):

adamretter:
I have now rebased this to remove the merge conflicts and it should be ready to be merged.

reinhapa (approved):

PR #4161, opened 2022-01-04
Run JUnit 4 tests via JUnit 5

adamretter:
Previously there was a dependency on `junit-jupiter-api`, but we instead need  `junit-vintage-engine` on the classpath to run the tests with JUnit5.

### DIFF
--- a/exist-ant/pom.xml
+++ b/exist-ant/pom.xml
@@ -75,8 +75,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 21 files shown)


dizzzz (approved):
wondering why we did not see issues? @reinhapa

adamretter:
@dizzzz I think overall the tests were run by Surefire which was configured for JUnit5. I was just updating this as part of my dependency cleanup work when I realised that none of the code had a dependency directly on JUnit 5 API. So I checked the JUnit 5 docs and they say to instead (see: https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4):

> Just make sure that the junit-vintage-engine artifact is in your test runtime path

dizzzz (approved):

dizzzz:
@adamretter  there is a branch conflict?

adamretter:
@dizzzz Thanks I just rebased after you merged my other PR :-) ...Assuming CI passes it should be all good :-)

PR #4164, opened 2022-01-04
Enforce dependency analysis of exist-core

adamretter:
This has been an outstanding piece of work for a while, but I finally got to it :-)
This switches on dependency analysis for exist-core, which is the only module for which it was previously disabled. We can now be confident that Java code of exist-core is only declaring and using those dependencies which are described in its pom.xml file.

This PR also removes a couple of dependencies that were never used.


exist-core/pom.xml
```diff
@@ -893,7 +887,23 @@ The BaseX Team. The original license statement is also included below.]]></pream
                             <goal>analyze-only</goal>
                         </goals>
                         <configuration>
-                            <failOnWarning>false</failOnWarning>  <!-- TODO(AR) this temporarily overrides the setting in exist-parent and this plugin should be removed from here -->
+                            <failOnWarning>true</failOnWarning>
+                            <ignoredUnusedDeclaredDependencies>
+                                <ignoredUnusedDeclaredDependency>org.apache.logging.log4j:log4j-core:jar:${log4j.version}</ignoredUnusedDeclaredDependency>
```
dizzzz, line 892: interesting

exist-core/src/test/java/org/exist/util/LeasableTest.java
```diff
@@ -22,8 +22,8 @@
 
 package org.exist.util;
 
+import org.easymock.EasyMock;
```
dizzzz, line 25: does this remove some of the runtime warnings?

dizzzz (approved):
LGTM!

Issue #4150, opened 2021-12-29
[BUG] fn:generate-id not spec compliant


martin-honnen:
**Describe the bug**
`generate-id()` run on nodes from different document fragments created with `parse-xml-fragment` or `parse-xml` returns the same generated id for nodes with a different node identity.

**Expected behavior**
`generate-id` is a function defined to "In other cases, the function returns a string that uniquely identifies a given node. More formally, it is guaranteed that within a single ·execution scope·, fn:codepoint-equal(fn:generate-id($N), fn:generate-id($M)) returns true if and only if ($M is $N) returns true". 

However, when I use it with eXist-db, precisely in the eXide online tool, I find that the result returned by `generate-id()` for nodes with a different identity is the same.

**To Reproduce**

```xquery
xquery version "3.1";

let $node1 := parse-xml-fragment('<node>a</node>')/node(),
    $node2 := parse-xml-fragment('<node>b</node>')/node()
return ($node1 is $node2, codepoint-equal(generate-id($node1), generate-id($node2)))
```
gives `false()
true()` in the eXide web site.

**Context (please always complete the following information):**

"Only" tested online at http://exist-db.org/exist/apps/eXide/index.html

dnovatchev:
This is an obvious bug.

Compare with two other, correct behaviors (BaseX and Saxon) - both produce `false(), false()`

dizzzz:
Not clear about the "eXide web site: ; the demo server is rather old AFAIR.

adamretter:
@martin-honnen Thanks very much for reporting, I have submitted a fix here - https://github.com/eXist-db/exist/pull/4167
As soon as it gets merged, it will appear in the next Nightly Build available from - http://www.evolvedbinary.com/afterdark.html

PR #4167, opened 2022-01-06
fn:generate-id must provide unique IDs for unique Nodes

adamretter:
Closes https://github.com/eXist-db/exist/issues/4150

### DIFF
--- a/exist-core/src/test/xquery/xquery3/generate-id.xqm
+++ b/exist-core/src/test/xquery/xquery3/generate-id.xqm
@@ -31,7 +31,8 @@ declare
 function gid:setup() {
     xmldb:create-collection("/db", "generate-id-test"),
     xmldb:store("/db/generate-id-test", "a.xml", <node>a</node>),
-    xmldb:store("/db/generate-id-test", "a.xml", <node>b</node>)
+    xmldb:store("/db/generate-id-test", "a.xml", <node>b</node>),
+    xmldb:store("/db/generate-id-test", "inner-a.xml", <node><inner>a</inner></node>)
 };
 
 declare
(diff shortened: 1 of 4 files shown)


martin-honnen:
I am wondering, however, whether the new value containing colons, if I read the code right, returns a valid `xs:ID` value. The function's spec says "The returned identifier must consist of ASCII alphanumeric characters and must start with an alphabetic character.". Not quite sure whether a colon `:` is allowed there and whether I read the Java code generating the id correctly, but it seems to do `nodeImplementationType + ":" + docId + ":" + nodeId`, i.e. concatenate the three parts with colons.

adamretter:
@martin-honnen You are right of course! Somehow I missed that bit when reading the spec. I have pushed a commit and tests which correct that.

exist-core/src/test/xquery/xquery3/generate-id.xqm
```diff
@@ -0,0 +1,141 @@
...
+
+declare namespace test = "http://exist-db.org/xquery/xqsuite";
+declare namespace xmldb = "http://exist-db.org/xquery/xmldb";
+
+declare
+    %test:setUp
+function gid:setup() {
+    xmldb:create-collection("/db", "generate-id-test"),
+    xmldb:store("/db/generate-id-test", "a.xml", <node>a</node>),
+    xmldb:store("/db/generate-id-test", "a.xml", <node>b</node>),
```
dizzzz, line 34: is document name correct?

exist-core/src/test/xquery/xquery3/generate-id.xqm
```diff
@@ -0,0 +1,141 @@
...
+        starts-with($id, "M"),
+        matches($id, "[a-zA-Z0-9]+")
+    )
+};
+
+declare
+    %test:assertEquals("false", "true", "true")
+function gid:persistent-dom-document-identity() {
+  let $node1 := doc("/db/generate-id-test/a.xml"),
+      $node2 := doc("/db/generate-id-test/b.xml")
```
dizzzz, line 109: this document does not exist , right?

dizzzz (approved):
I have a question, inline; not sure if I understand the tests, fully;
if I am... why the test results? are these reliable?

exist-core/src/test/xquery/xquery3/generate-id.xqm
adamretter: Nope. I just pushed a commit to fix that’ll . Thanks

exist-core/src/test/xquery/xquery3/generate-id.xqm
adamretter: Indeed.

adamretter:
> if I am... why the test results? are these reliable?

There was a mistake in the test, but the test still passed because the missing document evaluated to an empty-sequence, which was valid for all of the tests I made.

I have now fixed the missing document, and added a guard to ensure all documents are present.

Issue #1562, opened 2017-09-26
util:expand alters node type of attribute and document nodes


joewiz:
### What is the problem

The `util:expand()` function alters the node types of attribute and document nodes instead of preserving their original node type.

### What did you expect

The function documentation describes the function as follows:

> Creates an in-memory copy of the passed node set, using the specified serialization options. 
> `$node` as `node()*` The node(s) to create in-memory copies of.
> Returns: `node()*` the results

Thus, this function should return an in-memory copy of the original node set—and the node type of each passed node should be preserved in the output.

### Describe how to reproduce or add a test

The following query illustrates the problem:

```xquery
xquery version "3.1";

(: these correctly return true :)
util:expand( comment { "x" } ) instance of comment(),
util:expand( processing-instruction x { "" } ) instance of processing-instruction(),
util:expand( text { "x" } ) instance of text(),
util:expand( element x {()} ) instance of element(),

(: these incorrectly return false :)
util:expand( attribute x {""} ) instance of attribute(),
util:expand( document { element x {()} } ) instance of document-node(),

(: these incorrectly return true :)
util:expand( attribute x {""} ) instance of text(),
util:expand( document { element x {()} } ) instance of element()
```

The results shows that attribute nodes are being transformed into text nodes, and document nodes are being transformed into element nodes.

The following XQSuite (attached here: [util-expand-test.xql.zip](https://github.com/eXist-db/exist/files/1333602/util-expand-test.xql.zip)) also demonstrates the problem.

```xquery
xquery version "3.1";

module namespace ue="http://joewiz.org/ns/test/exist/util/expand";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare
    %test:assertTrue
function ue:attribute() as xs:boolean {
    util:expand( attribute foo { "" } ) instance of attribute()
};

declare
    %test:assertTrue
function ue:comment() as xs:boolean {
    util:expand( comment { "foo" } ) instance of comment()
};

declare
    %test:assertTrue
function ue:document() as xs:boolean {
    util:expand( document { element foo {()} } ) instance of document-node()
};

declare
    %test:assertTrue
function ue:element() as xs:boolean {
    util:expand( element foo {()} ) instance of element()
};

declare
    %test:assertTrue
function ue:pi() as xs:boolean {
    util:expand( processing-instruction foo { "" } ) instance of processing-instruction()
};

declare
    %test:assertTrue
function ue:text() as xs:boolean {
[...]

Issue #1486, opened 2017-07-19
util:expand turns attributes into text nodes


joewiz:
### What is the problem

When passing an attribute node to `util:expand()`, the function returns a `text()` node instead of an `attribute()` node. 

This is responsible for a bug in eXide. When results are serialized with the "expand matches" option enabled, attributes which should be adaptive-serialized as `foo="bar"` are serialized instead as `bar`. 

### What did you expect

I expected it to return the same node type as the set passed in. The function docs for `util:expand()` say it "Creates an in-memory copy of the passed node set, using the specified serialization options." See http://exist-db.org/exist/apps/fundocs/view.html?uri=http://exist-db.org/xquery/util#expand.1.

The code in eXide where the bug is exhibited is https://github.com/wolfgangmm/eXide/blob/develop/modules/session.xql#L51-L52. Changing the conditional on line 51 to `element()` fixes the problem, but based on the function documentation for `util:expand()`, this function shouldn't be limited to just the `element()` node type.

### Describe how to reproduce or add a test

Here is an xqsuite test:

```xquery
xquery version "3.1";

module namespace et="http://exist-db.org/xquery/util/expand";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare variable $et:test := attribute foo { "bar" };

declare %test:assertEquals("attribute") function et:util-expand-arity-1() {
    let $expanded := util:expand($et:test)
    return
        typeswitch ($expanded) 
            case attribute() return "attribute"
            case text() return "text"
            default return "unexpected instance type"
};

declare %test:assertEquals("attribute") function et:util-expand-arity-2() {
    let $expanded := util:expand($et:test, "method=adaptive")
    return
        typeswitch ($expanded) 
            case attribute() return "attribute"
            case text() return "text"
            default return "unexpected instance type"
};
```

### Context information

- eXist-db version + Git Revision hash: 3.4.0-SNAPSHOT+201707142259 (af161fd52) 
- Java version: 1.8.0_131
- Operating system: macOS 10.12.5
- 32 or 64 bit: 64 bit
- Any custom changes in e.g. conf.xml: No

joewiz:
(I updated the issue with a better test.)

joewiz:
Superceded by #1562.

PR #4172, opened 2022-01-07
util:expand must handle Document and Attribute Nodes correctly

adamretter:
Previously `util:expand` did not handle Document or AttributeNodes correctly.

Closes https://github.com/eXist-db/exist/issues/1562
Closes https://github.com/eXist-db/exist/issues/1486

For Document Nodes instead it would just return the first node from the document. If the first node was an element, you likely would not have noticed a problem (as it is a container of child nodes), however if the first node was a comment, text node, or processing instruction, then you would not see any of the other descendant nodes of the document.

For example, previously:
```xquery
let $doc1 := document { <!-- comment 1 before --><!-- comment 2 before --><doc2>doc2</doc2> }
let $doc1-uri := xmldb:store("/db", "doc1.xml", $doc1, "application/xml")
return
  util:expand(doc($doc1-uri))
```
would incorrectly return only the first node in the document, i.e.:
```xml
<!-- comment 1 before -->
```

After this fix, it now correctly returns:
```xml
<!-- comment 1 before --><!-- comment 2 before --><doc2>doc2</doc2>
```

For Attribute Nodes, it would previously incorrectly return a Text Node holding the value of the Attribute.

### DIFF
--- a/exist-core/src/main/java/org/exist/dom/memtree/MemTreeBuilder.java
+++ b/exist-core/src/main/java/org/exist/dom/memtree/MemTreeBuilder.java
@@ -26,11 +26,13 @@ import org.exist.Namespaces;
 import org.exist.dom.QName;
 import org.exist.dom.persistent.NodeProxy;
 import org.exist.xquery.Constants;
+import org.exist.xquery.XQuery;
 import org.exist.xquery.XQueryContext;
 import org.w3c.dom.DOMException;
 import org.w3c.dom.Node;
 import org.xml.sax.Attributes;
 
+import javax.annotation.Nullable;
 import javax.xml.XMLConstants;
 import java.util.Arrays;
 
@@ -42,7 +44,7 @@ import java.util.Arrays;
  */
 public class MemTreeBuilder {
 
-    private final XQueryContext context;
+    private XQueryContext context;
     private DocumentImpl doc;
     private short level = 1;
     private int[] prevNodeInLevel;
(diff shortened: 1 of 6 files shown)


joewiz:
@adamretter This is great! I re-ran the tests I originally wrote for #1562, and whereas before the test checking the handling of document nodes failed, now it passes. (The attribute test still fails, but this PR focuses on document nodes.)

adamretter:
@joewiz I see! No problem, I just added a commit that fixes the problem with Attribute Nodes too! I also incorporated your XQSuite tests from https://github.com/eXist-db/exist/issues/1562 - thank you :-)

joewiz (approved):
Testing the updated PR, all of my tests pass:

```xml
<testsuite package="http://joewiz.org/ns/test/exist/util/expand"
    timestamp="2022-01-08T09:59:57.144-05:00" tests="6" failures="0" errors="0" pending="0"
    time="PT0.095S">
    <testcase name="attribute" class="ue:attribute"/>
    <testcase name="comment" class="ue:comment"/>
    <testcase name="document" class="ue:document"/>
    <testcase name="element" class="ue:element"/>
    <testcase name="pi" class="ue:pi"/>
    <testcase name="text" class="ue:text"/>
</testsuite>
```

And, for the first time ever, eXide correctly serializes attributes when the "highlight index matches" option (enabled by default) is checked - since it runs all results through `util:expand`.

Thank you, @adamretter!

dizzzz (approved):

PR #4174, opened 2022-01-09
[4.x.x] Update libs

adamretter:
Updates all libraries to the latest versions that are compatible with eXist-db 4.x.x.
The most visible changes for end-users will likely be:
1. The update of Tika for the Content Extraction module
2. The update of Apache FOP for the XSL:FO module


joewiz:
Wow. What a service to those still on 4.x!

dizzzz (approved):

PR #4188, opened 2022-01-14
[4.x.x] Fix issue with Collection structures being overwritten during startup

adamretter:
This fixes a serious issue in 4.x.x where during the database startup, some Collection data could be overwritten. This happened because the Collection Creation algorithm expected the Collection Cache to be enabled, however it was disabled during startup. To fix this, I simply replaced the Collection Cache in 4.x.x with my new implementation from 5.x.x, which enables to the Collection Cache to be used during startup.

Test which shows that this doesn't happen in 5.x.x is in https://github.com/eXist-db/exist/pull/4187


dizzzz (approved):

dizzzz:
Do we need to update Monex for this?

adamretter:
> Do we need to update Monex for this?

@dizzzz Honestly, I don't know.

dizzzz (approved):

adamretter:
Thanks @dizzzz for merging :-)

PR #4196, opened 2022-01-18
Import jetty Properties from SystemProperties

agh2342:
This unifies handling of jetty variables set via JAVA_OPTS environment variable.
The following jetty Properties can now be set via JAVA_OPTS: [jetty.host, jetty.port, ssl.port, jetty.http.host, jetty.http.port, jetty.ssl.host, jetty.ssl.port].

to clarify: the PR enables e.g. the following inside an existdb home directory
```
$ JAVA_OPTS='-Djetty.host=127.0.0.2 -Djetty.http.port=9999' ./bin/startup.sh
```
or putting the options directly on the java command
```
$ /usr/bin/java ... \
   -Djetty.host=127.0.0.2 \
   -Djetty.http.port=9999 \
   ... org.codehaus.mojo.appassembler.booter.AppassemblerBooter
```

### DIFF
--- a/exist-jetty-config/src/main/resources/org/exist/jetty/etc/jetty-http.xml
+++ b/exist-jetty-config/src/main/resources/org/exist/jetty/etc/jetty-http.xml
@@ -34,8 +34,8 @@
             </Item>
           </Array>
         </Arg>
-        <Set name="host"><Property name="jetty.http.host" deprecated="jetty.host" /></Set>
-        <Set name="port"><Property name="jetty.http.port" deprecated="jetty.port"><Default><SystemProperty name="jetty.port" default="8080"/></Default></Property></Set>
+        <Set name="host"><Property name="jetty.http.host" deprecated="jetty.host"><Default><SystemProperty name="jetty.http.host" deprecated="jetty.host"/></Default></Property></Set>
+        <Set name="port"><Property name="jetty.http.port" deprecated="jetty.port"><Default><SystemProperty name="jetty.http.port" deprecated="jetty.port" default="8080"/></Default></Property></Set>
         <Set name="idleTimeout"><Property name="jetty.http.idleTimeout" deprecated="http.timeout" default="30000"/></Set>
         <Set name="acceptorPriorityDelta"><Property name="jetty.http.acceptorPriorityDelta" deprecated="http.acceptorPriorityDelta" default="0"/></Set>
         <Set name="acceptQueueSize"><Property name="jetty.http.acceptQueueSize" deprecated="http.acceptQueueSize" default="0"/></Set>
--- a/exist-jetty-config/src/main/resources/org/exist/jetty/etc/jetty-ssl.xml
+++ b/exist-jetty-config/src/main/resources/org/exist/jetty/etc/jetty-ssl.xml
@@ -26,7 +26,7 @@
           </Array>
         </Arg>
 
-        <Set name="host"><Property name="jetty.ssl.host" deprecated="jetty.host" /></Set>
+        <Set name="host"><Property name="jetty.ssl.host" deprecated="jetty.host"><Default><SystemProperty name="jetty.ssl.host" deprecated="jetty.host"/></Default></Property></Set>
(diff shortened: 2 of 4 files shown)


ljo (approved):

line-o (approved):

dizzzz:
@adamretter didn't we have an alternative mechanism for this?

adamretter:
@dizzzz to be honest I thought this was already supported using '-D' syntax... I'll take a look when I can find some time

dizzzz:
@adamretter  I had the same remembering (if that is correct English and not Dunglish)

dizzzz:
Pretty sure it should work out of the box
- https://stackoverflow.com/questions/24644114/change-jetty-default-port
- https://jike.in/?qa=1024172/java-how-to-override-jetty-xml-with-jetty-port
-

dizzzz (commented):
I think this is not needed

line-o:
@agh2342 could you test if using `-Djetty.port=9999` work on current develop?

agh2342:
@dizzzz thanks for your feedback/investigation.
with latest develop:
* `-Djetty.port=9999` works (for HTTP), but is deprecated
* `-Djetty.http.port=9999` does not work
* `-Dssl.port=9999` works, but is deprecated
* `-Djetty.ssl.port=9999` works
* `-Djetty.host=127.0.0.2` does not work, and is deprecated
* `-Djetty.http.host=127.0.0.2` does not work
* `-Djetty.ssl.host=127.0.0.2` does not work

agh2342:
i think having consistent behaviour for different schemes and being able to set listening IP address and port via JAVA_OPTS unsing non-deprecated settings is desirable.

line-o:
I want to add that being able to set the host is crucial. This will allow development on a local machine be restricted to local loopback adapter (127.0.0.1)

adamretter:
If this involves renaming existing and documented properties, then this needs to go into a major release e.g. 7.0.0 and it needs to be documented

line-o:
I would like to learn where the current flags are documented.

adamretter:
> I would like to learn where the current flags are documented.

I don't think they are in the exist-documentation AFAIK, but if you search exist-open mailing list you will find them mentioned there and a few times.

line-o:
So, to summarise the information given here:

- the current java tool options are 
  - not documented properly and 
  - are in part only available via deprecated names and 
  - some are even missing
- this PR exposes valuable options 
  -  through SystemProperties not available at the moment (specifically setting the host)

What is the best way forward? 

I would propose to have the java tool options work as they are intended, specifically
  - `jetty.http.host`
  - `jetty.ssl.host`
  - `jetty.http.port`

That way new users are able to set those with information readily available in numerous online resources.

adamretter:
> are in part only available via deprecated names and

I don't agree with that statement.

If we want the new names, we also need to continue to support the old names, as these have been in-use for at least 10 years!

line-o:
So it turns out to be already in the form everyone agrees to be correct.

line-o:
And just to make it extra explicit: all current flags are still supported in the same way as they were before.

line-o:
I will merge this first thing tomorrow unless somebody objects.

line-o:
@joewiz would you be so kind as to propose a good place to add the startup options as there is no documentation about the existing ones.

line-o:
I found vaguely matching articles
- https://github.com/eXist-db/documentation/blob/master/src/main/xar-resources/data/deployment/deployment.xml
- https://github.com/eXist-db/documentation/blob/master/src/main/xar-resources/data/advanced-installation/advanced-installation.xml (mentions startup.sh)
- https://github.com/eXist-db/documentation/blob/master/src/main/xar-resources/data/basic-installation/basic-installation.xml (also mentions startup.sh but may be a little less basic)

joewiz:
The corresponding documentation has now been merged, but, as discussed during Monday's Community Call, please hold on merging this until 5.4.1 and 6.0.1 have been released.

line-o:
Now that the releases are done I merged this PR

PR #4204, opened 2022-01-23
Fix NPE in FunctionCall for Deferred Forward Reference

adamretter:
Fixes a `NullPointerException` when analyze was called on a `FunctionCall` that was a Deferred Function Reference that had not yet been resolved.

### DIFF
--- a/exist-core/pom.xml
+++ b/exist-core/pom.xml
@@ -676,6 +676,7 @@
                                 <exclude>src/test/java/org/exist/storage/txn/TransactionManagerTestHelper.java</exclude>
                                 <exclude>src/test/java/org/exist/storage/txn/TxnTest.java</exclude>
                                 <exclude>src/main/java/org/exist/test/DiffMatcher.java</exclude>
+                                <exclude>src/test/java/org/exist/test/Util.java</exclude>
                                 <exclude>src/main/java/org/exist/util/CollectionOfArrayIterator.java</exclude>
                                 <exclude>src/test/java/org/exist/util/CollectionOfArrayIteratorTest.java</exclude>
                                 <exclude>src/main/java/org/exist/util/IPUtil.java</exclude>
@@ -818,6 +819,7 @@ The original license statement is also included below.]]></preamble>
                                 <include>src/test/java/org/exist/storage/txn/TransactionManagerTestHelper.java</include>
                                 <include>src/test/java/org/exist/storage/txn/TxnTest.java</include>
                                 <include>src/main/java/org/exist/test/DiffMatcher.java</include>
+                                <include>src/test/java/org/exist/test/Util.java</include>
                                 <include>src/main/java/org/exist/util/CollectionOfArrayIterator.java</include>
                                 <include>src/test/java/org/exist/util/CollectionOfArrayIteratorTest.java</include>
                                 <include>src/main/java/org/exist/util/IPUtil.java</include>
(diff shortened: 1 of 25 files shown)


line-o:
Good stuff.
I have two questions: 
1. You introduce a yet another [utility class](exist-core/src/test/java/org/exist/test/Util.java) to run xqueries in jUnit tests. Why is is this class necessary or do you consider this implementation to be superior? Can and should other tests be rewritten to use this? And if yes why is that so?
2. The tests use a very specific use case that makes it hard to follow and run XQsuite tests inside of jUnit tests. Why is this necessary? Can the tests be rewritten to more clearly communicate what is being tested? Naming the function whose call is deferred `deferred-function` could improve the clarity already a lot.

adamretter:
>You introduce a yet another utility class to run xqueries in jUnit tests. Why is is this class necessary or do you consider this implementation to be superior? Can and should other tests be rewritten to use this? And if yes why is that so?

When you say "yet another" - I am not sure which other ones you are talking about?

> The tests use a very specific use case that makes it hard to follow and run XQsuite tests inside of jUnit tests. Why is this necessary?

This was an incredibly hard to reproduce issue, which is why the test is complex. It is not possible to reproduce this using only XQSuite as XQSuite is written in XQuery and it gets in the way of controlling this.

> Naming the function whose call is deferred deferred-function could improve the clarity already a lot.

Sorry which function are you suggesting should be renamed?

line-o:
> When you say "yet another" - I am not sure which other ones you are talking about?

`exist-core/src/main/java/org/exist/test/XQueryCompilationTest.java` was introduced in #3605 .
You were actively involved in getting it merged into this project.

> This was an incredibly hard to reproduce issue, which is why the test is complex. It is not possible to reproduce this using only XQSuite as XQSuite is written in XQuery and it gets in the way of controlling this.

And you have to run XQSuite tests inside of jUnit tests in order to reproduce this issue? Wasn't there an issue opened regarding this? If yes could you please reference it here.

> > Naming the function whose call is deferred deferred-function could improve the clarity already a lot.
> Sorry, which function are you suggesting should be renamed?

If I could tell from the tests which call to what function is deferred here I would not have to ask for the renaming.

adamretter:
> exist-core/src/main/java/org/exist/test/XQueryCompilationTest.java was introduced in #3605 .
You were actively involved in getting it merged into this project.

I never liked that approach as it involves extending a base class, and the helper methods could be more helpful. I helped get it in solely for the purpose of getting your actual fix in.

> And you have to run XQSuite tests inside of jUnit tests in order to reproduce this issue?

Yes.

> Wasn't there an issue opened regarding this? If yes could you please reference it here.

No issue that I am aware of.

> If I could tell from the tests which call to what function is deferred here I would not have to ask for the renaming.

I am not really sure what you are asking me to rename - can you be clearer please?

line-o:
The first code smell should be fixed
https://sonarcloud.io/project/issues?id=eXist-db_exist&open=AX6HG7mRnYFqFi-YdV5D&pullRequest=4204&resolved=false&types=CODE_SMELL

line-o:
I added this to todays agenda for the community call.

adamretter:
> The first code smell should be fixed

Done.

dizzzz:
I will review tonight

dizzzz (approved):

PR #4215, opened 2022-01-30
Restore libraries required for Monex's Remote Console pane

joewiz:
### Description:

Restores jetty-annotations, ~~jetty-jndi, and jetty-plus~~ - libraries that Monex's Remote Console pane requires to be in place.

### Reference:

https://github.com/eXist-db/monex/issues/175#issuecomment-1025252020

### Type of tests:

None included, but I tested by opening http://localhost:8080/exist/apps/monex/console.html and confirming that Monex's Remote Console now displays, "Connected" - whereas previously it could never connect and instead displayed "Disconnected."

### DIFF
--- a/exist-core/pom.xml
+++ b/exist-core/pom.xml
@@ -551,6 +551,14 @@
             <version>1.4</version>
         </dependency>
         <dependency>
+            <!--
+                 NOTE(AR) This is needed to enable loading of Servlets etc
+                          that use Java Annotations for configuration.
+                          This is less than ideal as it allows any extension to
+                          serve arbitary Web requests, and should ultimately be
+                          removed. Unfortunately, at this time, it is required for
+                          Monex's Remote Console to function.
+            -->
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-annotations</artifactId>
         </dependency>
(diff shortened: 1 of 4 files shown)


adamretter:
@joewiz the entries from the `exist-core/pom.xml` will need to be reinstated and they will need to be added to the ignore part of the dependency plugin with a note of why they are required. Also, are both of these actually required for Monex? I was hopeful that if we did need to reinstate, then only `jetty-annotations` would actually be needed...

joewiz:
@adamretter Ah, I see! I'll try that and will work on an updated PR.

joewiz:
Ok, here's a second attempt. It drops the jndi and plus, keeping only annotations - and registering it as an ignoredUnusedDeclaredDependency. 

On my local build, Monex's Remote Console pane is able to connect and receive messages sent via `console:log()`.

joewiz:
I also tracked down when jetty-annotations was first added to the code base: https://github.com/eXist-db/exist/commit/5b023589aca1d80443fdf77d917fc2017e2b304b. It was between 2.0 and 2.1, and it was added without jetty-jndi and jetty-plus.

line-o (approved):
That is great news @joewiz  
I believe that this will need backporting for 5.4.0 and we have to discuss if this also needs a 5.4.1 and 6.0.1 release as both broke existing behaviour.

adamretter:
Thanks @joewiz I added a commit with a note about why the dependency is needed so that it doesn't get removed until Monex is fixed.

PR #4220, opened 2022-02-03
[upgrade] eXist changes for next XML:DB API version

reinhapa:
This PR contains all changes done for the next [XMLDB:API](https://github.com/xmldb-org/xmldb-api) version.

### DIFF
--- a/exist-ant/src/main/java/org/exist/ant/RestoreTask.java
+++ b/exist-ant/src/main/java/org/exist/ant/RestoreTask.java
@@ -43,11 +43,11 @@ import java.nio.file.Path;
  */
 public class RestoreTask extends AbstractXMLDBTask
 {
-    private Path zipFile         = null;
-    private Path   dir             = null;
-    private DirSet dirSet          = null;
-    private String restorePassword = null;
-    private boolean overwriteApps = false;
+    private Path zipFile;
+    private Path dir;
+    private DirSet dirSet;
+    private String restorePassword;
+    private boolean overwriteApps;
 
     @Override
     public void execute() throws BuildException
@@ -61,7 +61,7 @@ public class RestoreTask extends AbstractXMLDBTask
         }
 
         if( ( dir != null ) && !Files.isReadable(dir)) {
-            final String msg = "Cannot read restore file: " + dir.toAbsolutePath().toString();
+            final String msg = "Cannot read restore file: " + dir.toAbsolutePath();
 
             if( failonerror ) {
                 throw( new BuildException( msg ) );
@@ -75,11 +75,11 @@ public class RestoreTask extends AbstractXMLDBTask
             try {
 
                 if( dir != null ) {
-                    log( "Restoring from " + dir.toAbsolutePath().toString(), Project.MSG_INFO );
+                    log( "Restoring from " + dir.toAbsolutePath(), Project.MSG_INFO );
                     final Path file = dir.resolve("__contents__.xml" );
 
                     if( !Files.exists(file)) {
-                        final String msg = "Could not find file " + file.toAbsolutePath().toString();
+                        final String msg = "Could not find file " + file.toAbsolutePath();
 
                         if( failonerror ) {
                             throw( new BuildException( msg ) );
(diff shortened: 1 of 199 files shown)


exist-ant/src/main/java/org/exist/ant/RestoreTask.java
```diff
@@ -47,7 +47,7 @@
     private Path   dir             = null;
     private DirSet dirSet          = null;
     private String restorePassword = null;
-    private boolean overwriteApps = false;
+    private final boolean overwriteApps = false;
```
dizzzz, line 50: interesting

exist-ant/src/main/java/org/exist/ant/XMLDBExtractTask.java
```diff
@@ -133,8 +134,8 @@ public void execute() throws BuildException {
      * @throws IOException if an I/O error occurs
      */
     private void extractResources(final Collection base, final String path) throws XMLDBException, IOException {
-        final String[] resources = base.listResources();
-        if (resources != null) {
+        final List<String> resources = base.listResources();
+        if (!resources.isEmpty()) {
```
dizzzz, line 138: no risk for nPE?

exist-core/src/main/java/org/exist/client/InteractiveClient.java
```diff
@@ -377,39 +377,38 @@ protected void getResources() throws XMLDBException {
...
-        resources = new String[childCollections.length + childResources.length];
+        resources = new String[childCollections.size() + childResources.size()];
         //Collection child;
         Permission perm;
 
         final List<ResourceDescriptor> tableData = new ArrayList<>(resources.length); // A list of ResourceDescriptor for the GUI
 
         int i = 0;
-        for (; i < childCollections.length; i++) {
+        for (; i < childCollections.size(); i++) {
```
dizzzz, line 391: looks like a better orlop can babe used, not using the index?

exist-core/src/main/java/org/exist/client/NewResourceDialog.java
```diff
@@ -287,7 +287,7 @@ private void createResource(final ResourceType resourceType, final String filena
         
         try {
             final String resName = URIUtils.urlEncodeUtf8((isNullOrEmpty(filename) ? DEFAULT_FILENAME : filename) + "." + resourceType.getFileExtension());
-            final String resType = resourceType == ResourceType.XML_DOCUMENT ? XMLResource.RESOURCE_TYPE : BinaryResource.RESOURCE_TYPE;
+            final Class<? extends Resource> resType = resourceType == ResourceType.XML_DOCUMENT ? XMLResource.class : BinaryResource.class;
```
dizzzz, line 290: for readability split over lines?

dizzzz (commented):
nice work

exist-core/src/main/java/org/exist/client/InteractiveClient.java
reinhapa: I was too tiered to tackle this in the first pass  😊

exist-ant/src/main/java/org/exist/ant/RestoreTask.java
reinhapa: jep, guess this change was made by the IDE though. Still interesting though...

exist-ant/src/main/java/org/exist/ant/XMLDBExtractTask.java
reinhapa: The API states that there no null results are allowed instead an empty list must be returned.

exist-core/src/main/java/org/exist/client/NewResourceDialog.java
reinhapa: I don't seem to get your point here... The `resType` is now of type `Class` type not a `String` any more and the generic definition is needed in order to work for the later call.

line-o: I believe @dizzzz was only referring to the formatting.
Maybe he had something like this in mind.

```
final Class<? extends Resource> resType = resourceType == ResourceType.XML_DOCUMENT 
    ? XMLResource.class
    : BinaryResource.class;
```

reinhapa: I see... check for that with the next batch of changes then

line-o:
@reinhapa do you plan to change the format of your commit messages to `[enhancement/bugifix/...] message` ?
More info on "Commit tagging" can be found in the [eXist-db Developer Manifesto]http://www.exist-db.org/exist/apps/doc/devguide_manifesto.xml

reinhapa:
> @reinhapa do you plan to change the format of your commit messages to `[enhancement/bugfix/...] message` ? More info on "Commit tagging" can be found in the [eXist-db Developer Manifesto](http://www.exist-db.org/exist/apps/doc/devguide_manifesto.xml)

I have this in mind, the question is on how many commits I should reduce the existing ones, as there are some intermittent changes that will no longer compile in that way as the XML:DB API has changed over time. Feedback also from @adamretter  would be appreciated there.

Issue #4213, opened 2022-01-28
[BUG] WebDAV error  409 "Conflict" when updating a file - which is the same like deleting these!


Spaxx71:
After updating from eXist-db 5.3.1 to 6.0.0 the webdav access often fails when updating an xml, xsl or xconf file.
Before updating eXist-db to Version 6 all worked fine, as long as the following two Java options are given:
`-Dorg.exist.webdav.PROPFIND_METHOD_XML_SIZE=EXACT -Dorg.exist.webdav.GET_METHOD_XML_SIZE=EXACT`

The exist.logs looks like this:

```
2022-01-28 12:18:03,028 [qtp1993812827-32] INFO  (HttpManager.java [process]:213) - PUT :: http://exist-proxy/exist/webdav/db/apps/mywebsite/general/templates/generate-html.xsl - http://exist-proxy/exist/webdav/db/apps/mywebsite/general/templates/generate-html.xsl 
2022-01-28 12:18:03,031 [qtp1993812827-32] INFO  (MiltonResource.java [authorise]:356) - PUT /db/apps/mywebsite/general/templates (write=true) 
2022-01-28 12:18:03,036 [qtp1993812827-32] ERROR (NativeBroker.java [dropDomNodes]:2862) - Caught NPE in DOMTransaction to actually be able to remove the document. 
2022-01-28 12:18:03,036 [qtp1993812827-32] WARN  (TransactionManager.java [close]:421) - Transaction was not committed or aborted, auto aborting! 
2022-01-28 12:18:03,037 [qtp1993812827-32] ERROR (ExistCollection.java [createFile]:327) - org.xml.sax.SAXException: fatal error at (-1,-1) : Premature end of file.
org.xml.sax.SAXParseException; Premature end of file. 
2022-01-28 12:18:03,037 [qtp1993812827-32] WARN  (StandardFilter.java [process]:62) - conflictException: Conflict exception: templates 
2022-01-28 12:18:04,992 [qtp1993812827-49] INFO  (HttpManager.java [process]:213) - PROPFIND :: http://exist-proxy/exist/webdav/db/apps/mywebsite/general/templates/ - http://exist-proxy/exist/webdav/db/apps/mywebsite/general/templates/ 
2022-01-28 12:18:04,993 [qtp1993812827-49] INFO  (MiltonResource.java [authorise]:356) - PROPFIND /db/apps/mywebsite/general/templates (write=false) 
[...]

PR #4230, opened 2022-02-08
XML documents were stored incorrectly by WebDAV 

adamretter:
Closes https://github.com/eXist-db/exist/issues/4213

### DIFF
--- a/extensions/webdav/src/test/resources/log4j2.xml
+++ b/extensions/webdav/src/test/resources/log4j2.xml
@@ -1,0 +1,251 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    eXist-db Open Source Native XML Database
+    Copyright (C) 2001 The eXist-db Authors
+
+    info@exist-db.org
+    http://www.exist-db.org
+
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2.1 of the License, or (at your option) any later version.
+
+    This library is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+    Lesser General Public License for more details.
+
+    You should have received a copy of the GNU Lesser General Public
+    License along with this library; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+-->
+<Configuration xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <Properties>
+        <Property name="logs">${log4j:configParentLocation}/../../target/test-logs-${date:yyyyMMddHHmmssSSS}</Property>
+        <Property name="rollover.max.size">10MB</Property>
+        <Property name="rollover.max">14</Property>
+        <Property name="rollover.file.pattern">%d{yyyyMMddHHmmss}</Property>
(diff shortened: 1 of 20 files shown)


joewiz:
@adamretter I've tested the PR and am seeing no regressions in my use of WebDAV to upload/download/list resources in eXist with these clients on macOS: oXygen, Transmit, and Finder.

extensions/webdav/src/test/resources/standalone-webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,116 @@
...
+        <!--
+            Some WebDAV clients send a "Expect: 100-continue" header before
+            uploading body data. Servlet containers (like tomcat and jetty) handle
+            the header in a wrong way, making a client not work OK.
+            Set value to TRUE to restore old behavior (FALSE is the new default
+            value, hardcoded in MiltonWebDAVServlet).
+        -->
+        <!--
+            <init-param>
+            <param-name>enable.expect.continue</param-name>
```
line-o, line 84: Are there any advantages in allowing this again? 
And why did it work prior to version 5.4.0?

extensions/webdav/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,251 @@
...
+
+        <RollingRandomAccessFile name="exist.xmldb" filePattern="${logs}/xmldb.${rollover.file.pattern}.log.gz" fileName="${logs}/xmldb.log">
+            <Policies>
+                <SizeBasedTriggeringPolicy size="${rollover.max.size}"/>
+            </Policies>
+            <DefaultRolloverStrategy max="${rollover.max}"/>
+            <PatternLayout pattern="${exist.file.pattern}"/>
+        </RollingRandomAccessFile>
+
+        <RollingRandomAccessFile name="exist.xmlrpc" filePattern="${logs}/xmlrpc.${rollover.file.pattern}.log.gz" fileName="${logs}/xmlrpc.log">
```
line-o, line 52: Do we need the XML-RPC logs for webdav testing? Can this and others be safely removed? Would that speed up testing (even a little bit)?

line-o:
[Newly added tests are currently failing](https://github.com/eXist-db/exist/runs/5118298188?check_suite_focus=true#step:5:9793) » FileSystemNotFound

line-o:
Codacy complains about the tests not containing an explicit assertion (`assert` or `fail`).
I believe this is a valid complaint and would like to ask to refactor the tests to make it obvious where the assertion is taking place within each case.

extensions/webdav/src/test/resources/standalone-webapp/WEB-INF/web.xml
adamretter: This is not related to the issue or the fix. i.e. this is unchanged - this is just config needed to run eXist-db WebServer in the WebDAV tests. This is used in other tests too.

extensions/webdav/src/test/resources/log4j2.xml
adamretter: This is unlikely to make any difference. It is simpler to have consistent log4j2.xml files, than lots of different ones.

adamretter:
> [Newly added tests are currently failing](https://github.com/eXist-db/exist/runs/5118298188?check_suite_focus=true#step:5:9793) » FileSystemNotFound

@line-o That is now fixed.

adamretter:
> Codacy complains about the tests not containing an explicit assertion (assert or fail).
> I believe this is a valid complaint

I am afraid that Codacy is wrong here. If you check how the tests work, the assertions are common to all test methods and so are in the common function (which is perfectly valid) - Codacy is failing to trace the code correctly.

dizzzz (approved):

line-o:
Renaming the function to include the keyword assert might be enough for codacy - and humans - to pick it up.

PR #4263, opened 2022-03-02
[sec] add run configuration for local jetty server

line-o:
The run configuration to start the jetty server within the IDEA IDE that ships with eXist-db exposes the server to the local network (by binding to 0.0.0.0).
A new configuration named "Local Jetty Server" is added which only binds to the host IP `127.0.0.1` aka localhost.
This was made possible by #4196, which adds recognition of the option `jetty.host`.

It is absolutely unnecessary to expose a development instance running inside an IDE to the internet in most cases.

### DIFF
--- a/.idea/runConfigurations/Local_Jetty_Server.xml
+++ b/.idea/runConfigurations/Local_Jetty_Server.xml
@@ -1,0 +1,13 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Local Jetty Server" type="Application" factoryName="Application">
+    <option name="ALTERNATIVE_JRE_PATH" value="1.8" />
+    <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
+    <option name="MAIN_CLASS_NAME" value="org.exist.start.Main" />
+    <module name="exist-distribution" />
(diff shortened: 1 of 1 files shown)


line-o:
I do not know if that is an option within the IDEA run configuration. But it is possible to overwrite the default by editing it in the extremely rare case one would need that.

adamretter:
Sorry but I am against this. I frequently test across multiple machines when debugging issues.

line-o:
While running the db from inside IDEA? @adamretter

adamretter:
@line-o Indeed.

line-o:
And it is too much to ask to run a proxy for that special use-case?

adamretter:
@line-o For me it is the normal use-case not a special use-case.

line-o:
I see, how about adding a separate runconf for your use-case?

adamretter:
> how about adding a separate runconf for your use-case?

You are assuming this is only my use-case. I suspect others work this way also.

Additionally I don't really understand what the advantage of your PR is over what we currently have?

line-o:
The configuration that runs the jetty server only on localhost is now available as a separate option "Local Jetty Server".
I believe this addresses the concerns of you both @ljo and @adamretter

adamretter:
I think this is better dealt with my the firewall of your system, in-built on macOS, Linux, and Windows for many years now.

However, as long as this is an additional option and not the default option, then it doesn't effect other developers.

PR #4269, opened 2022-03-14
[test] Increase timeout for slower CI VMs to 60s

reinhapa:
Increase CollectionLockTest timeout


joewiz:
The Windows builds still fail, but while the JDK 11 run returned the same "Error: CollectionLocksTest.singleWriter:180->singleWriter:261 singleWriter test likely showed a thread deadlock", the JDK 8 build returned a new one (at least it's new to me):

> CopyCollectionTest.copyToNonExistentAsSelf:88->checkAttributes:404 Created value is not correct
Expected: not <1647360175061L> but: was <1647360175061L>

dizzzz (approved):

PR #4274, opened 2022-03-15
Fix a security privilege escalation issue with XQueryTrigger

adamretter:
This fixes a security issue with XQueryTrigger, and also as a nice consequence enables setUid and setGid to be used on XQuery Triggers that are stored as modules in the database.

### DIFF
--- a/exist-core/src/main/java/org/exist/collections/triggers/XQueryTrigger.java
+++ b/exist-core/src/main/java/org/exist/collections/triggers/XQueryTrigger.java
@@ -41,13 +41,14 @@ import org.exist.source.Source;
 import org.exist.source.SourceFactory;
 import org.exist.source.StringSource;
 import org.exist.storage.DBBroker;
-import org.exist.storage.ProcessMonitor;
 import org.exist.storage.txn.Txn;
 import org.exist.xmldb.XmldbURI;
 import org.exist.xquery.*;
 import org.exist.xquery.value.AnyURIValue;
 import org.exist.xquery.value.Sequence;
 import org.exist.xquery.value.StringValue;
+
+import static com.evolvedbinary.j8fu.tuple.Tuple.Tuple;
 
 /**
  * A trigger that executes a user XQuery statement when invoked.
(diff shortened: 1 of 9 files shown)


exist-core/src/main/java/org/exist/collections/triggers/XQueryTrigger.java
```diff
@@ -71,7 +72,7 @@
 */
 public class XQueryTrigger extends SAXTrigger implements DocumentTrigger, CollectionTrigger {
 
-    protected Logger LOG = LogManager.getLogger(getClass());
+    private static final Logger LOG = LogManager.getLogger(XQueryTrigger.class);
```
dizzzz, line 75: LogManager.getLogger() works as well no?

exist-core/src/main/java/org/exist/collections/triggers/XQueryTrigger.java
```diff
@@ -434,71 +435,58 @@ private CompiledXQuery getScript(boolean isBefore, DBBroker broker, Txn transact
...
-	            call.analyze(new AnalyzeContextInfo());
-
-				final Sequence contextSequence;
-				final ContextItemDeclaration cid = call.getContext().getContextItemDeclartion();
-				if(cid != null) {
-					contextSequence = cid.eval(null);
-				} else {
-					contextSequence = NodeSet.EMPTY_SET;
+            final int nParams;
+            if (dst != null) {
```
dizzzz, line 450: for me it would be ok to wrap this into a oneliner inline :-)

exist-core/src/main/java/org/exist/xquery/XQuery.java
```diff
@@ -226,19 +230,27 @@ private CompiledXQuery compile(final XQueryContext context, final Reader reader,
...
+            if (ast == null) {
                 throw new XPathException("Unknown XQuery parser error: the parser returned an empty syntax tree.");
             }
             
 //            LOG.debug("Generated AST: " + ast.toStringTree());
-            final PathExpr expr = new PathExpr(context);
-            if(xpointer) {
+
+            final PathExpr expr;
+            if (XQueryTreeParser.MODULE_DECL == ast.getType()) {
```
dizzzz, line 240: I'd allow the inline construction...

exist-core/src/main/java/org/exist/collections/triggers/XQueryTrigger.java
adamretter: Yes they are equivalent, the old version requires an unnecessary function call, whereas the second can be done statically by the compiler.

exist-core/src/main/java/org/exist/collections/triggers/XQueryTrigger.java
adamretter: I find for myself that if/else without braces, or with braces but inline is very hard to spot.

dizzzz (approved):

PR #4290, opened 2022-03-25
Fix default encoding and file module tests on Windows

line-o:
### Description:

With this PR applied the file module tests reliably pass. It also sets the correct default encoding `UTF-8` where needed and reinstates the full XQuery3 test suite.

### Reference:

CI tests fail on Windows 

### Type of tests:

XQSuite

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
+++ b/exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
@@ -97,7 +97,7 @@ public class SerializerUtils {
         CDATA_SECTION_ELEMENTS(OutputKeys.CDATA_SECTION_ELEMENTS, Type.QNAME, Cardinality.ZERO_OR_MORE, Sequence.EMPTY_SEQUENCE),
         DOCTYPE_PUBLIC(OutputKeys.DOCTYPE_PUBLIC, Type.STRING, Cardinality.ZERO_OR_ONE, Sequence.EMPTY_SEQUENCE),   //default: () means "absent"
         DOCTYPE_SYSTEM(OutputKeys.DOCTYPE_SYSTEM, Type.STRING, Cardinality.ZERO_OR_ONE, Sequence.EMPTY_SEQUENCE),   //default: () means "absent"
-        ENCODING(OutputKeys.ENCODING, Type.STRING, Cardinality.ZERO_OR_ONE, new StringValue("utf-8")),
+        ENCODING(OutputKeys.ENCODING, Type.STRING, Cardinality.ZERO_OR_ONE, new StringValue("UTF-8")),
         ESCAPE_URI_ATTRIBUTES("escape-uri-attributes", Type.BOOLEAN, Cardinality.ZERO_OR_ONE, BooleanValue.TRUE),
         HTML_VERSION(EXistOutputKeys.HTML_VERSION, Type.DECIMAL, Cardinality.ZERO_OR_ONE, new DecimalValue(5)),
         INCLUDE_CONTENT_TYPE("include-content-type", Type.BOOLEAN, Cardinality.ZERO_OR_ONE, BooleanValue.TRUE),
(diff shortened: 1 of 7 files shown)


adamretter:
@line-o I have pushed a couple of commits that I think help improve this a little.

I am just wondering about the reason for introducing the `%pending` annotations, i.e. are there still issues on Windows that need to be addressed?

line-o:
Well, thanks @adamretter I was just about to do add the path separator myself.  But I am still in the process of understanding the root cause of the changed behaviour on Windows.


Please understand that the added functionality is not in scope for this PR, I am afraid. As they are features they will need documentation. Waiting for that to be merged will keep us from having running tests on windows for too long.
So, please open a new PR with those two valuable commits, that we can then discuss them properly.

Not so sure about the temp directory and also they might be better put into the `system` module.

line-o:
There is some quirkiness involved in the string handling I still can't point to the exact location. Fortunately the problem goes away when comparing binary values.
I think it is best to discuss the best way forward in next weeks community call.
The tests now pass reliably it seems but their readability has suffered in the process.

adamretter:
> As they are features they will need documentation.

They are documented in the function docs. They are also documented in the EXPath spec - http://expath.org/spec/file#sys-props

> Waiting for that to be merged will keep us from having running tests on windows for too long.

Why do we need to wait? - I was happy for them to be merged as part of this PR!

line-o:
All tests green. What a look to behold. :tada:

extensions/modules/file/src/test/resources/util/fixtures.xqm
```diff
@@ -23,17 +23,85 @@ xquery version "3.1";
...
+            <nested
+                attr = "with lots of whitespace"
+            />
+            The next word <hi>is</hi> highlighted.
+        </item>
+        <p/>
+    </root>
+;
+
+(: FIXME(JL) cannot use StringConstructor here because that will cause all comparisons to fail on Windows :)
```
adamretter, line 57: Could you open a GitHub issue for this and then place the link here so that we link the two things please?

line-o:
This rather cosmetic thing is keeping our builds to fail. In my opinion no other PR should be merged before this one is applied.

line-o:
So, here you go - with a link to the new issue.

adamretter:
p.s. I think we need to merge this one first - https://github.com/eXist-db/exist/pull/4302
and then rebase your PR to then hopefully see all tests pass :-)

extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
```diff
@@ -122,6 +122,7 @@
         DEFAULT_PROPERTIES.put(OutputKeys.INDENT, "yes");
         DEFAULT_PROPERTIES.put(OutputKeys.OMIT_XML_DECLARATION, "no");
         DEFAULT_PROPERTIES.put(EXistOutputKeys.EXPAND_XINCLUDES, "no");
+        DEFAULT_PROPERTIES.put(OutputKeys.ENCODING, "UTF-8");
```
dizzzz, line 125: https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html#UTF_8 ?

exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
```diff
@@ -97,7 +97,7 @@
         CDATA_SECTION_ELEMENTS(OutputKeys.CDATA_SECTION_ELEMENTS, Type.QNAME, Cardinality.ZERO_OR_MORE, Sequence.EMPTY_SEQUENCE),
         DOCTYPE_PUBLIC(OutputKeys.DOCTYPE_PUBLIC, Type.STRING, Cardinality.ZERO_OR_ONE, Sequence.EMPTY_SEQUENCE),   //default: () means "absent"
         DOCTYPE_SYSTEM(OutputKeys.DOCTYPE_SYSTEM, Type.STRING, Cardinality.ZERO_OR_ONE, Sequence.EMPTY_SEQUENCE),   //default: () means "absent"
-        ENCODING(OutputKeys.ENCODING, Type.STRING, Cardinality.ZERO_OR_ONE, new StringValue("utf-8")),
+        ENCODING(OutputKeys.ENCODING, Type.STRING, Cardinality.ZERO_OR_ONE, new StringValue("UTF-8")),
```
dizzzz, line 100: shouldn't we use https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html#UTF_8 ?

dizzzz (approved):

dizzzz:
image
wohoooooooo!

exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
adamretter: Yes, `UTF8.name()` is preferred when you need the string name.

PR #4294, opened 2022-03-30
Optimise URI Encoding of Path Components

adamretter:
This optimises the URI Encoding of Path Components in eXist-db, and specifically affects `fn:encode-for-uri`.

Through using the YourKit Java Profiler, we found that a great deal of time in a specific XQuery was unexpectedly taken up by the `fn:encode-for-uri` XQuery function. A quick examination of the code found it to be correct, yet definetly sub-optimal.

By adding JMH Tests for this in `URIUtilsBenchmark.java` we are able to confidently show that before this PR, the throughput of `URIUtils#encodeForURI(String)` (which is used by `fn:encode-for-uri`) was:

```
293623.937 ± 16251.838  ops/s
```

This PR increases the throughput to:
```
2056617.057 ± 56774.934  ops/s
``` 

Overall performance is increased by a factor of `~7x` (or `~600%` if you prefer)!

Such techniques could be applied to other URI/URL/IRI encoding/decoding methods within eXist-db in future if desired.

### DIFF
--- a/exist-core/src/test/java/org/exist/xquery/util/URIUtilsTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/util/URIUtilsTest.java
@@ -1,0 +1,195 @@
+/*
+ * Copyright (C) 2014, Evolved Binary Ltd
+ *
+ * This file was originally ported from FusionDB to eXist-db by
+ * Evolved Binary, for the benefit of the eXist-db Open Source community.
+ * Only the ported code as it appears in this file, at the time that
+ * it was contributed to eXist-db, was re-licensed under The GNU
+ * Lesser General Public License v2.1 only for use in eXist-db.
+ *
+ * This license grant applies only to a snapshot of the code as it
+ * appeared when ported, it does not offer or infer any rights to either
+ * updates of this source code or access to the original source code.
+ *
+ * The GNU Lesser General Public License v2.1 only license follows.
+ *
+ * ---------------------------------------------------------------------
+ *
+ * Copyright (C) 2014, Evolved Binary Ltd
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; version 2.1.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


exist-core/src/test/java/org/exist/xquery/util/URIUtilsTest.java
```diff
@@ -0,0 +1,195 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.util;
+
+import org.junit.Test;
```
reinhapa, line 35: I would write new JUnit tests in JUnit 5 (org.junit.jupiter.api.Test)

reinhapa (changes requested):
Would it a big issue to convert the new test into JUnit 5?

exist-core/src/test/java/org/exist/xquery/util/URIUtilsTest.java
adamretter: @reinhapa I can try :-) I hadn't in the past as I wasn't sure if the Junit 4 and Junit 4 tests would report the results side-by-side to Surefire. Let's see...

adamretter: Done :-)

adamretter:
@reinhapa I pushed an additional commit to do that.

dizzzz (approved):

dizzzz (approved):

exist-core/src/test/java/org/exist/xquery/util/URIUtilsTest.java
```diff
@@ -0,0 +1,195 @@
...
+package org.exist.xquery.util;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * @author <a href="mailto:adam@evolvedbinary.com">Adam Retter</a>
+ */
+public class URIUtilsTest {
```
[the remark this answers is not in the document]
adamretter: Interesting! I have now pushed an additional commit to do this.

reinhapa (approved):

Issue #4301, opened 2022-04-01
[BUG] Compare StringConstructor literal with xs:string on Windows 


line-o:
**Describe the bug**

Comparison of a StringConstructor literal with a xs:string fails on Windows while both values are equal on Mac and linux.

This was uncovered while working on #4290. There, the contents of a file was read from the file system. But I believe this
should be visible by just comparing the string itself.

**Expected behavior**

The comparison to work regardless of the operatig system in use.

**To Reproduce**

```Xquery
xquery version "3.1";

module namespace t="http://exist-db.org/xquery/test";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare variable $t:XML := 
<root>
    <nested/>
</root>
;

declare variable $t:SC := 
``[<root>
    <nested/>
</root>]``
;

declare
    %test:assertTrue
function t:test() {
    serialize($t:XML, map{"indent": true()}) eq $t:SC
};
```

**Context (please always complete the following information):**
 - OS: windows runner on GHA
 - eXist-db version: 6.1.0-SNAPSHOT
 - Java Version: 1.8.0_322 (Temurin)

**Additional context**
- How is eXist-db installed? built from source 77875bed96f07930cced8ce41408ed22f208e301
- Any custom changes in e.g. `conf.xml`? none

line-o:
As evidence you can see that the newly added test in #4303 only fails on windows

PR #4303, opened 2022-04-02
[wip] Test string constructor with string comparison on windows

line-o:
### Description:

added test to show that string constructor literal is not comparable to xs:string on windows 

### Reference:

refs #4301 

### Type of tests:

XQSuite


line-o:
With the new test setup this will need to be changed into an integration test.

line-o:
@line-o Note to self: retest

PR #4311, opened 2022-04-04
Replace utf8 literals

line-o:
refs #4290 

@dizzzz and @adamretter proposed to use `UTF_8.name()` instead of `"UTF-8"`. 

It turned out to be all over the codebase, that is why I thought it is better to be part of a separate PR.


line-o:
If merged, supersedes #4290

dizzzz (approved):

adamretter:
> It turned out to be all over the codebase, that is why I thought it is better to be part of a separate PR.

We were only suggesting within your changes themselves... Cool that you decided to fix all the others too tough :-)

PR #4315, opened 2022-04-05
Drop build workflow

line-o:
The build workflow does not add value anymore now that tests reliably pass. Well, for the most part :)
But since we are now able to just re-run a single failed job it is the better strategy than to duplicate builds (in build and test).

We can also expect slightly faster startup times of tests, since less workers need to be allocated each run.


duncdrum (approved):
Lgtm

PR #4316, opened 2022-04-05
Use builtin maven-cache of setup-java action

line-o:
The caching key is built hashing all pomfiles (as fileHash).
That is in line with the current strategy.

- was : `${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}`
- is  : `setup-java-${{ platform }}-${{ packageManager }}-${{ fileHash }}`

Documentation: https://github.com/actions/setup-java#caching-packages-dependencies


duncdrum (approved):
2nd run found the cache

PR #4322, opened 2022-04-08
[bugfix] file:sync#3 updates only changed files

### LINKED ISSUE #4321: [BUG] file:sync() returns <update/> elements for all files synced
**Describe the bug**
`file:sync()` returns `<update/>` elements for all files synced.

**Expected behavior**
Previously (5.3.0) it would return `<update/>` elements only for files that had changed. 
My code evaluates this output to determine if there were changes and then do subsequent tasks.

**To Reproduce**
Call `file:sync('/db/file-module-test','/home/existdb/exist-sync/test-sync',())` twice and the result will be the same even though no updates have happened in the meantime.

line-o:
### Description:

- Sync.saveFile checks for modification of resource in a more readable and
  safer way while also returning even earlier than before
- add test that runs file:sync twice and ensures nothing is written to disk
- add test with multiple exclude patterns
- fix expectation in tests that were false positives before,
  due to the issue in helper:assert-sync-result
- fix and enhance helper:assert-sync-result to
  - throw when empty sequence is expected but something was returned
  - add label to each test so that it is clear where the error occurred
- helper:create-db-resource does not set modified time (is unnecessary)
- $fixtures:mod-date is now equal to current time

### Reference:

Fixes #4321

### Type of tests:

XQSuite

### DIFF
--- a/extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
+++ b/extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java
@@ -371,27 +371,30 @@ public class Sync extends BasicFunction {
     }
 
     private void saveFile(final Path targetFile, final DocumentImpl doc, final Date startDate, final MemTreeBuilder output) throws LockException {
+        // the resource has not changed in the selected period
+        if (startDate != null && doc.getLastModified() <= startDate.getTime()) {
+            return;
+        }
         try (final ManagedLock<MultiLock[]> lock = context.getBroker().getBrokerPool().getLockManager().acquireDocumentReadLock(doc.getURI())) {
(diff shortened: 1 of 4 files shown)


line-o:
5.x.x Backport is part of #4324

dizzzz (approved):

adamretter:
@line-o I am going to merge this. In future when creating a backport, can you do so as a 1-1 mapping between PRs please. i.e. if you want to backport a PR then create a single backport PR which contains only those changes and uses the same commit messages. Please do NOT mix multiple original PRs into a single backport PR.

This helps when we look back at the histories, so that we can clearly and easily compare the two branches and understand what was backported from where. Thanks

PR #4333, opened 2022-04-12
Run XQTS against eXist-db in CI

adamretter:
This will run the `exist-xqts-runner` against eXist-db in CI, and attach the XQTS results as an artifact of the build.

### DIFF
--- a/.github/workflows/ci-xqts.yml
+++ b/.github/workflows/ci-xqts.yml
@@ -1,0 +1,26 @@
+name: XQTS
+on: [push, pull_request]
+jobs:
+  xqts:
+    name: W3C XQuery Test Suite
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 8
+        uses: actions/setup-java@v2
+        with:
+          distribution: temurin
+          java-version: '8'
+          cache: 'maven'
+      - name: Maven Build
+        timeout-minutes: 60
+        run: mvn -V -B clean package -DskipTests -Ddependency-check.skip=true -Ddocker=false -P !mac-dmg-on-unix,!installer,!concurrency-stress-tests,!micro-benchmarks,!build-dist-archives
+      - name: Run XQTS
+        run: find exist-xqts/target -name exist-xqts-runner.sh -exec {} --xqts-version HEAD --output-dir /tmp/xqts-output \;
+      - name: Archive XQTS Logs
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: xqts-logs
+          retention-days: 14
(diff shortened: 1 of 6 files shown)


dizzzz (commented):
There is again this license thinghy

dizzzz (commented):
image

line-o:
I would like to have this as an activatable action. I fear running this on every build will prolong the build unnecessarily.

line-o:
Here is the documentation how to setup a GitHub action that is triggered manually 
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

joewiz:
@line-o Neat idea, but if the current CI is any indication, it doesn't seem the XQTS action adds much time to the build:

> Successful in 6m

On the other hand, if we trigger it manually, wouldn't that open us up to inadvertent regressions in XQTS? Divergence of eXist from XQuery 3.1 specs has been a persistent problem for eXist. One might argue that we should keep this run enabled by default until we are confident that we're conforming better. 

Looking at this from another angle, why shouldn't all of our actions be activatable pending some assertion by the author or reviewer about what the nature of a changeset?

adamretter:
I think this is better setup as an automated job for the reasons that Joe outlined. Also the execution is quite fast so far, faster than running the JUnit and XQSuite tests - so I don't see this adding any delay.

line-o:
When I read the GitHub action setup correctly then XQTS is _not_ built against the current build of Existdb. From all that I know I had to rebuild the xqts runner each time I wanted to run it against a new build.
The steps to get to comparable results were:

- mvn install in existdb working directory
- switch to xqts runner working directory
- build xqts
- run tests

Is this no longer necessary, have there been changes in xqts runner I am not aware of?

dizzzz (approved):

dizzzz:
@line-o is the PR good to go? I am not sure how to read your comment....

line-o:
@dizzzz I do not think this is ready since the line you screens hotted earlier is still in the run. So nothing is retained.

line-o:
@joewiz about having all the checks as activatable: I am all for it, let's discuss this on the community call.

adamretter:
@dizzzz This is now correctly retaining the results from the XQTS run as an artifact of the build.

line-o:
But is there a plan for every contributor to have easy access to the results of a single run?
The least thing I would expect is to have some actionable log output in the job itself.

adamretter:
> But is there a plan for every contributor to have easy access to the results of a single run?

They can easily click the link on the CI test run page to download the results.

> The least thing I would expect is to have some actionable log output in the job itself.

That could perhaps be added in future, but as it is not a feature currently provided by the system and tools available, it is out of scope for this PR.

PR #4335, opened 2022-04-14
Log xsl:fo render errors to exist.log

adamretter:
Without this it can be very hard to diagnose when your XSL:FO processor is misconfigured.


joewiz:
I remember well the pain of debugging XSL-FO from eXist. I always advised to use fop on the commandline first, then move a working configuration to eXist for testing. It sounds like this could help avoid that need!

dizzzz (approved):

PR #4351, opened 2022-04-24
[4.x.x] Improvements to system:as-user

adamretter:
**NOTE** This should not yet be merged as it shows an unexpected issue in 4.x.x that needs further investigation

Backport of https://github.com/eXist-db/exist/pull/4349


line-o:
Propose to close since exist-db v4 is EOL and no longer maintained

PR #4360, opened 2022-04-25
Implement fn:default-language#0

alanpaxton:
### Description:

This is an implementation of the XQuery 3 default-language function missing from eXist-db.

### Reference:

This has been reported as part of https://github.com/eXist-db/exist/issues/3635
The function is defined in https://www.w3.org/TR/xpath-functions-31/#func-default-language

### Type of tests:

A fnLanguage.xqm test is provided which confirms that a default language is returned and that it is one of the known languages. The language is fetched in Java from the underlying default Locale, and will naturally vary depending where the tests are run.

A defaultLanguage() test is added to XQueryFunctionsTest.java which confirms that running the method in the embedded server returns the same language as directly accessing the default locale through Java.


exist-core/src/main/java/org/exist/xquery/functions/fn/FnDefaultLanguage.java
```diff
@@ -0,0 +1,66 @@
...
+                    "during the evaluation of a query or transformation in which " +
+                    "fn:default-language() is executed.",
+            FunctionDSL.returns(Type.LANGUAGE, Cardinality.EXACTLY_ONE, "the default language within query execution time span"));
+
+    public FnDefaultLanguage(final XQueryContext context) {
+        super(context, FS_DEFAULT_LANGUAGE);
+    }
+
+    public Sequence eval(final Sequence[] args, final Sequence contextSequence) throws XPathException {
+        if (context.getProfiler().isEnabled()) {
```
dizzzz, line 49: basic function already implements this profiler

dizzzz (approved):

dizzzz (changes requested):
LGTM but the profiler check is not needed...

exist-core/src/main/java/org/exist/xquery/functions/fn/FnDefaultLanguage.java
adamretter: Ah right! So we could remove this from our `eval`. Also I just noticed, we should add an `@Override` to our eval

dizzzz (approved):

joewiz:
@alanpaxton For good measure, could you please post the results of relevant XQTS tests for this function? Thanks!

PR #4372, opened 2022-05-02
Improve XSLT errors

adamretter:
Previously errors that could be raised in XSLT transformations (by Saxon) and appear in eXist-db's log file did not include information about where the error occurred. This PR simply includes the location information provided by the `TransformerException`.

### DIFF
--- a/exist-core/src/main/java/org/exist/xslt/XSLTErrorsListener.java
+++ b/exist-core/src/main/java/org/exist/xslt/XSLTErrorsListener.java
@@ -21,6 +21,7 @@
  */
 package org.exist.xslt;
 
+import javax.annotation.Nullable;
 import javax.xml.transform.ErrorListener;
 import javax.xml.transform.TransformerException;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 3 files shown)


dizzzz (approved):

exist-core/src/main/java/org/exist/xslt/XSLTErrorsListener.java
```diff
@@ -31,64 +32,70 @@
...
       case ERROR:
         if (stopOnError) {
-          raiseError("XSL transform reported error: " + exception.getMessage(), exception);
+          raiseError("XSL transform reported error: " + exception.getMessageAndLocation(), exception);
         }
         break;
       case FATAL:
-        raiseError("XSL transform reported error: " + exception.getMessage(), exception);
+        raiseError("XSL transform reported error: " + exception.getMessageAndLocation(), exception);
     }
```
reinhapa, line 72: Could you add the missing break here?

reinhapa (changes requested):

exist-core/src/main/java/org/exist/xslt/XSLTErrorsListener.java
adamretter: @reinhapa I would like to keep this PR the same as the other two that were merged if possible. The missing break there will not actually cause any harm.

dizzzz (approved):

reinhapa (approved):

PR #4375, opened 2022-05-02
Do not override Bearer Authentication with Basic Authentication

adamretter:
Previously Bearer Authentication was incorrectly intercepted by the `AbstractExistHttpServlet` and/or XQuery Controller (`XQueryURLRewrite`), which made custom handling of Bearer Authentication (e.g. OAuth, JWT, etc.) impossible from XQuery.

It is now possible to work with Bearer Authentication tokens directly in XQuery, so you may implement your own authentication schemes in XQuery if you wish.

### DIFF
--- a/exist-core/src/main/java/org/exist/http/servlets/AbstractExistHttpServlet.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/AbstractExistHttpServlet.java
@@ -246,9 +246,9 @@ public abstract class AbstractExistHttpServlet extends HttpServlet {
 	        }
         }
 
-        // Secondly try basic authentication
+        // Secondly try basic authentication if there is no Authorization header or the Authorization header does not indicate Basic auth
         final String auth = request.getHeader("Authorization");
-        if (auth == null && getDefaultUser() != null) {
+        if ((auth == null || !auth.startsWith("Basic ")) && getDefaultUser() != null) {
             return getDefaultUser();
         }
         return getAuthenticator().authenticate(request, response, true);
--- a/exist-core/src/main/java/org/exist/http/servlets/AbstractExistHttpServlet.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/AbstractExistHttpServlet.java
@@ -248,7 +248,7 @@ public abstract class AbstractExistHttpServlet extends HttpServlet {
 
         // Secondly try basic authentication if there is no Authorization header or the Authorization header does not indicate Basic auth
         final String auth = request.getHeader("Authorization");
-        if ((auth == null || !auth.startsWith("Basic ")) && getDefaultUser() != null) {
+        if ((auth == null || !auth.toLowerCase().startsWith("basic ")) && getDefaultUser() != null) {
             return getDefaultUser();
         }
         return getAuthenticator().authenticate(request, response, true);
--- a/exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
+++ b/exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
@@ -179,7 +179,7 @@ public class XQueryURLRewrite extends HttpServlet {
         } else {
             // Secondly try basic authentication
             final String auth = request.getHeader("Authorization");
-            if (auth != null && auth.startsWith("Basic ")) {
+            if (auth != null && auth.toLowerCase().startsWith("basic ")) {
                 requestUser = authenticator.authenticate(request, response, sendChallenge);
                 if (requestUser != null) {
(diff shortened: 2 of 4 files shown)


dizzzz (changes requested):

exist-core/src/main/java/org/exist/http/servlets/AbstractExistHttpServlet.java
```diff
@@ -246,9 +246,9 @@ protected Subject authenticate(HttpServletRequest request, HttpServletResponse r
 	        }
         }
 
-        // Secondly try basic authentication
+        // Secondly try basic authentication if there is no Authorization header or the Authorization header does not indicate Basic auth
         final String auth = request.getHeader("Authorization");
-        if (auth == null && getDefaultUser() != null) {
+        if ((auth == null || !auth.startsWith("Basic ")) && getDefaultUser() != null) {
```
reinhapa, line 251: the auth comperation should be case-insensitive to be specification compliant by writing it somewhat like this:
`!auth.toLowerCase().startsWith("basic ")`

exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
```diff
@@ -179,7 +179,7 @@ protected void service(final HttpServletRequest request, final HttpServletRespon
         } else {
             // Secondly try basic authentication
             final String auth = request.getHeader("Authorization");
-            if (auth != null) {
+            if (auth != null && auth.startsWith("Basic ")) {
```
reinhapa, line 182: Same as before

exist-core/src/test/java/org/exist/http/RESTServiceTest.java
```diff
@@ -692,6 +705,308 @@ public void storedQuery() throws IOException {
...
+                    "<authorization>\n" +
+                            "    <sm:id xmlns:sm=\"http://exist-db.org/xquery/securitymanager\">\n" +
+                            "        <sm:real>\n" +
+                            "            <sm:username>admin</sm:username>\n" +
+                            "            <sm:groups>\n" +
+                            "                <sm:group>dba</sm:group>\n" +
+                            "            </sm:groups>\n" +
+                            "        </sm:real>\n" +
+                            "    </sm:id>\n" +
+                            "    <header>Basic YWRtaW46</header>\n" +
```
reinhapa, line 805: Here it would make sense to change the case of the header to something like `BaSiC` to make sure that the comparison is done specification compliant.

exist-core/src/main/java/org/exist/http/servlets/AbstractExistHttpServlet.java
```diff
@@ -246,9 +246,9 @@ protected Subject authenticate(HttpServletRequest request, HttpServletResponse r
 	        }
         }
 
-        // Secondly try basic authentication
+        // Secondly try basic authentication if there is no Authorization header or the Authorization header does not indicate Basic auth
         final String auth = request.getHeader("Authorization");
-        if (auth == null && getDefaultUser() != null) {
```
reinhapa, line 251: Would it make sense to extract the `getDefaultUser()` value to a local variable?

reinhapa (changes requested):

exist-core/src/main/java/org/exist/http/servlets/AbstractExistHttpServlet.java
adamretter: Good spot @reinhapa! I have now added the case-insensitive match

exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
adamretter: Good spot @reinhapa! I have now added the case-insensitive match

exist-core/src/test/java/org/exist/http/RESTServiceTest.java
adamretter: @reinhapa I have now added a test specifically for that: `RESTServiceTest#execQueryWithBasicAuthCaseInsensitive()`

exist-core/src/main/java/org/exist/http/servlets/AbstractExistHttpServlet.java
adamretter: I think as there are only 2 calls and the underlying function implementation just accesses a known class member it will make little difference... also I guess the compiler will do some optimisation anyway?

dizzzz (approved):

reinhapa (approved):

PR #4376, opened 2022-05-02
[5.x.x] Do not override Bearer Authentication with Basic Authentication

adamretter:
Backport of https://github.com/eXist-db/exist/pull/4375


dizzzz (changes requested):

adamretter:
@dizzzz Indeed you are correct, I have updated all 3-PR's to fix that now.

PR #4377, opened 2022-05-02
[4.x.x] Do not override Bearer Authentication with Basic Authentication

adamretter:
Backport of https://github.com/eXist-db/exist/pull/4375

### DIFF
--- a/exist-core/src/main/java/org/exist/http/servlets/BasicAuthenticator.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/BasicAuthenticator.java
@@ -69,9 +69,9 @@ public class BasicAuthenticator implements Authenticator {
 		String password = null;
 
 		try {
-			if (credentials != null && credentials.startsWith("Basic")) {
+			if (credentials != null && credentials.toLowerCase().startsWith("basic ")) {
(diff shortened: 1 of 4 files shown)


dizzzz (changes requested):

adamretter:
@dizzzz Thanks :-) Indeed... you are correct, I have updated all 3-PR's to fix that now.

exist-core/src/main/java/org/exist/http/servlets/BasicAuthenticator.java
```diff
@@ -67,21 +67,22 @@ public Subject authenticate(
...
-                    }
-                } catch(final IllegalArgumentException iae) {
-                    LOG.warn("Invalid BASIC authentication header received: " + iae.getMessage(), iae);
-                    credentials = null;
-                }
+
+		try {
+			if (credentials != null && credentials.toLowerCase().startsWith("basic ")) {
+				final Base64Decoder dec = new Base64Decoder();
+				dec.translate(credentials.substring("basic ".length()));
```
dizzzz, line 74: can still contain Basic ?

dizzzz (approved):

PR #4379, opened 2022-05-03
[bugfix] override resource type from backup metadata

### LINKED ISSUE #4368: [BUG] Some files are not restored from backup starting with version 5.4.0 up to the latest
**Describe the bug**

From a complete backup from the same or an earlier version of eXist, some files are not restored.  They seem to be parsed as XML, although they were stored and marked as "BinaryRessource" during the backup process.  The problem shows with certain types of files, including HTML5.  The attached backup file (created with version 5.4.0) demonstrates the problem - [eXist-backup.zip](https://github.com/eXist-db/exist/files/8600979/eXist-backup.zip)

This problem has been discussed in the Slack channel: https://app.slack.com/client/TG47XS7E2/CG2MRUZ35/thread/CG2MRUZ35-1651390531.411839

**Expected behavior**

I expect these files to be processed in a way that makes them function identical to the situation before the backup/restore process

**To Reproduce**

1. copy the attached backup file into  $PWD/backup
2. Run the following sequence of commands shows the problem:
```
#!/bin/sh
NAM=res-$(date +%Y-%m-%d-%H-%M)
#IMG=existdb/existdb:latest
IMG=existdb/existdb:5.4.0
CWD=$(pwd)
BKF=$(ls $CWD/backup/)
docker run -it -d -p 8080:8080 -p 8443:8443 -v $CWD/backup:/opt/backup --name $NAM $IMG
#docker run -it -d -p 8080:8080 -p 8443:8443 -v $CWD/etc:/exist/etc -v $CWD/backup:/opt/backup --name $NAM $IMG
sleep 60s
docker exec $NAM java org.exist.start.Main client --no-gui -u admin  --xpath "system:restore('/opt/backup/$BKF', '', '')"
```
4. The backup contains a collection with one file "/db/temp/dependencies.html"  (this file is part of the dashboard, it is here used to demonstrate the problem).  With the image variable above set to 5.3.1 the backup is successful, but with 5.4.0 and later it fails with the error message:
```
<system:restore xmlns:system="http://exist-db.org/xquery/system">
<system:info>Starting restore of backup...</system:info>
<system:info>Processing backup: /opt/backup/eXist-backup.zip#db/temp/__contents__.xml</system:info>
<system:collection>/db/temp</system:collection><system:warn>Failed to restore resource 'dependencies.html'
from file '/opt/backup/eXist-backup.zip#db/temp/dependencies.html'.
Reason: The XML parser reported a problem: fatal error at (1,2045) : The element type "link" must be terminated by the matching end-tag "&lt;/link&gt;".</system:warn>
<system:info>Finished restore of backup.</system:info>
</system:restore>

```

 - OS: the above docker is running on Ubuntu 20.04, but the problem has also be verified with 5.4.0 running directly on Ubuntu 20.04
 - eXist-db version: tested various versions as they are contained in the docker images: 5.3.1, 5.4.0, 6.0.0 and "latest" 
 - Java Version:   Java 1.8.0_275 [Oracle Corporation (OpenJDK 64-Bit Server VM) in /usr/lib/jvm/java-8-openjdk-amd64/jre

**Additional context**
- How is eXist-db installed? docker image
- Any custom changes in e.g. `conf.xml`?  no changes

reinhapa:
### Description:
Fixes #4368 by using the resource type defined by the backup meta data if it differs the one defined on the mime type definition.

The current implementation does always calculate the resource type (`XMLResource`/`BinaryResource`) based on the bakcup meta data content type using the `MimeTable` instance. In the failing case the resulting resource type will be set to `XMLResource` using the currently defined type for `text/html` within the `mime-types.xml` instead of the resource type `BinaryResource` that is defined in the backup meta data. This leads to the fact that the restore task will change the resource type to a different one as has being wrote to the backup.

### DIFF
--- a/exist-core/src/main/java/org/exist/backup/restore/RestoreHandler.java
+++ b/exist-core/src/main/java/org/exist/backup/restore/RestoreHandler.java
@@ -304,11 +304,16 @@ public class RestoreHandler extends DefaultHandler {
         }
 
         MimeType mimeType = null;
-        if (mimeTypeStr != null) {
+        if (xmlType) {
+            mimeType = MimeType.XML_TYPE;
+        } else if (mimeTypeStr != null) {
             mimeType = MimeTable.getInstance().getContentType(mimeTypeStr);
-        }
-        if (mimeType == null) {
-            mimeType = xmlType ? MimeType.XML_TYPE : MimeType.BINARY_TYPE;
+            if (mimeType == null) {
+                mimeType = MimeType.BINARY_TYPE;
+            } else if (mimeType.isXMLType() !=  xmlType) {
+                listener.warn("Calculated mime type " + mimeType + " type differs from the one specified in the backup, adjusting it");
+                mimeType = new MimeType(mimeType.getName(), xmlType ? MimeType.XML : MimeType.BINARY);
+            }
         }
 
         Date dateCreated = null;
--- a/exist-core/src/test/java/org/exist/backup/XMLDBRestoreTest.java
+++ b/exist-core/src/test/java/org/exist/backup/XMLDBRestoreTest.java
@@ -28,11 +28,9 @@ import org.exist.security.Account;
 import org.exist.security.MessageDigester;
 import org.exist.security.SecurityManager;
 import org.exist.test.ExistWebServer;
+import org.exist.util.MimeTable;
 import org.exist.xmldb.*;
-import org.junit.ClassRule;
-import org.junit.Ignore;
-import org.junit.Rule;
-import org.junit.Test;
+import org.junit.*;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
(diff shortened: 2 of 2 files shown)


dizzzz:
why are these "reduced" pom files checked-in? they are derived no?

exist-core/src/main/java/org/exist/backup/restore/RestoreHandler.java
```diff
@@ -304,11 +304,16 @@ private DeferredPermission restoreResourceEntry(final Attributes atts) throws SA
...
-        if (mimeTypeStr != null) {
+        if (xmlType) {
+            mimeType = MimeType.XML_TYPE;
+        } else if (mimeTypeStr != null) {
             mimeType = MimeTable.getInstance().getContentType(mimeTypeStr);
-        }
-        if (mimeType == null) {
-            mimeType = xmlType ? MimeType.XML_TYPE : MimeType.BINARY_TYPE;
+            if (mimeType == null) {
+                mimeType = MimeType.BINARY_TYPE;
```
dizzzz, line 312: this value should probably be taken from the default in mimetypes.xml ?

reinhapa: No, as we want to have the resource type as defined within the backup meta data to take precedence over those within the mime type configuration.

dizzzz (approved):

PR #4386, opened 2022-05-12
Implement fn:transform

adamretter:
Implements the `fn:transform` function [specification](https://www.w3.org/TR/xpath-functions-31/#func-transform)

The transform function is complex, so we have chosen to structure it as a number of top level classes in their own
package, `org.exist.xquery.functions.fn.transform` which are invoked by the main `org.exist.xquery.functions.fn.FnTransform` class.

The implementation relies on the [Saxon XSLT processing library](https://www.saxonica.com/products/products.xml) to compile referenced stylesheet(s).

In highly simplified terms, the implementation:
* Reads and checks all the parameters/options to the function
* Invokes Saxon to compile the identified stylesheet
* Retrieves a transformer from the compilation
* Figures out which form of application of the stylesheet is required
* Constructs (a) delivery object(s) to build output according to the requested delivery format
* Calls Saxon again (the transformer this time) applying the stylesheet to the input
* Builds the result map using the contents of the delivery objects, and returns it

# Unit Tests
A number of unit tests have been added, driven by the an attempt to test the same functionality as some of the individual XQTS tests efficiently during development. These tests do not provide complete coverage of the features
of `fn:transform`, and the XQTS tests should be considered as the truth.

# XQTS Tests
The goal of the development of `fn:transform` has been to pass all the tests in the `fn-transform` test set of [XQTS tests](https://www.w3.org/XML/xquery/test-suite/) for transformation.

The tests are run with the [eXist XQTS Runner](https://github.com/eXist-db/exist-xqts-runner)

* The test `fn-transform-68` currently fails. We are satisfied that the behaviour exhibited by our
code is correct, and that the test is deficient [Issue](https://github.com/w3c/qt3tests/issues/44)

* All other tests from the `fn:transform` test set pass, except for 4 tests which are not run because they require features that eXist-db does not support (e.g. custom schema types in XQuery) .

----


brihaye:
Be careful, you copied the W3C's "FOXT0001" error code too many times, I'm afraid ;-)

adamretter:
Thanks @brihaye I have fixed that now.

exist-core/src/main/java/org/exist/xquery/functions/fn/FnTransform.java
```diff
@@ -244,7 +256,13 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
     private MapType makeResultMap(final float xsltVersion, final MapType options, final NodeValue outputDocument) throws XPathException {
 
         final MapType outputMap = new MapType(context);
-        final StringValue outputKey = FnTransform.BASE_OUTPUT_URI.get(xsltVersion, options).orElse(new StringValue("output"));
+        final AtomicValue outputKey;
+        final Optional<AnyURIValue> baseOutputURI = FnTransform.BASE_OUTPUT_URI.get(xsltVersion, options);
+        if (baseOutputURI.isPresent()) {
```
dizzzz, line 261: probably better to use  `Optional.orElse()` ?

exist-core/src/main/java/org/exist/xquery/functions/fn/FnTransform.java
```diff
@@ -396,7 +447,7 @@ private float staxExtractXsltVersion(final Source xsltStylesheet) throws XPathEx
     private static final Option<Item> INITIAL_MATCH_SELECTION = new Option<>(
             Type.ITEM,"initial-match-selection", v3_0);
     private static final Option<QNameValue> INITIAL_MODE = new Option<>(
-            Type.QNAME,"initial-match-selection", v1_0, v2_0, v3_0);
+            Type.QNAME,"initial-mode", v1_0, v2_0, v3_0);
```
dizzzz, line 450: there are many quality reports on ordering; is it possible to correct them? 
image

exist-core/src/test/java/xquery/xquery3/XQuery3Tests.java
```diff
@@ -26,7 +26,8 @@
 
 @RunWith(XSuite.class)
 @XSuite.XSuiteFiles({
-        "src/test/xquery/xquery3/fnTransform11.xqm"
+        //"src/test/xquery/xquery3/fnTransform11.xqm",
```
dizzzz, line 29: commented line, please remove? or is this intended?

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3446,4 +3446,112 @@ public HttpContext setSession(final SessionWrapper newSession) {
             return new HttpContext(request, response, newSession);
         }
     }
+
+    @Override public String toString() {
+        return getStringValue();
+    }
+
+    public String getStringValue() {
```
dizzzz, line 3454: nice!

joewiz:
@alanpaxton Thank you for this PR and the truly Herculean effort that went into implementing this complex function!

exist-core/src/test/xquery/xquery3/transform/fnTransformErr16.xqm
```diff
@@ -0,0 +1,52 @@
...
+ : MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ : Lesser General Public License for more details.
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+module namespace testTransform="http://exist-db.org/xquery/test/function_transform";
```
dizzzz, line 24: reuse of namespace : in case of issues once it was very hard to trace back which xqm file was causing it, esp. when there was a syntax error or so. if the reporting works, great!

exist-core/src/test/java/xquery/xquery3/XQuery3Tests.java
```diff
@@ -26,7 +26,9 @@
 
 @RunWith(XSuite.class)
 @XSuite.XSuiteFiles({
-        "src/test/xquery/xquery3"
+        "src/test/xquery/xquery3",
+        "src/test/xquery/xquery3/transform",
+        //"src/test/xquery/xquery3/transform/fnTransform13a.xqm",
```
dizzzz, line 31: should be removed or for good reasons?

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
```diff
@@ -0,0 +1,662 @@
...
+
+    final Tuple2<String, Source> xsltSource;
+    final MapType stylesheetParams;
+    final Map<net.sf.saxon.s9api.QName, XdmValue> staticParams;
+    final float xsltVersion;
+    final Optional<AnyURIValue> resolvedStylesheetBaseURI;
+    final Optional<QNameValue> initialFunction;
+    final Optional<ArrayType> functionParams;
+    final Map<net.sf.saxon.s9api.QName, XdmValue> templateParams;
+    final Map<net.sf.saxon.s9api.QName, XdmValue> tunnelParams;
```
dizzzz, line 113: quality: Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes.

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Transform.java
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    public Sequence eval(final Sequence[] args, final Sequence contextSequence) throws XPathException {
+
+        final Options options = new Options(context, fnTransform, (MapType) args[0].itemAt(0));
+
+        //TODO(AR) Saxon recommends to use a <code>StreamSource</code> or <code>SAXSource</code> instead of DOMSource for performance
+        final Optional<Source> sourceNode = Transform.getSourceNode(options.sourceNode, context.getBaseURI());
+
+        if (options.xsltVersion == 1.0f || options.xsltVersion == 2.0f || options.xsltVersion == 3.0f) {
```
dizzzz, line 118: surprised that this works, as 3 could be 2.99999999 hence the comparison fails. Should use Flood.equals() instead?

dizzzz (approved):
Very impressed by the work, and the needed complexity and wrapping.

dizzzz:
it would be nice if we could get of the old unmaintained xalan lib.....

exist-core/src/test/java/xquery/xquery3/XQuery3Tests.java
alanpaxton: I wanted a record/reminder of how to run a single test. Edited comment to make that clear.

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
alanpaxton: Re-ordered.

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Transform.java
alanpaxton: We're probably OK if the version is never subjected to arithmetic ? Anyway, I got paranoid and built an XSLTVersion class with major/minor which reads from a BigDecimal.

dizzzz (approved):

PR #4390, opened 2022-05-13
Fix additional issues with ResourceType and MediaType during Restore

adamretter:
This fixes https://github.com/eXist-db/exist/issues/4368 and builds on the work by @reinhapa here: https://github.com/eXist-db/exist/pull/4379

If approved, I can add these commits to https://github.com/eXist-db/exist/pull/4380 for eXist-db 5.x.x.


exist-core/src/main/java/org/exist/backup/restore/RestoreHandler.java
```diff
@@ -303,17 +303,12 @@ private DeferredPermission restoreResourceEntry(final Attributes atts) throws SA
...
-            } else if (mimeType.isXMLType() !=  xmlType) {
-                listener.warn("Calculated mime type " + mimeType + " type differs from the one specified in the backup, adjusting it");
-                mimeType = new MimeType(mimeType.getName(), xmlType ? MimeType.XML : MimeType.BINARY);
-            }
+        final MimeType mimeType;
+        if (mimeTypeStr == null || mimeTypeStr.trim().isEmpty()) {
+            mimeType = xmlType ? MimeType.XML_TYPE : MimeType.BINARY_TYPE;
+            listener.warn("Missing mimetype attribute in the backup __contents__.xml file for: " + name + ", assuming: " + mimeType);
+        } else {
+            mimeType = new MimeType(mimeTypeStr.trim(), xmlType ? MimeType.XML : MimeType.BINARY);
```
reinhapa, line 311: Could this be moved into a protected method on a new parent class to be used on this and the `SystemImportHandler` in order to have no code duplication? In a later step there are some other duplicate methods, that could be pulled up in a second step later...

exist-core/src/main/java/org/exist/backup/restore/SystemImportHandler.java
```diff
@@ -321,12 +321,12 @@ private DeferredPermission restoreResourceEntry(final Attributes atts) throws SA
...
-        if (mimeTypeStr != null) {
-            mimeType = MimeTable.getInstance().getContentType(mimeTypeStr);
-        }
-        if (mimeType == null) {
+        final MimeType mimeType;
+        if (mimeTypeStr == null || mimeTypeStr.trim().isEmpty()) {
             mimeType = xmlType ? MimeType.XML_TYPE : MimeType.BINARY_TYPE;
+            listener.warn("Missing mimetype attribute in the backup __contents__.xml file for: " + name + ", assuming: " + mimeType);
+        } else {
+            mimeType = new MimeType(mimeTypeStr.trim(), xmlType ? MimeType.XML : MimeType.BINARY);
```
reinhapa, line 329: Same code as within `RestoreHandler` see comment above...

exist-core/src/test/java/org/exist/backup/XMLDBRestoreTest.java
```diff
@@ -290,57 +298,55 @@ private static void restoreBackup(final XmldbURI uri, final Path backup, @Nullab
...
+        col1Contents.append("</collection>");
 
         final Path dbContentsFile = Files.write(db.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR), dbContents.getBytes(UTF_8));
-        final Path col1ContentsFile = Files.write(col1.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR), col1Contents.getBytes(UTF_8));
-        Files.write(col1.resolve("doc1.xml"), doc1.getBytes(UTF_8));
-        Files.write(col1.resolve("doc2.xml"), doc2.getBytes(UTF_8));
-        Files.write(col1.resolve("doc3.xml"), doc3.getBytes(UTF_8));
-        Files.write(col1.resolve("doc4.html"), doc4.getBytes(UTF_8));
-        Files.write(col1.resolve("doc5.html"), doc5.getBytes(UTF_8));
+        final Path col1ContentsFile = Files.write(col1.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR), col1Contents.toString().getBytes(UTF_8));
```
reinhapa, line 346: I guess the local variable `col1ContentsFile`can be dropped as it is not used by itself (keeping the write operation)

reinhapa (changes requested):

exist-core/src/main/java/org/exist/backup/restore/RestoreHandler.java
adamretter: @reinhapa I think that sort of refactoring is more than I want to achieve in this PR. So I have opened a new PR here - https://github.com/eXist-db/exist/pull/4392 that takes care of the refactoring and could be merged first if you like? I could then rebase this atop it.

reinhapa (approved):

dizzzz (approved):

PR #4394, opened 2022-05-20
Feature/sql to use jdbc xars

lcahlander:
### Description:
Loads the JDBC driver using the Class.forName() call.  It allows for the JDBC driver files that I have created as packages to be utilized from the SQL connect.


lcahlander:
I do not think that the test failures are related to my code change.  The failures are on the windows-latest tests and that appears to be the same with other pull requests.

extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/GetConnectionFunction.java
```diff
@@ -132,10 +127,25 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
...
 
+    /**
+     * This accesses the JDBC driver that has been loaded as an XAR package.
+     *
+     * @param jdbcDriver The classpath of the JDBC driver
+     * @return true if the driver is successfully loaded.
+     */
+    private boolean testClass(final StringValue jdbcDriver) {
+        try {
+            return (Class.forName(jdbcDriver.getStringValue()) != null);
```
adamretter, line 138: As far as I understand it @lcahlander since JDBC 4.0 using `Class.forName` is no longer required.

lcahlander:
@adamretter In order for the libraries in the XARs to be accessible, the Class.forName() is necessary.  I am thinking of adding a driver available function in the SQL to determine if the library is available.

adamretter:
@lcahlander I think it depends... If your drivers are JDBC 4 compliant then I don't think you actually need the `Class.forName` even in a Xar. I might be wrong though...

If they are older drivers you will need that. Do you have an example of an older driver you are using?

lcahlander:
Does it hurt anything to do the call?

adamretter:
> Does it hurt anything to do the call?

I don't know, but my preference would be not to add it if we don't need it. Can you tell me which driver you are using that shows the problem?

lcahlander:
I have installed my XAR for MySQL in the current eXist-db 6.0.1

image

I then try calling my MySQL database through the standard sql:get-connection()

image

When I connect to the MySQL database in my model harvester, I use the Class.forName() call to have the JDBC driver available to my code.  I can change things so that there is a call to see if the driver is loaded into eXist-db.

extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/GetConnectionFunction.java
dizzzz: no, unloading a class is  not possible (yet) ..... I typically have to restart the java process.

Issue #3463, opened 2020-06-24
[BUG] Invalid function parameter type check on document element


adamretter:
```
declare namespace cc = "http://exist-db.org/collection-config/1.0";

declare function local:fun1($x as document-node(element(cc:collection))) {
 	$x
};

local:fun1(
	document {
		<collection xmlns="http://exist-db.org/collection-config/1.0"/>
	}
)
```

This might be related to in-memory nodes only... need to investigate...

joewiz:
The error in 5.3.0-SNAPSHOT:

> exerr:ERROR Type error in expression: required node name is cc:collection; got: #document [at line 7, column 1]

PR #4395, opened 2022-05-21
Fix the Node Test for document-node(element(name))

adamretter:
Closes https://github.com/eXist-db/exist/issues/3463

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/DynamicNameCheck.java
+++ b/exist-core/src/main/java/org/exist/xquery/DynamicNameCheck.java
@@ -21,11 +21,14 @@
  */
 package org.exist.xquery;
 
+import org.exist.dom.INode;
 import org.exist.dom.persistent.DocumentSet;
 import org.exist.dom.persistent.NodeProxy;
 import org.exist.dom.QName;
 import org.exist.xquery.util.ExpressionDumper;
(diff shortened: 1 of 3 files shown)


exist-core/src/main/java/org/exist/xquery/DynamicNameCheck.java
```diff
@@ -77,56 +80,52 @@ public Sequence eval(Sequence contextSequence, Item contextItem) throws XPathExc
...
-        final String localName = name.getLocalPart();
-        if (prefix == null) {
-            final String namespaceURI = name.getNamespaceURI();
-            if (namespaceURI == null) {
-                return localName;
-            } else {
-                return "{'" + namespaceURI + "'}:" + localName;
-            }
-        } else if (prefix.isEmpty()) {
-            return "{''}:" + localName;
```
dizzzz, line 127: i Remember that `{ .. }` is asserted in the XQTS suite?

adamretter: @dizzzz This is only used for the error message. I don't think that is checked in the tests... or if you can show me an example/link please?

dizzzz:
@adamretter  did you check the xqts suite?

PR #4408, opened 2022-05-25
[4.x.x] Improve out-of-scope detection for binary variables

adamretter:
This PR fixes *some* cases where `xs:base64Binary` value variables within an XQuery are cleared before they should be.

# Diagnosis of the problem

If you have a function which returns a Map containing an `xs:base64Binary` (specifically the result of an EXPath HTTP Client response), and then you try and use that value in a calling function you will encounter an error where the temporary file backing the xs:base64Binary value has incorrectly already been deleted.

What is happening is:

1. The EXPath HTTP Client response constructs an `xs:base64Binary` value using `org.exist.xquery.value.BinaryValueFromFile`.
2. `BinaryValueFromFile` stores the binary data in a temporary file.
3. When the xs:base64BinaryValue is returned from the function inside a Map, eXist-db does not recognise it is inside a Map, and so incorrectly assumes it is no longer required and calls `destroy` on the variable; this is done to clean up variables that are no longer needed.
4. The destroy call on the variable, causes `BinaryValueFromFile` to delete the temporary file.
5. When we try and use the value in the Map returned by the function, the binary value cannot be read from the temporary file because it was deleted in the previous step.

# Further work in Future
Ultimately, how eXist-db determines whether a Variable is still "in-scope" or "out-of-scope" within an XQuery is far too limited and does not match the semantics of the XQuery language specifications.
The current implementation of this in eXist-db is mostly within `XQueryContext#markLocalVariables` and `XQueryContext#popLocalVariables`.

Fixing this completely would require substantiate changes to both eXist-db's XQuery Parser and Interpreter.

# Post-script
It is not possible to write a test for this with the way that eXist-db is currently structured. A test for this should be in the Core module. However, the only way I have found to easily trigger it is via the EXPath HTTP Client which is itself in Extension Modules. To introduce a dependency between Core and Extension Modules, when there is already a dependency between Extension Modules and Core would cause a cycle in the build graph (and the build would be rejected by Maven).

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapExpr.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapExpr.java
@@ -77,12 +77,21 @@ public class MapExpr extends AbstractExpression {
     @Override
     public void dump(final ExpressionDumper dumper) {
         dumper.display("map {");
-        for (final Mapping mapping : this.mappings) {
+        for (int i = 0; i < this.mappings.size(); i++) {
+            final Mapping mapping = this.mappings.get(i);
+            if (i > 0) {
+                dumper.display(", ");
+            }
             mapping.key.dump(dumper);
-            dumper.display(" := ");
+            dumper.display(" : ");
             mapping.value.dump(dumper);
         }
         dumper.display("}");
+    }
+
+    @Override
+    public String toString() {
+        return ExpressionDumper.dump(this);
     }
 
     @Override
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
@@ -98,7 +98,10 @@ public class MapType extends AbstractMapType {
     public boolean containsReference(final Item item) {
         for (final Iterator<Map.Entry<AtomicValue, Sequence>> it = map.iterator(); it.hasNext();) {
             final Sequence value = it.next().getValue();
-            return value == item;
+            if (value == item) {
+                return true;
+            }
+            return value.containsReference(item);
         }
         return false;
     }
@@ -107,7 +110,10 @@ public class MapType extends AbstractMapType {
     public boolean contains(final Item item) {
         for (final Iterator<Map.Entry<AtomicValue, Sequence>> it = map.iterator(); it.hasNext();) {
             final Sequence value = it.next().getValue();
-            return value.equals(item);
+            if (value.equals(item)) {
+                return true;
+            }
+            return value.contains(item);
         }
         return false;
     }
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
@@ -94,6 +94,24 @@ public class MapType extends AbstractMapType {
         return this.map.containsKey(key);
     }
 
+    @Override
+    public boolean containsReference(final Item item) {
+        for (final Iterator<Map.Entry<AtomicValue, Sequence>> it = map.iterator(); it.hasNext();) {
+            final Sequence value = it.next().getValue();
+            return value == item;
+        }
+        return false;
+    }
+
+    @Override
+    public boolean contains(final Item item) {
+        for (final Iterator<Map.Entry<AtomicValue, Sequence>> it = map.iterator(); it.hasNext();) {
+            final Sequence value = it.next().getValue();
+            return value.equals(item);
+        }
+        return false;
+    }
+
     public Sequence keys() {
         final ValueSequence seq = new ValueSequence();
         for (final Map.Entry<AtomicValue, Sequence> entry: this.map) {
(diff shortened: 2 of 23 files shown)


exist-core/src/main/java/org/exist/xquery/value/SubSequence.java
```diff
@@ -423,6 +423,36 @@ public void clearContext(final int contextId) throws XPathException {
...
+        try {
+            for (final SequenceIterator it = iterate(); it.hasNext(); ) {
+                final Item i = it.nextItem();
+                if (i == item) {
+                    return true;
+                }
+            }
+            return false;
+        } catch (final XPathException e) {
+            return false;
```
reinhapa, line 437: Should we not log the error somewhere here?

exist-core/src/main/java/org/exist/xquery/value/SubSequence.java
```diff
@@ -423,6 +423,36 @@ public void clearContext(final int contextId) throws XPathException {
...
+        try {
+            for (final SequenceIterator it = iterate(); it.hasNext(); ) {
+                final Item i = it.nextItem();
+                if (i.equals(item)) {
+                    return true;
+                }
+            }
+            return false;
+        } catch (final XPathException e) {
+            return false;
```
reinhapa, line 452: Should we not log the error somewhere here?

exist-core/src/main/java/org/exist/xquery/functions/map/MapExpr.java
```diff
@@ -77,14 +77,23 @@ public void accept(final ExpressionVisitor visitor) {
...
+            final Mapping mapping = this.mappings.get(i);
+            if (i > 0) {
+                dumper.display(", ");
+            }
             mapping.key.dump(dumper);
-            dumper.display(" := ");
+            dumper.display(" : ");
             mapping.value.dump(dumper);
         }
         dumper.display("}");
```
reinhapa, line 89: I could imagine using a single boolean would remove the total amount of additional calls eventually:
```
boolean firstMapping = true;
for (final Mapping mapping : this.mappings) {
    mapping.key.dump(dumper);
    dumper.display(" := ");
    dumper.display(" : ");
    mapping.value.dump(dumper);
    if (firstMapping) {
      firstMapping = false;
    } else {
        dumper.display(", ");
    }
}
```

reinhapa (changes requested):
I have some minor suggestions... ;-)

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
```diff
@@ -94,6 +94,30 @@ public boolean contains(AtomicValue key) {
...
     }
 
+    @Override
+    public boolean containsReference(final Item item) {
+        for (final Iterator<Map.Entry<AtomicValue, Sequence>> it = map.iterator(); it.hasNext();) {
+            final Sequence value = it.next().getValue();
+            if (value == item) {
+                return true;
+            }
+            return value.containsReference(item);
```
reinhapa, line 104: I guess this could have a problem, if there are more than one sequence entry.  Should may be written like so:
```
if (value == item || value.containsReference(item)) {
  return true;
}
```

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
```diff
@@ -94,6 +94,30 @@ public boolean contains(AtomicValue key) {
...
+    }
+
+    @Override
+    public boolean contains(final Item item) {
+        for (final Iterator<Map.Entry<AtomicValue, Sequence>> it = map.iterator(); it.hasNext();) {
+            final Sequence value = it.next().getValue();
+            if (value.equals(item)) {
+                return true;
+            }
+            return value.contains(item);
```
reinhapa, line 116: I guess this could have a problem, if there are more than one sequence entry.  Should may be written like so:
```
if (value.equals(item) || value.containsReference(item)) {
  return true;
}
```

reinhapa (changes requested):
I found some spots, that may need a second look....

exist-core/src/main/java/org/exist/xquery/value/SubSequence.java
adamretter: Okay fixed.

exist-core/src/main/java/org/exist/xquery/value/SubSequence.java
adamretter: Okay fixed.

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
adamretter: Okay fixed.

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
adamretter: Okay fixed.

exist-core/src/main/java/org/exist/xquery/functions/map/MapExpr.java
adamretter: If I am not mistaken you code will end up with a trailing `,` as it postfixes with the comma character, whereas my code prefixes with the comma character to ensure no "dangling" comma.

adamretter:
Thanks @reinhapa I have pushed an additional commit which I think addresses the issues that you found.

reinhapa (approved):

exist-core/src/main/java/org/exist/xquery/functions/map/MapExpr.java
reinhapa: Sorry, the if/else block should of course go prior the key/value rendering...

PR #4409, opened 2022-05-25
Improve out-of-scope detection for binary variables

adamretter:
Forward port of https://github.com/eXist-db/exist/pull/4408

### DIFF
--- a/exist-core/src/main/java/org/exist/dom/persistent/AVLTreeNodeSet.java
+++ b/exist-core/src/main/java/org/exist/dom/persistent/AVLTreeNodeSet.java
@@ -352,6 +352,48 @@ public class AVLTreeNodeSet extends AbstractNodeSet {
         return searchData(root, proxy) != null;
     }
 
+    @Override
+    public boolean containsReference(final Item item) {
+        if (root == null) {
+            return false;
+        }
+
+        return containsReference(root, item);
+    }
+
+    private boolean containsReference(@Nullable final Node node, final Item item) {
+        if (node == null) {
+            return false;
+        }
+
+        if (node.data == item) {
+            return true;
+        }
+
+        return containsReference(node.leftChild, item) || containsReference(node.rightChild, item);
+    }
+
+    @Override
+    public boolean contains(final Item item) {
+        if (root == null) {
+            return false;
+        }
+
+        return containsReference(root, item);
+    }
+
+    private boolean contains(@Nullable final Node node, final Item item) {
+        if (node == null) {
+            return false;
+        }
+
+        if (node.data.equals(item)) {
+            return true;
+        }
+
+        return containsReference(node.leftChild, item) || containsReference(node.rightChild, item);
+    }
+
     public void removeNode(final Node node) {
         --size;
         Node tempNode = node;
--- a/exist-core/src/main/java/org/exist/dom/persistent/VirtualNodeSet.java
+++ b/exist-core/src/main/java/org/exist/dom/persistent/VirtualNodeSet.java
@@ -113,6 +113,32 @@ public class VirtualNodeSet extends AbstractNodeSet {
         return false;
     }
 
+    @Override
+    public boolean containsReference(final Item item) {
+        if (!(item instanceof NodeProxy)) {
+            return false;
+        }
+
+        final NodeProxy firstParent = getFirstParent((NodeProxy) item, null, axis == Constants.SELF_AXIS, 0);
+        if (firstParent == item) {
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public boolean contains(final Item item) {
+        if (!(item instanceof NodeProxy)) {
+            return false;
+        }
+
+        final NodeProxy firstParent = getFirstParent((NodeProxy) item, null, axis == Constants.SELF_AXIS, 0);
+        if (firstParent.equals(item)) {
+            return true;
+        }
+        return false;
+    }
+
     public void setInPredicate(final boolean predicate) {
         inPredicate = predicate;
     }
--- a/exist-core/src/main/java/org/exist/dom/persistent/VirtualNodeSet.java
+++ b/exist-core/src/main/java/org/exist/dom/persistent/VirtualNodeSet.java
@@ -120,10 +120,7 @@ public class VirtualNodeSet extends AbstractNodeSet {
         }
 
         final NodeProxy firstParent = getFirstParent((NodeProxy) item, null, axis == Constants.SELF_AXIS, 0);
-        if (firstParent == item) {
-            return true;
-        }
-        return false;
+        return firstParent == item;
     }
 
     @Override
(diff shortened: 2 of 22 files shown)


exist-core/src/main/java/org/exist/dom/persistent/AVLTreeNodeSet.java
```diff
@@ -352,6 +352,48 @@ public final boolean contains(final NodeProxy proxy) {
...
+            return false;
+        }
+
+        if (node.data.equals(item)) {
+            return true;
+        }
+
+        return containsReference(node.leftChild, item) || containsReference(node.rightChild, item);
+    }
+
```
reinhapa, line 396: Seems that this method is not used.. remove?

exist-core/src/main/java/org/exist/dom/persistent/VirtualNodeSet.java
```diff
@@ -113,6 +113,32 @@ public boolean contains(final NodeProxy p) {
...
+        if (!(item instanceof NodeProxy)) {
+            return false;
+        }
+
+        final NodeProxy firstParent = getFirstParent((NodeProxy) item, null, axis == Constants.SELF_AXIS, 0);
+        if (firstParent == item) {
+            return true;
+        }
+        return false;
+    }
```
reinhapa, line 127: why not `return firstParent == item` ?

exist-core/src/main/java/org/exist/dom/persistent/VirtualNodeSet.java
```diff
@@ -113,6 +113,32 @@ public boolean contains(final NodeProxy p) {
...
+    public boolean contains(final Item item) {
+        if (!(item instanceof NodeProxy)) {
+            return false;
+        }
+
+        final NodeProxy firstParent = getFirstParent((NodeProxy) item, null, axis == Constants.SELF_AXIS, 0);
+        if (firstParent.equals(item)) {
+            return true;
+        }
+        return false;
```
reinhapa, line 139: why not return `firstParent.equals(item)` ?

exist-core/src/main/java/org/exist/xquery/functions/array/ArrayType.java
```diff
@@ -382,6 +382,30 @@ public static Sequence flatten(Sequence input) throws XPathException {
...
     }
 
+    @Override
+    public boolean containsReference(final Item item) {
+        for (int i = 0; i < vector.length(); i++) {
+            final Sequence value = vector.nth(i);
+            if (value == item) {
+                return true;
+            }
+            return value.containsReference(item);
```
reinhapa, line 392: I think this will not work if the reference is not the first entry. I guess it should be:
```
if (value == item || value.containsReference(item)) {
  return true;
}
```

exist-core/src/main/java/org/exist/xquery/functions/array/ArrayType.java
```diff
@@ -382,6 +382,30 @@ public static Sequence flatten(Sequence input) throws XPathException {
...
+    @Override
+    public boolean contains(final Item item) {
+        for (int i = 0; i < vector.length(); i++) {
+            final Sequence value = vector.nth(i);
+            if (value.equals(item)) {
+                return true;
+            }
+            return value.contains(item);
+        }
+        return false;
```
reinhapa, line 406: Same here:
```
if (value.equals(item) || value.containsReference(item)) {
  return true;
}
```

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
```diff
@@ -216,6 +216,24 @@ public boolean contains(AtomicValue key) {
         return map.contains(key);
     }
 
+    @Override
+    public boolean containsReference(final Item item) {
+        for (final Iterator<Sequence> it = map.values().iterator(); it.hasNext();) {
+            final Sequence value = it.next();
+            return value == item;
```
reinhapa, line 223: Same here:
```
if (value==item) {
  return true;
}
```

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
```diff
@@ -216,6 +216,24 @@ public boolean contains(AtomicValue key) {
...
+        }
+        return false;
+    }
+
+    @Override
+    public boolean contains(final Item item) {
+        for (final Iterator<Sequence> it = map.values().iterator(); it.hasNext();) {
+            final Sequence value = it.next();
+            return value.equals(item);
```
reinhapa, line 232: Same here but so:
```
if (value.equals(item)) {
  return true;
}
```

reinhapa (changes requested):
I found some spots, that may need a second look....

exist-core/src/main/java/org/exist/dom/persistent/VirtualNodeSet.java
adamretter: Okay fixed.

exist-core/src/main/java/org/exist/dom/persistent/VirtualNodeSet.java
adamretter: Okay fixed.

exist-core/src/main/java/org/exist/xquery/functions/array/ArrayType.java
adamretter: AFAIK the logic you have suggested is exactly the same as what is already there. What am I missing?

exist-core/src/main/java/org/exist/xquery/functions/array/ArrayType.java
adamretter: AFAIK the logic you have suggested is exactly the same as what is already there. What am I missing?

exist-core/src/main/java/org/exist/xquery/functions/array/ArrayType.java
adamretter: Okay fixed.

exist-core/src/main/java/org/exist/xquery/functions/array/ArrayType.java
adamretter: Okay fixed.

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
adamretter: Okay fixed.

exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
adamretter: Okay fixed.

adamretter:
Thanks @reinhapa I have pushed an additional commit which I think addresses the issues that you found.

reinhapa (approved):

Issue #4412, opened 2022-05-28
Invalid predicate optimisation when calling function by reference


adamretter:
The following query incorrectly raises an error as it tries to send the entire sequence `("a", "b", "c")` into the argument `$y`:

```xquery
let $predicate := function($x as xs:string, $y as xs:string) as xs:boolean { $x eq $y }
return
	("a", "b", "c")[$predicate("b", .)]
```

Yielding the error:
```
An exception occurred during query execution: exerr:ERROR XPTY0004: The actual cardinality for parameter 2 does not match the cardinality declared in the function's signature: ($x as xs:string, $y as xs:string) xs:boolean. Expected cardinality: exactly one, got 3.
```

This query also fails due to the same cause:
```xquery
declare function local:pred($x as xs:string, $y as xs:string) as xs:boolean {
	$x eq $y
};

let $predicate := local:pred#2
return
	("a", "b", "c")[$predicate("b", .)]
```

Yielding the error:
```
An exception occurred during query execution: exerr:ERROR XPTY0004: The actual cardinality for parameter 2 does not match the cardinality declared in the function's signature: local:pred($x as xs:string, $y as xs:string) xs:boolean. Expected cardinality: exactly one, got 3.
```

The problem can be simplified to just:
```xquery
("a", "b", "c")[function($x as xs:string, $y as xs:string) as xs:boolean { $x eq $y }("b", .)]
```
Also yielding the error:
```
The actual cardinality for parameter 2 does not match the cardinality declared in the function's signature: ($x as xs:string, $y as xs:string) xs:boolean. Expected cardinality: exactly one, got 3
```

---
The following query which uses a function call as opposed to calling a function reference does work as expected:
```xquery
declare function local:pred($x as xs:string, $y as xs:string) as xs:boolean {
	$x eq $y
};

("a", "b", "c")[local:pred("b", .)]
```
Yielding the result `b`.

----
The problem is present in at least eXist-db `4.11.0-SNAPSHOT` through to `6.1.0-SNAPSHOT`.

I suspect this is another predicate optimisation issue, whereby it is attempting to incorrectly optimise the call to the Function Reference.

adamretter:
As suspected, for a Function Reference, within the predicate it fails to detect the dependency on the Context Item and attempts a "positional promotion" here - https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/xquery/Predicate.java#L235

joewiz:
[...]

PR #4414, opened 2022-05-29
Fix the use of Function Reference from with a Predicate

adamretter:
Closes https://github.com/eXist-db/exist/issues/4412

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
+++ b/exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
@@ -117,4 +117,31 @@ public class DynamicFunctionCall extends AbstractExpression {
         functionExpr.resetState(postOptimization);
         arguments.forEach(arg -> arg.resetState(postOptimization));
     }
+
+    @Override
+    public String toString() {
+        final StringBuilder builder = new StringBuilder();
+        if (functionExpr != null) {
+            builder.append(functionExpr);
+        } else {
+            builder.append("DynamicFunctionCall{arguments=");
+        }
+
+        builder.append('(');
+        if (arguments != null) {
+            for (int i = 0; i < arguments.size(); i++) {
+                if (i > 0) {
+                    builder.append(", ");
+                }
+                builder.append(arguments.get(i).toString());
+            }
+        }
+        builder.append(')');
+
+        if (functionExpr == null) {
+            builder.append('}');
+        }
+
+        return builder.toString();
+    }
 }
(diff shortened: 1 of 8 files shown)


exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
```diff
@@ -117,4 +128,31 @@ public void resetState(final boolean postOptimization) {
...
+        }
+
+        builder.append('(');
+        if (arguments != null) {
+            for (int i = 0; i < arguments.size(); i++) {
+                if (i > 0) {
+                    builder.append(", ");
+                }
+                builder.append(arguments.get(i).toString());
+            }
```
reinhapa, line 148: This could may be written in this way to reduce the amount of calls:
```
boolean firstArgument = true;
for (final Expression argument : arguments) {
    builder.append(argument.toString());
    if (firstArgument) {
      firstArgument = false;
    } else {
        builder.append(", ");
    }
}
```

reinhapa (changes requested):
I got a small suggestion here

exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
adamretter: I think this will lead to one too many comma characters, i.e. a dangling comma at the end... or am I mistaken?

reinhapa: Got the wrong order should be like this:
```
boolean firstArgument = true;
for (final Expression argument : arguments) {
    if (firstArgument) {
        firstArgument = false;
    } else {
        builder.append(", ");
    }
    builder.append(argument.toString());
}
```

exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
```diff
@@ -31,20 +31,31 @@
...
         this.arguments = args;
         this.isPartial = partial;
     }
 
+    @Override
+    public int getDependencies() {
+        int dependencies = functionExpr.getDependencies();
+        if (arguments != null) {
+            for (final Expression argument : arguments) {
+                dependencies |= argument.getDependencies();
```
dizzzz, line 53: "|=" .... need to check :-=)

dizzzz (approved):

exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
adamretter: @reinhapa This intrigued me so I decided to benchmark it with JMH, results are here:
Screenshot 2022-06-10 at 17 05 05

The JMH code looks like this: https://github.com/evolvedbinary/exist/blob/hotfix/predicate-function-reference/exist-core-jmh/src/main/java/org/exist/xquery/utils/StringJoinBenchmark.java#L48

Ultimately it seems that your and my approaches are so similar in terms of performance that the deviation indicated by JMH would account for them being identical in performance.

As the performance is the same, do you still want me to change the code?

reinhapa (approved):

PR #4416, opened 2022-05-29
[4.x.x] Fix the use of Function Reference from with a Predicate

adamretter:
Backport of https://github.com/eXist-db/exist/pull/4414

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
+++ b/exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
@@ -10,13 +10,13 @@ import org.exist.xquery.value.Type;
 
 public class DynamicFunctionCall extends AbstractExpression {
 
-    private Expression functionExpr;
-    private List<Expression> arguments;
-    private boolean isPartial = false;
+    private final Expression functionExpr;
+    private final List<Expression> arguments;
+    private final boolean isPartial;
     
     private AnalyzeContextInfo cachedContextInfo;
 
-    public DynamicFunctionCall(XQueryContext context, Expression fun, List<Expression> args, boolean partial) {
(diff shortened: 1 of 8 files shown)


exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
```diff
@@ -96,4 +107,31 @@ public void resetState(final boolean postOptimization) {
...
+        }
+
+        builder.append('(');
+        if (arguments != null) {
+            for (int i = 0; i < arguments.size(); i++) {
+                if (i > 0) {
+                    builder.append(", ");
+                }
+                builder.append(arguments.get(i).toString());
+            }
```
reinhapa, line 127: Following would reduce the amount of calls:
```
boolean firstArgument = true;
for (final Expression argument : arguments) {
    builder.append(argument.toString());
    if (firstArgument) {
        firstArgument = false;
    } else {
        builder.append(", ");
    }
}
```

reinhapa (changes requested):
Just a small suggestion

exist-core/src/main/java/org/exist/xquery/DynamicFunctionCall.java
adamretter: @reinhapa I think this will lead to one too many comma characters, i.e. a dangling comma at the end... or am I mistaken?

reinhapa: @adamretter got my order wrong, should be:
```
boolean firstArgument = true;
for (final Expression argument : arguments) {
    if (firstArgument) {
        firstArgument = false;
    } else {
        builder.append(", ");
    }
    builder.append(argument.toString());
}
```

dizzzz: So.... wassup?

PR #4433, opened 2022-06-07
Fix numerous bugs in numeric operations

adamretter:
It seems eXist-db had quite a lot of bugs in its numeric operations, most of these were around missing type promotions. This PR adds 72 tests and a bunch of fixes.

This was previously leading to false negatives in seemingly unrelated XQTS tests.
Closes https://github.com/eXist-db/exist-xqts-runner/issues/23

https://i.pinimg.com/originals/9f/7d/89/9f7d89ea2c67cb6cb0daf85fa7e52e0a.png

### DIFF
--- a/exist-core/src/test/xquery/numericOp/numeric-add.xqm
+++ b/exist-core/src/test/xquery/numericOp/numeric-add.xqm
@@ -1,0 +1,71 @@
+(:
+ : eXist-db Open Source Native XML Database
+ : Copyright (C) 2001 The eXist-db Authors
+ :
+ : info@exist-db.org
+ : http://www.exist-db.org
+ :
+ : This library is free software; you can redistribute it and/or
+ : modify it under the terms of the GNU Lesser General Public
+ : License as published by the Free Software Foundation; either
+ : version 2.1 of the License, or (at your option) any later version.
+ :
+ : This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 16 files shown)


exist-core/src/test/xquery/numericOp/numeric-add.xqm
```diff
@@ -0,0 +1,71 @@
...
+declare
+    %test:args("1.13", "1.13")
+    %test:assertEquals("2.2599999952316283")
+function ona:numeric-add-float-double($f as xs:float, $d as xs:double) {
+    $f + $d
+};
+
+declare
+    %test:args("1.13", "1.13")
+    %test:assertEquals("2.2599999952316283")
```
dizzzz, line 40: is this the same for apples M1 and intel?

dizzzz (approved):

exist-core/src/test/xquery/numericOp/numeric-add.xqm
adamretter: Why would it not be?

dizzzz: cause i think these 'rounding' can be processor specific.....

adamretter: Ah no. In this instance I think it is defined IEEE for floating point numbers and the W3C specs

Issue #3623, opened 2020-11-19
[BUG] error location for function type in sub-expression in element content 


line-o:
**Describe the bug**

The location of an expression evaluating to a function type inside of a sub-expression inside of element content must be reported.

Example:

```xquery
element test { true(), map{} }
```

**Expected behavior**

Throws an error at runtime but without location information.

**To Reproduce**

There are already pending tests in place in #3550, which adds location to the first set of errors of this type.

joewiz:
The example above is missing the enclosed content expression required by the computed element constructor. To match the description of the issue, the test would need an enclosed content expression. For example:

```xquery
element foo { map{} }
```

The error raised by this revised query is:

> err:XQTY0105 Enclosed expression contains function item

The references pending tests can be found here:

https://github.com/eXist-db/exist/blob/8d2dcbccde87de5adcedfd34b4aad033fa78a6e8/exist-core/src/test/xquery/errors.xql#L328-L360.

line-o:
> The example above is missing the enclosed content expression required by the computed element constructor. 
@joewiz of course! That was just a type, I changed it in the description. Thanks for pointing it out.

line-o:
fixed by c553667221306f4df149c622e0e4f2074052da60

PR #4444, opened 2022-06-15
[bugfix] Add missing location information in errors for functions inside element constructors (err:XQTY0105)

chris-evolvedbinary:
### Description:

Add location information to map expressions, and pass this information to the `XQTY0105` exception that occurs when an enclosed expression contains a function item

### Reference:

https://github.com/eXist-db/exist/issues/3623

### Type of tests:

`subexpression-in-enclosed-expression-evaluates-to-map` and `enclosed-expression-evaluates-to-map` in `exist-core\src\test\xquery\errors.xql`

--

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/value/FunctionReference.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/FunctionReference.java
@@ -192,4 +192,25 @@ public class FunctionReference extends AtomicValue implements AutoCloseable {
     public String toString() {
         return "anonymous-function#" + functionCall.getArgumentCount();
     }
+
+    /** the expression from which this value derives */
+    private Expression expression = null;
+
+    /**
+     * Gets the expression from which this value derives.
+     *
+     * @return  the expression from which this value derives
+     */
+    public Expression getExpression() {
+        return expression;
+    }
+
+    /**
+     * Sets the expression from which this value derives.
+     *
+     * @param   expression  the expression to use
+     */
+    public void setExpression(Expression expression) {
+        this.expression = expression;
+    }
 }
--- a/exist-core/src/main/java/org/exist/xquery/value/FunctionReference.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/FunctionReference.java
@@ -70,6 +70,7 @@ public class FunctionReference extends AtomicValue implements AutoCloseable {
      *
      * @return  the expression from which this type derives
      */
+    @Override
     public Expression getExpression() {
         return expression;
     }
--- a/exist-core/src/main/java/org/exist/xquery/value/FunctionReference.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/FunctionReference.java
@@ -36,6 +36,9 @@ import java.util.List;
  */
 public class FunctionReference extends AtomicValue implements AutoCloseable {
 
+    /** the expression from which this type derives */
+    private Expression expression;
+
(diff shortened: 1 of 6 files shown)


exist-core/src/main/java/org/exist/xquery/value/FunctionReference.java
```diff
@@ -170,4 +170,25 @@ public AtomicValue min(Collator collator, AtomicValue other)
     public AtomicValue atomize() throws XPathException {
         throw new XPathException(ErrorCodes.FOTY0013, "A function item other than an array cannot be atomized");
     }
+
+    /** the expression from which this value derives */
+    private Expression expression = null;
```
dizzzz, line 175: is this null assignment really needed?

dizzzz (approved):

dizzzz:
image

adamretter (changes requested):
This solves the problem for a Map expr inside an element constructor. But I think there are several more types of expr and function that we need to at least add tests for, and then fix potentially if they don't work already:

1. What about `ArrayExpr`?
2. What about evaluation of a`FunctionReference`? e.g. `let $f := function($x) { fn:error(xs:QName("local:my-error") } return element { $x() }`
3. What about `UserDefinedFunction`? i.e. functions implemented in XQuery? You could implement an XQuery function that wraps `fn:error` for testing this.
4. What about `BasicFunction`? i.e. functions for XQuery implemented in Java? You could use fn:error for testing this
5. What about other types of expression inside the element constructor... as just one example: Attribute Constructor:
`element e1 { attribute a1 { fn:error(xs:QName("local:my-error")) } }`

chris-evolvedbinary:
> This solves the problem for a Map expr inside an element constructor. But I think there are several more types of expr and function that we need to at least add tests for, and then fix potentially if they don't work already:
> 
> 1. What about `ArrayExpr`?
> 2. What about evaluation of a`FunctionReference`? e.g. `let $f := function($x) { fn:error(xs:QName("local:my-error") } return element { $x() }`
> 3. What about `UserDefinedFunction`? i.e. functions implemented in XQuery? You could implement an XQuery function that wraps `fn:error` for testing this.
> 4. What about `BasicFunction`? i.e. functions for XQuery implemented in Java? You could use fn:error for testing this
> 5. What about other types of expression inside the element constructor... as just one example: Attribute Constructor:
>    `element e1 { attribute a1 { fn:error(xs:QName("local:my-error")) } }`

The `exist-core/src/test/xquery/errors.xql` file already contains tests for these cases, except for the last case ("other types of expression inside the element constructor") for which I added a test.  All the tests pass.

The issue is that the `UserDefinedFunction`, `BasicFunction`, and `OpNumeric` class all derive from the `PathExpr` class, which has a list of expressions (as "steps").  So when errors occur for a `PathExpr` object, the location information is obtainable for an expression of the object.  The `MapType` class, however, derives from the `FunctionReference` class, which does not derive from the `PathExpr` class and therefore does not have information about the expression from which it came.  So the code for this pull request adds an expression property to the `FunctionReference` class.

Also, there does not appear to be an `ArrayExpr` class.  Instead, the `XQueryTreeParser` class uses the `ArrayConstructor` class, which like the `PathExpr` class has a list of expressions (as "arguments").

adamretter:
@chris-evolvedbinary I have pushed one commit which adds some additional tests to cover the issues I mentioned. Hopefully that is helpful?

4 of the tests I added do fail, so I think there are one or two more places where you need to expose the location information.

adamretter (approved):
LGTM

PR #4449, opened 2022-06-20
[4.x.x] Improve security and configurability of functions that access the Environment and System Properties

adamretter:
This adds better security controls for accessing Environment Variables and Java System Properties.

* Fixed the return cardinality type of `fn:environment-variable#1`.
* `fn:available-environment-variables#0` and `fn:enviroment-variable#1` are now deterministic.
* Added the function `util:available-system-properties#0`.
* `util:available-system-properties#0` and `util:system-property#1` are now deterministic.
* Allow access to specific Environment Variables when using fn:available-environment-variables#0 and fn:environment-variable#1 to be configured from conf.xml
* Allow access to specific Java System Properties  when using fn:available-system-properties#0 and fn:system-property#1 to be configured from conf.xml

PR #4452, opened 2022-06-21
Improve security and configurability of functions that access the Environment and System Properties

adamretter:
Forward port of https://github.com/eXist-db/exist/pull/4449


adamretter:
@line-o There is some documentation for this in the `conf.xml` file now. If you think there should also be some for the website, could you let me know where you think it should go and then I will try and add something

adamretter:
Found an issue that I need to fix...

line-o:
I assume this PR contains important changes to exist-db that we should pull in after finalising.
@adamretter would you still happen to know what needed fixing here?

adamretter:
@line-o Sure, just run the changes and the test suite you will see the issue clearly I think, let me know if you don't.

Issue #4438, opened 2022-06-08
Deadlock


adamretter:
When running XQTS (exist-xqts-runner) against eXist-db 6.1.0-SNAPSHOT we can encounter a deadlock in eXist-db.

Running jstack reveals:

```
Found one Java-level deadlock:
========================
"io-compute-42":
  waiting to lock monitor 0x00007f85207a0e08 (object 0x00000007b2918688, a org.exist.storage.BrokerPool),
  which is held by "io-compute-32"
"io-compute-32":
  waiting for ownable synchronizer 0x00000007b0839710, (a java.util.concurrent.locks.ReentrantReadWriteLock$NonfairSync),
  which is held by "global.BrokerPools.ShutdownHook"
"global.BrokerPools.ShutdownHook":
  waiting to lock monitor 0x00007f85207a0e08 (object 0x00000007b2918688, a org.exist.storage.BrokerPool),
  which is held by "io-compute-32"
```

```
Java stack information for the threads listed above:
=======================================
"io-compute-42":
	at org.exist.storage.BrokerPool.get(BrokerPool.java:1258)
	- waiting to lock <0x00000007b2918688> (a org.exist.storage.BrokerPool)
	at org.exist.storage.BrokerPool.getBroker(BrokerPool.java:1200)
	at org.exist.xqts.runner.ExistServer.$anonfun$getConnection$1(ExistServer.scala:112)
	at org.exist.xqts.runner.ExistServer$$Lambda$5145/2103542716.apply(Unknown Source)
	at cats.effect.unsafe.WorkerThread.blockOn(WorkerThread.scala:624)
	at scala.concurrent.package$.blocking(package.scala:124)
	at cats.effect.IOFiber.runLoop(IOFiber.scala:967)
	at cats.effect.IOFiber.execR(IOFiber.scala:1332)
	at cats.effect.IOFiber.run(IOFiber.scala:139)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:425)
"io-compute-32":
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x00000007b0839710> (a java.util.concurrent.locks.ReentrantReadWriteLock$NonfairSync)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
[...]

PR #4471, opened 2022-07-05
Adds test to produce deadlock

reinhapa:
Try to reproduce #4438 using a JUnit test


reinhapa:
@adamretter I might got a test that reproduces the deadlock. Could you look into it and check if this reproduces the problem before I go further trying to solve it? (Might require your help too, as I do not completely understand the state machinery behind it)

dizzzz:
any update?

line-o:
@reinhapa what's the status of this PR. Do you plan to work on this or rather close it?

Issue #4466, opened 2022-07-01
[BUG] local-name is invoked when it shouldn't be 


PieterLamers:
**Describe the bug**
with
```xquery
let $test as element(a):= 
  <a>
    <b></b>
  </a>
  
return $test/c/local-name()
```
`local-name#1` is invoked returning an empty string.

**Expected behavior**
empty-sequence() should be returned, similar to a call to `$test/c/string()`

**Context (please always complete the following information):**
 - OS: Windows, Linux
 - eXist-db version: 6.0.1
 - Java Version Java 8

Cf discussion on Slack: https://xmlcom.slack.com/archives/C011NLXE4DU/p1656648721760709

joewiz:
In the linked xml.com Slack thread, Michael Kay confirmed:

> For the expression X/Y, if X is an empty sequence, then Y doesn't get evaluated and the result is empty.

PR #4473, opened 2022-07-06
Corrects issues with fn:local-name and unnecessary evaluation of path steps on empty sequences

adamretter:
Closes https://github.com/eXist-db/exist/issues/4466

Previously eXist-db was doing more work than it needed, i.e. it was evaluating path steps on empty sequences. This led to both incorrect results and sub-optimal performance. eXist-db also has some tests for this previous behaviour which were incorrect, these tests have also been corrected.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/AbstractExpression.java
+++ b/exist-core/src/main/java/org/exist/xquery/AbstractExpression.java
@@ -183,4 +183,9 @@ public abstract class AbstractExpression implements Expression {
     public Expression getParent() {
         return null;
     }
+
+    @Override
+    public boolean evalNextExpressionOnEmptyContextSequence() {
+        return false;
+    }
 }
--- a/exist-core/src/main/java/org/exist/xquery/DebuggableExpression.java
+++ b/exist-core/src/main/java/org/exist/xquery/DebuggableExpression.java
@@ -194,4 +194,9 @@ public class DebuggableExpression implements Expression, RewritableExpression {
     public Expression getParent() {
         return null;
     }
+
+    @Override
+    public boolean evalNextExpressionOnEmptyContextSequence() {
+        return false;
+    }
 }
--- a/exist-core/src/main/java/org/exist/xquery/Expression.java
+++ b/exist-core/src/main/java/org/exist/xquery/Expression.java
@@ -255,4 +255,13 @@ public interface Expression {
     public boolean allowMixedNodesInReturn();
 
     public Expression getParent();
+
+    /**
+     * Return true only if the next expression within a path expression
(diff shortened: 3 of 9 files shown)


dizzzz (approved):

joewiz:
FYI, I noticed that the merge of this PR has caused eXide’s directory pane and file manager browsing to break in develop HEAD. 

Specifically, eXide's AJAX query to:

> http://localhost:8080/exist/apps/eXide/modules/collections.xq?root=%2Fdb&view=r

… returns:

```json
{
    "total": 37,
    "items": null
}
```

… instead of the same object whose “items” entry contains an array of the 37 child collections and resources of /db. 

The part of the query that responds to this request is https://github.com/eXist-db/eXide/blob/develop/modules/collections.xq#L110. I haven’t traced it further to understand why the “total” appears correct (37, in my case), whereas the “items” entry that should iterate over these 37 items somehow returns empty.

But resetting HEAD back to 4074014c49952ccd2451a835645bdc32b84ba195 (the commit immediately before the merge of #4473) clears the problem up completely.

Another symptom of the same bug is that the CI now [fails](https://github.com/eXist-db/eXide/runs/7383110523?check_suite_focus=true) on [Younes’s PR](https://github.com/eXist-db/eXide/pull/489) whereas before #4473 it [passed](https://github.com/eXist-db/eXide/actions/runs/2672968666).

I didn’t see any other problems in the stock apps - so the documentation, function documentation, and monex all opened up just fine. So it seems to be something about eXide that’s stumbling on #4473.

I've got to go to bed now, but I will investigate further in the morning to see if I can trace this down further in eXide that might’ve been affected by #4473.

joewiz:
I’ve narrowed down the issue with my build of HEAD (which includes #4473) to the following scenario: when an enclosed expression evaluates to an empty sequence, the evaluation of any following expressions is skipped. For example, this query:

```xquery
<abc>
    { 
        ()
    }
    123
</abc>
```

… should evaluate to `<abc>123</abc>`, but with #4473, it evaluates to `<abc/>`.

This explains the situation above, since the "items" entry contains an enclosed expression (at https://github.com/eXist-db/eXide/blob/develop/modules/collections.xq#L113-L128) that evaluates to the empty sequence (since we're asking for the contents of `/db`, which has no parent). And thus it causes the second expression to be skipped - causing the entire "items" entry to return `null` instead of the expected results. 

A similar issue affects eXide's controller.xql (at https://github.com/eXist-db/eXide/blob/develop/controller.xq#L175), which calls a login function that returns the empty sequence. As a result of the following expressions not being evaluated, any call to eXide's "execute" endpoint returns an error. Removing the call to the login function makes the issue evaporate.

Issue #4458, opened 2022-06-28
[BUG] Rest client can't handle file names with special characters


marmoure:
seems like rest client can't handle filenames with special characters in them per example AéB
```
xquery version "3.1";

xmldb:create-collection("/db","AéB"),
xmldb:store(xmldb:encode("/db/AéB"), xmldb:encode("AéB.xml"), <foo/>)
```
after the fixes introduced in https://github.com/eXist-db/eXide/pull/489
it's now possible to create and edit files with special characters and while trying to access these files via the rest client it return document not found
eg url : http://localhost:8080/exist/rest/db/A%C3%A9B/A%C3%A9B.xml to view /db/AéB/AéB.xml
the url is correct and seems like the issue is coming from name encoding on the rest client side

PR #4474, opened 2022-07-06
fix Rest client file encoding 

marmoure:
this Fixes: https://github.com/eXist-db/exist/issues/4458
as mentioned in the issue above the rest client was unable to handle filenames with special characters. This fixes that for all HTTP request GET,POST,PUT,DELETE
now storing resource like 
```xquery
xquery version "3.1";

xmldb:create-collection("/db","AéB"),
xmldb:store(xmldb:encode("/db/AéB"), xmldb:encode("AéB.xml"), <foo/>)
```
and performing operations on it from the rest client works fine eg:
GET: http://localhost:8080/exist/rest/db/A%C3%A9B/A%C3%A9B.xml

### DIFF
--- a/exist-core/src/main/java/org/exist/http/RESTServer.java
+++ b/exist-core/src/main/java/org/exist/http/RESTServer.java
@@ -393,7 +393,7 @@ public class RESTServer {
         // Process the request
         LockedDocument lockedDocument = null;
         DocumentImpl resource = null;
-        final XmldbURI pathUri = XmldbURI.createInternal(path);
+        final XmldbURI pathUri = XmldbURI.create(path);
         try {
             // check if path leads to an XQuery resource
             final String xquery_mime_type = MimeType.XQUERY_TYPE.getName();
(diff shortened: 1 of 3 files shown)


dizzzz:
Quite some tests fail :-(

This is a tricky PR, this subject has been 'a challenge' since ages. Eg with this change, is it still possible to get the document consistently using webdav, and using `doc(..)', and in the Java client?

dizzzz (changes requested):
Please check tests

adamretter:
@dizzzz So this one specifically focuses on improving the handling in the REST Server. It does improve the current situation. We do have some work scheduled here at Evolved Binary in the near-future to improve the situation for other interfaces too. However, I think this PR can standalone and be merged as it is an improvement. Does that sound reasonable?

dizzzz (approved):

dizzzz:
great thnx

dizzzz (approved):

PR #4481, opened 2022-07-13
Lazy loading model for module imports

### LINKED ISSUE #1848: Adopt lazy loading model for module imports
### What is the problem

As described by @adamretter in the 2018-03-19 Community Call:

> For every query, we compile every module loaded in conf.xml. Modules like the `file` module are already bound. Even if you haven’t bound it in your query, it’s already bound. This is inefficient, slows things down, and prevents users from supplying their own binding to the EXPath File Module.

See also https://github.com/eXist-db/expath-file-module/issues/6, in which the only way to import the EXPath File Module is to comment out the native file module in conf.xml and restart eXist, but having done this, there is no way to use the native file module.

The solution proposed by @wolfgangmm is to move to "lazy loading" of modules.

### What did you expect

I expected to be able to freely override default module namespace prefix bindings set in conf.xml.

1. Install the EXPath File Module xar (built via directions in the README.md file at https://github.com/eXist-db/expath-file-module)
2. Execute the following query:
    ```xquery
    inspect:inspect-module-uri(xs:anyURI("http://expath.org/ns/file"))
    ```
3. Currently, this query produces an error, instead of return a representation of the EXPath File Module. The error:
    > err:XQST0033 Cannot bind prefix 'file' to 'http://expath.org/ns/file' it is already bound to 'http://exist-db.org/xquery/file' [source: xquery version "3.1"; inspect:inspect-module-uri(xs:anyURI("http://expath.org/ns/file"))]

### Context information

- eXist-db version + Git Revision hash: eXist 4.2.0-SNAPSHOT develop e63daf7f0
- Java version: 1.8.0_162-b12
- Operating system: macOS 10.13.4
- 32 or 64 bit: 64 bit

rhubner:
This PR fixes #1848 by allowing the user to configure which modules should be loaded lazily.

It adds a new attribute `load` on the `module` element of `conf.xml`. If unspecified, modules are loaded automatically (i.e. eagerly) to maintain backwards compatibility with the previous eXist-db mode of operation.

The `exist-distribution` config file was updated to take advantage of this new feature. Only those modules that are deemed as required by eXist-db's XQSuite tests and Apps are loaded by default. Other modules can be loaded with `import module namespace`, for example:

```xquery
import module namespace validation = "http://exist-db.org/xquery/validation";
```

----

### DIFF
--- a/exist-core/src/main/java/org/exist/util/Configuration.java
+++ b/exist-core/src/main/java/org/exist/util/Configuration.java
@@ -472,13 +472,16 @@ public class Configuration implements ErrorHandler
                         final Class<?> moduleClass = lookupModuleClass(uri, clazz);
 
                         // Store class if the module class actually exists
-                        if(moduleClass != null && (load == null || BUILT_IN_MODULE_LOAD_ATTRIBUTE_VALUE_EAGER.equals(load))) {
-                            eagerModuleClassMap.put(uri, moduleClass);
-                        } else if(moduleClass != null) {
-                            lazyModuleClassMap.put(uri, moduleClass);
-                        }
-                        if(LOG.isDebugEnabled()) {
-                            LOG.debug("Configured module '{}' implemented in '{}'", uri, clazz);
+                        if (moduleClass != null) {
+                            if (XQueryContext.BUILT_IN_MODULE_LOAD_ATTRIBUTE_VALUE_LAZY.equals(load)) {
+                                lazyModuleClassMap.put(uri, moduleClass);
+                            } else {
+                                eagerModuleClassMap.put(uri, moduleClass);
+                            }
+
+                            if (LOG.isDebugEnabled()) {
+                                LOG.debug("Configured module '{}' implemented in '{}'", uri, clazz);
+                            }
                         }
                     }
 
--- a/exist-core/src/main/java/org/exist/util/Configuration.java
+++ b/exist-core/src/main/java/org/exist/util/Configuration.java
@@ -452,8 +452,8 @@ public class Configuration implements ErrorHandler
                         throw(new DatabaseConfigurationException("element 'module' requires either an attribute " + "'class' or 'src'" ));
                     }
 
-                    if(load != null && !("always".equals(load) || "lazily".equals(load))) {
-                        throw(new DatabaseConfigurationException("parameter 'load' on module can contains only 'always' and 'lazily' values"));
+                    if (load != null && !(XQueryContext.BUILT_IN_MODULE_LOAD_ATTRIBUTE_VALUE_EAGER.equals(load) || XQueryContext.BUILT_IN_MODULE_LOAD_ATTRIBUTE_VALUE_LAZY.equals(load))) {
+                        throw new DatabaseConfigurationException(String.format("parameter '{}' on module can only contain the value: '{}' or '{}'", XQueryContext.BUILT_IN_MODULE_LOAD_ATTRIBUTE, XQueryContext.BUILT_IN_MODULE_LOAD_ATTRIBUTE_VALUE_EAGER, XQueryContext.BUILT_IN_MODULE_LOAD_ATTRIBUTE_VALUE_LAZY));
                     }
 
                     if(source != null) {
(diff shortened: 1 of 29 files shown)


exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -460,10 +469,11 @@ private void loadModuleClasses( Element xquery, Map<String, Class<?>> modulesCla
                         final Class<?> moduleClass = lookupModuleClass(uri, clazz);
 
                         // Store class if thw module class actually exists
-                        if( moduleClass != null) {
-                            modulesClassMap.put(uri, moduleClass);
+                        if( moduleClass != null && (load == null || "always".equals(load))) {
+                            eagerModuleClassMap.put(uri, moduleClass);
+                        }else if(moduleClass != null) {
```
dizzzz, line 474: formatting is not consistent

dizzzz:
Great work! Being non-native, I find the term "lazily" quite hard to remember, I think I see "lazy" more often, is shorter and IMO more clear whilst still correct?

exist-core/src/test/java/org/exist/xquery/XQueryTest.java
```diff
@@ -935,7 +935,7 @@ public void namespaceWithTransform() throws XMLDBException {
 
         String query =
                 "xquery version \"1.0\";\n" +
-                "declare namespace transform=\"http://exist-db.org/xquery/transform\";\n" +
+                "import module namespace transform=\"http://exist-db.org/xquery/transform\";\n" +
```
dizzzz, line 938: interesting change.....

dizzzz:
Great work! @adamretter are we completely sure that the code is backwards compatible? This line might show something different? (not sure)

image

 is this PR a v7 candidate ? will old code still work?

dizzzz (approved):

exist-core/src/main/java/org/exist/util/Configuration.java
adamretter: I pushed a commit to fix this.

adamretter:
> Great work! Being non-native, I find the term "lazily" quite hard to remember, I think I see "lazy" more often, is shorter and IMO more clear whilst still correct?

@dizzzz "lazily" is the adverb (modified a verb), whilst "lazy" is the adjective (describes a state).

In this instance `load="lazily"` - so the verb is "load", and the adverb ("lazily") modifies that.

I think it makes sense, although for sure my English is not perfect either.

We can change it if you still want us to?

joewiz:
I see benefits in standardizing on "lazy". Now the code has instances of both "lazy" and "lazily" (but only "eager" - not "eagerly" - instead "always"). Standardizing on just "lazy" and "eager" (and doing away with "lazily" and "always") would probably prevent people mixing up the different verb forms/words on their code/configurations.

exist-core/src/main/resources/org/exist/test/runner/xquery-test-runner.xq
```diff
@@ -24,6 +24,8 @@ xquery version "3.1";
 import module namespace test = "http://exist-db.org/xquery/xqsuite"
     at "resource:org/exist/xquery/lib/xqsuite/xqsuite.xql";
 
+import module namespace inspect = "http://exist-db.org/xquery/inspection";
```
line-o, line 27: This change is a clear hint that this constitutes a breaking change.
Library and main modules that use one of the functions in the inspect namespace will no longer function until they are imported.

line-o (commented):
If I understand this PR correctly each module with `load=lazy` needs to imported explicitly in any module that wants to use them. Looking at the list where this is applied - like `lucene`, `http-client`, `security-manager` to name a few - this will create a lot of work on any library and app that is doing a little more than showing static content. 
I believe we need to discuss the modules that are lazy-loaded by default.
For this to be merged into 6.x.x my answer would be: none.

adamretter (approved):
LGTM

adamretter:
> Standardizing on just "lazy" and "eager"

@joewiz We have now switched this to `lazy` and `eager`

adamretter:
> This change is a clear hint that this constitutes a breaking change.

@line-o Sure, the target for this is eXist-db 7.0.0.

line-o:
Tests are failing

duncdrum (changes requested):
We should not lazily omit the documentation  that shows the new need for explicit import, that makes this a major release candidate. 

Should also be rebased to get some fresh ci results

Issue #2381, opened 2019-01-10
XQuery parser chokes on legal attribute names


joewiz:
### What is the problem

eXist throws an XPST0003 error if it encounters an direct element constructor with attributes named `copy-namespaces`, `empty-sequence`, or `schema-element`. 

I suspect there may be other cases since these are tokens defined in the XQuery parser. Other tokens that also contain hyphens do not raise this error.

### What did you expect

I expected these legal attribute names to raise no such error.

### Describe how to reproduce or add a test

Open a new query window in eXide or your editor of choice, and paste in any or all of the following elements:

```xml
<foo copy-namespaces="bar"/>,
<foo empty-sequence="bar"/>,
<foo schema-element="bar"/>
```

Submit the query for evaluation, and instead of seeing the literal element returned as expected, you'll see the following error:

> Cannot compile xquery: err:XPST0003 unexpected token: copy-namespaces (while expecting closing tag for element constructor: foo)

Needless to say, BaseX and Saxon do not raise this error.

Full error from exist.log:

```text
2019-01-10 12:30:35,757 [qtp1977003571-110] ERROR (XQueryServlet.java [process]:552) - Cannot compile xquery: err:XPST0003 unexpected token: copy-namespaces (while expecting closing tag for element constructor: foo) [at line 3, column 6] 
org.exist.EXistException: Cannot compile xquery: err:XPST0003 unexpected token: copy-namespaces (while expecting closing tag for element constructor: foo) [at line 3, column 6]
	at org.exist.http.servlets.XQueryServlet.process(XQueryServlet.java:445) [exist-optional.jar:4.6.0-SNAPSHOT]
	at org.exist.http.servlets.XQueryServlet.doPost(XQueryServlet.java:192) [exist-optional.jar:4.6.0-SNAPSHOT]
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:707) [servlet-api-3.1.jar:3.1.0]
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790) [servlet-api-3.1.jar:3.1.0]
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:867) [jetty-servlet-9.4.14.v20181114.jar:9.4.14.v20181114]
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:542) [jetty-servlet-9.4.14.v20181114.jar:9.4.14.v20181114]
[...]

PR #4498, opened 2022-07-19
Add missing XQuery Reserved Names

adamretter:
Closes https://github.com/eXist-db/exist/issues/2381

### DIFF
--- a/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
+++ b/exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
@@ -2224,6 +2224,12 @@ reservedKeywords returns [String name]
 	"map" { name = "map"; }
 	|
 	"array" { name = "array"; }
+	|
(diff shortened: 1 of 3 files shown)


exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g
```diff
@@ -2224,6 +2224,12 @@ reservedKeywords returns [String name]
 	"map" { name = "map"; }
 	|
 	"array" { name = "array"; }
+	|
+	"copy-namespaces" { name = "copy-namespaces"; }
+	|
+    "empty-sequence" { name = "empty-sequence"; }
```
line-o, line 2230: It appears as if the indentation is not correct.

adamretter: @line-o Thanks! I just force pushed an update that fixes that.

dizzzz (approved):

Issue #4537, opened 2022-09-02
[BUG] typed Lucene field values cannot be cast to xs:dateTime  


line-o:
related #4535 

```xquery
xquery version "3.1";

module namespace lfdc="http://exist-db.org/xquery/lucene-field-datetime-cast";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare variable $lfdc:data := document {
<items>
    <item datetime="2022-08-07T10:34:56.789-02:00"/>
    <item datetime="2022-08-07T14:34:56.789+02:00"/>
    <item datetime="2022-08-07T12:34:56.789Z"/>
    <item datetime="2022-08-07T12:34:56.789"/>
</items>
};

declare variable $lfdc:xconf :=
<collection xmlns="http://exist-db.org/collection-config/1.0">
    <index xmlns:xs="http://www.w3.org/2001/XMLSchema">
        <lucene>
            <text qname="item">
                <field name="time" expression="@time" type="xs:time"/>
            </text>
        </lucene>
    </index>
</collection>;

declare variable $lfdc:collection := "lucene-field-datetime-cast";

declare
    %test:setUp
function lfdc:setup() {
    let $testCol := xmldb:create-collection("/db", $lfdc:collection)
    let $indexCol := xmldb:create-collection("/db/system/config/db", $lfdc:collection)
    return (
        xmldb:store("/db/" || $lfdc:collection, "test.xml", $lftn:data),
        xmldb:store("/db/system/config/db/" || $lfdc:collection, "collection.xconf", $lfdc:xconf),
        xmldb:reindex("/db/" || $lfdc:collection)
    )
};

declare
(:    %test:tearDown:)
function lfdc:tearDown() {
    xmldb:remove("/db/" || $lfdc:collection),
    xmldb:remove("/db/system/config/db/" || $lfdc:collection)
};

declare
    %test:assertEquals("2022-08-07T12:34:56.789Z", "2022-08-07T12:34:56.789Z", "2022-08-07T12:34:56.789Z", "2022-08-07T12:34:56.789Z")
function lfdc:datetime-values-cast() {
    collection("/db/" || $lfdc:collection)
        //item[
            ft:query(., "datetime:*", map{ 
                "leading-wildcard": "yes",
                "fields": "datetime"
            })]
        ! ft:field(., "datetime", "xs:dateTime")
};
```

**Tested on:**
 - OS: MacOS 12.5.1
 - eXist-db version: 6.1.0-SNAPSHOT
[...]

Issue #4536, opened 2022-09-02
[BUG] not all timestrings are normalised to UTC in typed Lucene field


line-o:
```xquery
xquery version "3.1";

module namespace lftn="http://exist-db.org/xquery/lucene-field-time-normalization";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare variable $lftn:data := document {
<items>
    <item time="10:34:56.789-02:00"/>
    <item time="14:34:56.789+02:00"/>
    <item time="12:34:56.789Z"/>
<!--
    time-strings without a timezone are normalized
    against the current server timezone which can
    be different from UTC!
    this is a separate issue
-->
    <item time="12:34:56.789"/>
</items>
};

declare variable $lftn:xconf :=
<collection xmlns="http://exist-db.org/collection-config/1.0">
    <index xmlns:xs="http://www.w3.org/2001/XMLSchema">
        <lucene>
            <text qname="item">
                <field name="time" expression="@time" type="xs:time"/>
            </text>
        </lucene>
    </index>
</collection>;

declare variable $lftn:collection := "lucene-field-time-normalization";

declare
    %test:setUp
function lfttt:setup() {
    let $testCol := xmldb:create-collection("/db", $lftn:collection)
    let $indexCol := xmldb:create-collection("/db/system/config/db", $lftn:collection)
    return (
        xmldb:store("/db/" || $lftn:collection, "test.xml", $lftn:data),
        xmldb:store("/db/system/config/db/" || $lftn:collection, "collection.xconf", $lftn:xconf),
        xmldb:reindex("/db/" || $lftn:collection)
    )
};

declare
(:    %test:tearDown:)
function lftn:tearDown() {
    xmldb:remove("/db/" || $lftn:collection),
    xmldb:remove("/db/system/config/db/" || $lftn:collection)
};

declare
    %test:assertEquals("12:34:56.789Z", "12:34:56.789Z", "12:34:56.789Z", "12:34:56.789Z")
function lftn:time-values-normalized() {
    collection("/db/" || $lfttt:collection)
        //item[
            ft:query(., "time:*", map{ 
                "leading-wildcard": "yes",
                "fields": "time"
            })]
        ! ft:field(., "time", "xs:time")
};

```

**Tested on:**
 - OS: MacOS 12.5.1
 - eXist-db version: 6.1.0-SNAPSHOT
 - Java Version Java8u342

**Additional context**
[...]

Issue #4535, opened 2022-09-02
xs:time and xs:dateTime stored in Lucene field will be normalised to UTC


line-o:
**Describe the bug**

When one configures a Lucene Field index to be of type `xs:dateTime` or `xs:time` the timezone is normalised to UTC before the value is added to the index. Two entries that mark the same moment in time but in different timezones have a frequency of two in the index.
I believe this is not a bug, but more a design decision that needs to be documented. But since it is undocumented developers might expect a different behaviour.

My tests whether **sorting** of xs:time takes the timezone into account were inconclusive. It appears that only the (normalised) time portion is used. That would mean that the sorting of field values is comparable and not affected by the current implementation.

Since the timezone information is gone upon retrieval it it is impossible to **search** for, **filter** or **group by** timezones with only the values stored in the index.

**NOTE:** The original information can still be retrieved from the hit itself.

**Expected behavior**

The fact that values typed as `xs:time` or `xs:dateTime` are normalised to UTC needs to be documented.
It could be worth to add an example how to store timezone information in a separate field or facet to enable search, filtering and grouping.

Alternatively, a way to efficiently store and retrieve the timezone should be investigated.

**To Reproduce**

```xquery
xquery version "3.1";

module namespace lfttt="http://exist-db.org/xquery/lucene-field-time-timezone-test";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare variable $lfttt:data := document {
<items>
    <item datetime="2022-08-07T10:34:56.789-02:00"/>
    <item datetime="2022-08-07T14:34:56.789+02:00"/>
    <item datetime="2022-08-07T12:34:56.789Z"/>
    <item datetime="2022-08-07T12:34:56.789"/>
    <item time="10:34:56.789-02:00"/>
    <item time="14:34:56.789+02:00"/>
    <item time="12:34:56.789Z"/>
<!--
    time-strings without a timezone are normalized
    against the current server timezone which can
    be different from UTC!
    this is a separate issue
    <item time="12:34:56.789"/>
-->
</items>
};

declare variable $lfttt:xconf :=
<collection xmlns="http://exist-db.org/collection-config/1.0">
    <index xmlns:xs="http://www.w3.org/2001/XMLSchema">
        <lucene>
            <text qname="item">
                <field name="datetime" expression="@datetime" type="xs:dateTime"/>
                <field name="time" expression="@time" type="xs:time"/>
            </text>
        </lucene>
    </index>
</collection>;

[...]

PR #4541, opened 2022-09-05
Field perf revised

line-o:
revised version of #4253

This PR will address 3 problems when using Lucene fields for sorting or iterating large node sets (> 100.000 elements):

1. eXist-db quickly runs out of memory: this happens as `ft:query` retrieves the fields for the entire result set at once and keeps them attached to each item in the set, no matter if the fields content is in fact accessed or not
2. Because fields are retrieved at query time, users have to know in advance which fields will be required later and list them in the `ft:query` options
3. Fixing 1 and 2 will further slow down field-based sort operations on large node sets. In general, those operations were already slow, but performance will degrade even more if we do not preload the fields.

The PR effectively fixes 1 by not preloading any fields. Instead, the loading will be deferred until `ft:field` is called. As a consequence, users no longer need to specify the fields to retrieve at query time, so 2 is solved as well (this doesn't change the API: you can still define fields in ft:query options, though the setting is unnecessary and will be ignored).

To address 3, an additional configuration attribute is introduced to field definitions in `collection.xconf`: adding `binary="yes"` will result in the field to be stored as a `BinaryDocValue`, which guarantees the fastest possible access in Lucene. This follows the example of other Lucene-based implementations (e.g. elastic search), which provide a similar option. Binary fields can only be retrieved, but not queried like other fields.

~~For strings and simple numeric types,~~ performance should be outstanding. ~~A bottleneck remains for complex data types though, in particular xs:date and xs:dateTime. Here the overhead to cast the stored binary field into a proper XQuery xs:date is always considerable, though not to be avoided. We're still trying to find a way to address this.~~

Connected documentation PR: https://github.com/eXist-db/documentation/pull/760

This PR adds some missing atomic types to typed lucene fields (`xs:decimal`, `xs:float`, `xs:short` and more).
Additional xqsuite tests would be reasonable to add to test for these new types (see also #4537).

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/ValueIndexFactory.java
+++ b/exist-core/src/main/java/org/exist/storage/ValueIndexFactory.java
@@ -127,6 +127,7 @@ public class ValueIndexFactory {
         return serialize(value, offset, true);
     }
 
+    // TODO(AR) switch implementation to various serialize methods in the AtomicValues (requires major version bump)
     public final static byte[] serialize(final Indexable value, final int offset, final boolean caseSensitive) throws EXistException {
         /* xs:string */
         if (Type.subTypeOf(value.getType(), Type.STRING)) {
--- a/exist-core/src/main/java/org/exist/util/ByteConversion.java
+++ b/exist-core/src/main/java/org/exist/util/ByteConversion.java
@@ -21,6 +21,8 @@
  */
 package org.exist.util;
 
+import java.nio.ByteBuffer;
+
 /**
  * A collection of static methods to write integer values from/to a
  * byte array.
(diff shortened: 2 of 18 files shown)


dizzzz (approved):

line-o:
To test binary compatibility I followed the instructions to upgrade http://exist-db.org/exist/apps/doc/upgrading.xml?field=all&id=D3.15#D3.15
I copied a data folder with Lucene Indexes in various formats over to an instance built with this PR applied.
All Lucene fields including typed ones and full-text indexes were still present and the values equal. The values were also castable back to the corresponding XDM value types.

joewiz:
Based on @line-o's findings and the passing CI tests, I believe this PR is ready to and should be merged into develop, for inclusion in a forthcoming minor release (6.1.0). It contains @wolfgangmm and @line-o's original vital fix to a memory usage bug and adds a binary fields option for enhanced performance, and incorporates @adamretter's addition of support for more atomic types to typed lucene fields (`xs:decimal`, `xs:float`, `xs:short`, etc.). After extensive review, @line-o confirmed that all data truncation issues identified by @adamretter in #4253 pre-existed, and @line-o analyzed and prepared new issues for each of these problems (as well as some others identified in the process): 

- https://github.com/eXist-db/exist/issues/4532
- https://github.com/eXist-db/exist/issues/4533
- https://github.com/eXist-db/exist/issues/4534
- https://github.com/eXist-db/exist/issues/4535
- https://github.com/eXist-db/exist/issues/4536
- https://github.com/eXist-db/exist/issues/4537

Thus, I think we now have a comprehensive understanding of the impact of this PR and the other issues uncovered in the process of review.

dizzzz:
discussed with Patrick pulled in; there is no incompatibility as JL has tested the backward compatibility well.

Issue #4542, opened 2022-09-06
[BUG] regression in fn:unparsed-text-lines and fn:unparsed-text-available 


line-o:
**Describe the bug**

CI shows hundreds of failing core tests (this is likely related to #4538). I can reproduce those test cases on my local machine. After some investigation it turns out that setting just two of them to pending allows the build to pass again.

https://github.com/eXist-db/exist/blob/f3424d7c44d2cc7d8eb722e0701709de6e497a34/exist-core/src/test/xquery/unparsed-text.xql#L151-L155

https://github.com/eXist-db/exist/blob/f3424d7c44d2cc7d8eb722e0701709de6e497a34/exist-core/src/test/xquery/unparsed-text.xql#L232-L236

I then executed the code in eXide and discovered that

`fn:unparsed-text-lines` returns an empty string instead of throwing an error
and `fn:unparsed-text-available` returns true instead of false.

Running `curl -v http://www.w3.org/fots/unparsed-text/does-not-exist.txt` reveals that the server responds with a permanent redirect to https://www.w3.org/fots/unparsed-text/does-not-exist.txt

`curl -v https://www.w3.org/fots/unparsed-text/does-not-exist.txt` then returns the 404 that the two tests seem to rely on.

**Expected behavior**

The tests to work as expected or adapted to reflect changed expected behaviour.
It might also be the response of the URL the test case queries has changed.

We can and should change the URL for the above tests to the new location. But it is equally important to have a test that tests against a URL that responds with a redirect.
It should fail either because we will not follow redirects and treat status codes in the 3XX range as errors
- or - it should fail because the redirect is followed and the server responds with 404.  

Related: https://github.com/qt4cg/qtspecs/issues/132

**To Reproduce**

Run `mvn clean compile test` on your local machine.

**Tested on:**
 - OS: MacOS 12.5.1
 - eXist-db version: 6.1.0-SNAPSHOT
 - Java Version Java8u342

**Additional context**
- How is eXist-db installed? built from source: commit f3424d7c44d2cc7d8eb722e0701709de6e497a34 (develop HEAD)
- Any custom changes in e.g. `conf.xml`? none

duncdrum:
@line-o the patch was merged. Can we close this, and open a new issue for what is left to do.

line-o:
I agree this is done. We could open a follow-up issue on redirect handling but it is definitely a separate issue.

PR #4543, opened 2022-09-06
[hotfix] change URL for two unparsed-text tests

line-o:
refs #4542

This is just a quick patch!

Two pending tests were added `upt:non-existent-redirect` and `upt:unparsed-text-available-non-existent-redirect` which will pass when the redirect is followed _or_ an HTTP status code in the 3XX range is treated as an error.

### DIFF
--- a/exist-core/src/test/xquery/unparsed-text.xql
+++ b/exist-core/src/test/xquery/unparsed-text.xql
@@ -150,7 +150,20 @@ function upt:invalid-uri2() {
 
 declare
     %test:assertError("FOUT1170")
-function upt:non-existant() {
+function upt:non-existent() {
+    unparsed-text-lines("https://www.w3.org/fots/unparsed-text/does-not-exist.txt")
+};
+
+(:~
+ : this test will fail as the server responds with a
+ : permanent redirect that is neither followed nor
+ : treated as an error
+ : @see https://github.com/eXist-db/exist/issues/4542
+ :)
+declare
+    %test:pending
+    %test:assertError("FOUT1170")
+function upt:non-existent-redirect() {
     unparsed-text-lines("http://www.w3.org/fots/unparsed-text/does-not-exist.txt")
 };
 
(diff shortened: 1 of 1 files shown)


dizzzz (approved):

line-o:
Tests running on Windows with JDK 11 remain flaky. They rarely pass. This warrants an issue of its own.

adamretter:
Would it be better to use local files rather than remote? Otherwise we will likely have this issue again...

line-o:
Again? When they are redirected another tule you mean?
We should definitely use the time to get redirect handling fixed.

adamretter:
> Again? When they are redirected another tule you mean?
I am not sure what you mean by this?

I mean to use a local file instead of something remote, e.g. `file:///`

line-o:
The test tests handling of network errors. Specifically HTTP status code 404

dizzzz:
Again I am confused in the discussion :-)

wolfgangmm:
How would testing a redirect work with local files?

adamretter:
@wolfgangmm You can't of course test redirects with the `file:///` scheme, but you can with `http://localhost:8080`

line-o:
I am still in favour of merging this now. Whatever improvements can be made can be made later. Builds are failing and we are blind. Especially changes in extensions are not being tested.

line-o:
Testing redirects is a followup task. This is just to reinstate the 404 handling tests.

adamretter:
@line-o Seems reasonable. Should these not also have `%test:assumeInternetAccess` annotations?

line-o:
As the annotation does not do anything at the moment I do not see this as important right now.

adamretter:
@line-o I think it is important as it is there to try and handle such cases, I understand that at the moment it is nothing more than a label, but that label does:
1.  allow us to easily find such test cases again in future.
2. give us the option of controlling such tests in future.

If you would not mind adding it please, then I am happy to merge your PR

PR #4545, opened 2022-09-07
[bugfix] - namespace binding conflict errors

alanpaxton:
### Description:

Report an error (`XUDY0023`) when an insertion is attempted with a node or attribute name in a namespace that conflicts with one which exists in the document.

### References:

Closes https://github.com/eXist-db/exist/issues/1482

See https://www.w3.org/TR/xquery-update-30/#dt-conflict

Recursively check node/attributes inserted into a document for namespace conflicts. This change does not look at namespaces introduced in the subtrees being inserted, it appears these correctly respect the namespace in the subtree rather than the declared namespace in the XQuery context in which the insertion happens.

### Type of tests:

Added unit tests in `binidingConflict.xqm` for correct reporting of binding conflict errors

### DIFF
--- a/exist-core/src/main/java/org/exist/dom/memtree/NodeImpl.java
+++ b/exist-core/src/main/java/org/exist/dom/memtree/NodeImpl.java
@@ -48,7 +48,7 @@ import java.util.Iterator;
 import java.util.Properties;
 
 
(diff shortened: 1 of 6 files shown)


adamretter (approved):
LGTM

Issue #2056, opened 2018-07-21
HTML5 doctype is removed when saving, editing or serving static HTML5 files


metacontext:
### What is the problem
When importing / saving static HTML5 files (e.g. via the collection browser or oXygen) the doctype gets removed. When editing those files in eXide the doctype is not present and also cannot be included manually (inserting and saving has no effect). Serving those files to the browser results in an XML-prologue being inserted instead. When exporting the files the former HTML5 doctype is also not present anymore.

Interesstingly, when generating and saving static HTML5 files via the serialize() function and xmldb:store() everything works as expected.

### What did you expect
eXist should not change imported static HTML5 files and always keep the HTML5 doctype

### Describe how to reproduce or add a test
Import any valid static HTML5 file in eXist with the collection browser and open it in eXide.

### Context information
- eXist-db version: all versions from 3.2.0 onwards
- Java 8 171 on Mac, Oracle java "1.8.0_181" on Debian
- Linux and Mac
- 64 bit
- JAR and DMG

joewiz:
See the mailing list discussion of the doctype issue from 2009 at https://markmail.org/thread/27fxcjicang6yk3q, and the recent discussion of this issue peeling apart the relevant layers at https://github.com/eXist-db/eXide/issues/198.

adamretter:
I would love to get to the bottom of the Doctype handling stuff and fix it once and for all; It seems from what I have read, that it hasn't been properly implemented since at least 2009, which is a bit bad really :-/

I want to fix the issues in eXist-db and it's API's and then App developers can fix anything they need themselves.

As I understand it, if you store a document into eXist-db, the Doctype declaration (or at least the `ExternalID` (public/private parts) of it) will be captured and stored. Most of the discussion in the mailing-list thread that @joewiz referenced (https://markmail.org/thread/27fxcjicang6yk3q) seemed to be around serialization.

I think we need to do a few things:

1. Store the entire Doctype declaration not just the ExternalID. As described by - https://www.w3.org/TR/xml/#dtd

2. Fix the Serializer (and possibly also the Indexer?).
    In @wolfgangmm's reply to thread of Oct 21st 2009, he states:
    > The serializer ignores the doctype declaration by default. Reason: we 
cannot be sure that the doctype declaration would be correct and 
complete. The original XML document may have declared an internal 
subset, which the indexer could not see and thus not store.

[...]

PR #4547, opened 2022-09-07
Enable Document Type Declaration Serialization

adamretter:
This PR now correctly enables the serialisation of Document Type Declarations if one is present in a document. Previously this behaviour was inconsistent across eXist-db, some APIs would serialise the DocType which others would not.

This has been implemented and tested for the following APIs:
* HTTP RESTServer
* XML-RPC
* XML:DB Local and Remote
* WebDAV

As part of the document, the Document Type Declaration is now visible and modifiable from:
* eXist-db Java Admin Client
* Oxygen XML Editor
* eXide

Closes https://github.com/eXist-db/exist/issues/2056
Related to https://github.com/eXist-db/eXide/issues/198


joewiz:
Is this option now available in serialization option declarations in query prologs? If so, should an entry be added under https://exist-db.org/exist/apps/doc/xquery#legacy-serialization?

As to the question of classifying the changes in this PR, recall that in a related thread on exist-open, https://markmail.org/message/w4g4kt3o4gh5dvy6, @wolfgangmm noted that eXist was already storing doctype information, but wasn't serializing it. This PR's addition of doctype serialization and enabling it by default in all APIs strikes me as something between a bugfix and a comprehensively implemented new feature—providing something that users have often expected, requested, and struggled to overcome. I have struggled to imagine any process that would *break* if a doctype were supplied rather than omitted. This would argue in favor of including it in 6.x, rather than requiring a new major version.

dizzzz (approved):

adamretter:
> Is this option now available in serialization option declarations in query prologs? If so, should an entry be added under https://exist-db.org/exist/apps/doc/xquery#legacy-serialization?

Good idea - I have opened a PR for that here - https://github.com/eXist-db/documentation/pull/829

> that users have often expected, requested, and struggled to overcome

That was indeed my own experience recently!

It seems to me that people confuse XML documents and their serialization, with XDM Serialization. As I understand it XML documents can have Doctype Declarations with no issue, whereas XDM has no concept of a Doctype Declaration, and Serialization of an XDM (i.e the result of an XQuery) only allows you to statically declare a doctype as a serialization option (unless using `fn:serialize` within an XDM).

This PR focuses on XML serialization through eXist-db's various APIs.

Additionally it does add an (eXist-db specific) serialization option for XDM via `fn:serialize` to output the DocType if you happen to be serializing a persistent document from the database, otherwise that option has no effect.

> This would argue in favor of including it in 6.x, rather than requiring a new major version.

Sounds good to me :-)

exist-core/src/main/java/org/exist/client/InteractiveClient.java
```diff
@@ -137,6 +138,7 @@ public class InteractiveClient {
         defaultProps.setProperty(USER, USER_DEFAULT);
         defaultProps.setProperty(EDITOR, EDIT_CMD);
         defaultProps.setProperty(INDENT, "true");
+        defaultProps.setProperty(OUTPUT_DOCTYPE, "true");
```
line-o, line 141: Could this break existing applications or was the default to output doctype declarations true by default before?

line-o (approved):
The changes do look very good and I am looking forward to when this is pulled in.
Having written that, I am curious if this might be a breaking change or not.

Issue #4546, opened 2022-09-07
[BUG] Unable to navigate ancestor axis to document node


yamahito:
**Describe the bug**
When working with a sequence of nodes returned by collection(), the document-node() is absent from the ancestor axis.

**Expected behavior**
Expected behaviour is to be able to access the ancestor document node.

**To Reproduce**

Create two functions in a module:

```Xquery
module namespace mod1 = "http://mod1";

declare function mod1:doc-node($node as node()) as document-node() {
    $node => document-uri() => doc()
};

declare function mod1:doc-node2($node as node()) as document-node() {
    $node/ancestor-or-self::document-node()
};
```

and call each of the functions for an element in the database:

```xquery
import module namespace mod1="http://mod1" at "/db/mod1.xqm";

collection($some-collection)[1] => mod1:doc-node(),
collection($some-collection)[1] => mod1:doc-node2()

```

The first function runs successfully.
The second will fail with a type error as the empty sequence is returned rather than the required document node.

**Context (please always complete the following information):**
 - OS: ubuntu
 - eXist-db version: 6.10-SNAPSHOT

**Additional context**
- How is eXist-db installed? existdb/existdb:latest docker image

adamretter:
I think this can be simplified down to some one line expressions, by removing the functions and arrow operators we can reduce the complexity. I have stripped this back to the following test-cases:

1. Should return an empty result (works correctly):
```xquery
collection("/db")[1]/document-node()
```

2. Should return a non-empty result (works correctly):
```xquery
collection("/db")[1]/doc(document-uri(.))
```

3. Should return a non-empty result (works correctly):
```xquery
collection("/db")[1]/self::document-node()
```

4. Should return a non-empty result (**Fails**):
```xquery
collection("/db")[1]/ancestor-or-self::document-node()
```

@yamahito So if I have capture your test correctly and isolated it, then I think the problem lies solely with the `ancestor-or-self` Path Step.

adamretter:
@yamahito I also found more based on your original enquiry in slack.

Aloo if these require `declare namespace config = "http://exist-db.org/collection-config/1.0";`; I have omitted it below to make it clearer what the issues are.

1. Should return a non-empty result (works correctly):
```xquery
(collection("/db/system/config/db")/config:collection)[1]/parent::node()
```
2. Should return an empty result (works correctly):
```xquery
(collection("/db/system/config/db")/config:collection)[1]/parent::*
```

3. Should return a non-empty result (**Fails**):
[...]

PR #4548, opened 2022-09-07
Fix issues with document-node lookups on various axis

adamretter:
Appears to have been caused by over zealous optimisations.
The fixes should not impact general performance at all.

Closes https://github.com/eXist-db/exist/issues/4546

### DIFF
--- a/exist-core/src/test/xquery/document-nodes.xq
+++ b/exist-core/src/test/xquery/document-nodes.xq
@@ -115,3 +115,75 @@ declare
 function dn:memtree-document-node-element-wrong-name() {
     document { element template {} } instance of document-node(element(wrong))
 };
+
+declare
+    %test:assertEmpty
+function dn:persistent-document-node-from-collection() {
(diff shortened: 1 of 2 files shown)


dizzzz (approved):

exist-core/src/test/xquery/document-nodes.xq
```diff
@@ -115,3 +115,75 @@ declare
...
+};
+
+declare
+    %test:assertExists
+function dn:persistent-document-node-from-collection-via-parent-axis-node() {
+    (collection($dn:TEST_COLLECTION)/template)[1]/parent::node()
+};
+
+declare
+    %test:assertEmpty
```
line-o, line 150: the test name suggest the axis should return something

line-o (approved):
I think this is good. 👍🏼

exist-core/src/test/xquery/document-nodes.xq
line-o: Any chance to get an answer on this @adamretter ?

PR #4563, opened 2022-09-19
[enhancement] Use final log4j-slf4j2-impl bridge

reinhapa:
Switch from alpha version to final Log4j bridge implementation.


line-o (approved):
Thanks!

dizzzz (approved):

PR #4564, opened 2022-09-19
Migrate to java11

dizzzz:
In an ambition to default to Java11 for eXist-db, changes are made in the build system.

According to @reinhapa  eg due to much more efficient String handling, performance could increase significantly.

this PR is (now) to test GHA.

Java8 tests are expected to fail.... or should not run at all?

### DIFF
--- a/.github/workflows/ci-deploy.yml
+++ b/.github/workflows/ci-deploy.yml
@@ -10,11 +10,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
-      - name: Set up JDK 8
+      - name: Set up JDK 11
         uses: actions/setup-java@v2
         with:
           distribution: temurin
-          java-version: '8'
+          java-version: '11'
       - name: Make buildkit default
         uses: docker/setup-buildx-action@v1
         id: buildx
--- a/.github/workflows/ci-docs.yml
+++ b/.github/workflows/ci-docs.yml
@@ -8,7 +8,7 @@ jobs:
     name: ${{ matrix.jdk }} Javadocs 
     strategy:
       matrix: 
-        jdk: ['8','11', '16']
+        jdk: ['11', '16']
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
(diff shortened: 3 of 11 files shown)


reinhapa (approved):
LGTM

dizzzz:
hmmm 
image

line-o:
The test execution time on Linux is quite good compared to java8, no?

duncdrum:
the docker JVM also needs to be switched to `J11` see the failing `tests images` steps

dizzzz:
@dependabot rebase

dizzzz:
@duncdrum  where needs that to be done?

duncdrum:
@dizzzz I'm doing it like this https://github.com/duncdrum/distroless-exist/blob/25338457a7902fb34d0010417bf50613b662c0d2/Dockerfile#L24 but that wasn't desirable here, so you need to check back with @adamretter

line-o:
@dizzzz I would favour you rebase on to current develop head instead of creating unnecessary merges.

dizzzz:
@line-o I actually never rebase code :-( I think I don't even know how to do so; I think I was once asked to do this way, and... I remember that rebasing can have (for me) unclear side effects? Can you teach me other wise, and avoid/understand these unwanted side effects?

line-o:
Sure! Screen share Monday before the community call?

dizzzz:
yikes the following in exist-parent.xml is faking the build results: 
image

dizzzz:
[patch.diff.txt](https://github.com/eXist-db/exist/files/9639473/patch.diff.txt)

dizzzz:
hmmm errors with Java16 but the run looks OK :-( 
image

line-o:
@dizzzz Just copying the docker file from @duncdrum did not seem to do the trick. The COPY operations refer to some folder `dump` that does not exist in our setup.

reinhapa:
> hmmm errors with Java16 but the run looks OK :-( image

@dizzzz It seems the following JVM Flag needs to be added `--add-opens=java.base/java.lang.reflect=ALL-UNNAMED` in order to run properly in that test environment.

line-o:
We should discuss the way we are going forward in light of #4574

reinhapa:
@dizzzz seems as @adamretter has already started a draft for JDK 17 with Jetty 10 here: #4575

line-o:
@reinhapa jetty 11 even

reinhapa:
> We should discuss the way we are going forward in light of #4574

For me JDK 17 would also be a way to go...

line-o:
I added this to next weeks agenda @reinhapa

PR #4570, opened 2022-09-22
[bugfix] fn:transform stylesheet document crash

alanpaxton:
### Description

The fn:transform code to find the path to the node within the document, for hashing the stylesheet cache, didn’t expect a document. Test for that case explicitly, and in the case of a document the URI of the document can serve as the path.

### Issue

Passing an eXist stored document as the stylesheet node to `fn:transform` fails with a ClassCastException.
`{ ... "stylesheet-node":doc("/db/regression-test-1/stylesheet.xsl") ... }`

### Tests

An XSuite regression unit test is included in this checkin.

### DIFF
--- a/exist-core/src/test/xquery/xquery3/transform/fnTransformRegression1.xqm
+++ b/exist-core/src/test/xquery/xquery3/transform/fnTransformRegression1.xqm
@@ -1,0 +1,60 @@
+(:
+ : eXist-db Open Source Native XML Database
+ : Copyright (C) 2001 The eXist-db Authors
+ :
+ : info@exist-db.org
+ : http://www.exist-db.org
+ :
+ : This library is free software; you can redistribute it and/or
+ : modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


exist-core/src/test/xquery/xquery3/transform/fnTransformRegression1.xqm
```diff
@@ -0,0 +1,60 @@
...
+};
+
+declare
+    %test:assertEquals("<v>2</v>")
+function testTransform:regression-test-1() {
+    let $in := parse-xml("<dummy/>")
+    let $result := ( fn:transform(map{
+        "source-node":$in,
+        "stylesheet-node":doc("/db/regression-test-1/stylesheet.xsl"),
+        "stylesheet-params": map { QName("","v"): "2" } } ) )?output
```
line-o, line 57: I switched to using `xs:QName("v")` instead.

line-o (approved):
LGTM (wether or not the QName construction is changed is less important than to fix this issue)

PR #4573, opened 2022-09-26
Should never synchronize on a Boolean

adamretter:
A Boolean only has two static instances True or False, and they are different instances depending on the value, and so synchronising on a Boolean actually leads to synchronising on one or two global instances which is not what was intended.

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/BrokerPool.java
+++ b/exist-core/src/main/java/org/exist/storage/BrokerPool.java
@@ -24,7 +24,6 @@ package org.exist.storage;
 import com.evolvedbinary.j8fu.fsm.AtomicFSM;
 import com.evolvedbinary.j8fu.fsm.FSM;
 import com.evolvedbinary.j8fu.lazy.AtomicLazyVal;
-import net.jcip.annotations.GuardedBy;
 import net.jcip.annotations.ThreadSafe;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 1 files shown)


exist-core/src/main/java/org/exist/storage/BrokerPool.java
```diff
@@ -975,29 +974,25 @@ public NotificationService getNotificationService() {
...
     public boolean isReadOnly() {
-        synchronized(readOnly) {
-            if(!readOnly) {
-                final long freeSpace = FileUtils.measureFileStore(dataLock.getFile(), FileStore::getUsableSpace);
-                if (freeSpace != -1 && freeSpace < diskSpaceMin) {
-                    LOG.fatal("Partition containing DATA_DIR: {} is running out of disk space [minimum: {} free: {}]. Switching eXist-db into read-only mode to prevent data loss!", dataLock.getFile().toAbsolutePath().toString(), diskSpaceMin, freeSpace);
-                    setReadOnly();
-                }
+        final boolean isReadOnly = readOnly.get();
+        if (!isReadOnly) {
```
reinhapa, line 983: I think this local variable is only used here and can be inlined

reinhapa (approved):

exist-core/src/main/java/org/exist/storage/BrokerPool.java
adamretter: I felt it was clearer to read in the way I formulated it. How strong is your objection?

dizzzz: for debugging is is more clear; I'd not care

dizzzz (approved):

exist-core/src/main/java/org/exist/storage/BrokerPool.java
reinhapa: @adamretter just a hint, not a requirement..

PR #4574, opened 2022-09-26
Upgrade to Java 17

adamretter:
This updates eXist-db to Java 17 as a minimum LTS version.

It does not introduce Java9+ style modules.

* It does pass all eXist-db tests.
* It does also update the Docker images.
* It does also switch to a new GC - Z1.


adamretter:
@dizzzz Apologies I only just saw your PR for updating to Java 11 now.
I undertook this work as I need access to the new SAX APIs that are only available in Java14+, and as Java 17 is the next LTS after 11, I figured going from 14 to 17 make sense.

.idea/runConfigurations/Java_Admin_Client.xml
```diff
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Java Admin Client" type="Application" factoryName="Application">
-    <option name="ALTERNATIVE_JRE_PATH" value="1.8" />
+    <option name="ALTERNATIVE_JRE_PATH" value="liberica-17" />
```
reinhapa, line 3: Maybe choose a more general name here...

.idea/runConfigurations/Jetty_Server.xml
```diff
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Jetty Server" type="Application" factoryName="Application">
-    <option name="ALTERNATIVE_JRE_PATH" value="1.8" />
+    <option name="ALTERNATIVE_JRE_PATH" value="liberica-17" />
```
reinhapa, line 3: Maybe choose a more general name here...

.idea/runConfigurations/Local_Jetty_Server.xml
```diff
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Local Jetty Server" type="Application" factoryName="Application">
-    <option name="ALTERNATIVE_JRE_PATH" value="1.8" />
+    <option name="ALTERNATIVE_JRE_PATH" value="liberica-17" />
```
reinhapa, line 3: Maybe choose a more general name here...

exist-parent/pom.xml
```diff
@@ -762,7 +762,7 @@
                               However it can make it hard to diagnose problems if tests leak state; If you experience
                               such a problem you may want to set it to `false` whilst debugging -->
                         <reuseForks>true</reuseForks>
-                        <argLine>@{jacocoArgLine} -XX:+IgnoreUnrecognizedVMOptions --illegal-access=debug --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.ref=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED -XX:StartFlightRecording=maxsize=5g,disk=true,dumponexit=true,settings=default,filename=${project.build.directory}/ -Dfile.encoding=${project.build.sourceEncoding}</argLine>
+                        <argLine>@{jacocoArgLine} -XX:+IgnoreUnrecognizedVMOptions --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.ref=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED -XX:StartFlightRecording=maxsize=5g,disk=true,dumponexit=true,settings=default,filename=${project.build.directory}/ -Dfile.encoding=${project.build.sourceEncoding}</argLine>
```
reinhapa, line 765: For JDK 11 and later the `-XX:+IgnoreUnrecognizedVMOptions` argument an also be dropped, this was inserted, as JDK 8 did not support `--illegal-access=debug` and `--add-opens=xxx`

reinhapa (changes requested):
Why are those casts from `ByteBuffer` to `Buffer` as `ByteBuffer` already extends from it? I'm a bit confused here. This would reduce the number of changed files significantly though:

```
exist-core/src/main/java/org/exist/storage/blob/BlobStoreDumpTool.java
exist-core/src/main/java/org/exist/storage/blob/BlobStoreImpl.java
exist-core/src/main/java/org/exist/storage/journal/Journal.java
exist-core/src/main/java/org/exist/storage/journal/JournalReader.java
exist-core/src/main/java/org/exist/storage/lock/FileLock.java
exist-core/src/main/java/org/exist/storage/statistics/DataGuide.java
exist-core/src/main/java/org/exist/util/io/ByteBufferInputStream.java
exist-core/src/main/java/org/exist/util/io/MemoryMappedFileFilterInputStreamCache.java
exist-core/src/main/java/org/exist/xquery/value/BinaryValueFromFile.java
exist-core/src/test/java/org/exist/storage/journal/JournalTest.java
exist-core/src/test/java/org/exist/storage/ValueIndexFactoryTest.java
extensions/debuggee/src/main/java/org/exist/debuggee/dbgp/ResponseEncoder.java
extensions/debuggee/src/main/java/org/exist/debugger/dbgp/ResponseEncoder.java
```

adamretter:
> Why are those casts from ByteBuffer to Buffer as ByteBuffer already extends from it? I'm a bit confused here. This would reduce the number of changed files significantly though:

Yes it puzzled me too! However, without them however we get `MethodNotFound` exceptions at runtime. From what I have read this is just "needed"

.idea/runConfigurations/Java_Admin_Client.xml
adamretter: I have pushed a commit to address this.

.idea/runConfigurations/Local_Jetty_Server.xml
adamretter: I have pushed a commit to address this.

.idea/runConfigurations/Jetty_Server.xml
adamretter: I have pushed a commit to address this.

exist-parent/pom.xml
adamretter: I have pushed a commit to address this.

exist-core/src/main/java/org/exist/source/SourceFactory.java
```diff
@@ -152,7 +152,7 @@ private static String firstPathSegment(final String path) {
     }
 
     private static Source getSource_fromClasspath(final String contextPath, final String location) throws IOException {
-        if (location.startsWith(ClassLoaderSource.PROTOCOL)) {
+        if (contextPath == null || location.startsWith(ClassLoaderSource.PROTOCOL)) {
```
line-o, line 155: It appears to me as if this case should be handled further up in `getSource` or a separate method `getSource_fromLocation`

line-o: `if (location.startsWith(ClassLoaderSource.PROTOCOL))` is duplicated in both methods now for now added value.

line-o: What do you think?

line-o:
As for the ByteBuffer shenanigans I found this after a quick search:
https://www.morling.dev/blog/bytebuffer-and-the-dreaded-nosuchmethoderror/

line-o:
Do we have `-source 8 -target 8` still somewhere which could be replaced by `--release 8`. Or can we drop java8 support completely. Is ajc keeping us from doing so?

line-o:
> It may be related to our use of ajc as well as javac

Can you point me to where the calls to javac and ajc happen?

line-o (commented):
🚀 straight to 17
That is a surprising but welcome development.

dizzzz:
@adamretter A jump to 17 makes sense to me; though you advocated in the past that we could not make these jumps due to limitations in (fill in any main stream linux distribution).

exist-parent/pom.xml
```diff
@@ -92,8 +92,8 @@
     </mailingLists>
 
     <properties>
-        <project.build.source>1.8</project.build.source>
-        <project.build.target>1.8</project.build.target>
+        <project.build.source>17</project.build.source>
```
dizzzz, line 95: please use the new style: 
image

exist-core/src/main/java/org/exist/storage/BrokerPool.java
```diff
@@ -975,29 +974,25 @@ public NotificationService getNotificationService() {
...
-                if (freeSpace != -1 && freeSpace < diskSpaceMin) {
-                    LOG.fatal("Partition containing DATA_DIR: {} is running out of disk space [minimum: {} free: {}]. Switching eXist-db into read-only mode to prevent data loss!", dataLock.getFile().toAbsolutePath().toString(), diskSpaceMin, freeSpace);
-                    setReadOnly();
-                }
+        final boolean isReadOnly = readOnly.get();
+        if (!isReadOnly) {
+            final long freeSpace = FileUtils.measureFileStore(dataLock.getFile(), FileStore::getUsableSpace);
+            if (freeSpace != -1 && freeSpace < diskSpaceMin) {
+                LOG.fatal("Partition containing DATA_DIR: {} is running out of disk space [minimum: {} free: {}]. Switching eXist-db into read-only mode to prevent data loss!", dataLock.getFile().toAbsolutePath().toString(), diskSpaceMin, freeSpace);
+                setReadOnly();
```
dizzzz, line 987: is there a reason why we cannot reset the readonly state?

exist-core/src/main/java/org/exist/storage/blob/BlobStoreDumpTool.java
```diff
@@ -127,19 +128,19 @@ public static void dump(final DigestType digestType, final Path persistentFile,
...
             printStream.println("blobId,referenceCount");
 
-            buffer.clear();
+            ((Buffer)buffer).clear();
 
             final byte[] id = new byte[digestBytesLen];
 
             while (channel.read(buffer) > -1) {
-                buffer.flip();
+                ((Buffer)buffer).flip();
```
dizzzz, line 136: is this still needed?

dizzzz:
please remove the java8 compatibility thingy from the parent pom, see https://github.com/eXist-db/exist/pull/4564/files#diff-e6685ec8347a52581b8394c0744d4fce1b0db723b9574dab889ef2cece5c1f24L1155 
image

exist-core/src/main/java/org/exist/storage/blob/BlobStoreDumpTool.java
dizzzz: I guess if you remove the java8 compatibility thing in parent pom, all should be ok?

dizzzz (changes requested):
some smaller suggestions and questions; LGTM!

reinhapa:
> Do we have `-source 8 -target 8` still somewhere which could be replaced by `--release 8`. Or can we drop java8 support completely. Is ajc keeping us from doing so?

I would prefer the `--release` aproach as it has been done by @dizzzz  on his JDK 11 PR

adamretter:
> This needs addressing
https://sonarcloud.io/project/issues?resolved=false&types=BUG&pullRequest=4574&id=eXist-db_exist&open=AYN_QOnx66Y7jh0OfK-r

> The if condition needs to change to guard against an NPE https://sonarcloud.io/code?id=eXist-db_exist&pullRequest=4574&selected=eXist-db_exist%3Aexist-core%2Fsrc%2Fmain%2Fjava%2Forg%2Fexist%2Fsource%2FSourceFactory.java&line=82

> Or contextPath must be nullable here as well https://sonarcloud.io/code?id=eXist-db_exist&pullRequest=4574&selected=eXist-db_exist%3Aexist-core%2Fsrc%2Fmain%2Fjava%2Forg%2Fexist%2Fsource%2FSourceFactory.java&line=154

@line-o I was addressing these as you wrote the message. All are taken care of now in a subsequent commit :-)

exist-core/src/main/java/org/exist/source/SourceFactory.java
adamretter: This is now fixed in a subsequent commit.

adamretter:
> As for the ByteBuffer shenanigans I found this after a quick search:
https://www.morling.dev/blog/bytebuffer-and-the-dreaded-nosuchmethoderror/

> Could the preferred method described in the article be also a solution for us?

@line-o No. That article is about running Java 11 compiled code targeting Java 8 on a Java 8 JVM. We are compiling Java 17 for Java 17 and running on JVM 17.

adamretter:
> Can you point me to where the calls to javac and ajc happen?

@line-o https://github.com/eXist-db/exist/blob/develop/exist-core/pom.xml#L980

exist-core/pom.xml
```diff
@@ -976,9 +1004,9 @@ The BaseX Team. The original license statement is also included below.]]></pream
             </plugin>
 
             <plugin>
-                <groupId>com.nickwongdev</groupId>
+                <groupId>dev.aspectj</groupId>
                 <artifactId>aspectj-maven-plugin</artifactId>
-                <version>1.12.6</version>
+                <version>1.13.1</version>
                 <dependencies>
                     <dependency>
```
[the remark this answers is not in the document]
adamretter: Yes. This is done further up in the `/project/build/dependencies` section.

exist-core/pom.xml
```diff
@@ -1001,7 +1029,7 @@ The BaseX Team. The original license statement is also included below.]]></pream
                         <configuration>
                             <forceAjcCompile>true</forceAjcCompile>  <!-- Required, otherwise the Aspects are not re-compiled when the src/main/java is recompiled for the test phase -->
                             <showWeaveInfo>true</showWeaveInfo>
-                            <complianceLevel>${project.build.source}</complianceLevel>
+                            <complianceLevel>8</complianceLevel>
```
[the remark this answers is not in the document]
adamretter: No. We only have `<complianceLevel>8</complianceLevel>` set for the ajc compiler, as anything higher causes it to error out at the moment. It is likely a bug that I will open with the Eclipse AspectJ project. So for the time being it restricts us a little bit, but it is not holding us back from compiling on Java 17 and running on Java 17. See the commit https://github.com/eXist-db/exist/commit/d94e219ff6015b2399e68332a7130f6f06015cf9 within the PR.

adamretter:
@dizzzz A jump to 17 makes sense to me; though you advocated in the past that we could not make these jumps due to limitations in (fill in any main stream linux distribution).

Indeed. However, since that time the situation has changed. All major modern linux versions now have OpenJDK 17 packages available. For older platforms, BellSoft Liberica and Zulu Azul provide builds of JDK 17 that are usable.

exist-parent/pom.xml
adamretter: I have kept them this way, as the property you have suggested is directly about the maven-compiler-plugin, whereas our properties are more generic and used for move than just the maven-compiler-plugin.

adamretter:
> please remove the java8 compatibility thingy from the parent pom, see https://github.com/eXist-db/exist/pull/4564/files#diff-e6685ec8347a52581b8394c0744d4fce1b0db723b9574dab889ef2cece5c1f24L1155

@dizzzz I have posted a follow-up commit to address that.

exist-core/src/main/java/org/exist/storage/blob/BlobStoreDumpTool.java
adamretter: Its related to the `<complianceLevel>8<complianceLevel>` of the ajc compile

exist-core/src/main/java/org/exist/storage/BrokerPool.java
adamretter: As far as I can tell I have not changed the existing logic here.

If you are asking, if after read-only we can switch back to read-write if disk space has been made available by the admin, then I have no idea without digging deeper into the BufferManager and Journal. However, I think that should be a separate feature request...

adamretter:
@dizzzz @line-o @reinhapa I have now been able to remove the `<complianceLevel>8</complianceLevel>` from the ajc compile.

This has the advantage that I have also been able to revert the `Buffer` casts commit as that is no longer required.

It has a slight disadvantage that the user now needs to enable an option in IntelliJ before compiling eXist-db otherwise they will receive ajc compile errors. The option is: `Module Settings` -> `Facets` -> `AspectJ (exist-core)`, and selecting `Post-compile weave mode`, and then rebuilding the project in Intellij.

Screenshot 2022-09-28 at 11 12 44

Otherwise without that option IntelliJ will yield these errors when compiling eXist-db:
```
ajc: Internal AspectJ compiler error: trouble in: 
public class org.exist.collections.LockedCollection extends java.lang.Object implements org.exist.collections.Collection:
  private final org.exist.storage.lock.ManagedCollectionLock managedCollectionLock
  private final org.exist.collections.Collection collection
  public void <init>(org.exist.storage.lock.ManagedCollectionLock, org.exist.collections.Collection)    org.aspectj.weaver.MethodDeclarationLineNumber: 65:2388
:
                    ALOAD_0     // Lorg/exist/collections/LockedCollection; this   (line 65)
                    INVOKESPECIAL java.lang.Object.<init> ()V
    constructor-execution(void org.exist.collections.LockedCollection.<init>(org.exist.storage.lock.ManagedCollectionLock, org.exist.collections.Collection))
    |               ALOAD_0     // Lorg/exist/collections/LockedCollection; this   (line 66)
    |               ALOAD_1     // Lorg/exist/storage/lock/ManagedCollectionLock; managedCollectionLock
    |               PUTFIELD org.exist.collections.LockedCollection.managedCollectionLock Lorg/exist/storage/lock/ManagedCollectionLock;
    |               ALOAD_0     // Lorg/exist/collections/LockedCollection; this   (line 67)
    |               ALOAD_2     // Lorg/exist/collections/Collection; collection
    |               PUTFIELD org.exist.collections.LockedCollection.collection Lorg/exist/collections/Collection;
    |               RETURN   (line 68)
    constructor-execution(void org.exist.collections.LockedCollection.<init>(org.exist.storage.lock.ManagedCollectionLock, org.exist.collections.Collection))
  end public void <init>(org.exist.storage.lock.ManagedCollectionLock, org.exist.collections.Collection)

  public static org.exist.collections.Collection unwrapLocked(org.exist.collections.Collection)    org.aspectj.weaver.MethodDeclarationLineNumber: 78:3013
:
                    ALOAD_0     // Lorg/exist/collections/Collection; collection   (line 80)
                    INSTANCEOF org.exist.collections.LockedCollection
                    IFEQ L0
                    ALOAD_0     // Lorg/exist/collections/Collection; collection   (line 82)
                    CHECKCAST org.exist.collections.LockedCollection
                    GETFIELD org.exist.collections.LockedCollection.collection Lorg/exist/collections/Collection;
                    ARETURN
                L0: ALOAD_0     // Lorg/exist/collections/Collection; collection   (line 84)
                    ARETURN
  end public static org.exist.collections.Collection unwrapLocked(org.exist.collections.Collection)

  public void close()    org.aspectj.weaver.MethodDeclarationLineNumber: 92:3498
:
                    ALOAD_0     // Lorg/exist/collections/LockedCollection; this   (line 93)
                    GETFIELD org.exist.collections.LockedCollection.collection Lorg/exist/collections/Collection;
                    INVOKEINTERFACE org.exist.collections.Collection.close ()V
                    ALOAD_0     // Lorg/exist/collections/LockedCollection; this   (line 94)
                    GETFIELD org.exist.collections.LockedCollection.managedCollectionLock Lorg/exist/storage/lock/ManagedCollectionLock;
                    INVOKEVIRTUAL org.exist.storage.lock.ManagedCollectionLock.close ()V
                    RETURN   (line 95)
  end public void close()

...
```

reinhapa (approved):

.github/workflows/ci-test.yml
```diff
@@ -4,28 +4,24 @@ jobs:
   test:
     name: (JDK ${{ matrix.jdk }} / ${{ matrix.os }}) Test
     env:
-      MAVEN_OPTS: -XX:+IgnoreUnrecognizedVMOptions --illegal-access=debug -XX:StartFlightRecording=maxsize=5g,disk=true,dumponexit=true,settings=default,filename=./
+      MAVEN_OPTS: --illegal-access=debug -XX:StartFlightRecording=maxsize=5g,disk=true,dumponexit=true,settings=default,filename=./
```
reinhapa, line 7: You can even drop `--illegal-access=debug` as this is no longer supported from Java 17+

reinhapa (approved):

.github/workflows/ci-test.yml
adamretter: @reinhapa Thanks, I have pushed an additional commit to address that.

exist-parent/pom.xml
dizzzz: euh? h,,, based om documentation the --release flag on java solves a few 'issues' as well? anyway... makes no difference I assume

exist-core/src/main/java/org/exist/storage/BrokerPool.java
dizzzz: correct, it was just a question from my side :-)

exist-core/src/main/java/org/exist/storage/blob/BlobStoreDumpTool.java
dizzzz: ah ok, thnx. and that is still needed?

dizzzz (approved):

dizzzz:
nice build results!

exist-parent/pom.xml
dizzzz: https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html  "This means that, unlike the [-source and -target options](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html), the compiler will detect and generate an error when using APIs that don't exist in previous releases of Java SE."

Issue #3233, opened 2020-01-28
exist-serialization PI is untested


duncdrum:
Partner issue to https://github.com/eXist-db/documentation/issues/265 

undocumented and untested is bad. Let's either deprecate the feature since no one seems to know what it does, or lets test and document it so ppl can actually use it.

PR #4582, opened 2022-10-05
remove exist-serialize option

marmoure:
Closes https://github.com/eXist-db/documentation/issues/265 and Closes https://github.com/eXist-db/exist/issues/3233
this feature was added in https://github.com/eXist-db/exist/commit/0cda7b6387aa4fb8ff82e6bb14a4de81bb210f74 and it was both `undocumented` and `untested` when trying to provide documentation for it we tried testing it, it didn't work as intended.
this feature is better removed as it doesn't work.


joewiz:
@adamretter Shouldn't we have left this open for 5 days before merging?

PR #4586, opened 2022-10-12
[hotfix] multipart body contents in XQuery

line-o:
This reverts commit c22ffe0ce3daf7034ec9069d86da8b42ae366b74.

It turns out that payloads in body encoded as  multipart/form-data have to be written to disk in order to be available in XQuery context. This is the reason the DiskFileItemFactory threshhold has to be zero.

Apparently, there are no tests covering this case. It was only revealed through testing Roaster against current develop.

### DIFF
--- a/exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
@@ -188,6 +188,9 @@ public class HttpRequestWrapper implements RequestWrapper {
         // Create a factory for disk-based file items
         final DiskFileItemFactory factory = new DiskFileItemFactory();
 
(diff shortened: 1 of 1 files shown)


adamretter:
Would it make sense to also add some tests at this time to prevent this happening again

reinhapa (commented):
Looks ok to me. Is the size threshold not zero as default?

dizzzz:
@reinhapa  no it is about 10kb

line-o:
Please merge this without tests, as the builds of current develop HEAD cannot parse multipart/form-data with small payloads.
I am happy to discuss how this can be tested, as this _needs_ to be tested. Especially, before the Jetty 11 upgrade.

line-o:
If you want to see the difference you can

- clone the Roaster repo
- cd into the working directory
- `npm i && npm test`

Issue #4588, opened 2022-10-15
[REGRESSION] Under certain circumstances importing XQSuite can cause an NPE


adamretter:
In eXist-db 4.x.x it is possible to setup an XQuery Module as an XQSuite using either of 3 formulations:

1.
```xquery
declare namespace test = "http://exist-db.org/xquery/xqsuite";
```

2.
```xquery
import module namespace test = "http://exist-db.org/xquery/xqsuite";
```

3.
```xquery
import module namespace test = "http://exist-db.org/xquery/xqsuite"
    at "resource:org/exist/xquery/lib/xqsuite/xqsuite.xql";
```

When an XQSuite is run via XSuite in eXist-db 5.x.x and 6.x.x the second (2) formulation fails with an `NullPointerException` due to the database not yet being initialised.

The change ecc89966235efdbe5b1fca3ac96b7e3db46c4fdb introduced to 5.x.x and 6.x.x attempted to remove the need to start up the database before initialising the XQSuite tests (resulting in a significant speed up when setting up test execution). Unfortunately this change also inadvertently caused the regression for the second (2) formulation.

PR #4589, opened 2022-10-15
Fixes an issue with XSuite not being able to run some formulations of XQsuite

adamretter:
Closes https://github.com/eXist-db/exist/issues/4588

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ModuleContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/ModuleContext.java
@@ -65,14 +65,15 @@ public class ModuleContext extends XQueryContext {
 
     public ModuleContext(final XQueryContext parentContext, final String modulePrefix, final String moduleNamespace,
             final String location) {
-        super();
+        super(parentContext != null ? parentContext.db : null, parentContext != null ? parentContext.getConfiguration() : null, null, false);
+
         this.modulePrefix = modulePrefix;
         this.moduleNamespace = moduleNamespace;
         this.location = location;
+
         setParentContext(parentContext);
 
-        loadDefaults(getConfiguration());
-        this.profiler = new Profiler(getBroker() != null ? getBroker().getBrokerPool() : null);
+        loadDefaults(this.configuration);
     }
 
     @Override
@@ -103,7 +104,6 @@ public class ModuleContext extends XQueryContext {
         this.parentContext = parentContext;
         //XXX: raise error on null!
         if (parentContext != null) {
-            this.db = parentContext.db;
             this.baseURI = parentContext.baseURI;
             try {
                 if (location.startsWith(XmldbURI.XMLDB_URI_PREFIX) ||
@@ -181,11 +181,6 @@ public class ModuleContext extends XQueryContext {
             LOG.error(e);
         }
         return ctx;
-    }
-
-    @Override
-    public Configuration getConfiguration() {
-        return parentContext.getConfiguration();
     }
 
     @Override
--- a/exist-core/src/main/java/org/exist/xquery/ModuleContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/ModuleContext.java
@@ -65,7 +65,10 @@ public class ModuleContext extends XQueryContext {
 
     public ModuleContext(final XQueryContext parentContext, final String modulePrefix, final String moduleNamespace,
             final String location) {
-        super(parentContext != null ? parentContext.db : null, parentContext != null ? parentContext.getConfiguration() : null, null, false);
+        super(parentContext != null ? parentContext.db : null,
+                parentContext != null ? parentContext.getConfiguration() : null,
+                null,
+                false);
 
         this.modulePrefix = modulePrefix;
         this.moduleNamespace = moduleNamespace;
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -78,6 +78,7 @@ import org.exist.security.PermissionDeniedException;
 import org.exist.security.Subject;
 import org.exist.source.*;
 import org.exist.stax.ExtendedXMLStreamReader;
+import org.exist.storage.BrokerPool;
 import org.exist.storage.DBBroker;
 import org.exist.storage.UpdateListener;
 import org.exist.storage.lock.Lock.LockMode;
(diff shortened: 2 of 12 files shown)


dizzzz (approved):

line-o:
Can you clarify what you mean by XSuite is it XQSuite? Can those terms be used interchangeably?

reinhapa (approved):

exist-core/src/test/resources-filtered/conf.xml
```diff
@@ -888,6 +888,10 @@
             raise-error-on-failed-retrieval="no">
         
         <builtin-modules>
+
+            <!-- XQSuite Test Framework for XQuery testing -->
```
line-o, line 892: Will this now effectively add XQSuite to the XQuery contexts of the next release?

exist-core/src/main/java/org/exist/xquery/ModuleContext.java
```diff
@@ -65,14 +65,15 @@ public class ModuleContext extends XQueryContext {
 
     public ModuleContext(final XQueryContext parentContext, final String modulePrefix, final String moduleNamespace,
             final String location) {
-        super();
+        super(parentContext != null ? parentContext.db : null, parentContext != null ? parentContext.getConfiguration() : null, null, false);
```
line-o, line 68: I find this very hard to read. Can this be refactored for readability?

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3012,12 +3061,10 @@ public Object getAttribute(final String attribute) {
     /**
      * Load the default prefix/namespace mappings table and set up internal functions.
      *
-     * @param config the configuration
+     * @param config the configuration if available
      */
     @SuppressWarnings("unchecked")
-    void loadDefaults(final Configuration config) {
-        this.watchdog = new XQueryWatchDog(this);
```
line-o, line 3019: where is the watchdog set now?

exist-core/src/test/xquery/xsuite-declare-xqsuite.xqm
```diff
@@ -0,0 +1,35 @@
...
+ : Lesser General Public License for more details.
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+(:~
+ : Test for using XQSuite from XSuite via `declare namespace`.
```
line-o, line 25: XSuite = XQSuite?
If this is just a typo please correct it as it has the potential to confuse people (it did confuse me).

exist-core/src/test/xquery/xsuite-import-xqsuite-location.xqm
```diff
@@ -0,0 +1,36 @@
...
+ : Lesser General Public License for more details.
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+(:~
+ : Test for using XQSuite from XSuite via `import module` with a location hint.
```
line-o, line 25: see above

line-o (commented):
LGTM. Please have a look at my notes.

exist-core/src/test/xquery/xsuite-declare-xqsuite.xqm
joewiz: See https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/test/runner/XSuite.java#L47-L59.

exist-core/src/test/resources-filtered/conf.xml
adamretter: No.

exist-core/src/main/java/org/exist/xquery/ModuleContext.java
adamretter: @line-o That is not possible in Java as a `super` has to be the first call in the constructor!

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
adamretter: You can see it in the main constructor.

exist-core/src/test/xquery/xsuite-declare-xqsuite.xqm
adamretter: @line-o XSuite is the Java test runner that calls XQSuite. This makes no change to eXide.

exist-core/src/test/xquery/xsuite-import-xqsuite-location.xqm
adamretter: A mentioned XSuite != XQSuite

exist-core/src/test/xquery/xsuite-import-xqsuite-no-location.xqm
```diff
@@ -0,0 +1,35 @@
...
+ : Lesser General Public License for more details.
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+(:~
+ : Test for using XQSuite from XSuite via `import module` with no location hint.
```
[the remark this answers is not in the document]
adamretter: A mentioned XSuite != XQSuite

exist-core/src/main/java/org/exist/xquery/ModuleContext.java
line-o: I did not know that. How about extracting the logic for the first parameter into a method?

line-o: Maybe reformatting would do the trick?
```java
super(
  parentContext != null ? parentContext.db : null,
  parentContext != null ? parentContext.getConfiguration() : null,
  null,
  false
);
```

line-o: Digging deeper, I wonder if an additional XQueryContext constructor would make sense.
The module context needs to set its XQueryContext here. This is almost like the constructor that copies the context:
https://github.com/eXist-db/exist/blob/cff6e205ff75f7154d037251a584685e4427b4a3/exist-core/src/main/java/org/exist/xquery/XQueryContext.java#L480 

It does not hand over the profiler and never loads defaults, so it is special.
In which cases does the ModuleContext get instantiated without a parentContext?

line-o: What happens when we call `super(parentContext)` instead?

adamretter: @line-o I have pushed a commit to reformat the call to the super-constructor as you requested.

> I wonder if an additional XQueryContext constructor would make sense.

I don't see how that would help.

> What happens when we call super(parentContext) instead?

It doesn't work.

adamretter:
@line-o (cc @dizzzz) I believe all review has been addressed. Would you mind merging it?

PR #4597, opened 2022-10-21
Fix issues with SMTP dialogue in the Mail Module when sending emails.

adamretter:
1. Emails with text, and html messages, and with attachments were not correctly serialized.
2. Make sure that header parameters that have non-token characters are correctly quoted.
3. Be polite and tell the SMTP server when we have finished.

### DIFF
--- a/extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/SendEmailFunction.java
+++ b/extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/SendEmailFunction.java
@@ -75,6 +75,8 @@ public class SendEmailFunction extends BasicFunction {
 
     static final String ERROR_MSG_NON_MIME_CLIENT = "Error your mail client is not MIME Compatible";
 
+    private static final Random RANDOM = new Random();
+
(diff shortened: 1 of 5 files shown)


extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/SendEmailFunction.java
```diff
@@ -332,10 +340,14 @@ private boolean[] sendBySMTP(final Mail[] mails, final String smtpServerArg) thr
 
             //write SMTP message(s)
             for (int i = 0; i < mails.length; i++) {
```
dizzzz, line 342: could use java5 styled loop :-P

dizzzz (approved):

extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/SendEmailFunction.java
adamretter: Yes, but that is existing code rather than new code that I added

adamretter:
Thanks @dizzzz

PR #4617, opened 2022-11-18
[4.x.x] Prevent OutOfMemory on JDK 17 when loading EXPath Modules

ohumbel:
We use 4.x.x with custom `org/exist/xquery/Function`s. 
This works fine on JDK 11. 
As soon as we use JDK 17, we get OutOfMemory in the Non-heap region (Lambdas from `org/exist/xquery/Function`).
6 does not show the problem.

If we pick the commit: 

4ec61b0af9d5cec91ebcfe7c3a79bef9d1155555 
([bugfix] EXPath Repo classes must be instantiable via reflection, however LamdaMetafactory cannot cross classloader boundaries. As only EXPath Repo classes are now loaded by the EXistClassLoader everything else is loaded by a URLClassLoader (via AppAssemblerBooter), we have to use reflection instead of invokeDynamic to load EXPath Jar Modules) 

from develop, the OutOfMemory goes away.

Forensics done together with @reinhapa.

### DIFF
--- a/exist-core/src/main/java/org/exist/repo/ExistRepository.java
+++ b/exist-core/src/main/java/org/exist/repo/ExistRepository.java
@@ -10,9 +10,7 @@
 package org.exist.repo;
 
 import java.io.IOException;
-import java.lang.invoke.LambdaMetafactory;
-import java.lang.invoke.MethodHandle;
-import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Constructor;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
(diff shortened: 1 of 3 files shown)


adamretter:
@ohumbel Thanks for the contribution, this is very interesting. Do you believe that this also occurs in eXist-db 5 and 6?

ohumbel:
> @ohumbel Thanks for the contribution, this is very interesting. Do you believe that this also occurs in eXist-db 5 and 6?

@adamretter We tested both eXist-db 4 and 6. 
eXist-db 6 does *not* expose the problem. 
On 4, the cherry pick of the one commit safely made OutOfMemory go away. 
The commit gets rid of `invokedynamic` stuff like `LambdaFactory` and `MethodHandle`s.

If the commit already is on 5 (which I did not check), we could assume the problem is not there.

reinhapa (approved):

PR #4621, opened 2022-11-28
allow all options in map:merge#2

### LINKED ISSUE #4620: [BUG] map:merge#2 with empty options not spec compliant
**Describe the bug**

The second parameter to map:merge#2 _may_ contain "duplicates" as a key.
The current implementation does not allow that.

`map:merge((), map{})` throws an error instead 

```
<error type="err:FOJS0005" message="Invalid options. value for duplicates key was not recognised: "/>
```

**Expected behavior**

To behave as defined in the spec https://www.w3.org/TR/xpath-functions-31/#func-map-merge

> 1. The effect of calling the single-argument function is the same as the effect of calling the two-argument function with an empty map as the value of $options.

**To Reproduce**

```Xquery
xquery version "3.1";

declare variable $t:maps := (
  map {"a": 1},
  map {"a": 2},
  map {"a": 3}
);

declare
    %test:assertTrue
function t:map-merge-empty-options() {
    let $expected := map:merge($t:maps)
    let $actual := map:merge($t:maps, map {})
    return $expected?a eq $actual?a
};
```

 - OS: macOS 12.6
 - Java Version Java8u352

**Additional context**

line-o:
- support options `map{ "duplicates": "combine" }` and `map{ "duplicates": "reject" }`
  Cherry-picked 3 commits from #3738 and squashed them 
  These enhancements are non-breaking and can be merged in to develop before the next major release. 
- allow an empty options map passed to map:merge#2 - fixes #4620

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/AbstractMapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/AbstractMapType.java
@@ -31,6 +31,7 @@ import org.exist.xquery.value.*;
 
 import javax.annotation.Nullable;
 import java.util.List;
+import java.util.function.BinaryOperator;
 
 /**
  * Abstract base class for map types. A map item is also a function item. This class thus extends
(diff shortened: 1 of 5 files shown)


reinhapa (approved):

adamretter:
@alanpaxton As you did the underlying work which is cherry-picked into here - would you mind taking a look and giving us a "Approve"/"Reject"? I would add you as a reviewer, but it seems to only let me choose core-team members!

alanpaxton:
Approve.
Thanks @adamretter, I had a quick look and it seems the existing tests will confirm we haven't changed the current (non-spec) default merge order, so LGTM.

dizzzz (approved):

Issue #4612, opened 2022-11-13
XQSuite incorrectly reports non-error as success for `%test:assertError`


adamretter:
It seems there is another problem in XQSuite whereby we cannot trust its results entirely.

Consider this test case from https://github.com/eXist-db/exist/blob/eXist-6.0.1/exist-core/src/test/xquery/arrays/arrays.xql#L1013:
```xquery
declare
    %test:assertError("err:FORG0006")
function arr:put_pos_double() {
    array:put(["a", "b", "c", "d"], 1.5, "x")?*
};
```

The test clearly states that the functions should cause an `FORG0006` error. However, if you run the code within the function, you will find that eXist-db instead produces the result:
```
("x", "b", "c", "d")
```

Regardless, XQSuite reports that the test passed, where it should have failed.

* The test was introduced in commit fb7ed4a (2019-11-04).
* The issue I have now revealed and which is shown by this test is first visible in eXist-db release version 5.2.0. It affects all versions after that up until, and including, the current `HEAD` of the `develop` branch.
* I also remember we had some issue like this in the past for which I prepared a fix (2019-03-12) - https://github.com/eXist-db/exist/commit/a753749bdf8daad851de9f68dac9f7e9ec004adb
* This issue could be much older but I have not investigated.

----

I discovered this, because the XQuery 3.1 spec says the error should be `XPTY0004` and not `FORG0006`... but that can be addressed once we fix XQSuite.

PR #4622, opened 2022-11-28
Fix bugs in XQSuite and XQuery tests written for XQSuite

adamretter:
Closes https://github.com/eXist-db/exist/issues/4612

Unfortunately fixing a number of bugs in XQSuite has revealed:
1. A number of tests written for XQSuite that are invalid XQuery (according to the W3C spec)
2. A number of sub-type promotion errors in eXist-db

### DIFF
--- a/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
+++ b/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
@@ -740,9 +740,9 @@ declare %private function test:check-assertions($assertions as element(annotatio
     return
         switch ($assert)
             case "assertEquals" return
-                test:assertEquals($annotation/value/string(), $result)
+                test:assertEquals($annotation/value, $result)
             case "assertEqualsPermutation" return
-                test:assertEqualsPermutation($annotation/value/string(), $result)
+                test:assertEqualsPermutation($annotation/value, $result)
             case "assertEmpty" return
                 test:assertEmpty($result)
             case "assertExists" return
@@ -763,9 +763,9 @@ declare %private function test:check-assertions($assertions as element(annotatio
  : Check for equality of the function's result with the value in the annotation.
  : This function transforms the result to a string before checking for equality.
  :)
-declare %private function test:assertEquals($values as item()*, $result as item()*) as element(report)? {
-    if (exists($values)) then
-        if (count($values) eq count($result)) then
+declare %private function test:assertEquals($values as element(value)*, $result as item()*) as element(report)? {
+    if (fn:exists($values)) then
+        if (fn:count($values) eq fn:count($result)) then
             let $tests := fn:for-each-pair($values, $result, test:equals#2)
             let $equal := every $test in $tests satisfies $test
             return
@@ -775,15 +775,15 @@ declare %private function test:assertEquals($values as item()*, $result as item(
                    <report>
                         <failure message="assertEquals failed."
                             type="failure-error-code-1">
-                        { $values }
+                        { $values ! test:xdm-value-from-annotation-value(.) }
                         </failure>
                         <output>{ $result }</output>
                     </report>
         else
             <report>
-                <failure message="assertEquals failed: wrong number of items returned by function. Expected: {count($values)}. Got: {count($result)}"
+                <failure message="assertEquals failed: wrong number of items returned by function. Expected: {fn:count($values)}. Got: {fn:count($result)}"
                     type="failure-error-code-1">
-                { $values }
+                { $values ! test:xdm-value-from-annotation-value(.) }
                 </failure>
                 <output>{ $result }</output>
             </report>
@@ -795,9 +795,9 @@ declare %private function test:assertEquals($values as item()*, $result as item(
  : Check for equality of the function's result with the value in the annotation.
  : This function transforms the result to a string before checking for equality.
  :)
-declare %private function test:assertEqualsPermutation($values as item()*, $result as item()*) as element(report)? {
-    if (exists($values)) then
-        if (count($values) eq count($result)) then
+declare %private function test:assertEqualsPermutation($values as element(value)*, $result as item()*) as element(report)? {
+    if (fn:exists($values)) then
+        if (fn:count($values) eq fn:count($result)) then
             let $tests :=
                 for $item in $result
                 return
@@ -810,15 +810,15 @@ declare %private function test:assertEqualsPermutation($values as item()*, $resu
                    <report>
                         <failure message="assertEqualsPermutation failed."
                             type="failure-error-code-1">
-                        { $values }
+                        { $values ! test:xdm-value-from-annotation-value(.) }
                         </failure>
                         <output>{ $result }</output>
                     </report>
         else
             <report>
-                <failure message="assertEqualsPermutation failed: wrong number of items returned by function. Expected: {count($values)}. Got: {count($result)}"
+                <failure message="assertEqualsPermutation failed: wrong number of items returned by function. Expected: {fn:count($values)}. Got: {fn:count($result)}"
                     type="failure-error-code-1">
-                { $values }
+                { $values ! test:xdm-value-from-annotation-value(.) }
                 </failure>
                 <output>{ $result }</output>
             </report>
@@ -826,13 +826,14 @@ declare %private function test:assertEqualsPermutation($values as item()*, $resu
         ()
 };
 
-declare %private function test:equals($value as item(), $result as item()) as xs:boolean {
+declare %private function test:equals($annotation-value as element(value), $result as item()) as xs:boolean {
     let $normResult :=
         typeswitch ($result)
             case node() return
                 test:normalize($result)
             default return
                 $result
+    let $value := test:xdm-value-from-annotation-value($annotation-value)
     let $normValue := test:cast-to-type($value, $result)
     return
         typeswitch ($normResult)
--- a/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
+++ b/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
@@ -57,6 +57,7 @@ declare variable $test:TEST_NAMESPACE := "http://exist-db.org/xquery/xqsuite";
 declare variable $test:UNKNOWN_ASSERTION := QName($test:TEST_NAMESPACE, "no-such-assertion");
 declare variable $test:WRONG_ARG_COUNT := QName($test:TEST_NAMESPACE, "wrong-number-of-arguments");
 declare variable $test:TYPE_ERROR := QName($test:TEST_NAMESPACE, "type-error");
+declare variable $test:UNKNOWN_ANNOTATION_VALUE_TYPE := QName($test:TEST_NAMESPACE, "unknown-annotation-value-type");
 
 (:~
  : Main entry point into the module. Takes a sequence of function items.
@@ -456,7 +457,7 @@ declare %private function test:get-assertions($meta as element(function), $first
 (:~
  : Collect %test:arg and %test:args for the current run.
  :)
-declare %private function test:get-run-args($firstArg as element(annotation)?) {
+declare %private function test:get-run-args($firstArg as element(annotation)?) as element(annotation)* {
     if ($firstArg) then
         let $nextBlock :=
             $firstArg/following-sibling::annotation[matches(@name, ":args?")]
@@ -481,7 +482,7 @@ declare %private function test:get-run-args($firstArg as element(annotation)?) {
 declare %private function test:call-test($func as function(*), $meta as element(function), $args as element(annotation)*) {
     let $funArgs :=
         if ($args[1]/@name = "test:args") then
-            test:map-arguments($args/value/string(), $meta/argument)
+            test:map-arguments($args/value, $meta/argument)
         else
             test:map-named-arguments($meta/argument, $args)
     return
@@ -511,12 +512,12 @@ declare %private function test:map-named-arguments($funcArgs as element(argument
             $mappedArgs
 };
 
-declare %private function test:map-named-argument($testArgs as element(value)*, $funcArg as element(argument)) {
+declare %private function test:map-named-argument($test-args as element(value)*, $func-arg as element(argument)) {
     let $data :=
         try {
-            test:cast($testArgs, $funcArg)
+            test:cast($test-args, $func-arg)
         } catch * {
-            error($test:TYPE_ERROR, "Failed to cast annotation arguments to required target type " || $funcArg/@type)
+            fn:error($test:TYPE_ERROR, "Failed to cast annotation arguments to required target type " || $func-arg/@type)
         }
     return
         (: If we need to return a sequence of values, enclose them into a closure function :)
@@ -526,47 +527,135 @@ declare %private function test:map-named-argument($testArgs as element(value)*, 
 (:~
  : For %args: transform each annotation parameter into the type required for the function parameter.
  :)
-declare %private function test:map-arguments($testArgs as xs:string*, $funcArgs as element(argument)*) {
-    if (exists($testArgs)) then
-        if (count($testArgs) != count($funcArgs)) then
-            error(
+declare %private function test:map-arguments($test-args as element(value)*, $func-args as element(argument)*) {
+    if (fn:not(fn:empty($test-args))) then
+        if (fn:count($test-args) ne fn:count($func-args)) then
+            fn:error(
                 $test:WRONG_ARG_COUNT,
                 "The number of arguments specified in test:args must match the arguments of the function to test"
             )
         else
-            fn:for-each-pair($testArgs, $funcArgs, function($targ as xs:string, $farg as element(argument)) {
-                let $data := test:cast($targ, $farg)
-                return
-                    function() { $data }
+            fn:for-each-pair($test-args, $func-args, function($targ as element(value), $farg as element(argument)) {
+                function() {
+                    test:cast($targ, $farg)
+                }
             })
     else
         ()
 };
 
-declare %private function test:cast($targs as xs:string*, $farg as element(argument)) {
+
+(:~
+ : Convert an annotation value into an XDM value.
+ :
+ : Annotation values can only be one of the following types: xs:string, xs:integer, xs:decimal or xs:double.
+ :
+ : @param $annotation-value the annotation value element as returned by util:inspect-function
+ :
+ : @return the atomic value of the annotation
+ :)
+declare %private function test:xdm-value-from-annotation-value($annotation-value as element(value)) {
+    switch ($annotation-value/string(@type))
+        case "xs:string"
+        return
+            xs:string($annotation-value/text())
+
+        case "xs:integer"
+        return
+            xs:integer($annotation-value/text())
+
+        case "xs:decimal"
+        return
+            xs:decimal($annotation-value/text())
+
+        case "xs:double"
+        return
+            xs:double($annotation-value/text())
+
+        default
+        return
+            fn:error(
+                $test:UNKNOWN_ANNOTATION_VALUE_TYPE,
+                "The annotation value has an unknown type"
+            )
+};
+
+declare %private function test:cast($targs as element(value)*, $farg as element(argument)) {
     for $targ in $targs
+    let $varg := test:xdm-value-from-annotation-value($targ)
     return
         switch (string($farg/@type))
-            case "xs:string" return
-                string($targ)
-            case "xs:integer" case "xs:int" return
-                xs:integer($targ)
-            case "xs:decimal" return
-                xs:decimal($targ)
-            case "xs:float" case "xs:double" return
-                xs:double($targ)
-            case "xs:date" return
-                xs:date($targ)
-            case "xs:dateTime" return
-                xs:dateTime($targ)
-            case "xs:time" return
-                xs:time($targ)
-            case "element()" return
-                parse-xml($targ)/*
-            case "text()" return
-                text { string($targ) }
-            default return
-                $targ
+            case "xs:string"
+            return
+                if ($varg instance of xs:string)
+                then
+                    $varg
+                else
+                    string($varg)
+
+            case "xs:integer"
+            return
+                if ($varg instance of xs:integer)
+                then
+                    $varg
+                else
+                    xs:integer($varg)
+
+            case "xs:int"
+            return
+                xs:int($varg)
+
+            case "xs:decimal"
+            return
+                if ($varg instance of xs:decimal)
+                then
+                    $varg
+                else
+                    xs:decimal($varg)
+
+            case "xs:float"
+            return
+                xs:float($varg)
+
+            case "xs:double"
+            return
+                if ($varg instance of xs:decimal)
+                then
+                    $varg
+                else
+                    xs:double($varg)
+
+            case "xs:date"
+            return
+                xs:date($varg)
+
+            case "xs:dateTimeStamp"
+            return
+                xs:dateTimeStamp($varg)
+
+            case "xs:dateTime"
+            return
+                xs:dateTime($varg)
+
+            case "xs:time"
+            return
+                xs:time($varg)
+
+            case "element()"
+            return
+                fn:parse-xml($varg)/element()
+
+            case "text()"
+            return
+                if ($varg instance of xs:string)
+                then
+                    text { $varg }
+                else
+                    text { xs:string($varg) }
+
+            default
+            return
+                $varg
 };
 
 declare function test:apply($func as function(*), $meta as element(function), $args as item()*) {
--- a/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
+++ b/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
@@ -801,7 +801,7 @@ declare %private function test:assertEqualsPermutation($values as element(value)
             let $tests :=
                 for $item in $result
                 return
-                    some $value in $values satisfies test:equals($item, $value)
+                    some $value in $values satisfies test:equals($value, $item)
             let $equal := every $test in $tests satisfies $test
             return
                 if ($equal) then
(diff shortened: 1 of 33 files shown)


exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
```diff
@@ -526,47 +527,135 @@ declare %private function test:map-named-argument($testArgs as element(value)*,
...
+            return
+                xs:dateTime($varg)
+
+            case "xs:time"
+            return
+                xs:time($varg)
+
+            case "element()"
+            return
+                fn:parse-xml($varg)/element()
```
joewiz, line 646: The `fn:parse-xml` function returns a `document-node`, so selecting `/element()` here will return an empty sequence. See https://www.w3.org/TR/xpath-functions-31/#func-parse-xml.

line-o: I think this should work as I usually select the first child node() when dealing with document-nodes

joewiz: Ah, you're quite right. Please disregard my comment above.

line-o:
Please have a look at the test failures.

dizzzz (approved):

adamretter:
> Please have a look at the test failures
@line-o All tests are now passing.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
```diff
@@ -801,7 +801,7 @@ declare %private function test:assertEqualsPermutation($values as element(value)
             let $tests :=
                 for $item in $result
                 return
-                    some $value in $values satisfies test:equals($item, $value)
+                    some $value in $values satisfies test:equals($value, $item)
```
joewiz, line 804: Good catch! We should consider renaming `$item` to `$result` in this function; `$result` to `$results`; and perhaps even `$value` and `$values` to `$annotation-value` and `$annotation-values`. This would make it quite apparent how these variables need to align with the `test:equals()` function signature: `test:equals($annotation-value as element(value), $result as item())`. (Perhaps in a future PR, that is.)

line-o: I would be in favour of changing the name as part of this PR

adamretter:
> What about the error in Deploy / Build and Test Images?

I don't know what the Bats tests do and I don't understand them. Any idea who added them and/or what they do?

duncdrum:
They test the docker container. In this case https://github.com/eXist-db/exist/actions/runs/3574037174/jobs/6008812699#step:11:12 if the logs are free of errors on startup. They are not in this branch

adamretter:
Okay thanks @duncdrum, do you know where I can see the logs?

adamretter:
@duncdrum Unless I am mistaken the `exist.log` or Docker console log is not available from the CI for this?

adamretter:
I build a Docker image locally, and it does indeed show that there was a hidden XQuery error in Monex which is shown by the fixes in this PR:

```
30 Nov 2022 22:03:45,321 [main] ERROR (AutoDeploymentTrigger.java [execute]:127) - Exception during deployment of app monex-3.0.4.xar: Error found while processing repo.xml: err:XPTY0004 xs:string(/db/apps/monex/data) is not a sub-type of xs:anyURI [at line 12, column 15, source: /exist/etc/../data/expathrepo/monex-3.0.4/post-install.xql]
org.expath.pkg.repo.PackageException: Error found while processing repo.xml: err:XPTY0004 xs:string(/db/apps/monex/data) is not a sub-type of xs:anyURI [at line 12, column 15, source: /exist/etc/../data/expathrepo/monex-3.0.4/post-install.xql]
	at org.exist.repo.Deployment.deploy(Deployment.java:469) ~[exist.uber.jar:6.1.0-SNAPSHOT]
	at org.exist.repo.Deployment.installAndDeploy(Deployment.java:281) ~[exist.uber.jar:6.1.0-SNAPSHOT]
```

Seems we will also need to fix Monex and get a new release of that out...

adamretter:
Okay I fixed this issue in a quick hotfix of Monex 3.0.5 and this has fixed the build here.

line-o (commented):
Since one of the bundled applications seems to choke on the changes in this PR we need to investigate first if other packages might also be affected.

line-o:
related to #4632

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
```diff
@@ -526,47 +527,135 @@ declare %private function test:map-named-argument($testArgs as element(value)*,
 (:~
  : For %args: transform each annotation parameter into the type required for the function parameter.
  :)
-declare %private function test:map-arguments($testArgs as xs:string*, $funcArgs as element(argument)*) {
-    if (exists($testArgs)) then
-        if (count($testArgs) != count($funcArgs)) then
-            error(
+declare %private function test:map-arguments($test-args as element(value)*, $func-args as element(argument)*) {
+    if (fn:not(fn:empty($test-args))) then
```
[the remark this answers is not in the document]
line-o: and in all following instances

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
```diff
@@ -706,13 +795,13 @@ declare %private function test:assertEquals($values as item()*, $result as item(
  : Check for equality of the function's result with the value in the annotation.
  : This function transforms the result to a string before checking for equality.
  :)
-declare %private function test:assertEqualsPermutation($values as item()*, $result as item()*) as element(report)? {
-    if (exists($values)) then
-        if (count($values) eq count($result)) then
+declare %private function test:assertEqualsPermutation($values as element(value)*, $result as item()*) as element(report)? {
+    if (fn:not(fn:empty($values))) then
```
line-o, line 799: why is `exists()` not sufficient here?

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
```diff
@@ -819,14 +909,27 @@ declare %private function test:assertTrue($result as item()*) as element(report)
...
 declare %private function test:assertError($value as xs:string, $result as item()*) as element(report)? {
-    if ($result) then
-        ()
-    else
-        <report>
-            <failure message="assertError failed. Expected error {$value}"
-                type="failure-error-code-1"/>
-            <output>{ $result }</output>
-        </report>
+    let $ebv :=
```
line-o, line 912: This code looks like it could be written in a more readable way.

exist-core/src/test/xquery/xqsuite/xqsuite-assertions-inline.xqm
```diff
@@ -0,0 +1,185 @@
...
+ : Inline Tests for XQSuite assertions.
+ :
+ : @author Adam Retter
+ :)
+module namespace xqtai = "http://exist-db.org/xquery/xqsuite/test/assertions-inline";
+
+(:~
+ : Resource functions to test through XQSuite.
+ :)
+import module namespace xqtares = "http://exist-db.org/xquery/xqsuite/test/assertions/resources"
```
line-o, line 45: Here it is imported and not just the namespace ignored.

exist-core/src/test/xquery/xquery3/parseIetfDate.xqm
```diff
@@ -51,8 +51,10 @@ declare
     %test:args('') %test:assertError("FORG0010")
     %test:args('asdfa') %test:assertError("FORG0010")
 
-    %test:args(0) %test:assertError("FORG0010")
-    %test:args(1) %test:assertError("FORG0010")
+    %test:args(0) %test:assertError("XPTY0004")
```
line-o, line 54: These test cases that will result in XPTY0004 can be removed. They just test the type system of exist and not the function itself.

exist-core/src/test/xquery/xquery3/serialize.xql
```diff
@@ -234,7 +234,7 @@ function ser:serialize-empty-sequence() {
 
 declare
     %test:args(1234)
-    %test:assertEquals('"1234"')
+    %test:assertEquals(1234)
```
line-o, line 237: Adaptive serialisation will always output a string. The test should reflect that.

exist-core/src/test/xquery/xquery3/serialize.xql
```diff
@@ -243,7 +243,7 @@ function ser:adaptive-simple-atomic($atomic as xs:anyAtomicType) {
...
```
line-o, line 246: see above

line-o (dismissed):
Please have a look at the comments.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
line-o: In which cases does effectiveBoolean value fail - and could that not be mitigated by a call to `exists()` ?

line-o:
Regarding problems with bundled applications: the necessity of commit a1a35e041d79d717f66729bfdf085df258ebcb96 indicates that there might be more problems after this PR is merged.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
line-o:
```
declare %private function test:assertError($value as xs:string, $result as item()*) as element(report)? {
    if (exists($result)) then ()
    else
        <report>
            <failure message="assertError failed. Expected error {$value}"
                type="failure-error-code-1"/>
            <output>{ $result }</output>
        </report>
};
```
is sufficient. I just tested it. This change alone shows that the array:put test actually fails. Nothing else needed.

line-o:
This PR contains several - welcome and good - changes that would warrant their own separate PRs. The risk is that we accidentally create more harm than we solve.

adamretter:
> Since one of the bundled applications seems to choke on the changes in this PR we need to investigate first if other packages might also be affected.

@line-o I am afraid that is not specific to this PR. That is just a general statement that: whenever we modify eXist-db's XQuery engine to be more spec compliant, then it is likely that any code that relied on behaviour that is not spec compliant will need to be fixed. This has been true for every XQuery change to eXist-db, and does not relate specifically to this PR.

This change, just like any other PR that changes eXist-db's XQuery compliance in a non-compatible way, is scheduled for a major release.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
```diff
@@ -511,12 +512,12 @@ declare %private function test:map-named-arguments($funcArgs as element(argument
...
 
-declare %private function test:map-named-argument($testArgs as element(value)*, $funcArg as element(argument)) {
+declare %private function test:map-named-argument($test-args as element(value)*, $func-arg as element(argument)) {
     let $data :=
         try {
-            test:cast($testArgs, $funcArg)
+            test:cast($test-args, $func-arg)
         } catch * {
-            error($test:TYPE_ERROR, "Failed to cast annotation arguments to required target type " || $funcArg/@type)
+            fn:error($test:TYPE_ERROR, "Failed to cast annotation arguments to required target type " || $func-arg/@type)
```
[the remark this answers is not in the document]
adamretter: Why? What do we gain for doing that?

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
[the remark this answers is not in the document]
adamretter: Why? What do we gain for doing that?

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
```diff
@@ -526,47 +527,135 @@ declare %private function test:map-named-argument($testArgs as element(value)*,
...
-                xs:time($targ)
-            case "element()" return
-                parse-xml($targ)/*
-            case "text()" return
-                text { string($targ) }
-            default return
-                $targ
+            case "xs:string"
+            return
+                if ($varg instance of xs:string)
```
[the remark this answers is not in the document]
adamretter: > Isn't this already being handled in test:xdm-value-from-annotation-value?

No it is not.

These are an optimisation, no need to attempt costly conversion if the type is already of the correct type.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
adamretter: I changed it back. There was a mix of both styles already present in `xqsuite.xql` before even my changes

exist-core/src/test/xquery/xquery3/serialize.xql
adamretter: This is not related to adaptive serialization.

This is related to the XQSuite use of XQuery annotations. The XQuery 3.1 spec says that an `Annotation` can have zero or more [Literal](https://www.w3.org/TR/xquery-31/#prod-xquery31-Literal) values. Further, it says that a `Literal` is either a `StringLiteral`, `NumericLiteral`, `IntegerLiteral`, `DecimalLiteral` or `DoubleLiteral`.

If you `assertEquals(1234)` and your test returns an integer, then the integers will now be correctly compared. There was some support for this before in XQSuite but it was broken/incomplete.

exist-core/src/test/xquery/xquery3/serialize.xql
adamretter: As above - this is not related to adaptive serialization.

extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/MailStoreFunctions.java
```diff
@@ -52,7 +52,7 @@
  *
  * @see org.exist.xquery.BasicFunction#BasicFunction(org.exist.xquery.XQueryContext, org.exist.xquery.FunctionSignature)
  */
-public class MailStoreFunctions extends BasicFunction
+public class 	MailStoreFunctions extends BasicFunction
```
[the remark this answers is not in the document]
adamretter: Done.

exist-core/src/test/xquery/xqsuite/xqsuite-assertions-dynamic.xqm
```diff
@@ -0,0 +1,234 @@
...
+ : Dynamic Tests for XQSuite assertions.
+ :
+ : @author Adam Retter
+ :)
+module namespace xqtad = "http://exist-db.org/xquery/xqsuite/test/assertions-dynamic";
+
+(:~
+ : Resource functions to test through XQSuite.
+ :)
+declare namespace xqtares = "http://exist-db.org/xquery/xqsuite/test/assertions/resources";
```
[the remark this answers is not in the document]
adamretter: @line-o They do different things. The hint is in the filename!

In the dynamic one we only declare a namespace so we can later *dynamically* import a module of that namespace using `fn:load-xquery-module`. If I instead used `import module` then the module would already be imported statically and I would not be able to import it dynamically later!

exist-core/src/test/xquery/xqsuite/xqsuite-assertions-inline.xqm
adamretter: I am afraid I don't understand your comment. Can you rephrase/elaborate please?

adamretter:
> This PR contains several - welcome and good - changes that would warrant their own separate PRs

As the author of the PR who spent weeks chasing and trying to resolve this issue, an issue that went undetected for years, I strongly disagree. All of the fixes and in this PR are intimately inter-related to each other.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
adamretter: `$result` is of type `item()*`, so in the previous code:

```
if ($result)
```

This is getting the *effective boolean value* of `$result`. That is not the same as your proposed `fn:exists($result)` which would instead just tell you if the sequence is non-empty.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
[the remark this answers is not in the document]
line-o: Clarity. It's unnnecessary clutter.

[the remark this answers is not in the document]
adamretter: I think that is subjective, personally I have been writing all XQuery code for sometime now using the `fn` prefix as we have found here that it increases clarity rather than reducing it.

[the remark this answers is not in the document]
line-o: I will put it on the agenda for the next community meeting then.

exist-core/src/test/xquery/xquery3/parseIetfDate.xqm
line-o: I believe this issue is now tested elsewhere enough.

exist-core/src/test/xquery/xquery3/serialize.xql
line-o: Sorry but I have to disagree, @adamretter. This test is for adaptive serialization, if it does _not_ return a string than that is a problem. A reason to reject this PR. So please change it to a string so that we can see that it does in fact return a string and compares them correctly.

exist-core/src/test/xquery/xquery3/serialize.xql
line-o: We have to be able to test that a string is returned. If this happens to be castable to an integer value should not affect the assertion.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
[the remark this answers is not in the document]
line-o: Yeah, we need a definitive way forward. The entire codebase does not use the prefix so it is necessary to decide whether it is used _everywhere_ in the future or _never_.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
line-o: OK, I understand now. In that case it would help future readers of the source code to extract the evaluation of a safe effective boolean value into a properly named function.

```xquery
declare %private function test:safeEffectiveBooleanValue ($result as item()*) as xs:boolean {
    try {
        not(not(($result)))
    }
    catch * {
        false()
    }
};

```

exist-core/src/test/xquery/xquery3/serialize.xql
adamretter: @line-o It may itself be a test for adaptive serialization, but after that the assert-equals comes into play

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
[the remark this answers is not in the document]
adamretter: > The entire codebase does not use the prefix

If you grep for `fn:` you will find it in plenty of XQuery files within eXist-db!

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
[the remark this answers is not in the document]
line-o: How costly is this conversion from string to string in contrast to type check or type check + conversion?
Is this something we have hard numbers on?

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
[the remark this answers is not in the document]
line-o: The status quo is troubling, I agree.
This PR adds prefixes to existing lines of codes that are otherwise unchanged and the use of prefixes is consistent within the module itself.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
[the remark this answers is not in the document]
line-o: No answer for a week. Will have to be checked afterwards, then.

joewiz:
@adamretter Could you confirm this is a breaking change and needs to wait for 7.0.0?

adamretter:
> Could you confirm this is a breaking change and needs to wait for 7.0.0?

@joewiz Yes, this is a breaking change and requires 7.0.0.

adamretter:
> Why was the hot fix release for money necessary when this code is marked for 7.0.0? Why the rush?

@line-o Without the hotfix to Monex, Monex would not run on 7.0.0-SNAPSHOT. As several people are already using 7.0.0-SNAPSHOT builds and requested Monex... I fixed Monex to work with eXist-db 6 and 7.

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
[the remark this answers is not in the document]
adamretter: @line-o I assumed that if you were interested, you could benchmark it.

line-o:
> Without the hotfix to Monex, Monex would not run on 7.0.0-SNAPSHOT. As several people are already using 7.0.0-SNAPSHOT builds and requested Monex... I fixed Monex to work with eXist-db 6 and 7.

Surprising news! Where is the branch for the 7.0.0-SNAPSHOT that is used by several people?

exist-core/src/test/xquery/xquery3/serialize.xql
line-o: Which should ensure that the expected value _and_ type is returned.

reinhapa (approved):

line-o (commented):
There is still an open question on the adaptive serialization test where the expected value changed from a string to a number literal.
I would like to have proof that XQsuite is able to assert a string '1234' was returned.

adamretter:
> There is still an open question on the adaptive serialization test where the expected value changed from a string to a number literal.

As I tried to explain above there is a sequence of operations that happens. This test is not only testing that a specific string is returned from `fn:serialize` for Adaptive Serialization, but it is also testing the steps that happen after that.

1. A Literal *un-quoted* annotation with value `1234` is passed as the argument to the function, and XQSuite has to parse this and cast it to the correct sub-type of `xs:anyAtomicType`. In this test it will be parsed and cast to `xs:integer`.
2. The function under test is executed by XQSuite.
3.  The function under test eventually calls `fn:serialize` with the Adaptive Serialization mode set; `fn:serialize` returns an `xs:string`.
4. The function under test returns the `xs:string` value.
5. XQSuite needs to convert that `xs:string` value to a type and/or value that is appropriate for comparison to the expected value instructed by the `%test:assert`* annotation.
6. For this test the assert annotation has an *un-quoted` value of `1234`. XQSuite has to coerce this and the result of the function to a type that it can compare according to the XQuery specs for type promotion and type derivation. In this test it will be parsed and cast to `xs:integer`.
7. Finally the two `xs:integer` values are compared for equality.

I hope that helps explain the purpose of this test (which I think is the thing that is not being understood here). To summarise - it covers both the `xs:string` case that you are concerned about, as well as the actual `xs:integer` case that I am trying to test.

> I would like to have proof that XQsuite is able to assert a string '1234' was returned.

That is already present on that test, both implicitly through the explanation above that uses:
```xquery
%test:args(1234)
%test:assertEquals(1234)
```

and also explicitly by the additional assertions on the same function under test:
```xquery
%test:args('Hello "world"!')
%test:assertEquals('"Hello ""world""!"')
```

Additionally... I think you are already aware of many tests, including some written by yourself, within the full XQSuite of tests that already cover that use-case.

line-o:
The two tests in question do test the raw output of serialize (they are part of the spec of our implementation of `fn:serialize`). The function under test _is_ `fn:serialize`. Thus _have to_ return `xs:string` correct?

```xquery
declare
    %test:args(1234)
    %test:assertEquals(1234)
    %test:args('Hello "world"!')
    %test:assertEquals('"Hello ""world""!"')
function ser:adaptive-simple-atomic($atomic as xs:anyAtomicType) {
    ser:adaptive($atomic)
};

declare
    %test:args(1234)
    %test:assertEquals(1234)
    %test:args('Hello "world"!')
    %test:assertEquals('"Hello ""world""!"')
function ser:adaptive-simple-atomic-map-params($atomic as xs:anyAtomicType) {
    ser:adaptive-map-params($atomic)
};
```

adamretter:
> Thus have to return xs:string correct?

I am afraid you are still missing the point of my changes.
Yes, `fn:serialize` returns an `xs:string`, I am not disputing that, I even said above:

> 3. The function under test eventually calls fn:serialize with the Adaptive Serialization mode set; fn:serialize returns an xs:string.

However, that's not the point of these additions that I have made. As I said above:

> > To summarise - it covers both the xs:string case that you are concerned about, as well as the actual xs:integer case that I am trying to test.

Can we please move on?

line-o:
>  actual xs:integer case that I am trying to test.

The test are for `fn:serialize`! They _must_ stay testing that the _return value_ is `xs:string`, otherwise they do not make sense for anyone looking at them in a few months. They were changed in this PR for no apparent reason. Please let's move on by reverting the changes to those two tests.
If you want test type promotion capabilities of XQSuite,  please add them separately to an xqsuite specific testsuite.

adamretter:
@dizzzz @reinhapa Would you be able to spare some time to help @line-o understand what I am trying to achieve here please?

reinhapa:
> > actual xs:integer case that I am trying to test.
> 
> The test are for `fn:serialize`! They _must_ stay testing that the _return value_ is `xs:string`, otherwise they do not make sense for anyone looking at them in a few months. They were changed in this PR for no apparent reason. Please let's move on by reverting the changes to those two tests. If you want test type promotion capabilities of XQSuite, please add them separately to an xqsuite specific testsuite.

@line-o I can not see no problem with the changes from @adamretter  here. In my opinion the fix is as much as important as the test. As I understand it your requirements where covered as explained by Adam earlier. I see no point of changing again.

line-o:
@reinhapa a test that has to ensure the function under test returns the expected value and type.
Since the modified tests test `fn:serialize` it has to be `xs:string`. At the moment this is the case. In this PR the expected return type is changed to a literal `xs:decimal` (`"1234"` was changed to `1234`).
If this does not make a difference technologically it does change the meaning of the test semantically. 
In worst case it can lead to uncaught regressions.

PR #4623, opened 2022-11-28
[bugfix] add delay so timing of collections are subtle different

dizzzz:
Result of live community code review . chasing a recurring and unpredictable build error


line-o (approved):
I would add a comment why this sleep is necessary. Not super crucial but might be good to have. Other than that, good to go.

reinhapa (approved):

adamretter:
@dizzzz Would you mind providing some more information on this in the PR description please. I cannot find any minutes/notes in the Community Call about what this is for. Also for other people that don't attend/read the community call minutes, some background here might be useful.

joewiz:
Also, whenever we need to merge a PR before our agreed-upon 5-day review period (as I assume was the case here), the reason should be shared.

PR #4635, opened 2022-12-01
[bugfix] analyzer conf doesn't accept primitives

marmoure:
currently per docs the analyzer config should accept primitive types as `int` & `char`
and this was not the case 
when using `int` as param type in the analyzer config it gets converted to `Integer` then checks against the existing `int` constructor which yields `no constructor found` error
this fixes it 
thanks to @adamretter for providing the fix code

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/AnalyzerConfig.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/AnalyzerConfig.java
@@ -168,9 +168,9 @@ public class AnalyzerConfig {
             // Classname is defined.
 
             // Probe class
-            Class<?> clazz = null;
+            final Class<?> untypedClazz;
             try {
-                clazz = Class.forName(className);
+                untypedClazz = Class.forName(className);
 
             } catch (ClassNotFoundException e) {
                 LOG.error(String.format("Lucene index: analyzer class %s not found. (%s)", className, e.getMessage()));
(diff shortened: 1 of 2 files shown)


extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/AnalyzerConfig.java
```diff
@@ -513,7 +529,7 @@ static KeyTypedValue<?> getConstructorParameter(final Element param) throws Para
      * @param param The parameter-configuration element.
      * @return Set of parameter values
      */
-    private static Set<String> getConstructorParameterSetValues(Element param) {
+    private static Set<String>  getConstructorParameterSetValues(Element param) {
```
line-o, line 532: Was this space added intentionally?

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/AnalyzerConfig.java
```diff
@@ -597,6 +613,7 @@ public T getValue() {
 
         public Class<T> getValueClass() {
             return valueClass;
+
```
line-o, line 616: please remove surplus blank lines

line-o:
This PR adds two pom.xml files that I have seen in my local checkouts as well. I wonder what these are and if they should be added or rathre ignored. If they should be added, this should happen in a separate commit. A separate PR even.

line-o (approved):
Other than the formatting this looks good to me.

line-o:
Well, I would prefer the two pom files to be removed. And I would like to see the commits squashed before merging.

marmoure:
> Well, I would prefer the two pom files to be removed. And I would like to see the commits squashed before merging.

oh yes sorry about those i added them by mistake

marmoure:
@line-o  it should be good now

adamretter (commented):
LGTM

dizzzz (approved):

line-o:
We have two approvals. Should we wait until 6th of December before merging this (5 days of grace period)?

dizzzz:
no, it is OK for me.

PR #4636, opened 2022-12-02
Fix Memory leaks caused by XQuery context and update listener leaks

adamretter:
This PR fixes a memory leak in eXist-db that can quickly cause all available memory to be consumed.

This came about due to Listeners subscribing to events produced by a running XQuery never being reset between executions of the query. Thus the list of listeners per-query grew indefinitely. 

The cause of this was due to several XQuery expressions not being reset and cleaned up after the query was executed. In particular these XQuery functions were incorrectly implemented and therefore failed to clean up after themselves and could contribute to the memory leak:

* `inspect:module-functions`
* `inspect:module-functions-by-uri`
* `inspect:inspect-module`
* `inspect:inspect-module-uri`
* `util:eval`
* `util:compile`
* `util:module-info`
* `fn:load-xquery-module`

----
Thanks to @PieterLamers for reporting to me, and testing this in production at [John Benjamins Publishing](https://benjamins.com/content/home). Also thanks to @nverwer for his work investigating and testing this.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/util/Eval.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/util/Eval.java
@@ -438,7 +438,7 @@ public class Eval extends BasicFunction {
     private void cleanup(final XQueryContext evalContext, final XQueryContext innerContext, final DocumentSet oldDocs,
             final LocalVariable mark, final Item expr, final Sequence resultSequence) {
         if (innerContext != evalContext) {
-            innerContext.reset(true);
+            evalContext.addImportedContext(innerContext);
         }
 
         if (oldDocs != null) {
--- a/exist-core/src/main/java/org/exist/xquery/functions/util/ModuleInfo.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/util/ModuleInfo.java
@@ -223,14 +223,19 @@ public class ModuleInfo extends BasicFunction {
 		} else {
 			final ValueSequence resultSeq = new ValueSequence();
             final XQueryContext tempContext = new XQueryContext(context.getBroker().getBrokerPool());
-			for(final Iterator<Module> i = tempContext.getRootModules(); i.hasNext(); ) {
-				final Module module = i.next();
-				resultSeq.add(new StringValue(this, module.getNamespaceURI()));
-			}
-			if (tempContext.getRepository().isPresent()) {
-			    for (final URI uri : tempContext.getRepository().get().getJavaModules()) {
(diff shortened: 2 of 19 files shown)


exist-core/src/main/java/org/exist/xquery/functions/util/Eval.java
```diff
@@ -438,7 +438,7 @@ private Sequence doEval(final XQueryContext evalContext, final Sequence contextS
     private void cleanup(final XQueryContext evalContext, final XQueryContext innerContext, final DocumentSet oldDocs,
             final LocalVariable mark, final Item expr, final Sequence resultSequence) {
         if (innerContext != evalContext) {
-            innerContext.reset(true);
+            evalContext.addImportedContext(innerContext);
```
dizzzz, line 441: so effectively the reset is removed here.... it is reset indirect?

exist-core/src/main/java/org/exist/xquery/functions/util/ModuleInfo.java
```diff
@@ -223,14 +223,19 @@ public Sequence eval(Sequence[] args, Sequence contextSequence)
...
+				for (final Iterator<Module> i = tempContext.getRootModules(); i.hasNext(); ) {
+					final Module module = i.next();
+					resultSeq.add(new StringValue(this, module.getNamespaceURI()));
+				}
+				if (tempContext.getRepository().isPresent()) {
+					for (final URI uri : tempContext.getRepository().get().getJavaModules()) {
+						resultSeq.add(new StringValue(this, uri.toString()));
+					}
+				}
+			} finally {
```
dizzzz, line 236: in other code these two lines have different order. Is that on purpose? or does it not matter?

dizzzz (approved):
Two questions, LGTM. great work!

exist-core/src/main/java/org/exist/xquery/functions/util/Eval.java
adamretter: Yes, but the reset happens in a different way, and the purpose of it is more explicit.

exist-core/src/main/java/org/exist/xquery/functions/util/ModuleInfo.java
adamretter: Order here is not especially important

line-o:
This was merged before the five days grace period, with tests on windows/jdk11 failing, only one approval and no reason given.

dizzzz:
Hmmm you are right. I missed the failing test actually...

Issue #4670, opened 2023-01-09
[BUG] NPE in org.exist.xquery.functions.securitymanager.IdFunction


nverwer:
**Describe the bug**
In `/exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java` an NPE can occur when the `sm:id()` function is called. This happens when `context.getRealUser()` returns `null`.

**Expected behavior**
When `context.getRealUser()` returns null, the `<sm:real>` element should be empty, or absent.

**To Reproduce**
This happens when an XQuery script is called from a web application, after eXist has been restarted. However, I cannot come up with a simple reliable example.

**Context (please always complete the following information):**
 - OS: Windows
 - eXist-db version: 6.1
 - Java Version 8 and 11

**Additional context**
The NPE occurs in  
```
    private void subjectToXml(final MemTreeBuilder builder, final Subject subject) {
        builder.startElement(new QName("username", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
        builder.characters(subject.getName());
    ...
  }
```
Here, `subject` may be null, since is is obtained from `context.getRealUser()`, which may be null.

**Proposed fix**
I will make a PR, with a proposed solution avoiding the NPE.

adamretter:
@nverwer Thanks for this, I appreciate the sentiment. However, the issue you describe indicates a bug at a lower-level in eXist-db as the `realUser` in the `XQueryContext` should never be `null`.

Unfortunately this PR would simply hide that bug. I think hiding such bugs is not really what we should do, and so I am reluctant to merge this PR.

Perhaps it would be worth looking at all sites where an `XQueryContext` is initialised or reused, and make sure that `XQueryContext#setRealUser` is called at those sites?

nverwer:
@adamretter 
> looking at all sites where an XQueryContext is initialised or reused
That would be feasible if I would have a way to remote debug eXist with JDWP/JDB, but I never found a way to get that working.

There is no indication in `XQueryContext.java` that `realUser` should not be `null`, so a bug like this is likely to occur.
I will just surround calls to `sm:id()` with `try - catch` from now on...

line-o:
We should always guard against NPEs. To not hide bugs we could throw a more meaningful error.
Obviously it was possible to trigger this.

PieterLamers:
[...]

PR #4671, opened 2023-01-09
[bugfix] Fix for NPE in xquery.functions.securitymanager.IdFunction

nverwer:
### Description:

This fixes a NPE in the `sm:id()` function, that caused problems in one of my applications.

### Reference:

See https://github.com/eXist-db/exist/issues/4670 .

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
@@ -77,14 +77,20 @@ public class IdFunction extends BasicFunction {
 
             builder.startElement(new QName("id", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
 
-            builder.startElement(new QName("real", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
-            subjectToXml(builder, context.getRealUser());
-            builder.endElement();
+            final Subject realUser = context.getRealUser();
+            if (realUser != null) {
+              builder.startElement(new QName("real", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
+              subjectToXml(builder, realUser);
+              builder.endElement();
+            }
 
-            if (!sameUserWithSameGroups(context.getRealUser(), context.getEffectiveUser())) {
-                builder.startElement(new QName("effective", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
-                subjectToXml(builder, context.getEffectiveUser());
-                builder.endElement();
+            final Subject effectiveUser = context.getEffectiveUser();
+            if (effectiveUser != null) {
+              if (realUser == null || !sameUserWithSameGroups(realUser, effectiveUser)) {
+                  builder.startElement(new QName("effective", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
+                  subjectToXml(builder, effectiveUser);
+                  builder.endElement();
+              }
             }
 
             builder.endElement();
--- a/.github/actions/install-mvnd/action.yml
+++ b/.github/actions/install-mvnd/action.yml
@@ -1,0 +1,89 @@
+name: 'install-mvnd'
+description: 'Install the Maven Daemon'
+inputs:
+  version:
+    description: 'The version of the Maven Daemon to install'
+    required: true
+    default: '0.9.0'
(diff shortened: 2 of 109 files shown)


exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
```diff
@@ -77,14 +77,20 @@ private org.exist.dom.memtree.DocumentImpl functionId() {
...
+              subjectToXml(builder, realUser);
+              builder.endElement();
+            }
 
-            if (!sameUserWithSameGroups(context.getRealUser(), context.getEffectiveUser())) {
-                builder.startElement(new QName("effective", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
-                subjectToXml(builder, context.getEffectiveUser());
-                builder.endElement();
+            final Subject effectiveUser = context.getEffectiveUser();
+            if (effectiveUser != null) {
```
line-o, line 88: I would prefer to combine the nested conditions to

```
if (effectiveUser != null && (
  realUser == null || !sameUserWithSameGroups(realUser, effectiveUser)
)) {
```

line-o: Additionally the method `sameUserWithSameGroups` could handle null values for both realUser and effectiveUser

nverwer: I thought about that, but to me the conditions mean different things, and I kept them separate.
Either way, the result, and maybe the bytecode, is exactly the same.

I also thought about checking for `null` inside `sameUserWithSameGroups`, which might make the intention of the code clearer.

line-o: If both are null then it should return `false`.

line-o: This would simplify the condition to 
```
if (sameUserwithSameGroups(realUser, effectiveUser)) {...}
```

nverwer: :D I'll shut up until you have completed your review.

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
```diff
@@ -77,14 +77,20 @@ private org.exist.dom.memtree.DocumentImpl functionId() {
...
+              builder.startElement(new QName("real", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
+              subjectToXml(builder, realUser);
+              builder.endElement();
+            }
 
-            if (!sameUserWithSameGroups(context.getRealUser(), context.getEffectiveUser())) {
-                builder.startElement(new QName("effective", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
-                subjectToXml(builder, context.getEffectiveUser());
-                builder.endElement();
+            final Subject effectiveUser = context.getEffectiveUser();
```
line-o, line 87: As this is always run it can be moved up to line 81

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
line-o: I should have combined them into one instead of brain-dumping :)

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
dizzzz: Code checkers suggest to declare a variable as close as possible to where it is used. keep it here...

dizzzz (approved):
OK to me

reinhapa:
@nverwer can you rebase this PR once again?

nverwer:
@reinhapa I think that I rebased correctly. But I have to admit that I have become a bit lost in all the branches that seem relevant.

reinhapa:
@nverwer seems something is wrong. Did you rebased your changes on the current `exist/develop` of `eXistdb` ?

reinhapa (changes requested):
@nverwer please do a proper rebase on the the current `develop` branch. Seems that you rebased on your local `develop` branch that is not on the same state as the current `develop` from [eXist](https://github.com/eXist-db/exist)...

line-o:
Thank you @nverwer I will try to get your additions merged in the followup PR #5158

PR #4672, opened 2023-01-09
[bugfix] add compatibitity for slf4j v1

line-o:
Add the activation library for packages still using slf4j v1

### DIFF
--- a/exist-parent/pom.xml
+++ b/exist-parent/pom.xml
@@ -253,6 +253,12 @@
                 <version>${log4j.version}</version>
                 <scope>runtime</scope>
             </dependency>
+            <dependency>
+                <groupId>org.apache.logging.log4j</groupId>
+                <artifactId>log4j-slf4j-impl</artifactId>
+                <version>${log4j.version}</version>
+                <scope>runtime</scope>
+            </dependency>
 
             <dependency>
                 <groupId>org.slf4j</groupId>


line-o:
Older versions of monex would need that @dizzzz but as discussed in the community meeting this activation library is solely for the purpose to minimise friction for libraries still built with slf4japi v1

line-o:
re-running tests hoping that the CVE-database of dependency-check-maven is now working again.

dizzzz (approved):

adamretter:
I don't yet really understand why this is required.

> but as discussed in the community meeting this activation library is solely for the purpose to minimise friction for libraries still built with slf4japi v1

For everyone who was not at the community call, can you please tell us which libraries you are referring to?

line-o:
Packages can make use of slf4j-api that is provided by exist. Since this is not a major release we need to ensure that packages still relying on the version 1 of slf4j-api can still be used with exist 6.1.0 after its release.

adamretter:
@line-o Interesting. Can you tell me which packages you tested this against please?

adamretter:
@line-o Ok, but... can you tell me which packages you tested it with please?

line-o:
The above will not help to reinstate compatibility with packages still depending on slf4j-api v1.
see https://www.slf4j.org/faq.html#changesInVersion200

line-o:
We are investigating if this

Issue #4528, opened 2022-08-27
Service memory settings on Windows


vincentml:
After struggling with eXist running out of memory despite there being plenty of memory available, I've found a part of the Windows installer setup code that might need to be updated: 

https://github.com/eXist-db/exist/blob/f3424d7c44d2cc7d8eb722e0701709de6e497a34/exist-core/src/main/java/org/exist/launcher/WindowsServiceManager.java#L89-L90

I used `prunmgr.exe //ES//eXist-db` to inspect the details of the service that was created by the eXist-db Launcher. The memory settings specified in the launcher were not being used.

I used prunsrv-x86_64.exe to manually install a Service for eXist-db duplicating the same settings that were created by the Launcher with one change. The minimum and maximum memory values need to be provided as integers. For example:

 --JvmMs=1024 --JvmMx=4096

The eXist-db Launcher adds "m" to the value ( --JvmMs=1024m --JvmMx=4096m) but the "m" causes the value to not be used.

joewiz:
Which version of eXist and JDK?

dizzzz:
Hmmm, interesting as the JVM really should be started with -Xmx1024m and -Xmx4096m - this will fail on 32bit JVMs though

vincentml:
The eXist version is 6.0.1 and JVM is OpenJDK11U-jdk_x64_windows_hotspot_11.0.16_8

dizzzz:
Very recently there was an upgrade to commons-daemon v1.3.1 ([PR](https://github.com/eXist-db/exist/pulls?q=is%3Apr+is%3Aclosed+daemon+)) - [Changelog](https://commons.apache.org/proper/commons-daemon/changes-report.html)

I cant see yet if there is a relevant change there... but checking...

<img width="637" alt="image" src="https://user-images.githubusercontent.com/1700062/187380511-c1b6e905-a783-4712-82cf-eb341ff46688.png"> ([link](https://commons.apache.org/proper/commons-daemon/procrun.html))

you are indeed right...

PR #4685, opened 2023-01-13
Windows Service installer - prunsrv.exe expects memory to be supplied as an integer

adamretter:
Closes https://github.com/eXist-db/exist/issues/4528


line-o (approved):
Looks good to me. 
Regarding the security hotspot reported by SonarCloud: I cannot spot any backtracking in the  constructed regular expression.

Issue #4686, opened 2023-01-14
[BUG] installation issues cannot be separated


line-o:
**Describe the bug**

If an error is encountered running `deploy:install-and-deploy-from-db('/db/path/to/my.xar', 'https://my-repo/find')`
will always throw an error with `experr:EXPATH00`.

When caught its `$err:description` will always be empty.

This code is defined in `org.exist.xquery.modules.expathrepo.EXPathErrorCode` as 

```java
    public final static ErrorCode EXPDY007 = new EXPathErrorCode("EXPATH00", null);
```

On top of that the error codes do not follow the schema defined in the same class.

**EXPATH specific errors**

Stated: `EXP(DY|SE|ST)[0-9]{4}`
Acutal: `EXPATH[0-9]{2}`

Sadly, the error codes do not seem to be defined by the [EXPath pkg specification](http://expath.org/spec/pkg).
Does that actually mean we are free to use whatever codes we like?

**Expected behavior**

I would expect that specific error codes are used depending on the problem that was encountered that are already defined in the 
class and also used in repo-module implementation code.

```java
    public final static ErrorCode EXPDY001 = new EXPathErrorCode("EXPATH001", "Package not found.");
    public final static ErrorCode EXPDY002 = new EXPathErrorCode("EXPATH002", "Bad collection URI.");
    public final static ErrorCode EXPDY003 = new EXPathErrorCode("EXPATH003", "Permission denied.");
    public final static ErrorCode EXPDY004 = new EXPathErrorCode("EXPATH004", "Error in descriptor found.");
    public final static ErrorCode EXPDY005 = new EXPathErrorCode("EXPATH005", "Invalid repo URI");
    public final static ErrorCode EXPDY006 = new EXPathErrorCode("EXPATH006", "Failed to connect to public repo");
```

**To Reproduce**

Just to highlight one of the many problems below is a XQSuite test that tries to install a package from a db-local XAR that is non-existent.

```Xquery
xquery version "3.1";

module namespace t="http://exist-db.org/xquery/test";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare
    %test:assertError("experr:EXPATH001")
function t:test() {
    deploy:install-and-deploy-from-db('/db/path/to/my.xar', 'https://my-repo/find')
};
```

**Context (please always complete the following information):**
Build: eXist-6.1.0-SNAPSHOT (${build-commit})
Java: 1.8.0_352 (Azul Systems, Inc.)
[...]

Issue #4694, opened 2023-01-20
[BUG] poms of two modules show up as untracked after build and are not updated on release


line-o:
**Describe the bug**

When building exist I have two changed pom.xml files in my working directory.

- exist-core-jcstress/dependency-reduced-pom.xml
- exist-core-jmh/dependency-reduced-pom.xml

**Expected behavior**

Building exist from source does not add untracked files to the working directory. They should either be ignored or stay unchanged.

**To Reproduce**

1. checkout develop
2. run `mvn clean package`
3. run `git status`
4. the two poms mentioned above are shown as untracked files

**Screenshots**

When checking out the current master branch and building it both subprojects still show up as 6.1.0-SNAPSHOT (see screenshot below): 

<img width="639" alt="Screenshot 2023-01-20 at 15 16 18" src="https://user-images.githubusercontent.com/657222/213721716-dcf81aaa-f73f-4cec-a0cb-8024ea4685d7.png">

**Context (please always complete the following information):**
- Build: eXist-6.1.0-SNAPSHOT (b8e73493b95e252dbe09f4b3be1833bd4c62371e)
- Java: 1.8.0_352 (Azul Systems, Inc.)
- OS: Mac OS X 12.6.2 (aarch64)

**Additional context**
- How is eXist-db installed? built from source (develop, master)
- Any custom changes in e.g. `conf.xml`? none

line-o:
This seems to even cause the test workflow in CI to fail see https://github.com/eXist-db/exist/actions/runs/3940322485/jobs/6741277473#step:4:21

joewiz:
It appears that the 6.1.0 release didn't touch the poms of `exist-core-jcstress` or `exist-core-jmh`.

Compare the files listing of the [6.0.0](https://github.com/eXist-db/exist/commit/70875a0bb0c6e3d03324c01db984100bd8c32fb0) and [6.1.0](https://github.com/eXist-db/exist/commit/f90ddb547162d205b293a982bbe989e39651c605) releases.

joewiz:
My results when running `mvn package` on develop HEAD:

> <img width="642" alt="Screenshot 2023-01-20 at 20 52 18" src="https://user-images.githubusercontent.com/59118/213835830-83cf0ca1-e159-44b9-b273-551a407d50d0.png">

joewiz:
[...]

PR #4695, opened 2023-01-21
[fix] permissions on install

line-o:
### Description:

Xquery modules and collections have to be set as executable on installation if the mode set in conf.xml or in the permission element in repo.xml does not set it. With this PR applied this will be limited to the owner if the group or others cannot read or write the resource or collection.

### Reference:

refs #4686
Completes a TODO in the code.

### Type of tests:

Junit-Tests

### DIFF
--- a/exist-core/src/main/java/org/exist/security/AbstractUnixStylePermission.java
+++ b/exist-core/src/main/java/org/exist/security/AbstractUnixStylePermission.java
@@ -23,6 +23,7 @@ package org.exist.security;
 
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.exist.util.SyntaxException;
@@ -40,7 +41,7 @@ public abstract class AbstractUnixStylePermission implements Permission {
 
     /**
      * The symbolic mode is described by the following grammar:
-     *
+     * <p>
      * mode         ::= clause [, clause ...]
      * clause       ::= [who ...] [action ...] action
      * action       ::= op [perm ...]
@@ -54,15 +55,15 @@ public abstract class AbstractUnixStylePermission implements Permission {
         //TODO expand perm to full UNIX chmod i.e. perm ::= r | s | t | w | x | X | u | g | o
 
         final String[] clauses = symbolicMode.split(",");
-        for(final String clause : clauses) {
+        for (final String clause : clauses) {
             final String[] whoPerm = clause.split("[+\\-=]");
 
             int perm = 0;
             boolean uidgid = false;
             boolean sticky = false;
             //process the op first
-            for(final char c : whoPerm[1].toCharArray()) {
-                switch(c) {
+            for (final char c : whoPerm[1].toCharArray()) {
+                switch (c) {
                     case READ_CHAR:
                         perm |= READ;
                         break;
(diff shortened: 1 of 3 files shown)


line-o:
tests failures related to #4694

reinhapa (changes requested):
Could the formatting and the actual fix be splitt into two separate PRs? Also there are two POM files where the version is still not updated (I created a separate PR #4696 for this issue)

line-o:
@reinhapa that's why I reference #4694

line-o:
I already separated the formatting fix into a separate commit. This is the agreed way, I believe. If you want to just examine the changes please look at the commits 6f52c590572b2fc4c919cfa65c4668cab95ff0f8, e83a2b775749651ee75602dc9a5c486311449c87 and 836d238c62bce6a2b71da6ee6d8bd4fb955e6400

exist-core/src/main/java/org/exist/security/AbstractUnixStylePermission.java
```diff
@@ -356,65 +349,97 @@ public static String modeToSimpleSymbolicMode(final int mode) {
...
             builder.append(EXECUTE_CHAR);
         } else {
             builder.append(UNSET_CHAR);
         }
-        
+
         return builder.toString();
     }
+
+    static int ownerExecute = EXECUTE << 6;
```
dizzzz, line 414: 😵

line-o: ?

adamretter:
We really should not do this. This sets things executable that it is the responsibility of the user or package maintainer to do. We cannot and should not assume that things should just be set executable for convenience.

line-o:
@adamretter What do you mean by "we should not do this"? Keep it as it is? 
8 years ago you added this TODO - This PR will add the executable bit _only_ to groups and users if they have read permissions. What is a collection worth if it is created with read but without executable rights?

adamretter:
>  Keep it as it is

No my preference would be to remove the security problem that was added where it should never have been added, i.e. this comment of mine:

```java
mode = mode | 0111;     //TODO(AR) Whoever did this - this is a really bad idea. You are circumventing the security of the system
```

line-o:
Sorry, I could not find a solution there. I provide one which might not be your preferred solution but removing this line will break more than it solves.

reinhapa:
> > Keep it as it is
> 
> No my preference would be to remove the security problem that was added where it should never have been added, i.e. this comment of mine:
> 
> ```java
> mode = mode | 0111;     //TODO(AR) Whoever did this - this is a really bad idea. You are circumventing the security of the system
> ```
@adamretter how would you remove this security problem in your opinion? Remove that line completely in the next major version and document as such?

line-o:
That is not a viable solution @adamretter 
Please provide a path forward that leaves installed applications usable.

reinhapa:
> That is not a viable solution @adamretter @reinhapa Please provide a path forward that leaves installed applications usable.

@line-o  What exactly is not a viable solution? Do not touch any access settings? From the discussion in the last community talk I thought, that this code is only executed on installation time... So from that point of view, that change would only have an impact if you install new XARs that have improper security definitions. Or do I not get the point here?

line-o:
There is no other way to set collections as executable on installation by default. The same is true for XQuery scripts.
There is a reason this line is there since eight years despite the scary comment.

reinhapa:
> There is no other way to set collections as executable on installation by default. The same is true for XQuery scripts.

So you mean there is no way to define the according access definition with a XAR file? Is this what you are trying to say? In that case I would propose to extend the XAR API in order to support  the access settings and only using the existing fallback method if no explicit definition is available to support backwards compatibility then...

line-o:
@reinhapa let me reiterate:

case 1:  a XAR without any application specific permissions set in repo.xml 
the default permissions (`rw-r--r--` at the time of writing) are applied to all resources and collections extracted from the xar. Leaving this unchanged will even grant the owner (admin/DBA by default) the ability to open any of its collections nor execute any of the xquery scripts.

case 2: a XAR with a permission element in repo.xml
The permissions element (I only have seen applications with a single one, including bundled applications) can set the permissions for a single owner and group and can have one mode which will be applied to all collections and resources.
Let's say this is even more restrictive and set to `r--r-----`. Previously this would have set the mode for collections and xquery scripts to `r-xr-x--x` and with this PR applied it is `r-xr-x---`. That is an improvement in security. Doing this does result in the application not being accessible for unauthenticated users, though. They will not even be able to access the controller.xq or a login page. 

Having both collections and xquery scripts executable for database users who can also read or write the same resource is reasonable. It _must_ be possible in a straightforward way for the vast majority of users of our platform to end up with a working application. Any other solution must provide this path.

line-o:
> I would propose to extend the XAR API in order to support the access settings and only using the existing fallback method if no explicit definition is available to support backwards compatibility then...

I agree, having fine-grained control over permissions at installation time is the way forward. But this will take time to introduce as we have not even started yet discussing it and is clearly out of scope of this PR. This fallback should still then use `safeSetExecutable`. So it does not touch the change introduced in this PR. I am convinced this change is a step in the right direction.

If we keep this preferably declarative permissions setting logic within repo.xml this can be done without having to change the expath packaging standard. I can open a feature-request in the following days so that we do not forget about it.

joewiz:
During last week and this week's Community Call, we discussed eXist's current behavior and what this PR changes. We understand that this PR is very narrow in focus. It does not break existing packages. It only gives package creators a new ability to control the "executable" flag on collections and XQuery files via the repo.xml. If packages don't specify this (i.e., any existing package), permissions will fall back on the defaults specified in conf.xml. These defaults align with what's hardcoded for package installation now. So effectively this doesn't change the current defaults. 

We concluded that the PR is acceptable.

The only thing remaining is that the existing documentation should be surveyed for necessary edits / expansion.

PR #4696, opened 2023-01-22
[bugfix] corrects version number on missed POMs

reinhapa:
### Description:
Fixes the version numbers in missed POM files after the last release.

### DIFF
--- a/exist-core-jcstress/pom.xml
+++ b/exist-core-jcstress/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.exist-db</groupId>
         <artifactId>exist-parent</artifactId>
-        <version>6.1.0-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


line-o:
Thanks for updating the two poms ( there is an open issue for that #4694 ).

I would like to point out that rather than having to fix it manually each time we release exist I would like to see a change that updates the maven process that updates those two poms like all the other ones.

line-o (approved):
This is still a helpful hot fix for us to continue working on the develop branch.
We still need to update the maven release configuration to do this automatically.

dizzzz (approved):

line-o (changes requested):
These two files are built artefacts and should be removed and ignored from version control.

line-o (approved):
I just realized, that this is not about the dependency-reduced-poms.
We still need to find out how to include the real POMs into the maven release process.

PR #4698, opened 2023-01-23
bump version to 7.0.0-SNAPSHOT

line-o:
In preparation of merging in breaking changes this updates the version number and the run configuration.

### DIFF
--- a/extensions/security/iprange/pom.xml
+++ b/extensions/security/iprange/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.exist-db</groupId>
         <artifactId>exist-parent</artifactId>
-        <version>6.2.0-SNAPSHOT</version>
+        <version>7.0.0-SNAPSHOT</version>
         <relativePath>../../../exist-parent</relativePath>
     </parent>
 
--- a/.idea/runConfigurations/Java_Admin_Client.xml
+++ b/.idea/runConfigurations/Java_Admin_Client.xml
@@ -5,7 +5,7 @@
     <option name="MAIN_CLASS_NAME" value="org.exist.start.Main" />
     <module name="exist-distribution" />
     <option name="PROGRAM_PARAMETERS" value="client" />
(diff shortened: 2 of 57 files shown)


reinhapa (approved):

joewiz:
CI is looking good so far... I'm curious, was `mvn release:update-versions -DautoVersionSubmodules=true` (i.e., the process at https://maven.apache.org/maven-release/maven-release-plugin/examples/update-versions.html) used to apply the changes in the version bump commit?

line-o:
The JavaDoc 8 job fails with:

> Error:  Failed to execute goal on project exist-service: Could not resolve dependencies for project org.exist-db:exist-service:jar:7.0.0-SNAPSHOT: org.exist-db:exist-start:jar:7.0.0-SNAPSHOT was not found in https://repo.evolvedbinary.com/repository/exist-db-snapshots/ during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of exist-db-snapshots has elapsed or updates are forced -> [Help 1]

I wonder why the other jobs are not affected.

line-o:
Apparently `exist-docker/pom.xml` was not updated. Maybe `-DautoVersionSubmodules=true` is necessary after all.

line-o:
I just retested `exist-docker/pom.xml` is not updated regardless if `-DautoVersionSubmodules=true` is set or not.

extensions/security/iprange/pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.exist-db</groupId>
         <artifactId>exist-parent</artifactId>
-        <version>6.2.0-SNAPSHOT</version>
+        <version>7.0.0-SNAPSHOT</version>
```
dizzzz, line 31: will give merge issues as module is to be removed...

dizzzz (approved):

line-o:
From the community call we found out that `mvn release:update-versions -Pdocker` will update all the POM files.

joewiz:
The changes in this PR constitute one important step in the remaining steps for advancing to a 7.0.0 develop, but there is another one. The other key step is to advance the milestones, as described in https://github.com/eXist-db/exist/blob/develop/exist-versioning-release.md#publishingpromoting-the-product-release:

> 13. Go to GitHub and move all issues and PRs which are still open for the release milestone to the next release milestone. Close the release milestone.

joewiz:
In addition, via @adamretter DM, quoted with permission:

> If possible I would prefer this pause until I can tackle it, some things look off and I think it likely we need a 6.1.1 immediately before 7.0.0 to address the two reported issues since 6.1.0.

reinhapa:
> In addition, via @adamretter DM, quoted with permission:
> 
> > If possible I would prefer this pause until I can tackle it, some things look off and I think it likely we need a 6.1.1 immediately before 7.0.0 to address the two reported issues since 6.1.0.

@joewiz @adamretter will the outstanding fixes for 6.1.1 not need an 6.1 branch  anyway? Or is it more convenient to create that release on the current `develop` branch to not have to forward/backward port?

line-o:
This PR is no longer needed since the change was force-pushed to develop yesterday night: https://github.com/eXist-db/exist/commit/55523020242381a79bc78bb8c7e818c08f9c3b68

Issue #3448, opened 2020-06-16
[BUG] RESTXQ infinite loop with circular dependency 


joewiz:
**Describe the bug**

It is possible to induce RESTXQ to fail a dependency check and cause system instability.

**Expected behavior**

A failed dependency check by RESTXQ should be registered once, and not cause an infinite loop, log bloat, and system instability. 

**To Reproduce**

Save the following query to `/db/test.xqm`:

```xquery
xquery version "3.1";

module namespace test = "test";
```

Notice no problems in exist.log or restxq.log.

Now, add the following line to the query and save it a 2nd time:

```xquery
import module namespace test = "test" at "test.xqm";
```

Again, notice no problems in exist.log or restxq.log. 

Now, delete the added line, and save the query a 3rd time. 

Now, notice the following two entries in restxq.log:

> 2020-06-16 17:14:38,175 [qtp554346832-55] WARN  (ExistXqueryRegistry.java [recordMissingDependency]:428) - Module '/db/test.xqm' has a missing dependency on '/db/test.xqm'. Will re-examine if the missing module is added. 
> 2020-06-16 17:14:38,179 [qtp554346832-55] INFO  (ExistXqueryRegistry.java [reexamineModulesWithResolvedDependencies]:321) - Missing dependency '/db/test.xqm' has been added to the database, re-examining '/db/test.xqm'... 

In addition, the 2nd log line ("reexamineModules...") is repeated 1312 times in restxq.log. 

In addition, the store fails, with the following error, reported in the Java admin client:

```txt
Failed to invoke method storeBinary in class org.exist.xmlrpc.RpcConnection: null

org.xmldb.api.base.XMLDBException: Failed to invoke method storeBinary in class org.exist.xmlrpc.RpcConnection: null
	at org.exist.xmldb.RemoteCollection.execute(RemoteCollection.java:127)
	at org.exist.xmldb.RemoteCollection.store(RemoteCollection.java:545)
	at org.exist.xmldb.RemoteCollection.storeResource(RemoteCollection.java:508)
	at org.exist.xmldb.RemoteCollection.storeResource(RemoteCollection.java:469)
	at org.exist.client.QueryDialog.saveToDb(QueryDialog.java:148)
	at org.exist.client.QueryDialog.lambda$1(QueryDialog.java:168)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
[...]

PR #4699, opened 2023-01-24
[fix] disable global restXQ trigger

line-o:
### Description:

This PR disables the restXQ trigger in collection.xconf.init. This is a breaking change as applications relying on this trigger to
be activated globally will have to add this trigger to their application's collection.xconf.

In my testing I was able to add two separate applications with both having the restXQ-trigger activated in their own
collection.xconf (both were based on https://gist.github.com/joewiz/28dd9b8454d14b4164a0). Modifying the code
showed up immediately when the exported routes were called with `/exist/restxq/{route}`. 

### Reference:

This trigger has led to several severe problems in the past with heap-memory exceptions when installing XAR-packages and 
hanging databases. There is also an unsolved issue with modules with circular dependencies.

- related #3448

### Type of tests:

In order to replicate my test you can check out this PR, build it and then create one or more test-applications using the template below.

```xquery
xquery version "3.1";

module namespace my-app = "http://my/app";

(: 
    Save this file into eXist anywhere (e.g., /db/restxq-template.xqm or 
    /db/apps/my-app/modules/restxq-template.xqm).
    
    Note that the collection configuration file where you store the module must invoke
    the restXQ trigger.

    Add a collection.xconf next to your script (or in any parent collection).
    It has to have the following trigger configuration (along with any other configurations).

            <triggers>
                <trigger class="org.exist.extensions.exquery.restxq.impl.RestXqTrigger"/>
            </triggers>
        </collection>
    
    run the handler with GET http://localhost:8080/exist/restxq/test
:)

declare namespace rest = "http://exquery.org/ns/restxq";

declare
    %rest:GET
    %rest:path("/test")
    %output:media-type("text/html")
    %output:method("html5")
function my-app:hello() {
    let $title := 'Hello world!'
    return
        <html xmlns="http://www.w3.org/1999/xhtml">
            <head>
                <title>{$title}</title>
            </head>
            <body>
                <h1>{$title}</h1>
            </body>
        </html>
};
```

### DIFF
--- a/exist-distribution/src/main/config/collection.xconf.init
+++ b/exist-distribution/src/main/config/collection.xconf.init
@@ -1,6 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <collection xmlns="http://exist-db.org/collection-config/1.0">
     <triggers>
-        <trigger class="org.exist.extensions.exquery.restxq.impl.RestXqTrigger"/>
+        <!--
+            The global restXQ trigger is now disabled by default
+            you can re-enable it here or just add the trigger to your applications
+            collection.xconf.
+            This was done due to the negative impact of this trigger on overall
+            database performance and memory footprint when installing packages.
+        -->
+        <!--<trigger class="org.exist.extensions.exquery.restxq.impl.RestXqTrigger"/>-->
     </triggers>
 </collection>


adamretter:
As I previously stated to you @line-o in the public eXist-db Slack, I will not accept this change

line-o:
I know

line-o:
Here you have the chance to argue why this is still needed or what I have not thought about. Just not accepting it is a little shallow of an argument.

adamretter:
Would you like me to copy and paste my arguments from Slack?

line-o:
> Would you like me to copy and paste my points from Slack?

Yes, please do so.

reinhapa:
> As I previously stated to you @line-o in the public eXist-db Slack, I will not accept this change

Hi @adamretter,
I tried to find the discussion about this but was not able to find it anymore. Could you explain what the problem is in your opinion as I do not really seem understand the actual problem with this change.

dizzzz:
@adamretter  in exist index configuration for a collection is taken from the collection.xconf for that collection, or when missing the one from the most close by parent collection.xconf .
How does that work for triggers?

duncdrum:
@line-o there is a test for docker using restxq. This will need to modified to either activate it by using a custom `conf.xml` for that temporary container; or disabled.

line-o:
@duncdrum Yes I saw that. It would make better to adapt it. But before that I would like to come to an agreement if this is the way forward, which is the more complex part.

windauer:
@adamretter in Slack you wrote that the people who are not using restxq should simply disable it. Problem with this approach is imho that endusers will not know whats the cause of their issues and that they should search for "disabling restxq". They will notice heavy RAM usage only due to installing a XAR (and eXist-db not freeing the RAM afterwards, even when you trigger a GC via Monex). Or that the installation of [frus](https://github.com/HistoryAtState/frus) takes 40min with a default eXist-db while it would take only 20min on the same machine with restxq disabled. In worst case they even step away from eXist-db without knowing that their issue could have simply been solved by disabling the restxq trigger. 
 
I would be all in for adding a "enable restxq" trigger to the app generator @duncdrum is maintaining and to add the information how and where to enable restxq prominent in the documentation but to disable it by default.

line-o:
@StephanMa are you using restxq - annotating functions to route http traffic - or the REST-like API of exist?
The easiest way to distinguish them is by the URLs you are calling 
- `[/exist]/restxq/...` -> restxq
- `[/exist]/rest/...` -> REST-like API
This is about restxq and does not disable the REST-like API of exist.

duncdrum:
See https://github.com/eXist-db/exist/pull/4855#issuecomment-1500203280

adamretter:
#4855 is unrelated to this!

PR #4661, opened 2022-12-22
Compare this XQTS run with previous XQTS run

adamretter:
Attempt at comparing results of previous XQTS run against this run.
See - https://github.com/eXist-db/exist/issues/3632

----
This open source contribution to the eXist-db project was commissioned by the Office of the Historian, U.S. Department of State, https://history.state.gov/.

PR #4735, opened 2023-02-07
CI: test xqts workflow

### LINKED ISSUE #4707: XQTS workflow no longer working
Judging from the output of the most recent run of our XQTS workflow 
https://github.com/eXist-db/exist/actions/runs/4024734250/jobs/6917098711

the runner throws an uncaught exception at the very beginning 
see https://github.com/eXist-db/exist/actions/runs/4024734250/jobs/6917098711#step:5:8

and no report is generated
and https://github.com/eXist-db/exist/actions/runs/4024734250/jobs/6917098711#step:6:10

**NOTE** The workflow reports success without producing a report.

line-o:
closes #4707
By adding a new step into the XQTS workflow that checks for the existence of the temporary folder
the workflow now signals failure if no output was created.
This approach was discussed and agreed on the community call on 6.2.2023

### DIFF
--- a/.github/workflows/ci-xqts.yml
+++ b/.github/workflows/ci-xqts.yml
@@ -22,8 +22,9 @@ jobs:
         env:
           JAVA_OPTS: -XX:+UseG1GC -XX:+UseStringDeduplication
         run: find exist-xqts/target -name exist-xqts-runner.sh -exec {} --xqts-version HEAD --output-dir /tmp/xqts-output \;
+      - name: Check XQTS Logs
+        run: ls /tmp/xqts-output 
       - name: Archive XQTS Logs
-        if: always()
         uses: actions/upload-artifact@v2
         with:
           name: xqts-logs


adamretter:
I am afraid that this is not really following the purpose of why I created the XQTS workflow. As described here in your issue:

> That is intentional as the desired workflow design has not yet been completed. I have been actively working on this https://github.com/eXist-db/exist/pull/4661. Please remain calm and give me the necessary time to finish that work in https://github.com/eXist-db/exist/pull/4661.

I am still working on this.

It's wonderful to see you excited about contributing to eXist-db. May I suggest that one of the other [464 issues](https://github.com/eXist-db/exist/issues) would welcome your energy and input

line-o:
This action will need to be discussed next Monday.

reinhapa:
> I am afraid that this is not really following the purpose of why I created the XQTS workflow. As described here in your issue:
> 
> > That is intentional as the desired workflow design has not yet been completed. I have been actively working on this #4661. Please remain calm and give me the necessary time to finish that work in #4661.
> 
> I am still working on this.
> 
> It's wonderful to see you excited about contributing to eXist-db. May I suggest that one of the other [464 issues](https://github.com/eXist-db/exist/issues) would welcome your energy and input

@adamretter I my opinion, it would be good to have a failing work flow if and only if we have **no test results** as in the sense of not even one test has passed and produced any output. This would mean that if only one test result is available the job would pass, even though we might faile rate of 99% in this case.

line-o:
@reinhapa and the simple solution to list the contents of the output folder with `/tmp/xqts-output` achieves exactly that.

duncdrum:
@adamretter this here:
> It's wonderful to see you excited about contributing to eXist-db. May I suggest that one of the other [464 issues](https://github.com/eXist-db/exist/issues) would welcome your energy and input

is massively inappropriate

dizzzz:
so from the teleconf meeting: this change will (should) result in a visible XQTS issue when there are _no_ results available, due to a low-level java issue (class cloader during initialization, etc... the app can not start at all).

If the start is ok, and eg only one test is executed, this check will not issue an error.

.github/workflows/ci-xqts.yml
```diff
@@ -22,8 +22,9 @@ jobs:
         env:
           JAVA_OPTS: -XX:+UseG1GC -XX:+UseStringDeduplication
         run: find exist-xqts/target -name exist-xqts-runner.sh -exec {} --xqts-version HEAD --output-dir /tmp/xqts-output \;
+      - name: Check XQTS Logs
+        run: ls /tmp/xqts-output
```
dizzzz, line 26: so this means...... (1) no directory is created or (2) no files in the directory?

line-o: `ls` will fail with exit code 2 if the directory is not found. If the directory is created but has no contents it will output `.` and `..` I believe. So there is room for improvement.

line-o: Maybe `ls /tmp/xqts-ouput/**/*.xml` is what we really want. As we expect the zip-file to contain
a subfolder with a bunch of test results in XML files.

reinhapa (approved):

.github/workflows/ci-xqts.yml
line-o: But since I am unsure what the actual contents of that folder will be I am hesitant to go this step right away.

dizzzz (approved):

PR #4765, opened 2023-02-23
Lock object optimisation

adamretter:
Backported from FusionDB - improvements in the area of Collection and Document locks - some small reductions in the number of object allocations, less array copies, and ultimately less GC.

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/lock/ManagedSingleLockDocumentLock.java
+++ b/exist-core/src/main/java/org/exist/storage/lock/ManagedSingleLockDocumentLock.java
@@ -32,9 +32,10 @@
  */
 package org.exist.storage.lock;
 
-import jakarta.annotation.Nullable;
 import org.exist.xmldb.XmldbURI;
 import uk.ac.ic.doc.slurp.multilock.MultiLock;
+
+import javax.annotation.Nullable;
 
 /**
  * @author <a href="mailto:adam@evolvedbinary.com">Adam Retter</a>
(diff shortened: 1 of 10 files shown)


exist-core/src/main/java/org/exist/storage/lock/ManagedSingleLockDocumentLock.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private ManagedSingleLockDocumentLock(final XmldbURI documentUri) {
+        this(documentUri, -1, null, Lock.LockMode.NO_LOCK, null);
+    }
+
+    @Override
+    public void close() {
+        if (!closed) {
+            if (lock != null) {  // NOTE(AR) only null when constructed from {@link #notLocked(XmldbURI)}.
```
dizzzz, line 63: 62/63 could be combined? maybe null check is faster than value comparison?

dizzzz:
certain parts of code are much more readable, great.
Are there any new class specific tests missing?

dizzzz (changes requested):
build fails....

exist-core/src/main/java/org/exist/storage/lock/ManagedSingleLockDocumentLock.java
adamretter: Thanks @dizzzz I just pushed a commit to adjust that as you suggested.

adamretter:
> Are there any new class specific tests missing? I guess that these classes are all implicitly tested?

The new classes are just replacements to existing implementations, and so they they are tested by the existing tests.

adamretter:
> build fails....

Should be fixed now I hope.

reinhapa (approved):

dizzzz (approved):

adamretter:
Thanks @dizzzz

PR #4773, opened 2023-02-28
[test] fix and add adaptive serialization tests

line-o:
### Description:

fn:serialize tests with method adaptive _have_ to return strings

### Reference:

reinstate tests modified in #4622 and add two extra

### Type of tests:

XQSuite tests

### DIFF
--- a/exist-core/src/test/xquery/xquery3/serialize.xql
+++ b/exist-core/src/test/xquery/xquery3/serialize.xql
@@ -254,7 +254,9 @@ function ser:serialize-empty-sequence() {
 
 declare
     %test:args(1234)
-    %test:assertEquals(1234)
+    %test:assertEquals('1234')
+    %test:args('1234')
+    %test:assertEquals('"1234"')
     %test:args('Hello "world"!')
     %test:assertEquals('"Hello ""world""!"')
 function ser:adaptive-simple-atomic($atomic as xs:anyAtomicType) {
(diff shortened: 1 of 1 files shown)


reinhapa (approved):

adamretter:
I would suggest adding extra test assertions if you need them instead of removing the numeric assertions. The numeric assertions are there to ensure that the result of fn:serialize can be correctly handled by XQSuite. I would be happy to see more tests/assertions; I fear that removing the numeric assertions will cause an important feature of serialization in XQSuite to no longer be tested.

reinhapa:
> I fear that removing the numeric assertions will cause an important feature of serialization in XQSuite to no longer be tested.

How is this concern covered currently? I do not exactly understand how this is addressed currently.

adamretter:
@reinhapa So currently the test covers both string and numeric assersions by the annotations:

```xquery
    %test:args(1234)
    %test:assertEquals(1234)
    %test:args('Hello "world"!')
    %test:assertEquals('"Hello ""world""!"')
```

However, if we change `%test:args(1234)` to `%test:args('1234')`, then we are no longer (a) testing the numeric realisation of `fn:serialize` by XQSuite, and (b) it becomes a duplicate of ` %test:args('Hello "world"!')` as a string is just a string!

line-o:
@adamretter `%test:args` annotation is left unchanged.

exist-core/src/test/xquery/xquery3/serialize.xql
```diff
@@ -254,7 +254,9 @@ function ser:serialize-empty-sequence() {
 
 declare
     %test:args(1234)
-    %test:assertEquals(1234)
+    %test:assertEquals('1234')
```
line-o, line 257: Please!

adamretter:
> @adamretter `%test:args` annotation is left unchanged.

@line-o Yes, but the `%test:assertEquals` has been changed! It should not be. The same explanation as I gave above applies to each of them.

line-o:
@adamretter the numeric realisation of `fn:serialize`, what is that exactly?

The signature is `fn:serialize($arg as item()*, $params as item()?) as xs:string` (see https://www.w3.org/TR/xpath-functions-31/#func-serialize)

line-o:
I would argue that XQSuite _must_ fail for functions that return a string but the assert annotation is a numeric literal.
It currently does not. I will open an issue on that.

PR #4777, opened 2023-03-01
[test] refactor serialization suite

line-o:
### Description:

Refactor serialization tests for maintainability and readability.

- combine item-separator with method tests into one
    `ser:item-separator-with-method`
- simplify test cases that used eq in combination with assertXPath
- make direct calls to function under test (serialize) in each test
- extract fixtures to module variables
- remove namespace from children of output:serialization-parameters
- convert QNames keys to strings for exist-db specific options
- convert string constructors to CDATA sections
- make HTML5 test data real world examples
- use unprefixed form of built-ins
- fix formatting

### Reference:

has to wait for

- #4773 
- #4816

### Type of tests:

XQSuite

### DIFF
--- a/exist-core/src/test/xquery/xquery3/serialize.xql
+++ b/exist-core/src/test/xquery/xquery3/serialize.xql
@@ -25,6 +25,8 @@ module namespace ser="http://exist-db.org/xquery/test/serialize";
 
 declare namespace test="http://exist-db.org/xquery/xqsuite";
 
+declare namespace output="http://www.w3.org/2010/xslt-xquery-serialization";
+
 declare %private variable $ser:opt-map-html5 :=
     map {
         "method": "html",
@@ -33,19 +35,17 @@ declare %private variable $ser:opt-map-html5 :=
 ;
 
 declare %private variable $ser:opt-xml-adaptive-no-indent :=
-    <output:serialization-parameters
-        xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
-        <method>adaptive</method>
-        <indent>no</indent>
+    <output:serialization-parameters>
+        <output:method>adaptive</output:method>
+        <output:indent>no</output:indent>
     </output:serialization-parameters>
 ;
 
 declare %private function ser:opt-xml-with-separator($item-separator as xs:string) {
-    <output:serialization-parameters
-            xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
-        <item-separator>{$item-separator}</item-separator>
-        <method>adaptive</method>
-        <indent>no</indent>
+    <output:serialization-parameters>
+        <output:item-separator>{$item-separator}</output:item-separator>
+        <output:method>adaptive</output:method>
+        <output:indent>no</output:indent>
     </output:serialization-parameters>
 };
 
@@ -207,10 +207,9 @@ declare
 function ser:serialize-with-params() {
     serialize(
         $ser:atomic,
-        <output:serialization-parameters
-               xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
-            <method value="xml"/>
-            <indent value="yes"/>
+        <output:serialization-parameters>
+            <output:method value="xml"/>
+            <output:indent value="yes"/>
         </output:serialization-parameters>
     )
 };
@@ -226,8 +225,7 @@ declare
     %test:assertXPath("contains($result,'atomic')")
 function ser:serialize-no-method() {
     serialize($ser:atomic,
-        <output:serialization-parameters
-                xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization" />)
+        <output:serialization-parameters />)
 };
 
 declare
@@ -260,9 +258,8 @@ declare
     %test:assertEquals("aaabbb")
 function ser:serialize-atomic-empty-separator-xml-options() {
     serialize(("aaa", "bbb"),
-        <output:serialization-parameters
-            xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
-            <item-separator value=""/>
+        <output:serialization-parameters>
+            <output:item-separator value=""/>
         </output:serialization-parameters>
     )
 };
@@ -759,10 +756,9 @@ declare
     %test:assertEquals("1--2")
 function ser:item-separator-with-method($method as xs:string) {
     serialize((1, 2),
-        <output:serialization-parameters
-                xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
-            <method>{$method}</method>
-            <item-separator>--</item-separator>
+        <output:serialization-parameters>
+            <output:method>{$method}</output:method>
+            <output:item-separator>--</output:item-separator>
         </output:serialization-parameters>
     )
 };
@@ -771,9 +767,9 @@ declare
     %test:assertEquals("1|2|3|4|5|6|7|8|9|10")
 function ser:serialize-xml-033() {
     let $params :=
-        <output:serialization-parameters xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
-            <method value="xml"/>
-            <item-separator value="|"/>
+        <output:serialization-parameters>
+            <output:method value="xml"/>
+            <output:item-separator value="|"/>
         </output:serialization-parameters>
     return serialize(1 to 10, $params)
 };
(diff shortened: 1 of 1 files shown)


line-o:
If this PR gets merged first, then #4773 can be closed.

dizzzz (approved):

dizzzz (approved):

line-o:
This PR was rebased on current develop HEAD and does not include #4773 anymore

exist-core/src/test/xquery/xquery3/serialize.xql
```diff
@@ -476,508 +448,447 @@ math:pi#0
...
 function ser:serialize-xml-034() {
-    let $params :=
+    serialize(1 to 4,
         <output:serialization-parameters xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
-            <output:method value="xml"/>
-            <output:omit-xml-declaration value="yes"/>
-            <output:item-separator value="=="/>
+            <method value="xml"/>
+            <omit-xml-declaration value="yes"/>
+            <item-separator value="=="/>
```
joewiz, line 794: The spec requires that these elements be in the `output` namespace (even if, evidently, eXist doesn't care). Please fix the namespace on the child elements.

exist-core/src/test/xquery/xquery3/serialize.xql
```diff
@@ -157,99 +193,84 @@ function ser:teardown() {
...
 function ser:serialize-atomic-empty-separator-xml-options() {
-    fn:serialize(("aaa", "bbb"),
-      <output:serialization-parameters
-           xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
-        <output:item-separator value=""/>
-      </output:serialization-parameters>
+    serialize(("aaa", "bbb"),
+        <output:serialization-parameters
+            xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
+            <item-separator value=""/>
```
joewiz, line 265: Same namespace issue here.

exist-core/src/test/xquery/xquery3/serialize.xql
```diff
@@ -157,99 +193,84 @@ function ser:teardown() {
...
-        <output:method value="xml"/>
-        <output:indent value="yes"/>
-      </output:serialization-parameters>
-    return fn:serialize($ser:atomic, $params)
+    serialize(
+        $ser:atomic,
+        <output:serialization-parameters
+               xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
+            <method value="xml"/>
+            <indent value="yes"/>
```
joewiz, line 213: Same namespace issue here.

exist-core/src/test/xquery/xquery3/serialize.xql
```diff
@@ -25,57 +25,44 @@ module namespace ser="http://exist-db.org/xquery/test/serialize";
...
-                    <output:item-separator>{$itemSep}</output:item-separator>
-                else
-                    ()
-            }
-        </output:serialization-parameters>
-    return
-        fn:serialize($data, $options)
+        <item-separator>{$item-separator}</item-separator>
+        <method>adaptive</method>
+        <indent>no</indent>
```
joewiz, line 48: Same namespace issue here.

exist-core/src/test/xquery/xquery3/serialize.xql
```diff
@@ -25,57 +25,44 @@ module namespace ser="http://exist-db.org/xquery/test/serialize";
...
+        "method": "html",
+        "html-version": 5.0
+    }
+;
+
+declare %private variable $ser:opt-xml-adaptive-no-indent :=
+    <output:serialization-parameters
+        xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
+        <method>adaptive</method>
+        <indent>no</indent>
```
joewiz, line 39: Same namespace issue here.

exist-core/src/test/xquery/xquery3/serialize.xql
line-o: My understanding of namespaces in XML is that elements are put into the namespace of the parent element.

joewiz: Elements *inherit their parent's in-scope namespaces*, but *not necessarily their namespace*. See https://www.w3.org/TR/xml-names/#scoping-defaulting.

In this case the `<output:serialization-parameters>` element does create a binding between the `output` namespace prefix and the `http://www.w3.org/2010/xslt-xquery-serialization` namespace URI, but unprefixed elements will remain in the empty namespace. 

To illustrate this, give this a try:

```xquery
xquery version "3.1";

let $root := 
    <output:serialization-parameters xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
        <method>adaptive</method>
return
    $root/output:*
```

The only way the parent could cause its child to be in its namespace would be to add an `xmlns` namespace node without a prefix - i.e., changing the default element namespace within the scope of the parent. 

Doing so here will place the parent's children in the `output` namespace, and the query that selects child elements in the `output` namespace will now return the child.

```xquery
xquery version "3.1";

let $root := 
    <output:serialization-parameters xmlns="http://www.w3.org/2010/xslt-xquery-serialization"
                                     xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization">
        <method>adaptive</method>
]    </output:serialization-parameters>
return
    $root/output:*
```

A useful function for finding out the namespace URI for a node is `fn:namespace-uri()`.

line-o: in 5a7e87fc5954ffc84e439ffa276bda51273f1a29 I addressed your review @joewiz 

- I opted for declaring the output namespace once at the top of the testuite. 
  This way it does not have to be redeclared in each test case. 
- All children of `output:serialization` are now also put back into that namespace
  I think we should open a separate issue on the finding that the namespace of child elements is effectively ignored.

exist-core/src/test/xquery/xquery3/serialize.xql
line-o: see above

exist-core/src/test/xquery/xquery3/serialize.xql
line-o: see above

exist-core/src/test/xquery/xquery3/serialize.xql
line-o: see above

exist-core/src/test/xquery/xquery3/serialize.xql
line-o: see above

line-o:
we have a problem with the docker images it seems (although this is completely unrelated to my changes):
```
DOCKER> Unable to build image [existdb/existdb:7.0.0-SNAPSHOT] : "failed to export image: failed to create image: failed to get layer sha256:89dbfd2aac2fd2efd0e12805e343c3a1139557ea1930ad204bdb7387b89fe3c4: layer does not exist"  ["failed to export image: failed to create image: failed to get layer sha256:89dbfd2aac2fd2efd0e12805e343c3a1139557ea1930ad204bdb7387b89fe3c4: layer does not exist" ]
```

exist-core/src/test/xquery/xquery3/serialize.xql
joewiz: @line-o Perfect! Thank you! And I agree, we should open a new issue about the failure to check namespace of serialization parameters.

joewiz (approved):

adamretter (changes requested):
Please consider that the following change is not equivalent:

Before this PR we had:
```
map { xs:QName("exist:output-doctype"): fn:false() }
```

But now you have:
```
map { "exist:output-doctype": false() })
```

Whilst the end result may be the same, we added the functionality and designed the tests to specifically test for the QName, this should not be removed please.

duncdrum:
@line-o no clue could be a hickup. I would recommend a local build of the image, but that only works on intel so …

joewiz:
I see that not all such QName constructions were removed - e.g., https://github.com/eXist-db/exist/pull/4777/files#diff-41536ff061981a585744759c566780f801e29b78405dd79d158d466f3272905dR374. I agree that it would be best to restore the others to their original form with the xs:QName construction. Perhaps the versions without the xs:QName construction could be preserved too as variations to ensure that both forms continue to work (as they do today).

line-o:
I would be in favour of adding one test that uses constructed QNames to ensure this form will continue to work.
I believe that using QNames as keys is the less common form to provide options to fn:serialize since all other options
do _not_ use them but plain strings.

adamretter:
> I believe that using QNames as keys is the less common form

Not for us. Please keep our QName tests, we worked hard to add this and ensure that it works for several groups of users. If you wish to add additional tests for strings as well, I have no objection, but please do not remove our tests!

line-o:
@adamretter all exist-specific serialization option keys are now tested as `xs:string` and `xs:QName`. Please re-review if that addresses your concerns.

line-o:
@adamretter out of curiosity: is it possible to use QNames for the other keys in the map as well?

line-o:
@dizzzz could you remove the build cache for this PR so that we can re-run the tests?

line-o:
@adamretter Your concerns were addressed 2 months ago, would you please re-review this PR

line-o:
I will bring this PR up on the community call today.

reinhapa (approved):

PR #4778, opened 2023-03-01
[refactor] XQSuite for readability

line-o:
Extract function test:safe-effective-boolean-value.

### DIFF
--- a/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
+++ b/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
@@ -905,31 +905,26 @@ declare %private function test:assertTrue($result as item()*) as element(report)
         </report>
 };
 
+declare %private function test:safe-effective-boolean-value ($result as item()*) as xs:boolean {
+    try {
+        boolean($result)
+    } catch err:FORG0006 {
+        false()
+    }
+};
+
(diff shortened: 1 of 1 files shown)


exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
```diff
@@ -905,31 +905,26 @@ declare %private function test:assertTrue($result as item()*) as element(report)
         </report>
 };
 
+declare %private function test:safe-effective-boolean-value ($result as item()*) as xs:boolean {
+    try {
+        boolean($result)
```
dizzzz, line 910: does this work? the same way? nice....

dizzzz (approved):

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
line-o: Yes I tested it extensively and also compared it with `not(not($result)`

line-o: https://gist.github.com/line-o/4f06be97455ec9e1049da5785ab372f9

reinhapa (approved):

PR #4799, opened 2023-03-15
[fix] XQSuite assertXPath annotation: Support default element namespace

joewiz:
### Description:

XQSuite's `assertXPath` annotation failed to account for results whose in-scope-namespaces changed the default element namespace (e.g., `<x xmlns="foo"/>`. As a result, users are forced to result to wildcards or Clark notation to select elements (e.g., `%test:assertXPath("$result/self::*:x")` or `%test:assertXPath("$result/self::Q{foo}x")`).

This PR adjusts how XQSuite detects namespaces in elements in functions' return values, and constructs a default element namespace declaration. In the example scenario described above, a test author could create the annotation, `%test:assertXPath("$result/self::x")`.

### Reference:

https://markmail.org/message/oxsuzh5elgtxlx23

### Type of tests:

XQSuite

### DIFF
--- a/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
+++ b/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
@@ -958,7 +958,7 @@ declare %private function test:assertXPath($annotation as element(annotation), $
                 function ($namespaces as map(*), $xml as element()) {
                     map:merge(($namespaces,
                 	    for $prefix in in-scope-prefixes($xml)
-                	    where $prefix != "" and $prefix != "xml"
+                	    where $prefix ne "xml"
                 	    return
                 	        map:entry($prefix, namespace-uri-for-prefix($prefix, $xml))
                     ))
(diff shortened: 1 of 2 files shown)


joewiz:
Regarding documentation, I propose expanding the description of the `%test:assertXPath` annotation in https://exist-db.org/exist/apps/doc/xqsuite#annotations with a sentence or two about how to handle different namespace scenarios. Also, as the exist-open thread indicates, users may be confused about the `$result` variable and what context it supplies. The revision should clarify this.

line-o:
Very nice work @joewiz !

exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
```diff
@@ -958,7 +958,7 @@ declare %private function test:assertXPath($annotation as element(annotation), $
                 function ($namespaces as map(*), $xml as element()) {
                     map:merge(($namespaces,
                 	    for $prefix in in-scope-prefixes($xml)
-                	    where $prefix != "" and $prefix != "xml"
```
adamretter, line 961: Interestingly this existing code probably didn't do what the author intended anyway (due to a common misunderstanding of the `!=` operator). If you want to see what I mean, just run:
```
let $prefix := ""
return
  $prefix != ""
,
let $prefix := "xml"
return
  $prefix != "xml"
```

adamretter (approved):

adamretter:
@joewiz Perhaps this one would be worth back-porting to `develop-6.x.x` as well?

PR #4801, opened 2023-03-16
[6.x.x] Add mail:get-mail-session#2 with authentication

nverwer:
### Description:

This PR adds the XQuery function `mail:get-mail-session#2`, which adds authentication to the existing function.

```
mail:get-mail-session($properties as element()?, $authentication as element()) as xs:long?
```
Opens a JavaMail session with authentication.

**Parameters:**
    $properties? An optional JavaMail session properties in the form <properties><property name="" value=""/></properties>. The JavaMail properties are spelled out in Appendix A of the JavaMail specifications.
    $authentication The username and password for authentication in the form <authentication username="" password=""/>.

**Returns:**
    xs:long? : an xs:long representing the session handle.

### Reference:

Pieter Lamers from _John Benjamins Publishing Company_ asked me to make this. Many email servers require authentication nowadays, so this should be useful to others, too.

### Type of tests:

I tested this with an email server. A unit test would require a mock email-server with TLS.

PR #4802, opened 2023-03-16
Add mail:get-mail-session#2 with authentication

nverwer:
### Description:

This PR adds the XQuery function `mail:get-mail-session#2`, which adds authentication to the existing function.
It is the same as #4801, which is for the 6.x.x branch.

```
mail:get-mail-session($properties as element()?, $authentication as element()) as xs:long?
```
Opens a JavaMail session with authentication.

**Parameters:**
    $properties? An optional JavaMail session properties in the form <properties><property name="" value=""/></properties>. The JavaMail properties are spelled out in Appendix A of the JavaMail specifications.
    $authentication The username and password for authentication in the form <authentication username="" password=""/>.

**Returns:**
    xs:long? : an xs:long representing the session handle.

### Reference:

Pieter Lamers from _John Benjamins Publishing Company_ asked me to make this. Many email servers require authentication nowadays, so this should be useful to others, too.

### Type of tests:

I tested this with an email server. A unit test would require a mock email-server with TLS.


adamretter:
Let's conduct the review in https://github.com/eXist-db/exist/pull/4801

PR #4816, opened 2023-03-21
[enhancement] allow more types in XQSuite

line-o:
Add support for following argument and return types

- xs:boolean
- xs:anyURI
- xs:numeric
- xs:positiveInteger
- xs:negativeInteger
- xs:duration
- xs:dayTimeDuration
- xs:yearMonthDuration

Add tests to assert that annotation values are cast to the respective input type and can be asserted against.

Thank you for your contribution to eXist-db! 

To help the community judge your pull request (PR), please include the following:

- A (short) description of the content of changes.
- A context reference to a [Github Issue](https://github.com/eXist-db/exist/issues), a message in the [eXist-open mailinglist](http://exist-open.markmail.org), or a [specification](https://www.w3.org/TR/xquery-31/).
- Tests. The [XQSuite - Annotation-based Test Framework for XQuery](http://exist-db.org/exist/apps/doc/xqsuite.xml) makes it very easy for you to create tests. These tests can be executed from the [eXide editor](http://exist-db.org/exist/apps/eXide/index.html) via XQuery > Run as Test.

Your PR will be tested using [GitHub Actions](https://github.com/eXist-db/exist/actions) against a number of operating systems and environments. The build status is visible in the PR. 

To detect errors in your PR before submitting it, please run eXist's full test suite on your own system via `mvn -V clean verify`.

------

### DIFF
--- a/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
+++ b/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
@@ -580,82 +580,55 @@ declare %private function test:xdm-value-from-annotation-value($annotation-value
             )
 };
 
-declare %private function test:cast($targs as element(value)*, $farg as element(argument)) {
-    for $targ in $targs
-    let $varg := test:xdm-value-from-annotation-value($targ)
+declare %private function test:cast(
+    $annotation-arg-values as element(value)*,
+    $function-argument as element(argument)
+) {
+    for $annotation-arg-value in $annotation-arg-values
+    let $literal-value := test:xdm-value-from-annotation-value($annotation-arg-value)
     return
-        switch (string($farg/@type))
-            case "xs:string"
-            return
-                if ($varg instance of xs:string)
-                then
-                    $varg
-                else
-                    string($varg)
+        switch (string($function-argument/@type))
+            case "xs:string" return
+                if ($literal-value instance of xs:string)
+                then $literal-value
+                else xs:string($literal-value)
 
-            case "xs:integer"
-            return
-                if ($varg instance of xs:integer)
-                then
-                    $varg
-                else
-                    xs:integer($varg)
+            case "xs:integer" return
+                if ($literal-value instance of xs:integer)
+                then $literal-value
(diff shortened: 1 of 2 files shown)


duncdrum (commented):
Should we also test against the numeric notation for `xs:Boolean` ?

line-o:
We could @duncdrum. I just don't think it adds much. If `"true"` gets cast to boolean then any other value will as well. The important part is to declare the type of the parameter in the function.

duncdrum:
I think adding two more test:args is worth the hassle, if my function declares Boolean it should not trip on `1`

duncdrum:
@adamretter the example in the op is quite clear. As xquery annotations are string only, xqsuite annotations struggle with functions whose return type isn’t xs:string. With this PR they struggle less. 

Your implication that this PR is not something “anybody serious” would do, are off the mark both in form and content.

adamretter:
@duncdrum I'm afraid that you have misunderstood my comments.

Also if it was "quite clear" to me, then I would not have asked a genuine question in the first place! I'm afraid it still remains unclear to me.

line-o:
> Do you mean that someone can annotate an existing function within a main or library module, to be able to make assertions about its inputs and outputs?

That is how XQSuite was advertised to me in a different PR.
This does, however not affect _this_ PR in any way.
I need a test function that can accept a boolean as its input. This is currently not possible.
While fixing this I noticed that a number of XDM types are not yet supported and added those as well.
This in turn led to the finding that not all output types were supported, yet.

line-o:
I found at least one important type that is still missing `xs:QName`. I will add that.

line-o:
I added xs:NCName and xs:QName and also added additional tests for xs:boolean (for you @duncdrum).

adamretter:
> That is how XQSuite was advertised to me in a different PR.

I see. I think also the docs suggest such a thing, although in practice that approach is highly undesirable. We should update the docs with better advice when we can.

dizzzz (approved):

reinhapa (approved):

joewiz:
With 2 approvals, this can now be merged. Thanks to all who reviewed!

PR #4817, opened 2023-03-21
Java/cleanup

line-o:
### Description:

Remove dead code and fix warnings and apply improvements that were presented by IntelliJ

- use new switch-statements
- return early were possible
- less indentation
- improve readability
- consistently use "an XMLDB"


exist-core/src/main/java/org/exist/dom/QName.java
```diff
@@ -137,6 +133,7 @@ public String getStringValue() {
      * @deprecated Use for debugging purpose only,
      * use {@link #getStringValue()} for production
      */
+    @Deprecated
```
adamretter, line 136: I am not sure that `@Deprecated` makes sense here from a Java API perspective, as every Java object can have a `toString` method. The `@deprecated` in the comments is more of a suggestion to the developer than anything else.

exist-core/src/main/java/org/exist/dom/QName.java
```diff
@@ -205,15 +202,14 @@ public int compareTo(final QName other) {
...
-        } else if(!(other instanceof QName)) {
+        }
+        if (!(other instanceof final QName qnOther)) {
             return false;
-        } else {
-            final QName qnOther = (QName)other;
-            return this.namespaceURI.equals(qnOther.namespaceURI)
-                    && this.localPart.equals(qnOther.localPart);
         }
+        return this.namespaceURI.equals(qnOther.namespaceURI)
```
adamretter, line 211: I find this a little tricky to read. Is `qnOther` in-scope here? I know they made some changes to `instance of` in newer Java versions so that you can bind variables, but just reading this code, I would expect `qnOther` to be out-of-scope as it is within the context of the `if` statement and this `return` argument is in a separate scope.

exist-core/src/main/java/org/exist/dom/QName.java
```diff
@@ -326,7 +325,7 @@ public static QName parse(final String namespaceURI, final String qname) throws
         }
     }
 
-    private final static Pattern ptnClarkNotation = Pattern.compile("\\{([^&{}]*)\\}([^&{}:]+)");
+    private final static Pattern ptnClarkNotation = Pattern.compile("\\{([^&{}]*)}([^&{}:]+)");
```
adamretter, line 328: Is this still the same pattern?

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
```diff
@@ -425,7 +361,7 @@ public URI getURI() {
     }
 
     /**
-     * This returns an xmldb uri. This is the most generic sort of uri - the only fields set in the uri are scheme and schemeSpecificPart
+     * This returns a xmldb uri. This is the most generic sort of uri - the only fields set in the uri are scheme and schemeSpecificPart
```
adamretter, line 364: In english, it should still be "an xmldb uri". The "a xmldb uri" does not make sense.

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
```diff
@@ -479,7 +414,7 @@ protected void recomputeURI() throws URISyntaxException {
     protected void safeRecomputeURI() {
         try {
             recomputeURI();
-        } catch (final URISyntaxException e) {
+        } catch (final URISyntaxException ignored) {
```
adamretter, line 417: Why `ignored` when the common syntax is `e`?

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
```diff
@@ -168,78 +167,46 @@ public boolean isPositive() {
     @Override
     public AtomicValue convertTo(final int requiredType) throws XPathException {
         switch (requiredType) {
-            case Type.ATOMIC:
-            case Type.ITEM:
-            case Type.NUMBER:
-            case Type.DOUBLE:
+            case Type.ATOMIC, Type.ITEM, Type.NUMBER, Type.DOUBLE -> {
```
adamretter, line 170: I think this `->` syntax is less readable than the previous c-style syntax. The `->` operator here doesn't really deliver anything and should be reserved for the operations it is intended for

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
```diff
@@ -168,78 +167,46 @@ public boolean isPositive() {
...
                 return this;
-            case Type.FLOAT:
-                //if (Float.compare(value, 0.0f) && (value < Float.MIN_VALUE || value > Float.MAX_VALUE)
-                //	throw new XPathException(getExpression(), "Value is out of range for type xs:float");
-                //return new FloatValue((float) value);
+            }
+            case Type.FLOAT -> {
```
adamretter, line 173: Same comment as above... and throughout this `switch` statement!

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
```diff
@@ -327,15 +294,11 @@ public ComputableValue plus(final ComputableValue other) throws XPathException {
...
-        switch (other.getType()) {
-            case Type.DOUBLE:
-                return new DoubleValue(getExpression(), value * ((DoubleValue) other).value);
-            case Type.DAY_TIME_DURATION:
-            case Type.YEAR_MONTH_DURATION:
-                return other.mult(this);
-            default:
-                return mult((ComputableValue) other.convertTo(getType()));
-        }
+        return switch (other.getType()) {
```
adamretter, line 297: Same comments as the above `switch` statement

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
```diff
@@ -506,21 +460,17 @@ public <T> T toJavaObject(final Class<T> target) throws XPathException {
             return (T) Boolean.valueOf(effectiveBooleanValue());
         }
 
-        throw new XPathException(getExpression(), 
-                "cannot convert value of type "
-                        + Type.getTypeName(getType())
-                        + " to Java object of type "
-                        + target.getName());
+        throw new XPathException(getExpression(), ErrorCodes.FORG0001, "cannot convert value of type "
```
adamretter, line 463: Is `FORG0001` the correct error code here?

adamretter (changes requested):
For the most part, I think these are excellent and welcome changes.
Where there are some minor issues I have commented, and would like to see those addressed, otherwise this is good work :-)

exist-core/src/main/java/org/exist/dom/QName.java
line-o: Then I'd suggest to remove the misleading @deprecated from the DocBlock as well.

exist-core/src/main/java/org/exist/dom/QName.java
line-o: I find this tricky as well, I just gave in as the IDE kept insisting this is the way to go. It can be easily refactored to be way easier to read. Maybe there is a setting to ignore this "optimisation" in IntelliJ.

exist-core/src/main/java/org/exist/dom/QName.java
line-o: Yes, as I understand it, without an opening curly brace, as it is still escaped, the closing one is treated as a literal.

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
line-o: Then IntelliJ's spell checker must be at fault.

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
line-o: Because the exception is ignored here, I do like to make it explicit to the reader. Also IntelliJ suggested this change.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
line-o: I do like the new syntax, here much more.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
line-o: Although this new style needs getting used to it does improve switch-statements:
- case blocks are enclosed in curly braces and can not fall-through by accident
- fall-through is now explicit, the number of "cases" is clearly visible

How about listing each type in a separate line as it was before? That could improve readability further.

exist-core/src/main/java/org/exist/dom/QName.java
line-o: qnOther must be in scope as all tests pass and I strongly suspect that QName equality is covered by several test cases.

exist-core/src/main/java/org/exist/dom/QName.java
adamretter: The `@deprecated` in the JavaDoc is not misleading. The `@deprecated` and `@Deprecated` annotations have different purposes, they don't always have to be used together!

exist-core/src/main/java/org/exist/dom/QName.java
adamretter: As it is difficult for both of us to read. I suggest we reverse this small change. Automated suggestions are not always correct/best

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
adamretter: Okay, so let's change it back

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
adamretter: Okay, I have no strong feeling either way.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
adamretter: It would be interesting to see if that helps.

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
joewiz: This case (determining when an "x" is voiced as a consonant or a vowel, for the purpose of selecting the appropriate article "a" vs. "an") is likely to trip up any spell checker, particularly with domain-specific terms like "XMLDB" that may be spelled in a lower case form. See a similar question about X-ray: https://english.stackexchange.com/questions/8181/does-one-use-a-or-an-before-the-word-x-ray. Lots of other good examples in the comments. Since machines may not get it right, we have community review!

line-o: What about HTTP and HTML?
My feeling says (now backed by @joewiz' explanation)

- an HTML page
- an http-request

correct?

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
line-o: It's the code throughout the class for the same type of error. So I assume that is the case.

exist-core/src/main/java/org/exist/dom/QName.java
line-o: I reworded the comment in JavaDoc. I think it the point is quite clear that it should not be used.
Why is not so clear on the other hand. It jus calls getStringValue internally anyway.

exist-core/src/main/java/org/exist/dom/QName.java
line-o: Please have a look at the current form. This is now quite obvious.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
line-o: Please have another look, then.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
reinhapa: @adamretter I would prefer to use of the new switch as function approach instead of the old style here too..

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
reinhapa: @adamretter I would prefer to use of the new switch as function approach instead of the old style here too as it gets more compact.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
reinhapa: Same as above..

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
reinhapa: @line-o I would at least log the exception (even in debug)

reinhapa (changes requested):

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
adamretter: It depends to a certain extent what follows "HTTP" or "HTML". I often have to read a phrase out loud (in my head) to decide if it is an "a" or an "an", I'm not sure of the exact rules... but for me personally one of the approaches will seem correct after I sound it out a couple of times.

exist-core/src/main/java/org/exist/dom/QName.java
```diff
@@ -49,33 +49,31 @@ public class QName implements Comparable<QName> {
     public static final QName DOCTYPE_QNAME = EMPTY_QNAME;
     public static final QName CDATA_SECTION_QNAME = EMPTY_QNAME;
 
+    private static final Pattern ptnClarkNotation = Pattern.compile("\\{([^&{}]*)}([^&{}:]+)");
+    private static final Pattern ptnEqNameNotation = Pattern.compile("Q" + ptnClarkNotation);
```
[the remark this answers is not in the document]
adamretter: In Java, by convention, member variables that are `static` should be in upper-case.

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
adamretter: To answer your question:

> an HTML page
> an http-request

I would write both of those as "a", and not "an". Context is important.

exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
line-o: I like it, too.

exist-core/src/main/java/org/exist/dom/QName.java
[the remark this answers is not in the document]
line-o: I want to point out that I did not touch the field names

reinhapa (approved):

dizzzz (approved):

exist-core/src/main/java/org/exist/xmldb/XmldbURI.java
joewiz: I remember once discussing this very "HTML" issue with Adam some years back! We were correcting each other's articles (i.e., "a HTML" vs. "an HTML"). And finally (I think?) we had a call and realized why. He pronounces the letter "h" as "haych", whereas I pronounce it "aych." This explains why he suggests "a HTML page" and I would write "an HTML page." So we're getting into regional differences. I'm pretty sure an en-US spell checked would say "an HTML page," but I'm not sure about an en-UK or en-AU, etc. All to say, English is a crazy language!

PR #4836, opened 2023-03-27
Cleanup/xquerycontext

line-o:
### Description:

Refactor and cleanup XQueryContext and adjacent classes for readability.

### Reference:

- add LOCAL_NAME field to pragma implementations
- use enhanced switch statement where possible
- reduce indentation by returning early
- reformat code (using IntelliJ defaults)
- convert to final variables where possible 

### Type of tests:

no new tests added

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -2946,34 +2946,26 @@ public class XQueryContext implements BinaryValueManager, Context {
         } catch (final QName.IllegalQNameException e) {
             throw new XPathException(rootExpression, ErrorCodes.XPST0081, "No namespace defined for prefix " + name);
         }
-
-        if (qname.getNamespaceURI().isEmpty()) {
-            throw new XPathException(rootExpression, "XPST0081: pragma's ('" + name + "') namespace URI is empty");
-        } else if (Namespaces.EXIST_NS.equals(qname.getNamespaceURI())) {
-            contents = StringValue.trimWhitespace(contents);
-
-            if (TimerPragma.TIMER_PRAGMA.equals(qname)) {
-                return new TimerPragma(rootExpression, qname, contents);
-            }
-
-            if (Optimize.OPTIMIZE_PRAGMA.equals(qname)) {
-                return new Optimize(rootExpression, this, qname, contents, true);
-            }
-
-            if (ForceIndexUse.EXCEPTION_IF_INDEX_NOT_USED_PRAGMA.equals(qname)) {
-                return new ForceIndexUse(rootExpression, qname, contents);
-            }
-
-            if (ProfilePragma.PROFILING_PRAGMA.equals(qname)) {
-                return new ProfilePragma(rootExpression, qname, contents);
-            }
-
-            if (NoIndexPragma.NO_INDEX_PRAGMA.equals(qname)) {
-                return new NoIndexPragma(rootExpression, qname, contents);
-            }
+        final String ns = qname.getNamespaceURI();
+        if (ns.isEmpty()) {
+            throw new XPathException(rootExpression, ErrorCodes.XPST0081, "XPST0081: pragma's ('" + name + "') namespace URI is empty");
         }
 
-        return null;
+        if (!Namespaces.EXIST_NS.equals(ns)) {
+            // TODO(JL): should we throw or a least warn here?
+            return null;
+        }
+
+        final String sanitizedContents = StringValue.trimWhitespace(contents);
+
+        return switch(qname.getLocalPart()) {
+            case Optimize.LOCAL_NAME -> new Optimize(rootExpression, this, qname, sanitizedContents, true);
+            case TimerPragma.LOCAL_NAME -> new TimerPragma(rootExpression, qname, sanitizedContents);
+            case ProfilePragma.LOCAL_NAME -> new ProfilePragma(rootExpression, qname, sanitizedContents);
+            case ForceIndexUse.LOCAL_NAME -> new ForceIndexUse(rootExpression, qname, sanitizedContents);
+            case NoIndexPragma.LOCAL_NAME -> new NoIndexPragma(rootExpression, qname, sanitizedContents);
+            default -> null;
+        };
     }
 
     @Override
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -198,10 +198,10 @@ public class XQueryContext implements BinaryValueManager, Context {
     private Deque<UserDefinedFunction> closures = new ArrayDeque<>();
 
     // List of options declared for this query at compile time - i.e. declare option
-    private List<Option> staticOptions = null;
+    private List<Option> staticOptions = new ArrayList<>();
 
     // List of options declared for this query at run time - i.e. util:declare-option()
-    private List<Option> dynamicOptions = null;
+    private List<Option> dynamicOptions = new ArrayList<>();
 
     //The Calendar for this context : may be changed by some options
     private XMLGregorianCalendar calendar = null;
@@ -713,11 +713,11 @@ public class XQueryContext implements BinaryValueManager, Context {
         ctx.staticNamespaces = new HashMap<>(this.staticNamespaces);
         ctx.staticPrefixes = new HashMap<>(this.staticPrefixes);
 
-        if (this.dynamicOptions != null) {
+        if (!this.dynamicOptions.isEmpty()) {
             ctx.dynamicOptions = new ArrayList<>(this.dynamicOptions);
         }
 
-        if (this.staticOptions != null) {
+        if (!this.staticOptions.isEmpty()) {
             ctx.staticOptions = new ArrayList<>(this.staticOptions);
         }
 
@@ -1490,7 +1490,7 @@ public class XQueryContext implements BinaryValueManager, Context {
             globalVariables.clear();
         }
 
-        if (dynamicOptions != null) {
+        if (!dynamicOptions.isEmpty()) {
             dynamicOptions.clear(); //clear any dynamic options
         }
 
@@ -2960,17 +2960,11 @@ public class XQueryContext implements BinaryValueManager, Context {
 
     @Override
     public void addOption(final String name, final String value) throws XPathException {
-        if (staticOptions == null) {
-            staticOptions = new ArrayList<>();
-        }
         addOption(staticOptions, name, value);
     }
 
     @Override
     public void addDynamicOption(final String name, final String value) throws XPathException {
-        if (dynamicOptions == null) {
-            dynamicOptions = new ArrayList<>();
-        }
         addOption(dynamicOptions, name, value);
     }
 
@@ -3030,19 +3024,14 @@ public class XQueryContext implements BinaryValueManager, Context {
 
     @Override
     public Option getOption(final QName qname) {
-        if (dynamicOptions != null) {
-            for (final Option option : dynamicOptions) {
-                if (qname.compareTo(option.getQName()) == 0) {
-                    return option;
-                }
+        for (final Option option : dynamicOptions) {
+            if (qname.compareTo(option.getQName()) == 0) {
+                return option;
             }
         }
-
-        if (staticOptions != null) {
-            for (final Option option : staticOptions) {
-                if (qname.compareTo(option.getQName()) == 0) {
-                    return option;
-                }
+        for (final Option option : staticOptions) {
+            if (qname.compareTo(option.getQName()) == 0) {
+                return option;
             }
         }
 
@@ -3247,22 +3236,18 @@ public class XQueryContext implements BinaryValueManager, Context {
     @Override
     public void checkOptions(final Properties properties) throws XPathException {
         checkLegacyOptions(properties);
-        if (dynamicOptions != null) {
-            for (final Option option : dynamicOptions) {
-                if (Namespaces.XSLT_XQUERY_SERIALIZATION_NS.equals(option.getQName().getNamespaceURI())) {
-                    SerializerUtils.setProperty(option.getQName().getLocalPart(), option.getContents(), properties,
-                            inScopeNamespaces::get);
-                }
+        for (final Option option : dynamicOptions) {
+            if (Namespaces.XSLT_XQUERY_SERIALIZATION_NS.equals(option.getQName().getNamespaceURI())) {
+                SerializerUtils.setProperty(option.getQName().getLocalPart(), option.getContents(), properties,
+                        inScopeNamespaces::get);
             }
         }
 
-        if (staticOptions != null) {
-            for (final Option option : staticOptions) {
-                if (Namespaces.XSLT_XQUERY_SERIALIZATION_NS.equals(option.getQName().getNamespaceURI())
-                        && !properties.containsKey(option.getQName().getLocalPart())) {
-                    SerializerUtils.setProperty(option.getQName().getLocalPart(), option.getContents(), properties,
-                            inScopeNamespaces::get);
-                }
+        for (final Option option : staticOptions) {
+            if (Namespaces.XSLT_XQUERY_SERIALIZATION_NS.equals(option.getQName().getNamespaceURI())
+                    && !properties.containsKey(option.getQName().getLocalPart())) {
+                SerializerUtils.setProperty(option.getQName().getLocalPart(), option.getContents(), properties,
+                        inScopeNamespaces::get);
             }
         }
     }
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -787,74 +787,70 @@ public class XQueryContext implements BinaryValueManager, Context {
     }
 
     @Override
-    public void declareNamespace(String prefix, String uri) throws XPathException {
-        if (prefix == null) {
-            prefix = "";
-        }
-
-        if (uri == null) {
-            uri = "";
-        }
-
+    public void declareNamespace(final String prefix, final String uri) throws XPathException {
         if (XML_NS_PREFIX.equals(prefix) || XMLNS_ATTRIBUTE.equals(prefix)) {
-            throw new XPathException(rootExpression, ErrorCodes.XQST0070, "Namespace predefined prefix '" + prefix + "' can not be bound");
+            throw new XPathException(rootExpression, ErrorCodes.XQST0070,
+                    "Namespace predefined prefix '" + prefix + "' can not be bound");
         }
 
-        if (uri.equals(XML_NS)) {
-            throw new XPathException(rootExpression, ErrorCodes.XQST0070, "Namespace URI '" + uri + "' must be bound to the 'xml' prefix");
+        if (XML_NS.equals(uri)) {
+            throw new XPathException(rootExpression, ErrorCodes.XQST0070,
+                    "Namespace URI '" + uri + "' must be bound to the 'xml' prefix");
         }
 
-        final String prevURI = staticNamespaces.get(prefix);
+        final String nonNullPrefix = prefix == null ?  "" : prefix;
+        final String nonNullUri = uri == null ? "" : uri;
 
         //This prefix was not bound
-        if (prevURI == null) {
-
-            if (uri.isEmpty()) {
+        if (!staticNamespaces.containsKey(nonNullPrefix)) {
+            if (nonNullUri.isEmpty()) {
                 //Nothing to bind
-
                 //TODO : check the specs : unbinding an NS which is not already bound may be disallowed.
+                // FIXME (JL): despite the log message nothing happens in this execution branch
                 LOG.warn("Unbinding unbound prefix '{}'", prefix);
             } else {
                 //Bind it
-                staticNamespaces.put(prefix, uri);
-                staticPrefixes.put(uri, prefix);
+                staticNamespaces.put(nonNullPrefix, nonNullUri);
+                staticPrefixes.put(nonNullUri, nonNullPrefix);
             }
-
-        } else {
-            //This prefix was bound
-
-            //Unbind it
-            if (uri.isEmpty()) {
-                // if an empty namespace is specified,
-                // remove any existing mapping for this namespace
-                //TODO : improve, since XML_NS can't be unbound
-                staticPrefixes.remove(uri);
-                staticNamespaces.remove(prefix);
-                return;
-            }
-
-            //those prefixes can be rebound to different URIs
-            if (("xs".equals(prefix) && Namespaces.SCHEMA_NS.equals(prevURI))
-                    || ("xsi".equals(prefix) && Namespaces.SCHEMA_INSTANCE_NS.equals(prevURI))
-                    || ("xdt".equals(prefix) && Namespaces.XPATH_DATATYPES_NS.equals(prevURI))
-                    || ("fn".equals(prefix) && Namespaces.XPATH_FUNCTIONS_NS.equals(prevURI))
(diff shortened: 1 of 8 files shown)


exist-core/src/main/java/org/exist/xquery/pragmas/Optimize.java
```diff
@@ -39,15 +39,17 @@
 import java.util.Iterator;
 import java.util.List;
 
-public class Optimize extends Pragma {
+import static java.lang.System.arraycopy;
```
dizzzz, line 42: is this needed?

line-o: arraycopy is used once

joewiz:
@dizzzz Could you confirm that you intended to merge this PR? I notice that the PR didn't have any explicit approvals. Thanks!

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -2961,17 +2862,11 @@ public boolean optimizationsEnabled() {
 
     @Override
     public void addOption(final String name, final String value) throws XPathException {
-        if (staticOptions == null) {
-            staticOptions = new ArrayList<>();
-        }
         addOption(staticOptions, name, value);
```
adamretter, line 2865: The issue with changing from lazy evaluation again.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -2961,17 +2862,11 @@ public boolean optimizationsEnabled() {
...
     }
 
     @Override
     public void addDynamicOption(final String name, final String value) throws XPathException {
-        if (dynamicOptions == null) {
```
adamretter, line 2972: The issue with changing from lazy evaluation again.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -2985,12 +2880,10 @@ private void addOption(final List<Option> options, final String name, final Stri
...
         final Option option = new Option(rootExpression, qn, value);
 
-        //if the option exists, remove it so we can add the new option
-        for (int i = 0; i < options.size(); i++) {
-            if (options.get(i).equals(option)) {
-                options.remove(i);
-                break;
-            }
+        // if the option exists, remove it so we can add the new option
+        final int pos = options.indexOf(option);
```
adamretter, line 2884: You don't need the double call into the List here. You can just use
```
options.remove(option);
```

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3031,19 +2924,14 @@ else if (Option.OPTIMIZE_IMPLICIT_TIMEZONE.compareTo(qn) == 0) {
 
     @Override
     public Option getOption(final QName qname) {
-        if (dynamicOptions != null) {
-            for (final Option option : dynamicOptions) {
-                if (qname.compareTo(option.getQName()) == 0) {
-                    return option;
-                }
+        for (final Option option : dynamicOptions) {
```
adamretter, line 2927: The issue with changing from lazy evaluation again.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3031,19 +2924,14 @@ else if (Option.OPTIMIZE_IMPLICIT_TIMEZONE.compareTo(qn) == 0) {
...
+                return option;
             }
         }
-
-        if (staticOptions != null) {
-            for (final Option option : staticOptions) {
-                if (qname.compareTo(option.getQName()) == 0) {
-                    return option;
-                }
+        for (final Option option : staticOptions) {
```
adamretter, line 2932: The issue with changing from lazy evaluation again.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3058,34 +2946,26 @@ public Pragma getPragma(final String name, String contents) throws XPathExceptio
...
+        if (!Namespaces.EXIST_NS.equals(ns)) {
+            // TODO(JL): should we throw or a least warn here?
+            return null;
         }
 
-        return null;
+        final String sanitizedContents = StringValue.trimWhitespace(contents);
+
+        return switch(qname.getLocalPart()) {
+            case Optimize.LOCAL_NAME -> new Optimize(rootExpression, this, qname, sanitizedContents, true);
```
adamretter, line 2962: These variables should not all be renamed to `LOCAL_NAME` otherwise they cannot be statically imported into the same class for use as they will conflict. Please rename them back.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3248,22 +3115,18 @@ protected void clearUpdateListeners() {
     @Override
     public void checkOptions(final Properties properties) throws XPathException {
         checkLegacyOptions(properties);
-        if (dynamicOptions != null) {
-            for (final Option option : dynamicOptions) {
-                if (Namespaces.XSLT_XQUERY_SERIALIZATION_NS.equals(option.getQName().getNamespaceURI())) {
-                    SerializerUtils.setProperty(option.getQName().getLocalPart(), option.getContents(), properties,
-                            inScopeNamespaces::get);
-                }
+        for (final Option option : dynamicOptions) {
```
adamretter, line 3118: The issue with changing from lazy evaluation again.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3248,22 +3115,18 @@ protected void clearUpdateListeners() {
...
         }
 
-        if (staticOptions != null) {
-            for (final Option option : staticOptions) {
-                if (Namespaces.XSLT_XQUERY_SERIALIZATION_NS.equals(option.getQName().getNamespaceURI())
-                        && !properties.containsKey(option.getQName().getLocalPart())) {
-                    SerializerUtils.setProperty(option.getQName().getLocalPart(), option.getContents(), properties,
-                            inScopeNamespaces::get);
-                }
+        for (final Option option : staticOptions) {
```
adamretter, line 3125: The issue with changing from lazy evaluation again.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3322,39 +3185,36 @@ public boolean requireDebugMode() {
...
-                        }
-                        destroyable.add(bv);
-                    }
-                } catch (final IOException e) {
-                    LOG.warn("Unable to close binary reference on exiting enclosed expression: {}", e.getMessage(), e);
+        if (binaryValueInstances == null) {
+            return;
+        }
+        final Iterator<BinaryValue> it = binaryValueInstances.iterator();
+        final List<BinaryValue> destroyable = new ArrayList<>();
```
adamretter, line 3202: Another issue with a change to eager evaluation from lazy evaluation. If there are no binary value instances which is likely true for many many XQuery's then this function will use more memory and be less performant. As this is called for every enclosed expression, not wasting memory and performance here is critical.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3322,39 +3185,36 @@ public boolean requireDebugMode() {
...
+        }
 
-            // eagerly cleanup those BinaryValues that are not used outside the EnclosedExpr (to release memory)
-            if (destroyable != null) {
-                for (final BinaryValue bvd : destroyable) {
-                    binaryValueInstances.remove(bvd);
-                }
-            }
+        // eagerly cleanup those BinaryValues that are not used outside the EnclosedExpr (to release memory)
+        for (final BinaryValue bvd : destroyable) {
```
adamretter, line 3216: Same problem as the eager evaluation above.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3322,39 +3185,36 @@ public boolean requireDebugMode() {
...
+            binaryValueInstances.remove(bvd);
```
adamretter, line 3217: Could be refactored to `binaryValueInstances.removeAll(destroyable)`

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3379,27 +3239,23 @@ public void registerBinaryValueInstance(final BinaryValue binaryValue) {
...
-                                removable = new ArrayList<>();
-                            }
-                            removable.add(bv);
-                        }
-                    } catch (final IOException e) {
-                        LOG.error("Unable to close binary value: {}", e.getMessage(), e);
+            if (context.binaryValueInstances == null) {
+                return;
+            }
+            final List<BinaryValue> removable = new ArrayList<>();
```
adamretter, line 3245: Another issue with a change to eager evaluation from lazy evaluation. If there are no binary value instances which is likely true for many many XQuery's then this function will use more memory and be less performant.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -3379,27 +3239,23 @@ public void registerBinaryValueInstance(final BinaryValue binaryValue) {
...
                 }
+            }
 
-                if (removable != null) {
-                    for (final BinaryValue bv : removable) {
-                        context.binaryValueInstances.remove(bv);
-                    }
-                }
+            for (final BinaryValue bv : removable) {
+                context.binaryValueInstances.remove(bv);
```
adamretter, line 3258: Same problem as the eager evaluation above.

exist-core/src/main/java/org/exist/xquery/pragmas/ForceIndexUse.java
```diff
@@ -27,13 +27,12 @@
 import org.exist.xquery.value.Sequence;
 
 public class ForceIndexUse extends Pragma {
-	
+    public static final String LOCAL_NAME = "force-index-use";
```
adamretter, line 30: See my comment above about why renaming this to `LOCAL_NAME` is problematic and needs to be reversed.

exist-core/src/main/java/org/exist/xquery/pragmas/NoIndexPragma.java
```diff
@@ -30,11 +30,11 @@
 import org.exist.xquery.value.Sequence;
 
 public class NoIndexPragma extends Pragma {
+    public final static String LOCAL_NAME = "no-index";
```
adamretter, line 33: See my comment above about why renaming this to `LOCAL_NAME` is problematic and needs to be reversed.

exist-core/src/main/java/org/exist/xquery/pragmas/Optimize.java
adamretter: Typically the Java convention is not to statically import this, the convention is to instead write `System.arrayCopy(...)` at the call-site.

exist-core/src/main/java/org/exist/xquery/pragmas/Optimize.java
```diff
@@ -39,15 +39,17 @@
...
 
-    public  final static QName OPTIMIZE_PRAGMA = new QName("optimize", Namespaces.EXIST_NS, "exist");
+public class Optimize extends Pragma {
+    public final static String LOCAL_NAME = "optimize";
```
adamretter, line 45: See my comment above about why renaming this to `LOCAL_NAME` is problematic and needs to be reversed.

exist-core/src/main/java/org/exist/xquery/pragmas/Optimize.java
```diff
@@ -261,22 +275,26 @@ public void after(XQueryContext context) throws XPathException {
...
-        } else {
-            Optimizable o[] = new Optimizable[optimizables.length + 1];
-            System.arraycopy(optimizables, 0, o, 0, optimizables.length);
-            o[optimizables.length] = optimizable;
-            optimizables = o;
+            return;
         }
+
+        Optimizable[] o = new Optimizable[optimizables.length + 1];
+        arraycopy(optimizables, 0, o, 0, optimizables.length);
```
adamretter, line 295: By convention, this should be `System.arrayCopy` please.

adamretter: However, this could be further improved by refactoring to:

```java
optimizables  = Arrays.copyOf(optimizables, optimizables.length + 1);
optimizables[optimizables.length - 1] = optimizable;
```

exist-core/src/main/java/org/exist/xquery/pragmas/ProfilePragma.java
```diff
@@ -27,8 +27,8 @@
 import org.exist.xquery.value.Sequence;
 
 public class ProfilePragma extends Pragma {
-
-    public final static QName PROFILING_PRAGMA = new QName("profiling", Namespaces.EXIST_NS, "exist");
+    public final static String LOCAL_NAME = "profiling";
```
adamretter, line 30: See my comment above about why renaming this to LOCAL_NAME is problematic and needs to be reversed.

exist-core/src/main/java/org/exist/xquery/pragmas/TimerPragma.java
```diff
@@ -30,8 +30,8 @@
 import org.exist.xquery.value.Sequence;
 
 public class TimerPragma extends Pragma {
-
-    public  final static QName TIMER_PRAGMA = new QName("timer", Namespaces.EXIST_NS, "exist");
+    public final static String LOCAL_NAME = "timer";
```
adamretter, line 33: See my comment above about why renaming this to LOCAL_NAME is problematic and needs to be reversed.

exist-core/src/test/java/org/exist/xquery/XQueryContextTest.java
```diff
@@ -27,18 +27,18 @@
...
 
 import static org.easymock.EasyMock.*;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 
-/**
- * @author <a href="mailto:adam.retter@googlemail.com">Adam Retter</a>
- */
 public class XQueryContextTest {
+    private static final List<String> INITIAL_NAMESPACES = Arrays.asList(
```
adamretter, line 40: 1. Rather than encoding strings here, can these be references to constants that are already defined elsewhere in eXist-db perhaps?
2. As it appears that these are only ever used in a `HashSet`, why not just declare a `HashSet` here instead of a `List` whose member references are later copied to a `HashSet` at each site of use?
3. As this is for expected values for an assertion, this list should ideally be made immutable, i.e. `Set.of(...)`
3.

adamretter (commented):
There are many changes in here which are quite sensible and welcome.
Unfortunately I think there are four problematic areas with this PR that need to be addressed:

1. Some changes in here that lead to increased memory use and performance degradation for the majority of XQuery use-cases. This is caused by changes in the code that now cause every query to evaluate additional code and construct objects that it likely won't ever use.

2. I believe at least one bug has been introduced which will cause an `ArrayIndexOutOfBoundsException` at run-time.

3. Some renaming of constants with conflicting namespaces means that they can no longer be referenced together via static imports. i.e. the ability to compose and reuse of these constants simply has been removed.

4. Changed from non-reifiable to reifiable types in the Module API.

dizzzz:
Hi, if you have serious concerns, then revering is indeed an option IMO. We (Patrick, me) had discussions on this PR though, and considered performance implications... @reinhapa

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
line-o: I did not rename any variable. They are all added.

line-o:
I think we can agree that this was hasitily merged. Another great reminder that we should always give all interested parties time to review new PRs before they are merged.

In this instance as there seems to be a few commits that introduce problematic changes I would propose to revert those instead of all.

I can create a PR that does this.

1. lazy initialisation of ArrayLists for two class fields 
2. lazy initialisation ArrayLists in binary removal functions

For the non-reifiable types vs. reifiable types ... I would like to learn about the implications and why this change should not - never ? - be done.

I also do not understand the issue with adding static field named `LOCAL_NAME` to the Pragmas.

line-o:
If reverting means a force-push to develop @adamretter this should _not_ be done.

line-o:
Regarding Non-reifiable types I found this: https://www.baeldung.com/java-generics-vs-extends-object

> [Reifiable](https://www.baeldung.com/java-super-type-tokens#1reification) types are those whose type is not erased at compile time.

> Non-reifiable types cannot be used in [certain situations](https://docs.oracle.com/javase/tutorial/java/generics/nonReifiableVarargsType.html) such as in an [instanceof](https://www.baeldung.com/java-instanceof) operator or as elements of an array.

So @adamretter you argue that we _want_ to have the types be removed at compile time, correct?

exist-core/src/main/java/org/exist/xquery/pragmas/TimerPragma.java
line-o: not renamed but added

exist-core/src/main/java/org/exist/xquery/pragmas/ProfilePragma.java
line-o: not renamed but added

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -198,10 +198,10 @@ public class XQueryContext implements BinaryValueManager, Context {
     private Deque<UserDefinedFunction> closures = new ArrayDeque<>();
 
     // List of options declared for this query at compile time - i.e. declare option
-    private List<Option> staticOptions = null;
+    private List<Option> staticOptions = new ArrayList<>();
 
     // List of options declared for this query at run time - i.e. util:declare-option()
-    private List<Option> dynamicOptions = null;
+    private List<Option> dynamicOptions = new ArrayList<>();
```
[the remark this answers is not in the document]
line-o: Is this a measurable difference in having it initialised with null vs. an empty ArrayList?

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -800,74 +787,70 @@ public int getExpressionCount() {
...
-        if (uri.equals(XML_NS)) {
-            throw new XPathException(rootExpression, ErrorCodes.XQST0070, "Namespace URI '" + uri + "' must be bound to the 'xml' prefix");
+        if (XML_NS.equals(uri)) {
+            throw new XPathException(rootExpression, ErrorCodes.XQST0070,
+                    "Namespace URI '" + uri + "' must be bound to the 'xml' prefix");
         }
 
-        final String prevURI = staticNamespaces.get(prefix);
+        final String nonNullPrefix = prefix == null ?  "" : prefix;
+        final String nonNullUri = uri == null ? "" : uri;
```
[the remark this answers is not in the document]
line-o: If this turns out to be a problematic change I can change it back to mutating the function arguments.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -1776,20 +1682,22 @@ Module initBuiltInModule(final String namespaceURI, final String moduleClass) {
...
             if (modules == null) {
-                modules = new Module[]{ module };
-            } else {
-                // check if the module is already present
-                for (final Module existingModule : modules) {
-                    if (existingModule == module) {  // NOTE: intentional object identity comparison
-                        return modules;  // already present, no further action needed
-                    }
-                }
+                return new Module[]{ module };
```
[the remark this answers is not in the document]
line-o: Unless these tools do not complain (and I haven't seen any complaints yet) I would like to leave like that.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -1776,20 +1682,22 @@ Module initBuiltInModule(final String namespaceURI, final String moduleClass) {
...
+                    return modules;  // already present, no further action needed
+                }
             }
-            return modules;
+
+            // add the module to the modules
+            final int currentLength = modules.length;
+            final Module[] newModules = Arrays.copyOf(modules, currentLength + 1);
+            // previous length is new last index
+            newModules[currentLength] = module;
```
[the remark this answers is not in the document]
line-o: This was `newModules[newModules.length - 1]` (disguised because modules was mutated / replaced in the line above).

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -2299,10 +2204,7 @@ public void setBaseURI(final AnyURIValue uri, final boolean setInProlog) {
             return;
         }
 
-        if (uri == null) {
-            baseURI = AnyURIValue.EMPTY_URI;
-        }
-        baseURI = uri;
+        baseURI = (uri == null) ? AnyURIValue.EMPTY_URI : uri;
```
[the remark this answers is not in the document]
line-o: Interesting, I agree. 
Which tests cover XQueryContext.declareVariable at the moment?

adamretter:
> Hi, if you have serious concerns, then reverting is indeed an option IMO. We (Patrick, me) had discussions on this PR though, and considered performance implications... @reinhapa

Thanks @dizzzz and @reinhapa for confirming to me that you were happy for me to roll-back this merge. I have now done so.

adamretter:
> For the non-reifiable types vs. reifiable types ... I would like to learn about the implications and why this change should not - never ? - be done.

In this instance it believe it will break compatibility will all existing Java modules that are outside of the eXist-db code-base and already released as compiled Jar files... think EXPath Modules and third-party modules.

> I also do not understand the issue with adding static field named LOCAL_NAME to the Pragmas.

You cannot do this now:
```xquery
import static x.y.A.LOCAL_NAME;
import static x.y.B.LOCAL_NAME;
```

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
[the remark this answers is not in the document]
adamretter: You would have to look

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -1896,49 +1801,49 @@ public Variable declareVariable(final QName qn, final Object value) throws XPath
...
-            return var;
+            return modules[0].declareVariable(qn, value);
         }
 
         final Sequence val = XPathUtil.javaObjectToXPath(value, this, rootExpression);
-        var = globalVariables.get(qn);
 
-        if (var == null) {
+        final Variable var;
+        if (globalVariables.containsKey(qn)) {
```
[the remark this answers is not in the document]
adamretter: Yes this is faster

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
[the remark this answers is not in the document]
adamretter: If you run PMD or FindBugs you will see the complaints. Have you run them?

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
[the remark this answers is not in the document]
adamretter: Yes, which is why it was refactored previously

exist-core/src/test/java/org/exist/xquery/XQueryContextTest.java
adamretter: If this list is modified by even a single test, then it should not be global shared state in the test class, otherwise you have potential side-effects between tests.

PR #4838, opened 2023-03-28
Further cleanup XQueryContext

line-o:
### Description:

additional cleanup refactorings 

### Reference:

refs #4836 

### Type of tests:

no new tests were added

### DIFF
--- a/exist-core/src/main/java/org/exist/Namespaces.java
+++ b/exist-core/src/main/java/org/exist/Namespaces.java
@@ -24,6 +24,7 @@ package org.exist;
 import org.exist.dom.QName;
 
 import javax.xml.XMLConstants;
+import java.util.Set;
 
 /**
  * Global namespace declarations.
@@ -92,4 +93,7 @@ public interface Namespaces {
     String XHTML_NS = "http://www.w3.org/1999/xhtml";
 
 	String XINCLUDE_NS = "http://www.w3.org/2001/XInclude";
+
+	Set<String> PROTECTED_NS = Set.of(
+			XML_NS, SCHEMA_NS, SCHEMA_INSTANCE_NS, XPATH_FUNCTIONS_NS);
 }
--- a/exist-core/src/main/java/org/exist/interpreter/Context.java
+++ b/exist-core/src/main/java/org/exist/interpreter/Context.java
@@ -308,8 +308,6 @@ public interface Context {
      */
     boolean lockDocumentsOnLoad();
 
-    void addLockedDocument(DocumentImpl doc);
-
     void setShared(boolean shared);
 
     boolean isShared();
--- a/exist-core/src/main/java/org/exist/util/MapUtil.java
+++ b/exist-core/src/main/java/org/exist/util/MapUtil.java
@@ -54,8 +54,9 @@ public interface MapUtil {
      *
      * @return The HashMap
      */
-    static <K, V> Map<K,V> HashMap(final Tuple2<K, V>... entries) {
-        return HashMap(Math.max(entries.length, 16), entries);
+    @SafeVarargs
+    static <K, V> Map<K,V> hashMap(final Tuple2<K, V>... entries) {
+        return hashMap(Math.max(entries.length, 16), entries);
     }
 
     /**
@@ -69,7 +70,8 @@ public interface MapUtil {
      *
      * @return The HashMap
      */
-    static <K, V> Map<K,V> HashMap(final int initialCapacity, final Tuple2<K, V>... entries) {
+    @SafeVarargs
+    static <K, V> Map<K,V> hashMap(final int initialCapacity, final Tuple2<K, V>... entries) {
         final Map<K, V> map = new HashMap<>(initialCapacity);
         for (final Tuple2<K, V> entry : entries) {
             map.put(entry._1, entry._2);
--- a/exist-core/src/main/java/org/exist/xquery/Optimizer.java
+++ b/exist-core/src/main/java/org/exist/xquery/Optimizer.java
@@ -58,13 +58,12 @@ public class Optimizer extends DefaultExpressionVisitor {
 
     private static final Logger LOG = LogManager.getLogger(Optimizer.class);
 
-    private XQueryContext context;
+    private final XQueryContext context;
+    private final List<QueryRewriter> rewriters;
 
     private int predicates = 0;
 
     private boolean hasOptimized = false;
-
-    private List<QueryRewriter> rewriters;
 
     public Optimizer(XQueryContext context) {
         this.context = context;
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -1198,7 +1198,7 @@ public class XQueryContext implements BinaryValueManager, Context {
     }
 
     /**
-     * Get's a document from the "Available documents" of the
+     * Gets a document from the "Available documents" of the
      * dynamic context.
      *
      * @param uri the URI by which the document was registered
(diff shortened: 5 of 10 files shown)


adamretter:
I think we need to pause this one please until we can address the issues with this one - https://github.com/eXist-db/exist/pull/4836#issuecomment-1486522704

Also, would it be possible to give the PR a different title from the other PR to reduce confusion for people. As a suggestion, perhaps this one could be renamed to "Further cleanup of the XQueryContext class"?

line-o:
@Adam I will start to address the issues you raised in #4836 here. Does that sound reasonable?

adamretter:
> I will start to address the issues you raised in https://github.com/eXist-db/exist/pull/4836 here. Does that sound reasonable?

It might be.... I can't really say, as I am not really clear on the distinction between that PR and this PR

line-o:
- Lazy ArrayList instantiation is now back in bbb8ca7a47b040d8b24b34e28956c6db4a1892b3
- the options.remove(option) is in 892f73e24986ff70181c898dd2f78d286dadadea
- all @Nullable annotations are now in front of the return type 2305da8215380da2e57ead235353a1aeb429c948
- the baseURI assignment is not setting the default anymore 8c1bcd3812a55785b90aaa2953129f9dd61804ee

line-o:
@adamretter as for the __newly added__ static fields named `LOCAL_NAME` to Pragmas, do you still think this is an issue?

line-o:
https://github.com/line-o/exist/blob/a9f49e4fd1df17fdb0b0109fb61f4108b9867eeb/exist-core/src/main/java/org/exist/xquery/pragmas/ForceIndexUse.java#L30
    
as you can see from the linked example above  the `LOCAL_NAME` static field in each Pragma is an addition. 
The field that was below it was changed only in so far that it uses the new field to populate the localPart of its QName value.

adamretter:
> as for the newly added static fields named LOCAL_NAME to Pragmas, do you still think this is an issue?

Yes, it still has the same problems that I mentioned in the previous PR - https://github.com/eXist-db/exist/pull/4836#discussion_r1150860145

line-o:
- Reifiable types now also reverted back to non-reifiable ones in 457807b0b26a2c0fa3d118e82a4e3bdb348c6738
  I can confirm that this breaks compatibility with modules.

line-o:
As for the last remaining discussion item: the newly added LOCAL_NAME properties. As they are not imported on their own anywhere, I have a hard time understanding where the issue is.

line-o:
@adamretter @reinhapa @dizzzz merging this PR will resolve all issues that were introduced in #4836 there is no need to rollback and it will make it in turn very hard to merge this PR.
Also note that the only critical issue is _not_ in develop but was only introduced here (switch from non-reifiable types to reifiable ones).

line-o:
the import for pragmas in XQueryContext class is `import org.exist.xquery.pragmas.*;`

adamretter:
> Yes you could not do that. But it is also not done and until there is a need to do that it should be fine, no?

You could do it, but it doesn't make sense. This is not the convention that is used everywhere else in this and other projects.

> there is no need to rollback

https://github.com/eXist-db/exist/pull/4836 was already rolled back yesterday morning after I received agreement from @dizzzz and @reinhapa

line-o:
@adamretter I still see my commits in develop. So please refrain from force-pushing onto develop.
As for the static fields: What would you propose as good names for those?

line-o:
@dizzzz could you remove the build-cache for this branch so that we can re-run the tests?

adamretter:
> I still see my commits in develop

Presumably you mean your "local clone of" `develop`, they are not in https://github.com/eXist-db/exist/commits/develop.

> So please refrain from force-pushing onto develop.

I agree that force pushing to `develop` is not desirable. However, it is sometimes the only tool for the job, and we only do that when something has gone wrong and has to be undone. Thankfully, this happens rarely in practice.

adamretter:
> As for the static fields: What would you propose as good names for those?

Something descriptive that does not clash when they are imported into the same context.
Although, regardless of their name, I am still not clear what benefit adding those actually brings to the codebase?

line-o:
@adamretter so instead of `ProfilePragma.LOCAL_NAME` you would be happy with `ProfilePragma.PROFILING_PRAGMA_LOCAL_NAME`?

line-o:
@adamretter I hope e421919b7b4e6e75a4c72a1af08122834c1ba892 is both what you were asking for and also makes it clear why they were added

line-o:
I would like to remove the now unused QName fields. Is that OK?

line-o:
As for the QNames in Pragmas, _some_ of them are in fact used. I will leave them all in for that reason.

line-o:
@adamretter All of your concerns are addressed. Please review.

dizzzz (approved):

line-o:
rebased onto current develop

reinhapa (approved):

line-o:
XQTS results indicate that with this PR applied 14 additional test cases pass. Somewhat surprising to me but certainly welcome.

adamretter (changes requested):
I think this is now almost ready to merge. There are just some release engineering tasks to figure out:

1. This PR still seems to contain aspects of the last PR that was rejected. i.e. there are commits that make a change, and then commits that undo that change. Does this need either (a) rebasing on `develop`, or (b) some manual rebasing to cleanup the history to remove the unnecessary churn?

2. I am concerned that more tests are passing after these changes. I am not unhappy about that in principle of course, but as this PR is only meant to be refactoring of existing code, there must be some unanticipated change. I am concerned that these additional passes may be false/positives, or do you know what has changed in this PR that is now causing the additional tests to pass?

line-o:
@adamretter re 1: since the original commits are still in develop its impossible to get rid of those changes otherwise than to revert them.

re 2: Is it possible to find out which test cases were now successful?
If you are concerned about the changes in XQTS test results you are certainly in a much better position to find out why those are _reported_ as passing. If these are false positive as you suspect (which is also more likely), it is not in scope of this PR to investigate the reporting.

adamretter:
> re 1: since the original commits are still in develop its impossible to get rid of those changes otherwise than to revert them.

They are no in develop, as you yourself recognised previously - they were removed via a rebase.

> re 2: Is it possible to find out which test cases were now successful?

Of course! Simply run the test suite before and after you changes and compare the results. This is true of any test suite for any application where you are making changes!

> If you are concerned about the changes in XQTS test results you are certainly in a much better position to find out why those are reported as passing.

Why is that?

>  If these are false positive as you suspect (which is also more likely), it is not in scope of this PR to investigate the reporting.

We cannot say if that is the case, until you furnish us with the details of what (and why) your PR is affecting in terms of the XQTS.

reinhapa:
@adamretter this PR is open for quite a while now and seems reasonable to me. Can this be merged?

adamretter:
@reinhapa There has been no response from the submitter to the comments here - https://github.com/eXist-db/exist/pull/4838#issuecomment-1508684403

Outstanding tasks:
1. Rebase onto `develop`
2. Disclose what has changed in the XQTS output (which has been alluded to)

line-o:
The XQTS test results as reported leave me with questions. I am wondering why 16 new failures are reported https://github.com/eXist-db/exist/actions/runs/5315040636/jobs/9622950623?pr=4838#step:16:29
and all of those tests reported as now failing are unrelated to the changes in this PR (all for `fn:matching` regular expressions).

Furthermore, if I take the first one of those tests reported as failing, `re00475`, and run it against this build of existdb, it passes.

```xml
   <test-case name="re00475">
      <description>Test regex syntax</description>
      <created by="Michael Kay" on="2011-07-04"/>
      <test>(every $s in tokenize('', ',') satisfies matches($s, '^(?:\p{IsMusicalSymbols})$')) and (every $s in tokenize('&#119039;', ',') satisfies not(matches($s, '^(?:\p{IsMusicalSymbols})$')))</test>
      <result>
         <assert-true/>
      </result>
   </test-case>
```

@adamretter can you explain what might be happening here?

adamretter:
@line-o I will take a look as soon as I can. My first suggestion would be to build this version of eXist-db, rebuild a version of XQTS for this eXist-db, and then run the exact same set of tests as GitHub CI to make sure you get the same results

PR #4839, opened 2023-03-28
Cleanup/configuration

line-o:
### Description:

cleanup of Configuration.java

### DIFF
--- a/exist-core/src/main/java/org/exist/util/Configuration.java
+++ b/exist-core/src/main/java/org/exist/util/Configuration.java
@@ -50,15 +50,12 @@ import org.exist.indexing.IndexManager;
 import org.exist.dom.memtree.SAXAdapter;
 import org.exist.scheduler.JobConfig;
 import org.exist.scheduler.JobException;
-import org.exist.storage.BrokerFactory;
 import org.exist.storage.BrokerPool;
 import org.exist.storage.DBBroker;
 import org.exist.storage.DefaultCacheManager;
 import org.exist.storage.IndexSpec;
 import org.exist.storage.NativeBroker;
-import org.exist.storage.NativeValueIndex;
 import org.exist.storage.XQueryPool;
-import org.exist.storage.journal.Journal;
 import org.exist.storage.serializers.CustomMatchListenerFactory;
 import org.exist.storage.serializers.Serializer;
 import org.exist.validation.GrammarPool;
@@ -95,6 +92,43 @@ import org.xmlresolver.Resolver;
 
 import static com.evolvedbinary.j8fu.tuple.Tuple.Tuple;
 import static javax.xml.XMLConstants.FEATURE_SECURE_PROCESSING;
+import static org.exist.Namespaces.XPATH_FUNCTIONS_NS;
+import static org.exist.scheduler.JobConfig.*;
+import static org.exist.storage.BrokerFactory.PROPERTY_DATABASE;
+import static org.exist.storage.BrokerPoolConstants.*;
+import static org.exist.Indexer.CONFIGURATION_INDEX_ELEMENT_NAME;
+import static org.exist.Indexer.PRESERVE_WS_MIXED_CONTENT_ATTRIBUTE;
+import static org.exist.Indexer.PROPERTY_INDEXER_CONFIG;
+import static org.exist.Indexer.PROPERTY_PRESERVE_WS_MIXED_CONTENT;
+import static org.exist.Indexer.PROPERTY_SUPPRESS_WHITESPACE;
+import static org.exist.Indexer.SUPPRESS_WHITESPACE_ATTRIBUTE;
+import static org.exist.collections.CollectionCache.PROPERTY_CACHE_SIZE_BYTES;
+import static org.exist.storage.DBBroker.*;
+import static org.exist.storage.DefaultCacheManager.*;
+import static org.exist.storage.NativeBroker.INDEX_DEPTH_ATTRIBUTE;
+import static org.exist.storage.NativeBroker.PROPERTY_INDEX_DEPTH;
+import static org.exist.storage.NativeValueIndex.INDEX_CASE_SENSITIVE_ATTRIBUTE;
+import static org.exist.storage.NativeValueIndex.PROPERTY_INDEX_CASE_SENSITIVE;
+import static org.exist.storage.XQueryPool.MAX_STACK_SIZE_ATTRIBUTE;
+import static org.exist.storage.XQueryPool.POOL_SIZE_ATTTRIBUTE;
+import static org.exist.storage.XQueryPool.PROPERTY_MAX_STACK_SIZE;
+import static org.exist.storage.XQueryPool.PROPERTY_POOL_SIZE;
+import static org.exist.storage.journal.Journal.PROPERTY_RECOVERY_JOURNAL_DIR;
+import static org.exist.storage.journal.Journal.PROPERTY_RECOVERY_SIZE_LIMIT;
+import static org.exist.storage.journal.Journal.PROPERTY_RECOVERY_SYNC_ON_COMMIT;
+import static org.exist.storage.journal.Journal.RECOVERY_JOURNAL_DIR_ATTRIBUTE;
+import static org.exist.storage.journal.Journal.RECOVERY_SIZE_LIMIT_ATTRIBUTE;
+import static org.exist.storage.journal.Journal.RECOVERY_SYNC_ON_COMMIT_ATTRIBUTE;
+import static org.exist.storage.serializers.Serializer.*;
+import static org.exist.util.HtmlToXmlParser.*;
+import static org.exist.util.ParametersExtractor.PARAMETER_ELEMENT_NAME;
+import static org.exist.util.XMLReaderObjectFactory.PROPERTY_VALIDATION_MODE;
+import static org.exist.util.XMLReaderPool.XmlParser.*;
+import static org.exist.xquery.FunctionFactory.*;
+import static org.exist.xquery.XQueryContext.*;
+import static org.exist.xquery.XQueryWatchDog.PROPERTY_OUTPUT_SIZE_LIMIT;
+import static org.exist.xquery.XQueryWatchDog.PROPERTY_QUERY_TIMEOUT;
+import static org.exist.xslt.TransformerFactoryAllocator.*;
 
 
 public class Configuration implements ErrorHandler {
@@ -219,9 +253,9 @@ public class Configuration implements ErrorHandler {
             }
 
             //db connection settings
-            final NodeList dbcon = doc.getElementsByTagName(BrokerPool.CONFIGURATION_CONNECTION_ELEMENT_NAME);
-            if (dbcon.getLength() > 0) {
-                configureBackend(existHomeDirname, (Element) dbcon.item(0));
+            final NodeList dbConnection = doc.getElementsByTagName(CONFIGURATION_CONNECTION_ELEMENT_NAME);
+            if (dbConnection.getLength() > 0) {
+                configureBackend(existHomeDirname, (Element) dbConnection.item(0));
             }
 
             // lock-table settings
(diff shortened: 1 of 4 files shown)


exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -98,36 +97,30 @@
...
     private static final String XQUERY_CONFIGURATION_ELEMENT_NAME = "xquery";
     private static final String XQUERY_BUILTIN_MODULES_CONFIGURATION_MODULES_ELEMENT_NAME = "builtin-modules";
     private static final String XQUERY_BUILTIN_MODULES_CONFIGURATION_MODULE_ELEMENT_NAME = "module";
-    
-    public final static String BINARY_CACHE_CLASS_PROPERTY = "binary.cache.class";
-    
+    protected Optional<Path> configFilePath = Optional.empty();
+    protected Optional<Path> existHome = Optional.empty();
+    protected DocumentBuilder builder = null;
+    protected HashMap<String, Object> config = new HashMap<>(); //Configuration
```
adamretter, line 109: 1. Should be `protected Map<String, Object> config = new HashMap<>();`
2. Why `protected` and not `private`?
3. Can this be `final`?

exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -98,36 +97,30 @@
...
+    private static final Logger LOG = LogManager.getLogger(Configuration.class); //Logger
     private static final String XQUERY_CONFIGURATION_ELEMENT_NAME = "xquery";
     private static final String XQUERY_BUILTIN_MODULES_CONFIGURATION_MODULES_ELEMENT_NAME = "builtin-modules";
     private static final String XQUERY_BUILTIN_MODULES_CONFIGURATION_MODULE_ELEMENT_NAME = "module";
-    
-    public final static String BINARY_CACHE_CLASS_PROPERTY = "binary.cache.class";
-    
+    protected Optional<Path> configFilePath = Optional.empty();
+    protected Optional<Path> existHome = Optional.empty();
+    protected DocumentBuilder builder = null;
```
adamretter, line 108: 1. Should be marked `@Nullable`
2. Why `protected` and not `private`?
3. Is this actually used anywhere?

exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -325,272 +346,290 @@ private void configureLockManager(final Element lockManager) throws DatabaseConf
...
+        final Element cache = (Element) nlCache.item(0);
+        final String binaryCacheClass = getConfigAttributeValue(cache, "class");
+        config.put(BINARY_CACHE_CLASS_PROPERTY, binaryCacheClass);
+        LOG.debug(BINARY_CACHE_CLASS_PROPERTY + ": {}", config.get(BINARY_CACHE_CLASS_PROPERTY));
     }
-    
-    private void configureXQuery( Element xquery ) throws DatabaseConfigurationException
-    {
+
+    private void configureXQuery(Element xquery) throws DatabaseConfigurationException {
```
adamretter, line 382: `xquery` can be `final`

exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -325,272 +346,290 @@ private void configureLockManager(final Element lockManager) throws DatabaseConf
...
+     * @param modulesClassMap  map containing all classes of modules
+     * @param modulesSourceMap map containing all source uris to external resources
+     * @throws DatabaseConfigurationException if one of the modules is configured incorrectly
      */
-    private void loadModuleClasses( Element xquery, Map<String, Class<?>> modulesClassMap, Map<String, String> modulesSourceMap, Map<String, Map<String, List<? extends Object>>> moduleParameters) throws DatabaseConfigurationException {
+    private void loadModuleClasses(final Element xquery,
+                                   final Map<String, Class<?>> modulesClassMap,
+                                   final Map<String, String> modulesSourceMap,
+                                   final Map<String, Map<String, List<?>>> moduleParameters
+    ) throws DatabaseConfigurationException {
```
adamretter, line 454: Changed from `non-reified` to `reified` types. Is that intentional?

exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -887,558 +929,472 @@ private void configureBackend( final Optional<Path> dbHome, Element con ) throws
...
+                    if ("org.exist.storage.StartupTrigger".equals(iface.getName())) {
+                        isStartupTrigger = true;
+                        break;
                     }
                 }
+
+                // if it actually is a StartupTrigger
+                if (isStartupTrigger) {
+                    // Parse additional parameters
+                    final Map<String, List<?>> params = ParametersExtractor.extract(trigger.getElementsByTagName(ParametersExtractor.PARAMETER_ELEMENT_NAME));
```
adamretter, line 1251: This `params` used to be of type `List<? extends Object>>`, changing this from a non-reified to a reified type, will break compiled compatibility with any existing Startup Triggers that are out there in the wild (e.g. EXPath packages etc); this is because it is used as the Interface in `StartupTrigger#execute`.

Please change this back.

exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -1448,157 +1404,147 @@ private void configureValidation(final Optional<Path> dbHome, final Element vali
...
-    	return( property.toString() );
-    }
+    private String getAttributeSystemPropertyName(Element element, String attributeName) {
+        final StringBuilder property = new StringBuilder(attributeName);
+        Node parent = element.getParentNode();
 
+        property.insert(0, ".");
+        property.insert(0, element.getLocalName());
+
+        while (parent instanceof Element) {
```
adamretter, line 1514: We seem to be missing ` parent != null` check here now?

exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -1448,157 +1404,147 @@ private void configureValidation(final Optional<Path> dbHome, final Element vali
...
-    		parent   = parent.getParentNode();
-    	}
-    	
-    	property.insert( 0, "org." );
-    	
-    	return( property.toString() );
-    }
+    private String getAttributeSystemPropertyName(Element element, String attributeName) {
+        final StringBuilder property = new StringBuilder(attributeName);
+        Node parent = element.getParentNode();
```
adamretter, line 1509: Should be declared `@Nullable`

adamretter (changes requested):
In general it looks good. I think it might be better to just replace it with a JAXB solution, which would mean we could basically delete all of this code.

I do have concerns about multiple `return` statements again, have you checked this with PMD and FindBugs?

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: Same over-eager optimisation as in #4838
I am gonna change that.

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: see above

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: You mean `parent`?

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: yes

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: No idea.

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: This declaration was just moved to a different location. If you think those changes can be made I can do this additional optimisation.

exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -887,558 +929,472 @@ private void configureBackend( final Optional<Path> dbHome, Element con ) throws
...
-            }
-            catch( final NumberFormatException e ) {
-                LOG.warn( e );
+            } catch (final NumberFormatException e) {
+                LOG.warn(e);
             }
         }
     }
-    
-    public static class StartupTriggerConfig {
```
dizzzz, line 1217: Removed..... can't find replacement?

line-o: The class `StartupTriggerConfig` was replaced with/ converted to a record.

line-o: in deae76a7cee3a4654c2cd577db3fd5c6c7573094

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: > "At run time, the result of the instanceof operator is true if the value of the RelationalExpression is not null and the reference could be cast to the ReferenceType without raising a ClassCastException. Otherwise the result is false."

line-o: @dizzzz you are correct

line-o:
@adamretter in #4838 you stated

> This is not the convention that is used everywhere else in this and other projects.

for the naming of static finals. So when I do encounter such clashing names they are considered a bug and should be renamed, correct?

adamretter:
> for the naming of static finals. So when I do encounter such clashing names they are considered a bug and should be renamed, correct?

When refactoring, if you are touching the code, the general approach that we (and many projects) have been following over the years is that of the [Boy Scout Rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html): "Always leave the campground cleaner than you found it"

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: done

exist-core/src/main/java/org/exist/util/Configuration.java
reinhapa: @line-o  If the visibility can be reduced, I would do. Also using the interface instead of the concrete Class would be preferable (when doing a clean up anyway)

exist-core/src/main/java/org/exist/util/Configuration.java
reinhapa: @line-o If the visibility can be reduced without side effects I would do so..

reinhapa (changes requested):

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: I am still unsure what to annotate as nullable here.

reinhapa (approved):

reinhapa (approved):

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: done

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: @adamretter unless you specify what you want this will stay unchanged

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: @dizzzz Please resolve if you do not have further questions

line-o:
It would be good to merge this rather soon now that all issues were addressed. I would want to avoid the pain to rebase all those commits again because of small changes that are added to Configuration.java in develop. Already had to do this twice now.

exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -898,722 +950,629 @@ private void configureBackend( final Optional<Path> dbHome, Element con ) throws
...
+
+            property.insert(0, ".");
+            property.insert(0, parentName);
 
+            parent = parent.getParentNode();
+        }
+
+        property.insert(0, "org.");
+
+        return (property.toString());
```
reinhapa, line 1553: The surrounding parentheses can be removed I guess..

exist-core/src/main/java/org/exist/util/Configuration.java
reinhapa: > Should be declared `@Nullable`

@adamretter what part should be annotated here? As far I see it, there is nothing nullable here or have I missed something?

reinhapa:
> In general it looks good. I think it might be better to just replace it with a JAXB solution, which would mean we could basically delete all of this code.

@adamretter  I see your point. I would like to see this change as one step ahead, but I would try to use JAXB as a separate step.

dizzzz (approved):

exist-core/src/main/java/org/exist/util/Configuration.java
line-o: I believe so, too.

reinhapa: @line-o could you remove the parentheses in order I can merge?

PR #4841, opened 2023-03-30
[ci] split test and verify into separate workflows

line-o:
### Description:

A lot of time consuming tasks such as static code analysis tasks are performed in the test workflow. They are performed per run in the test matrix even if they are independent of the operating system. 

- dependency checks
- vulnerability checking
- code coverage of tests

This PR separates both concerns into separate workflows.

- one for testing and
- one for verification, static code analysis and code coverage 

This is the result of discussions in the community meetings on the 20th and 27th of March 2023

### Reference:

This will also allow us to see if tests in a PR succeed even if we experience issues with the dependency checking plugin.
We should discuss if it is beneficial to check if packaging succeeds on each operating system.
This would mean to call `mvn package` instead of `mvn test`.

### Type of tests:

None added

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -1,4 +1,4 @@
-name: Test
+name: Test (Package)
 on: [push, pull_request]
 jobs:
   test:
@@ -21,7 +21,7 @@ jobs:
           cache: 'maven'
       - name: Maven Test
         timeout-minutes: 60
-        run: mvn -V -B -DtrimStackTrace=false -D"maven.resolver.transport"=wagon clean test
+        run: mvn -V -B -DtrimStackTrace=false -D"maven.resolver.transport"=wagon clean package
       - name: Archive build logs
         if: always()
         uses: actions/upload-artifact@v2
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -21,21 +21,12 @@ jobs:
           cache: 'maven'
       - name: Maven Test
         timeout-minutes: 60
-        run: mvn -V -B -DtrimStackTrace=false -D"maven.resolver.transport"="wagon" clean verify
-      - name: Maven Code Coverage
-        if: ${{ github.ref == 'refs/heads/develop' && matrix.jdk == '8' && matrix.os == 'ubuntu-latest' }}
-        env:
-          CI_NAME: github
-          BRANCH_NAME_OR_REF: ${{ github.head_ref || github.ref }}
-          CI_BUILD_NUMBER: ${{ github.run_id }}
-          CI_BUILD_URL: https://github.com/${{ github.repository }}/commit/${{ github.event.after }}/checks
-          COVERALLS_TOKEN: ${{ secrets.COVERALLS_TOKEN }}
-        run: mvn -V -B jacoco:report coveralls:report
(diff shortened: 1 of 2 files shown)


line-o:
Interestingly a behaviour I noticed on my development machine is reproducible in CI as well:
`mvn clean test` never succeeds because of `PackageTriggerTest` not working in isolation.
Switching to `mvn clean package` will get us past that with the downside of adding execution time _and_ increased artefact size.

line-o:
Side note: the `Maven Code Coverage` job will never run since it checks for the idk version to be `8` and we moved to `17`

line-o:
@duncdrum this PR changes the test workflow to run `mvn package` instead of `mvn verify`. This does _not_ do the dependency checks that fail consistently on second run.
Which also means that license header checking and OWASP DB updating and code coverage reporting will all only run once in the Verify workflow.
These JDK and OS independent checks currently run once per test matrix run which is a waste of VM uptime, cache storage and network bandwidth.
The full effect is visible when build caches are turned on again.

line-o:
Code coverage is _covered_ by SonarCloud not Codacy.

line-o:
If you want to compare  execution times:
- last successful test run with the old setup https://github.com/eXist-db/exist/actions/runs/4539942886
  Screenshot 2023-03-30 at 21 21 34
- first run with the new setup https://github.com/eXist-db/exist/actions/runs/4563948788
  Screenshot 2023-03-30 at 15 49 43

I do agree that the times are not that different and also vary wildly but there is still no need to do these types of checks more than once.

reinhapa:
@line-o when splitting away the dependency check then also the license check can be done there I think. Also I was wondering, why we split the all into separate workflow files?

duncdrum:
@line-o sorry I was comparing three wrong runs. This makes sense, also ruining them in parallel

line-o:
@reinhapa I agree, there is room for further improvements. 
My goal was to get to a setup were we would finally see which PRs are passing tests again. The `Verify` workflow now does it all (testing, license and dependency check) whereas `Test (Package)` only runs tests and packages the application. 

A next step would then be separate workflows for

- licence header check
- dependency check
- code coverage (can be combined with tests to not have to run them an additional time)
- did I forget one?

adamretter:
At the moment as far as I can see in this PR, these changes cause all of the tests to be ran twice. I don't think that was what you were aiming for!

> Should this be removed, as Codacy already includes code coverage

The Maven Code Coverage is fro JaCoCo and should remain, it also integrates with Coveralls.io.

adamretter:
> It was me who removed the redundant build workflow

Thanks :-)

> the decision to separate the verify step was made in the community call

That's not at issue. I just wanted to point out that you may not get the results you are hoping for.

adamretter:
> What I see here is exactly what I was aiming for. Ensure tests (and packaging) works on all supported operating systems and run verify once.

Yes, you are now only calling the goals bound to the `verify` lifecycle once. However, you are calling the goals bound to the `test` lifecycle twice. As the `test` lifecycle is the slowest, you have in-effect doubled the amount of work and time required.

reinhapa:
@adamretter would this https://github.com/reinhapa/exist/tree/githubactions be a better approach?

line-o:
As discussed on the community call we will go with the better approach that @reinhapa came up with.

PR #4842, opened 2023-03-30
Fixes dependabot definition

reinhapa:
Fixed the dependabot definition for github actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,11 +1,13 @@
 version: 2
 updates:
-- package-ecosystem:
-  - github-actions
-  - maven
+- package-ecosystem: "github-actions"
   directory: "/"
   schedule:
-    interval: daily
+    interval: "weekly"
+- package-ecosystem: "maven"
+  directory: "/"
+  schedule:
+    interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,8 +1,11 @@
 version: 2
 updates:
-- package-ecosystem:
-  - github-actions
-  - maven
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: daily
```
line-o, line 6: Do you believe it is necessary to check for new versions of GitHub actions daily?
`weekly` without specifying the time should be good enough for those, I think.
also the time can be omitted

dizzzz (approved):

.github/dependabot.yml
reinhapa: > Do you believe it is necessary to check for new versions of GitHub actions daily? `weekly` without specifying the time should be good enough for those, I think. also the time can be omitted

I just used the same settings as for the maven dependencies. Why do we have a time on the daily check anyway?

adamretter (approved):

PR #4847, opened 2023-04-02
Simplify Source interface

adamretter:
Simplifies an interface used for providing XQuery to eXist-db by removing its dependency on DBBroker.


reinhapa (approved):

line-o:
Cool stuff! This should wait for #4838 and rebased (as it modifies XQueryContext among other classes).

adamretter:
I didn't think #4838 was read yet? Does it need another round of review then? If #4838 is not ready, this PR should not wait for it IMHO.

line-o:
It's been ready for a few days now.

PR #4850, opened 2023-04-04
[bugfix] tails of subsequences off by one

### LINKED ISSUE #4830: [BUG] XQuery bug of node selection of tail() after subsequence()
**Describe the bug**
Give XPath Query
```
tail(subsequence((1 to 2),1,2))
```

Exist returns 0 element

**Expected behavior**

Should return one element 2

**To Reproduce**

Can just execute the query statement in http://exist-db.org/exist/apps/eXide/index.html

 - Build: latest development version built from source, commit dc0bfc8
 - Java: 17
 - OS: [Windows]

**Additional context**
- How is eXist-db installed? Jar installer

line-o:
### Description:

Tails of subsequences are no longer off by one.

### Reference:

fixes #4830

### Type of tests:

XQSuite tests


duncdrum (commented):
Ouch quite a find needs a rebase

line-o:
rebased onto current develop

duncdrum:
@line-o thx looks good to me

duncdrum (approved):

exist-core/src/main/java/org/exist/xquery/value/SubSequence.java
```diff
@@ -204,7 +204,7 @@ public Sequence tail() {
             return Sequence.EMPTY_SEQUENCE;
         }
 
-        return new SubSequence(fromInclusive + 1, toExclusive -1, sequence);
+        return new SubSequence(fromInclusive + 1, toExclusive, sequence);
```
adamretter, line 207: Nice!

exist-core/src/test/xquery/subsequence.xq
```diff
@@ -30,3 +30,26 @@ declare
...
+    %test:args(5,2)
+    %test:assertEmpty(4)
+    %test:args(1,3)
+    %test:assertEquals(2,3)
+    %test:args(3,3)
+    %test:assertEquals(4,5)
+    %test:args(1,5)
+    %test:assertEquals(2,3,4,5)
+function ss:tail($start, $length) {
+    subsequence((1 to 5), $start, $length) => tail()
```
adamretter, line 54: Can you remove the unnecessary arrow operator please? (for the reasons previously discussed about the ability to debug such constructs from Java).

adamretter (changes requested):

exist-core/src/test/xquery/subsequence.xq
line-o: Do you really think this warrants to reject this PR?

line-o: The test above has an arrow operator.
I also do used _this_ test to debug _and_ fix the issue. It is certainly possible.

reinhapa (approved):

exist-core/src/test/xquery/subsequence.xq
adamretter: > his warrants to reject this PR?

I am not rejecting the PR, I am asking for a very small change. Surely the purpose of the peer-review process is exactly for for these matters!

> The existing XQSuite test case above the newly added one has an arrow operator as well.

I don't see what bearing that has on your changes. If the existing code was perfect, then we would never need to make changes to eXist-db! When we are making changes, these days we follow the peer-review process, which hopefully leads to agreed improvements to the code base.

adamretter (changes requested):
The test-case can be simplified to its actual purpose by removing the (unnecessary) arrow operator.

line-o:
For the sake of getting this PR, and it's backport #4851, merged I will change the test.
I will open a separate issue to get to a consensus how to write XQSuite tests and have that documented somewhere.

adamretter (approved):
Thanks for the improvements. Now LGTM :-)

PR #4852, opened 2023-04-04
Separate static analyzer task/javadoc from test execution

reinhapa:
Combines Test & JavaDoc Workflow to one separating single tasks as follows:

- Licence check
- OWASP Scan
- Java Doc
- Test matrix OS for test executions only


line-o (approved):
This improves our current workflow. Thank you!

line-o:
Should be rebased onto current develop after #4853 was merged.

duncdrum (approved):

.github/workflows/ci-test.yml
```diff
@@ -1,15 +1,53 @@
...
+    name: Javadoc
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          distribution: temurin
+          java-version: ${{ env.DEV_JDK }}
+          cache: 'maven'
+      - run: mvn -V -B -q -T 2C install javadoc:javadoc -DskipTests -D'dependency-check.skip' -D'license.skip' --projects '!exist-distribution,!exist-installer' --also-make
```
adamretter, line 43: I am not sure if it makes sense to have the javadoc in a separate task.
As it is setup in this PR, the javadoc CI job will also call these Maven phases:

1. validate
2. initialize
3. generate-sources
4. process-sources
5. generate-resources
6. process-resources
7. compile
8. process-classes 
9. generate-test-sources
10. process-test-sources
11. generate-test-resources
12. process-test-resources
13. test-compile
14. process-test-classes
15. test
16. prepare-package
17. package
18. pre-integration-test
19. integration-test
20. post-integration-test
21. verify
22. install

Only phases 15 and 19, i.e. `test` and `integration test` will be reduced (but not omitted) in this job due to the `-DskipTests` short-cutting the `process` methods of the `maven-surefire-plugin` and `maven-failsafe-plugin`. As all of these phasse are already performed in the `Maven Test` CI job, separating this into a separate job has increased the amount of work required by the CI due to duplication of phases.

I think it would be better to reinstate it into the `Maven Test` CI job with a condition that it is only executed on the Linux runner (as that is the fastest runner).

.github/workflows/ci-test.yml
```diff
@@ -1,15 +1,53 @@
...
   test:
     name: (JDK ${{ matrix.jdk }} / ${{ matrix.os }}) Test
-    env:
-      MAVEN_OPTS: -XX:StartFlightRecording=maxsize=5g,disk=true,dumponexit=true,settings=default,filename=./
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        jdk: ['17']
+        jdk: [ '17' ]
```
adamretter, line 50: Should this be - `${{ env.DEV_JDK }}` ?

exist-parent/pom.xml
```diff
@@ -755,7 +755,7 @@
                 <plugin>
                     <groupId>org.jacoco</groupId>
                     <artifactId>jacoco-maven-plugin</artifactId>
-                    <version>0.8.9</version>
+                    <version>0.8.8</version>
```
adamretter, line 758: Why was the version of the jacoco plugin downgraded?

.github/workflows/ci-test.yml
```diff
@@ -21,9 +59,9 @@ jobs:
           cache: 'maven'
       - name: Maven Test
         timeout-minutes: 60
-        run: mvn -V -B -DtrimStackTrace=false -D"maven.resolver.transport"="wagon" clean verify
+        run: mvn -V -B verify -D'dependency-check.skip' -D'license.skip'
```
adamretter, line 62: Can we reinstate `-DtrimStackTrace=false` please?

adamretter (commented):
Some good changes, but the separate javadoc CI job doesn't make sense as it duplicates the majority of the work that has to be done.

exist-parent/pom.xml
adamretter: Okay I understand that it was due to a separate issue addressed in https://github.com/eXist-db/exist/pull/4855

.github/workflows/ci-test.yml
dizzzz: would this mean no unit tests anymore on windows/macos?

adamretter: No, it would mean that only the `javadoc` task would be run on Linux.

PR #4865, opened 2023-04-10
[6.x.x] Fixes to fn:replace, fn:tokenize, and fn:analyze-string

adamretter:
Backport of https://github.com/eXist-db/exist/pull/4864


reinhapa (approved):

dizzzz (approved):

joewiz:
@line-o Since the original PR for develop was merged in April, the only issue we've noticed with the bundled apps was a problem with a regex in the function-documentation app, which I fixed in https://github.com/eXist-db/exist/pull/4864. To ensure function-documentation continues to work for develop as well as 5.x and 6.x after these backports (this and https://github.com/eXist-db/exist/pull/4866) have been merged, I'll need to publish a new release of function-documentation. Since no other problems have been reported under develop, I think these are safe to merge. And I'll happily cut a new release of function-documentation with the fix.

As discussed on the Community Call just now, @reinhapa, @dizzzz, and I are in favor of merging this and the 5.x backport. As I noted in the notes, Patrick noted that when new releases of 5.x and 6.x are issued, the release notes should probably say something like, “Beware that this release fixes bugs in eXist's implementation of regular expressions used in fn:replace, fn:tokenize, and fn:analyze-string. If you rely on the previous buggy behavior, please do not upgrade.” And perhaps it could link to the relevant issues so users can investigate if their code does rely on the buggy behavior.

joewiz:
@line-o This was good to go. I should cut a new release of function-documentation, for folks using any of the affected develop-* branches. At latest, I should do so before the next release of the versions of eXist that receive this fix.

Issue #4609, opened 2022-11-07
[BUG] fn:transform not evaluating global params correctly


yamahito:
**Describe the bug**
When running a transform using `fn:transform()`, parameters set with a default value (e.g. `@select='/*/@xml:id'`) return an error.  Possibly this is because the context for the transformation?

```
err:XTTE0590 Could not transform input: An empty sequence is not allowed as the value of parameter $documentID
```

Additionally, when running the transformation using the `stylesheet-location` option, I get the following error (not sure whether or not it's related):

```
Failed to invoke method queryPT in class org.exist.xmlrpc.RpcConnection: org.exist.dom.persistent.NodeProxy cannot be cast to org.w3c.dom.Node
```

**Expected behavior**
I expect a successful transformation: here is the output when run with Saxon PE (some whitespace added):
```xquery
map{
    "output":
        <div class="toc">
            <div class="toc__header">
                <h4 class="title">Contents</h4>
            </div>
            <nav aria-label="Side navigation,,," class="toc__chapters">
                <ul class="chapters js-smoothscroll">
                    <li data-tei-id="pressrelease">
                        <a href="/output/pressrelease">Press Release</a>
                    </li>
                    <li data-tei-id="comp">
                        <a href="/output/comp">Main Book</a> (Documents 1 - 28)
                        <ul class="chapters__nested">
                            <li data-tei-documents="d1 d2 d3 d4 d5 d6 d7" data-tei-id="ch1">
                                <a href="/output/ch1">Chapter One</a> (Documents 1 - 7)
                            </li>
                            <li data-tei-documents="d8 d9 d10 d11 d12 d13 d14 d15" data-tei-id="ch2">
                                <a href="/output/ch2">Chapter Two</a> (Documents 8 - 15)
                            </li>
                            <li data-tei-documents="d16 d17 d18 d19 d20 d21" data-tei-id="ch3">
                                <a href="/output/ch3">Chapter Three</a> (Documents 16 - 21)
                            </li>
                            <li data-tei-documents="d22 d23 d24 d25 d26 d27 d28" data-tei-id="ch4">
                                <a href="/output/ch4">Chapter Four</a> (Documents 22 - 28)
                            </li>
                        </ul>
                    </li>
                </ul>
            </nav>
        </div>
}
```

**To Reproduce**

I've attached a sample collection that can be loaded to exist to replicate.
[fn_transform.zip](https://github.com/eXist-db/exist/files/9950043/fn_transform.zip)

```Xquery
[...]

PR #4883, opened 2023-04-25
Fix missing fn:transform global parameters

alanpaxton:
### Description:

Transformation fails to load global parameters from the stylesheet when it is loaded from eXist-db (rather than in memory).

Stylesheet document fetching into sources is not recognising as documents the documents it received from the DB (as opposed to ones which are constructed in memory). All our testing has been done with temporary in memory documents and we missed the case of a transformation source coming from the persistent db.

To fix this, we just have to recognise `org.exist.dom.persistent.DocumentImpl` as well as `org.exist.dom.memtree.DocumentImpl` as a valid document source; similarly, the case where a document can be a `NodeProxy` wasn’t recognised, we just need to fetch the node from the proxy when we recognise it as such, and proceed with using a source wrapping the resulting node as the stylesheet source.

### Reference:

[Github Issue](https://github.com/eXist-db/exist/issues/4609)

### Type of tests:

- eXide files detailed in the bug report now produce the correct transformation, rather than an error
- Added an XQSuite unit test transformation loading stylesheet from a resource: and input document from database added; confirmed this Unit Test failed before the changes and now passed.


adamretter (approved):
Thanks @alanpaxton LGTM

PR #4885, opened 2023-04-26
[hotfix] set enforce-index-use default value to strict

### LINKED ISSUE #4884: [BUG] enforce-index-use is omitting search results
`enforce-index-use` is a value that is set in `etc/conf.xml`  and documented as follows:
```
 enforce-index-use
               When set to "strict", eXist will not use a range index unless all
                collections in the context sequence define it. When set to
                "always", the query engine will still use an index, even if only
                one collection has it defined. It thus leaves it to the user to
                properly define indexes and if you forget to specify an index on
                a particular collection, it will be missing in the results.
``` 
The value, by default, is set to `always`, and according to the documentation, this means it will ignore any collection without an index defined.
Most users are unaware of this, for eg: #3620. I found it while trying to fix the issue just mentioned. The following is the code that lead me to this discovery: https://github.com/eXist-db/exist/blob/165bbc58a3cf126a1749000f9a794395e80c39c0/exist-core/src/main/java/org/exist/xquery/pragmas/Optimize.java#L342
I don’t think the default value should be `always`. Ignoring search results without users knowledge should not be the default case.
@joewiz @adamretter

marmoure:
Fixes #4884 
As discussed in #4884 we want to make the default value to be `strict`.

---
This open source contribution to the [exist](https://github.com/eXist-db/exist) project was commissioned by the Office of the Historian, U.S. Department of State, https://history.state.gov/.

### DIFF
--- a/extensions/indexes/indexes-integration-tests/src/test/java/org/exist/indexing/EnforceIndexUseTest.java
+++ b/extensions/indexes/indexes-integration-tests/src/test/java/org/exist/indexing/EnforceIndexUseTest.java
@@ -1,0 +1,194 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.indexing;
+
+import org.exist.EXistException;
+import org.exist.collections.Collection;
+import org.exist.collections.CollectionConfigurationException;
+import org.exist.collections.CollectionConfigurationManager;
+import org.exist.collections.triggers.TriggerException;
+import org.exist.dom.persistent.DefaultDocumentSet;
+import org.exist.dom.persistent.DocumentSet;
+import org.exist.dom.persistent.MutableDocumentSet;
+import org.exist.security.PermissionDeniedException;
+import org.exist.storage.BrokerPool;
+import org.exist.storage.DBBroker;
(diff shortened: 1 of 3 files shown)


joewiz:
Naturally, this PR would require a new major release. Today, this would mean it could be a candidate for eXist 7. So I'll set it for that milestone.

adamretter:
Tests will follow shortly...

extensions/indexes/indexes-integration-tests/src/test/java/org/exist/indexing/EnforceIndexUseTest.java
```diff
@@ -0,0 +1,156 @@
...
+            Collection oldRange = broker.getOrCreateCollection(transaction, XmldbURI.xmldbUriFor(TestConstants.TEST_COLLECTION_URI + "old-range-index"));
+            Collection noRange = broker.getOrCreateCollection(transaction, XmldbURI.xmldbUriFor(TestConstants.TEST_COLLECTION_URI + "no-range-index"));
+            broker.saveCollection(transaction, test);
+            broker.saveCollection(transaction, newRange);
+            broker.saveCollection(transaction, oldRange);
+            broker.saveCollection(transaction, noRange);
+
+            transact.commit(transaction);
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
```
reinhapa, line 130: Remove this catch and add an additional throws declaration instead

extensions/indexes/indexes-integration-tests/src/test/java/org/exist/indexing/EnforceIndexUseTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.exist.test.ExistEmbeddedServer;
+import org.exist.test.TestConstants;
+import org.exist.util.LockException;
+import org.exist.util.MimeType;
+import org.exist.util.StringInputSource;
+import org.exist.xmldb.XmldbURI;
+import org.exist.xquery.XPathException;
+import org.exist.xquery.XQuery;
+import org.exist.xquery.XQueryContext;
+import org.exist.xquery.value.Item;
```
reinhapa, line 26: Remove this unused import

reinhapa (changes requested):

extensions/indexes/indexes-integration-tests/src/test/java/org/exist/indexing/EnforceIndexUseTest.java
```diff
@@ -11,6 +11,7 @@
 import org.exist.security.PermissionDeniedException;
 import org.exist.storage.BrokerPool;
 import org.exist.storage.DBBroker;
+import org.exist.storage.txn.TransactionException;
```
reinhapa, line 14: The `TransactionException` import seems to be unused and should be removed...

reinhapa (changes requested):

reinhapa (approved):

joewiz:
@marmoure Thank you for this PR and adding the tests!

adamretter (approved):
LGTM. Thanks @marmoure

PR #4886, opened 2023-04-26
[bugfix] fn:matches error codes

line-o:
### Description:

Read the error code raised by the Saxon regular expression implementation.
This will now correctly raise FORX0002 when the pattern is invalid.

### Reference:

https://www.w3.org/TR/xpath-functions-31/#func-matches

### Type of tests:

added XQSuite tests

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
@@ -151,14 +151,10 @@ public class FunAnalyzeString extends BasicFunction {
             }
         } catch (final net.sf.saxon.trans.XPathException e) {
             switch (e.getErrorCodeLocalPart()) {
-                case "FORX0001":
-                    throw new XPathException(this, ErrorCodes.FORX0001, e.getMessage());
-                case "FORX0002":
-                    throw new XPathException(this, ErrorCodes.FORX0002, e.getMessage());
-                case "FORX0003":
-                    throw new XPathException(this, ErrorCodes.FORX0003, e.getMessage());
-                default:
-                    throw new XPathException(this, e.getMessage());
+                case "FORX0001" -> throw new XPathException(this, ErrorCodes.FORX0001, e.getMessage());
+                case "FORX0002" -> throw new XPathException(this, ErrorCodes.FORX0002, e.getMessage());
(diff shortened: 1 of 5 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
```diff
@@ -151,14 +151,10 @@ private void analyzeString(final MemTreeBuilder builder, final String input, Str
...
-                case "FORX0002":
-                    throw new XPathException(this, ErrorCodes.FORX0002, e.getMessage());
-                case "FORX0003":
-                    throw new XPathException(this, ErrorCodes.FORX0003, e.getMessage());
-                default:
-                    throw new XPathException(this, e.getMessage());
+                case "FORX0001" -> throw new XPathException(this, ErrorCodes.FORX0001, e.getMessage());
+                case "FORX0002" -> throw new XPathException(this, ErrorCodes.FORX0002, e.getMessage());
+                case "FORX0003" -> throw new XPathException(this, ErrorCodes.FORX0003, e.getMessage());
+                default -> throw new XPathException(this, e.getMessage());
```
joewiz, line 157: Should this match the change in https://github.com/eXist-db/exist/pull/4886/commits/5f974298fa6f8eb83d654fe9c9f8f941200909b4#diff-b83837e14cf949b3eff92fc056203d1a62e5f5c444fb8891703e1b220c70e4e5R144 ? (I see `ErrorCodes.ERROR` is not present here.)

line-o: Oh, good catch!

line-o:
XQTS results are in:

> failures="-307"

Astounding what one can achieve with the smallest of changes.

line-o:
I would really love to see the percentage of passing tests in the XQTS results.
To save you time I calculated it:

| before | after |
|---|---| 
| 79,759% | 80,726% |

line-o:
now an additional two XQTS tests pass ... :eyes:

duncdrum (commented):
Nice one again. Could you add one valid pattern to the two test functions for prosperities sake?

dizzzz (approved):

joewiz:
With two approvals from core devs, this can be merged. Thanks @line-o and thanks @adamretter and @dizzzz!

Issue #4729, opened 2023-02-06
[BUG] System configuration leads to empty jetty-http.xml and jetty-ssl.xml


ahenket:
**Describe the bug**
The macOS version of eXist-db comes with a minimal UI that allow access to System Configuration. This, among other things lets you change the ports that eXist-db runs on. Our preferred ports are 8877 for http and 8843 for https.

However setting them using the eXist-db system bar feature System Configuration is unreliable.

**Expected behavior**
When you set the new ports you expect those to actually be applied

**To Reproduce**

- Copy eXist-db.app to Applications folder and start it up from there (no other changes)
  - eXist-db.app will present you the system configuration screen
- Change the Jetty Ports and Save. http = 8877 and ssl = 8843
  - eXist-db will now continue to load. In the exist.log you will see it loads using the default ports and ignore what you Saved
- While the db is running, choose System Configuration from the eXist-db menubar icon
- Change the Jetty Ports and Save again. http = 8877 and ssl = 8843
  - As soon you hit Save you will see that the once 1kb jetty-http.xml and jetty-ssl.xml will be turned into 0kb files

**Screenshots**

**Context (please always complete the following information)**
 - Build: [eXist-6.2.0]
 - Java: [java version "15.0.2" 2021-01-19]
 - OS: [Mac OS X 13.1 Ventura]

**Additional context**
- Installed from dmg

ahenket:
Side note that might be related: opening the exist.log from this same system bar utility also fails to happen since 6.0.1/6.2.0. A much beloved feature as the log is deep inside the .app.

bansp:
I'd like to confirm this for: 

[eXist Version : 6.2.0] 
[eXist Build : 2023-02-04T22:42:29Z] 
[Git commit : c8fa4958b6d4a50bd0cba7f3e76a150226414187] 
[Git commit timestamp : 20230204223916] 
[Operating System : Linux 4.15.0-206-generic amd64]
java version "1.8.0_111"  

Installed from .jar


This is the state after having adjusted the ports in the configuration window that shows up at first run (and without that action having consequences for jetty, but you see the consequences in the renamed files:

![image](https://user-images.githubusercontent.com/312137/227071876-b4bb54bd-f677-4106-bc1d-72f958b013a9.png)

Then I repeated the action that @ahenket mentions: invoked the configuration window, modified the ports, saved, agreed to a restart, and noticed an error at the moment that jetty http configuration loaded: premature end of file. On the disk, it looked as follows:

[...]

PR #4899, opened 2023-05-02
[fix] jetty configuration changes don't stick, and eventually crash eXist-db

alanpaxton:
### Description:

Updated jetty-http.xml and jetty-ssl.xml were being written by the configurator with a bad DTD, which on subsequent read by the configurator caused an exception, so that the NEXT versions of jetty-http.xml and jetty-ssl.xml were made empty. This then caused a startup crash.

1. Fix the DTD reference.

Also, configurator was reading the wrong field(s) for the http configuration, so

2. make configurator read the correct fields for jetty http configuration, so that subsequent edits of an already re-configured jetty don’t end up re-introducing the default (8080)

### Reference:

Closes https://github.com/eXist-db/exist/issues/4729

### Type of tests:

Unfortunately it is not simple to automate this testing. We have built a test distribution with the changes, installed it (`java -jar ./exist-installer/target/exist-installer-7.0.0-SNAPSHOT.jar`) and confirmed through the use of ./bin/launcher.sh in the new installation that modifications of the jetty configuration are stored, restored and used on startup as the jetty ports. A distribution built without the change allowed us to reproduce the failed configuration changes, crashes and empty files produced in the initial problem report.

### DIFF
--- a/exist-core/src/main/java/org/exist/launcher/ConfigurationDialog.java
+++ b/exist-core/src/main/java/org/exist/launcher/ConfigurationDialog.java
@@ -97,6 +97,12 @@ public class ConfigurationDialog extends JDialog {
             if (ports.containsKey("jetty.port")) {
                 httpPort.setValue(ports.get("jetty.port"));
             }
+            if (ports.containsKey("jetty.http.port")) {
+                httpPort.setValue(ports.get("jetty.http.port"));
+            }
+            if (ports.containsKey("ssl.port")) {
+                sslPort.setValue(ports.get("ssl.port"));
+            }
             if (ports.containsKey("jetty.ssl.port")) {
                 sslPort.setValue(ports.get("jetty.ssl.port"));
             }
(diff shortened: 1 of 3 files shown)


adamretter (approved):
LGTM. Thanks @alanpaxton

dizzzz (approved):

adamretter:
@duncdrum How would you see using the Launcher Configuration GUI working from inside a Docker container in CI exactly?

duncdrum:
@adamretter 
> Also, configurator was reading the wrong field(s) for the http configuration, so

make configurator read the correct fields for jetty http configuration, so that subsequent edits of an already re-configured jetty don’t end up re-introducing the default (8080)

does this only refer to the configurator of the initial launch agent, or to the configurator that is launched whenever you start exist?

alanpaxton:
The configurator that is started on initial, and whenever you run "System Configuration" from the menu bar, is the same configurator code with the same problem.

Issue #4518, opened 2022-08-16
[BUG] The first argument of `castable as` is apparently not being atomized


cmsmcq:
**Describe the bug**
When the first argument to `castable as` is a reference to an attribute, eXist evaluates the expression as false even when the value of the attribute is in fact castable as the given type.   

For example, given an element `$e` bound to the element `<e n="xyz"/>`, the expression `$e/@n castable as xs:NCName` returns false, although the string "`xyz`" is indeed castable to xs:NCName.  The expression `$e/@n/string() castable as xs:NCName` returns the expected value (true), but the call to `string()` should not be necessary.

This suggests (at least to someone unfamiliar with the internals of eXist) that the left-hand argument of `castable as` is not being atomized as required by sections 3.18.4 and 3.18.3 of the XQuery spec.

**Expected behavior**
The left-hand argument of `castable as` should be evaluated (in the example, this yields an attribute node) and then atomized (which involves invoking the data() function to find the typed value of the attribute node), and the value thus found should be checked to see if it can be cast to the indicated type. 

**To Reproduce**
Run the query
````
xquery version "3.1";

(: castability test :)
let $doc := <e n="e">
    <f n="ixml"/>
    <f n="comment"/>
    <f n="apple-jack"/>
    <f n="123"/>
    <f n="élève"/>
 </e>
 
return <tests>{
for $e in $doc/descendant-or-self::*[exists(@n)] 
let $n := $e/@n,
    $s := string($n)
return <test name="{$s}" n-castable="{$n castable as xs:NCName}" s-castable="{$s castable as xs:NCName}"/>
}</tests>
````
In my tests, running the query shown in eXist 6.0.1 produces
````
<tests>
    <test name="e" n-castable="false" s-castable="true"/>
    <test name="ixml" n-castable="false" s-castable="true"/>
    <test name="comment" n-castable="false" s-castable="true"/>
    <test name="apple-jack" n-castable="false" s-castable="true"/>
    <test name="123" n-castable="false" s-castable="false"/>
    <test name="élève" n-castable="false" s-castable="true"/>
</tests>
````
Note that when the first argument of `castable as` is an attribute, the value returned is always false; when the first argument is a string, the value returned depends on whether the string is in the value space of xs:NCName.

For comparison, BaseX and Saxon (HE) return the same value for the attribute and the string in each case:
````
<?xml version="1.0" encoding="UTF-8"?>
<tests>
   <test name="e" n-castable="true" s-castable="true"/>
[...]

PR #4900, opened 2023-05-02
[fix] castable as for untyped atomic value(s)

alanpaxton:
### Description:

It appears perfectly acceptable for UntypedAtomicValue.convertTo to convert to subtypes of Type.STRING as StringValue(… value, requiredType). The conversion was omitted, and the omission was breaking “castable as” for Type.NCNAME in particular.

These conversions are already acceptable for StringValue(s), and are borrowed from there.

### Reference:

Closes https://github.com/eXist-db/exist/issues/4518

### Type of tests:

xquery unit test added which tests a variety of atomic values `castable as` an NCName successfully. Failing prior to the code change.


dizzzz (approved):

reinhapa (approved):

adamretter:
There are two additional XQTS failures caused by this PR that need to be examined...

Issue #4901, opened 2023-05-03
[BUG] Installed EXPath (Java) Packages are not portable between file-systems


adamretter:
When an EXPath package is installed into eXist-db, a copy of it is written to a folder in `$EXIST_HOME/data/expathrepo`.
For EXPath packages that provide Java class files (e.g. XQuery extension modules), an additional file is written to `$EXIST_HOME/data/expathrepo/<my-package>/.exist/classpath.txt`.

* This has been tested against eXist-db 6.1.0-SNAPSHOT.

Unfortunately the `classpath.txt` file appears to contain absolute paths. For example with the EXPath Crypto 6.0.1 module, the `classpath.txt` file contains:
```
E:\Apps\eXist-db\data\expathrepo\crypto-6.0.1\content\crypto-java-lib-1.8.0.jar
E:\Apps\eXist-db\data\expathrepo\crypto-6.0.1\content\expath-crypto-module-exist-java-lib-6.0.1.jar
```

This means that should you copy your `data` directory to a different location (and/or Operating System e.g. Windows <-> Linux) and try to start up eXist-db, it will fail to load all of these Java Classes for an EXPath Module as the .jar files will be in the wrong location (i.e. they do not match the absolute paths in `.exist/classpath.txt`).

Neither does eXist-db report at startup that it cannot access these paths. Instead they are silently ignored if they cannot be accessed by eXist-db.

At the very least, (a) the inaccessible paths should be logged, and (b) the `classpath.txt` file should use a path that is relative to the package directory (or perhaps the content sub-directory) instead to enable portability. The above would then instead look something like:
```
content/crypto-java-lib-1.8.0.jar
content/expath-crypto-module-exist-java-lib-6.0.1.jar
```

However, I wonder why the `.exist/classpath.txt` file is needed at? We can see that the required information is already present in the file `$EXIST_HOME/data/expathrepo/<my-package>/exist.xml`, e.g.:
```xml
<package xmlns="http://exist-db.org/ns/expath-pkg" xmlns:xs="http://www.w3.org/2001/XMLSchema">
   <java>
      <namespace>http://expath.org/ns/crypto</namespace>
      <class>org.expath.exist.crypto.ExistExpathCryptoModule</class>
   </java>
   <jar>expath-crypto-module-exist-java-lib-6.0.1.jar</jar>
   <jar>crypto-java-lib-1.8.0.jar</jar>
</package>
```

[...]

PR #4902, opened 2023-05-03
Ensure that EXPath packages installed in $EXIST_HOME/data/expathrepo are filesystem portable

adamretter:
Closes https://github.com/eXist-db/exist/issues/4901

Previously classpath locations were stored as absolute paths, they are now stored as relative paths. Care has been taken to ensure that the code remains backwards compatible with existing eXist-db data dirs that are already using absolute paths.

### DIFF
--- a/exist-core/src/main/java/org/exist/repo/ClasspathHelper.java
+++ b/exist-core/src/main/java/org/exist/repo/ClasspathHelper.java
@@ -170,10 +170,17 @@ public class ClasspathHelper implements BrokerPoolService {
                 try (final BufferedReader reader = Files.newBufferedReader(cp)) {
                     String line;
                     while ((line = reader.readLine()) != null) {
-                        if (!Files.exists(Paths.get(line))) {
-                            LOG.warn("Unable to add '" + line + "' to the classpath for EXPath package: " + pkg.getName() + ", as the file does not exist!");
+                        Path p = Paths.get(line);
+                        if (!p.isAbsolute()) {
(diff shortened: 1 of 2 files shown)


dizzzz (approved):

exist-core/src/main/java/org/exist/repo/ClasspathHelper.java
```diff
@@ -169,7 +170,18 @@ private static void scanPackageDir(Classpath classpath, Path module) throws IOEx
...
+                            final FileSystemStorage.FileSystemResolver res = (FileSystemStorage.FileSystemResolver) pkg.getResolver();
+                            p = res.resolveComponentAsFile(line);
+                        }
+                        p = p.normalize().toAbsolutePath();
+
+                        if (!Files.exists(p)) {
+                            LOG.warn("Unable to add '" + p + "' to the classpath for EXPath package: " + pkg.getName() + ", as the file does not exist!");
+                        } else {
+                            classpath.addComponent(p.toString());
+                        }
```
reinhapa, line 184: it may would be easier to read if you turn the if statement and remove the inversion within the if clause like so:
```
 if (Files.exists(p)) {
     classpath.addComponent(p.toString());
 } else {
     LOG.warn("Unable to add '" + p + "' to the classpath for EXPath package: " + pkg.getName() + ", as the file does not exist!");
 }
 ```

reinhapa (changes requested):
@adamretter just a minor issue

reinhapa (approved):

PR #4909, opened 2023-05-08
🦄 refactor(docker-pom): no tagging from name

duncdrum:
cleanup docker and deploy step on CI

### DIFF
--- a/exist-docker/pom.xml
+++ b/exist-docker/pom.xml
@@ -49,7 +49,7 @@
         <assemble.dir>${project.build.directory}/exist-docker-${project.version}-docker-dir</assemble.dir>
         <exist.uber.jar.filename>exist.uber.jar</exist.uber.jar.filename>
         <docker.tag>latest</docker.tag>
-        <docker.debug.tag>debug</docker.debug.tag>
+<!--        <docker.debug.tag>debug</docker.debug.tag>-->
     </properties>
 
     <dependencies>
@@ -213,10 +213,11 @@
                   <verbose>true</verbose>
                     <images>
                         <image>
-                            <name>existdb/existdb:%v</name>
+                            <name>existdb/existdb</name>
                             <alias>exist</alias>
                             <build>
                                 <tags>
+                                    <tag>%v</tag>
                                     <tag>${docker.tag}</tag>
                                 </tags>
                                 <dockerFile>${project.build.outputDirectory}/Dockerfile</dockerFile>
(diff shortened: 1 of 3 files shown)


duncdrum:
Disabled experimental tags, no clue why I can no longer push to docker hub, either a change in the credentials, or problem with secret sharing. 

More importantly, since the build was changed to include two `<image>` elements, there was a problem with overwriting of production images with the debug images contents. This is no longer the case, and fixed by the changes to `<name>`

reinhapa (approved):

exist-docker/pom.xml
```diff
@@ -213,22 +213,23 @@
...
                             </build>
                         </image>
                         <image>
-                            <name>existdb/existdb:%v-DEBUG</name>
+                            <name>existdb/existdb:debug</name>
                             <alias>exist-debug</alias>
                             <build>
                                 <tags>
-                                    <tag>${docker.debug.tag}</tag>
+                                    <tag>%v-debug</tag>
```
adamretter, line 232: This is not the same as before. The name was previously `x.x.x-DEBUG` or `x.x.x-SNAPSHOT-DEBUG`.

adamretter (changes requested):
There is a breaking change in the name from `DEBUG` to `debug` that needs to be fixed.

Also as far as I can see the change between putting the primary tag in the `<name>` or the `<tag>` field doesn't actually change anything. Can you explain what this changes in the outcome in practice?

duncdrum:
@adamretter yes the build command that was there overwrote the `:latest` image with the contents of the `debug` image, it also let to errors on the CI when trying to push images tagged with just the version number.

no caps `debug` is now applied consistently how is that a breaking change?

exist-docker/pom.xml
duncdrum: the names were  `existdb/existdb:debug` or `existdb/existdb:X.X.X-DEBUG` with `%v` also creating `existdb/existdb:latest` for semVers ending in `-SNAPSHOT` which was causing problems. Since that should no be created using the Debug Dockerfile

adamretter: > with %v also creating existdb/existdb:latest
Yes, but that was in a separate`<image>` block and so should not impact this.

adamretter:
> no caps debug is now applied consistently how is that a breaking change?

Because one of them should not be changed to be in lower-case as it will break existing tooling.

adamretter:
> yes the build command that was there overwrote the :latest image with the contents of the debug image, it also led to errors on the CI when trying to push images tagged with just the version number.

Okay, I understand what you were saying was happening previously *may* have been a problem (I have not attempted to reproduce yet), but I don't understand how your changes would change the behaviour at all, as far as I can see your changes just change the names of various tags.

exist-docker/pom.xml
duncdrum: Can you point me to the docs that speak of this isolation? What we saw on ci here, matches my reading of their docs on multiple images.

line-o: If this change fixes a problem maybe it is better to change the tooling to now look for the new tag?
Alternatively, is it OK and sufficient to just uppercase from `%v-debug` to `%v-DEBUG`?

dizzzz:
what is the state of this discussion?

duncdrum:
@dizzzz the overwriting no longer happens during normal operation, the images are based on the same commit, but clearly different in contents, as expected. 

Screenshot 2025-01-06 at 21 26 28

 it might still happen during releases but that ll now require a different fix.

PR #4910, opened 2023-05-09
Fix concurrency issues in Performance Stats code

adamretter:
Cleanup the code and fix concurrency issues in the Performance Stats code


exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
```diff
@@ -330,8 +330,7 @@ public NodeSet preSelect( Sequence contextSequence, boolean useContext ) throws
         
         // if the right hand sequence has more than one item, we need to merge them
         // into preselectResult
-        if (rightSeq.getItemCount() > 1)
-        	{preselectResult = new NewArrayNodeSet();}
+        if (rightSeq.getItemCount() > 1) {preselectResult = new NewArrayNodeSet();}
```
dizzzz, line 333: formatting is not worse or better :-)

dizzzz:
Codewise it looks OK to me. Better JMX support?

exist-core/src/test/xquery/optimizer/expressions.xqm
```diff
@@ -89,39 +89,39 @@ function ot:cleanup() {
 declare
     %test:stats
     %test:args("Rudi Rüssel")
-    %test:assertXPath("$result//stats:index[@type = 'range'][@optimization = 2]")
+    %test:assertXPath("$result//stats:index[@type eq 'range'][@optimization-level eq 'Optimized']")
```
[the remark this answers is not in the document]
adamretter: @joewiz yes, indeed

exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
adamretter: Hmm I think IntelliJ did that one by itself - I don't like it!

line-o:
Please update the description of this PR to include information about the breaking change that it introduces.

line-o (changes requested):
This PR introduces breaking changes and fails to communicate that clearly in the commit messages.
Please extract the breaking changes into a single commit so that we can discuss them separately.

adamretter:
> Please extract the breaking changes into a single commit so that we can discuss them separately.

This is not something we have ever done before! So why do you believe that this PR is "special" in some way to any others that target an upcoming major release?

line-o:
As the changes are not necessary to fix concurrency issues they should not be part of this PR.

dizzzz (approved):

dizzzz (commented):
ah ok, so some additional updates need to be put in place in apps to keep things compatible.

exist-core/src/test/xquery/optimizer/expressions.xqm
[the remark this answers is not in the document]
dizzzz: hmmmm ok. then we need to have a plan to update monex here as well, whilst keeping that app compatible with older exist versions?

exist-core/src/test/xquery/indexing/range.xml
```diff
@@ -78,35 +78,35 @@
         <task>Test empty element content</task>
         <code>doc("/db/test/text1.xml")//a[. = 'A1']</code>
         <xpath xmlns:prof="http://exist-db.org/xquery/profiling">
-            //prof:index[@type = 'range'][@optimization = 2]
+            //prof:index[@type eq 'range'][@optimization = 'Optimized']
```
joewiz, line 81: You might consider changing `=` to `eq` - to match the analogous changes in this PR to `optimizer/expressions.xqm`.

exist-core/src/main/java/org/exist/xquery/PerformanceStatsImpl.java
```diff
@@ -0,0 +1,421 @@
...
+                builder.endElement();
+            }
+            for (final IndexStats stats : indexStats.values()) {
+                attrs.clear();
+                attrs.addAttribute("", "type", "type", "CDATA", stats.indexType);
+                attrs.addAttribute("", "source", "source", "CDATA", stats.source + " [" + stats.line + ":" +
+                        stats.column + "]");
+                attrs.addAttribute("", "elapsed", "elapsed", "CDATA", Double.toString(stats.executionTime / 1000.0));
+                attrs.addAttribute("", "calls", "calls", "CDATA", Integer.toString(stats.usageCount));
+                attrs.addAttribute("", "optimization-level", "optimization", "CDATA", stats.indexOptimizationLevel.name());
```
line-o, line 397: We could have the old behaviour with
```
attrs.addAttribute("", "optimization", "optimization", "CDATA", stats.indexOptimizationLevel.ordinal());
```

exist-core/src/main/java/org/exist/xquery/PerformanceStats.java
```diff
@@ -19,359 +19,100 @@
...
-                return source.equals(stats.source) && type == stats.type &&
-                        line == stats.line && column == stats.column;
-            }
-            return false;
-        }
-    }
-
-    public enum OptimizationType {
+    enum OptimizationType {
         PositionalPredicate
```
reinhapa, line 40: This enumeration value should be UPPER in accordance to the java standards.

exist-core/src/test/xquery/indexing/range.xml
adamretter: Good spot!

exist-core/src/main/java/org/exist/xquery/PerformanceStats.java
adamretter: Agreed! Not sure what I was thinking there!

exist-core/src/main/java/org/exist/xquery/PerformanceStats.java
```diff
@@ -19,359 +19,100 @@
...
-        functions.clear();
-        indexStats.clear();
-        optimizations.clear();
-    }
-}
\ No newline at end of file
+    enum IndexOptimizationLevel {
+        None,
+        Basic,
+        Optimized;
```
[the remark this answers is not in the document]
adamretter: Agreed! Not sure what I was thinking there!

exist-core/src/test/xquery/indexing/range.xml
adamretter: Done.

exist-core/src/main/java/org/exist/xquery/PerformanceStats.java
adamretter: Done.

exist-core/src/main/java/org/exist/xquery/PerformanceStats.java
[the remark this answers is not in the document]
adamretter: Done.

exist-core/src/main/java/org/exist/xquery/PerformanceStatsImpl.java
adamretter: You should never use the `ordinal` of an Enum for serialization or deserialization.

reinhapa: @line-o I think we agreed at the last meeting to leave the names in favor of numerical values

line-o:
@adamretter In yesterdays community call I was convinced that speaking labels are favourable over magical integer constants. We also found that the attribute name returned by the stats package should stay `@optimization` and not be changed to `@optimization-level` as it is both shorter and also better to understand. This is especially true for the case where no index is found and thus no optimisation can be used.
What are your thoughts on this?

adamretter:
> @optimization and not be changed to @optimization-level

I intentionally changed this because it was misleading. It does not describe the optimization that was performed, which is what it previously implies. Rather it describes the "level of optimization" that was performed - and this is how it was described in the Java code. So I corrected it so that it actually reflects what it is.

adamretter:
@reinhapa @dizzzz Is this one now ready to go?

dizzzz:
during the teleconference we talked about the naming:

   enum IndexOptimizationLevel {
         NONE,
         BASIC,
         OPTIMIZED;
     }

OptimizationLevel= OPTIMIZED does not reflect i kind of "amount" of optimisation was applied. "FULL" would do that much better, or "OPTIMAL". Maybe it is a kind language thingy ?

joewiz:
How about if we expressed this information using 2 boolean attributes instead of 1:

1. index-present=yes|no ("yes" when index can be found; "no" when not)
2. fully-optimized=yes|no ("yes" when all dependencies present; "no" when further preparation is needed or when index is not present)

So there are still 3 possibilities, but they're expressed explicitly through the combination of these two boolean attributes:

1. No index: index-present=no + fully-optimized=no
2. Basic: index-present=yes + fully-optimized=no
3. Full: index-present=yes + fully-optimized=yes

line-o:
@joewiz I like your proposed approach but would like to shorten the attribute names.

`<stats:index used="yes|no" optimized="yes|no" />`

| level | label | @used | @optimized |
| ---- | ---- | ---- | ---- |
| 0 | No Index | no | no |
| 1 | Basic | yes  | no |
| 2 | Full | yes | yes |

adamretter:
@joewiz My thoughts:

> index-present=yes|no ("yes" when index can be found; "no" when not)

That seems fine.

> fully-optimized

I don't think this is a good term. There are "degrees" of optimisation. There are also optimisations that can be applied without an index, and IMHO we should not close the door of being able to report on them in future.

adamretter:
@line-o My thoughts:

> optimized="yes|no" 

This has the same problem as before. This does not express the degrees of optimisation possible.

line-o:
@adamretter there is only one level of optimisation possible and only if an index is found/used.

adamretter:
> there is only one level of optimisation possible and only if an index is found/used.

I am afraid that is not correct.

joewiz:
As discussed on the Community Call today, we seem agreed on the idea of returning two boolean attributes and that one should be called `index-present` (or `used`), and we are still seeking an acceptable name for the attribute I suggested, `fully-optimized`. We noted that these attributes are children of the `<stats:index>` element, and therefore that the attributes only describe index-related performance statistics - and not any other "optimisations that can be applied without an index". Perhaps this simplifies the criteria for these attributes' names? Does anyone have any suggestions for a good name?

adamretter:
As far as I am aware from studying the code it is not possible to present a result like `index-present=yes|no` at the position in the code where the index is queries. There is really only 1 big Lucene Index behind the scenes that may or may not contain things you are looking for.

The Lucene Index in eXist-db in fact does not actually know if there is an index or not for the particular expression it is looking up, instead it queries Lucene based on just the element or attribute name, which often returns more results than it should, later in other expressions these results are filtered out.

line-o:
If the index use is not really consistent, then it is misleading to even present that data under `<stats:index>`.
How about renaming the element then to `stats:optimization` to free it from the notion it is strictly tied to indexes?
```xml
<stats:optimiziation level="[none|basic|full]">
```

adamretter:
>  to even present that data under <stats:index>.
> How about renaming the element then to stats:optimization

I am not sure that makes sense as these optimizations are index guided.

As this has been open now for 6 months, and is a clear improvement on the status quo, I suggest we just merge this rather than splitting hairs over naming, and move on. It can always be improved in the future as can everything else in eXist-db.

@dizzzz @reinhapa Can you discuss together, and consider merging this please?

dizzzz:
Does this require an money update? if not, I am happy.

dizzzz (approved):

adamretter:
> Does this require an money update?
I am happy to receive any money

dizzzz:
I ment.... Monex  :-P

adamretter:
@dizzzz Yes, see the discussion here: https://github.com/eXist-db/exist/pull/4910#discussion_r1188852687

adamretter:
@reinhapa Would you mind adding a review please?

reinhapa (approved):

dizzzz:
Thnx @adamretter;

One (important) question remains... with the change, is Monex still working? ( @joewiz )

adamretter:
Yes, I think it may be possible that Monex will need a small change, but I have not tested Monex and I am not really involved in its development; so it is probably best to get one of the Monex developers to test/fix it once this PR has been merged. I did try and answer this before, but perhaps I wasn't clear enough (See [here](https://github.com/eXist-db/exist/pull/4910#issuecomment-2324240004) and [here](https://github.com/eXist-db/exist/pull/4910#discussion_r1188852687))

joewiz:
The changes to Monex should be straightforward. I've opened an issue for this: https://github.com/eXist-db/monex/issues/273.

dizzzz:
With the merge there were still issues ?

Can't check well on my phone ...

PR #4913, opened 2023-05-10
[6.x.x] Ensure that EXPath packages installed in $EXIST_HOME/data/expathrepo are filesystem portable

adamretter:
Backport of https://github.com/eXist-db/exist/pull/4902


reinhapa (approved):

dizzzz:
the situation with "org.owasp:dependency-check-maven" is getting annoying, as in GHA most of the builds fail;
is it a matter of removing caches? or should we disable it?

PR #4927, opened 2023-05-16
Remove pre-1.0 syntax for variable declarations

joewiz:
### Description:

This syntax has been deprecated in eXist [since 2006](https://github.com/eXist-db/exist/commit/1b99e4544d00d4c5602eb8b777498a6b2e30691f), and not allowed in XQuery [since 1.0](https://www.w3.org/TR/2010/REC-xquery-20101214/#id-variable-declarations).

As a breaking change, this needs to wait until a major release. I've tagged it as belonging to eXist 7.0.0.

I searched all eXist stock apps, as well as my own collection of app and library packages, and found none using the deprecated syntax.

### Reference:

- See above

### Type of tests:

- Updated tests included


duncdrum (approved):
Nice find, and yes 7.0.0 is good place

adamretter (approved):
LGTM

PR #4969, opened 2023-06-28
[build] Use original CVE files

dizzzz:
For a long time the retrieval of the owasp data is not reliable. This is a test to verify if the details configuration works more reliable.

... wait for build

### DIFF
--- a/exist-parent/pom.xml
+++ b/exist-parent/pom.xml
@@ -599,10 +599,6 @@
                     <artifactId>dependency-check-maven</artifactId>
                     <version>8.3.1</version>
                     <configuration>
-                        <!-- Use Evolved Binary's mirror of nvd.nist.gov -->
-                        <cveUrlModified>https://nvd.mirror.evolvedbinary.com/feeds/json/cve/1.1/nvdcve-1.1-modified.json.gz</cveUrlModified>
-                        <cveUrlBase>https://nvd.mirror.evolvedbinary.com/feeds/json/cve/1.1/nvdcve-1.1-%d.json.gz</cveUrlBase>
-
                         <!-- The OSS Index Server (https://ossindex.sonatype.org) can sometimes be flaky -->
                         <ossIndexWarnOnlyOnRemoteErrors>true</ossIndexWarnOnlyOnRemoteErrors>
 


adamretter:
Interesting!

Before the mirror we would hit the problem that we exceeded the number of requests allowed to that server from github.com.

The mirror that we operate uses this tool to sync from NIST - https://github.com/stevespringett/nist-data-mirror
I have also enquired previously about an updated tool using the newer v2 API - https://github.com/EugenMayer/nist-data-mirror/issues/2

The errors that I was seeing seemed to indicate that the problem was not the mirror but rather that there is some concurrency bug in the [OWASP Dependency Plugin for Maven](https://github.com/jeremylong/DependencyCheck) whereby its database is becoming corrupt and/or some of the files are corrupt. I have opened this issue - https://github.com/jeremylong/DependencyCheck/issues/5798

My feeling it is likely not the mirror, but that clearing the Maven caches in GItHub Actions that often fixes this. Updating the pom.xml files causes GitHub Actions to invalidate the caches.

dizzzz:
thnx adam, I remember now what you wrote.
previously I think that we had a lot of similar HTTP calls during the many builds we had, but that is reduced to one. Maybe that helps.... but it could also be that this build is accidently working OK :-/

thnx for the issue report!

dizzzz:
could it be that your mirrored files are somehow corrupted?

adamretter:
> could it be that your mirrored files are somehow corrupted?

@dizzzz I did check this in the past to make sure it wasn't the case, let me just check again quickly...

adamretter:
@dizzzz So I do see some differences between my mirror (synced every 2 hours) and the latest upstream:

```
diff /tmp/old-clone.txt /tmp/new-clone.txt
25,26c25,26
< MD5(nvdcve-1.1-2010.json)= 3826b41b21ffdb4ba8dc914c3cba3a66
< MD5(nvdcve-1.1-2010.json.gz)= 8915d5111e00b591cc975c2aa6d9c958
---
> MD5(nvdcve-1.1-2010.json)= a17c75e6d28cae47efe905bb2d68c94e
> MD5(nvdcve-1.1-2010.json.gz)= aaf9acf04d2cbe3da7129ed17fd639b7
28,29c28,29
< MD5(nvdcve-1.1-2011.json)= b7b4c80ad04c0f624e2b0c2f344c1b4c
< MD5(nvdcve-1.1-2011.json.gz)= d2c0868258eee6d551c48a978e1c280f
---
> MD5(nvdcve-1.1-2011.json)= 690e559d5af21af0843061f99eb9b97d
> MD5(nvdcve-1.1-2011.json.gz)= 4eb08e7b6b6b549ff92ef28ef2b5e924
34,35c34,35
< MD5(nvdcve-1.1-2013.json)= 4b156f483aac63024931f698ed4600e2
< MD5(nvdcve-1.1-2013.json.gz)= 08961501e523325f6892abc8f72f2566
---
> MD5(nvdcve-1.1-2013.json)= 736f1ab24c93c177fd36d1ba1260f186
> MD5(nvdcve-1.1-2013.json.gz)= 81be04af85757fab823153cf0eeb36f5
```

This does not necessarily mean the files that I have are corrupt. It could mean that they have been changed upstream in the last 2 hours.

line-o:
Can this be merged? It looks like it would finally fix the dependency check task.

reinhapa (approved):

adamretter:
@reinhapa My fear is that if we merge this then we will go back to getting frequent `HTTP 429 Too Many Requests` errors.

I have been trying to diagnose the root cause of the error here: https://github.com/jeremylong/DependencyCheck/issues/5798#issuecomment-1630758033

I am happy however for you to proceed however you wish :-)

dizzzz:
I agree the 'real' solution you are chasing is the best option; On the otherhand, we did already reduced the nr of calls in our build dramatically, potentially having a (just) better working situation.

reinhapa:
> I have been trying to diagnose the root cause of the error here: [jeremylong/DependencyCheck#5798 (comment)](https://github.com/jeremylong/DependencyCheck/issues/5798#issuecomment-1630758033)

@adamretter I will be happy to see a PR as soon this fix for https://github.com/jeremylong/DependencyCheck/issues/5798 is available,,

adamretter:
> On the otherhand, we did already reduced the nr of calls in our build dramatically

That won't make any difference at all. This isn't about how many calls *our* build makes. The `HTTP 429 Too Many Requests` occurs because of *all* projects from *everyone* on GitHub.com that call the NIST CVE service.

PR #4973, opened 2023-07-06
[fix] cardinality of CastExpression#toFunction

### LINKED ISSUE #4971: [BUG] type cast after arrow operator with empty-sequence as input causes cardinality error
**Describe the bug**
Forwarding an non-existent node to a function that accepts an empty-sequence returns err:XPTY0004 
`Invalid cardinality for parameter $vp0. Expected exactly one, got 0 [at line 1, column 6, source: String/267377182041421505]`

**Expected behavior**
I expect an empty-sequence

**To Reproduce**
```
let $test := <foo/>
return $test/bar => xs:integer()
```
produces the error where it should return empty sequence. In contrast,

```
let $test := <foo/>
return $test/bar ! xs:integer(.)
``` 
works as expected.

 - Build: [eXist-6.2.0]
 - Java: openjdk 11.0.17
 - OS: Windows 10, Linux

**Additional context**
none relevant (I think)

line-o:
### Description:

Type constructors like `xs:string` or `xs:integer` did wrongly check for EXACTLY_ONE item passed to them when used as a function. This happens for instance when used on the right hand side of an arrow operator.

```xquery
() => xs:string()
```

would fail with a cardinality error. The error message itself is hard to understand which should be addressed in a following PR.

This PR fixes the cast function to allow an empty sequence as its parameter.

### Reference:

fixes #4971 

### Type of tests:

XQSuite tests

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/CastExpression.java
+++ b/exist-core/src/main/java/org/exist/xquery/CastExpression.java
@@ -190,7 +190,7 @@ public class CastExpression extends AbstractExpression {
 	    try {
             final QName qname = QName.parse(context, typeName);
             final FunctionSignature signature = new FunctionSignature(qname);
-            final SequenceType argType = new SequenceType(Type.ITEM, Cardinality.ZERO_OR_ONE);
+            final SequenceType argType = new SequenceType(Type.ATOMIC, Cardinality.ZERO_OR_ONE);
             signature.setArgumentTypes(new SequenceType[]{argType});
             signature.setReturnType(new SequenceType(CastExpression.this.requiredType, CastExpression.this.cardinality));
             return new FunctionWrapper(this, signature);
--- a/exist-core/src/main/java/org/exist/xquery/CastExpression.java
+++ b/exist-core/src/main/java/org/exist/xquery/CastExpression.java
@@ -190,7 +190,7 @@ public class CastExpression extends AbstractExpression {
 	    try {
             final QName qname = QName.parse(context, typeName);
             final FunctionSignature signature = new FunctionSignature(qname);
-            final SequenceType argType = new SequenceType(Type.ITEM, Cardinality.EXACTLY_ONE);
+            final SequenceType argType = new SequenceType(Type.ITEM, Cardinality.ZERO_OR_ONE);
             signature.setArgumentTypes(new SequenceType[]{argType});
             signature.setReturnType(new SequenceType(CastExpression.this.requiredType, CastExpression.this.cardinality));
             return new FunctionWrapper(this, signature);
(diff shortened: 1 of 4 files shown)


dizzzz (approved):

adamretter:
Does this change now allow the user to incorrectly write just `xs:string()` in an XQuery?

reinhapa (approved):

line-o:
> Does this change now allow the user to incorrectly write just `xs:string()` in an XQuery?

No

adamretter:
@line-o Thanks. So it looks like the problem is not only related to Arrow Operators, but is actually a bad interpretation of the XQuery F+O 3.1 specification.

[18.1 Constructor functions for XML Schema built-in atomic types](https://www.w3.org/TR/xpath-functions-31/#constructor-functions-for-xsd-types) states that the argument type should be `xs:anyAtomicType?`.

So if you also change `Type.ITEM` in your PR to `Type.ATOMIC`, then your PR will nicely bring this inline with the XQuery specification.

exist-core/src/main/java/org/exist/xquery/CastExpression.java
```diff
@@ -190,7 +190,7 @@ public Function toFunction() throws XPathException {
 	    try {
             final QName qname = QName.parse(context, typeName);
             final FunctionSignature signature = new FunctionSignature(qname);
-            final SequenceType argType = new SequenceType(Type.ITEM, Cardinality.EXACTLY_ONE);
+            final SequenceType argType = new SequenceType(Type.ITEM, Cardinality.ZERO_OR_ONE);
```
adamretter, line 193: I think that the specification: [18.1 Constructor functions for XML Schema built-in atomic types](https://www.w3.org/TR/xpath-functions-31/#constructor-functions-for-xsd-types) says that this should be `Type.ATOMIC` and not `Type.ITEM`. Otherwise, LGTM.

adamretter (changes requested):

exist-core/src/main/java/org/exist/xquery/CastExpression.java
line-o: Both Saxon and basex allow `item()?` to be passed to cast constructor functions.
```
xs:integer([1]), 
xs:string(<foo>test</foo>)
``` 
are both valid

line-o: I asked for clarification on XML Slack https://xmlcom.slack.com/archives/C011NLXE4DU/p1688742869580379?thread_ts=1688552697.034799&cid=C011NLXE4DU

line-o: OK, so atomisation will be done before the argument is passed in. Will change the type.

line-o:
The XQTS workflow seems to be broken https://github.com/eXist-db/exist/actions/workflows/ci-xqts.yml

adamretter (approved):

PieterLamers:
Could this fix also be merged into the 6.x branch of develop?

line-o:
Yes, @PieterLamers, I will prepare a back port.

PR #4980, opened 2023-07-19
[bugfix] Correct function signatures that return empty sequences

joewiz:
### Description:

This PR corrects the return type of functions that return `empty-sequence()`. Before this PR, the function signatures of several functions were described as returning a sequence of type `item()` with a cardinality of "empty sequence":

```java
SequenceType(Type.ITEM, Cardinality.EMPTY_SEQUENCE)
```

With this PR, the function signatures are now updated to return a sequence of type `empty-sequence()` and an unchanged cardinality:

```java
SequenceType(Type.EMPTY, Cardinality.EMPTY_SEQUENCE)
```

### Reference:

I shared my observation about this issue in Slack on July 18, 2023, at https://exist-db.slack.com/archives/CG2MRUZ35/p1689734379824689. Referencing the two code snippets above, I wrote:

> I don’t see why any function that returns an empty sequence should use the former instead of the latter. Shall I prepare a PR replacing all instances of functions returning the former with the latter?

@adamretter replied:

> It is indeed a bug. The second form is correct.

### Type of tests:

No new tests added, but `mvn test` passes locally.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/securitymanager/GroupManagementFunction.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/securitymanager/GroupManagementFunction.java
@@ -51,7 +51,7 @@ public class GroupManagementFunction extends BasicFunction {
         new SequenceType[]{
             new FunctionParameterSequenceType("group-name", Type.STRING, Cardinality.EXACTLY_ONE, "The name of the group to create.")
         },
-        new SequenceType(Type.ITEM, Cardinality.EMPTY_SEQUENCE)
+        new SequenceType(Type.EMPTY, Cardinality.EMPTY_SEQUENCE)
     );
 
     public final static FunctionSignature FNS_CREATE_GROUP_WITH_METADATA = new FunctionSignature(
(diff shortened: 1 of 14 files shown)


line-o (approved):
You found quite a number of places with this declaration.
Is there one instance where it actually is declared correctly, I wonder.
Good work!

joewiz:
@line-o Thanks! Yes, it was correctly declared in 48 signatures in `exist-core` (e.g., [fn:error](https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/xquery/functions/fn/FunError.java#L56)). With this PR, we have 78 such functions that return empty sequence.

I just extended the search to the entire repo and found one more in `extensions/modules/mail`. I'll add that to the PR... so 79.

joewiz:
Screenshots from function-documentation illustrate the change...

**Before:**

> Screenshot 2023-07-19 at 12 12 19@2x

**After:** (using the just-released [function-documentation 1.2.1](https://github.com/eXist-db/function-documentation/releases/tag/v1.2.1) on develop)

> Screenshot 2023-07-19 at 12 12 46@2x

reinhapa (approved):

PR #4991, opened 2023-07-26
Log exception if the response is already committed

adamretter:
The exception was missing for a `BadRequestException` in this one location


dizzzz (approved):

dizzzz:
pulling in right now, no risks

Issue #1010, opened 2016-05-13
Recursion RestXQ: deploying web app fails 


peterstadler:
I just upgraded my eXist instance via git and did a `build.sh rebuild`.
All went fine, but now I’m having problems installing my [web app](https://github.com/Edirom/WeGA-WebApp/releases/download/v2.0.0/WeGA-WebApp-2.0.0.xar). The deployment process simply hangs, whether I upload via browser or try via autodeploy.
I updated from commit 72b4bc7d3112593df2d508ca563ad23a31189c3e (February 1st) and I can confirm that going back to this commit indeed fixes the issue for me.

My system:
OS X 10.11.4
Java(TM) SE Runtime Environment (build 1.8.0_91-b14)
eXist develop branch at 5d6fc97102e27f5404f22ddf3e90e41b4310ae02

@wolfgangmm confirmed on [exist-open](http://exist.2174344.n4.nabble.com/Problem-deploying-web-app-tp4670225p4670226.html):

> I can see an endless loop in the thread dump (the first lines are repeated over and over again):

```
at org.exist.extensions.exquery.restxq.impl.ExistXqueryRegistry.findServices(ExistXqueryRegistry.java:239)
at org.exist.extensions.exquery.restxq.impl.ExistXqueryRegistry.reexamineModulesWithResolvedDependencies(ExistXqueryRegistry.java:320)
at org.exist.extensions.exquery.restxq.impl.ExistXqueryRegistry.findServices(ExistXqueryRegistry.java:239)
at org.exist.extensions.exquery.restxq.impl.ExistXqueryRegistry.reexamineModulesWithResolvedDependencies(ExistXqueryRegistry.java:320)
at org.exist.extensions.exquery.restxq.impl.ExistXqueryRegistry.findServices(ExistXqueryRegistry.java:239)
at org.exist.extensions.exquery.restxq.impl.RestXqTrigger.after(RestXqTrigger.java:113)
at org.exist.extensions.exquery.restxq.impl.RestXqTrigger.afterCreateDocument(RestXqTrigger.java:51)
at org.exist.collections.triggers.DocumentTriggers.afterCreateDocument(DocumentTriggers.java:236)
at org.exist.collections.Collection.addBinaryResource(Collection.java:1955)
at org.exist.collections.Collection.addBinaryResource(Collection.java:1865)
at org.exist.collections.Collection.addBinaryResource(Collection.java:1858)
at org.exist.repo.Deployment.storeFiles(Deployment.java:771)
[...]

PR #4996, opened 2023-07-28
Fixes to eXist-db regarding circular imports

adamretter:
These errors although mostly in eXist-db itself were easy to manifest through using RESTXQ.

Closes https://github.com/eXist-db/exist/issues/3448
Closes https://github.com/eXist-db/exist/issues/1010
Closes https://github.com/eXist-db/exist/pull/4699

1. The first thing that this PR deals with is a Library Module that attempts to import itself. Whilst BaseX raises an error, and Saxon allows it. With this PR eXist-db now follows Saxon in basically skipping the superfluous import and continues. This is inline with the W3C XQuery 3.1 specification which states:

    > The namespace prefix specified in a module declaration must not be xml or xmlns [err:XQST0070], and must not be the same as any namespace prefix bound in the same module by a schema import, by a namespace declaration, or by a module import with a different target namespace [err:XQST0033]. [^1]

    and:

    > A module may import its own target namespace (this is interpreted as importing an implementation-defined set of other modules that share its target namespace.) [^2]

2. The second thing that this PR deals with is cyclic imports of Library Modules. i.e. `M1 -> M2 -> M1`. Before this PR an XQuery executed on eXist-db with a cyclic set of imports in its library modules would crash with a Java `StackOverflowError`. The expected behaviour has changed here between XQuery 1.0 and XQuery 3.1. In XQuery 1.0 cyclic imports were prohibited:

    > It is a static error [err:XQST0093] to import a module M1 if there exists a sequence of modules M1 ... Mi ... M1 such that each module directly depends on the next module in the sequence (informally, if M1 depends on itself through some chain of module dependencies.) [^3]

    however, in XQuery 3.1 cyclic imports are allowed:

    > Implementations must resolve cycles in the import graph, either at the level of target namespace URIs or at the level of location URIs, and ensure that each module is imported only once. [^4]

    In this PR, the XQuery version is now checked. If the XQuery is 1.0 then XQST0093 is raised, if the XQuery is newer (i.e. 3.0 or 3.1) then the cyclic import is resolved and allowed without issue.

[^1]: https://www.w3.org/TR/xquery-31/#id-module-declaration
[^2]: https://www.w3.org/TR/xquery-31/#dt-module-import
[^3]: https://www.w3.org/TR/2010/REC-xquery-20101214/#id-module-import
[^4]: https://www.w3.org/TR/xquery-31/#id-module-handling-cycles

----

### DIFF
--- a/exist-core/pom.xml
+++ b/exist-core/pom.xml
@@ -514,6 +514,17 @@
             <artifactId>ant</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.jgrapht</groupId>
+            <artifactId>jgrapht-core</artifactId>
+            <version>1.5.2</version>
+        </dependency>
+
+	<dependency>
+            <groupId>org.jgrapht</groupId>
+            <artifactId>jgrapht-opt</artifactId>
+            <version>1.5.2</version>
+        </dependency>
 
         <dependency>
             <groupId>junit</groupId>
--- a/exist-core/src/main/java/org/exist/interpreter/Context.java
+++ b/exist-core/src/main/java/org/exist/interpreter/Context.java
@@ -760,8 +760,6 @@ public interface Context {
      */
     boolean tailRecursiveCall(FunctionSignature signature);
 
-    void mapModule(String namespace, XmldbURI uri);
-
     /**
      * Import one or more library modules into the function signatures and in-scope variables of the importing module.
      *
@@ -783,7 +781,7 @@ public interface Context {
      *      XQST0070
      *      XQST0088
      */
-    Module[] importModule(@Nullable String namespaceURI, @Nullable String prefix, @Nullable AnyURIValue[] locationHints) throws XPathException;
+    @Nullable Module[] importModule(@Nullable String namespaceURI, @Nullable String prefix, @Nullable AnyURIValue[] locationHints) throws XPathException;
 
     /**
      * Returns the static location mapped to an XQuery source module, if known.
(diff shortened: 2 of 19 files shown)


reinhapa (approved):

joewiz:
@adamretter To aid folks reviewing the PR, could you assess the CI failures and whether they are due to the PR or issues specific to the CI environment?

adamretter:
@joewiz All resolved now.

duncdrum:
@adamretter looks like the license header checker doesn't like one of the license headers
https://github.com/eXist-db/exist/actions/runs/5751766777/job/15611509940?pr=4996#step:4:1871

```
Error:  Failed to execute goal com.mycila:license-maven-plugin:4.2:check (default-cli) on project exist-restxq: Some files do not have the expected license header -> [Help 1]
```

adamretter:
> looks like the license header checker doesn't like one of the license headers

@duncdrum I have fixed that now.

> Could you please explain the effect of these changes to https://github.com/eXist-db/exist/pull/4699

PR #4699 was never the correct approach and is now redundant due to the fixes in this PR. This PR does not break any backwards compatibility.

dizzzz (approved):

reinhapa:
@adamretter could you resolve the existing conflicts in order to merge your're PR?

adamretter:
I have just rebased this now. @dizzzz @reinhapa I believe this is now ready to merge please

PR #5008, opened 2023-08-14
[6.x.x] Allow the user to override the JDWP Suspend at Docker build time

adamretter:
Backport of https://github.com/eXist-db/exist/pull/5007

### DIFF
--- a/exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
+++ b/exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
@@ -46,10 +46,11 @@ LABEL org.label-schema.build-date=${maven.build.timestamp} \
 
 EXPOSE 8080 8443 5005
 
-# make CACHE_MEM, MAX_BROKER, and JVM_MAX_RAM_PERCENTAGE available to users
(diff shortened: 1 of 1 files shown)


exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
```diff
@@ -70,7 +71,7 @@ ENV JAVA_TOOL_OPTIONS \
   -XX:+UseContainerSupport \
   -XX:MaxRAMPercentage=${JVM_MAX_RAM_PERCENTAGE:-75.0} \
   -XX:+ExitOnOutOfMemoryError \
-  -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005
+  -agentlib:jdwp=transport=dt_socket,server=y,suspend=${JVM_JDWP_SUSPEND:-n},address=5005
```
reinhapa, line 74: Could it be that there is a unintended minus sign?

adamretter: @reinhapa It is intentional. It means if there is no value then use `n`

reinhapa (approved):

PR #5021, opened 2023-08-22
Rename Function flag overloaded to variadic

line-o:
### Description:

The _overloaded_ flag is set to `true` for functions like `fn:concat`, which has an unlimited arity.
The term does however have a different meaning and therefore lead to confusion when I touched code in #3389.
It is therefore better to call this flag variadic (see the references for details).

The getter function and constructor parameter names are renamed to reflect this.

`response:stream-binary` and `util:get-fragment-between` are both non-variadic but flagged as such. This flag is removed.

`fn:namespace-uri#0` and `fn:namespace-uri#1` were flagged as non-variadic, which is the default. Therefore, setting of the flag is removed.

**Note:** There are variadic functions in the `range` module (FieldLookup.java). Removing the flag for `range:field-eq` for example results in failing tests. I believe they are flagged as variadic because you are allowed to query **n** fields for **m** values.

Two failing tests can be found in src/test/xquery/range/range.xql (L197, L369)

### Reference:

Definition of variadic in the Wikipedia: https://en.wikipedia.org/wiki/Variadic_function

Usage of variadic in the XQuery Specification: 
> Although the base function referred to may be variadic, the result of evaluating the function reference is a function that has fixed arity.

### Type of tests:

None added

refs #3389 (discussion in thread)

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/Function.java
+++ b/exist-core/src/main/java/org/exist/xquery/Function.java
@@ -219,7 +219,7 @@ public abstract class Function extends PathExpr {
      * @throws XPathException if an error occurs setting the arguments
      */
     public void setArguments(final List<Expression> arguments) throws XPathException {
-        if ((!mySignature.isOverloaded()) && arguments.size() != mySignature.getArgumentCount()) {
+        if ((!mySignature.isVariadic()) && arguments.size() != mySignature.getArgumentCount()) {
             throw new XPathException(this, ErrorCodes.XPST0017,
                     "Number of arguments of function " + getName() + " doesn't match function signature (expected " +
                             mySignature.getArgumentCount() + ", got " + arguments.size() + ')');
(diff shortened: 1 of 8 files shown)


dizzzz (approved):

reinhapa (approved):

line-o:
An example of a variadic call to field-eq taken from range.xql testsuite (all tests are with only one value per field).
```xquery
collection("/db/rangetest")
  //range:field-eq(
    ("type", "subtype", "name"), 
    "main", "official", "Hofthiergarten"
  )/text()
```

line-o:
As expected the function allows to query for more than one value per field:
```xquery
declare
    %test:assertEquals("Hofthiergarten", "Dorfprozelten")
function rt:equality-field-nested-multi() {
    collection("/db/rangetest")
        //range:field-eq(
            ("type", "subtype", "name"),
            "main", "official", ("Hofthiergarten", "Dorfprozelten"))
        /text()
};
```

line-o:
a non-variadic version of range field lookups would therefore need to either:
- wrap value sequences in an array
- pass on a sequence of arrays of values

Both constitute breaking changes in my opinion, as those functions can also be called directly in application code.

reinhapa (approved):

reinhapa:
@line-o ready to merge?

line-o:
I say: yes it is.
Anything else can be added in additional PRs.

PR #5027, opened 2023-08-29
Upgrading apache's FOP

dizzzz:
Combining #5020 and #5019


line-o:
@dizz the content extraction module declares a dependency on FOP 2.9

PR #5037, opened 2023-09-04
fn:parse-xml-fragment#1 should return a document-node() when there is an XML Declaration

adamretter:
WIP - I need to add tests yet


line-o:
The PR - as is - does not present a sufficient solution. As I assume that it will not receive additional work I propose to close it without further action.
I will open a PR fixing this issue by the end of this week.

Issue #5028, opened 2023-08-29
[BUG] file:directory-list#2 does not properly report an error on a non-existent directory


PieterLamers:
**Describe the bug**
The Java code of `DirectoryList` throws an `IllegalStateException` when a non-existent directory is supplied as its first argument, which is an unchecked exception and therefore not caught in Java. From Oxygen, this results in a `Failed to invoke method queryPT in class org.exist.xmlrpc.RpcConnection:` message.

**Expected behavior**
We expect an XQuery error to be thrown with module, line number and column information.

**To Reproduce**
Assuming a non-existent path `/foo` run the following : 
`file:directory-list('/foo', '*')`

 - Build: [eXist-6.2.0]
 - Java: [1.8.0_347]
 - OS:  Windows 10

nverwer:
I think that this can be solved by adding the following in DirectoryList.java, line 175:

```
        } catch (final  IllegalStateException e) {
            throw new XPathException(this, e.getMessage());
```

PR #5061, opened 2023-09-15
fix for #5028

nverwer:
The Java code of DirectoryList throws an IllegalStateException when a non-existent directory is supplied as its first argument, which is an unchecked exception and therefore not caught in Java.


dizzzz (approved):

joewiz:
Thank you for this fix! One request: Could you change the title for this PR to be more descriptive, in keeping with our general practice. Something like:

> Prevent IllegalStateException in DirectoryList when a non-existent directory is supplied

This helps us distinguish between various PRs and helps with release notes, which are drawn from PR titles.

PR #5068, opened 2023-09-22
Parameterise the XML-RPC parse method with the Media Type

### LINKED ISSUE #5067: [BUG] Xmlrpc: XML documents without an .xml extension are stored as binary
When an XML document without an explicit `.xml` extension is stored via xmlrpc, its mime type is converted to binary.
Despite the fact that the provided resource type is `XMLResource.class`.

Here is a sketch for a test:
```
@Test
public void testXMLMimeType() throws XMLDBException {
    // store an XML document without an .xml extension
    try(Collection collection = DatabaseManager.getCollection(getBaseUri() + "/db/" + COLLECTION_NAME, ...)) {
        final Class<? extends Resource> xmlResourceType = XMLResource.class;
        final XMLResource resource = (XMLResource)collection.createResource(DOCUMENT_NAME, xmlResourceType);
        resource.setContent(XML_CONTENT);
        collection.storeResource(resource);
        assertEquals(ResourceType.XML_RESOURCE, resource.getResourceType());
    }
	
    // retrieve the document and verify its resource type
    try(Collection collection = DatabaseManager.getCollection(getBaseUri() + "/db/" + COLLECTION_NAME, ...)) {
        Resource resource = collection.getResource(DOCUMENT_NAME);
        assertEquals(ResourceType.XML_RESOURCE, resource.getResourceType()); // <== fails with binary
    }
}
```

The expected behaviour would be, that an XML document is stored as `XMLResource` if so requested.
I will try to provide a pull request with a complete test and a possible solution.

The fix is needed for `develop` and `develop-6.x.x`.
On `develop-4.x.x` the issue does not manifest itself.

ohumbel:
### Description:
Closes #5067.

### Reference:
Since XML resources are always and only stored via the `parse()` method, we need to pass the Media Type as a parameter to this method.

### Type of tests:
A unit test in `exist-core/src/test/java/org/exist/xmlrpc`: `MimeTypeTest`

I successfully ran the full maven reactor locally.

### DIFF
--- a/exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
+++ b/exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
@@ -1336,18 +1336,24 @@ public class RpcConnection implements RpcAPI {
 
     @Override
     public boolean parse(byte[] xml, String documentPath, int overwrite) throws URISyntaxException, EXistException, PermissionDeniedException {
-        return parse(xml, documentPath, overwrite, null, null);
+        return parse(xml, documentPath, null, overwrite, null, null);
+    }
+
+    @Override
+    public boolean parse(final byte[] xml, final String documentPath, final String mimeType, final int overwrite) throws URISyntaxException, EXistException, PermissionDeniedException {
+        return parse(xml, documentPath, mimeType, overwrite, null, null);
     }
 
     @Override
     public boolean parse(final byte[] xml, final String documentPath,
                          final int overwrite, final Date created, final Date modified) throws URISyntaxException, EXistException, PermissionDeniedException {
-        return parse(xml, XmldbURI.xmldbUriFor(documentPath), overwrite, created, modified);
+        return parse(xml, documentPath, null, overwrite, created, modified);
     }
 
-    private boolean parse(final byte[] xml, final XmldbURI docUri,
-                          final int overwrite, @Nullable final Date created, @Nullable final Date modified) throws EXistException, PermissionDeniedException {
-
+    @Override
+    public boolean parse(final byte[] xml, final String documentPath, @Nullable String mimeType,
+                         final int overwrite, @Nullable final Date created, @Nullable final Date modified) throws URISyntaxException, EXistException, PermissionDeniedException {
+        final XmldbURI docUri = XmldbURI.xmldbUriFor(documentPath);
         return this.<Boolean>writeCollection(docUri.removeLastSegment()).apply((collection, broker, transaction) -> {
 
             try(final ManagedDocumentLock lockedDocument = broker.getBrokerPool().getLockManager().acquireDocumentWriteLock(docUri)) {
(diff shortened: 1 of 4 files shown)


reinhapa (approved):

exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
```diff
@@ -1366,7 +1366,10 @@ private boolean parse(final byte[] xml, final XmldbURI docUri,
 
                 final long startTime = System.currentTimeMillis();
 
-                final MimeType mime = MimeTable.getInstance().getContentTypeFor(docUri.lastSegment());
+                MimeType mime = MimeTable.getInstance().getContentTypeFor(docUri.lastSegment());
```
adamretter, line 1369: I think it would be better to add an additional media type parameter to the XML-RPC API for parse (in a backward compatible fashion) so that the Media Type is sent correctly from the XML-DB RemoteCollection class.

adamretter (changes requested):

exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
reinhapa: @adamretter  I do not see the _backward_ part here as the content type seems to be added with 6.x.x and was not available prior. Either way those method will only invoked with XML content from the `RemoteCollection`..

adamretter: > I do not see the backward part here

By backward compatible, I mean that if we add a parameter then we effectively have an additional procedure available for RPC, so we must also keep the previous procedure as well. For example, at present there are 6 `parse` procedures available for RPC, we should add a 7th rather than remove/change an existing one.

> Either way those method will only invoked with XML content from the RemoteCollection

Sure, but by hardcoding the Media Type to just `application/xml` in the RPC procedure, we limit ourselves to just that. Instead the Media Type should be a procedure parameter that is passed in from the RPC client.

> where should the existing content type being taken on the RemoteCollection

It comes from the resource, i.e. `EXistResource#getMomeType()`.

ohumbel:
@reinhapa Thanks for the feature!

ohumbel:
Regarding backport to `6.x.x`: I think it is necessary, because:

1. We initially found and fixed the problem in `6.2.0`
2. The slightly modified `MimeTypeTest` also fails on `6.x.x` with:
`[ERROR]   MimeTypeTest.testXMLMimeType:68 expected:<[XML]Resource> but was:<[Binary]Resource>`

[MimeTypeTest.java.txt](https://github.com/eXist-db/exist/files/12705980/MimeTypeTest.java.txt)

The relevant parts are:
`XMLResource resource = (XMLResource)collection.createResource(DOCUMENT_NAME, XMLResource.RESOURCE_TYPE);`
and
`assertEquals(XMLResource.RESOURCE_TYPE, resource.getResourceType());`

reinhapa:
@adamretter could you do review my enhancement adding the mime type again?

adamretter (approved):
LGTM!

adamretter:
I think we are just awaiting a backport to 6.x.x? Are backports to 5.x.x and 4.x.x also needed?

ohumbel:
> @ohumbel I think we are just awaiting a backport to 6.x.x? Are backports to 5.x.x and 4.x.x also needed?

@adamretter I know for sure that backport to 6.x.x is necessary - I'll try to take care of that as soon as the develop one is merged. For 5.x.x I need to check first.

dizzzz (approved):

ohumbel:
> Actually it's better to provide the backports before we merge please.... Otherwise they tend to get mixed up or forgotten!

Both PR's (#5070 and #5071) are now created.

PR #5077, opened 2023-10-04
[6.x.x] New exist:time XQuery Pragma

adamretter:
Backported from FusionDB - A new XQuery Pragma called `exist:time` that is backwards compatible with the previous limited `exist:timer` pragma.

Like its predecessor `exist:timer` it can be used for timing an XQuery expression. However, its main advantage is that it can also be used to log multiple iterations of the same expression using the parameter `measurement-mode=multiple`. In addition to that the new `exist:time` pragma also adds:

1. Nanosecond precision (useful for measuring very fast expressions or multiple iterations thereof)
2. A humane format output of the total time elapsed.
3. Automatically adapts output between `ns` and `ms` units depending on the magnitude of the measurement
4. A `logging-level` parameter so that you may set the log-level dynamically on invocation.
5. A`log-message-prefix` so that you may specify a prefix to be appended to the log message; this can make it much easier for you when trying to locate specific log messages.

## Example 1 - Single measurement
```xquery
(# exist:time logging-level=INFO #) {
    doc("/db/hindawi-example.xml")//p
}
```
produces log output like:
```
Elapsed: 21 ms.
```

## Example 2 - Single iteration measurements inside FLWOR
```xquery
for $i in (1 to 2)
return
  (# exist:time logging-level=INFO #) {
      $i * 10
  }
```
produces log output like:
```
Elapsed: 35967 ns.
Elapsed: 18028 ns.
```

## Example 3 - Multiple iteration measurement inside FLWOR

This new facility is extremely useful for helping diagnose performance issues when you have a loop that is slow, but the expressions inside the loop itself appear fast. In such a scenario you can use `exist:time` to measure the performance of one or more expressions within the loop.

```xquery
for $i in (1 to 2)
return
  (# exist:time logging-level=INFO measurement-mode=multiple #) {
      $i * 10
  }
```
produces log output like:
```
Elapsed: 53995 ns (53995 ns) [iterations=2 first=35967 ns, min=18028 ns, avg=26997.50 ns, max=35967 ns, last=18028 ns].
```

### DIFF
--- a/exist-core/pom.xml
+++ b/exist-core/pom.xml
@@ -712,6 +712,7 @@
                                 <exclude>src/test/java/org/exist/xquery/functions/session/AttributeTest.java</exclude>
                                 <exclude>src/test/java/org/exist/xquery/functions/xmldb/XMLDBAuthenticateTest.java</exclude>
                                 <exclude>src/main/java/org/exist/xquery/functions/util/Eval.java</exclude>
+                                <exclude>src/main/java/org/exist/xquery/pragmas/TimePragma.java</exclude>
                                 <exclude>src/test/java/org/exist/xquery/util/URIUtilsTest.java</exclude>
                                 <exclude>src/main/java/org/exist/xquery/value/ArrayListValueSequence.java</exclude>
                                 <exclude>src/test/java/org/exist/xquery/value/BifurcanMapTest.java</exclude>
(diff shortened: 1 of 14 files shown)


dizzzz (approved):
builds are not clean, rebase needed?

adamretter:
@dizzzz The build errors are unrelated, they are from downloading the NVE CVE database.

reinhapa (approved):

PR #5124, opened 2023-11-07
[5.x.x] Backport build scripts

reinhapa:
Backport `develop` Github actions


adamretter:
We should probably exclude exist-xqts from this, as it isn't really designed still to work with eXist-db 5.x.x

reinhapa:
> We should probably exclude exist-xqts from this, as it isn't really designed still to work with eXist-db 5.x.x

Ok, that's fine for me. I will tidy up the commits accordingly..

adamretter:
> Do you have any ideas about the failing dependency check though?

I am afraid not, my only observation is that it is trying to download it rather than use the built artefact from the local repo

line-o:
Even though that there still seems to be an issue with dependency checks I would want to merge this PR as it brings back test execution to 5.x.x

line-o (approved):
Dependency checks should be addressed but can be done in a separate PR

line-o:
@reinhapa you made it! All green. Good work.

adamretter (approved):
LGTM

PR #5147, opened 2023-11-30
[refactor] Some automatic refactoring using IntelliJ

dizzzz:
Some automatic refactoring using IntelliJ: use java 7/8/14/16 language features.

Grouped per commit.


reinhapa:
> @reinhapa yeah I spotted them, I will check if they were not already there.

Seems only one class being the issue, that you may can change with low effort...

dizzzz:
there are 3 classes, all previously risking the NPE :-(

the largest I could solve by simply not returning the null value, but the already defined QName.EMPTY_QNAME

```
@Override
    public final QName getQName() {
        return switch (getNodeType()) {
            case Node.ATTRIBUTE_NODE -> document.attrName[nodeNumber];
            case Node.ELEMENT_NODE, Node.PROCESSING_INSTRUCTION_NODE -> document.nodeName[nodeNumber];
            case NodeImpl.NAMESPACE_NODE -> document.namespaceCode[nodeNumber];
            case Node.DOCUMENT_NODE -> QName.EMPTY_QNAME;
            case Node.COMMENT_NODE -> QName.EMPTY_QNAME;
            case Node.TEXT_NODE -> QName.EMPTY_QNAME;
            case Node.CDATA_SECTION_NODE -> QName.EMPTY_QNAME;
            default -> null;
        };
    }
```

looks not to be of a problem right here, but it might be downstream.

for the other two... few... @adamretter @wolfgangmm  ?

adamretter:
> for the other two... pfew... @adamretter 

@dizzzz I see 5 listed. Which of those are the one or two you would like me to take a look at?

dizzzz:
so there are 3 identical situations in NodeImpl.java, where we could just return the 'null' Qname
for the other 2.... no idea.

exist-core/src/main/java/org/exist/dom/memtree/NodeImpl.java
```diff
@@ -151,21 +139,21 @@ public String getLocalName() {
...
                     return nsUri;
                 }
-
-            case NodeImpl.NAMESPACE_NODE:
+            }
+            case NodeImpl.NAMESPACE_NODE -> {
                 return XMLConstants.XMLNS_ATTRIBUTE_NS_URI;
-
-            default:
+            }
```
reinhapa, line 153: `case NodeImpl.NAMESPACE_NODE -> XMLConstants.XMLNS_ATTRIBUTE_NS_URI;`

exist-core/src/main/java/org/exist/dom/memtree/NodeImpl.java
```diff
@@ -151,21 +139,21 @@ public String getLocalName() {
...
+            default -> {
                 return null;
+            }
```
reinhapa, line 156: `default -> null;`

reinhapa (changes requested):
Some small parts that can be simplified even more...

reinhapa (approved):

reinhapa (approved):

line-o (approved):
I like the changes :tada:

reinhapa (commented):
@dizzzz getting better every day 💪

PR #5158, opened 2023-12-11
[bugfix] guard against NPE in securitymanager

### LINKED ISSUE #4670: [BUG] NPE in org.exist.xquery.functions.securitymanager.IdFunction

line-o:
### Reference:

supersedes #4671
fixes #4670


dizzzz (approved):

reinhapa (approved):

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
```diff
@@ -77,13 +77,18 @@ private org.exist.dom.memtree.DocumentImpl functionId() {
...
+            if (realUser != null) {
+                builder.startElement(new QName("real", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
+                subjectToXml(builder, realUser);
+                builder.endElement();
+            }
 
-            if (!sameUserWithSameGroups(context.getRealUser(), context.getEffectiveUser())) {
+            final Subject effectiveUser = context.getEffectiveUser();
+            if (effectiveUser != null && (
+                    realUser == null || !sameUserWithSameGroups(realUser, effectiveUser))) {
```
adamretter, line 89: `realUser` will never be `null`

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
```diff
@@ -77,13 +77,18 @@ private org.exist.dom.memtree.DocumentImpl functionId() {
...
+            final Subject realUser = context.getRealUser();
+            if (realUser != null) {
+                builder.startElement(new QName("real", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
+                subjectToXml(builder, realUser);
+                builder.endElement();
+            }
 
-            if (!sameUserWithSameGroups(context.getRealUser(), context.getEffectiveUser())) {
+            final Subject effectiveUser = context.getEffectiveUser();
+            if (effectiveUser != null && (
```
adamretter, line 88: It would be much cleaner to move the null check into the function `sameUserWithSameGroups`

adamretter (changes requested):

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
```diff
@@ -77,13 +77,18 @@ private org.exist.dom.memtree.DocumentImpl functionId() {
 
             builder.startElement(new QName("id", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
 
-            builder.startElement(new QName("real", SecurityManagerModule.NAMESPACE_URI, SecurityManagerModule.PREFIX), null);
-            subjectToXml(builder, context.getRealUser());
-            builder.endElement();
+            final Subject realUser = context.getRealUser();
+            if (realUser != null) {
```
adamretter, line 81: realUser will never be null

line-o: see comment above

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
adamretter: If context.realUser returns null - then that indicates a serious bug elsewhere that needs to be solved. Changing the code here will only hide that bug and make it much harder to find!

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
dizzzz: throw XPathException in this case?

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
dizzzz: throw XPathexception instead ?

adamretter: I would consider actually throwing a RuntimeException and shutting down - as at this point the security of the system is completely compromised

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
line-o: done

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
line-o: I'll try that and see if tests still pass

exist-core/src/main/java/org/exist/xquery/functions/securitymanager/IdFunction.java
dizzzz: A null value should not happen but... defensive programming in a complex application is not a too bad idea I'd say.

A log to detect the situation/throw exception ... both would be OK for me.

line-o:
@duncdrum this PR can be pulled in

PR #5186, opened 2024-01-10
[6.x.x] Allow setting the namespace on the result of sql:execute

nverwer:
### Description:

This change introduces two new signatures for the `sql:execute` function:

```
sql:execute($connection-handle as xs:long, $sql-statement as xs:string, $make-node-from-column-name as xs:boolean, $ns-prefix as xs:string, $ns-uri as xs:string) as element()?
sql:execute($connection-handle as xs:long, $statement-handle as xs:long, $parameters as element()?, $make-node-from-column-name as xs:boolean, $ns-prefix as xs:string, $ns-uri as xs:string) as element()?
```

These functions allow setting a namespace on the output of `sql:execute` that is different from the default.
This may be useful in contexts where namespace processing is unwanted, or different namespaces for different result sets are wanted.

### DIFF
--- a/extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/ExecuteFunction.java
+++ b/extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/ExecuteFunction.java
@@ -21,24 +21,17 @@
  */
 package org.exist.xquery.modules.sql;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-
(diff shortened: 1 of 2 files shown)


line-o:
Thank you @nverwer for your contribution. I would like to see the reorganisation of imports to be in a separate commit and a few tests (either jUnit or XQSuite) as well as a tiny bit of documentation explaining how to use that feature.
Also the commit message needs to be prepended with `[enhancement]`.

dizzzz (approved):

PR #5188, opened 2024-01-10
Adds in memory cache for RPC query results

reinhapa:
Enables in-memory query result values less than 4 MB instead of writing them to a temporary file.

By using the in-memory-buffer no potential virus scanner is trigger either on the server as on the client side. This almost doubled the performance on our productive systems, when using this change.

### DIFF
--- a/exist-core/src/main/java/org/exist/util/io/VirtualTempPath.java
+++ b/exist-core/src/main/java/org/exist/util/io/VirtualTempPath.java
@@ -52,6 +52,10 @@ public final class VirtualTempPath implements ContentFile {
     @GuardedBy("lock")
     private Path contentFile;
 
+    public VirtualTempPath(TemporaryFileManager tempFileManager) {
+        this(DEFAULT_IN_MEMORY_SIZE, tempFileManager);
+    }
+
     public VirtualTempPath(int inMemorySize, TemporaryFileManager tempFileManager) {
         this.inMemorySize = inMemorySize;
         this.lock = new StampedLock();
@@ -97,6 +101,7 @@ public final class VirtualTempPath implements ContentFile {
         }
     }
 
+    @Override
     public InputStream newInputStream() throws IOException {
         long stamp = lock.readLock();
         try {
@@ -106,7 +111,7 @@ public final class VirtualTempPath implements ContentFile {
             if (content != null) {
                 return new MemoryContentsInputStream(content);
             }
-            return new ByteArrayInputStream(EMPTY_BUFFER);
+            return InputStream.nullInputStream();
         } finally {
             lock.unlockRead(stamp);
         }
@@ -129,6 +134,7 @@ public final class VirtualTempPath implements ContentFile {
         }
     }
 
+    @Override
     public long size() {
         long stamp = lock.readLock();
         try {
@@ -141,6 +147,7 @@ public final class VirtualTempPath implements ContentFile {
         }
     }
 
+    @Override
     public byte[] getBytes() {
         long stamp = lock.readLock();
         try {
--- a/exist-core/src/main/java/org/exist/xmlrpc/RpcAPI.java
+++ b/exist-core/src/main/java/org/exist/xmlrpc/RpcAPI.java
@@ -78,7 +78,7 @@ public interface RpcAPI {
      * @throws PermissionDeniedException If the current user does not have the permission to shut down the database
      * @deprecated Use {@link org.exist.xmlrpc.RpcAPI#shutdown(long)}
      */
-    @Deprecated
+    @Deprecated(since = "7.0")
     boolean shutdown(String delay) throws PermissionDeniedException;
 
     /**
@@ -388,7 +388,7 @@ public interface RpcAPI {
      * @throws EXistException If an internal error occurs
      * @throws PermissionDeniedException If the current user is not allowed to perform this action
      */
-    @Deprecated
+    @Deprecated(since = "7.0")
     Map<String, Object> queryP(byte[] xpath, Map<String, Object> parameters)
             throws EXistException, PermissionDeniedException;
 
@@ -404,7 +404,7 @@ public interface RpcAPI {
      * @throws PermissionDeniedException If the current user is not allowed to perform this action
      * @throws URISyntaxException If the URI contains syntax errors
      */
-    @Deprecated
+    @Deprecated(since = "7.0")
     Map<String, Object> queryP(byte[] xpath, String docName, String s_id, Map<String, Object> parameters)
             throws EXistException, PermissionDeniedException, URISyntaxException;
 
@@ -453,7 +453,7 @@ public interface RpcAPI {
      *
      * @deprecated use {@link #queryPT(byte[], Map)} or int {@link #executeQuery(byte[], Map)} instead.
      */
-    @Deprecated
+    @Deprecated(since = "7.0")
     byte[] query(
             byte[] xquery,
             int howmany,
@@ -496,7 +496,7 @@ public interface RpcAPI {
      * @throws PermissionDeniedException If the current user is not allowed to perform this action
      * @deprecated use List query() or int executeQuery() instead
      */
-    @Deprecated
+    @Deprecated(since = "7.0")
     Map<String, Object> querySummary(String xquery)
             throws EXistException, PermissionDeniedException;
 
@@ -729,7 +729,7 @@ public interface RpcAPI {
      *
      * @deprecated Use {@link #executeT(String, Map)} instead.
      */
-    @Deprecated
+    @Deprecated(since = "7.0")
     Map<String, Object> execute(String path, Map<String, Object> parameters)
             throws EXistException, PermissionDeniedException;
 
@@ -749,7 +749,7 @@ public interface RpcAPI {
      *
      * @deprecated Use {@link #executeT(String, Map)} instead.
      */
-    @Deprecated
+    @Deprecated(since = "7.0")
     Map<String, Object> executeT(String path, Map<String, Object> parameters)
             throws EXistException, PermissionDeniedException;
 
--- a/exist-core/src/main/java/org/exist/xmlrpc/RpcAPI.java
+++ b/exist-core/src/main/java/org/exist/xmlrpc/RpcAPI.java
@@ -31,7 +31,6 @@ import org.exist.EXistException;
 import org.exist.security.PermissionDeniedException;
 import org.exist.security.internal.aider.ACEAider;
 import org.exist.util.LockException;
-import org.exist.xmldb.XmldbURI;
 import org.exist.xquery.XPathException;
 import org.xml.sax.SAXException;
 
@@ -112,7 +111,7 @@ public interface RpcAPI {
     /**
      * Retrieve document by name. XML content is indented if prettyPrint is set
      * to &gt;=0. Use supplied encoding for output.
-     *
+     * <p>
      * This method is provided to retrieve a document with encodings other than
      * UTF-8. Since the data is handled as binary data, character encodings are
      * preserved. byte[]-values are automatically BASE64-encoded by the XMLRPC
@@ -132,7 +131,7 @@ public interface RpcAPI {
      * Retrieve document by name. XML content is indented if prettyPrint is set
      * to &gt;=0. Use supplied encoding for output and apply the specified
      * stylesheet.
-     *
+     * <p>
      * This method is provided to retrieve a document with encodings other than
      * UTF-8. Since the data is handled as binary data, character encodings are
      * preserved. byte[]-values are automatically BASE64-encoded by the XMLRPC
@@ -152,7 +151,7 @@ public interface RpcAPI {
     /**
      * Retrieve document by name. All optional output parameters are passed as
      * key/value pairs int the <code>parameters</code>.
-     *
+     * <p>
      * Valid keys may either be taken from
      * {@link javax.xml.transform.OutputKeys} or
      * {@link org.exist.storage.serializers.EXistOutputKeys}. For example, the
@@ -247,10 +246,10 @@ public interface RpcAPI {
     List<String> getDocumentListing(String collection)
             throws EXistException, PermissionDeniedException, URISyntaxException;
 
-    Map<String, List> listDocumentPermissions(String name)
+    Map<String, List<Object>> listDocumentPermissions(String name)
             throws EXistException, PermissionDeniedException, URISyntaxException;
 
-    Map<XmldbURI, List> listCollectionPermissions(String name)
+    Map<String, List<Object>> listCollectionPermissions(String name)
             throws EXistException, PermissionDeniedException, URISyntaxException;
 
     /**
@@ -497,6 +496,7 @@ public interface RpcAPI {
      * @throws PermissionDeniedException If the current user is not allowed to perform this action
      * @deprecated use List query() or int executeQuery() instead
      */
+    @Deprecated
     Map<String, Object> querySummary(String xquery)
             throws EXistException, PermissionDeniedException;
 
@@ -612,7 +612,7 @@ public interface RpcAPI {
 
     /**
      * Parse a file previously uploaded with upload.
-     *
+     * <p>
      * The temporary file will be removed.
      *
      * @param localFile temporary file name
@@ -749,6 +749,7 @@ public interface RpcAPI {
      *
      * @deprecated Use {@link #executeT(String, Map)} instead.
      */
+    @Deprecated
     Map<String, Object> executeT(String path, Map<String, Object> parameters)
             throws EXistException, PermissionDeniedException;
 
@@ -926,7 +927,7 @@ public interface RpcAPI {
 
     List<String> getGroups() throws EXistException, PermissionDeniedException;
 
-    List<List> getIndexedElements(String collectionName, boolean inclusive)
+    List<List<Object>> getIndexedElements(String collectionName, boolean inclusive)
             throws EXistException, PermissionDeniedException, URISyntaxException;
 
     boolean releaseQueryResult(int handle);
--- a/exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
+++ b/exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
@@ -72,7 +72,9 @@ import org.exist.storage.txn.Txn;
 import org.exist.util.*;
 import org.exist.util.crypto.digest.DigestType;
 import org.exist.util.crypto.digest.MessageDigest;
+import org.exist.util.io.ContentFile;
 import org.exist.util.io.TemporaryFileManager;
+import org.exist.util.io.VirtualTempPath;
 import org.exist.util.serializer.SAXSerializer;
 import org.exist.util.serializer.SerializerPool;
 import org.exist.validation.ValidationReport;
@@ -163,6 +165,14 @@ public class RpcConnection implements RpcAPI {
         }
     }
 
+    private TemporaryFileManager temporaryFileManager() {
+        return TemporaryFileManager.getInstance();
+    }
+
+    private VirtualTempPath createVirtualTempPath() {
+        return new VirtualTempPath(temporaryFileManager());
+    }
+
     private boolean createCollection(final XmldbURI collUri, final Date created) throws PermissionDeniedException, EXistException {
         withDb((broker, transaction) -> {
             Collection current = broker.getCollection(collUri);
@@ -216,22 +226,11 @@ public class RpcConnection implements RpcAPI {
 
     private String createId(final XmldbURI collUri) throws EXistException, PermissionDeniedException {
         return this.<String>readCollection(collUri).apply((collection, broker, transaction) -> {
-            XmldbURI id;
             final Random rand = new Random();
-            boolean ok;
+            XmldbURI id;
             do {
-                ok = true;
                 id = XmldbURI.create(Integer.toHexString(rand.nextInt()) + ".xml");
-                // check if this id does already exist
-                if (collection.hasDocument(broker, id)) {
-                    ok = false;
-                }
-
-                if (collection.hasChildCollection(broker, id)) {
-                    ok = false;
-                }
-
-            } while (!ok);
+            } while (collection.hasDocument(broker, id) || collection.hasChildCollection(broker, id));
             return id.toString();
         });
     }
@@ -320,7 +319,7 @@ public class RpcConnection implements RpcAPI {
             context.setModuleLoadPath(moduleLoadPath);
         }
         final Map<String, String> namespaces = (Map<String, String>) parameters.get(RpcAPI.NAMESPACES);
-        if (namespaces != null && namespaces.size() > 0) {
+        if (namespaces != null && !namespaces.isEmpty()) {
             context.declareNamespaces(namespaces);
         }
         //  declare static variables
@@ -406,7 +405,7 @@ public class RpcConnection implements RpcAPI {
     public int executeQuery(final byte[] xpath, final String encoding, final Map<String, Object> parameters) throws EXistException, PermissionDeniedException {
         final Charset charset = Optional.ofNullable(encoding).map(Charset::forName).orElse(DEFAULT_ENCODING);
         final String xpathString = new String(xpath, charset);
-        if(LOG.isDebugEnabled()) {
+        if (LOG.isDebugEnabled()) {
             LOG.debug("query: {}", xpathString);
         }
         return executeQuery(xpathString, parameters);
@@ -428,15 +427,6 @@ public class RpcConnection implements RpcAPI {
                 throw new EXistException(e);
             }
         });
-    }
-
-    protected String formatErrorMsg(final String message) {
-        return formatErrorMsg("error", message);
-    }
-
-    protected String formatErrorMsg(final String type, final String message) {
-        return ("<exist:result xmlns:exist=\"" + Namespaces.EXIST_NS + "\" ") + "hitCount=\"0\">" +
-                '<' + type + '>' + message + "</" + type + "></exist:result>";
     }
 
     @Override
@@ -565,10 +555,10 @@ public class RpcConnection implements RpcAPI {
 
     /**
      * The method <code>describeCollection</code>
-     *
+     * <p>
      * Returns details of a collection - collections (list of sub-collections) -
      * name - created - owner - group - permissions - acl
-     *
+     * <p>
      * If you do not have read access on the collection, the list of
      * sub-collections will be empty, an exception will not be thrown!
      *
@@ -715,15 +705,15 @@ public class RpcConnection implements RpcAPI {
 
             // A tweak for very large resources, VirtualTempFile
             final Map<String, Object> result = new HashMap<>();
-            final TemporaryFileManager temporaryFileManager = TemporaryFileManager.getInstance();
-            final Path tempFile = temporaryFileManager.getTemporaryFile();
+            final VirtualTempPath tempFile = createVirtualTempPath();
 
             if (document.getResourceType() == DocumentImpl.XML_FILE) {
-                try (final Writer writer = Files.newBufferedWriter(tempFile, encoding)) {
+                try (final OutputStream out = tempFile.newOutputStream();
+                     final BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(out, encoding))) {
                     serialize(broker, toProperties(parameters), saxSerializer -> saxSerializer.toSAX(document), writer);
                 }
             } else {
-                try (final OutputStream os = new BufferedOutputStream(Files.newOutputStream(tempFile))) {
+                try (final OutputStream os = new BufferedOutputStream(tempFile.newOutputStream())) {
                     broker.readBinaryResource(transaction, (BinaryDocument) document, os);
                 }
             }
@@ -732,14 +722,14 @@ public class RpcConnection implements RpcAPI {
 
             result.put("data", firstChunk);
             int offset = 0;
-            if (Files.size(tempFile) > MAX_DOWNLOAD_CHUNK_SIZE) {
+            if (tempFile.size() > MAX_DOWNLOAD_CHUNK_SIZE) {
                 offset = firstChunk.length;
 
-                final int handle = factory.resultSets.add(new SerializedResult(tempFile));
+                final int handle = factory.resultSets.add(new CachedContentFile(tempFile));
                 result.put("handle", Integer.toString(handle));
                 result.put("supports-long-offset", Boolean.TRUE);
             } else {
-                temporaryFileManager.returnTemporaryFile(tempFile);
+                tempFile.close();
             }
             result.put("offset", offset);
 
@@ -747,16 +737,19 @@ public class RpcConnection implements RpcAPI {
         });
     }
 
-    private byte[] getChunk(final Path file, final int offset) throws IOException {
-        final long available = Files.size(file);
+    private byte[] getChunk(final ContentFile file, final int offset) throws IOException {
+        final long available = file.size();
         final int len = (int)Math.min(Math.min(available - offset, MAX_DOWNLOAD_CHUNK_SIZE), Integer.MAX_VALUE);
 
         final byte[] chunk = new byte[len];
-        try (final InputStream is = Files.newInputStream(file)) {
-            is.skip(offset);
+        try (final InputStream is = file.newInputStream()) {
+            long remainingSkipped = offset;
+            do {
+                remainingSkipped -=  is.skip(remainingSkipped);
+            } while (remainingSkipped > 0);
             final int read = is.read(chunk);
             if(read != len) {
-                throw new IOException("Unable to read full chunk at offset: " + offset + ", from file: " + file.toAbsolutePath().toString());
+                throw new IOException("Unable to read full chunk at offset: " + offset + ", from file: " + file);
             }
         }
         return chunk;
@@ -767,16 +760,16 @@ public class RpcConnection implements RpcAPI {
             throws EXistException, PermissionDeniedException {
         try {
             final int resultId = Integer.parseInt(handle);
-            final SerializedResult sr = factory.resultSets.getSerializedResult(resultId);
+            final CachedContentFile sr = factory.resultSets.getCachedContentFile(resultId);
 
             if (sr == null) {
                 throw new EXistException("Invalid handle specified");
             }
             // This will keep the serialized result in the cache
             sr.touch();
-            final Path tempFile = sr.result;
+            final ContentFile tempFile = sr.getResult();
 
-            if (offset <= 0 || offset > Files.size(tempFile)) {
+            if (offset <= 0 || offset > tempFile.size()) {
                 factory.resultSets.remove(resultId);
                 throw new EXistException("No more data available");
             }
@@ -786,7 +779,7 @@ public class RpcConnection implements RpcAPI {
             final Map<String, Object> result = new HashMap<>();
             result.put("data", chunk);
             result.put("handle", handle);
-            if (nextChunk > (long) Integer.MAX_VALUE || nextChunk >= Files.size(tempFile)) {
+            if (nextChunk > Integer.MAX_VALUE || nextChunk >= tempFile.size()) {
                 factory.resultSets.remove(resultId);
                 result.put("offset", 0);
             } else {
@@ -803,17 +796,17 @@ public class RpcConnection implements RpcAPI {
             throws EXistException, PermissionDeniedException {
         try {
             final int resultId = Integer.parseInt(handle);
-            final SerializedResult sr = factory.resultSets.getSerializedResult(resultId);
+            final CachedContentFile sr = factory.resultSets.getCachedContentFile(resultId);
 
             if (sr == null) {
                 throw new EXistException("Invalid handle specified");
             }
             // This will keep the serialized result in the cache
             sr.touch();
-            final Path tempFile = sr.result;
+            final ContentFile tempFile = sr.getResult();
 
             final long longOffset = Long.parseLong(offset);
-            if (longOffset < 0 || longOffset > Files.size(tempFile)) {
+            if (longOffset < 0 || longOffset > tempFile.size()) {
                 factory.resultSets.remove(resultId);
                 throw new EXistException("No more data available");
             }
@@ -823,7 +816,7 @@ public class RpcConnection implements RpcAPI {
             final Map<String, Object> result = new HashMap<>();
             result.put("data", chunk);
             result.put("handle", handle);
-            if (nextChunk >= Files.size(tempFile)) {
+            if (nextChunk >= tempFile.size()) {
                 factory.resultSets.remove(resultId);
                 result.put("offset", Long.toString(0));
             } else {
@@ -879,10 +872,10 @@ public class RpcConnection implements RpcAPI {
 
     private int xupdate(final XmldbURI collUri, final String xupdate) throws PermissionDeniedException, EXistException {
         return withDb((broker, transaction) -> {
-            final Collection collectionRef = this.<Collection>readCollection(collUri).apply((collection, broker1, transaction1) -> collection);
-            //TODO : register a lock (which one ?) in the transaction ?
-            final DocumentSet docs = collectionRef.allDocs(broker, new DefaultDocumentSet(), true);
-            try(final Reader reader = new StringReader(xupdate)) {
+            try (final Collection collectionRef = this.<Collection>readCollection(collUri).apply((collection, broker1, transaction1) -> collection);
+                 final Reader reader = new StringReader(xupdate)) {
+                //TODO : register a lock (which one ?) in the transaction ?
+                final DocumentSet  docs = collectionRef.allDocs(broker, new DefaultDocumentSet(), true);
                 final XUpdateProcessor processor = new XUpdateProcessor(broker, docs);
                 final Modification modifications[] = processor.parse(new InputSource(reader));
                 long mods = 0;
@@ -1025,7 +1018,7 @@ public class RpcConnection implements RpcAPI {
     /**
      * Creates a unique name for a database resource Uniqueness is only
      * guaranteed within the eXist instance
-     *
+     * <p>
      * The name is based on a hex encoded string of a random integer and will
      * have the format xxxxxxxx.xml where x is in the range 0 to 9 and a to f
      *
@@ -1134,14 +1127,14 @@ public class RpcConnection implements RpcAPI {
     }
 
     @Override
-    public Map<String, List> listDocumentPermissions(final String name)
+    public Map<String, List<Object>> listDocumentPermissions(final String name)
             throws EXistException, PermissionDeniedException, URISyntaxException {
         return listDocumentPermissions(XmldbURI.xmldbUriFor(name));
     }
 
-    private Map<String, List> listDocumentPermissions(final XmldbURI collUri) throws EXistException, PermissionDeniedException {
-        return this.<Map<String, List>>readCollection(collUri).apply((collection, broker, transaction) -> {
-            final Map<String, List> result = new HashMap<>(collection.getDocumentCount(broker));
+    private Map<String, List<Object>> listDocumentPermissions(final XmldbURI collUri) throws EXistException, PermissionDeniedException {
+        return this.<Map<String, List<Object>>>readCollection(collUri).apply((collection, broker, transaction) -> {
+            final Map<String, List<Object>> result = new HashMap<>(collection.getDocumentCount(broker));
             for (final Iterator<DocumentImpl> i = collection.iterator(broker); i.hasNext(); ) {
                 final DocumentImpl doc = i.next();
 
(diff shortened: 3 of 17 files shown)


exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
```diff
@@ -732,31 +722,34 @@ public Map<String, Object> getDocumentData(final String docName, final Map<Strin
...
-    private byte[] getChunk(final Path file, final int offset) throws IOException {
-        final long available = Files.size(file);
+    private byte[] getChunk(final ContentFile file, final int offset) throws IOException {
+        final long available = file.size();
         final int len = (int)Math.min(Math.min(available - offset, MAX_DOWNLOAD_CHUNK_SIZE), Integer.MAX_VALUE);
 
         final byte[] chunk = new byte[len];
-        try (final InputStream is = Files.newInputStream(file)) {
-            is.skip(offset);
+        try (final InputStream is = file.newInputStream()) {
```
adamretter, line 745: Should we buffer the I/O here perhaps?

exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
```diff
@@ -1155,21 +1148,21 @@ private Map<String, List> listDocumentPermissions(final XmldbURI collUri) throws
     }
 
     @Override
-    public Map<XmldbURI, List> listCollectionPermissions(final String name)
+    public Map<String, List<Object>> listCollectionPermissions(final String name)
```
adamretter, line 1151: I am interested to know the thinking behind why this should be changed from `XmldbURI` to `String` please?

adamretter (commented):
This sounds interesting and like you are seeing a good performance improvement. If I understand correctly, when the result of an RPC call is less than 4 MB in size, then instead of writing to a file, you keep it in memory for a finite period of time. Do I have that correct?

If that is the case, then in general that sounds fine, but I have some questions which I can't spot easily in your code (perhaps you can point me to specific parts of your PR):

1. Presumably the data coming from the result of RPC call is in one of two forms - either (a) an InputStream that is read directly from the network socket, or (b) a buffer that is already in-memory?

If its scenario (a), how do you know the size without reading all of the data? or do you just read up to 4 MB, and if it is 4MB or more, you flush the already read 4 MB to a file and then keep appending to the file until the stream is completely read?

If its scenario (b), are you avoiding making a second copy of the same data in-memory?

2. How are you bounding the total size of the cache? i.e. How do I make sure I don't have lots of 4 MB chunks of memory in-flight that ultimately add up to a large number and cause memory starvation?

Kind regards.

exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
reinhapa: @adamretter the buffering is done in the ContentFile itself if not based on memory buffer already...

reinhapa:
> This sounds interesting and like you are seeing a good performance improvement. If I understand correctly, when the result of an RPC call is less than 4 MB in size, then instead of writing to a file, you keep it in memory for a finite period of time. Do I have that correct?

Correct.

> If that is the case, then in general that sounds fine, but I have some questions which I can't spot easily in your code (perhaps you can point me to specific parts of your PR):
> 
> 1. Presumably the data coming from the result of RPC call is in one of two forms - either (a) an InputStream that is read directly from the network socket, or (b) a buffer that is already in-memory?
> 
> If its scenario (a), how do you know the size without reading all of the data? or do you just read up to 4 MB, and if it is 4MB or more, you flush the already read 4 MB to a file and then keep appending to the file until the stream is completely read?

Absolutely correct. This is done within the existing `VirtualTempPath` implementation introduced earlier on the client side within `AbstractRemoteResource`, `RemoteResourceSet`, `RemoteXMLResource` and `InternalXMLSerializer`

> If its scenario (b), are you avoiding making a second copy of the same data in-memory?
> 
> 2. How are you bounding the total size of the cache? i.e. How do I make sure I don't have lots of 4 MB chunks of memory in-flight that ultimately add up to a large number and cause memory starvation?

At the moment, there is no such handling for permitting such logic. The current `MemoryContentsImpl` class holding the actual data stores those in 4k blocks within a two dimensional array instead of using one big 4MB  byte array.

exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
adamretter: Presumably this was a new test that you added that was previously missing?

reinhapa: Correct, among a lot of others 😎

adamretter:
> Absolutely correct. This is done within the existing VirtualTempPath implementation introduced earlier on the client side

I think you are referring specifically to the `OverflowToDiskStream` class, is that correct?

> At the moment, there is no such handling for permitting such logic that may be added later...

Just my opinion, but if we are going to Cache things in memory we should have an upper bound on the total size of that cache otherwise it will lead to problems sooner or later.

> The current `MemoryContentsImpl` class holding the actual data stores those in 4k blocks within a two dimensional array instead of using one big 4MB byte array.

An interesting design! One concern, with that sizing, that is potentially a lot of small allocations, and as these buffers (as I think I understand it) are not from a pool, it will also mean a lot of GC.

eXist-db does have bounded caches already based on Caffeine, and also has some Pooling based on custom mechanisms such as the simplistic `ByteArrayPool` or the more complete Apache Commons Pool.

reinhapa:
> I think you are referring specifically to the `OverflowToDiskStream` class, is that correct?

Yes, that's exactly the implementation...

> Just my opinion, but if we are going to Cache things in memory we should have an upper bound on the total size of that cache otherwise it will lead to problems sooner or later.

That is correct, I do not know how much the RPC option is actually used and more enhancements in that diection are shure an option.

> An interesting design! One concern, with that sizing, that is potentially a lot of small allocations, and as these buffers (as I think I understand it) are not from a pool, it will also mean a lot of GC.

In case of pooling the already used memory instances would be beneficial in that regard I guess...

> eXist-db does have bounded caches already based on Caffeine, and also has some Pooling based on custom mechanisms such as the simplistic `ByteArrayPool` or the more complete Apache Commons Pool.

I can look into this a bit more in the next days. The question is where to configure and monitor all those parts... Do monitor exising caches that you mentioned?

adamretter:
> I do not know how much the RPC option is actually used

Hmm... My awareness is that it is used by Oxygen XML Editor, The Java Admin Client, and xst.
However they are all pretty much single-user uses. So not bounding the cache there won't make much difference.

But I guess like you, others are using it for integration - if I recall correctly, I think the Python libraries for eXist-db use the XML-RPC interface (as of course does the XML:DB interface that you wrote/updated).

reinhapa:
> For monitoring - some of it is exposed via JMX (particularly Caffeine and some of the Apache Commons pools). For sizing and configuration of this stuff, I think its normally done from `conf.xml` at startup.

@adamretter do you got any suggestions, where there is the best place to configure that to be created `ContentFilePool` within the `conf.xml`?

adamretter:
> do you got any suggestions, where there is the best place to configure that to be created ContentFilePool within the conf.xml?

Hmm... I am not certain at the moment I don't think there are any API specific settings in the `conf.xml` file. The closest thing in concept is perhaps the `binary-manager` section, so perhaps a new section that is similar to that? or perhaps conf.xml is not even the right place for this?

reinhapa:
> Hmm... I am not certain at the moment I don't think there are any API specific settings in the `conf.xml` file. The closest thing in concept is perhaps the `binary-manager` section, so perhaps a new section that is similar to that? or perhaps conf.xml is not even the right place for this?

At least for the server part I think `conf.xml` seems to be the right place for me. The question is if we would need a new section as it is used within the RPC servlet only...

@dizzzz @wolfgangmm @line-o  any other ideas on this?

reinhapa:
I created a separate PR #5195 adding a `ContentFile` pooling mechanism. It still has to be decided, where the configuration actually needs to go.

exist-core/src/main/java/org/exist/xmlrpc/RpcAPI.java
```diff
@@ -926,7 +927,7 @@ String hasUserLock(String path)
 
     List<String> getGroups() throws EXistException, PermissionDeniedException;
 
-    List<List> getIndexedElements(String collectionName, boolean inclusive)
+    List<List<Object>> getIndexedElements(String collectionName, boolean inclusive)
```
dizzzz, line 930: this works? 
image

dizzzz (approved):
LGTM....

reinhapa:
@adamretter I've integrated the pooling for those ContentFile instances now...

adamretter (changes requested):
LGTM. Can you address these Sonarcloud issues, then I think it's good to merge:

1. https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=5188&id=eXist-db_exist&open=AYz6Yodr6PRvCQqwhtF0
2. https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=5188&id=eXist-db_exist&open=AYz5OtO43AIjI1_SDloE
3. https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=5188&id=eXist-db_exist&open=AYz8jIkUHDD9adKHbyf3
4. https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=5188&id=eXist-db_exist&open=AY2Dy9Eqk9o22oitb7OP

Also I think these should be improved to be less fragile to time sensitivity (see Collection lock tests for an example), by polling the status or a latch, and a timeout?
1. https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=5188&id=eXist-db_exist&open=AYz5OtOi3AIjI1_SDloC
2. https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=5188&id=eXist-db_exist&open=AYz5OtOj3AIjI1_SDloD

reinhapa:
@adamretter I addressed all the changes you mentioned. Could you re-review it and approve in order to merge?

line-o (approved):
The issues reported by SonarCloud are unrelated to the actual changes in the code.
It was SonarCloud itself that suggested to add the `@Deprecated` annotation and now it complains about missing since and forRemoval. Finding out since which version that method is deprecated is a lot of extra work.  
I accept the changes as is. If you, @reinhapa, decide to add those this is just an extra.

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -480,78 +487,145 @@ public void testQueryWithStylesheet() throws XmlRpcException, MalformedURLExcept
...
     public void testExecuteQuery() throws XmlRpcException, MalformedURLException {
-        storeData();
+        final XmlRpcClient xmlrpc = storeData();
+
         List<Object> params = new ArrayList<>();
         String query = "distinct-values(//para)";
         params.add(query.getBytes(UTF_8));
-        params.add(new HashMap<>());
-        XmlRpcClient xmlrpc = getClient();
+        params.add(Map.of());
```
adamretter, line 604: Perhaps better to use `Collections.emptyMap()` ?

exist-core/src/main/java/org/exist/util/io/VirtualTempPath.java
```diff
@@ -52,6 +52,10 @@ public final class VirtualTempPath implements ContentFile {
     @GuardedBy("lock")
     private Path contentFile;
 
+    public VirtualTempPath(TemporaryFileManager tempFileManager) {
```
adamretter, line 55: Param can be `final`.

exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
```diff
@@ -3057,33 +3035,34 @@ private Map<String, Object> summaryToMap(final long queryTime, @Nullable final S
...
+                    final List<Object> temp = new ArrayList<>(4);
+                    temp.add(qname.getLocalPart());
+                    temp.add(qname.getNamespaceURI());
+                    temp.add(qname.getPrefix() == null ? "" : qname.getPrefix());
+                    temp.add(occurrence.getOccurrences());
+                    result.add(temp);
+                }
+                return result;
+            } else {
+                return List.of();
```
adamretter, line 3061: Perhaps `Collections.emptyList()` is better?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -116,58 +127,69 @@ private static String getUri() {
     @After
     public void tearDown() throws XmlRpcException, MalformedURLException {
         XmlRpcClient xmlrpc = getClient();
```
adamretter, line 129: Variable could be `final`

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -116,58 +127,69 @@ private static String getUri() {
...
-        params.add(TARGET_COLLECTION.toString());
-        @SuppressWarnings("unused")
-        Boolean b = (Boolean) xmlrpc.execute("removeCollection", params);
+        assertThat(xmlrpc.execute("removeCollection", List.of(TARGET_COLLECTION.toString()))).isInstanceOf(Boolean.class);
     }
 
     @Test
     public void testStoreAndRetrieve() throws XmlRpcException, IOException {
-        XmlRpcClient xmlrpc = getClient();
+        XmlRpcClient xmlrpc = createCollection(TARGET_COLLECTION);
```
adamretter, line 135: Variable could be `final`

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -116,58 +127,69 @@ private static String getUri() {
...
+
         List<Object> params = new ArrayList<>();
```
adamretter, line 137: Variable could be `final`

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -116,58 +127,69 @@ private static String getUri() {
...
+        assertThat(xmlrpc.execute("getDocumentAsString", params)).isInstanceOf(String.class);
 
         params.clear();
         params.add(TARGET_RESOURCE.toString());
-        params.add(options);
+        params.add(Map.of());
         Map table = (Map) xmlrpc.execute("getDocumentData", params);
 
-        try (final UnsynchronizedByteArrayOutputStream os = new UnsynchronizedByteArrayOutputStream()) {
+        try (final ByteArrayOutputStream os = new ByteArrayOutputStream()) {
```
adamretter, line 190: Why replace `UnsynchronizedByteArrayOutputStream` with less performant `ByteArrayOutputStream`?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -116,58 +127,69 @@ private static String getUri() {
...
             int offset = (int) table.get("offset");
-            data = (byte[]) table.get("data");
+            byte[] data = (byte[]) table.get("data");
```
adamretter, line 192: Variables could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -179,58 +201,53 @@ public void testStoreAndRetrieve() throws XmlRpcException, IOException {
...
         params.clear();
         params.add(TARGET_RESOURCE.toString());
-        Boolean b = (Boolean) xmlrpc.execute("hasDocument", params);
-        assertTrue(b);
+        assertThat(xmlrpc.execute("hasDocument", params)).isEqualTo(TRUE);
     }
 
-    private void storeData() throws XmlRpcException, MalformedURLException {
+    private XmlRpcClient createCollection(XmldbURI collection) throws XmlRpcException, MalformedURLException {
         XmlRpcClient xmlrpc = getClient();
```
adamretter, line 213: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -179,58 +201,53 @@ public void testStoreAndRetrieve() throws XmlRpcException, IOException {
...
-        params.add(TARGET_COLLECTION.toString());
-        Boolean result = (Boolean) xmlrpc.execute("createCollection", params);
-        assertTrue(result);
+        assertThat(xmlrpc.execute("createCollection", List.of(collection.toString()))).isEqualTo(TRUE);
+        return xmlrpc;
+    }
 
-        params.clear();
+    private XmlRpcClient storeData() throws XmlRpcException, MalformedURLException {
+        XmlRpcClient xmlrpc = createCollection(TARGET_COLLECTION);
```
adamretter, line 219: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -179,58 +201,53 @@ public void testStoreAndRetrieve() throws XmlRpcException, IOException {
...
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 221: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -179,58 +201,53 @@ public void testStoreAndRetrieve() throws XmlRpcException, IOException {
...
-        final XmlRpcClient xmlrpc = getClient();
-        List<Object> params = new ArrayList<>();
-        params.add(TARGET_COLLECTION.toString());
-        Boolean result = (Boolean) xmlrpc.execute("createCollection", params);
-        assertTrue(result);
+        final XmlRpcClient xmlrpc = createCollection(TARGET_COLLECTION);
 
-        params.clear();
-        final String generatedXml = generateXml((int)(MAX_DOWNLOAD_CHUNK_SIZE * 1.5));
+        List<Object> params = new ArrayList<>();
```
adamretter, line 245: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -249,30 +266,25 @@ public void getDocumentDataChunked_nextChunk() throws IOException, XmlRpcExcepti
...
-        final XmlRpcClient xmlrpc = getClient();
-        List<Object> params = new ArrayList<>();
-        params.add(TARGET_COLLECTION.toString());
-        Boolean result = (Boolean) xmlrpc.execute("createCollection", params);
-        assertTrue(result);
+        final XmlRpcClient xmlrpc = createCollection(TARGET_COLLECTION);
 
-        params.clear();
-        final String generatedXml = generateXml((int)(MAX_DOWNLOAD_CHUNK_SIZE * 1.75));
+        List<Object> params = new ArrayList<>();
```
adamretter, line 282: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -356,46 +368,36 @@ public void uploadCompressedAndDownload() throws IOException, XmlRpcException {
...
         }
     }
 
     @Test
     public void testRemoveCollection() throws XmlRpcException, MalformedURLException {
-        storeData();
-        XmlRpcClient xmlrpc = getClient();
+        final XmlRpcClient xmlrpc = storeData();
+
         List params = new ArrayList(1);
```
adamretter, line 379: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -356,46 +368,36 @@ public void uploadCompressedAndDownload() throws IOException, XmlRpcException {
...
+        assertThat(xmlrpc.execute("hasCollection", params)).isEqualTo(FALSE);
     }
 
     @Test
     public void testRemoveDoc() throws XmlRpcException, MalformedURLException {
-        storeData();
-        XmlRpcClient xmlrpc = getClient();
+        final XmlRpcClient xmlrpc = storeData();
+
         List params = new ArrayList(1);
```
adamretter, line 390: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -356,46 +368,36 @@ public void uploadCompressedAndDownload() throws IOException, XmlRpcException {
...
+        assertThat(xmlrpc.execute("remove", params)).isEqualTo(TRUE);
+        assertThat(xmlrpc.execute("hasDocument", params)).isEqualTo(FALSE);
     }
 
     @Test
     public void testRetrieveDoc() throws XmlRpcException, MalformedURLException {
-        storeData();
+        final XmlRpcClient xmlrpc = storeData();
+
         Map<String, String> options = new HashMap<>();
```
adamretter, line 401: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -407,50 +409,56 @@ public void testRetrieveDoc() throws XmlRpcException, MalformedURLException {
...
+                "\u00E4\u00E4\u00F6\u00F6\u00FC\u00FC\u00C4\u00C4\u00D6\u00D6\u00DC\u00DC\u00DF\u00DF",
+                "\uC5F4\uB2E8\uACC4");
     }
 
     @Test
     public void testQuery() throws XmlRpcException, MalformedURLException {
-        storeData();
+        final XmlRpcClient xmlrpc = storeData();
+
         List<Object> params = new ArrayList<>();
```
adamretter, line 438: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -407,50 +409,56 @@ public void testRetrieveDoc() throws XmlRpcException, MalformedURLException {
...
+        params.add(Map.of(RpcAPI.PROTECTED_MODE, "/db"));
+
+        assertThat((byte[]) xmlrpc.execute("query", params)).isNotNull().isNotEmpty();
+    }
+
+    @Test
+    public void testQuerySummary() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 452: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -407,50 +409,56 @@ public void testRetrieveDoc() throws XmlRpcException, MalformedURLException {
...
+
+        assertThat((Map<String, Object>) xmlrpc.execute("querySummary", params)).isNotEmpty();
     }
 
     @Test
     public void testQueryWithStylesheet() throws XmlRpcException, MalformedURLException, SAXException, IOException {
-        storeData();
+        final XmlRpcClient xmlrpc = storeData();
+
         Map<String, String> options = new HashMap<>();
```
adamretter, line 462: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -480,78 +487,145 @@ public void testQueryWithStylesheet() throws XmlRpcException, MalformedURLExcept
...
-        assertFalse(diff.toString(), diff.hasDifferences());
+        assertThat(diff.hasDifferences()).withFailMessage(diff.toString()).isFalse();
     }
 
     @Test
     public void testCompile() throws XmlRpcException, MalformedURLException {
-        storeData();
+        final XmlRpcClient xmlrpc = storeData();
+
         List<Object> params = new ArrayList<>();
```
adamretter, line 497: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -480,78 +487,145 @@ public void testQueryWithStylesheet() throws XmlRpcException, MalformedURLExcept
...
     }
 
     @Test
-    public void testAddAccount() throws MalformedURLException, XmlRpcException {
-        String user = "rudi";
-        String passwd = "pass";
+    public void testAccount() throws MalformedURLException, XmlRpcException {
+        final String user = "rudi";
+        final String passwd = "pass";
         String simpleMd5 = MessageDigester.md5(passwd, true);
```
adamretter, line 511: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -480,78 +487,145 @@ public void testQueryWithStylesheet() throws XmlRpcException, MalformedURLExcept
...
+        params.clear();
+        params.add(user);
+        params.add(simpleMd5);
+        params.add(digest);
+        params.add(new String[]{"guest"});
+        assertThat(xmlrpc.execute("updateAccount", params)).isEqualTo(TRUE);
+
+        assertThat(xmlrpc.execute("setUserPrimaryGroup", List.of(user, "guest"))).isEqualTo(TRUE);
 
         XmlRpcClientConfigImpl config = (XmlRpcClientConfigImpl) xmlrpc.getClientConfig();
```
adamretter, line 535: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -480,78 +487,145 @@ public void testQueryWithStylesheet() throws XmlRpcException, MalformedURLExcept
...
+        xmlrpc.execute("addGroupManager", List.of("admin", "testGroup"));
+        xmlrpc.execute("removeGroup", List.of("testGroup"));
     }
 
     @Test
     public void testExecuteQuery() throws XmlRpcException, MalformedURLException {
-        storeData();
+        final XmlRpcClient xmlrpc = storeData();
+
         List<Object> params = new ArrayList<>();
```
adamretter, line 601: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -480,78 +487,145 @@ public void testQueryWithStylesheet() throws XmlRpcException, MalformedURLExcept
...
-        item = (byte[]) xmlrpc.execute("retrieve", params);
+        assertThat(xmlrpc.execute("retrieve", params)).isInstanceOf(byte[].class);
     }
 
     @Test
     public void testQueryModuleExternalVar() throws XmlRpcException, MalformedURLException {
-        storeData();
+        final XmlRpcClient xmlrpc = storeData();
+
         List<Object> params = new ArrayList<>();
```
adamretter, line 629: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -609,7 +673,7 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
                 break;
             }
         }
-        assertTrue("added collection not found", foundMatch);
+        assertThat(foundMatch).withFailMessage("added collection not found").isTrue();
 
         Map<String, String> options = new HashMap<>();
```
adamretter, line 678: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
         params.add(options);
 
         // execute the call
-        byte[] data = (byte[]) xmlrpc.execute("getDocument", params);
+        assertThat(xmlrpc.execute("getDocument", params)).isInstanceOf(byte[].class);
+    }
+
+    @Test
+    public void testGetVersion() throws XmlRpcException, MalformedURLException {
+        XmlRpcClient xmlrpc = getClient();
```
adamretter, line 694: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        assertThat(xmlrpc.execute("getVersion", List.of())).isEqualTo(Version.getVersion());
```
adamretter, line 695: Perhaps  better to use `Collections.emptyList()`?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+    }
+
+    @Test
+    public void testConfigureCollection() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = createCollection(TARGET_COLLECTION);
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 702: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_COLLECTION.toString());
+        params.add("<configuration/>");
+        assertThat(xmlrpc.execute("configureCollection", params)).isEqualTo(TRUE);
+    }
+
+    @Test
+    public void testCreateId() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = createCollection(TARGET_COLLECTION);
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 712: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_COLLECTION.toString());
+        assertThat((String) xmlrpc.execute("createId", params)).endsWith(".xml");
+    }
+
+    @Test
+    public void testCreateResourceId() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = createCollection(TARGET_COLLECTION);
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 721: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_COLLECTION.toString());
+        assertThat((String) xmlrpc.execute("createResourceId", params)).endsWith(".xml");
+    }
+
+    @Test
+    public void testGetCollectionDesc() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 730: Variables could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_COLLECTION.toString());
+        Map<String, Object> response = (Map<String, Object>) xmlrpc.execute("getCollectionDesc", params);
+        assertThat(response).isNotEmpty();
+    }
+
+    @Test
+    public void testExistsAndCanOpenCollection() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 740: Variables could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_COLLECTION.toString());
+        assertThat(xmlrpc.execute("existsAndCanOpenCollection", params)).isEqualTo(TRUE);
+    }
+
+    @Test
+    public void testDescribeResource() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 749: Variables could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_RESOURCE.toString());
+        Map<String, Object> result = (Map<String, Object>) xmlrpc.execute("describeResource", params);
+        assertThat(result).isNotEmpty();
+    }
+
+    @Test
+    public void testGetContentDigest() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 759: Variables could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        assertThat(result).isNotEmpty();
+    }
+
+    @Test
+    public void testDocumentListing() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        assertThat((Object[]) xmlrpc.execute("getDocumentListing", List.of())).isNotEmpty();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 772: Variables could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_COLLECTION.toString());
+        assertThat((Object[]) xmlrpc.execute("getDocumentListing", params)).isNotEmpty();
+    }
+
+    @Test
+    public void testGetCollectionListing() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 781: Variables could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add("/db");
+        Object[] result = (Object[]) xmlrpc.execute("getCollectionListing", params);
+        assertThat(result).isNotEmpty();
+    }
+
+    @Test
+    public void testGetResourceCount() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 791: Variables could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(MODULE_RESOURCE.toString());
+        params.add("guest");
+        assertThat(xmlrpc.execute("chgrp", params)).isEqualTo(TRUE);
+    }
+
+    @Test
+    public void testChown() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 855: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+
+        params.add("guest");
+        assertThat(xmlrpc.execute("chown", params)).isEqualTo(TRUE);
+    }
+
+    @Test
+    public void testGetBinaryResource() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 868: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(MODULE_RESOURCE.toString());
+        assertThat((byte[]) xmlrpc.execute("getBinaryResource", params)).isNotEmpty();
+    }
+
+    @Test
+    public void testListDocumentPermissions() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 877: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_COLLECTION.toString());
+        assertThat((Map<String, List>) xmlrpc.execute("listDocumentPermissions", params)).isNotEmpty();
+    }
+
+    @Test
+    public void testListCollectionPermissions() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 886: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add("/db");
+        params.add("xmlrpc");
+        assertThat((Map<String, List>) xmlrpc.execute("getSubCollectionPermissions", params)).isNotEmpty();
+    }
+
+    @Test
+    public void testGetSubCollectionCreationTime() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 905: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add("/db");
+        assertThat((Map<String, List>) xmlrpc.execute("listCollectionPermissions", params)).isNotEmpty();
+    }
+
+    @Test
+    public void testGetSubCollectionPermissions() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 895: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+
+        params.set(0, "//test".getBytes());
+        assertThat((Map<String, Object>) xmlrpc.execute("queryP", params)).isNotEmpty();
+    }
+
+    @Test
+    public void testQueryPT() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 969: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(Map.of());
+        assertThat((String) xmlrpc.execute("printDiagnostics", params))
+                .contains("$tm:imported-external-string, $tm-query:local-external-string");
+    }
+
+    @Test
+    public void testQueryP() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 954: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_RESOURCE.toString());
+        assertThat((Object[]) xmlrpc.execute("getTimestamps", params)).isNotEmpty();
+    }
+
+    @Test
+    public void testPrintDiagnostics() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 943: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add(TARGET_COLLECTION.toString());
+        params.add(TEST_XML_URI.toString());
+        assertThat((Map<String, List>) xmlrpc.execute("getSubResourcePermissions", params)).isNotEmpty();
+    }
+
+    @Test
+    public void testGetCreationDate() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 925: Variable could be final?

exist-core/src/test/java/org/exist/xmlrpc/XmlRpcTest.java
```diff
@@ -622,7 +686,331 @@ public void testCollectionWithAccentsAndSpaces() throws XmlRpcException, Malform
...
+        params.add("/db");
+        params.add("xmlrpc");
+        assertThat((Long)xmlrpc.execute("getSubCollectionCreationTime", params)).isPositive();
+    }
+
+    @Test
+    public void testGetSubResourcePermissions() throws XmlRpcException, MalformedURLException {
+        final XmlRpcClient xmlrpc = storeData();
+
+        List<Object> params = new ArrayList<>();
```
adamretter, line 915: Variable could be final?

adamretter (commented):
Great piece of work thanks @reinhapa :-)
Just 3 observations:

1. Some variables/params could be marked `final` please
2. Empty collections are maybe better created with Collections.emptyXXX to make intention clear.
3. Not sure why the change from `UnsynchronizedByteArrayOutputStream` to the less performant `ByteArrayOutputStream`. I spent time previously chasing out ByteArrayOutput/InputStream from the code base.

reinhapa:
Hi @adamretter 

> * Some variables/params could be marked `final` please

Most places where already that way before my change but should now be changed.

> * Empty collections are maybe better created with Collections.emptyXXX to make intention clear.

I've changed those accordingly
 
> * Not sure why the change from `UnsynchronizedByteArrayOutputStream` to the less performant `ByteArrayOutputStream`. I spent time previously chasing out ByteArrayOutput/InputStream from the code base.

That was done because of the deprecation warning... I reverted that change, we might need to fix those warnings by other means...

PR #5195, opened 2024-01-23
[feature] Adds ContentFilePool

reinhapa:
In preparation for #5188 this PR adds a pooling mechanism for `ContentFile` objects.


reinhapa:
Would be something to configure the pool and the content file ok?

```
<exist>
    ...
    <rpc-server>
        <content-file max-in-memory-size="4196"/>
        <content-file-pool size="10"/>
    </rpc-server>
</exist>
```

exist-core/pom.xml
```diff
@@ -589,6 +589,16 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
adamretter, line 592: Do we really need to add another Mocking framework? Is it not possible to use EasyMock here (there is already code dependent on that)

exist-core/src/main/java/org/exist/util/Configuration.java
```diff
@@ -1582,10 +1610,14 @@ public void removeProperty(final String name) {
     }
 
     public int getInteger(final String name) {
+        return getInteger(name, -1);
+    }
+
+    public int getInteger(final String name, int defaultValue) {
```
adamretter, line 1616: I think `defaultValue` could be marked `final` too

exist-core/src/main/java/org/exist/util/io/ContentFilePool.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+public final class ContentFilePool extends GenericObjectPool<ContentFile> {
+    public static final String PROPERTY_IN_MEMORY_SIZE = "rpc-server.content-file.in-memory-size";
+    public static final String PROPERTY_POOL_SIZE = "rpc-server.content-file-pool.size";
+
+    public ContentFilePool(final TemporaryFileManager tempFileManager, final Configuration config, final int maxIdle) {
+        super(new ContentFilePoolObjectFactory(tempFileManager, toInMemorySize(config)), toPoolConfig(config, maxIdle));
+    }
+
+    private static int toInMemorySize(Configuration config) {
```
adamretter, line 40: Parameter could be marked final

exist-core/src/main/java/org/exist/util/io/ContentFilePool.java
```diff
@@ -0,0 +1,75 @@
...
+        poolConfig.setMaxTotal(config.getInteger(PROPERTY_POOL_SIZE));
+        poolConfig.setJmxNameBase("org.exist.management.exist:type=ContentFilePool");
+        return poolConfig;
+    }
+
+    @Override
+    public ContentFile borrowObject() {
+        try {
+            return super.borrowObject();
+        } catch (Exception e) {
```
adamretter, line 58: Exception could be marked final.

exist-core/src/main/java/org/exist/util/io/ContentFilePool.java
```diff
@@ -0,0 +1,75 @@
...
+            throw new IllegalStateException("Error while borrowing ContentFile", e);
+        }
+    }
+
+    @Override
+    public void returnObject(ContentFile obj) {
```
adamretter, line 64: parameter and exception could be marked final.

exist-core/src/main/java/org/exist/util/io/ContentFilePoolObjectFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.commons.pool2.impl.DefaultPooledObject;
+
+/**
+ * @author <a href="mailto:patrick@reini.net">Patrick Reinhart</a>
+ */
+public final class ContentFilePoolObjectFactory extends BasePooledObjectFactory<ContentFile> {
+    private final TemporaryFileManager tempFileManager;
+    private final int inMemorySize;
+
+    public ContentFilePoolObjectFactory(final TemporaryFileManager tempFileManager, int inMemorySize) {
```
adamretter, line 36: Needs some Javadoc please

exist-core/src/main/java/org/exist/util/io/ContentFilePoolObjectFactory.java
```diff
@@ -0,0 +1,50 @@
...
+package org.exist.util.io;
+
+import org.apache.commons.pool2.BasePooledObjectFactory;
+import org.apache.commons.pool2.PooledObject;
+import org.apache.commons.pool2.impl.DefaultPooledObject;
+
+/**
+ * @author <a href="mailto:patrick@reini.net">Patrick Reinhart</a>
+ */
+public final class ContentFilePoolObjectFactory extends BasePooledObjectFactory<ContentFile> {
```
adamretter, line 32: Needs some class Javadoc please

exist-core/src/main/java/org/exist/util/io/ContentFilePoolObjectFactory.java
adamretter: inMemorySize parameter could also be final

exist-core/src/main/java/org/exist/util/io/ContentFilePool.java
```diff
@@ -0,0 +1,75 @@
...
+import org.exist.util.Configuration;
+
+/**
+ * @author <a href="mailto:patrick@reini.net">Patrick Reinhart</a>
+ */
+public final class ContentFilePool extends GenericObjectPool<ContentFile> {
+    public static final String PROPERTY_IN_MEMORY_SIZE = "rpc-server.content-file.in-memory-size";
+    public static final String PROPERTY_POOL_SIZE = "rpc-server.content-file-pool.size";
+
+    public ContentFilePool(final TemporaryFileManager tempFileManager, final Configuration config, final int maxIdle) {
```
adamretter, line 36: Needs some Javadoc please

exist-core/src/main/java/org/exist/util/io/ContentFilePool.java
```diff
@@ -0,0 +1,75 @@
...
+
+import org.apache.commons.pool2.impl.GenericObjectPool;
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+import org.exist.util.Configuration;
+
+/**
+ * @author <a href="mailto:patrick@reini.net">Patrick Reinhart</a>
+ */
+public final class ContentFilePool extends GenericObjectPool<ContentFile> {
+    public static final String PROPERTY_IN_MEMORY_SIZE = "rpc-server.content-file.in-memory-size";
```
adamretter, line 33: Public static members need Javadoc please

exist-core/src/main/java/org/exist/util/io/ContentFilePool.java
```diff
@@ -0,0 +1,75 @@
...
+package org.exist.util.io;
+
+import org.apache.commons.pool2.impl.GenericObjectPool;
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+import org.exist.util.Configuration;
+
+/**
+ * @author <a href="mailto:patrick@reini.net">Patrick Reinhart</a>
+ */
+public final class ContentFilePool extends GenericObjectPool<ContentFile> {
```
adamretter, line 32: Needs some class Javadoc please

exist-core/src/test/java/org/exist/util/ConfigurationTest.java
```diff
@@ -0,0 +1,124 @@
...
+    Path existHomeDir;
+
+    @Test
+    void testConfigurationConstructors() throws Exception {
+        assertThatNoException().isThrownBy(Configuration::new);
+        assertThatNoException().isThrownBy(() -> new Configuration(null));
+        assertThatNoException().isThrownBy(() -> new Configuration(writeConfig(existHomeDir.resolve("myconf.xml")), Optional.of(existHomeDir)));
+    }
+
+    private String writeConfig(Path confFile) throws IOException {
```
adamretter, line 46: I think it would be better to add this as a file to the test resources folder, name it appropriately and load it at test execution time. There should be some existing examples in the tests of how to do this...

adamretter (changes requested):
Good first draft and the approach is fine - just needs some cleanup IMHO.

exist-core/pom.xml
reinhapa: Any way to mock non interfaces with EasyMock?

adamretter: Yes. It used to be an extension, but I think it is just built-in in modern versions of EasyMock.

adamretter: One such existing example of this - https://github.com/eXist-db/exist/blob/eXist-6.2.0/exist-core/src/test/java/org/exist/storage/journal/JournalTest.java#L282

reinhapa:
@adamretter I tried to process all suggested issues now..

exist-core/src/main/java/org/exist/util/io/ContentFile.java
```diff
@@ -25,6 +25,7 @@
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 
 /**
  * @author <a href="mailto:patrick@reini.net">Patrick Reinhart</a>
```
dizzzz, line 31: please could you elaborate on the responsibility of this interface?

dizzzz (approved):

dizzzz:
I'd like to have some explanation on the mocking.....

exist-core/src/main/java/org/exist/util/io/ContentFile.java
```diff
@@ -42,4 +43,7 @@ default InputStream newInputStream() throws IOException {
        return new ByteArrayInputStream(getBytes());
     }
 
+    default OutputStream newOutputStream() throws IOException {
```
adamretter, line 46: Whilst in this area, could you add the missing Javadoc to the interface class and methods please?

exist-core/src/test/resources/CompleteConf.xml
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
adamretter, line 1: Just wondering about the naming of this file. What is it that makes it `complete` compared to other configs? I realise it is used by your `ConfigurationTest.java` - but is there another reason why this is separate to the existing `conf.xml` used for test resources? or could you just add the xml-rpc settings to one of the existing files?

exist-core/src/main/java/org/exist/util/io/ContentFile.java
reinhapa: @adamretter I will do that according to the comment of @dizzzz in the next moments...

exist-core/src/test/resources/CompleteConf.xml
reinhapa: > Just wondering about the naming of this file. What is it that makes it `complete` compared to other configs? I realise it is used by your `ConfigurationTest.java` - but is there another reason why this is separate to the existing `conf.xml` used for test resources? or could you just add the xml-rpc settings to one of the existing files?

@adamretter what would be the best file to refer instead of using a custom one?

adamretter: Can you explain what you mean by `complete` here? Then I can advise better...

reinhapa: I tried to use  a better name

adamretter (approved):
LGTM. Thanks

PR #5197, opened 2024-01-24
[bugfix] Fixed vunerable lib & wrong variable name

reinhapa:
- Fixes known vulnerability using an newer library
- Fixed a wrong version variable


adamretter (approved):
LGTM

PR #5246, opened 2024-03-11
[refactor] Mechanical code improvements extensions

dizzzz:
Generic Improvement of code quality.

The code coverage is not fair :-)

PR #5256, opened 2024-03-19
[refactor] RangeIndexConfigAttributeCondition for readability

line-o:
- less indentation
- use final variables where possible
- return early
- remove this. where possible
- switch negated conditions where it made sense

Followup PR to #5246

### DIFF
--- a/extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
+++ b/extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
@@ -29,7 +29,6 @@ import org.exist.xquery.*;
 import org.exist.xquery.modules.range.RangeQueryRewriter;
 import org.exist.indexing.range.RangeIndex.Operator;
 import org.exist.xquery.value.AtomicValue;
-import org.exist.xquery.value.NumericValue;
 import org.exist.xquery.value.Sequence;
 import org.exist.xquery.value.StringValue;
 import org.w3c.dom.Element;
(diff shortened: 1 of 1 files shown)


reinhapa (approved):

line-o:
I broke 4 tests so I need to fix those first.

extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
```diff
@@ -56,143 +56,136 @@ public class RangeIndexConfigAttributeCondition extends RangeIndexConfigConditio
...
 
-
         final String caseString = elem.getAttribute("case");
-        final String numericString = elem.getAttribute("numeric");
+        caseSensitive = (caseString != null && !caseString.equalsIgnoreCase("no"));
 
-        this.caseSensitive = (caseString != null && !caseString.equalsIgnoreCase("no"));
-        this.numericComparison = (numericString != null && numericString.equalsIgnoreCase("yes"));
+        final String numericString = elem.getAttribute("numeric");
+        numericComparison = (numericString != null && numericString.equalsIgnoreCase("yes"));
```
dizzzz, line 101: I tend to use StringUtils here for readability... but that is another dependency.

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#equalsIgnoreCase-java.lang.CharSequence-java.lang.CharSequence-

dizzzz (approved):

extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
line-o: Since this pattern is littered throughout exist-db's code-base it would either make sense to:
- create our own utility function
- use string-utils everywhere

PR #5259, opened 2024-03-26
[refactor] Restructure RangeIndexAttributeCondition

line-o:
### Description:

These private properties of RangeIndexAttributeCondition

- String value
- boolean caseSensitive
- boolean numericComparison
- Double numericValue
- String lowercaseValue
- Pattern pattern

were replaced by

- boolean commutative
- boolean commutativeNegate
- Predicate<String> indexPredicate
- Predicate<AtomicValue> queryPredicate

indexPredicate and queryPredicate are references to one-line lambdas.
Before, a lot of switching and checking happened on each call to `find` and `matches`

### DIFF
--- a/extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
+++ b/extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
@@ -34,9 +34,10 @@ import org.exist.xquery.value.Sequence;
 import org.exist.xquery.value.StringValue;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
+import org.apache.commons.lang3.StringUtils;
 
 import javax.xml.XMLConstants;
-import java.util.function.BiPredicate;
+import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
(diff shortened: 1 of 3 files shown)


extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
```diff
@@ -34,293 +34,217 @@
...
-                if (caseSensitive) {
-                    return testString.equals(value);
-                }
-                return testString.equalsIgnoreCase(value);
-            }
-        } catch (XPathException e) {
-            RangeIndex.LOG.error(
-                    "Value conversion error when testing predicate for condition, value: {}", testValue.toString());
-            RangeIndex.LOG.error(e);
+    private static java.util.function.Predicate<AtomicValue> getTester (final String value, final boolean caseSensitive) {
```
reinhapa, line 206: Here you could remove the fully qualified name `java.util.function.Predicate` as it's already imported.

reinhapa (changes requested):
Ony a small change found ;-)

dizzzz (approved):

extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
line-o: Ah, I missed one :)

dizzzz (approved):

line-o:
@reinhapa I believe the changes you requested were addressed in the meantime

line-o:
@reinhapa could you please re-review, I believe it's good to go

extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexConfigAttributeCondition.java
```diff
@@ -34,293 +34,217 @@
...
-                pattern = Pattern.compile(value, flags);
-            } catch (PatternSyntaxException e) {
-                RangeIndex.LOG.error(e);
-                throw new DatabaseConfigurationException(
-                        "Range index module: Invalid regular expression in condition: " + value);
-            }
-        }
+        operator = getOperator(elem);
+        commutative = (operator == Operator.EQ || operator == Operator.NE);
+        commutativeNegate = (operator == Operator.GT || operator == Operator.LT || operator == Operator.GE || operator == Operator.LE);
```
adamretter, line 89: `commutativeNegate` is a very strange name, at least in English. Perhaps `notCommutative` would be better to describe what you are trying to model?

adamretter (changes requested):
I am not sure if this also addresses a bug or not? If it is just refactoring of existing code, then I have concerns:
1. The newer code is more complex and difficult to follow; I don't see what advantage it offers over the existing code.
2. The newer code exhibits symptoms of "information hiding". e.g. the variable `caseSensitive` (amongst others) is no longer knowable within an instance of the class at runtime; one side effect of that is that it will make debugging any future issues much more difficult.
3. I haven't tested it with JMH myself, but my intuition is that the new code will perform worse that the older code. Has this been tested with JMH yet?

line-o:
commutativeNegate could be renamed to converse (see https://en.m.wikipedia.org/wiki/Converse_relation and https://en.m.wikipedia.org/wiki/Inequality_(mathematics) ).

duncdrum:
@line-o this would need a rebase, is this still something you want to pursue for 7?

PR #5274, opened 2024-04-09
Fix NPE in NativeBroker

### LINKED ISSUE #5273: [BUG] NPE thrown on frequent updates of an XML resource
### Description:

Frequent updates on any XML resource stored in the database will cause a NullPointerException when it hits the fragmentation limit.

```xquery
update replace $stored-node with $new-node
```

This will result in

1. the document effectively being removed from the database
2. the data of that document to be no longer present in any configured index

### Reference:

Affected versions 6.1.0 and later (6.2.0, 6.3.0-SNAPSHOT, 7.0.0-SNAPSHOT)
Regression introduced in c553667

### Type of tests:

Evaluate following query

```xquery
xquery version "3.1";

(:sm:create-account("test", "test", "test"),:)
(:sm:create-group("test"),:)
xmldb:create-collection("/db", "test"),
sm:chgrp(xs:anyURI("/db/test"), "test"),
sm:chmod(xs:anyURI("/db/test"), "rwxrws---"),
xmldb:store("/db/test", "test.xml", <test><foo/></test>),
sm:chmod(xs:anyURI("/db/test/test.xml"), "rw-rw----"),
xmldb:create-collection("/db/apps", "test"),
sm:chmod(xs:anyURI("/db/apps/test"), "rwxrwxrwx"),
xmldb:store("/db/apps/test", "update.xql", "update replace doc('/db/test/test.xml')/test/foo with <foo updated='{util:system-time()}' />, doc('/db/test/test.xml')/test/foo", "application/xquery"),
sm:chown(xs:anyURI("/db/apps/test/update.xql"), "test"),
sm:chgrp(xs:anyURI("/db/apps/test/update.xql"), "test"),
sm:chmod(xs:anyURI("/db/apps/test/update.xql"), "rwxr-sr-x")
```

Then open `/exist/apps/test/update.xql` as guest

line-o:
### Description:

Frequent updates on any XML resource stored in the database will cause a NullPointerException when it hits the fragmentation limit.

This will have dire consequences on the consistency of the stored data:

1. the document is effectively removed from the database
2. the document is no longer present in the index

### Reference:

fixes #5273 

### Type of tests:

TBD

### DIFF
--- a/exist-distribution/pom.xml
+++ b/exist-distribution/pom.xml
@@ -715,6 +715,18 @@
                 <configuration>
                     <showfiles>true</showfiles>
                 </configuration>
+                <dependencies>
+                    <dependency>
+                        <groupId>commons-io</groupId>
+                        <artifactId>commons-io</artifactId>
+                        <version>2.11.0</version>
+                    </dependency>
+                    <dependency>
+                        <groupId>org.codehaus.plexus</groupId>
+                        <artifactId>plexus-utils</artifactId>
+                        <version>1.1</version>
+                    </dependency>
+                </dependencies>
                 <executions>
                     <execution>
                         <id>autodeploy-expath-pkgs-for-appassembler</id>
--- a/exist-core/src/main/java/org/exist/storage/NativeBroker.java
+++ b/exist-core/src/main/java/org/exist/storage/NativeBroker.java
@@ -3033,7 +3033,7 @@ public class NativeBroker extends DBBroker {
                 }
             }.run();
             // create a copy of the old doc to copy the nodes into it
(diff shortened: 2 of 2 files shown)


exist-distribution/pom.xml
```diff
@@ -715,6 +715,18 @@
                 <configuration>
                     <showfiles>true</showfiles>
                 </configuration>
+                <dependencies>
```
adamretter, line 718: Better to use the newer plugin where this was fixed some time ago.

adamretter: Although this change should be in a separate PR and not this PR really

Issue #4890, opened 2023-04-27
Memory Leak when restoring a Backup


adamretter:
When restoring a clean instance of eXist-db 6.2.0 from a backup we are able to observe what we believe to be a memory leak in the Full-Text and Range indexes of eXist-db; to be clear I am referring to the implementations of both of those indexes that are backed by Lucene.

The database backup is approximately 40,000 XML documents (35GB).

Having configured the JVM for eXist-db to `-XX:+HeapDumpOnOutOfMemoryError` and `-XX:+ExitOnOutOfMemoryError`, we have captured the heap dumps from the JVM and analyzed them with YourKit Java Profiler.

## Heap Dump 1
@dariok provided me with the file: `java_pid236689.hprof` (md5sum: `54c669ddba7532bee2449922f2fba1d5`)

![dump1-part1](https://user-images.githubusercontent.com/1264057/234953838-113d4778-6b43-46b1-bb30-67371b66f5b9.png)
![dump1-part2](https://user-images.githubusercontent.com/1264057/234953844-20a54610-d335-4dc2-951c-08b2601b06f7.png)

Shows that of the 2.9GB available to eXist-db, 1.1GB is retained (at the time that eXist-db ran out of memory) by the `RangeIndex`.

## Heap Dump 2
@dariok provided me with the file: `java_pid318495.hprof` (md5sum: `9a99f30986fe2579b8c54efe6547e05b`)

![dump2](https://user-images.githubusercontent.com/1264057/234954613-759f6757-de7f-460c-8c1a-1d04e4e2461b.png)
Re-running with a clean eXist-db instance and restoring the same backup again but with 6GB of memory available to eXist-db, we see that 2.5GB and 1.1GB are retained (at the time that eXist-db ran out of memory) by the `LuceneIndex` and `RangeIndex` respectively.

# Observations
1. The problem is not reproducible with eXist-db 6.0.1.
2. The git history shows that the only changes to those indexes were the PR https://github.com/eXist-db/exist/pull/4541 which first shipped in eXist-db 6.1.0.

----
For reference, the `collection.xconf` in use looks like:
```xml
<collection xmlns="http://exist-db.org/collection-config/1.0">
  <index xmlns:mets="http://www.loc.gov/METS/" xmlns:meta="https://github.com/dariok/wdbplus/wdbmeta" xmlns:tei="http://www.tei-c.org/ns/1.0" xmlns:xs="http://www.w3.org/2001/XMLSchema">
    <lucene>
[...]

PR #5295, opened 2024-04-30
Issue #4890 trigger leak analysis

reinhapa:
Analysis and simplifications to tackle the leaking trigger states.


reinhapa:
@line-o I did some clean ups in preparation for looking into the actual leak. As soon I got your minimal restore example, I try to setup a according test, that may show the issue and can be used to reproduce.

reinhapa:
> @reinhapa can you tell me a little about your thinking around removing the ThreadLocal please?

It's to get more easy access to all currently known per-thread entries in order to find the actual leak behaviour as those can be found more easy as within the thread local variable. It's not intended to stay that way when I got to the actual ground of the problem.

line-o:
@reinhapa here is the backup to restore
[full-backup-restore-with-triggers.zip](https://github.com/eXist-db/exist/files/15179694/full-backup-restore-with-triggers.zip)

The collection where an update trigger is configured is `/db/apps/restore-with-trigger-test`

adamretter:
> to find the actual leak behaviour

What is the leak that you are mentioning? is there a GitHub issue for this I can take a look at?

reinhapa:
> What is the leak that you are mentioning? is there a GitHub issue for this I can take a look at?

@adamretter its issue #4890 where trigger state objects accumulate and only released after the full backup is complete (as I understand @line-o)

dizzzz:
disable triggers -> breaks stuff ?

adamretter:
See: https://github.com/eXist-db/exist/issues/4890#issuecomment-2314672530

dizzzz:
Patrick agrees to close

PR #5304, opened 2024-05-13
[ci] test on Java 11

line-o:
### Description:

Test on Java 11

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -4,7 +4,7 @@ permissions:
   contents: read
 env:
   MAVEN_OPTS: -XX:StartFlightRecording=maxsize=5g,disk=true,dumponexit=true,settings=default,filename=./ -DtrimStackTrace=false -D'maven.resolver.transport=wagon'
-  DEV_JDK: '8'
+  DEV_JDK: '11'
 jobs:
   license:
     name: License check


dizzzz (commented):
afaik / have seen, java11 causes an issue runtime (tar.gz) due to a signing issue/inconsistency. Only one jar file has this issue.

adamretter:
Be aware, eXist-db 6 is significantly less performant on JDK 11 than on JDK 8.

reinhapa:
> Be aware, eXist-db 6 is significantly less performant on JDK 11 than on JDK 8.

@adamretter unfortunately JDK 8 is no longer available for the Mac OS Github action...

adamretter:
> @adamretter unfortunately JDK 8 is no longer available for the Mac OS Github action...

Is it not possible to have a step that install a JDK? If I recall there is even a pre-build step for this in the GitHub Marketplace that can be used

dizzzz:
> Be aware, eXist-db 6 is significantly less performant on JDK 11 than on JDK 8.

I was not aware of that..... or maybe I am. Any idea why?

line-o:
Runtime of tests was reasonable between 19 to 27 minutes for the entire test suite on java 11.
If there is a known slow-down of exist-db on java 11 I would like to see an issue with some more details to be shared. 
I would also like to add that we do use the java setup action.

In order for us to make sure that the existdb compiles and runs on java 8 as well we could add a separate test on java8 on linux workflow. How does that sound?

duncdrum (approved):

adamretter:
I don't understand the point of this. eXist-db 6 targets Java 8, so changing CI to Java 11 makes no sense

reinhapa:
> I don't understand the point of this. eXist-db 6 targets Java 8, so changing CI to Java 11 makes no sense

@adamretter I think using Java 11 to produce Java 8 byte code should be still ok instead of having investing a lot of time to handle an alternative Java setup instead. I think it would be better to consentrate on the next release instead and not wasting to much time in doing work-arounds for no longer supported build JDKs IMHO...

reinhapa (approved):

dizzzz:
I read/hear that the performance decline of java11 has been addresses in the JDK some time ago.

dizzzz:
OK to close?

PR #5276, opened 2024-04-11
[bugfix] fix errors on update operations

line-o:
### Description:

By setting the allowed-fragmentation attribute value to `-1` any modification will trigger the defragmentation of the dom.dbx (AKA BTree pages).

This setting triggers the NPE described in #5273 on the very first call to

```xquery
update (replace|insert) $nodeA (with|into) $nodeB
```

Unfortunately, it also triggers a totally different class of NPEs in XQueryTrigger and other subsystems of the database.

<details>
  <summary>Failures before d145f29857be1cf7be5ab15226f8239b5e5aac3d</summary>

```
[ERROR] Failures:
[ERROR]   TriggerConfigTest.removeDocument:137 Cannot invoke "org.exist.storage.BrokerPool.getSecurityManager()" because "pool" is null
[ERROR]   TriggerConfigTest.removeDocument:137 Cannot invoke "org.exist.storage.BrokerPool.getSecurityManager()" because "pool" is null
[ERROR]   TriggerConfigTest.removeDocument:137 Cannot invoke "org.exist.storage.BrokerPool.getSecurityManager()" because "pool" is null
[ERROR]   TriggerConfigTest.storeDocument:114 Cannot invoke "org.exist.storage.BrokerPool.getSecurityManager()" because "pool" is null
[ERROR]   TriggerConfigTest.storeDocument:114 Cannot invoke "org.exist.storage.BrokerPool.getSecurityManager()" because "pool" is null
[ERROR]   TriggerConfigTest.storeDocument:114 Cannot invoke "org.exist.storage.BrokerPool.getSecurityManager()" because "pool" is null
[ERROR]   TriggerConfigTest.updateTriggers:183 Cannot invoke "org.exist.storage.BrokerPool.getSecurityManager()" because "pool" is null
[ERROR]   TriggerConfigTest.updateTriggers:183 Cannot invoke "org.exist.storage.BrokerPool.getSecurityManager()" because "pool" is null
[ERROR]   TriggerConfigTest.updateTriggers:183 Cannot invoke "org.exist.storage.BrokerPool.getSecurityManager()" because "pool" is null
[ERROR]   XQueryTriggerTest.collectionCreate:522 expected:<1> but was:<0>
[ERROR]   XQueryTriggerTest.collectionMove:626 expected:<1> but was:<0>
[ERROR]   XQueryTriggerTest.documentBinaryDelete:475 expected:<1> but was:<0>
[...]

PR #5296, opened 2024-05-02
[6.x.x] Fix errors on update operations

line-o:
Backport of #5276
refs #5273

PR #5311, opened 2024-05-23
Fix PermissionDeniedException  in NativeBroker.defragXMLresource

line-o:
### Description:

Additional bugfix for NativeBroker.defragXMLresource where in specific setups with modules and collections with set group ID set a PermissionDeniedException is thrown when the modified XML resource is defragmented.

### Reference:

refs #5276

### Type of tests:

Code review was addressed in UpdateInsertTriggersDefrag.java

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/NativeBroker.java
+++ b/exist-core/src/main/java/org/exist/storage/NativeBroker.java
@@ -3171,9 +3171,9 @@ public class NativeBroker implements DBBroker {
                     return null;
                 }
             }.run();
-            // create a copy of the old doc to copy the nodes into it
+            // Create a copy of the old document to copy the nodes into it.
+            // tempDoc serves as a storage container for the child nodes.
             final DocumentImpl tempDoc = new DocumentImpl(null, doc.getDocId(), doc);
-            tempDoc.copyOf(this, doc, doc);
             final StreamListener listener = getIndexController().getStreamListener(doc, ReindexMode.STORE);
             // copy the nodes
             final NodeList nodes = doc.getChildNodes();
@@ -3198,15 +3198,17 @@ public class NativeBroker implements DBBroker {
                     return null;
                 }
             }.run();
+            // assign new child nodes from the temporary document
             doc.copyChildren(tempDoc);
             doc.setSplitCount(0);
             doc.setPageCount(tempDoc.getPageCount());
+            // store defragmented document
             storeXMLResource(transaction, doc);
             closeDocument();
             if (LOG.isDebugEnabled()) {
                 LOG.debug("Defragmentation took {} ms.", (System.currentTimeMillis() - start));
             }
-        } catch(final PermissionDeniedException | IOException e) {
+        } catch(final IOException e) {
             LOG.error(e);
         }
     }
--- a/exist-core/src/test/java/org/exist/xquery/update/UpdateInsertTriggersDefrag.java
+++ b/exist-core/src/test/java/org/exist/xquery/update/UpdateInsertTriggersDefrag.java
@@ -21,66 +21,56 @@
  */
 package org.exist.xquery.update;
 
-import org.exist.storage.DBBroker;
 import org.exist.test.ExistXmldbEmbeddedServer;
-import org.exist.test.TestConstants;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.xmldb.api.base.Collection;
 import org.xmldb.api.base.ResourceSet;
-import org.xmldb.api.base.XMLDBException;
 import org.xmldb.api.modules.CollectionManagementService;
 import org.xmldb.api.modules.XMLResource;
-import org.xmldb.api.modules.XPathQueryService;
 import org.xmldb.api.modules.XQueryService;
 
+import static org.exist.storage.DBBroker.PROPERTY_XUPDATE_FRAGMENTATION_FACTOR;
+import static org.exist.test.TestConstants.TEST_COLLECTION_URI;
+import static org.exist.test.TestConstants.TEST_XML_URI;
 import static org.exist.util.PropertiesBuilder.propertiesBuilder;
 import static org.junit.Assert.assertEquals;
 
 public class UpdateInsertTriggersDefrag {
     @ClassRule
-    public static final ExistXmldbEmbeddedServer exist = new ExistXmldbEmbeddedServer(false, true, true, propertiesBuilder().put(DBBroker.PROPERTY_XUPDATE_FRAGMENTATION_FACTOR, -1).build());
-    final String path = TestConstants.TEST_COLLECTION_URI + "/" + TestConstants.TEST_XML_URI.toString();
-    final String xml = "<list><item>initial</item></list>";
-    Collection testCollection;
-    XQueryService queryService;
-    CollectionManagementService collectionService;
-
-    /**
-     * stores XML String and get Query Service
-     *
-     * @param documentName to be stored in the DB
-     * @param content      to be stored in the DB
-     * @throws XMLDBException if an error occurs storing the document
-     */
-    private void storeXML(final String documentName, final String content) throws XMLDBException {
-        final XMLResource doc = testCollection.createResource(documentName, XMLResource.class);
-        doc.setContent(content);
-        testCollection.storeResource(doc);
-    }
+    public static final ExistXmldbEmbeddedServer exist = new ExistXmldbEmbeddedServer(false, true, true,
+            propertiesBuilder().put(PROPERTY_XUPDATE_FRAGMENTATION_FACTOR, -1).build());
+    private final String path = TEST_COLLECTION_URI + "/" + TEST_XML_URI.toString();
+    private Collection testCollection;
+    private CollectionManagementService collectionService;
 
     @Before
     public void setUp() throws Exception {
         collectionService = exist.getRoot().getService(CollectionManagementService.class);
-        testCollection = collectionService.createCollection(TestConstants.TEST_COLLECTION_URI.toString());
-        queryService = (XQueryService) testCollection.getService(XPathQueryService.class);
-        storeXML(TestConstants.TEST_XML_URI.toString(), xml);
+        testCollection = collectionService.createCollection(TEST_COLLECTION_URI.lastSegment().toString());
+        final XMLResource doc = testCollection.createResource(TEST_XML_URI.toString(), XMLResource.class);
+
+        doc.setContent("<list><item>initial</item></list>");
+        testCollection.storeResource(doc);
     }
 
     @After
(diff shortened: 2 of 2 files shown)


exist-core/src/main/java/org/exist/storage/NativeBroker.java
```diff
@@ -3171,9 +3171,9 @@ public Object start() {
                     return null;
                 }
             }.run();
-            // create a copy of the old doc to copy the nodes into it
+            // Create a copy of the old document to copy the nodes into it.
+            // tempDoc serves as a storage container for the child nodes.
             final DocumentImpl tempDoc = new DocumentImpl(null, doc.getDocId(), doc);
-            tempDoc.copyOf(this, doc, doc);
```
adamretter, line 3176: I don't think just removing this is the correct or secure solution. Can you write a bit more detail in the description of the PR about exactly what the problem was please, that might help me to suggest a better solution.

adamretter (dismissed):
Likely needs a more nuanced fix

exist-core/src/main/java/org/exist/storage/NativeBroker.java
line-o: The commit message body has some more detail.

> With this patch applied, this can no longer happen as the call to DocumentImpl#copyOf is removed in defragXMLresource.
> The temporary document is just a container to realign nodes on one BTree page and is destroyed within the method.

I hope this clarifies the motivation and justification.
I would like to point out that we arrived at this solution together with @wolfgangmm

line-o: > Likely needs a more nuanced fix

Please only block PRs where you have either proof that it creates a problem or at least share the thoughts what the issue might be.

adamretter: > Please only block PRs where you have either proof that it creates a problem

Obviously!

> or at least share the thoughts what the issue might be

I am concerned that the temporary document which looks persistent to me is not correctly initialised, also I am concerned that the security on it is now too relaxed.

Can you clearly describe the initial problem that you are trying to solve in the PR description please. For example, something like: There is an active user `U` with the security profile `s(U)`, there is document `D` with the permissions `p(D)`, when the action `A` is  executed by the subject `AS` with security profile `s(AS)`, it is rejected because `s(AS)` is incompatible with `p(D)`.
Obviously you need to replace the symbols above with the actual details.

>  hope this clarifies the motivation and justification.

It just discusses the solution you arrived at, it does not clearly describe the problem you were trying to solve.

wolfgangmm: I was debugging this code with Juri yesterday and as the original author of the defrag method, I confirm that the offending line `tempDoc.copyOf(this, doc, doc);` is not only unnecessary here, but can cause database corruption. The `tempDoc` is only used as a temporary container for copying nodes during defrag and is thrown away afterwards! Original permissions and ownership on the resource are not affected by the `defragXMLResource` operation in any way, because the `tempDoc` is never written. What is written is the original DocumentImpl, which still has the proper permissions and ownership.

It seems I originally added the call to `tempDoc.copyOf` 19 years ago and it was preserved during a series of refactorings later though the implementation of `DocumentImpl` had changed to the point that it now may actually cause a corruption when called in this case.

Please be aware that this issue is a ticking time bomb that will lead to data wipe out!!! Together with the even more fatal #5296 (because it's easier to run into) this is absolutely worth a hotfix, so I would give it a thumbs up, not excluding that more improvements could be done at a later point.

adamretter: Hi @wolfgangmm thanks for your reply. It is interesting to hear about these symptoms, and it does sound like something needs to be fixed. However, the actual original issue and its cause remain unclear. I would like further information please so that we may all understand it...

> but can cause database corruption

Could you explain how that is possible please? As I understood it from @line-o's description there was a `PermissionDeniedException` which has not yet been explained. I am not sure how a `PermissionDeniedException`, which is a genuine way of forbidding a subject from doing something they are not allowed to do, could cause a database corruption here? 

> Original permissions and ownership on the resource are not affected by the defragXMLResource operation in any way, because the tempDoc is never written.

I did not expect that they would be.

> What is written is the original DocumentImpl, which still has the proper permissions and ownership.

Okay.

What I (and presumably others) don't yet understand, and what has not yet been explained, is how can the current approach that places the same permissions onto `tempDoc` as the original resource that is to be defragmented, cause a `PermissionDeniedException` (and then also if I now understand you, a subsequent database corruption)?

> Please be aware that this issue is a ticking time bomb that will lead to data wipe out!!!

I agree that it sounds like it needs to be fixed. At the moment I don't have enough information to see if the proposed fix correctly addresses the cause. I trust you and @line-o will provide the requested information about the underlying issue and cause...

> Together with the even more fatal https://github.com/eXist-db/exist/pull/5296 (because it's easier to run into)

Let's consider that separately. As at the moment https://github.com/eXist-db/exist/pull/5296 is blocked by work that is still needed to address issues introduced in https://github.com/eXist-db/exist/pull/5276 which itself was rushed and merged in violation of the agreed community process for reviewing and merging PRs.

> this is absolutely worth a hotfix,

From what you have said so far, I agree. We just need more information to be able to complete the review of this PR.

line-o:
@adamretter Please note that the test was adapted to address your review in #5273 and also note that the test was written based on src/test/java/org/exist/xquery/update/AbstractTestUpdate.java
Any improvements in this _family of tests_ for updates should be addressed separately. The seriousness of the problem fixed weighs higher in my opinion.

adamretter:
> Any improvements in this family of tests for updates should be addressed separately. The seriousness of the problem fixed weighs higher in my opinion.

I don't agree.

exist-core/src/main/java/org/exist/storage/NativeBroker.java
wolfgangmm: @line-o added some information about how to reproduce the issue. It requires a certain combination of permissions. The corruption occurs because the PermissionDeniedException aborts the defrag process after some steps have already been applied. Therefore another way to prevent this might be to move the offending code further up to the start of the method. We did not follow this route as dropping the line is the cleaner way to solve the issue.

The PermissionDeniedException is eventually thrown in line 664 of `DocumentImpl#copyOf`: 

```
permissions.setOwner(prev.get_1().getOwner());
```

why this happens I cannot explain entirely and it seems very weird that the owners would be different. It must be related to the setgid. But in itself it has nothing to do with the actual defrag routine and points to another bug in permission handling.

adamretter: > it seems very weird that the owners would be different

Yes that was my feeling from the start, which is why I wanted more information from yourselves. Based on that knowledge it seems to me that the this PR is a workaround and not a fix for the root issue.

If @line-o could provide the information I previously requested [here](https://github.com/eXist-db/exist/pull/5311#discussion_r1612355788) please:

> Can you clearly describe the initial problem that you are trying to solve in the PR description please. For example, something like: There is an active user U with the security profile s(U), there is document D with the permissions p(D), when the action A is executed by the subject AS with security profile s(AS), it is rejected because s(AS) is incompatible with p(D).
Obviously you need to replace the symbols above with the actual details.

Then I should be able to understand, and then help you both resolve the real issue instead...

line-o:
@adamretter What do you mean with this statement?

> Your code fixed one thing you were interested in but also breaks another thing that you are not interested in.

What did the code break and which code are you talking about, the test?

line-o:
Maybe let's look at this the other way around. What is the use of the call to `DocumentImpl#copyOf` here? @adamretter

exist-core/src/main/java/org/exist/storage/NativeBroker.java
adamretter: @wolfgangmm I think you have accidentally deleted or overwritten my comments
Screenshot 2024-05-24 at 16 39 27

Are you able to restore my comments please? If not, I can try and re-create them...

wolfgangmm:
@adamretter please watch your language, which I perceive as disrespectful in the last comments answering @line-o. We have no direct interest in this bugfix as we're not using any code affected by it. It's all just for the community's benefit.

adamretter:
@wolfgangmm I will. No disrespect was or is intended. I'm sorry you felt that way. I have constantly written offers to help throughout this thread. I trust you have seen them?
I think I've also made it clear that I'd like to see this fixed. It sounds like it could be important. But I want to see a fix to the problem and not to the symptom.
To be doubly clear - I can't accept this PR when the problem (the root cause) is not understood.

line-o:
@adamretter I am wondering how to put two recent quotes from you together:

last week in this PR:

> Your code fixed one thing you were interested in but also breaks another thing that you are not interested in.

By the way, you still haven't answered what I broke, allegedly.

In Slack 3 days ago:

> My general advice to my customers still with eXist-db is not to use XQuery Update, and to instead always overtime the complete document 

So I fixed a thing users in our community are interested in: exist-db's XQuery update extension. You clearly stated that you are not interested in it. So what is your vested interest here, as this change only affects defragmentation which can only ever run after modification?

adamretter:
> By the way, you still haven't answered what I broke, allegedly.
Yes I did... it's been in the review comments for weeks - your newly added tests are leaking resources.


>> My general advice to my customers still with eXist-db is not to use XQuery Update, and to instead always overtime the complete document
> So I fixed a thing users in our community are interested in: exist-db's XQuery update extension. You clearly stated that you are not interested in it. So what is your vested interest here, as this change only affects defragmentation which can only ever run after modification?

I don't really see what you are getting at here! My interest is advising eXist-db users as to the best way of using eXist-db. Your fix addresses an issue with defragmentation, it does not fix other issues with XQuery Update.

line-o:
OK, so then have a look at the __test__ in this PR as it is __changed__ to address your review.  

>  it does not fix other issues with XQuery Update.

But it does fix an __existing, reproducable and severe__ issue with XQuery Update. Other issues, you hinted at, are out of scope.

adamretter:
> OK, so then have a look at the test in this PR as it is changed to address your review.

I'm not seeing that in this PR.

line-o:
Well it is the other changed file.
And your review so far did __only__ address the removal of the __unnecessary__ call of copyOf.
I would still like to hear why you think it cannot be removed.

dizzzz (approved):
LGTM

As it addresses a 'timebomb' I'd vote to pull the PR in ASAP.

line-o:
> No it isn't. It does not fully address my review and is still missing several things.

I would need some more info on what you think is missing.

reinhapa (approved):

dizzzz:
The community agrees to have it in...

duncdrum:
@line-o could you rebase please

line-o:
I will rebase this as we need to get it in.

PR #5385, opened 2024-08-01
[ci] Repair 6.x.x build for macOs

dizzzz:
Java8 on macOS seems really impossible to find in GHA. Spending more time is IMO not worth the effort.

Personally I do not expect OpenJDK8 on macOS to be really used, it is a corner case, Apple ditched Java8 support ages ago.

Whilst OpenJDK11 might be a bit slower [not sure if this is true] , keeping Java8 support on this OS is IMHO not worth the effort of the eXist-db contributors.

More valuable is to have green builds again.
OpenJDK works just fine as reported by end-users.

It is up to the community (monday eve) to decide and agree.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -39,6 +39,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
+        jvm: ['8']
+        exclude:
+          - os: macOS-latest
+            version: '8'
+        include:
+          - os: macOS-latest
+            version: '11'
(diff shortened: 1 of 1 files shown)


dizzzz:
@reinhapa can we use the key of 'develop' here as well?

image

dizzzz:
The windows builds show an interesting issue:

image

this is certainly related:

image

For one of the builds I saw an build / fork issue, I expect this might be the problem.

My suggestion for now is: update mvnd as in develop

dizzzz:
Plugins are upgraded in different PR

duncdrum (approved):

reinhapa (approved):

PR #5386, opened 2024-08-01
[ci] Upgrade all build related maven plugins

dizzzz:
Upgrade all build related maven plugins.

macOS build is expected to fail, this is addressed in #5385

### DIFF
--- a/exist-parent/pom.xml
+++ b/exist-parent/pom.xml
@@ -560,12 +560,12 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-clean-plugin</artifactId>
-                    <version>3.3.2</version>
+                    <version>3.4.0</version>
                 </plugin>
                 <plugin>
                     <groupId>com.mycila</groupId>
                     <artifactId>license-maven-plugin</artifactId>
-                    <version>4.3</version>
+                    <version>4.5</version>
                 </plugin>
                 <plugin>
                     <groupId>org.owasp</groupId>
@@ -609,7 +609,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-dependency-plugin</artifactId>
-                    <version>3.6.1</version>
+                    <version>3.7.1</version>
                 </plugin>
                 <plugin>
                     <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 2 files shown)


exist-parent/pom.xml
```diff
@@ -560,19 +560,19 @@
...
+                    <version>4.5</version>
                 </plugin>
                 <plugin>
                     <groupId>org.owasp</groupId>
                     <artifactId>dependency-check-maven</artifactId>
-                    <version>9.0.9</version>
+                    <version>10.0.3</version>
                     <configuration>
-                        <nvdApiKey>${env.NVD_API_KEY}</nvdApiKey>
+                        <nvdApiKey>cb452226-8de1-4d27-9032-7c9b8741c880</nvdApiKey>
```
duncdrum, line 575: Should this be hardcoded?

dizzzz: I tried and tried and tried.... it should be injected, which is working for develop, but I spent hours on the develop-6.x.x branche, nothing worked. I hardcoded it, because it was the only way to get it working.

The key is not really sensitive, so it should be ok-ish

duncdrum: I see @dizzzz  could you open an issue and assign it to me. I'm afk for a while and would like to gave a go at it once I m back. Should be ok to leave this for a limited time.

duncdrum (approved):

exist-parent/pom.xml
dizzzz: done. thnx (CC: @reinhapa )

adamretter:
The `<nvdApiKey>cb452226-8de1-4d27-9032-7c9b8741c880</nvdApiKey>` should NOT appear in Git! @dizzzz can you remove this please?

dizzzz:
Hi @adamretter, 

I agree 100% with you that keys etc should not be in checked-in code. 

However I spent a rather significant amount of time in trying to have the 6.x build working with the key 'injected' as a secret, like it works for 'develop'.

After many hours I gave up, nothing worked. Somehow GH deals with develop differently than with a branch. No idea why. My sole purpose was to have a decent build again.

In the weekly community call I presented the issue, and we agreed to solve it (for now) pragmatically, as the key is not really secret, absolutely no harm can be done if it is used elsewhere. Obtaining a key is simple, and free. I registered it with an exist-db account.

If I remember correctly @duncdrum would have a look at it to solve it in a correct way.

reinhapa:
> The `<nvdApiKey>cb452226-8de1-4d27-9032-7c9b8741c880</nvdApiKey>` should NOT appear in Git! @dizzzz can you remove this please?

@adamretter do you have an idea how to proved a secret key for PRs? In those cases, that key is also not available and I have no idea how to solve that (could solve this issue too)

PR #5436, opened 2024-08-30
Register binary attribute in collection.xconf.xsd

### LINKED ISSUE #5342: Node Comparison returns no data with nested predicate
The data (`example-1.xml`) looks like:
```xml
<text>
  <lg>
      <w wid="_1_d2e15" id="d1e14" lemma="thalla">Thalla</w>
      <pc join="left">,</pc>
      <w wid="_1_d2e23" id="d1e20" lemma="voc">a</w>
      <w wid="_1_d2e26" id="d1e23" pos="nx" lemma="leabhar">leabhair</w>
      <w wid="_1_d2e30" id="d1e26" pos="a" lemma="beag">bhig</w>
      <w wid="_1_d2e33" id="d1e30" pos="px" lemma="neo">neo</w>
      <pc join="both">-</pc>
      <w wid="_1_d2e41" id="d1e36" pos="a" lemma="euchdach">euchdaich</w>
      <pc join="left">:</pc>
      <lb/>
      <w wid="_1_d2e50" id="d1e45" pos="V" lemma="amhairc">amhairc</w>
      <w wid="_1_d2e55" id="d1e54" pos="A" lemma="a-steach">a-steach</w>
      <w wid="_1_d2e58" id="d1e61" pos="p" lemma="an">’na</w>
      <w wid="_1_d2e61" id="d1e64" pos="ns" lemma="sùil">sùilean</w>
      <w wid="_1_d2e65" id="d1e67" pos="a" lemma="leugach">leugach</w>
      <pc join="left">:</pc>
      <lb/>
      <w wid="_1_d2e75" id="d1e76" pos="z" lemma="ge">ge</w>
      <w wid="_1_d2e78" id="d1e79" pos="a" lemma="bacach">bacach</w>
      <w wid="_1_d2e84" id="d1e82" pos="D" lemma="thu">thu</w>
      <pc join="left">,</pc>
      <w wid="_1_d2e90" id="d1e88" pos="z" lemma="cha">chan</w>
      <w wid="_1_d2e93" id="d1e92" pos="v" lemma="bi">eil</w>
  </lg>
</text>
```

The initial query setup looks like:
```xquery
let $n := 7
let $word := doc("/db/meanma/example-1.xml")//w[@wid eq "_1_d2e50"]
let $following-n-words := $word/following-sibling::w[fn:position() le $n]
```

This expression returns no data:
```xquery
$word/following-sibling::node()[. << $following-n-words[last()]]
```

Refactoring it to this (to remove the inner predicate), does return data:
```xquery
let $last-following-n-word := $following-n-words[last()]
return
  $word/following-sibling::node()[. << $last-following-n-word]
```

Tested with eXist-db 7.0.0-SNAPSHOT (commit 097087bb020637b2c9d293d82522a2a5689aedc0)

### LINKED ISSUE #5432: [BUG] Register binary attribute in collection.xconf.xsd
**Describe the bug**

PR #4541 added support for indexing fields as binary values, controlled by a new `@binary` attribute on the `<field>` element. The PR neglected to register this new attribute in collection.xconf.xsd.

As a result, eXide, for example, flags uses of this attribute. See https://github.com/eXist-db/exist/issues/5431#issuecomment-2317631397.

**Expected behavior**

The xconf schema should define the attribute and its allowed values.

**To Reproduce**

Validate this document from the linked issue against collection.xconf.xsd:

```xml
    <!-- Full-text indexing with Lucene -->
    <lucene>
      <!-- Elements upon which to build an index. -->
      <text qname="div">
        <field name="sortable" expression="./test/string()" type="xs:string" binary="yes"/>
      </text>
    </lucene>
  </index>
</collection>
```

Via the linked issue:

> If I try to apply the collection.xconf with eXide this error occurs: 
> > Failed to apply configuration: DocValuesField "sortable" appears more than once in this document (only one value is allowed per field)

**Screenshots**

n/a

 - Build: eXist-6.2.0
 - Java: 1.8.0_422
 - OS: Ubuntu 22.04.4 LTS - Linux 6.8.0-40-generic amd64

**Additional context**
- How is eXist-db installed? n/a
- Any custom changes in e.g. `conf.xml`? n/a

joewiz:
### Description:

Closes #5432 by adding schema definitions for the `field` element's `binary` attribute in the `collection.xconf.xsd` schema. 

### Reference: 

#5432

### Type of tests:

n/a

### DIFF
--- a/schema/collection.xconf.xsd
+++ b/schema/collection.xconf.xsd
@@ -183,6 +183,7 @@
         <xs:attributeGroup ref="cc:typeOpt"/>
         <xs:attributeGroup ref="cc:analyzerOpt"/>
         <xs:attributeGroup ref="cc:storeOpt"/>
+        <xs:attributeGroup ref="cc:binaryOpt"/>
     </xs:complexType>
 
     <xs:complexType name="matchAttrBoostType">
@@ -287,6 +288,24 @@
     </xs:attributeGroup>
     <xs:attributeGroup name="attributeReq">
         <xs:attribute name="attribute" type="xs:NCName" use="required"/>
+    </xs:attributeGroup>
+    <xs:attributeGroup name="binaryOpt">
+        <xs:attribute name="binary" use="optional">
+            <xs:simpleType>
+                <xs:restriction base="xs:token">
+                    <xs:enumeration value="yes">
+                        <xs:annotation>
+                            <xs:documentation>Index as a binary field</xs:documentation>
+                        </xs:annotation>
+                    </xs:enumeration>
+                    <xs:enumeration value="no">
+                        <xs:annotation>
+                            <xs:documentation>Do not index as a binary field</xs:documentation>
+                        </xs:annotation>
+                    </xs:enumeration>
+                </xs:restriction>
+            </xs:simpleType>
+        </xs:attribute>
     </xs:attributeGroup>
     <xs:attributeGroup name="boostOpt">
         <xs:attribute name="boost" type="xs:double" use="optional" form="unqualified"/>


joewiz:
SonarCloud's failure:

> Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.20.0:site (default-site) on project exist: Failed to render site: Error generating maven-surefire-report-plugin:3.5.0:report report: Cannot invoke "Object.toString()" because "value" is null -> [Help 1]

The windows-latest checks haven't failed yet; they're still going, 30 minutes after submitting the PR...

dizzzz:
it is a recurring issue.... I guess a plugin update is needed...

dizzzz (approved):

duncdrum:
@joewiz did you mean issue to link to issue 5342 in the op, I corrected that as I don't see how that would be related

joewiz:
@duncdrum Thanks, yes! 

@dizzzz The `windows-latest` checks finished... after 40 minutes - wow!

adamretter (changes requested):
In principle this is fine, but I think perhaps we should just restrict the choices to "yes" and "no" so that it is consistent with the other elements in the schema.

line-o (commented):
Thanks @joewiz for taking this on. I second @adamretter's comment that the options should be limited to `yes` and `no` for consistency.
I am writing this knowing that collection configurations that include this option do set it to `true`. So we need to check first if `yes` will be recognised by the implementation and to change the documentation accordingly.

joewiz:
Thanks, all. I've removed `true` & `false` from the schema-allowed options and updated my PR branch. 

In [the issue](https://github.com/eXist-db/exist/issues/5432#issuecomment-2317738084) I linked to the implementation and its tests, which still respectively support and use `true` & `false`. I think those fixes should come in a separate PR. If that PR removes these values, it would constitute a breaking change, but of course the PR could also deprecate them, etc. 

(I think this PR still counts as a bugfix, since the old schema complained about the very presence of the `@binary` attribute.)

line-o (approved):

PR #5466, opened 2024-10-02
[6.x.x] Switch CI from Temurin to Liberica

adamretter:
Closes https://github.com/eXist-db/exist/pull/5304

Liberica should still provide JDK 8 on macOS

PR #5468, opened 2024-10-02
[4.x.x] Switch CI from Temurin to Liberica

adamretter:
Backport of https://github.com/eXist-db/exist/pull/5466

Liberica should still provide JDK 8 on macOS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,14 +10,14 @@ jobs:
         jdk: ['8']
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Set up JDK ${{ matrix.jdk }}
         uses: actions/setup-java@v4
         with:
           distribution: liberica
           java-version: ${{ matrix.jdk }}
       - name: Cache local Ivy repository
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: ~/.ivy/cache
           key: ${{ runner.os }}-ivy-${{ hashFiles('**/ivy.xml') }}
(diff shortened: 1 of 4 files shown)


dizzzz (commented):
Interesting build error....

   [junit] 1407 [WARN] DefaultCacheManager: The cacheSize="256" setting in conf.xml is too large. Java has only 466432k available. Cache manager will not use more than 155477k to avoid memory issues which may lead to database corruptions.
Error: nit] 1704 [ERROR] PluginsManagerImpl: java.lang.ClassNotFoundException: org.exist.storage.md.MDStorageManager

adamretter:
> WARN] DefaultCacheManager: The cacheSize="256" setting in conf.xml is too large. Java has only 466432k available. Cache manager will not use more than 155477k to avoid memory issues which may lead to database corruptions.

@dizzzz I think this one may be due to the limited amount of RAM available in CI and that we need not worry about it.

> Error: nit] 1704 [ERROR] PluginsManagerImpl: java.lang.ClassNotFoundException: org.exist.storage.md.MDStorageManager

@dizzzz I have appended a commit to change this from an Error into a Warning in the log.

dizzzz:
@adamretter could it be that a 32bit VM is used?

adamretter:
> @adamretter could it be that a 32bit VM is used?
> 

No. I don't think so

adamretter:
@dizzzz @reinhapa I believe this is now ready to merge please?

reinhapa:
@adamretter the windows build still fails...

reinhapa (approved):

adamretter:
> the windows build still fails. Wilk this be resolve with the folllow up restore output change?
@reinhapa I didn't have the Ant script quite correct for passing the property to disable the journal recovery log output. I have fixed that now, and CI is all passing now :-)

PR #5467, opened 2024-10-02
[5.x.x] Switch CI from Temurin to Liberica

adamretter:
Backport of https://github.com/eXist-db/exist/pull/5466

Liberica should still provide JDK 8 on macOS

PR #5469, opened 2024-10-02
Hide journal recovery messages during test suite run

adamretter:
Stops the CI logs from getting polluted

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -53,7 +53,7 @@ jobs:
         id: install-mvnd
         uses: ./.github/actions/install-mvnd
         with:
-          version: '1.0.1'
+          version: '1.0.2'
           file-version-suffix: ''
           cache: 'true'
       - name: Maven Build
--- a/exist-core/src/main/java/org/exist/storage/recovery/RecoveryManager.java
+++ b/exist-core/src/main/java/org/exist/storage/recovery/RecoveryManager.java
@@ -149,7 +149,7 @@ public class RecoveryManager {
                     // if the last checkpoint record is not the last record in the file
                     // we need a recovery.
                     if ((lastCheckpoint == null || !lastCheckpoint.getLsn().equals(lastLsn)) &&
-                            txnsStarted.size() > 0) {
+                        !txnsStarted.isEmpty()) {
                         LOG.info("Dirty transactions: {}", txnsStarted.size());
                         // starting recovery: reposition the log reader to the last checkpoint
                         if (lastCheckpoint == null) {
@@ -298,7 +298,7 @@ public class RecoveryManager {
                 {
                     LOG.info("Second pass: undoing dirty transactions. Uncommitted transactions: {}", runningTxns.size());}
             // see if there are uncommitted transactions pending
-            if (runningTxns.size() > 0) {
+            if (!runningTxns.isEmpty()) {
                 // do a reverse scan of the log, undoing all uncommitted transactions
                 try {
                     final long lastSize = FileUtils.sizeQuietly(last);
@@ -307,9 +307,10 @@ public class RecoveryManager {
                         if (next.getLogType() == LogEntryTypes.TXN_START) {
                             if (runningTxns.get(next.getTransactionId()) != null) {
                                 runningTxns.remove(next.getTransactionId());
-                                if (runningTxns.size() == 0)
+                                if (runningTxns.isEmpty()) {
                                     // all dirty transactions undone
-                                    {break;}
+                                    break;
+                                }
                             }
                         } else if (next.getLogType() == LogEntryTypes.TXN_COMMIT) {
                             // ignore already committed transaction
--- a/exist-core/src/main/java/org/exist/storage/recovery/RecoveryManager.java
+++ b/exist-core/src/main/java/org/exist/storage/recovery/RecoveryManager.java
@@ -120,10 +120,10 @@ public class RecoveryManager {
                     Lsn lastLsn = Lsn.LSN_INVALID;
                     Loggable next;
                     try {
-                        final ProgressBar progress = new ProgressBar("Scanning journal ", FileUtils.sizeQuietly(last));
+                        final long lastSize = FileUtils.sizeQuietly(last);
+                        final ProgressBar scanProgressBar = new ProgressBar("Scanning journal ", lastSize);
                         while ((next = reader.nextEntry()) != null) {
 //	                        LOG.debug(next.dump());
-                            progress.set(next.getLsn().getOffset());
                             if (next.getLogType() == LogEntryTypes.TXN_START) {
                                 // new transaction starts: add it to the transactions table
                                 txnsStarted.put(next.getTransactionId(), next);
@@ -135,7 +135,10 @@ public class RecoveryManager {
                                 lastCheckpoint = (Checkpoint) next;
                             }
                             lastLsn = next.getLsn();
+
+                            scanProgressBar.set(next.getLsn().getOffset());
                         }
+                        scanProgressBar.set(lastSize);  // 100%
                     } catch (final LogException e) {
                         if (LOG.isDebugEnabled()) {
                             LOG.debug("Caught exception while reading log", e);
@@ -250,10 +253,11 @@ public class RecoveryManager {
             if (LOG.isInfoEnabled())
                 {
                     LOG.info("First pass: redoing {} transactions...", txnCount);}
-            final ProgressBar progress = new ProgressBar("Redo ", FileUtils.sizeQuietly(last));
             Loggable next = null;
             int redoCnt = 0;
             try {
+                final long lastSize = FileUtils.sizeQuietly(last);
+                final ProgressBar redoProgressBar = new ProgressBar("Redo ", lastSize);
                 while ((next = reader.nextEntry()) != null) {
                     SanityCheck.ASSERT(next.getLogType() != LogEntryTypes.CHECKPOINT,
                             "Found a checkpoint during recovery run! This should not ever happen.");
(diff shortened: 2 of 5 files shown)


exist-core/src/main/java/org/exist/storage/recovery/RecoveryManager.java
```diff
@@ -135,6 +139,14 @@ public boolean recover() throws LogException {
                                 lastCheckpoint = (Checkpoint) next;
                             }
                             lastLsn = next.getLsn();
+
```
dizzzz, line 142: I was thinking of this, great catch!

.github/workflows/sonarcloud.yml
```diff
@@ -35,4 +35,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-        run: mvn -V -B -Dsurefire.useFile=false -DtrimStackTrace=false -Ddependency-check.skip=true -Ddocker=false -P \!mac-dmg-on-mac,\!codesign-mac-dmg,\!mac-dmg-on-unix,\!installer,\!concurrency-stress-tests,\!micro-benchmarks,\!build-dist-archives verify site org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
+        run: mvn -V -B -Dsurefire.useFile=false -DtrimStackTrace=false -Ddependency-check.skip=true -Ddocker=false -P \!mac-dmg-on-mac,\!codesign-mac-dmg,\!mac-dmg-on-unix,\!installer,\!concurrency-stress-tests,\!micro-benchmarks,\!build-dist-archives verify site org.sonarsource.scanner.maven:sonar-maven-plugin:4.0.0.4121:sonar
```
dizzzz, line 38: is it an good idea to have this versionnr hardcoded in the scripts?

dizzzz (commented):
It looks like the PR contains content of another PR?

.github/workflows/ci-deploy.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
```
reinhapa, line 16: How does a different Java implementation impact the journal recovery output?

.github/workflows/ci-test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
```
reinhapa, line 16: How does a different Java implementation impact the journal recovery output?

.github/workflows/ci-test.yml
```diff
@@ -25,7 +25,7 @@ jobs:
...
```
reinhapa, line 28: How does a different Java implementation impact the journal recovery output?

.github/workflows/ci-test.yml
```diff
@@ -46,14 +46,14 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
```
reinhapa, line 49: How does a different Java implementation impact the journal recovery output?

.github/workflows/ci-xqts.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
```
reinhapa, line 15: How does a different Java implementation impact the journal recovery output?

.github/workflows/sonarcloud.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
```
reinhapa, line 20: How does a different Java implementation impact the journal recovery output?

reinhapa (approved):
Great enhancement to have less output during tests!

.github/workflows/ci-deploy.yml
dizzzz: the PR combines 2 PRs in essence; the change is in the recoverymanager

.github/workflows/sonarcloud.yml
adamretter: It was using a really old version before by default. The only way I could see to get it to use the latest was to add it to the command here.

adamretter:
@dizzzz @reinhapa I kept the journal improvements and CI improvements in a single PR, as that is what I had already done for 4.x.x, 5.x.x, and 6.x.x; albeit in those cases I needed to do both in a single PR to be able to see what was going on. If you feel this PR here must be split into two - let me know and I will do so...

.github/workflows/sonarcloud.yml
adamretter: @dizzzz I have managed to move the version into `exist-parent/pom.xml` now, which means it should be updated by dependabot etc.

dizzzz:
It is OK for me, as it is intentional

dizzzz (approved):

duncdrum:
I m also curious what the JDK has to do with it. Since we use temurin not just on CI here, but also in Docker and all other repos in the org, I m not a fan of changing it without reason. If there is a reason to change, we should stay consistent elsewhere in the org.

reinhapa:
I'm ok changing the JDK for the sake of having the old versions, but on the `develop` branch I would prefer to leave it using the temurin one...

dizzzz:
It looks like that switching to the JVM Adam suggested all kind of strange test/fork/... issues occur.

Looking at the discussions it is probably better to split the jvm change and the recovery log code delta.

reinhapa:
> It looks like that switching to the JVM Adam suggested all kind of strange test/fork/... issues occur.
> 
> Looking at the discussions it is probably better to split the jvm change and the recovery log code delta.

We then at least can rule out side effects from the JDK change... Just to be clear I'm full in the hiding suff!

line-o (changes requested):
I do appreciate hiding the journal progress bar in CI!
I would like to see as @dizzz and @reinhapa  that only the changes necessary for that to work would be included in this PR.
I would also like to see them in just one or two commits.

adamretter:
> I m also curious what the JDK has to do with it.

@duncdrum (cc @reinhapa) As I wrote before you posted, see: "as that is what I had already done for 4.x.x, 5.x.x, and 6.x.x"

> Since we use temurin not just on CI here, but also in Docker and all other repos in the org, I m not a fan of changing it without reason.

@duncdrum We are now using Liberica (and not Temurin) for eXist-db [4.x.x](https://github.com/eXist-db/exist/pull/5468), [5.x.x](https://github.com/eXist-db/exist/pull/5467), and [6.x.x](https://github.com/eXist-db/exist/pull/5466). I am also a fan of consistency and so I wanted to use Liberica also for 7.x.x in THIS repo.
Regards Docker, I think you may be mistaken, we are not using Temurin and we never have:

1. For eXist-db 7.x.x in Docker we are using OpenJDK (from `gcr.io/distroless/java17:latest` base image), see: https://github.com/GoogleContainerTools/distroless/blob/main/java/README.md
2. For eXist-db 6.x.x in Docker we are using OpenJDK, see: https://github.com/eXist-db/exist/blob/develop-6.x.x/exist-docker/src/main/resources-filtered/Dockerfile#L28
3. For eXist-db 5.x.x in Docker we are using OpenJDK, see: https://github.com/eXist-db/exist/blob/develop-5.x.x/exist-docker/src/main/resources-filtered/Dockerfile#L28
4. For eXist-db 4.x.x in Docker we are using OpenJDK, see: https://github.com/eXist-db/docker-existdb/blob/master/Dockerfile#L28

>  If there is a reason to change

Yes, the features of eXist-db 4.x.x, 5.x.x, and 6.x.x only reliably run 100% on Java 8. Whilst you can run some parts of eXist-db 4.x.x, 5.x.x, and 6.x.x on newer Java, you will find various parts that throw exceptions due to changes in the JDK after Java 8; JAXB, Mail, Activation, Image, etc.

There does not appear to be a Temurin build available for Java 8 in macOS (at least in CI) so we simply can't use Temurin in CI, however there is one for Liberica. That is the reason for the change.

I see a reason to drop Temurin, but no reason not to use Liberica (which is also based on OpenJDK).

> we should stay consistent elsewhere in the org.

I am not so concerned about other repos in this org as they each have different requirements that I expect to be set by the owners of those repos.

As the JDK change seems to be of concern, I will move it to a separate PR. However I think that we should choose one JDK vendor across 4.x.x, 5.x.x, 6.x.x, and 7.x.x and stick to it in this repo. As Temurin won't work for us (no JDK 8 support on macOS in CI) we need to choose one of the others.

.github/workflows/ci-deploy.yml
adamretter: Moved JDK change to PR - https://github.com/eXist-db/exist/pull/5480

.github/workflows/ci-test.yml
adamretter: https://github.com/eXist-db/exist/pull/5480

.github/workflows/ci-test.yml
adamretter: Moved JDK change to PR - https://github.com/eXist-db/exist/pull/5480

.github/workflows/ci-xqts.yml
adamretter: Moved JDK change to PR - https://github.com/eXist-db/exist/pull/5480

.github/workflows/ci-test.yml
adamretter: Moved JDK change to PR - https://github.com/eXist-db/exist/pull/5480

.github/workflows/sonarcloud.yml
adamretter: Moved JDK change to PR - https://github.com/eXist-db/exist/pull/5480

adamretter:
@dizzzz @reinhapa I have moved JDK change out of this PR and into a new PR - https://github.com/eXist-db/exist/pull/5480

PR #5513, opened 2024-10-22
[6.x.x] Fix CI Windows stream corruption

adamretter:
* Remove JFR from CI
* Use Windows specific CI profile for testing in CI

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -20,6 +20,7 @@ jobs:
         timeout-minutes: 60
   dependencies:
     name: Dependency checks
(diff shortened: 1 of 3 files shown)


dizzzz (approved):
somehow still issues during builds, randomly. 
Have seen them before...

reinhapa:
For me it would be better that we would try to only change one thing at the time in order to know what acctually solves the build failures. As the situation represents itself right now, it seems that the current changes sadly did not yield the expected results so far.

reinhapa (approved):

dizzzz (approved):

PR #5529, opened 2024-10-31
Allow module imports in one-off xqueries

### LINKED ISSUE #5525: [BUG] 'other' is different type of Path
**Describe the bug**
With a simple query, be it via Java Admin, via rest, or in eXide, one gets a `'other' is different type of Path` 
error under certain conditions.

**Expected behavior**
Getting the result of the query.

**To Reproduce**
Open eXide -> new XQuery. Query a collection of XML document with a simple for loop importing the `functx` module.

```xquery
xquery version "3.1";

import module namespace functx = "http://www.functx.com";

for $x in collection("/db/apps/yourfolder")//tei:TEI
return $x
```

Copy-pasting the same query in the Java Admin console returns the same error.

Full error:

```txt
org.xmldb.api.base.XMLDBException: Failed to invoke method compile in class org.exist.xmlrpc.RpcConnection: 'other' is different type of Path
	at org.exist.xmldb.RemoteCollection.execute(RemoteCollection.java:129)
	at org.exist.xmldb.RemoteXPathQueryService.compileAndCheck(RemoteXPathQueryService.java:151)
	at org.exist.xmldb.RemoteXPathQueryService.compile(RemoteXPathQueryService.java:128)
	at org.exist.client.QueryDialog$QueryRunnable.run(QueryDialog.java:577)
	at java.lang.Thread.run(Thread.java:750)
Caused by: org.apache.xmlrpc.XmlRpcException: Failed to invoke method compile in class org.exist.xmlrpc.RpcConnection: 'other' is different type of Path
	at org.apache.xmlrpc.client.XmlRpcStreamTransport.readResponse(XmlRpcStreamTransport.java:214)
	at org.apache.xmlrpc.client.XmlRpcStreamTransport.sendRequest(XmlRpcStreamTransport.java:173)
	at org.apache.xmlrpc.client.XmlRpcHttpTransport.sendRequest(XmlRpcHttpTransport.java:142)
	at org.apache.xmlrpc.client.XmlRpcSunHttpTransport.sendRequest(XmlRpcSunHttpTransport.java:70)
	at org.apache.xmlrpc.client.XmlRpcClientWorker.execute(XmlRpcClientWorker.java:56)
	at org.apache.xmlrpc.client.XmlRpcClient.execute(XmlRpcClient.java:167)
	at org.apache.xmlrpc.client.XmlRpcClient.execute(XmlRpcClient.java:158)
	at org.apache.xmlrpc.client.XmlRpcClient.execute(XmlRpcClient.java:147)
	at org.exist.xmldb.RemoteCollection.execute(RemoteCollection.java:127)
	... 4 more
```

 - Build: [eXist-6.3.0]
 - Java: openjdk 17.0.13 2024-10-15
 - Debian

**Additional context**
- How is eXist-db installed? .jar

### LINKED ISSUE #5530: [BUG] one-off queries cannot import modules with location hint
**Describe the bug**

One-off queries that import library modules with a location hint (`import module namespace ... at ...;`) raise error `err:XQST0059` that the source could not be found regardless if the library module is available at the location or not.

**Expected behavior**

One-off queries evaluated in exist-db must allow imports of library modules with a location hint.

**To Reproduce**

1. Evaluate the any query importing a library with a location hint in eXide or over XML-RPC (eg. `xst run`)

    ```xquery
    import module namespace packages="http://exist-db.org/apps/existdb-packages" at "/db/apps/packageservice/modules/packages.xqm";
    packages:get-local-packages()"
    ```

    ```bash
    xst run "import module namespace functx='http://www.functx.com' at '/db/system/repo/functx-1.0.1/functx/functx.xq';functx:atomic-type(4)"
    ```
2. The query will raise error err:XQST0059 that the module could not be found
3. running 
    ```bash
    xst run "util:binary-doc-available('/db/system/repo/functx-1.0.1/functx/functx.xq')"
   ```
4. returns `true()`

- Build: eXist-7.0.0-SNAPSHOT (9f881d7442457fca3918814574ff88a64667aefa)
- Java: 17.0.6 (Azul Systems, Inc.)
- OS: Mac OS X 14.5 (aarch64)

**Additional context**

line-o:
### Description:

- add modified conf.xml ModuleImport tests
- add tests for one-off queries with module imports
  - using a modified conf.xml that loads functx into package repo via auto deploy
  - of a registered module without location hint
  - of a module with location hint
- change XQueryContext to allow imports again
- change SourceFactory to work with contextPath set to "."

### Reference:

fixes #5525
fixes #5530

### Type of tests:

Java tests added in src/test/java/org/exist/xquery/ModuleImportTest.java

### DIFF
--- a/exist-core/src/main/java/org/exist/source/SourceFactory.java
+++ b/exist-core/src/main/java/org/exist/source/SourceFactory.java
@@ -111,7 +111,7 @@ public class SourceFactory {
                 && ((location.startsWith("/db") && !Files.exists(Paths.get(firstPathSegment(location))))
                 || (contextPath != null && contextPath.startsWith("/db") && !Files.exists(Paths.get(firstPathSegment(contextPath)))))) {
             final XmldbURI pathUri;
-            if (contextPath == null) {
+            if (contextPath == null || ".".equals(contextPath)) {
                 pathUri = XmldbURI.create(location);
             } else {
                 pathUri = XmldbURI.create(contextPath).append(location);
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -587,33 +587,34 @@ public class XQueryContext implements BinaryValueManager, Context {
             return null;
         }
 
-        // use the resolved file or return null
-        if (resolved != null) {
-
-            String location = "";
-
-            try {
-
-                // see if the src exists in the database and if so, use that instead
-                Source src = repo.get().resolveStoredXQueryModuleFromDb(getBroker(), resolved);
-                if (src != null) {
-                    // NOTE(AR) set the location of the module to import relative to this module's load path - so that transient imports of the imported module will resolve correctly!
-                    location = Paths.get(XmldbURI.create(moduleLoadPath).getCollectionPath()).relativize(Paths.get(((DBSource)src).getDocumentPath().getCollectionPath())).toString();
+        // use the resolved file
+        String location = "";
+        try {
+            // see if the src exists in the database and if so, use that instead
+            Source src = repo.get().resolveStoredXQueryModuleFromDb(getBroker(), resolved);
+            if (src == null) {
+                // fallback to load the source from the filesystem
+                src = new FileSource(resolved, false);
+            } else {
+                final String sourceCollection = ((DBSource)src).getDocumentPath().getCollectionPath();
+                if (".".equals(moduleLoadPath)) {
+                    // module is a string passed to the xquery context, has therefore no location of its own
+                    location = sourceCollection;
                 } else {
-                    // else, fallback to the one from the filesystem
-                    src = new FileSource(resolved, false);
+                    // NOTE(AR) set the location of the module to import relative to this module's load path
+                    // - so that transient imports of the imported module will resolve correctly!
+                    final Path collectionPath = Paths.get(XmldbURI.create(moduleLoadPath).getCollectionPath());
+                    final Path sourcePath = Paths.get(sourceCollection);
+                    location = collectionPath.relativize(sourcePath).toString();
                 }
-
-                // build a module object from the source
-                final ExternalModule module = compileOrBorrowModule(namespace, prefix, location, src);
-                return module;
-
-            } catch (final PermissionDeniedException e) {
-                throw new XPathException(e.getMessage(), e);
             }
-        }
 
-        return null;
+            // build a module object from the source
+            return compileOrBorrowModule(namespace, prefix, location, src);
+
+        } catch (final PermissionDeniedException | IllegalArgumentException e) {
+            throw new XPathException(e.getMessage(), e);
+        }
     }
 
     /**
--- a/exist-core/src/test/java/org/exist/xquery/ModuleImportTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/ModuleImportTest.java
@@ -1,0 +1,174 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery;
+
+import com.evolvedbinary.j8fu.Either;
+
+import org.exist.EXistException;
+import org.exist.security.PermissionDeniedException;
+import org.exist.storage.BrokerPool;
+import org.exist.storage.DBBroker;
+import org.exist.test.ExistEmbeddedServer;
+import org.exist.xquery.value.Sequence;
+import org.exist.xquery.value.StringValue;
+
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import static com.evolvedbinary.j8fu.Either.Left;
+import static com.evolvedbinary.j8fu.Either.Right;
+import static com.ibm.icu.impl.Assert.fail;
+import static org.exist.test.XQueryAssertions.assertThatXQResult;
+import static org.exist.test.XQueryAssertions.assertXQStaticError;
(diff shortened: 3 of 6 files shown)


exist-core/src/main/java/org/exist/source/SourceFactory.java
```diff
@@ -111,7 +111,7 @@ public class SourceFactory {
                 && ((location.startsWith("/db") && !Files.exists(Paths.get(firstPathSegment(location))))
                 || (contextPath != null && contextPath.startsWith("/db") && !Files.exists(Paths.get(firstPathSegment(contextPath)))))) {
             final XmldbURI pathUri;
-            if (contextPath == null) {
+            if (contextPath == null || ".".equals(contextPath)) {
```
adamretter, line 114: I don't want to upset anyone, but I just wanted to point out that this is the wrong approach to fixing this issue.

line-o: This is fixing an additional bug I found while creating tests:

see `importLibraryFromDbLocation` and `importLibraryFromXMLDBLocation`

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -587,33 +587,34 @@ public Optional<ExistRepository> getRepository() {
...
+        String location = "";
+        try {
+            // see if the src exists in the database and if so, use that instead
+            Source src = repo.get().resolveStoredXQueryModuleFromDb(getBroker(), resolved);
+            if (src == null) {
+                // fallback to load the source from the filesystem
+                src = new FileSource(resolved, false);
+            } else {
+                final String sourceCollection = ((DBSource)src).getDocumentPath().getCollectionPath();
+                if (moduleLoadPath.equals(".")) {
```
dizzzz, line 600: better to swap left/right

line-o:
Adding the functx library to auto deploy in the conf.xml used for all tests leads to test failures.
There are, however, other configurations that do have functx loaded, see src/test/resources-filtered/org/exist/xquery/functions/transform/conf.xml.

Should I, either
1. limit the application of the conf.xml __with__ functx to only xquery tests
2. limit the application of the conf.xml __with__ functx to only moduleImport tests
3. try to import other modules that are present without adding functx

joewiz:
If the issue isn't functx-specific, then choice 3 sounds like it would address the issue.

exist-core/src/main/java/org/exist/source/SourceFactory.java
adamretter: Yes, I understood that, but it still is the wrong approach.

adamretter:
You should not have to modify conf.xml for this.

exist-core/src/main/java/org/exist/source/SourceFactory.java
reinhapa: > Yes, I understood that, but it still is the wrong approach.

@adamretter Could you explain why you think this is the wrong approach and how do you think it should be solved correctly?

line-o:
@joewiz module imports are certainly not functx specific. 
For the tests to work we need do need to add a pure library package to exist. It just happens that the functx-1.0.1.xar is already part of the test fixtures in 
`src/test/resources/functx/functx-1.0.1.xar`.
With that we have a library module that can be imported **with** and **without** location hint.

So far as I can see there is no other way than to add another configuration with functx added to autodeploy.
We could change both `src/test/java/org/exist/xquery/functions/transform/TransformFromPkgTest.java` and `src/test/java/org/exist/xquery/ModuleImportTest.java` test classes to both use a `conf-with-functx.xml` to make its use-case more obvious.

line-o:
I ended up using option 2, only use conf.xml with functx loaded via autodeploy for ModuleImportTest.

exist-core/src/test/java/org/exist/xquery/ModuleImportTest.java
```diff
@@ -0,0 +1,162 @@
...
+        try {
+            return Paths.get(loader.getResource(packagePath + separator + "conf.xml").toURI());
+        } catch (final URISyntaxException e) {
+            fail(e);
+            return null;
+        }
+    }
+
+    @ClassRule
+    public static ExistEmbeddedServer server = new ExistEmbeddedServer(null, getConfigFile(), null, false, true);
```
reinhapa, line 93: Please move the field declaration to the top of the file to resolve this Codacy issue...

reinhapa (changes requested):

exist-core/src/test/java/org/exist/xquery/ModuleImportTest.java
line-o: Will do

exist-core/src/test/resources-filtered/conf.xml
```diff
@@ -768,7 +777,7 @@
             Set the parameter to "yes" to enable this feature.
 
     -->
-    <serializer add-exist-id="none" compress-output="no" output-doctype="yes" enable-xinclude="yes"
```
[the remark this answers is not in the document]
line-o: I'll re-add this.

exist-core/src/main/java/org/exist/source/SourceFactory.java
line-o: `SourceFactory.getSource` is called with contextPath set to `""`, `"."` and `"/"` throughout the codebase in exist-db.
`"."` being the **default** value for `XQueryContext.moduleLoadPath`.
Its unit tests do not test for any of these (see https://github.com/eXist-db/exist/blob/develop/exist-core/src/test/java/org/exist/source/SourceFactoryTest.java).

line-o: From all that I know the SourceFactory would like to get a `null` for contextPath but it isn't getting that except for a few call-sites where that is hard-coded.

adamretter: The null is significant

line-o: I agree

exist-core/src/test/java/org/exist/xquery/ModuleImportTest.java
```diff
@@ -0,0 +1,162 @@
...
+        final String query = "import module namespace functx='http://www.functx.com'" +
+                " at 'unknown:///db/system/repo/functx-1.0.1/functx/functx.xq';" +
+                "functx:atomic-type(4)";
+        final String expectedMessage = "error found while loading module functx: Source for module 'http://www.functx.com' not found module location hint URI 'unknown:///db/system/repo/functx-1.0.1/functx/functx.xq'.";
+
+        assertXQStaticError(ErrorCodes.XQST0059, -1,-1, expectedMessage, compileQuery(query));
+    }
+
+    @Test
+    public void importLibraryFromRelativeLocation() throws EXistException, PermissionDeniedException {
```
dizzzz, line 153: this could help on the exist://  urls 

https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/protocolhandler/URLStreamHandlerStartupTrigger.java

dizzzz (approved):

reinhapa (approved):

line-o:
@dizzzz I will see if I can address all issue Sonar complains about. 
I think that "Replace these 3 tests with a single Parameterized one." might not work, as I think I tried it already and it ended up having some race condition. But I might be confusing this with my other PR. Let's see.

dizzzz:
Addressing the imports is sufficient (as discussed Monday)

line-o:
Note to self: Add a note to the source code why this special handling was added: "ContextPath should be null but isn't; find a way to do that instead"

Issue #5450, opened 2024-09-25
[BUG] CICD deploy has been failing on `develop` for the past 7 months


duncdrum:
https://github.com/eXist-db/exist/actions/runs/7955959388 was the first failing run. 

We are attempting to push to ghcr.io , while the push should go to docker hub

duncdrum:
the problem is fabric8io/docker-maven-plugin#378 . Because we are pulling a base image from google container registry, the plugin now always attempts a push to the same registry.

PR #5533, opened 2024-11-02
Feat-multi-arch-containers

### LINKED ISSUE #4151: [feature] multi-arch docker images
to better support arm processors, such as Apple Silicon, but also in cloud. We should produce native images for each platform. While `linux/amd64, linux/arm64` are the most important native images we could potentially add more architectures, provided that there is actual demand and use cases for them. 

There are two problems preventing us from simply doing so. 
The first is that the maven plugin [#1267](https://github.com/fabric8io/docker-maven-plugin/issues/1267) doesn't support buildx.  So for maven to spit out an ARM64 image at all, it needs to run on an ARM chip. There doesn't seem to be an easy way to get our hands on one with the current CI provider. 

The second problem is the that changes to the `Dockerfile` are required. There are architecture specific paths in the builder stage. That being said, all images used by us are already available as multi-arch. 

Some more research is necessary, but as of now our options are :
1. to ditch maven as intermediary for docker on CI, and publish multi-arch images without running on actual ARM chips, 
1. switch CI provider or find a way to use GitHub Actions on ARM [for free](https://github.com/actions/virtual-environments/issues/2187) or, 
1. to replace the current  maven-plugin with an [alternative plugin](https://github.com/GoogleContainerTools/jib) that supports v2 manifest files required by buildx, 

Happy to hear thoughts and suggestions, but these are in my current order of preference based on work overall feasibility and work required.

### LINKED ISSUE #4153: does this need to be there ?
https://github.com/eXist-db/exist/blob/ca91ad0d237f3b5d7963dbc54d0d4388fc7a1ebd/exist-docker/pom.xml#L188

### LINKED ISSUE #5532: [CI] Change docker setup
Create docker containers that do run on arm architectures as well

see duncdrum/exist-db images

duncdrum:
enables a long overdue feature: multi-arch containers for `existdb/existdb` 
backport will remove a release blocker for 6.3.1

sidessteps docker maven plugin bug, `deploy` is green again.
see fabric8io/docker-maven-plugin#1835

fixes outdated JRE patches, bumps base image to Debian 12 
see #5450
close #4151
close #5532
close #4153

### DIFF
--- a/exist-docker/src/main/resources-filtered/Dockerfile
+++ b/exist-docker/src/main/resources-filtered/Dockerfile
@@ -48,10 +48,10 @@ ARG CACHE_MEM
 ARG MAX_BROKER
 ARG JVM_MAX_RAM_PERCENTAGE
 
-ENV EXIST_HOME "/exist"
+ENV EXIST_HOME=/exist
 ENV CLASSPATH=/exist/lib/${exist.uber.jar.filename}
 
-ENV JAVA_TOOL_OPTIONS \
+ENV JAVA_TOOL_OPTIONS="\
   -Dfile.encoding=UTF8 \
   -Dsun.jnu.encoding=UTF-8 \
   -Djava.awt.headless=true \
@@ -67,7 +67,7 @@ ENV JAVA_TOOL_OPTIONS \
   -XX:+UseStringDeduplication \
   -XX:+UseContainerSupport \
   -XX:MaxRAMPercentage=${JVM_MAX_RAM_PERCENTAGE:-75.0} \
-  -XX:+ExitOnOutOfMemoryError
+  -XX:+ExitOnOutOfMemoryError"
 
 HEALTHCHECK CMD [ "java", \
     "org.exist.start.Main", "client", \
--- a/exist-docker/src/main/resources-filtered/Dockerfile
+++ b/exist-docker/src/main/resources-filtered/Dockerfile
@@ -20,25 +20,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #
 
-# Use Debian Bullseye (which is the base of gcr.io/distroless/java:17) for additional library dependencies that we need
-# FROM debian:bullseye-slim as debian-slim
-# RUN apt-get update && apt-get -y dist-upgrade
-# RUN apt-get install -y openjdk-17-jre-headless
-# RUN apt-get install -y expat fontconfig     # Install tools required by FOP
-
 FROM gcr.io/distroless/java17:latest
-
-# Copy over dependencies for Apache FOP, missing from GCR's JRE
-# COPY --from=debian-slim /usr/lib/x86_64-linux-gnu/libfreetype.so.6 /usr/lib/x86_64-linux-gnu/libfreetype.so.6
-# COPY --from=debian-slim /usr/lib/x86_64-linux-gnu/liblcms2.so.2 /usr/lib/x86_64-linux-gnu/liblcms2.so.2
-# COPY --from=debian-slim /usr/lib/x86_64-linux-gnu/libpng16.so.16 /usr/lib/x86_64-linux-gnu/libpng16.so.16
-# COPY --from=debian-slim /usr/lib/x86_64-linux-gnu/libfontconfig.so.1 /usr/lib/x86_64-linux-gnu/libfontconfig.so.1
-
-# Copy dependencies for Apache Batik (used by Apache FOP to handle SVG rendering)
-# COPY --from=debian-slim /etc/fonts /etc/fonts
-# COPY --from=debian-slim /lib/x86_64-linux-gnu/libexpat.so.1 /lib/x86_64-linux-gnu/libexpat.so.1
-# COPY --from=debian-slim /usr/share/fontconfig /usr/share/fontconfig
-# COPY --from=debian-slim /usr/share/fonts/truetype/dejavu /usr/share/fonts/truetype/dejavu
 
 # Copy eXist-db
 COPY LICENSE /exist/LICENSE
--- a/exist-docker/src/main/resources-filtered/Dockerfile
+++ b/exist-docker/src/main/resources-filtered/Dockerfile
@@ -20,7 +20,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #
 
-FROM gcr.io/distroless/java17:latest
+FROM gcr.io/distroless/java17-debian12:latest
 
 # Copy eXist-db
 COPY LICENSE /exist/LICENSE
--- a/exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
+++ b/exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
@@ -20,7 +20,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #
 
-# Use JDK 17 in Debian Bullseye (as our production image gcr.io/distroless/java:17 is based on Debian Bullseye with just a JRE)
+# Use JDK 17 in Debian as our production image gcr.io/distroless/java is based on Debian with just a JRE)
 FROM debian:bookworm-slim
 RUN apt-get update && apt-get -y dist-upgrade
 RUN apt-get install -y openjdk-17-jdk-headless
--- a/exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
+++ b/exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
@@ -52,10 +52,10 @@ ARG MAX_BROKER
 ARG JVM_MAX_RAM_PERCENTAGE
 ARG JVM_JDWP_SUSPEND
 
-ENV EXIST_HOME "/exist"
+ENV EXIST_HOME=/exist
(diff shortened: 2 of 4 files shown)


duncdrum:
The codacy issue here is pointless. We are trying to generate moving tag `latest` images. So they are based on `latest` base images from upstream on purpose

dizzzz:
Ubuntu build fails, any idea ?
Don't have a laptop around to check the logs ...

duncdrum:
@dizzzz some timeout with building distribution packages, I restarted it. Pretty sure it's unrelated to the changes in here.

reinhapa (approved):

line-o (commented):
@duncdrum is this good to go? Should we wait for the back port to 6.x.x?

duncdrum:
@line-o this is good to go. There are problems with the back port, but I m 99,9% certain that these will not affect the contents of this PR, which also fixes our broken ci.

line-o:
@duncdrum would you have the time to edit the commit message  of [edfc10e](https://github.com/eXist-db/exist/pull/5533/commits/edfc10ee74d80991e1b1a5ecb9db8423f640c4f2)

to
```
[bugfix] workaround to deploy images to dockerhub

fixes #5450

some explanation what the workaround does
```

If you do not have time I would volunteer to do this.

duncdrum:
@line-o go ahead I ll be afk a lot this week. Thanks for the help

exist-docker/src/main/resources-filtered/Dockerfile
```diff
@@ -20,25 +20,7 @@
...
 
-# Use Debian Bullseye (which is the base of gcr.io/distroless/java:17) for additional library dependencies that we need
-# FROM debian:bullseye-slim as debian-slim
-# RUN apt-get update && apt-get -y dist-upgrade
-# RUN apt-get install -y openjdk-17-jre-headless
-# RUN apt-get install -y expat fontconfig     # Install tools required by FOP
-
-FROM gcr.io/distroless/java17:latest
-
-# Copy over dependencies for Apache FOP, missing from GCR's JRE
```
adamretter, line 31: This change breaks XSL:FO in eXist-db when using Apache FOP. Unfortunately, the `gcr.io/distroless/java17-debian12:latest` image you have chosen to update to does not include these libraries. Please test a proper FOP transformation with these changes, and you will see the error.

exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
```diff
@@ -21,7 +21,7 @@
 #
 
 # Use JDK 17 in Debian Bullseye (as our production image gcr.io/distroless/java:17 is based on Debian Bullseye with just a JRE)
```
adamretter, line 23: I think this comment is out of date if you are switching the non-debug build to bookworm also.

exist-docker/src/main/resources-filtered/Dockerfile
```diff
@@ -85,7 +67,7 @@ ENV JAVA_TOOL_OPTIONS \
   -XX:+UseStringDeduplication \
```
adamretter, line 67: Isn't `-XX:+UseStringDeduplication` invalid on JDK 17?

adamretter (changes requested):

exist-docker/src/main/resources-filtered/Dockerfile
reinhapa: This option is still valid see the [JDK 24](https://docs.oracle.com/en/java/javase/23/docs/specs/man/java.html)  Java command options, but only useful in conjunction with the G1 garbage collector... (`-XX:+UseG1GC`)

reinhapa: > @reinhapa Yes you are correct. However, we are using the ZGC garbage collector, and it is not supported there and prints a warning.

@adamretter in this case the option `-XX:+UseStringDeduplication` is indeed useless and will be ignored...

exist-docker/src/main/resources-filtered/Dockerfile
line-o: This PR switches from  `gcr.io/distroless/java17:latest` to `gcr.io/distroless/java17-debian12:latest`
The lines stating they would be needed for FOP are commented out (lines 24-27 and 32-41) and select a different base image `debian:bullseye-slim`.
So, even if those additional libraries were needed for PDF-generation using XSL:FO they are already missing and nothing is changed from the status quo.  
It, therefore, would be beneficial to create an additional Dockerfile that has all that is needed for this specific use-case. This could also be provided by a third party.
This is all just under the pretense that this is really needed for PDF generation and the adoption rate would also show us how common it is.

exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
line-o: Yeah, better remove the name of a specific build of Debian from the comment, otherwise we will need to change it everytime the base image changes

exist-docker/src/main/resources-filtered/Dockerfile
duncdrum: If they are no longer included that's a bug likely introduced when moving to debian 12. see https://github.com/GoogleContainerTools/distroless/issues/321 or https://github.com/GoogleContainerTools/distroless/issues/789 from 2019 and 2021 for relevant discussions. I will check the java-base image and move the discussion to the distroless repo

exist-docker/src/main/resources-filtered/Dockerfile
duncdrum: Yes with ZGC the setting is ignored, just as when you introduced the switch to ZGC. This PR does not change the JAVA_OPTIONS. It only addresses the outdated syntax warning from the Dockerfile. Revisiting the JAVA_OPTIONS makes more sense, with the PR for java21

exist-docker/src/main/resources-filtered/Dockerfile
adamretter: I don't see it as a bug with Debian 12 or GCR. Simply, GCR has never included these libraries as many Java Applications don't need them. However, we do need them for eXist-db, which is why previously explicitly installed them into a base Debian image and then copied them into our derivative of GCR.

duncdrum: @adamretter they were included in the gcr java-base image.

adamretter: @duncdrum I can't see it in the GCR java-base images (see below)... Where are you seeing it?

Screenshot 2024-11-21 at 15 43 57

Screenshot 2024-11-21 at 15 41 33

adamretter: @duncdrum Interesting that the files even though listed in that BUILD file don't seem to have ever been in the GCR images as far as I can see. Also, I think even if they were there, from what I understand of that BUILD file ,they would still be missing libexpat which we need for Apache FOP (via Apache Batik)

duncdrum: Just to be sure @adamretter    You are checking in subsequent layers as well, and not seeing the dependencies?

adamretter: > You are checking in subsequent layers as well, and not seeing the dependencies?

@duncdrum Yes, I think so. What do you see when you look in the image?

duncdrum: @adamretter now that I'm back at my laptop. On the image published as part of this PR I see all the dependencies present in the base image, as would be expected based on the build file.
Screenshot 2024-11-25 at 19 35 36

duncdrum: @adamretter nope WAI. There was some discussion about the reasoning behind separate vs single layers in distroless slack. The overall size hasn't changed much, so any differences should only be noticeable on the first download.

adamretter: > WAI

What does "WAI" mean?

line-o: Having more layers means they might be better cached, I suppose.

adamretter: I think that would only work if your lower levels are stable, i.e. don't change. As this now proposes using a `latest` as the base image, that means any change from that explodes upwards and changes all of our upper layers, so therefore they are cache invalidated.

duncdrum: @adamretter WAI mean working as intended. There are pro and cons to layer squashing. The distroless images now come with more layers. `latest` was used before and after this PR. There is no change to cache busting or change inheritance mechanism.

dizzzz (approved):
All look OK, as discussed last monday.

exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
duncdrum, line 23:
```suggestion
# Use JDK 17 in Debian as our production image gcr.io/distroless/java is based on Debian with just a JRE)
```

line-o:
This PR is now good to go as all issues have been addressed.

line-o:
rebased on current develop

dizzzz:
OK to pull? @line-o @ljo @reinhapa

line-o:
@dizzzz yes, from my side

reinhapa:
> @dizzzz yes, from my side

Go!

PR #5552, opened 2024-11-14
[WIP] attempt to make sure streams are closed

### LINKED ISSUE #5345: [BUG] resource leak on Ubuntu 22.04
**Describe the bug**

On a server running exist-db 6.2.0 on Ubuntu 22.04 the monitoring reported a constant growing number of open file handles. On further inspection it became apparent that these were all temporary files created by exist-db. These files would only be purged from the filesystem when exist-db was restarted. The same query executed on the same version of eXist-db running on Ubuntu 20.04 _did not show this behaviour_. 

The smallest reproducible XQuery triggering this behaviour we found is: 

```xquery
"TEST" => util:string-to-binary() => compression:deflate() => response:stream-binary()
```

The above will create a temporary file in eXist-db's temporary file folder. The folder can be found by searching for a subfolder whose name starts with `exist-db-temp-file-manager-` followed by a number of digits. The parent folder can be queried with `util:system-property("java.io.tmpdir")`

The java implementation of `util:string-to-binary` might be the issue here as it is opening a stream (`UnsynchronizedByteArrayInputStream`) without using the form usually used throughout the exist-db codebase (`try( ... ){ }`).

```java
protected BinaryValue stringToBinary(String str, String encoding) throws XPathException {
  try {
    return BinaryValueFromInputStream.getInstance(context, new Base64BinaryValueType(), new UnsynchronizedByteArrayInputStream(str.getBytes(encoding)), this);
  } catch(final UnsupportedEncodingException e) {
    throw new XPathException(this, "Unsupported encoding: " + encoding);
  }
}
```

**Expected behavior**

All temporary files associated with a query to be removed after the query has finished running.

**To Reproduce**

**Leaking**

1. evaluate the test in eXide
    ```xquery
    "TEST" => util:string-to-binary() => compression:deflate() => response:stream-binary()
    ```
2. list the contents of the temporary directory
3. there is __one new__ temporary file 
4. it will remain until exist-db is restarted

**Safe**

```xquery
"TEST" => util:base64-encode() => xs:base64Binary() => response:stream-binary()
```

and

```xquery
reponse-stream-binary(compression:deflate(util:string-to-binary("TEST")))
```

1. evaluate one of the safe examples above
2. list the contents of the temporary directory
3. there is no __new__ temporary file

**Context**

 - Build: eXist-6.2.0
 - Java: openjdk version "1.8.0_312"
 - OS: Ubuntu 22.04.3 LTS

**Additional context**
- How is eXist-db installed? built from source, running as a service
- Any custom changes in e.g. `conf.xml`? several

line-o:
### Description:

This is work in progress as we have still a test failure in `BinaryValueFromBinaryStringTest.cast_hexBinary_to_base64`

### Reference:

fixes #5345

### Type of tests:

none added

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ArrowOperator.java
+++ b/exist-core/src/main/java/org/exist/xquery/ArrowOperator.java
@@ -95,23 +95,7 @@ public class ArrowOperator extends AbstractExpression {
         }
         contextSequence = leftExpr.eval(contextSequence, null);
 
-        final FunctionReference fref;
-        if (fcall != null) {
-            fref = new FunctionReference(this, fcall);
-        } else {
-            final Sequence funcSeq = funcSpec.eval(contextSequence, contextItem);
-            if (funcSeq.getCardinality() != Cardinality.EXACTLY_ONE)
-            {throw new XPathException(this, ErrorCodes.XPTY0004,
-                    "Expected exactly one item for the function to be called, got " + funcSeq.getItemCount() +
-                            ". Expression: " + ExpressionDumper.dump(funcSpec));}
-            final Item item0 = funcSeq.itemAt(0);
-            if (!Type.subTypeOf(item0.getType(), Type.FUNCTION_REFERENCE)) {
-                throw new XPathException(this, ErrorCodes.XPTY0004,
-                    "Type error: expected function, got " + Type.getTypeName(item0.getType()));
-            }
-            fref = (FunctionReference)item0;
-        }
-        try {
+        try (final FunctionReference fref = getFunctionReference(contextSequence, contextItem)) {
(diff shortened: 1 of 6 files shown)


adamretter:
@line-o This approach won't work, you need to first better understand the lifetime of how these streams are designed to be used.

reinhapa:
> @line-o Just to warn you that this approach won't work, you need to first better understand the lifetime of how these streams are designed to be used.

@adamretter I was tinkering with @line-o a bit on that specific  part of the code without better knowledge (as the draft state also hopefully refelcts 😎) 

Any insights and tips to approach the problems are welcome of course

dizzzz:
Community Call: Agreed to close PR.

PR #5574, opened 2024-12-09
[hotfix] fix and extend TransformFromPkgTest

line-o:
- load the module from existdb instead of an external source
- test with and without xmldb: scheme
- fix template under test to actually do something
- ensure that the output is checked for correctness

### DIFF
--- a/exist-core/src/test/java/org/exist/xquery/functions/transform/TransformFromPkgTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/functions/transform/TransformFromPkgTest.java
@@ -39,6 +39,10 @@ import static org.junit.Assert.assertNotNull;
  */
 public class TransformFromPkgTest {
 
+    private static final String moduleLocation = "/db/system/repo/functx-1.0.1/functx/functx.xsl";
+    private static final String inputXml = "<x>bonjourno</x>";
+    private static final String expectedOutput = "<r xmlns:functx=\"http://www.functx.com\">hello</r>";
+
     private static Path getConfigFile() {
         final ClassLoader loader = TransformFromPkgTest.class.getClassLoader();
         final char separator = System.getProperty("file.separator").charAt(0);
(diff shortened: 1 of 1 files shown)


line-o:
XQTS test action fails because it cannot find the zip file with previous actions. This might be due to this being removed from the cache by accident.

duncdrum (approved):

reinhapa (approved):

adamretter:
@line-o This test no longer serves the purpose of what it was designed to test, please revert your changes, if you want to add additional tests for loading from the db then please do so by adding additional test functions. The test function `transformWithModuleFromPkg` needs to be restored. The whole purpose of this test was to load something directly from a package - changing this to simply point at the path in the database of the XSL completely skips the point of this test and means that the test is now deleted and that functionality does not work.

reinhapa:
> @line-o This test no longer serves the purpose of what it was designed to test, please revert your changes, if you want to add additional tests for loading from the db then please do so by adding additional test functions. The test function `transformWithModuleFromPkg` needs to be restored. The whole purpose of this test was to load something directly from a package - changing this to simply point at the path in the database of the XSL completely skips the point of this test and means that the test is now deleted and that functionality does not work.

@adamretter  Unfortunately, it wasn't really clear to us  (@line-o , @duncdrum and myselt ) what exactly the purpose of the test was.
The reason for the adjustment was the dependency on external resources, which in this case led to the test failing because there was an error in the external resource on which the test was based. If you see an opportunity to implement it this way without having to access external resources that we don't have under control ourselves, we'll be happy to take it.

line-o:
OK, so I found the underlying issue. And will post a follow-up PR that fixes it.

line-o:
The feature introduced in https://github.com/eXist-db/exist/commit/8b71534330e63f316994d2778bd0237f7c29b1cf only worked for stylesheets that happened to be available at the import-uri.
When datypic botched their redirects this became visible as `http://www.functx.com/functx.xsl` no longer returns the stylesheet.

adamretter:
> @adamretter Unfortunately, it wasn't really clear to us (@line-o , @duncdrum and myselt ) what exactly the purpose of the test was.

@reinhapa Then why not just ask?

> The reason for the adjustment was the dependency on external resources, which in this case led to the test failing because there was an error in the external resource. If you see an opportunity to implement it a way without relaying external resources, we'll be happy to take it.

Sure. The point was that it should use the external resource - so removing that defeats the point. If you want to skip it if the resource is not available for some reason then you should instead use the appropriate JUnit annotation for that assumption.

> had a single test that loaded the XSL provided by the package from an external source http://functx.com/functx.xsl.
> Since the external source is no longer available at that location the test failed.

@line-o Then it should instead be handled by the Catalog and/or EXPath Pkg registry.

> We found it logical to use the existing package that is provided by the db instead.

That defeats the purpose of the test.

> When datypic botched their redirects this became visible as http://www.functx.com/functx.xsl no longer returns the stylesheet.

I doubt that Datypic "botched" their redirects! I know Priscilla Walmsley to be quite careful about such things. I would rather have thought that this was an unexpected regression or an intentional change on her part. Regardless, if the lack of external resources is an issue then this should be handled instead by the Catalog or EXPath Pkg repo.

PR #5575, opened 2024-12-09
[6.x.x] Backport #5574

line-o:


duncdrum (approved):

reinhapa (approved):

adamretter:
This is actually a regression and should be reverted - as described here - https://github.com/eXist-db/exist/pull/5574#issuecomment-2531068566

reinhapa:
> This is actually a regression and should be reverted - as described here - [#5574 (comment)](https://github.com/eXist-db/exist/pull/5574#issuecomment-2531068566)

@adamretter  I saw your comment. If there is a way to test this regression without external resources we would be happy to take it.

line-o:
> This is actually a regression

The external stylesheet that was retrieved is no longer available at that location.

PR #5576, opened 2024-12-10
[bugfix] registered import-uris have precedence

line-o:
When importing stylesheets the URIs are resolved in order

- registered import-uris in package repo
- XMLDB-locations (relative, absolute and with or without scheme)
- any other location is treated as an exteranal source

The test class and its cases were renamed to reflect their purpose. Some inline comments were added to help describing the intent.

follow-up for #5574

### DIFF
--- a/exist-core/src/main/java/org/exist/xslt/XsltURIResolverHelper.java
+++ b/exist-core/src/main/java/org/exist/xslt/XsltURIResolverHelper.java
@@ -52,13 +52,15 @@ public class XsltURIResolverHelper {
             @Nullable final URIResolver defaultResolver, @Nullable final String base, final boolean avoidSelf) {
         final List<URIResolver> resolvers = new ArrayList<>();
 
+        // EXpath Pkg resolver
+        // This resolver needs to be the first one to prevent
+        // HTTP requests for registered package names (e.g. http://www.functx.com/functx.xsl)
+        brokerPool.getExpathRepo().map(repo -> resolvers.add(new PkgXsltModuleURIResolver(repo)));
+
         if (base != null) {
             // database resolver
-            resolvers.add(new EXistURISchemeURIResolver(new EXistURIResolver(brokerPool, base)));
+            resolvers.add(new EXistURIResolver(brokerPool, base));
         }
-
-        // EXpath Pkg resolver
-        brokerPool.getExpathRepo().map(repo -> resolvers.add(new PkgXsltModuleURIResolver(repo)));
 
         // default resolver
         if (defaultResolver != null) {
--- a/exist-core/src/test/java/org/exist/xquery/functions/transform/ImportLocalStylesheetTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/functions/transform/ImportLocalStylesheetTest.java
@@ -35,18 +35,22 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 /**
+ * Test transform:transform with an imported stylesheet from various
+ * locations
+ *
  * @author <a href="mailto:adam@evolvedbinary.com">Adam Retter</a>
+ * @author <a href="mailto:juri.leino@existsolutions.com">Juri Leino</a>
  */
-public class TransformFromPkgTest {
+public class ImportLocalStylesheetTest {
 
-    private static final String moduleLocation = "/db/system/repo/functx-1.0.1/functx/functx.xsl";
-    private static final String inputXml = "<x>bonjourno</x>";
-    private static final String expectedOutput = "<r xmlns:functx=\"http://www.functx.com\">hello</r>";
+    private static final String XSL_DB_LOCATION = "/db/system/repo/functx-1.0.1/functx/functx.xsl";
+    private static final String INPUT_XML = "<in>bonjourno</in>";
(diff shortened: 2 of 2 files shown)


adamretter:
I don't think this works... What about the catalog?

line-o:
@adamretter If you have a test case that shows an issue please post it here. Otherwise, I can only judge by the tests that are already present and I can come up with.

line-o:
At least now the registered import uri is really resolved against the repo instead of trying to resolve it via http request.

adamretter:
You're missing the point of the test, please restore the original test

reinhapa:
> You're missing the point of the test, please restore the original test

@adamretter could you please explain to us the point that  we missing, instead of just telling us to restore a test as it was that we did no understand in the fist place. 

As from #5574
> @reinhapa Then why not just ask?

I personally would like to learn from that. Just rolling back will not help me (and may others in the future)  in getting reasoning behind the current test. as the discussion seems to prove already.  I would like see this issue as a chance to improve the test and the understanding for regression it wants to cover.

line-o:
> You're missing the point of the test, please restore the original test

The first test case is importing the stylesheet from the repo using the registered import-uri.

line-o:
To better understand the current situation and the proposed solution:

The feature, to be able to import stylesheets using an import-uri that is registered in the databases package repository, __did not work__. The order of URI resolvers prevented that from happening, as the first resolver captured the URI and resolved it as an external resource.

By putting the `PkgXsltModuleURIResolver` first, all stylesheet import URIs will be matched against the ones registered in the package repo. That way we can be sure that they take precedence.

adamretter:
> I would like see this issue as a chance to improve the test

@reinhapa But it doesn't improve the original test. Instead it removes the original test and adds different tests that prove something completely different.

> in getting reasoning behind the current test

@reinhapa You could just talk to me. Even after all the nonsense and abuse I have been shovelled lately, I am still hear and trying to help. If you want to understand the original purpose of the test, then the obvious thing would be to have a conversation with me (the original author of the test)

> did not work

@line-o That is not my impression. It looks to me like this test only started recently failing, before that it was passing just fine.

line-o:
> this test only started recently failing, before that it was passing just fine

@adamretter because of `curl -i http://www.functx.com/functx.xsl`

exist-core/src/test/java/org/exist/xquery/functions/transform/ImportStylesheetTest.java
```diff
@@ -57,51 +61,72 @@ private static Path getConfigFile() {
...
     }
 
     @ClassRule
     public static ExistXmldbEmbeddedServer existXmldbEmbeddedServer = new ExistXmldbEmbeddedServer(true, false, true, getConfigFile());
 
     @Test
-    public void testImportNoScheme() throws XMLDBException {
-        final String xquery = getQuery(moduleLocation);
+    public void fromRegisteredImportUri() throws XMLDBException {
+        final String xquery = getQuery("http://www.functx.com/functx.xsl");
```
reinhapa, line 92: @line-o I found  a working document here: https://www.datypic.com/xsl/functx-1.0.1-doc.xsl that we may should use instead...

reinhapa (changes requested):

dizzzz:
@line-o by default curl does not follow redirects; I think the -L flag will make tha happen...

`curl -Li http://www.functx.com/functx.xsl`

(cant check right now)

If so, we could consider if we can enforce the HTTP-resolver in the test to follow redirects too?

line-o:
But the test is clearly about importing the Stylesheet from within the db. The fact that an HTTP connection es even attempted is a failure.
The test would need to test for that as well. Sadly, it's easier said than done. We would need to either mock a lot or spy on the expected resolver to be called.

adamretter:
> Absolutely not. The test is about importing a locally available Stylesheet and the fact that it was resolved from an external source was the problem I fixed.

> But the test is clearly about importing the Stylesheet from within the db. 

@line-o Nope! That is absolutely not what I wrote this test for. This test was to ensure the loading of XSLT imports from within EXPath packages. This PR deletes that work and important use-case and goes in a completely different direction. This PR is a regression.

The correct approach would have been to fix this in the XML Resolver and/or EXPath package resolver.

line-o:
> This test was to ensure the loading of XSLT imports from within EXPath packages.

OK, so we are in agreement here, good. The test case to do that is back in now that this feature is fixed and more cases were added.

> The correct approach would have been to fix this in the XML Resolver and/or EXPath package resolver.

This is what I did, by changing the order of the resolvers everything is in order.

exist-core/src/test/java/org/exist/xquery/functions/transform/ImportStylesheetTest.java
line-o: I just coincidentally was _also_ available at that address and not just registered with that _namespace URI_ in exist-db's package repository.
It is non-obvious to use HTTP-addresses to uniquely name packages without _also_ guaranteeing the source is available there but this is the status-quo, not the use-case this test is about and also a different discussion.

line-o:
The http-resolver should not be used for any of the test cases within `ImportStylesheetTest`. Maybe `ImportLocalStylesheetTest` would communicate that clearer?

dizzzz (approved):

reinhapa (approved):

PR #5586, opened 2024-12-20
[6.x.x] allow module imports in one-off xqueries

line-o:
backport of #5529

### DIFF
--- a/exist-core/src/main/java/org/exist/source/SourceFactory.java
+++ b/exist-core/src/main/java/org/exist/source/SourceFactory.java
@@ -111,7 +111,7 @@ public class SourceFactory {
                 && ((location.startsWith("/db") && !Files.exists(Paths.get(firstPathSegment(location))))
                 || (contextPath != null && contextPath.startsWith("/db") && !Files.exists(Paths.get(firstPathSegment(contextPath)))))) {
             final XmldbURI pathUri;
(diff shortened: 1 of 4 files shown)


exist-core/src/test/java/org/exist/xquery/AbsolutePathTests.java
```diff
@@ -133,4 +133,5 @@ public void topLevelAbsolutePath() throws EXistException, PermissionDeniedExcept
 
         assertThatXQResult(actual, equalTo(expected));
     }
+
```
reinhapa, line 136: This change should be reverted as it does provide any change...

reinhapa (changes requested):

line-o:
@reinhapa more importantly the test-failure needs to be addressed.

line-o:
Turns out that the order of the first two parameters of `XqueryContext.compileOrBorrowModule` is reversed on develop
(see 4b92888ce3a787b3abe03882559cbf29b88b967a)

line-o:
Windows tests are super flaky. Each run something fails but each time a different test. Is that known?

dizzzz:
Interesting, I see indeed sometimes just a test fail....

dizzzz (approved):

reinhapa (approved):

PR #5588, opened 2024-12-24
[bugfix] upgrade transient library commons-beanutils

dizzzz:
- run tests -> OK
- login webdav -> OK

### DIFF
--- a/extensions/webdav/pom.xml
+++ b/extensions/webdav/pom.xml
@@ -53,12 +53,22 @@
         </dependency>
 
         <dependency>
+            <groupId>commons-beanutils</groupId>
+            <artifactId>commons-beanutils</artifactId>
+            <version>1.9.4</version>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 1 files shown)


dizzzz:
sync to 6.x ?

duncdrum (approved):
Merry x-mas. Nice fix, I have no stake in a backport. @dizzzz would be one less thing where they diverge.

extensions/webdav/pom.xml
```diff
@@ -52,6 +52,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>commons-beanutils</groupId>
```
adamretter, line 56: Perhaps add a comment saying why this is hear - as it is transient, it typically should not be here - which means that without extra information, someone could easily remove this in future by mistake.

dizzzz: @adamretter yes makes sense

reinhapa (approved):

Issue #5235, opened 2024-02-27
[BUG] ForClaues's PositionalVar returns incorrect value on empty sequence when allowing empty


adamretter:
The following query in eXist-db returns the wrong result:

```xquery
for $x allowing empty at $y in ("a", "b", "c")[4]
return
  $x || ":" || $y
```

The result should be `:0`, but eXist-db returns `:1`.

Tested against eXist-db 7.0.0-SNAPSHOT.

PR #5589, opened 2024-12-25
Fix ForClause PositionalVar empty sequence handling

Robin6939:
### Description:  
Resolved an issue where ForClause's PositionalVar incorrectly returned values when faced with an empty sequence while allowing emptiness. This adjustment ensures accurate handling of edge cases in sequence evaluations, aligning behavior with expected standards.

### Reference: 
https://github.com/eXist-db/exist/issues/5235

### Tests: 
Added tests which verifies the same

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ForExpr.java
+++ b/exist-core/src/main/java/org/exist/xquery/ForExpr.java
@@ -229,7 +229,8 @@ public class ForExpr extends BindingExpression {
         context.proceed(this);
         context.setContextSequencePosition(p, in);
         if (positionalVariable != null) {
-            at.setValue(new IntegerValue(this, p + 1));
+            final int position = contextItem == AtomicValue.EMPTY_VALUE ? 0 : p + 1;
+            at.setValue(new IntegerValue(this, position));
         }
         final Sequence contextSequence = contextItem.toSequence();
         // set variable value to current item
--- a/exist-core/src/test/xquery/xquery3/flwor.xql
+++ b/exist-core/src/test/xquery/xquery3/flwor.xql
@@ -156,3 +156,32 @@ function flwor:allowing-empty($n as xs:integer) {
     return concat("[", $x, "]")
 };
 
+declare
+    %test:args(4)
+    %test:assertEquals(":0")
+    %test:args(2)
+    %test:assertEquals("b:1")
+    %test:args(1)
+    %test:assertEquals("a:1")
+    %test:args(5)
+    %test:assertEquals(":0")
+function flwor:allowing-empty-fix($n as xs:integer) {
+    let $sequence := ("a", "b", "c")[$n]
+    for $x allowing empty at $y in $sequence
+    return $x || ":" || $y
+};
+
+declare
+    %test:args(4)
+    %test:assertEquals("")
+    %test:args(2)
+    %test:assertEquals("b:1")
+function flwor:no-allow-empty($n as xs:integer) {
+    let $sequence := ("a", "b", "c")[$n]
+    return
+        if (empty($sequence)) then
+            ""
+        else
+            for $x at $y in $sequence
+            return $x || ":" || $y
+};


line-o (commented):
This looks really great, thank you for your contribution! What a 🎁 
Would it be OK for me to fix the formatting and commit message format or do you want to do it?

Robin6939:
Hi @line-o,
Thank you for your review 😄 
I tried to fix the commit message. Can you help me fix the formatting. Since this is my first PR here, I am not familiar with it.

adamretter:
@Robin6939 Wonderful to see that you made some progress after our call yesterday... Quick work!

Regards formatting, IntelliJ has a Java code formatter on the `Code` menu item, there is `Reformat Code`, but if the existing file is not correctly formatted, then this will cause more noise than you need. So I will comment directly on the formatting of your code separately...

exist-core/src/main/java/org/exist/xquery/ForExpr.java
```diff
@@ -229,7 +229,10 @@ private void processItem(LocalVariable var, Item contextItem, Sequence in, Seque
         context.proceed(this);
         context.setContextSequencePosition(p, in);
         if (positionalVariable != null) {
-            at.setValue(new IntegerValue(this, p + 1));
+            if(contextItem.getType() == Type.ANY_ATOMIC_TYPE)
```
adamretter, line 232: @Robin6939 For formatting:

1. Insert a space between `if` and `(`
2. Add braces even for one line if statements.

e.g.:

```java
if (contextItem.getType() == Type.ANY_ATOMIC_TYPE) {
    at.setValue(new IntegerValue(this, 0));
} else {
    at.setValue(new IntegerValue(this, p + 1));
}
```

adamretter (changes requested):
Small formatting issue.

reinhapa (approved):

exist-core/src/main/java/org/exist/xquery/ForExpr.java
line-o: how about 
```java
final int nextPosition = contextItem.getType() == Type.ANY_ATOMIC_TYPE ? 0 : p + 1;
at.setValue(new IntegerValue(this, nextPosition));
```

exist-core/src/main/java/org/exist/xquery/ForExpr.java
```diff
@@ -229,7 +229,11 @@ private void processItem(LocalVariable var, Item contextItem, Sequence in, Seque
         context.proceed(this);
         context.setContextSequencePosition(p, in);
         if (positionalVariable != null) {
-            at.setValue(new IntegerValue(this, p + 1));
+            if (contextItem.getType() == Type.ANY_ATOMIC_TYPE) {
```
line-o, line 232: how about
```
final int position = contextItem.getType() == Type.ANY_ATOMIC_TYPE ? 0 : p + 1;
at.setValue(new IntegerValue(position));
```

exist-core/src/test/xquery/xquery3/flwor.xql
```diff
@@ -156,3 +156,17 @@ function flwor:allowing-empty($n as xs:integer) {
     return concat("[", $x, "]")
 };
 
+declare
```
line-o, line 159: Do we have a test that shows the correct behaviour if `allowing empty` is not set?

exist-core/src/main/java/org/exist/xquery/ForExpr.java
line-o: I am also wondering, if there is a better way to check for this case.
Since we want to avoid that the position is set to 0 in any other case.
Since the function `processItem` is called with contextItem set to `AtomicValue.EMPTY_VALUE` if and only if `allowing empty` is set _and_ there are no items in the binding sequence we should best check for that.
What do you think?

line-o (commented):
I added a few suggestions and questions.

exist-core/src/test/xquery/xquery3/flwor.xql
Robin6939: There isn't any right now
Should I add them?

exist-core/src/main/java/org/exist/xquery/ForExpr.java
Robin6939: Yeah that is better
Will make that change

dizzzz (approved):

line-o (approved):

PR #5596, opened 2025-01-02
[feature] gh package snapshots...

### LINKED ISSUE #5592: [feature] use GitHub registry for nightly build artefacts (maven)
- [x] allow  `-SNAPSHOT` artifacts to be pulled by maven for developing
- [x] add GH registry to list of repos for maven builds across the org

[Communit Call Notes](https://docs.google.com/document/d/103QKAAjoDiaTS0va1dPLh2SEJmWVJVEKRamq0JxEdTU/edit?tab=t.0)

duncdrum:
from CI (hopefully)

The previous SNAPSHOT repo hasn't seen any traffic for over a year. This automates release snapshots to GitHub on push events on develop. 

We could make the step, run only after tests have passed (as they should always be green on develop anyway) but that would mean that every time the OWASP dependency check times out, we won't publish artefacts. So I decided against it. What do others think?

This will also not publish artefacts from PRs by design, but this could be adjusted. 

Screenshot 2025-01-02 at 12 00 20

close #5592

### DIFF
--- a/.github/workflows/ci-snapshots.yml
+++ b/.github/workflows/ci-snapshots.yml
@@ -1,0 +1,30 @@
+name: Publish Snaphots to Github Packages
+
+on:
+  push:
+    branches:
+      - develop
+  workflow_dispatch: # Allow manual triggering
+
+jobs:
+  publish-snapshots:
+    name: Deploy Snapshots
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/develop'
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          distribution: liberica
+          java-version: '17'
+      - name: Cache Maven packages
+        uses: actions/cache@v4
+        with:
+          path: ~/.m2
+          key: deploy-${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: deploy-${{ runner.os }}-maven
+      - name: Deploy SNAPSHOT maven artefacts
+        run: mvn -V -B -q -DskipTests -Ddependency-check.skip=true -P !mac-dmg-on-unix,!installer,!concurrency-stress-tests,!micro-benchmarks,skip-build-dist-archives clean deploy
--- a/exist-parent/pom.xml
+++ b/exist-parent/pom.xml
@@ -1177,17 +1177,28 @@
     </reporting>
 
     <repositories>
-        <repository>
-            <id>exist-db-snapshots</id>
-            <name>Evolved Binary - eXist-db Snapshots</name>
-            <url>https://repo.evolvedbinary.com/repository/exist-db-snapshots/</url>
-            <releases>
-                <enabled>false</enabled>
-            </releases>
-            <snapshots>
-                <enabled>true</enabled>
-            </snapshots>
-        </repository>
+            <repository>
+                <id>github</id>
+                <name>GitHub Registry for Snapshots</name>
+                <url>https://maven.pkg.github.com/eXist-db/exist</url>
+                <releases>
+                    <enabled>false</enabled>
+                </releases>
+                <snapshots>
+                    <enabled>true</enabled>
+                </snapshots>
+            </repository>
+<!--        <repository>-->
+<!--            <id>exist-db-snapshots</id>-->
+<!--            <name>Evolved Binary - eXist-db Snapshots</name>-->
+<!--            <url>https://repo.evolvedbinary.com/repository/exist-db-snapshots/</url>-->
+<!--            <releases>-->
+<!--                <enabled>false</enabled>-->
+<!--            </releases>-->
+<!--            <snapshots>-->
+<!--                <enabled>true</enabled>-->
+<!--            </snapshots>-->
+<!--        </repository>-->
         <repository>
             <id>exist-db</id>
             <name>Evolved Binary - eXist-db Releases</name>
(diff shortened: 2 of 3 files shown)


exist-parent/pom.xml
```diff
@@ -1177,17 +1177,28 @@
     </reporting>
 
     <repositories>
-        <repository>
```
adamretter, line 1180: Please leave the existing repository in place as it is needed for a number of purposes including Nightly Builds and Releases.

adamretter (changes requested):
The existing repo should remain, you can add the additional one beside it.

exist-parent/pom.xml
duncdrum:
```shell
The project org.exist-db:exist:7.0.0-SNAPSHOT (/Users/hal1000/Documents/GitHub/exist/pom.xml) has 1 error
[ERROR]     Non-parseable POM /Users/hal1000/Documents/GitHub/exist/exist-parent/pom.xml: Duplicated tag: 'snapshotRepository' 
```

can your automation not pull the snapshots from the GitHub package registry, these builds will be automated?

exist-parent/pom.xml
```diff
@@ -1232,10 +1254,15 @@
 
     <distributionManagement>
         <snapshotRepository>
-            <id>exist-db-snapshots</id>
-            <name>Evolved Binary - eXist-db Snapshots</name>
-            <url>https://repo.evolvedbinary.com/repository/exist-db-snapshots/</url>
+            <id>github</id>
+            <name>GitHub Registry for Snapshots</name>
```
line-o, line 1258: Is it valid to set the `snapshotRepository` by just the **id**? Would be worth to try.

exist-parent/pom.xml
```diff
@@ -1217,10 +1228,21 @@
                 <enabled>false</enabled>
             </snapshots>
         </pluginRepository>
+<!--        <pluginRepository>-->
```
line-o, line 1231: we could leave this in for the moment

exist-parent/pom.xml
```diff
@@ -1232,10 +1254,15 @@
...
     <distributionManagement>
         <snapshotRepository>
-            <id>exist-db-snapshots</id>
-            <name>Evolved Binary - eXist-db Snapshots</name>
-            <url>https://repo.evolvedbinary.com/repository/exist-db-snapshots/</url>
+            <id>github</id>
+            <name>GitHub Registry for Snapshots</name>
+            <url>https://maven.pkg.github.com/eXist-db/exist</url>
         </snapshotRepository>
+<!--        <snapshotRepository>-->
```
line-o, line 1261: This can be removed, for this repo and branch.

.github/workflows/ci-snapshots.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          distribution: liberica
+          java-version: '17'
+      - name: Cache Maven packages
+        uses: actions/cache@v4
+        with:
+          path: ~/.m2
+          key: deploy-${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
```
line-o, line 27: Is this key hash not the default?

.github/workflows/ci-snapshots.yml
```diff
@@ -0,0 +1,30 @@
...
+          restore-keys: deploy-${{ runner.os }}-maven
```
line-o, line 28: the restore-keys seems to be different from the cache key. Is that intended or am I reading it wrong?

line-o (commented):
Great work. Much less changes than anticipated. Is the snapshot already downloadable somewhere?

duncdrum:
SNAPSHOTS: https://github.com/orgs/eXist-db/packages?repo_name=exist

dizzzz (approved):

exist-parent/pom.xml
```diff
@@ -1228,13 +1239,24 @@
                 <enabled>true</enabled>
             </snapshots>
         </pluginRepository>
+        <pluginRepository>
```
adamretter, line 1242: This is not needed as there are no plugins that we publish.

exist-parent/pom.xml
```diff
@@ -1228,13 +1239,24 @@
...
+            </snapshots>
+        </pluginRepository>
     </pluginRepositories>
 
     <distributionManagement>
         <snapshotRepository>
-            <id>exist-db-snapshots</id>
-            <name>Evolved Binary - eXist-db Snapshots</name>
-            <url>https://repo.evolvedbinary.com/repository/exist-db-snapshots/</url>
+            <id>github</id>
```
adamretter, line 1257: Please also restore the old entry for `Evolved Binary - eXist-db Snapshots` as this is needed for the nightly builds and releases.

adamretter (changes requested):
This needs a bit more work as it breaks some things and has some unnecessary things too.

exist-parent/pom.xml
duncdrum: @adamretter see my earlier comment. There can only be one `snapshotRepository`

exist-parent/pom.xml
line-o: @duncdrum was this needed for snapshots of the sub-projects to be published to the registry?

line-o (approved):
Cool!

exist-parent/pom.xml
adamretter: @duncdrum So how can we fix this then?

PR #5598, opened 2025-01-03
Backport multi arch 5533

duncdrum:
we are being bitten by see fabric8io/docker-maven-plugin#1289

### DIFF
--- a/.github/workflows/ci-deploy.yml
+++ b/.github/workflows/ci-deploy.yml
@@ -65,12 +65,12 @@ jobs:
         run: mvn -q -Ddocker.tag=release -Ddocker.username=$DOCKER_USERNAME -Ddocker.password=$DOCKER_PASSWORD docker:build docker:push
         working-directory: ./exist-docker
       # NOTE (DP): This is for debugging, publishes an experimental image from inside PRs against develop
-      - name: Publish experimental images
-        if: github.base_ref == 'develop-6.x.x'
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
-        run: mvn -q -Ddocker.tag=experimental6 -Ddocker.username=$DOCKER_USERNAME -Ddocker.password=$DOCKER_PASSWORD docker:build docker:push
-        working-directory: ./exist-docker  
+      # - name: Publish experimental images
+      #   if: github.base_ref == 'develop-6.x.x'
+      #   env:
+      #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


duncdrum:
to test:

```
docker pull existdb/existdb:experimental6
docker run -dit -p 8080:8080 --name exp6 --rm exisdb/existdb:experimental6
```

[all green](https://github.com/duncdrum/exist/actions/runs/12600539288/job/35119787461?pr=2)

dizzzz (approved):

.github/workflows/ci-deploy.yml
```diff
@@ -5,7 +5,8 @@ jobs:
     name: Build and Test Images
     runs-on: ubuntu-latest
     # NOTE (DP): Publish on develop and master, test on PRs against these
-    if: github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/master' || github.base_ref ==  'develop' || github.base_ref ==  'master'
+    # TODO(DP) Reinstate CRONed release builds to update stock apps regularly
+    if: github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/develop-6.x.x' || github.base_ref ==  'develop' || github.base_ref ==  'develop-6.x.x'
```
line-o, line 9: Why should we remove master from the list of branches to deploy from?

line-o: I would say we can remove develop here _but not_ master.

line-o (commented):
Thanks for this. Really good that you seem to have worked out the problems with the maven plugin.

I would like to understand the conditions under which the deploy workflow runs before this is merged.

reinhapa (approved):

.github/workflows/ci-deploy.yml
duncdrum: @line-o not really. For now `develop` deploys `latest`, and this here replaces `release` aka `master`. 

There is a question is if we plan further merges to `master` from `develop-6.x.x` or if the next merge will be from `develop` given the divergent histories of dev and dev6.

line-o: One thing is for sure: this code will never be in `develop` 
So we can remove it from the condition.
Whether the code ends on master is a different question.
Just to keep that option I would want to see the condition to change to
```
if: github.ref == 'refs/heads/develop-6.x.x' || github.ref == 'refs/heads/master' || github.base_ref == 'develop-6.x.x' || github.base_ref == 'master'
```

line-o (changes requested):
Let's discuss this tonight.

duncdrum:
@line-o this would be a breaking change and start publishing `existv6` as `:latest` which has been `existv7` for more than a year.

line-o:
Ah! I rhink it's best that we have a brief chat then. I might be completely on the wrong track.

.github/workflows/ci-deploy.yml
line-o: As discussed in tonights call we will have all three branches in here

line-o (approved):
Cool!

PR #5642, opened 2025-02-24
[cicd] Remove unused java and library checkers

dizzzz:
Remove of
- detection of buggy java15 versions (pre 17 & 21)
- detection of XML libraries (needed for war file deployment only, in combination with the late 'endorsed' directory)


exist-start/pom.xml
```diff
@@ -118,6 +118,7 @@ The original license statement is also included below.]]></preamble>
                             <failOnWarning>true</failOnWarning>
                             <ignoredUnusedDeclaredDependencies>
                                 <ignoredUnusedDeclaredDependency>org.hamcrest:hamcrest:jar</ignoredUnusedDeclaredDependency>
+                                <ignoredUnusedDeclaredDependency>junit:junit:jar</ignoredUnusedDeclaredDependency>
```
reinhapa, line 121: Better remove dependency if not used....

reinhapa (changes requested):

dizzzz:
I will delete the PR actually :-)

PR #5668, opened 2025-03-14
[refactor] use isEmpty() where possible - readability

dizzzz:
[refactor] use isEmpty() where possible - readability

### DIFF
--- a/exist-core/src/main/java/org/exist/dom/QName.java
+++ b/exist-core/src/main/java/org/exist/dom/QName.java
@@ -344,7 +344,7 @@ public class QName implements Comparable<QName> {
     public static QName parse(final Context context, final String qname, final String defaultNS)
             throws IllegalQNameException {
 
-        final char firstChar = qname.length() > 0 ? qname.charAt(0) : 0;
+        final char firstChar = !qname.isEmpty() ? qname.charAt(0) : 0;
 
         // quick test if qname is in clark notation
         if (firstChar == '{') {
--- a/exist-core/src/main/java/org/exist/dom/QName.java
+++ b/exist-core/src/main/java/org/exist/dom/QName.java
@@ -344,7 +344,7 @@ public class QName implements Comparable<QName> {
     public static QName parse(final Context context, final String qname, final String defaultNS)
             throws IllegalQNameException {
 
-        final char firstChar = !qname.isEmpty() ? qname.charAt(0) : 0;
+        final char firstChar = qname.isEmpty() ? 0 : qname.charAt(0);
 
         // quick test if qname is in clark notation
         if (firstChar == '{') {
--- a/exist-core/src/main/java/org/exist/source/ClassLoaderSource.java
+++ b/exist-core/src/main/java/org/exist/source/ClassLoaderSource.java
@@ -65,7 +65,7 @@ public class ClassLoaderSource extends URLSource {
     public String type() {
         final String protocol = url.getProtocol();
         final String host = url.getHost();
-        if (protocol.equals("file") && (host == null || host.length() == 0 || "localhost".equals(host) || "127.0.0.1".equals(host))) {
+        if (protocol.equals("file") && (host == null || host.isEmpty() || "localhost".equals(host) || "127.0.0.1".equals(host))) {
             return "File";
         }
         return "Classloader";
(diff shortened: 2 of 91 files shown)


reinhapa (approved):

exist-core/src/main/java/org/exist/dom/QName.java
```diff
@@ -344,7 +344,7 @@ public static QName parse(final String namespaceURI, final String qname) throws
     public static QName parse(final Context context, final String qname, final String defaultNS)
             throws IllegalQNameException {
 
-        final char firstChar = qname.length() > 0 ? qname.charAt(0) : 0;
+        final char firstChar = !qname.isEmpty() ? qname.charAt(0) : 0;
```
line-o, line 347: this might benefit from reversing the condition:
```java
final char firstChar = qname.isEmpty() ? 0 : qname.charAt(0);
```

exist-core/src/main/java/org/exist/source/ClassLoaderSource.java
```diff
@@ -65,7 +65,7 @@ private static URL sourceToUrl(String source) throws IOException {
     public String type() {
         final String protocol = url.getProtocol();
         final String host = url.getHost();
-        if (protocol.equals("file") && (host == null || host.length() == 0 || "localhost".equals(host) || "127.0.0.1".equals(host))) {
+        if (protocol.equals("file") && (host == null || host.isEmpty() || "localhost".equals(host) || "127.0.0.1".equals(host))) {
```
line-o, line 68: Do we actually need the two null and empty checks of host?
```java
if (protocol.equals("file") && ("localhost".equals(host) || "127.0.0.1".equals(host))) {
```

exist-core/src/main/java/org/exist/source/URLSource.java
```diff
@@ -65,7 +65,7 @@ public URLSource(final URL url) {
     public String path() {
         final String protocol = url.getProtocol();
         final String host = url.getHost();
-        if ("file".equals(protocol) && (host == null || host.length() == 0 || "localhost".equals(host) || "127.0.0.1".equals(host))) {
+        if ("file".equals(protocol) && (host == null || host.isEmpty() || "localhost".equals(host) || "127.0.0.1".equals(host))) {
```
line-o, line 68: same here

exist-core/src/main/java/org/exist/source/URLSource.java
```diff
@@ -75,7 +75,7 @@ public String path() {
     public String type() {
         final String protocol = url.getProtocol();
         final String host = url.getHost();
-        if ("file".equals(protocol) && (host == null || host.length() == 0 || "localhost".equals(host) || "127.0.0.1".equals(host))) {
+        if ("file".equals(protocol) && (host == null || host.isEmpty() || "localhost".equals(host) || "127.0.0.1".equals(host))) {
```
line-o, line 78: same here

exist-core/src/main/java/org/exist/xslt/XsltURIResolverHelper.java
```diff
@@ -73,7 +73,7 @@ public class XsltURIResolverHelper {
             }
         }
 
-        if (resolvers.size() > 0) {
+        if (!resolvers.isEmpty()) {
```
line-o, line 76: I would suggest to reverse the condition here
```java
if (resolvers.isEmpty()) { return null; }
return new URIResolverHierarchy(resolvers);
```

line-o (approved):
I added a few suggestions that you could address otherwise I'll send a followup PR.

dizzzz:
thnx for the reviews! I will rework the PR tonight
image
not sure if this can be solved with a rebase or just a manual fix.

exist-core/src/main/java/org/exist/source/ClassLoaderSource.java
dizzzz: The host==null check protects the .isEmpty() in case of a null value. If no host is provided this means 'File'. The URL class does not mention possible values so we should handle this.

dizzzz: think so yes

exist-core/src/main/java/org/exist/source/URLSource.java
dizzzz: Think so yes

exist-core/src/main/java/org/exist/source/URLSource.java
dizzzz: think so yes

dizzzz:
@line-o there is just a small conflict to to the toString() PR.

PR #5671, opened 2025-03-14
[refactor] replace null check with method call

dizzzz:
[refactor] replace null check with method call


line-o (approved):
Nice

reinhapa (changes requested):
I have some suggestions noted. Also the use of static imports may lead to better readability of the code...

reinhapa (approved):

PR #5673, opened 2025-03-16
No static typing

### LINKED ISSUE #2445: self:: axis inside predicate throws XPST0005
### What is the problem
When used inside a predicate, the `self::` axis throws error XPST0005 in some cirumstances, even though nothing is wrong with the expression. This happens with the most recent eXist releases 4.5.0 and 5.0.0-RC5 (and earlier releases).

### What did you expect
I would expect the query to run without errors (confirmed with Saxon).

Behaviour is the same for in-memory and stored nodes, so these stand-alone test cases will illustrate the problem.

#### [1]
    let $node := 
      <test>
        <el>test <el1>test</el1> test <el2>test</el2></el>
      </test>
    return 
        $node/el[self::el1]

This should return an empty result set. Instead following error is being thrown: 

    org.xmldb.api.base.XMLDBException: exerr:ERROR err:XPST0005 Self::{}el1 called on set of nodes which do not contain any nodes of this name. [at line 5, column 23, source: let $node := 
      <test>
        <el>test <el1>test</el1> test <el2>test</el2></el>
      </test>
    return $node/el[self::el1]] [at line 5, column 23]
      at org.exist.xmldb.RemoteXPathQueryService.compile(RemoteXPathQueryService.java:130)
      at org.exist.client.QueryDialog$QueryThread.run(QueryDialog.java:558)
    Caused by: org.exist.xquery.XPathException: exerr:ERROR err:XPST0005 Self::{}el1 called on set of nodes which do not contain any nodes of this name. [at line 5, column 23, source: let $node := 
      <test>
        <el>test <el1>test</el1> test <el2>test</el2></el>
      </test>
    return $node/el[self::el1]] [at line 5, column 23]
      at org.exist.xmldb.RemoteXPathQueryService.throwXPathException(RemoteXPathQueryService.java:179)
      at org.exist.xmldb.RemoteXPathQueryService.compileAndCheck(RemoteXPathQueryService.java:155)
      at org.exist.xmldb.RemoteXPathQueryService.compile(RemoteXPathQueryService.java:128)
      ... 1 more
    Caused by: org.exist.xquery.XPathException: exerr:ERROR err:XPST0005 Self::{}el1 called on set of nodes which do not contain any nodes of this name. [at line 5, column 23, source: let $node := 
      <test>
        <el>test <el1>test</el1> test <el2>test</el2></el>
      </test>
    return $node/el[self::el1]] [at line 5, column 23]
      at org.exist.xmldb.RemoteXPathQueryService.throwXPathException(RemoteXPathQueryService.java:179)
      at org.exist.xmldb.RemoteXPathQueryService.compileAndCheck(RemoteXPathQueryService.java:155)
      at org.exist.xmldb.RemoteXPathQueryService.compile(RemoteXPathQueryService.java:128)
      at org.exist.client.QueryDialog$QueryThread.run(QueryDialog.java:558)

Of course, this example is contrived, in order to present a simple test case. It seems, however, that the empty node set returned by the test condition in the predicate is significant. If the `self::` axis does select a node (e.g. `$node/el[self::el]`), the query does run without error.

#### [2]
I've found another failing variant XQuery where the `self::` axis occurs inside a predicate and doesn't select a node. Here, parentheses seem to be triggering the same error:

    let $node := 
      <test>
        <el>test <el1>test</el1> test <el2>test</el2></el>
      </test>
    return $node/el[descendant::*/(self::el1)]

This should return the <el> node:

    <el>test <el1>test</el1> test <el2>test</el2></el>

Instead, above error XPST0005 is being thrown. 

Here, the parentheses seem to be crucial. Compare with the same predicate without parentheses:

    let $node := 
      <test>
        <el>test <el1>test</el1> test <el2>test</el2></el>
      </test>
    return $node/el[descendant::*/self::el1]

...which does return the correct result:

    <el>test <el1>test</el1> test <el2>test</el2></el>

I've found a reference to this error message in the ML archives (https://sourceforge.net/p/exist/mailman/message/32496312/), and a bug fix in Github. So there's a chance this issue is related to https://github.com/eXist-db/exist/pull/249.

### Context information
- eXist-db version + Git Revision hash:
    - eXist-4.5.0 / e29b409
    - eXist-5.0.0-RC5 / 4414b1a
- Java version: oracle 1.8.0_191
- Operating system: Windows 10
- 32 or 64 bit: 64 bit
- How is eXist-db installed? JAR installer

line-o:
### Description:

fixes #2445

### Reference:

with XQuery 4.0 static typing feature is removed entirely

### Type of tests:

none new added


dizzzz (changes requested):
Build is failing;
can we add tests eg from the original bug report?

line-o:
I applied commits from an older PR that targeted `develop-6.x.x` and it took 3 attempts to get it to compile.
I will look to add one or two tests from the original issue.

line-o:
I would expect for one to many additional XQTS test cases to now pass.

line-o:
There seems to be zero tests that assert error `XPST0005` to be raised. ;)

dizzzz:
I am wondering.... should this code change be 'protected' to it would not be effective for 'query 3.1' (and older) based queries?

line-o:
The errors that were raised are the result of a incomplete beginning to have the static typing feature. exist-db never claimed to support this optional xquery feature which is why we do not need to guard against it.
It is now that queries will behave much more like in any other Xquery processor (namely baseX and Saxon).

dizzzz (approved):

duncdrum (approved):
as discussed needs documentation

adamretter:
An interesting idea of how to fix the issue... But it must be recognised that this small change is very significant!

Before it is merged two things need to be checked carefully in great detail:

1. No regression in the XQTS results
2. No performance regression (looking at the code change, I suspect this PR will result in a regression in its current form)

line-o:
I see a minor - but consistent - performance improvement. 

Execution times of the entire test suite goes down from  693.4 seconds to 667.3 in the last XQTS run in CI. That is 3.75% faster.
I observed similar results on my local machine.

dizzzz (approved):

PR #5680, opened 2025-03-18
[bugfix] fn:transform: Conversion and treeIndex problems

nverwer:
# Description

The `fn:transform` function sometimes produced `NullPointerException`s. This happened for documents containing nodes of type `org.exist.dom.memtree.ElementImpl` and other subtypes of `org.exist.dom.memtree.NodeImpl`.

The function `org.exist.xquery.functions.fn.transform.Convert.ToSaxon.ofNode(Node)` uses `org.exist.xquery.functions.fn.transform.TreeUtils.treeIndex(Node)` to get a 'path' (as a list of indexes) of the `Node` in its containing document. This 'path' is used to find the node in a newly built document that is suitable for use by Saxon.
Using the 'path' that the `treeIndex` function produces assumes that it starts at the document node.

However, some nodes do not have a containing document. Specifically, in `org.exist.dom.memtree.NodeImpl.getParentNode()`:
```
if (parent.getNodeType() == DOCUMENT_NODE && !((DocumentImpl) parent).isExplicitlyCreated()) {
            /*
                All nodes in the MemTree will return an Owner document due to how the MemTree is implemented,
                however the explicitlyCreated flag tells us whether there "really" was a Document Node or not.
                See https://github.com/eXist-db/exist/issues/1463
             */
            return null;
```
In this case, the 'path' returned by `treeIndex` does not contain the index (0) for the root node, and `org.exist.xquery.functions.fn.transform.TreeUtils.xdmNodeAtIndex(XdmNode, List<Integer>)` will return `null` when a node is expected.

The proposed fix tests for this case. Going up the ancestor chain, when a node that is not a document node, and that does not have a parent is found, the index 0 for this root element is inserted in the 'path'.

# Testing

The` NullPointerException` happened in a rather complicated XQuery with documents composed from several sources.
I have not yet been able to make a simple test case, but I will keep trying.

### DIFF
--- a/.github/workflows/ci-deploy.yml
+++ b/.github/workflows/ci-deploy.yml
@@ -1,11 +1,15 @@
 name: Deploy
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+      - cron: "0 6 * * *"
+  workflow_dispatch:
 jobs:
   build:
     name: Build and Test Images
     runs-on: ubuntu-latest
     # NOTE (DP): Publish on develop and master, test on PRs against these
-    # TODO(DP) Reinstate CRONed release builds to update stock apps regularly
     if: github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/develop-6.x.x' || github.ref == 'refs/heads/master'|| github.base_ref ==  'develop' || github.base_ref ==  'develop-6.x.x' || github.base_ref ==  'master'
     steps:
       - uses: actions/checkout@v4
@@ -46,6 +50,23 @@ jobs:
           sleep 35s
       - name: Run tests
         run: bats --tap exist-docker/src/test/bats/*.bats
+
+      - name: debug logs
+        if: failure()
+        run: docker logs exist-ci | grep 'ERROR'
+
+      - name: Copy logs on failure
+        if: failure()
+        run: |
+          echo "Copying logs..."
+          docker cp exist-ci:/exist/logs/exist.log ./exist.log
+
+      - name: Upload logs artifact
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: exist-core-failed-log
+          path: exist.log
       # NOTE (DP): When on master push release, when on develop push latest: Version is included automatically
       # TODO (DP): Confirm that releases triggered from maven publish images with the non SNAPSHOT version
       - name: Publish latest images


dizzzz:
@wolfgangmm please can you check?

nverwer:
It appears there is another (different) problem with the way `fn:transform` handles its input, see #5682.

nverwer:
I have added two more commits.
* Show the location when an error occurs during an XSL transformation (because my client who uses `fn:transform` wanted this).
* Fix an error when transforming a document containing 'NodeProxy` nodes into an XdmValue for Saxon.

The second commit fixes another bug in the handling of `NodeProxy` instances.

nverwer:
There was another problem with the the 'path' returned by `treeIndex`.
The indexes are determined by counting previous-siblings.
It turns out that a `org.exist.dom.persistent.StoredNode` returns attributes of an element as previous siblings of the element's children.
When later the tree is traversed using the indexes, attributes are not in the `children` list.

I think that what `StoredNode` does is wrong, but decided to not try to solve it there (yet).
Instead, I made a work-around in `TreeUtils`.

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/TreeUtils.java
```diff
@@ -64,20 +65,41 @@ static StringBuilder pathTo(final Node node) {
...
+     */
+    private static Node previousSiblingNotAttribute(Node node) {
+      Node sibling = node.getPreviousSibling();
+      if (sibling instanceof Attr) {
+        return null;
+      }
+      return sibling;
+    }
+
     static XdmNode xdmNodeAtIndex(final XdmNode xdmNode, final List<Integer> index) {
```
DrRataplan, line 103: Should this method not be adapted? This loops over `children()`, are these not including attributes?

line-o:
Thanks @nverwer again for your contribution. We will pull it in once we have at least one positive and one negative test for it.
Any pointers you could give how to test this scenario would be much appreciated.

line-o:
Or maybe it is easier to test in XQuery?

duncdrum:
@nverwer thank you for the time you spend on this. We would like to include this PR in a patch release of exist. Could you rebase the PR please so, we can start with a cleaner commit history? I will try to help you with adding a test case. Did you succeed in the meantime in wiggling things down? Could you share the `xsl` that triggered the NPE?

nverwer:
> @nverwer Could you rebase the PR please so, we can start with a cleaner commit history?

What should I rebase this on? EDIT: I see, it is develop-6.x.x.

About a test case: We had the problem several months ago in a very large application. I could not make a concise test case back then, and it would be even more difficult now. Designing a test case requires a thorough understanding of the internal representations of XML in eXist-db, which I do not have.

I apologize for not coming up with a useful test. I am aware that eXist can only advance by the efforts of the community around it.
My main concern is to keep our application running, and this PR has helped us to use `fn:transform` instead of `transform:transform`.

nverwer:
After spending some time trying to rebase, it all seems to be terribly messed up, even resulting in automatically closing this PR. There are close to a 100 files with merge conflicts, most of which are totally unrelated to this PR. I have no idea why this went horribly wrong. I will give it one more try before giving up.

nverwer:
I tried to apply the changes to develop-6.x.x and will reopen this PR.

duncdrum:
@nverwer thank you, gave me quite a scare there for a minute

nverwer:
> @nverwer thank you, gave me quite a scare there for a minute
I should not have pushed prematurely. :fearful: 
I could not reconstruct exactly what was in the original pull request, but it does include the fixes for the NPE. I added the code that allows maps and arrays to be passed as parameters to an XSL transformation. I probably should have made that into a separate pull request. :grimacing:

nverwer:
In the meantime, my version of develop-6.x.x evolved, and I pushed it. I seems that this pull request was botched as a result.
Since nothing has happened with this pull request, I propose that it is closed without merging.

nverwer:
This PR became a mess after I changed my develop-6.x.x branch a few times, and these changes became visible here.
I got fed up with this PR hanging around without any progress, and closed it.
Then I decided to once more find the actual changes, and made a new PR: #5882

PR #5709, opened 2025-04-15
[library] upgrade pkg-java library

dizzzz:
update pkg-java to most recent version, as addresses potential XAR loading issues for windows (10).

note it does not fix all issues, but there are significant improvements

### DIFF
--- a/exist-parent/pom.xml
+++ b/exist-parent/pom.xml
@@ -396,7 +396,7 @@
             <dependency>
                 <groupId>org.expath.packaging</groupId>
                 <artifactId>pkg-java</artifactId>
-                <version>2.0.1</version>
+                <version>2.1.1</version>
             </dependency>
 
             <dependency>


duncdrum:
@dizzzz so despite this https://github.com/eXist-db/exist/pull/5335#issuecomment-2178714945 ?

adamretter:
FYI - This is not compatible with eXist-db as it requires Saxon 12.

dizzzz:
Ow interesting......

dizzzz:
Hi @adamretter thanks for your input;

Re re-checked the repo ( https://github.com/expath/expath-pkg-java/blob/main/pkg-java/pom.xml ) but I do not see any obvious reference to Saxon; I do see some XSLs but these are al marked version v1.0;

please could you elaborate why Saxon 12 is needed?

adamretter:
@dizzzz If you follow the parent chain you will find it...

PR #5723, opened 2025-05-02
[build] only include specific props into izpack installer

line-o:
We would otherwise potentially leak secrets with the next release


duncdrum (approved):
㊙️

PR #5725, opened 2025-05-04
[fix] replace test SSL certificate

line-o:
### Description:

The current bundled certs for testing SSL connections have expired several years ago.
They are replaced with a self-signed cert created with exist-jetty-config/src/main/scripts/create-keystore.sh
That way
1. it is clear these certs cannot be used for production
2. the path to re-new certificates is clearly documented

### Type of tests:

The entire xst test suite does pass with the new test certificate. The only change was the error message (see https://github.com/eXist-db/xst/pull/266)


duncdrum (commented):
> the path to re-new certificates is clearly documented 

is it? I'd prefer to add a line or two to the release documentation about the cert.

dizzzz (approved):

duncdrum:
see #3962

PR #5740, opened 2025-05-14
Align DOM implementations with org.w3c.dom

### LINKED ISSUE #5672: [BUG] org.w3c.dom.Element#getAttribute can't return null
according to the javadoc org.w3c.dom.Element#getAttribute can not return null;

we have 2 implementations:

Image

however org.exist.dom.memtree.ElementImpl#getAttribute can return null, 
as where org.exist.dom.persistent.ElementImpl#getAttribute correctly returns ""

line-o:
### Description:

This PR fixes inconsistencies in
- org.exist.dom.memtree.ElementImpl
- org.exist.dom.persistent.ElementImpl

with org.w3c.dom.ElementImpl

Because of that _a lot_ of code throughout exist-db core _and_ its extensions needed to be adapted.
So this is definitely a breaking change that developers of Java extensions need to take into account.

On the plus side this also enables us to get rid of a lot of extra checks that were necessary before.
As a result the dependency on apache.commons.lang3.StringUtils could be dropped in several modules.

### Reference:

fixes #5672

### Type of tests:

None added.


extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/MessageListFunctions.java
```diff
@@ -476,68 +475,56 @@ private SearchTerm parseSearchTerms( Node terms ) throws XPathException
...
+			} else {
+				throw new XPathException(this, "Invalid Search Term type specified: " + type );
+			}
+		}
 		
-		if( st == null ) {
-			throw( new XPathException(this, "Invalid Search Terms specified" ) );
+		if (st == null) {
+			throw new XPathException(this, "Invalid Search Terms specified" );
 		}
```
reinhapa, line 513: Maybe turn into a an switch statement. Also the null check seems to be obsolete here as the `getAttribute` should not return `null`

extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/MessageListFunctions.java
```diff
@@ -557,227 +544,166 @@ private SearchTerm[] parseChildSearchTerms( Node terms ) throws XPathException
...
+		if ("to".equalsIgnoreCase(type) ) {
+			rtype = Message.RecipientType.TO;
+		} else if( "cc".equalsIgnoreCase(type) ) {
+			rtype = Message.RecipientType.CC;
+		} else if( "bcc".equalsIgnoreCase(type) ) {
+			rtype = Message.RecipientType.BCC;
 		} else {
-			throw( new XPathException(this, "Pattern attribute must be specified for term with type: " + ((Element)terms).getAttribute( "type" ) ) );
+			throw new XPathException(this, "Invalid recipientType: " + type + ", for term with type: " + ((Element)terms).getAttribute("type"));
 		}
```
reinhapa, line 596: Here also a switch as function could be used to simplify the logic...

extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/MessageListFunctions.java
```diff
@@ -557,227 +544,166 @@ private SearchTerm[] parseChildSearchTerms( Node terms ) throws XPathException
...
+		} else if( "draft".equalsIgnoreCase(flag) ) {
+			flags = new Flags( Flags.Flag.DRAFT );
+		} else if( "recent".equalsIgnoreCase(flag) ) {
+			flags = new Flags( Flags.Flag.RECENT );
+		} else if( "seen".equalsIgnoreCase(flag) ) {
+			flags = new Flags( Flags.Flag.SEEN );
 		} else {
-			throw( new XPathException(this, "flag attribute must be specified for term with type: " + ((Element)terms).getAttribute( "type" ) ) );
+			throw new XPathException(this, "Invalid flag: " + flag + ", for term with type: " + ((Element)terms).getAttribute("type"));
 		}
```
reinhapa, line 637: Here switch as function would make sense

extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/MessageListFunctions.java
```diff
@@ -557,227 +544,166 @@ private SearchTerm[] parseChildSearchTerms( Node terms ) throws XPathException
...
+			cp = ComparisonTerm.LT;
+		} else if( "ne".equalsIgnoreCase(comp) ) {
+			cp = ComparisonTerm.NE;
 		} else {
-			throw( new XPathException(this, "comparison attribute must be specified for term with type: " + ((Element)terms).getAttribute( "type" ) ) );
+			throw( new XPathException(this, "Invalid comparison: " + comp + ", for term with type: " + ((Element)terms).getAttribute( "type" ) ) );
 		}
-		
-		return( cp );
+		return cp;
```
reinhapa, line 707: .. switch as function

reinhapa (changes requested):
some parts could be converted to switch expressions

adamretter:
Interestingly, this work looks very similar to the previous work done in Elemental on March 17 2025, see here - https://github.com/evolvedbinary/elemental/commit/f8a2781ecab67ff7b64cb111c940099139f5c70e#diff-5b0c3e472fee91090782d0d1535d50f290ac6f450e8fbe68b30c9dcfad1b6505

I have no problem with you (or anyone else) taking code from Elemental, as long as you follow the license terms. That particular code in Elemental is partly under the LGPL 2.1 ONLY license with an 'Evolved Binary' Copyright. This is not the same license used by eXist-db (which is straight LGPL 2.1 (without the ONLY clause)), and it is not the same Copyright.

If you wish to reuse Elemental code in eXist-db, please do, but you must follow the terms of the LGPL, so as I understand it, at a minimum, you would need to include our license header too in each file where you use our code, and a statement from yourselves if you have made any modifications to our code. You may also need to seek guidance on combining 'LGPL 2.1 ONLY' and 'LGPL 2.1' code as I suspect the outcome of that product is 'LGPL 2.1 ONLY' which is not the license that eXist-db currently claims.

extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/MessageListFunctions.java
line-o: @reinhapa yes, there is still a lot of code that can be optimised. But given the amount of code that is changed in this PR I would propose to do it in a separate PR

extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/MessageListFunctions.java
line-o: I propose to change this in a followup PR

extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/MessageListFunctions.java
line-o: I propose to change this in a followup PR

extensions/modules/mail/src/main/java/org/exist/xquery/modules/mail/MessageListFunctions.java
line-o: I propose to change this in a followup PR

adamretter:
> The license changes in the referenced commit are questionable

@line-o I don't yet understand what you mean by "questionable". Do you have questions about them? or do you believe there is a problem with our approach to licensing? We have worked hard to make sure we have followed and complied with the terms of the LGPL 2.1 license. If you believe that there is an issue with our licensing, please feel free to open an issue at https://github.com/evolvedbinary/elemental/issues with the specific details of your concerns in reference to the LGPL license terms.

adamretter:
> I would be very careful pointing to different licenses

@reinhapa Can you tell me what I need to be careful of please? The code for eXist-db is 'LGPL 2.1', the modifications in Elemental 6.4.0 and 7.0.0 are 'LGPL 2.1 ONLY'. I don't see how clearly pointing that out should upset anyone, in fact quite to the contrary, it is important for both developers and users to understand the licensing of the software they contribute to and/or use.

> and accusing @line-o to have copied from Elemental here.

I have not made any accusations, I have simply made an observation. If you compare the approach taken, and then also the git diffs you will likely find, as I did, that they are almost identical.

> Changing the implementation in order to correctly implement an external defined API will naturally result in similar changes that might also exist in Elemental.

I don't disagree... but, there is more to it than just that here.

> Also: How comes that Elemental code contains eXist Source but not the GIT history that lead to it?

Elemental does contain the entire Git History of eXist-db - `git log` will show you that. What do you believe is missing?

dizzzz (approved):

reinhapa (approved):

duncdrum:
@line-o can you rebase this. The container test shows an error in `exist.log` we should check that first.

duncdrum:
@line-o the logs aren't uploaded anymore automatically we should bring that back. 

You would have to build locally start the container and inspect its 'exist.log' There is an 'ERROR' that wasn't there before.

exist-core/src/main/java/org/exist/repo/Deployment.java
```diff
@@ -914,6 +881,37 @@ public Optional<DocumentImpl> getDescriptor(final DBBroker broker, final XarSour
...
+        final Optional<String> group;
+        final Either<Integer, String> permissions;
+
+        private RequestedPerms(final String user, final String password, final Optional<String> group, final Either<Integer, String> permissions) {
+            this.user = user;
+            this.password = password;
+            this.group = group;
+            this.permissions = permissions;
+        }
+    }
```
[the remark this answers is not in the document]
dizzzz: makes sense, but I thought a 'switch' change would be in an upcoming PR... but I am wrong as I don't see a remark here.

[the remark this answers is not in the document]
line-o: Yes let's not do too much in one PR.

PR #5782, opened 2025-06-30
[bugfix] Repair JMXservlet for ipv6 addresses.

dizzzz:
Repair JMXservlet for ipv6 addresses.

Interestingly the servlet container returns [0:0:0:0:0:1] as string.

### DIFF
--- a/exist-core/src/main/java/org/exist/management/client/JMXServlet.java
+++ b/exist-core/src/main/java/org/exist/management/client/JMXServlet.java
@@ -32,6 +32,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.*;
+import java.util.function.Predicate;
 import javax.management.*;
 import jakarta.servlet.ServletConfig;
 import jakarta.servlet.ServletException;
@@ -92,7 +93,7 @@ public class JMXServlet extends HttpServlet {
 
     private Path dataDir;
     private Path tokenFile;
-    
+
     @Override
     protected void doGet(final HttpServletRequest request, final HttpServletResponse response) throws ServletException, IOException {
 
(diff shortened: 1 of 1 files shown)


line-o (commented):
Looks good! I want to play around with it before I give my approval.

exist-core/src/main/java/org/exist/management/client/JMXServlet.java
```diff
@@ -218,32 +214,34 @@ public void init(ServletConfig config) throws ServletException {
...
-            for (InetAddress address : Inet6Address.getAllByName("localhost") ){
-                localhostAddresses.add(address.getHostAddress());
             }
-        } catch (UnknownHostException ex) {
-            LOG.warn("Unable to retrieve ipaddresses (v6) for localhost: {}", ex.getMessage());
+            }
+
+        } catch (final SocketException e) {
+            LOG.error("Unable to determine localhost loopback addresses: {}  ", e.getMessage(), e);
         }
```
reinhapa: Could be rewritten using streams and a simple additional method:
```
NetworkInterface.networkInterfaces()
    .filter(SimpleTest::isLoopback)
    .flatMap(NetworkInterface::inetAddresses)
    .map(InetAddress::getHostAddress)
    .map(a -> a.replaceFirst("%.+", ""))
    .forEach(localhostAddresses::add);
```
Additonal Method would be:
```
static boolean isLoopback(NetworkInterface ne) {
    try {
        return ne.isLoopback();
    } catch (SocketException e) {
        LOG.error("Unable to determine if {} is a loopback interface", ne, e);
        return false;
    }
}
```

reinhapa (commented):
Could be written simpler, given that  we require Java > 9

exist-core/src/main/java/org/exist/management/client/JMXServlet.java
dizzzz: @reinhapa I doubted about this due to readability with the "%" but your suggestion really looks  clean!

dizzzz:
I agree that a test would be very nice.... but I had no sustainable idea how to test that, as potentially the loopback IP-adresses vary from system to os to ....

Issue #5681, opened 2025-03-18
[BUG] failing transformer test for develop


dizzzz:
needs investigation (develop)

```
Error: core] [ERROR] Tests run: 754, Failures: 1, Errors: 0, Skipped: 4, Time elapsed: 13.35 s <<< FAILURE! -- in xquery.xquery3.XQuery3Tests
Error: core] [ERROR] xqts.org.exist-db.xquery.test.function_transform.transform-67 -- Time elapsed: 0.015 s <<< FAILURE!
org.junit.ComparisonFailure: expected:<[ (assertTrue failed.)]> but was:<[false]>
[exist-core] [INFO] 
[exist-core] [INFO] Results:
[exist-core] [INFO] 
Error: core] [ERROR] Failures: 
Error: core] [ERROR]   expected:<[ (assertTrue failed.)]> but was:<[false]>
[exist-core] [INFO] 
Error: core] [ERROR] Tests run: 5784, Failures: 1, Errors: 0, Skipped: 105
[exist-core] [INFO] 
```

looks reproducible, maybe timing issue

line-o:
It looks like a race condition. This transformer test 67 is flaky since a long time. Re running it several times will make it pass eventually. I would be open to skip it if we can't make it pass reliably. This is of course only possible after making sure there is no real issue.

line-o:
same goes for transform tests 66 and 11-2

duncdrum:
would be good to investigate after  #6036

line-o:
This was fixed by my rewrites of the transformer tests https://github.com/eXist-db/exist/pull/6546

PR #5788, opened 2025-07-07
fix: version in two submodule poms

line-o:
### Description:

Building is broken because the commit to prepare for the next development iteration does not update two submodules by default. This is a separate issue.

This PR just updates the submodules to match the others and allows to build again.


line-o:
@duncdrum the transform tests are flaky on windows see also #5681

duncdrum (approved):

PR #5814, opened 2025-07-23
Bugfix/6xx allow j7 and j11 builds pt2

dizzzz:
An attempt to support scenario: build Java11 run Java8 ;

probably not relevant anymore as dev focus is as agreed on develop/v7

### DIFF
--- a/exist-core-jmh/pom.xml
+++ b/exist-core-jmh/pom.xml
@@ -125,6 +125,7 @@
                     <compilerVersion>${project.build.target}</compilerVersion>
                     <source>${project.build.source}</source>
                     <target>${project.build.target}</target>
+                    <release>${maven.compiler.release}</release>
                 </configuration>
             </plugin>
             <plugin>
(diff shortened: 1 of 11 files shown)


adamretter:
Here is one of the problems with this that I have spoken about...
```
org.exist.EXistException: java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer;
        at org.exist.storage.BrokerPool.initialize(BrokerPool.java:464)
        at org.exist.storage.BrokerPools.configure(BrokerPools.java:177)
        at org.exist.storage.BrokerPools.configure(BrokerPools.java:112)
        at org.exist.jetty.JettyStart.run(JettyStart.java:217)
        at org.exist.jetty.JettyStart.main(JettyStart.java:103)
```

Also, additional runtime problems exist if you use some of the `extensions/modules`.

reinhapa:
But as long as you not use methods only available in later versions, you would not have such issues, even when compiled on a later JDK using Java 8 compliance Or do I miss a point therre?

adamretter:
@dizzzz https://en.wikipedia.org/wiki/Application_binary_interface

dizzzz:
As discussed.

PR #5817, opened 2025-07-27
Reinstate integration tests in maven

### LINKED ISSUE #5805: [BUG] No XQSuite tests are executed
When the full test suits is run on the `develop` branch, none of the XSuite unit tests (i.e. those written in XML or XQuery) are executed. This can be observed in the output of the test suite run where you will see messages like:

```
Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 12.15 s -- in xquery.numbers.NumberTests
```

This has also been the case in your CI on the `develop` branch https://github.com/eXist-db/exist/actions/runs/16343506189/job/46171471898#step:6:2720 since March 2025.

line-o:
The latest release of maven-surefire-plugin lead to a large amount of integration tests not running.

Integration test for us means tests written in XQuery and also those defined XML.
I could confirm that deliberately breaking a test in core did not lead to a failing build.

**NOTE:** The build of this PR will break because some tests are really broken.
Those tests will be fixed in separate PRs.

fixes #5805 

possibly related issues from the maven plugins in us: 
- https://github.com/apache/maven-surefire/issues/3156
- https://github.com/apache/maven-surefire/issues/2634

### DIFF
--- a/exist-core/src/main/java/org/exist/test/runner/XMLTestRunner.java
+++ b/exist-core/src/main/java/org/exist/test/runner/XMLTestRunner.java
@@ -142,7 +142,7 @@ public class XMLTestRunner extends AbstractTestRunner {
 
     private static @Nullable String getIdValue(final Node test) {
         final String id = ((Element)test).getAttribute("id");
-        return id.isEmpty() ? null : id;
+        return id.isBlank() ? null : id;
     }
 
     private static @Nullable String getTaskText(final Node test) {
--- a/exist-parent/pom.xml
+++ b/exist-parent/pom.xml
@@ -807,7 +807,8 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>3.5.3</version>
+                    <!-- NOTE: version 3.5.3 fails to run XSuite tests -->
+                    <version>3.5.2</version>
                     <dependencies>
                         <dependency>
                             <groupId>org.junit.platform</groupId>


adamretter:
> Root cause: https://github.com/apache/maven-surefire/issues/3156

@line-o that's not the correct root cause. https://github.com/apache/maven-surefire/issues/3156 is about the maven-failsafe-plugin, that is not used for XSuite tests in eXist-db. The correct root cause is actually this one: https://github.com/apache/maven-surefire/issues/2634

exist-core/src/main/java/org/exist/test/runner/XMLTestRunner.java
```diff
@@ -142,7 +142,7 @@ private static XMLTestInfo extractTestInfo(final Path path, final Document doc)
 
     private static @Nullable String getIdValue(final Node test) {
         final String id = ((Element)test).getAttribute("id");
-        return id.isEmpty() ? null : id;
+        return id.isBlank() ? null : id;
```
adamretter, line 145: This isn't needed.

line-o:
On the other hand the problem did not surface in 3.5.2 whereas https://github.com/apache/maven-surefire/issues/2634 looks a lot older and mentions 3.0.0 instead.

exist-core/src/main/java/org/exist/test/runner/XMLTestRunner.java
line-o: It is an extra little bit I saw I got wrong earlier when changing attribute implementation to be compliant with w3c. I think it does not hurt to keep it in this PR but more people think it should not be here then I am fine moving it.

exist-parent/pom.xml
```diff
@@ -807,7 +807,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>3.5.3</version>
+                    <version>3.5.2</version>
```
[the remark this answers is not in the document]
line-o: Good idea.

dizzzz (approved):

exist-core/src/main/java/org/exist/test/runner/XMLTestRunner.java
adamretter: @line-o This could actually cause an issue, in so far as it would hide errors in the test case implementation.

line-o: so, instead of returning null you would say it is better to return the string containing only space characters?

adamretter: > so, instead of returning null you would say it is better to return the string containing only space characters

@line-o No. If there is a string that is only whitespace characters, then an error should be raised.

PR #5824, opened 2025-07-30
[bugfix] consistently raise errors in fn:analyze-string

line-o:
# Description

`fn:analyze-string` was implemented such that an invalid pattern or flag or a pattern matching an empty string would only raise an error if the value was a non-empty string.

With this PR applied the error will be raised regardless of the provided value.

# References

> Error Conditions
> A dynamic error is raised [[err:FORX0002](https://www.w3.org/TR/xpath-functions-31/#ERRFORX0002)] if the value of $pattern is invalid according to the rules described in section [5.6.1 Regular expression syntax](https://www.w3.org/TR/xpath-functions-31/#regex-syntax).
>
> A dynamic error is raised [[err:FORX0001](https://www.w3.org/TR/xpath-functions-31/#ERRFORX0001)] if the value of $flags is invalid according to the rules described in section [5.6.2 Flags](https://www.w3.org/TR/xpath-functions-31/#flags).
>
> A dynamic error is raised [[err:FORX0003](https://www.w3.org/TR/xpath-functions-31/#ERRFORX0003)] if the supplied $pattern matches a zero-length string, that is, if fn:matches("", $pattern, $flags) returns true.

source: https://www.w3.org/TR/xpath-functions-31/#func-analyze-string

# Tests

XQSuite tests added

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
@@ -103,18 +103,15 @@ public class FunAnalyzeString extends BasicFunction {
             final MemTreeBuilder builder = context.getDocumentBuilder();
             builder.startDocument();
             builder.startElement(new QName("analyze-string-result", Function.BUILTIN_FUNCTION_NS), null);
-            String input = "";
-            if (!args[0].isEmpty()) {
-                input = args[0].itemAt(0).getStringValue();
+
+            final String input = args[0].isEmpty() ? "" : args[0].itemAt(0).getStringValue();
+            final String pattern = args[1].itemAt(0).getStringValue();
+
+            String flags = "";
+            if (args.length == 3) {
+                flags = args[2].itemAt(0).getStringValue();
             }
-            if (input != null && !input.isEmpty()) {
-                final String pattern = args[1].itemAt(0).getStringValue();
-                String flags = "";
-                if (args.length == 3) {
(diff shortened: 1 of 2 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
```diff
@@ -103,18 +103,15 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
...
-            }
-            if (input != null && !input.isEmpty()) {
-                final String pattern = args[1].itemAt(0).getStringValue();
-                String flags = "";
-                if (args.length == 3) {
-                    flags = args[2].itemAt(0).getStringValue();
-                }
-                analyzeString(builder, input, pattern, flags);
+
+            final String input = args[0].isEmpty() ? "" : args[0].itemAt(0).getStringValue();
```
dizzzz, line 107: StingUtils.isBlank() would make this more readable. Fee NPE check :-)

line-o: In this specific function and this specific argument we should under no circumstances modify the string value.
as " " does match `\s+` but "" does not
:)

line-o: Also, let's try to get rid of StringUtils where possible

reinhapa (approved):

PR #5830, opened 2025-08-05
[fix] always serialize statistics

### LINKED ISSUE #5829: [BUG] system:trace looses its data after disabling tracing
**Describe the bug**

In order to inspect and debug calls to the database one needs to 
- enable tracing
- do some things
- disable tracing
- show the recorded trace

Unfortunately, with exist 7.0.0-SNAPSHOT this no longer works as all data is lost when tracing is disabled

see [exist-db/monex/#314](https://github.com/eXist-db/monex/issues/314)

**Expected behavior**

The tracing data to be retained until `system:clear-trace()` is called

**To Reproduce**

```Xquery
xquery version "3.1";

declare namespace stats="http://exist-db.org/xquery/profiling";

declare
    %test:assertTrue
function t:test() {
  let $_ := system:enable-tracing(true())
  let $_ := util:uuid()
  let $_ := system:enable-tracing(false())
  return exists(system:trace()/stats:calls/stats:*)
};
```

- Build: eXist-7.0.0-SNAPSHOT (d7b9c4ae81c4e53c737299a6c349fdaaab9514a0)
- Java: 21.0.6 (Azul Systems, Inc.)
- OS: Mac OS X 15.5 (aarch64)

**Additional context**

reinhapa:
Always serialize statistics data in order to have access to the last values.

fixes #5829

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/PerformanceStatsImpl.java
+++ b/exist-core/src/main/java/org/exist/xquery/PerformanceStatsImpl.java
@@ -48,6 +48,13 @@ import java.util.Set;
  */
 @NotThreadSafe
 public class PerformanceStatsImpl implements PerformanceStats {
+    public static final String CDATA = "CDATA";
+    public static final String SOURCE = "source";
+    public static final String ELAPSED = "elapsed";
+    public static final String CALLS = "calls";
+    public static final String QUERY = "query";
+    public static final String NAME = "name";
+    public static final String TYPE = "type";
 
     private static class IndexStats {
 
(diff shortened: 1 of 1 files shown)


line-o (approved):
Nice! Is there a specific reason you decided to handle "optimization " and "~-level" differently?

reinhapa:
> Nice! Is there a specific reason you decided to handle "optimization " and "~-level" differently?

@line-o I did not touch this part. You're right this seems to be is wrong and should be either `optimization` or `optimization-level` for both `localName` and `qName`.  Can you tell me what attribute do you use actually?

dizzzz:
I remember code changes relevant on this some time ago?

dizzzz (approved):

line-o:
Those attributes changed recently (at least their content). An attribute whose local name does not match its QName is very odd indeed.

dizzzz:
https://github.com/eXist-db/exist/pull/4910

reinhapa:
> #4910

@dizzzz should I change both to `optimization-level` ?

dizzzz:
Not sure, I'd need to dive a bit deeper...

dizzzz:
We might want to revert it to keep compatibility with v6?  Need to check a bit more...

https://github.com/eXist-db/monex/issues/273

line-o:
From all I read in the related issues and PRs we should go with `@optimization`

line-o:
And we should see it blow up immediately if range index tests expect `@optimization-level` instead (I doubt it)

reinhapa:
> And we should see it blow up immediately if range index tests expect `@optimization-level` instead (I doubt it)

When looking into [expressions.xqm](https://github.com/eXist-db/exist/blob/7a5025b9f5a28a168a01f97987bd1201ce93aa7e/exist-core/src/test/xquery/optimizer/expressions.xqm#L92) it tends to be `@optimization-level`

line-o:
Can I merge @reinhapa ?

PR #5856, opened 2025-09-01
[feature] use caffeine library as cache for XML grammar

dizzzz:
**Summary:**
Use caffeine library as cache for XML grammars (DTDs, XSDs, ....)

**Description:**
- the main change is in `GrammarPool.java` - it uses caffeine now instead of the Xerces provided one.
  - the one of Xerces does not (automatically) clean up old grammars.
  - tests are added into `exist-core/src/test/java/org/exist/validation/GrammarPoolTest.java`
- `Configuration.java` has been changed so settings are read from `conf.xml`.
  - some old methods are split-up for readability.
  - if config element is left out, old behavior is there, with new library.
- all code in `org.exist.validation` has been reformatted, and with static quality checks improved.

### DIFF
--- a/exist-distribution/src/main/config/conf.xml
+++ b/exist-distribution/src/main/config/conf.xml
@@ -25,6 +25,8 @@
             serializer
             transformer
             validation
+                entity-resolver
+                grammar-cache
             xquery
                 builtin-modules
                     module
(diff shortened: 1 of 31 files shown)


reinhapa (approved):

exist-distribution/src/main/config/conf.xml
duncdrum, line 1: The code listing for `conf.xml` in the documentation app needs to be updated.

duncdrum (changes requested):
Looking good. Needs documentation, otherwise good to merge.

duncdrum (approved):

PR #5881, opened 2025-09-23
Make unique query history keys for monex

nverwer:
This PR solves https://github.com/eXist-db/monex/issues/299

In the "recent queries", Monex shows only one query for a source URL, even if there have been more queries with the same URL.
The reason is that  the `QueryKey` class has a method `int compareTo(final QueryKey other)` that does not take all parts of the query key into account.
This PR cahnges the `compareTo` method, so that the whole key is compared.

See also the analysis in https://github.com/eXist-db/monex/issues/299#issuecomment-3322905361

### DIFF
--- a/exist-core/src/main/java/org/exist/management/impl/ProcessReportMXBean.java
+++ b/exist-core/src/main/java/org/exist/management/impl/ProcessReportMXBean.java
@@ -123,7 +123,9 @@ ProcessReportMXBean extends PerInstanceMBean {
             if (other == null) {
                 return 1;
             }
-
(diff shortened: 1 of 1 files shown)


dizzzz (approved):

dizzzz:
apologies for reviewing so late, I just returned from a summer break, I missed the PR

reinhapa (approved):

line-o:
Thanks a ton @nverwer for your contribution. Ideally, we would like to see this tested as well before pulling it in. But this will have to be done later in the develop port.

PR #5882, opened 2025-09-23
[bugfix] fn:transform: Conversion and treeIndex problems

nverwer:
# Description

This is a retry for #5680 which was botched after I changed my develop-6.x.x branch.

The `fn:transform` function sometimes produced `NullPointerException`s. This happened for documents containing nodes of type `org.exist.dom.memtree.ElementImpl` and other subtypes of `org.exist.dom.memtree.NodeImpl`.

The function `org.exist.xquery.functions.fn.transform.Convert.ToSaxon.ofNode(Node)` uses `org.exist.xquery.functions.fn.transform.TreeUtils.treeIndex(Node)` to get a 'path' (as a list of indexes) of the `Node` in its containing document. This 'path' is used to find the node in a newly built document that is suitable for use by Saxon.
Using the 'path' that the `treeIndex` function produces assumes that it starts at the document node.

However, some nodes do not have a containing document. Specifically, in `org.exist.dom.memtree.NodeImpl.getParentNode()`:
```
if (parent.getNodeType() == DOCUMENT_NODE && !((DocumentImpl) parent).isExplicitlyCreated()) {
            /*
                All nodes in the MemTree will return an Owner document due to how the MemTree is implemented,
                however the explicitlyCreated flag tells us whether there "really" was a Document Node or not.
                See https://github.com/eXist-db/exist/issues/1463
             */
            return null;
```
In this case, the 'path' returned by `treeIndex` does not contain the index (0) for the root node, and `org.exist.xquery.functions.fn.transform.TreeUtils.xdmNodeAtIndex(XdmNode, List<Integer>)` will return `null` when a node is expected.

The proposed fix tests for this case. Going up the ancestor chain, when a node that is not a document node and that does not have a parent is found, the index 0 for this root element is inserted in the 'path'.

The PR also adds support for using maps and arrays as parameters in a XSLT transformation.

# Testing

The` NullPointerException` happened in a rather complicated XQuery with documents composed from several sources.
I have not yet been able to make a simple test case.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Convert.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Convert.java
@@ -26,17 +26,23 @@ import net.sf.saxon.s9api.*;
 import net.sf.saxon.type.BuiltInAtomicType;
 import org.exist.dom.QName;
 import org.exist.dom.memtree.DocumentImpl;
+import org.exist.dom.persistent.NodeProxy;
 import org.exist.xquery.ErrorCodes;
 import org.exist.xquery.XPathException;
 import org.exist.xquery.functions.array.ArrayType;
 import org.exist.xquery.functions.fn.FnTransform;
+import org.exist.xquery.functions.map.AbstractMapType;
 import org.exist.xquery.value.*;
 import org.w3c.dom.Document;
 import org.w3c.dom.Node;
 
+import io.lacuna.bifurcan.IEntry;
+
 import javax.xml.transform.dom.DOMSource;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Type conversion to and from Saxon
@@ -119,11 +125,18 @@ class Convert {
         }
 
         XdmValue of(final Item item) throws XPathException {
+            if (item instanceof NodeProxy) {
+              return ofNode(((NodeProxy) item).getNode());
+            }
             final int itemType = item.getType();
             if (Type.subTypeOf(itemType, Type.ATOMIC)) {
                 return ofAtomic((AtomicValue) item);
             } else if (Type.subTypeOf(itemType, Type.NODE)) {
                 return ofNode((Node) item);
+            } else if (Type.subTypeOf(itemType, Type.MAP)) {
+                return ofMap((AbstractMapType) item);
+            } else if (Type.subTypeOf(itemType,  Type.ARRAY)) {
+                return ofArray((ArrayType) item);
             }
             throw new XPathException(ErrorCodes.XPTY0004,
                     "Item " + item + " of type " + Type.getTypeName(itemType) + COULD_NOT_BE_CONVERTED + "XdmValue");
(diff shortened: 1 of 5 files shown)


dizzzz:
I recognize that under some conditions there is not a "DocumentNode"; I need to recall the context, it is some time ago.

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Convert.java
```diff
@@ -119,11 +125,18 @@ static net.sf.saxon.s9api.QName of(final QNameValue qName) {
         }
 
         XdmValue of(final Item item) throws XPathException {
+            if (item instanceof NodeProxy) {
+              return ofNode(((NodeProxy) item).getNode());
+            }
```
reinhapa, line 130: As of Java 21 you could use this:

```suggestion
            if (item instanceof NodeProxy nodeProxy) {
              return ofNode(nodeProxy.getNode());
            }
```

reinhapa (commented):
Just a small note...

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Convert.java
dizzzz: this PR is to 6.x.x

duncdrum:
results against 6.4.0

```xml
<testsuites>
    <testsuite package="http://exist-db.org/xquery/test/function_transform" timestamp="2026-02-17T12:55:05.283Z" tests="3" failures="0" errors="3" pending="0" time="PT0.006S">
        <testcase name="memtree-second-child" class="testTransform:memtree-second-child">
            <error type="java:java.lang.NullPointerException" message=""/>
        </testcase>
        <testcase name="stored-doc-with-attributes" class="testTransform:stored-doc-with-attributes">
            <error type="java:java.lang.ClassCastException" message="class org.exist.dom.persistent.NodeProxy cannot be cast to class org.w3c.dom.Node (org.exist.dom.persistent.NodeProxy is in unnamed module of loader 'app'; org.w3c.dom.Node is in module java.xml of loader 'bootstrap')"/>
        </testcase>
        <testcase name="stylesheet-params-map-array" class="testTransform:stylesheet-params-map-array">
            <error type="err:XPTY0004" message="It is a type error if, during the static analysis phase, an expression is found to have a static type that is not appropriate for the context in which the expression occurs, or during the dynamic evaluation phase, the dynamic type of a value does not match a required type as specified by the matching rules in 2.5.4 SequenceType Matching. Item [ x, y, z ] of type array(*) could not be converted to an eXist XdmValue"/>
        </testcase>
    </testsuite>
</testsuites>
```

if ci is green this is ready from my view

line-o (approved):
Thanks to you both @nverwer and @duncdrum  :tada:

duncdrum:
see #6055 

@nverwer many thanks for the fix. I added tests and created a forward port. During rebasing the commit author seems to be wrongly displayed here on GitHub. Many apologies, but I have no clue, it seems to be a display issues only. The commit log names you as the author of the fix.

duncdrum (approved):

dizzzz (approved):

PR #5934, opened 2025-11-10
[ci] require alignment of test plugins

reinhapa:
Ties maven plugins together, that need to be updated at once


duncdrum:
@reinhapa if these need to be updated together we should group them in the dependabot config as well. 

That way we get a single pr with multiple deps updated together.

PR #5935, opened 2025-11-10
Fix/debug docker image

line-o:
### Description:

Since November 3rd the base image used to create the DEBUG docker image no longer exists.
Therefore this PR switches the base image to ~~maven:3-eclipse-temurin-21 (the same one used by duncdrum/existdb:7.0.0-SNAPSHOT-DEBUG)~~ eclipse-temurin:21-jdk-noble (suggested by @duncdrum)

### Reference:

last failure https://github.com/eXist-db/exist/actions/runs/19256716288

### Type of tests:

Manual tests with the running container

### DIFF
--- a/exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
+++ b/exist-docker/src/main/resources-filtered/Dockerfile-DEBUG
@@ -20,7 +20,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #
 
-FROM openjdk:21-jdk-slim-bullseye
+FROM eclipse-temurin:21-jdk-noble
 RUN apt-get update && apt-get -y dist-upgrade
 RUN apt-get install -y expat fontconfig     # Install tools required by FOP
 


reinhapa (approved):

duncdrum (changes requested):
TBH I don't think the maven image is a good replacement here something like `eclipse-temurin:21-jdk-noble` makes more sense. The `:debug` images here and at `duncdrum` are very different.

reinhapa:
> TBH I don't think the maven image is a good replacement here something like `eclipse-temurin:21-jdk-noble` makes more sense. The `:debug` images here and at `duncdrum` are very different.

@duncdrum as you have not been at the call this was the best we have been able to come up with... 😄

line-o:
> TBH I don't think the maven image is a good replacement here something like eclipse-temurin:21-jdk-noble makes more sense. The :debug images here and at duncdrum are very different.

I would appreciate if you could expand on:

1. Why is the maven image a bad replacement?
2. How are the debug images different?

It might be even useful to provide the information in the Dockerfile itself

duncdrum:
What debug images are is actually covered on each read me. It's just a naming collision. Duncdrum is using the debug tag from distroless, and passing it on to the final image stage. 

Existdb uses a distroless compatible base image with extended deps and Linux env for use with JAC, shell scripts etc. 

I don't see why the maven image would have the same requirements for the container env as exist. Maybe it uses a jdk maybe a jre, maybe be it comes with a system shell maybe not, are it's FOP deps on the same version as our base image? Who knows are they even there?

So at a minimum we need versioned jdk and Debian based env. So we can assemble the rest.

line-o:
Although I can't find any information about what the DEBUG images are and how they differ I will take @duncdrum word for it and change the base image to what he suggested.

reinhapa (approved):

duncdrum (approved):

PR #5951, opened 2025-11-30
[feature] get rid of  appassembler-maven-plugin

dizzzz:
### Context
Until now, eXist used the outdated _appassembler-maven-plugin_ to create the startup code the application. It created a set of shell scripts, each invoked a boot loader, which read an XML file to find the relevant JAR files, created its own classloader and invoked the class `org.exist.start.Main` to get eXist started.  

As a drawback it was difficult to add an extra jar file or to apply an hotfix by updating one jar file.

Since java 8 or so the 'java' commands allows an * in the classpath. This PR effectively make use of this feature.

@duncdrum I think this might be useful for our docker deployments.

----

old content of `etc` directory:

image

old content of XML file:

```xml
<daemon>
  <id>client</id>
  <mainClass>org.exist.start.Main</mainClass>
  <classpath>
    <dependencies>
      <dependency>
        <groupId>org.exist-db</groupId>
        <artifactId>exist-distribution</artifactId>
        <version>7.0.0-SNAPSHOT</version>
        <relativePath>exist-distribution-7.0.0-SNAPSHOT.pom</relativePath>
      </dependency>
      
	  <!-- 1000's of generated lines -->
	  
      <dependency>
        <groupId>stax</groupId>
        <artifactId>stax-api</artifactId>
        <version>1.0.1</version>
        <relativePath>stax-api-1.0.1.jar</relativePath>
      </dependency>
    </dependencies>
  </classpath>
  <commandLineArguments>
    <commandLineArgument>client</commandLineArgument>
  </commandLineArguments>
  <configurationDirectory>etc</configurationDirectory>
  <repositoryName>lib</repositoryName>
  <licenseHeaderFile>/....../LGPL-21-license.txt</licenseHeaderFile>
</daemon>
```

### DIFF
--- a/exist-distribution/src/main/templates/template.bat
+++ b/exist-distribution/src/main/templates/template.bat
@@ -82,11 +82,18 @@ if NOT "%CLASSPATH_PREFIX%" == "" set CLASSPATH=%CLASSPATH_PREFIX%;%CLASSPATH%
 @REM Reaching here means variables are defined and arguments have been captured
 :endInit
 
+@REM Normalize classpath and property path quoting for Windows
+set "CLASSPATH=%BASEDIR%\etc;%REPO%\*"
+if NOT "%CLASSPATH_PREFIX%" == "" set "CLASSPATH=%CLASSPATH_PREFIX%;%CLASSPATH%"
+
 %JAVACMD% %JAVA_OPTS% -Xms128m -XX:+UseNUMA -XX:+UseZGC -XX:+UseStringDeduplication ^
-    -Dfile.encoding=UTF-8 -Dlog4j.configurationFile="%BASEDIR%"/etc/log4j2.xml ^
-    -Dexist.home="%BASEDIR%" -Dexist.configurationFile="%BASEDIR%"/etc/conf.xml ^
-    -Djetty.home="%BASEDIR%" -Dexist.jetty.config="%BASEDIR%"/etc/jetty/standard.enabled-jetty-configs ^
-    -classpath %CLASSPATH% org.exist.start.Main @PLACEHOLDER@ %CMD_LINE_ARGS%
+    -Dfile.encoding=UTF-8 ^
+    -Dlog4j.configurationFile="%BASEDIR%\etc\log4j2.xml" ^
+    -Dexist.home="%BASEDIR%" ^
+    -Dexist.configurationFile="%BASEDIR%\etc\conf.xml" ^
+    -Djetty.home="%BASEDIR%" ^
+    -Dexist.jetty.config="%BASEDIR%\etc\jetty\standard.enabled-jetty-configs" ^
+    -classpath "%CLASSPATH%" org.exist.start.Main @PLACEHOLDER@ %CMD_LINE_ARGS%
 if %ERRORLEVEL% NEQ 0 goto error
 goto end
 
--- a/exist-distribution/src/main/templates/template.bat
+++ b/exist-distribution/src/main/templates/template.bat
@@ -75,7 +75,7 @@ set EXIST_COMMAND=
 if /i "%SCRIPT_NAME%"=="client" set EXIST_COMMAND=client
 if /i "%SCRIPT_NAME%"=="backup" set EXIST_COMMAND=backup
 if /i "%SCRIPT_NAME%"=="export-gui" set EXIST_COMMAND=export-gui
-if /i "%SCRIPT_NAME%"=="jmx-client" set EXIST_COMMAND=jmx-client
+if /i "%SCRIPT_NAME%"=="jmxclient" set EXIST_COMMAND=jmxclient
 if /i "%SCRIPT_NAME%"=="launcher" set EXIST_COMMAND=launcher
 if /i "%SCRIPT_NAME%"=="shutdown" set EXIST_COMMAND=shutdown
 if /i "%SCRIPT_NAME%"=="startup" set EXIST_COMMAND=jetty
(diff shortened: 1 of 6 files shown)


dizzzz:
I might need some help on the quality reporting. I tried the implement the suggestions but it resulted in that the main class could not be found.

dizzzz:
- [x] check build log, directory "log" is missing, probably more
- [x] make JAVA_OPTS configurable
- [x] introduce EXIST_OPTS

duncdrum:
very cool, a bit too much to think about for the comments here. Yes wildcard works even with J8, been running it for 4 years now. We would need to keep an eye on the backup, recovery, etc workflows on all three OSs to make sure they still work. We have no integration tests for this, so we need to do it manually. 

The Dockerfiles here and at `duncdrum/existdb` will need adjustments. 

I really like the idea of improvements to modifying `JAVA_TOOL_OPTIONS` and introducing `EXIST_OPTS`. Note that for java 21 you can use `JDK_JAVA_OPTIONS` to get some compostability going see [tp-app](https://github.com/eeditiones/tei-publisher-app/blob/master/Dockerfile#L88C5-L88C21) as far as I understand it `JAVA_OPTS` is not a thing.

dizzzz:
I have seen 'JAVA_OPTS' mostly in old scripts, it looks like it was a pattern once used in the [tomcat ](https://stackoverflow.com/questions/5241743/how-do-i-use-the-java-opts-environment-variable) scripts. 

setting this details in an environment variable will will make the scripts more clean and readable.

I will revise the PR with the ideas.

adamretter:
> Since java 8 or so the 'java' commands allows an * in the classpath. This PR effectively make use of this feature.

That could open up a security issue, as an attacker can just drop a jar file into a folder.

> I have seen 'JAVA_OPTS' mostly in old scripts, it looks like it was a pattern once used in the [tomcat ](https://stackoverflow.com/questions/5241743/how-do-i-use-the-java-opts-environment-variable)scripts

JAVA_OPTS is pretty much defacto for all Java applications these days, and has been for many years. Users and developers have come to expect it.

> JAVA_TOOL_OPTIONS
> JDK_JAVA_OPTIONS

You should avoid these as they have much wider implications.

dizzzz:
I considered the potential security issue already. When the variables are locally defined, and the java command is the last one in sequence, there should be no side effects IMO. If one has access for writing access anyway, an existing jar file (or the xml) can be updated any way. The security risk is merely changed not different.

I checked a bit more on 'JAVA_OPTS' it seems to be originated from the Catalina scripts, hence an implementation decision? I will check a bit more though. I will follow this pattern.

dizzzz:
note 1: to myself: as symlink alternative

```
<echo file="${assemble.dir}/bin/backup.bat">@echo off
call "%~dp0template.bat" %*
</echo>
```

note 2: eXist-db.app does not start

exist-distribution/src/main/templates/template.bat
```diff
@@ -0,0 +1,141 @@
...
+
+@REM Get the script name and determine the command to run
+set SCRIPT_NAME=%~n0
+set EXIST_COMMAND=
+
+@REM Check script name and set appropriate command
+if /i "%SCRIPT_NAME%"=="client" set EXIST_COMMAND=client
+if /i "%SCRIPT_NAME%"=="backup" set EXIST_COMMAND=backup
+if /i "%SCRIPT_NAME%"=="export-gui" set EXIST_COMMAND=export-gui
+if /i "%SCRIPT_NAME%"=="jmxclient" set EXIST_COMMAND=jmxclient
```
dizzzz, line 78: not ok

line-o (approved):
I am in favour of this change. It might be good to have a look at the code analysis warnings to see if there is a useful improvement to be made.

duncdrum (approved):
much cleaner I like it. not sure how it would affect docker though, there is no shell in our base images.

dizzzz:
I think that for Docker there are no differences, like you Said

reinhapa (commented):
@dizzzz there seems to be some warnings from static code analysis left. Could you look into these as well?

line-o:
@reinhapa at this point as @dizzzz already did go the extra mile and addressed the ones he found to be important to be fixed I would like you to go through the remaining ones and point out which you think should be fixed.

line-o:
Screenshot 2026-02-11 at 09 45 31

exist-distribution/src/main/templates/template.sh
```diff
@@ -0,0 +1,157 @@
...
+  echo "################"
+  echo JAVA_OPTS="$_JAVA_OPTS"
+  echo "################"
+  echo EXIST_OPTS="$_EXIST_OPTS"
+  echo "################"
+  echo EXIST_COMMAND="$EXIST_COMMAND"
+  echo "################"
+fi
+
+# Dot not double quote variables below
```
line-o, line 150: the following two codacy warnings _must_ _be_ _ignored_

line-o:
2 of those warnings _have to be itgnored_ as we cannot put double quotes around them

line-o:
That leaves one warning about the Darwin property never being used. Is that so @dizzzz ?

dizzzz:
Right :) but maybe we should report the detected modus in debug mode?

Interestingly the original scripts have the same unused item. I decided to leave it in :)

line-o:
Then @dizzzz  either remove it or expose it as you choose.  I suspect that we will not stop here to further improve exist-db's startup.

reinhapa (approved):

line-o:
can be merged once tests pass

PR #5980, opened 2025-12-17
[Feature] Revise default webDAV size reporting parameters

dizzzz:
In webDav details matter. Each and every client implementation has different assumptions on what shall and shall not be reported on document sizes.

In the past, for performance optimization, I decided to report mostly estimated XML document sizes. Today I found out that Cyberduck does not accept this anymore.

This PR defaults the parameters to the slowest but 100% accurate settings.

----

IMO the extension needs to be replaces. Funding is needed.


line-o (approved):
I agree

duncdrum (approved):

PR #5982, opened 2025-12-21
Revive build of oracle driver

dizzzz:
Until some time ago, the oracle driver needed to be downloaded manually. Additional license agreements needed to be accepted. This is not the case anymore.

This PR re-enables the build of the oracle-sql module, this makes sure the module is always in sync with eXist-db.

Note that the jars are not included in the eXist-db distribution, as I think the nr of users is limited.


extensions/modules/sql-oracle/src/main/java/org/exist/xquery/modules/oracle/ExecuteFunction.java
```diff
@@ -21,417 +21,364 @@
...
 import org.exist.Namespaces;
 import org.exist.dom.QName;
-import org.apache.commons.io.output.ByteArrayOutputStream;
 import org.exist.dom.memtree.MemTreeBuilder;
-import org.exist.xquery.BasicFunction;
-import org.exist.xquery.Cardinality;
-import org.exist.xquery.FunctionSignature;
-import org.exist.xquery.XPathException;
-import org.exist.xquery.XQueryContext;
+import org.exist.xquery.*;
```
dizzzz: expand * import

reinhapa (approved):

duncdrum (approved):
@dizzzz nice work. Wouldn't it make sense to fully externalization this as a separate package? 

How can ppl use this now:

- Change `conf.xml`
- manually copy jar into place 
- Reboot
- Profit?

(Step 2 won't work on containers)

Sounds like it would be easier to maintain and use if we had a xar package that included the jars.

reinhapa (approved):

dizzzz:
@duncdrum valid points, moving it into a separate XAR file would be nice indeed. Let's have some thoughts about that.

In general I don't like the structure of the 'extensions' directory. The subdirs indexes (indices :-) ), modules, security don't really match with the other dirs.

A top level directory called 'xars' with underneath modules that generate XARs would already be a step.

PR #5993, opened 2026-01-02
[Feature] revise webdav configuration mechanism

dizzzz:
Make configuration code of the webDAV extension more simple and consistent.

As a bonus the jdom 1.x dependency is removed for 'production'. For tests it is is still needed.

Code has been improved mechanically a bit.


joewiz:
A question, asked out of pure curiosity, total ignorance, and everlasting gratitude for your work on WebDAV in eXist: is there a reason that WebDAV configuration is stored in webdav.properties and not in conf.xml?

dizzzz:
@joewiz that is a very valid question :-)  I never thought about it, I think I added this file as a 'easy fix' a long time ago.

That said, I am not sure if extensions should be configured from conf.xml, shouldn't it be the core only?
what if I'd make a XAR of an extension, how to configure it then? #noIdeaYet

adamretter:
I added a mechanism to conf.xml years ago that allows you to specify a config for each module, maybe @dizzzz that could help you to move your config into there within the WebDAV section?

dizzzz:
thnx @adamretter I know about the config section.

joewiz:
@dizzzz Ah! Interesting to consider how a jar-only .xar package could have an editable configuration. Perhaps it could (a) ship with a default configuration that would be instantiated without the presence of a config file, and (b) it could check for the presence of a config file on disk (or in `/db/system`). I suspect that storing the config on disk may have some benefits - namely, surviving database corruption.

extensions/webdav/src/main/java/org/exist/webdav/MiltonDocument.java
```diff
@@ -115,61 +132,35 @@ public MiltonDocument(final Properties configuration, String host, XmldbURI uri,
...
-            getSizeMethod = getSizeMethod(systemProp);
-        }
-
-        if (getSizeMethod == null) {
-            LOG.info("Alternatively try to obtain {} from properties file", GET_METHOD_XML_SIZE);
-            String fileProp = configuration.getProperty(GET_METHOD_XML_SIZE);
-            getSizeMethod = getSizeMethod(fileProp);
+        if(getSizeMethod==null) {
+            final String getMethod = configuration.getProperty(GET_METHOD_XML_SIZE);
+            getSizeMethod = getSizeMethod(getMethod);
```
reinhapa, line 143: Address Codacy issue: _Possible unsafe assignment to a non-final static field in a constructor._

reinhapa (changes requested):

dizzzz:
Not needed anymore. Jack rabbit it is

dizzzz:
superseded by #6364

PR #6005, opened 2026-01-08
[enhancement] deprecate Cardinality _MANY

reinhapa:
### Description:
Deprecate `_MANY` as it is not an available XPath indicator. 

Use `EMPTY_SEQUENCE` and `EXACTLY_ONE`  to achieve the same behavior if needed.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/Cardinality.java
+++ b/exist-core/src/main/java/org/exist/xquery/Cardinality.java
@@ -47,7 +47,10 @@ public enum Cardinality {
     //TODO(AR) can we eliminate this?
     EXACTLY_ONE(ONE),
 
-    //TODO(AR) eliminate this in favour of probably ONE_OR_MORE
+    /**
+     * @deprecated for removal. Use #EMPTY_SEQUENCE and #EXACTLY_ONE to determine many records instead.
+     */
+    @Deprecated
     _MANY(MANY),
 
     /**


adamretter:
@reinhapa I think your code comment could  benefit from being a little less assertive as _MANY has the value 4 whilst ONE_OR_MORE has the value 6. So it's unfortunately in most places not just a case of replacing one with the other, but instead at those call sites some changes may need to be made

line-o (approved):
_MANY will be removed without an (exposed) replacement. In places where _MANY is still used within the codebase it needs to be adapted to make use of ONE_OR_MORE instead.

I think the comment is good enough to transport that information

adamretter:
@line-o actually no, that's not how you would fix the call sites, which is why I commented that Patrick's comment in the code is incorrect/misleading

dizzzz (approved):

reinhapa:
> @reinhapa I think your code comment could benefit from being a little less assertive as _MANY has the value 4 whilst ONE_OR_MORE has the value 6. So it's unfortunately in most places not just a case of replacing one with the other, but instead at those call sites some changes may need to be made

@adamretter  What would be your suggestion instead?

I had a look into the code and to occurrences of those values and for me at least  there is a mix up between the following use cases

- XML schema definition: `ZERO_OR_ONE`, `ONE_OR_MORE`, `ZERO_OR_MORE`)
- Function arguments:  `EXACTLY_ONE`, `ZERO_OR_MORE`
- Functions results: `EMTY_SEQUENCE`. `EXATLY_ONE`, `ZERO_OR_MORE` 

IMHO this can not actually be captured in a single Enum in general.

adamretter:
> I had a look into the code and to occurrences of those values and for me at least there is a mix up between the following use cases
>
> XML schema definition: ZERO_OR_ONE, ONE_OR_MORE, ZERO_OR_MORE)
> Function arguments: EXACTLY_ONE, ZERO_OR_MORE
> Functions results: EMTY_SEQUENCE. EXATLY_ONE, ZERO_OR_MORE

The cardinality Enum should only have the cardinalities applicable to XDM, i.e.:
* `EMTY_SEQUENCE` `()`
* `ZERO_OR_ONE` (`?`)
* `EXACTLY_ONE`
* `ONE_OR_MORE` (`+`)
* `ZERO_OR_MORE` (`*`)

My suggestion is to fix the calling code by refactoring it, the solution to fixing that is not to move those call sites to `ONE_OR_MORE` as that will just break the current behaviour.

reinhapa:
> My suggestion is to fix the calling code by refactoring it, the solution to fixing that is not to move those call sites to `ONE_OR_MORE` as that will just break the current behaviour.

That is actually the plan. But I would suggest to move the discussion to #6007 ...

dizzzz (approved):

line-o (approved):

PR #6014, opened 2026-01-19
[fix] bump container OS to debian 13

duncdrum:
Debian 12 will be dropped in June or so. No reason not to jump to the new hotness.

Distroless will drop deb 12 sooner, see [support policy](https://github.com/GoogleContainerTools/distroless/blob/main/SUPPORT_POLICY.md)

### DIFF
--- a/.github/workflows/ci-container.yml
+++ b/.github/workflows/ci-container.yml
@@ -41,7 +41,7 @@ jobs:
         # Hack around #5450
       - name: pull base image
         run: |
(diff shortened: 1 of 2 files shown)


line-o (approved):

dizzzz (approved):

adamretter:
@duncdrum I think you may also want to update `exist-docker/src/main/resources-filtered/Dockerfile-DEBUG`

duncdrum:
@adamretter `eclipse-temurin:21-jdk-noble` is ubuntu 24 which is already based on Debian 13, no changes necessary.

Issue #2393, opened 2019-01-15
Include open issue test in main repo


duncdrum:
## What is the Problem
We are drowning in open issues going back to 2013, with no reliable means of seeing if and how later changes to the code base affect open issues. Similarly, the significant number of tests locked in issue comments, is not used to help us detect regressions. The question therefore is how to best utilise the 100+ tests that have been written over the years in response to user issues. I can see three technical means of going about this, for Junit and XQsuite tests. 

### Known-issue test folder
Ask users to submit test into a folder which is executed by our Ci in separate step that has the `allowed-to-fail` flag set. In pseudo code:

```bash
ant test backlog
```

Run only tests in the `test/src/backlog` folder, we would call it e.g. on Travis like this:

```yaml
env:
matrix:
  include:
    - jdk: openjdk8
    - jdk: openjdk9
    - jdk: openjdk10
    - jdk: openjdk11
    - jdk: oraclejdk11
  allow_failures:
    - jdk: openjdk8
      env: backlog=true
      script: ant test backlog
```

Once a backlog test passes it should be pulled into the main test-suite and the corresponding issue closed.

#### Pro:
-   clear where to submit code for submitters
-   test don’t need to be written expecting fails, but can contain the expected value from the get go. 
#### Con:
-   NPE causing test could potentially remain in the backlog folder marked as `%pending` indefinitely. 
-   Our Test reporter on Travis makes it very hard to see if and how individual tests failed, so this will be some needle in the haystack work. 
-   Requires active maintenance when moving tests between test-suites

### Enable `%test:warn` aka yellow test for XQsuite tests
Yellow test, in addition to fail=red, pass=green, is similar to `@ignore` in [Junit4](https://www.tutorialspoint.com/junit/junit_ignore_test.htm), or [c](https://www.boost.org/doc/libs/1_67_0/libs/test/doc/html/boost_test/testing_tools/expected_failures.html#l_expected_failure). This are different from `%test:pending`. Pending tests, are not executed at all, whereas warnings are run but don’t halt execution of the test-suite on failure. 

```Xquery
(: more test here :)
declare
  %test:warn
  %test:assertEquals(3)
function ao:wrap-element-sequence () {
  let $xml := <root><i/><i/><i/></root>
  return $xml/node()
    => for-each(ao:wrap#1)
    => for-each(ao:get-i-elements#1)
    => count()
};	
```
The main challenge would be in how to report these. 

```xml
<testsuites>
[...]

PR #6033, opened 2026-02-04
[test] index test updates OP#318 

duncdrum:
while going though old issue and checking their test coverage I found the mix of existing xqsuite test in xml testSet format, and xquery format irritating. This now unifies on XQ format, for easier portability, and better test isolation. I could also unskip a few test which hadn't been running since the spice girls, but which are now passing.  All test from xml have been refactored keeping their syntax and intention as intact as possible. 

What actually go me going are the first 4 commits. These are pending tests for open issue, confirmed to still exist in our code base since long before the spice girls. 

I expect further refactoring of the indexing tests and more additions, so it would be good to get this merged as the basis for that. Focusing on: 

-  older test relying on assert `true` where more expressive assertions, makes it easier to see what went wrong when something does go wrong. 
- More open issues I still have to reproduce and see if we should add pending tests for them
- Some cleanup and shuffling of redundant tests where xml and xq were testing the same thing.

see #2393


adamretter:
In terms of debugging tests and fixing issues in the actual underlying Java code, this is a bad idea. XQSuite is horribly convoluted when trying to understand the real problems that are in the Java code. This will make the developers work much harder. You would have been better going in the opposite direction and converting the XQSuite tests into the older XML format, it's runner implement is much simpler and much easier to debug the Java code when it executes

duncdrum:
execution times for last test run on [linux develop](https://github.com/eXist-db/exist/actions/runs/21604781897/job/62625801354#step:6:9340) and [here](https://github.com/eXist-db/exist/actions/runs/21719305322/job/62644151688#step:6:9340) are `2:14m` and `2:15m` for the lucene tests. So with slightly more tests performance hasn't suffered. 

Turns out some of the existing tests were relying on state from the xml specs or from previous tests, which mavend exaggerated on CI. Sharing state does not strike me as conducive to a good developer experience. 

@adamretter the horrible convolutedness of a mix of xml and xq tests while working on the underlying java code is what motivated this PR. The work won't stop here.

line-o:
I do appreciate the effort you poured into this @duncdrum !
Especially the tests that aim to give more meaningful assertion errors would certainly benefit from the ability to create custom assertions that we will get once my feature PR #4332  is merged.

adamretter:
> adamretter the horrible convolutedness of a mix of xml and xq tests while working on the underlying java code is what motivated this PR. The work won't stop here.

But switching to XQSuite makes this much much worse. Are you doing to instead switch all the tests to the XML format instead then to improve this?

extensions/indexes/lucene/src/test/xquery/lucene/unicode-supplementary-lucene.xql
```diff
@@ -0,0 +1,182 @@
...
+    xmldb:remove("/db/system/config/db/" || $unic-smp-l:COLLECTION_NAME)
+};
+
+(:~
+ : Counts how many codepoints in the given group Lucene finds (ft:query).
+ :
+ : @param $group group name: smp-indexed | smp-dropped | sip-indexed | sip-dropped
+ : @return xs:string "group: count" e.g. "smp-indexed: 5"
+ :)
+declare function unic-smp-l:lucene-finds-count($group as xs:string) as xs:string {
```
line-o, line 128: This here could be rewritten to a custom assertion

duncdrum: yes and it might in a future refactoring of the lucene tests, but this is the closest to the issue report. So I d rather leave this commit as is and focus on a larger refactoring of the entire Lucene suite, including this spec.

dizzzz (approved):
LGTM ; nice job

line-o (commented):
(sorry that is not the case) ✂️

duncdrum:
@line-o so good to merge then?

Issue #4334, opened 2022-04-12
[BUG] FileLock has insufficient synchronisation


adamretter:
```
Caused by: java.lang.IllegalArgumentException
	at java.nio.Buffer.position(Buffer.java:244) ~[?:1.8.0_292]
	at java.nio.HeapByteBuffer.put(HeapByteBuffer.java:196) ~[?:1.8.0_292]
	at java.nio.ByteBuffer.put(ByteBuffer.java:867) ~[?:1.8.0_292]
	at org.exist.storage.lock.FileLock.save(FileLock.java:241) ~[exist-core-5.5.0-SNAPSHOT.jar:5.5.0-SNAPSHOT]
	at org.exist.storage.lock.FileLockHeartBeat.execute(FileLockHeartBeat.java:71) ~[exist-core-5.5.0-SNAPSHOT.jar:5.5.0-SNAPSHOT]
```

and

```
java.nio.BufferOverflowException: null
	at java.nio.Buffer.nextPutIndex(Buffer.java:533) ~[?:1.8.0_292]
	at java.nio.HeapByteBuffer.putLong(HeapByteBuffer.java:438) ~[?:1.8.0_292]
	at org.exist.storage.lock.FileLock.save(FileLock.java:242) ~[exist-core-5.5.0-SNAPSHOT.jar:5.5.0-SNAPSHOT]
	at org.exist.storage.lock.FileLockHeartBeat.execute(FileLockHeartBeat.java:71) ~[exist-core-5.5.0-SNAPSHOT.jar:5.5.0-SNAPSHOT]
```

duncdrum:
Without reproduction steps the log excerpts aren't very helpful. 

I don't doubt that there are issues with lock synchronization. But the issue report does little to help us address them.

duncdrum:
Ping @eXist-db/core

line-o:
I agree that the issue report is not sufficient. I would throw this into a prompt and see what falls out :)

Issue #4140, opened 2021-12-21
util:get-fragment-between() produces not well-formed XML and behaves not as in eXistdb 4.3.1 (or minor)


StefanDumont:
### What is the problem

Apparently, after the fix, `util:get-fragment-between()` no longer works as it did in 4.3.1. If the second parameter `$ending-node` is empty (because it is e.g. the last milestone element), the function produces not well-formed XML (as xs:string of course). In the example below, only  `&lt;body&gt;` is returned. In existdb versions <= 4.3.1 the functions returned the content between the last milestone and the closing tag of the ancestor element(s).

### What did you expect

... and which was still the case in 4.3.1:
```
&lt;body&gt;
&lt;pb&gt;&lt;/pb&gt;&lt;p&gt;Test p 3&lt;/p&gt;&lt;/body&gt;
```

### Describe how to reproduce or add a test

Adapted from @joewiz example above:

```
xquery version "3.1";

let $data := 
    <body>
        <pb/>
        <p>Test p 1</p>
        <pb/>
        <p>Test p 2</p>
        <pb/>
        <p>Test p 3</p>
    </body>
let $store := xmldb:store("/db", "test.xml", $data)
let $doc := doc("/db/test.xml")
let $elems := $doc/body/pb
let $fragment := util:get-fragment-between($elems[3], $elems[4], true(), true())
return 
    $fragment
``` 

### Context information

- eXist-db version: 5.3.0
- Java version: 1.8.0_301
- Operating system: Windows 10
- 64 bit
- JAR installer

StefanDumont:
Tests for this problems, based on @joewiz test in https://github.com/eXist-db/exist/issues/2316#issuecomment-442924414) - see the test results below in different versions. Please note also my two comments in the test code :-) 

```xquery version "3.1";

module namespace gfb = "http://exist-db.org/test/util/get-fragment-between";

declare namespace test="http://exist-db.org/xquery/xqsuite";

declare variable $gfb:DATA := 
    <root xmlns="http://exist-db.org/xquery/xqsuite">
        <pb/>
        <p>Test p 1</p>
        <pb/>
        <p>Test p 2</p>
        <pb/>
        <p>Test p 3</p>
    </root>;
    
declare 
    %test:setUp 
function gfb:setup() {
    xmldb:store("/db", "test.xml", $gfb:DATA)
};

declare 
    %test:tearDown 
function gfb:teardown() {
    xmldb:remove("/db/test.xml")
};

declare 
[...]

Issue #3685, opened 2020-12-30
[BUG] util:get-fragment-between returns unparsable XML when attribute contains square brackets


joewiz:
**Describe the bug**

When calling `util:get-fragment-between()` on a tree that contains an attribute with square brackets, the function returns a string that cannot be parsed.

**Expected behavior**

I expected the function to return parseable XML, as the source is valid XML.

For example, given a document, `test.xml`:

```xquery
<root foo="[bar]">
    <x/>
    <y/>
</root>
```

... then the following query:

```xquery
let $doc := doc("/db/test.xml")/root
let $beginning-node := $doc/x
let $ending-node := $doc/y
let $make-fragment := true()
let $display-root-namespace := true()
return
    util:get-fragment-between($beginning-node, $ending-node, $make-fragment, $display-root-namespace)
```

... will return a string that cannot be parsed as XML:

```xml
<root foo=" bar">
<x></x>
    </root"]>
```

Notice that:

1. The attribute doesn't match the original (`" bar"` instead of `"[bar]"`)
2. The close tag is mangled (`</root"]>` instead of `</root>`)

**To Reproduce**

The following test extends tests already in https://github.com/eXist-db/exist/blob/develop/exist-core/src/test/xquery/util/get-fragment-between.xqm (added in https://github.com/eXist-db/exist/pull/3328).

```xquery
xquery version "3.1";

module namespace gfb = "http://exist-db.org/test/util/get-fragment-between";

declare namespace test = "http://exist-db.org/xquery/xqsuite";

declare variable $gfb:DOC3 :=
    <root foo="[bar]" xmlns="http://exist-db.org/xquery/xqsuite">
        <x/>
        <y/>
        <z/>
    </root>;

declare
    %test:setUp
function gfb:setup() {
    xmldb:create-collection("/db", "test-gfb"),
    xmldb:store("/db/test-gfb", "doc3.xml", $gfb:DOC3)
};

declare
    %test:tearDown
function gfb:teardown() {
    xmldb:remove("/db/test-gfb")
};

declare
    %test:assertTrue
function gfb:wrapped-fragment-with-bracked-in-attribute-is-parseable() {
    let $doc := doc("/db/test-gfb/doc3.xml")
    let $elems := $doc/test:root/*
    let $fragment := util:get-fragment-between($elems[1], $elems[2], true(), false())
    let $parsed := try { parse-xml($fragment) } catch * { $err:code }
    return
        $parsed instance of document-node(element(root))
};
```

[...]

Issue #1705, opened 2018-01-26
Collection Deadlock


adamretter:
Can be demonstrated by running `RESTServiceTest.java` in parallel, see: https://github.com/eXist-db/exist/blob/develop/test/src/org/exist/http/RESTServiceTest.java#L65

adamretter:
@duncdrum Yes, it is also possible to make happen with `EvalTest.java`. However, if we do that, then the test suite never completes because of the deadlock.

duncdrum:
well at least the testsuite would be working ;-)

PR #6045, opened 2026-02-11
Fix IT tests

### LINKED ISSUE #2727: CollectionLocksTest hangs on AppVeyor/Windows/JDK11
See https://github.com/eXist-db/exist/blob/develop/exist-core/pom.xml#L732

### LINKED ISSUE #5903: [BUG] CI and site not running all tests
At least the following test suites are missing from the CI and site runs:

1. ConcurrentResourceTest3
2. ConcurrentResourceTest2
3. DOMTestJUnit
4. CachingFilterInputStreamTest_NonMarkableByteArrayInputStream
5. XQueryProcessingInstruction
6. ParseDtdTestNOK
7. ParseXsdTestNOK
8. DbStoreTest2
9. GetReleaseBrokerDeadlocks
10. RecoverBinaryTest2
11. RecoveryTest2
12. RecoveryTest3
13. MemtreeInXQuery
14. IndexerTest3
15. IndexerTest2
16. SendEmailIT
17. JndiConnectionIT
18. ImplicitConnectionCloseIT
19. ConnectionIT
20. ExecuteIT
21. ConnectionPoolIT

Fixing this adds > 200 tests to the test suite.

duncdrum:
Bring back previously excluded IT tests and execute on CI
close #5903

Additional refactoring to ensure that test in the unit test phase do not depend on artefacts only available in later phases.

Separate Unit Test and Integration Test execution on CI

Test for OS integration are IT tests executed in the OS matrix. Unit tests are OS agnostic and only executed on linux. 

This lowers overall execution time. Win specific exceptions within `RESTServiceTest.testPutPlus` , and `ConcurrencyTest`, `FragmentsTest` are maintained for local runs of `mvn test` on windows to stay unaffected. 
see #1705

`ConcurrencyTest` & `FragmentsTest`  are globally excluded because of deadlocks, remains unchanged.
see #4140
see #3685

previously we excluded `CollectionLocksTest` units test via config from running in parallel on windows. On Ci this is no longer necessary. I've run this tests repeatedly on windows CI and it passes. So I removed the now dead config. 
close #2727
see #4334

before:
Screenshot 2026-02-11 at 16 50 30

after:
see below

Improved test isolation, no more hard-coded calls to `~user/wolf/documents/...` on powerpc era laptops ;-)


duncdrum:
So that last codacy warning is a head scratcher. I vote to ignore, unless someone has a better idea.

line-o:
@duncdrum do you mean

> extensions/modules/persistentlogin/src/test/java/org/exist/xquery/modules/persistentlogin/LoginModuleIT.java#L55
> The class name 'LoginModuleIT' doesn't match '^Test.*$|^[A-Z][a-zA-Z0-9]*Test(s|Case)?$'

This can be ignored in my humble opinion

exist-core/pom.xml
```diff
@@ -1191,6 +1191,7 @@ The BaseX Team. The original license statement is also included below.]]></pream
                     </dependency>
                 </dependencies>
                 <configuration>
+                    <skip>${skipUnitTests}</skip>
```
line-o, line 1194: How does this differ from  `-DskipTests=true`

If this stays in we need to document it

exist-parent/pom.xml
```diff
@@ -133,6 +133,9 @@
         <junit.toolbox.version>2.4</junit.toolbox.version>
         <hamcrest.version>3.0</hamcrest.version>
 
+        <!-- CI: skip unit tests when running integration-only (e.g. -DskipUnitTests=true) -->
+        <skipUnitTests>false</skipUnitTests>
```
line-o, line 137: Ah I see, so this is necessary here.
Would `-DskipTests=true` skip all tests?

exist-core/src/test/java/org/exist/xmldb/concurrent/ConcurrentResourceTest2.java
```diff
@@ -39,11 +43,11 @@ public class ConcurrentResourceTest2 extends ConcurrentTestBase {
     
     private static final String QUERY0 =
         "declare default element namespace 'http://www.loc.gov/mods/v3';" +
-        "collection(\"" + XmldbURI.ROOT_COLLECTION + "\")//mods[titleInfo/title &= 'germany']";
+        "collection(\"" + XmldbURI.ROOT_COLLECTION + "\")//mods[contains(lower-case(string(titleInfo/title)), 'germany')]";
```
line-o: Although the old query is certainly wrong I think that the cast to string is not needed.

line-o: I am wondering what this is about since only subject/topic includes 'germany'

exist-core/src/test/java/org/exist/xmldb/concurrent/FragmentsTest.java
```diff
@@ -24,12 +24,18 @@
...
+import org.xmldb.api.base.Collection;
+import org.xmldb.api.base.XMLDBException;
 
 import java.util.Arrays;
 import java.util.List;
 
 public class FragmentsTest extends ConcurrentTestBase {
-    
+
+    private static final String C2 = "C2";
```
line-o: Wouldn't a more meaningful name like `OTHER_COLLECTION` help us better understand what it's for two months from now?

line-o (commented):
Really good stuff. 💯 
I left a few comments and questions inline.

exist-core/pom.xml
duncdrum: Good point. This was necessary when we had different exclusion configs, but I think it no longer is. I ll check.

exist-parent/pom.xml
duncdrum: s.a.

exist-core/src/test/java/org/exist/xmldb/concurrent/ConcurrentResourceTest2.java
duncdrum: Yup I'll check. Looks like we can simplify more here

exist-core/src/test/java/org/exist/xmldb/concurrent/FragmentsTest.java
duncdrum: The test still results in a deadlock. I just noticed that this collection is required by the test logic but never created within this spec. I would leave it so the original test can be run in the future, before we start optimizing it.

exist-core/pom.xml
duncdrum: nope `skipTests` skips all tests, which is not what we want.

exist-core/src/test/java/org/exist/xmldb/concurrent/FragmentsTest.java
duncdrum: i m still not sure what the intended tree structure is like, all nested all sibling we seem to have a mix (which might have been on purpose)

dizzzz (approved):
LGTM

dizzzz:
image

so 'integration' is the default, what is the role of the ubuntu unit test?

duncdrum:
`mvn test` runs our surefire "unit tests" as before. Files are called `*Test.java` This was the majority of tests. As a rule these should be OS independent and only require JVM stuff. 

`mvn verify` includes failsafe integration tests. Files are called `*IT.java`. These weren't all executed before. Files were renamed to follow the convention. Some long running tests (`@slow` is a junit5 thing for a future PR). And tests that were wrongly treated as unit tests before, are now properly treated as IT. 

Our old setup executed both unit and integration tests on the full OS matrix sequentially on every push. Thus hiding undesirable dependencies between test suites and build phases. The new setup executes only IT per OS, these run parallel (as before) but the unit tests are excluded. They run parallel in their own runner. So overall build time decreases. 

So after the PR is merged. `mvn test` will run fewer tests than before. `mvn verify` will act the same as before but execute more tests. On ci we use a new flag to not repeat the unit test run when running IT.

line-o:
Really good stuff!
We have to keep in mind that there is at least one unit test that _is_ OS dependent. So when we encounter those they must be moved to IT tests.
The one test I know of is in an open PR. So the PR needs to change. But I half expect there to be a few more.

duncdrum:
@line-o all the OS specific stuff I found has been moved and is listed in the OP. If you know of more unit tests that should be IT, please let me know which tests.

line-o:
Well then even better

line-o (approved):
Thanks for bringing us yet another step forward.

PR #6052, opened 2026-02-16
[fix] parse-xml-fragment spec compliant

line-o:
### Description:

With this PR applied the implementation of `parse-xml-fragment` is more spec compliant.
Specifically fragments that contain a xml declaration can now be parsed.
Raises an error when the special case of the standalone pseudo attribute to be set.

**NOTE:** The error reporting was greatly simplified. This might be controversial.

Parsing an invalid fragment like `"<a>no"` will result in a _changed_ error description.

Testcode:

```xquery
try {
    parse-xml-fragment('<a>no')
} catch * {
    $err:description
}
```

**Before:**

```
"String passed to fn:parse-xml is not a well-formed XML document. String passed to fn:parse-xml is not a well-formed XML document.: Document is not valid.
Fatal (1,32) : The element type ""a"" must be terminated by the matching end-tag ""</a>"".
Validated in -1 millisec.
"
```

**After:**

```
"String passed to fn:parse-xml is not a well-formed XML document. parse-xml-fragment failed with: The element type ""a"" must be terminated by the matching end-tag ""</a>""."
```

As one can see, the error report in the current implementation tries to report a location within the fragment that is parsed.
Unfortunately, the location is rather misleading which led me to believe it is better left out entirely.

This change will also affect errors raised by `fn:parse-xml`! 

### Reference:

[XQFO 3.1 fn:parse-xml-fragment Spec](https://www.w3.org/TR/xpath-functions-31/#func-parse-xml-fragment)

supersedes  #5037 
refs #2181 

### Type of tests:

Extended parse-xml-fragment tests (XQSuite)


line-o:
I am puzzled why two lucene tests fail on ubuntu. There seems to be no connection but they are failing consistently. I cannot reproduce that locally.

line-o:
needs to be rebase on top of current develop HEAD

line-o:
Done

duncdrum (approved):

dizzzz (approved):

line-o:
This PR let us pass `parse-xml-fragment-013`, `parse-xml-fragment-014`, `parse-xml-fragment-015`,

It fails `parse-xml-fragment-016` because the encoding pseudo attribute is not enforced.

PR #6055, opened 2026-02-17
Port of #5882: "[bugfix] fn:transform: Conversion and treeIndex problems"

duncdrum:
forward port of fix

see #5882


dizzzz (approved):

exist-core/src/test/xquery/xquery3/transform/fnTransform5882.xqm
```diff
@@ -0,0 +1,126 @@
...
+};
+
+(:~
+ : Select the second child of a MemTree root. Without the fix, treeIndex(b) = [1]
+ : (missing index 0 for root), but doc has only one child; xdmNodeAtIndex returns null.
+ :)
+declare
+    %test:assertEquals('<out>second</out>')
+function testTransform:memtree-second-child() {
+    let $source := element root { <a>first</a>, <b>second</b> }
```
line-o, line 81: Please, do not mix direct element constructors with computed element constructors here.

```xquery
let $source := <root><a>first</a><b>second</b></root>
```

exist-core/src/test/xquery/xquery3/transform/fnTransform5882.xqm
```diff
@@ -0,0 +1,126 @@
...
+    let $selection := $source/b
+    let $result := (fn:transform(map{
```
line-o: The parentheses around the call to transform are not needed. Please remove them

line-o (commented):
Looks good.
Please have a look at my  comments to improve `testTransform:memtree-second-child`.

exist-core/src/test/xquery/xquery3/transform/fnTransform5882.xqm
duncdrum: Please look at the inline documentation. The direct element constructor is important, and part of the test. If this works, pretty will work too.

line-o: OK, so you say it _has to be_ a mixture of computed and direct element constructors in order to trigger the issue.

duncdrum: Pure constructor would work to. But the mix is intentional. If this passes we are in the clear.

line-o: So you say

```xquery
let $source := element root { element a { "first" }, element b { "second" } }
```

would work but 

```xquery
let $source := <root><a>first</a><b>second</b></root>
```

does not?

line-o:
@duncdrum The commits do not adhere to the commit message format.
I propose to squash the last two commits and change the commit message subject to

```
[fix] NPEs in fn:transform with maps, arrays, and treeIndex 
```

line-o:
I will propose a revised test suite later.

PR #6060, opened 2026-02-23
[test] improve tests in fnTransform5882.xqm

line-o:
Change module prefix to match module file name.

Prove that direct element constructors, computed element constructors and mixtures all work when
passed to fn:transform.

Re-use direct-element-constructor for stored document contents.

Simplify test function bodies.

Use xs:QName constructor where applicable instead of QName.

Follow-up on #6055


dizzzz (approved):

exist-core/src/test/xquery/xquery3/transform/fnTransform5882.xqm
```diff
@@ -21,106 +21,127 @@
...
+module namespace fnTransform5882="http://exist-db.org/xquery/test/function_transform";
 import module namespace xmldb="http://exist-db.org/xquery/xmldb";
 
 declare namespace test="http://exist-db.org/xquery/xqsuite";
 declare namespace xsl="http://www.w3.org/1999/XSL/Transform";
 
-(:~
- : Regression test for PR 5882: fn:transform with MemTree nodes.
- : A bare element constructor (element root { ... }) creates a MemTree node whose
- : getParentNode() returns null (see issue #1463). We use initial-match-selection
```
duncdrum, line 33: the only thing I would like to keep is the reference to issue 1463. Did you delete that on purpose or would you be willing to keep it around for future reference.

line-o: Oh yes that's important

line-o: The connection to #1463 is not obvious. Do we really need it here?

duncdrum: I d like to keep it in this file, I don't really care about the exact location

line-o: I don't see what context the  linked issue adds
It seems to be unrelated. If you know how it is related please explain

duncdrum: >  creates a MemTree node whose getParentNode() returns null

it was part of my trying to devise tests that reproduce the behaviour described but not reproduced in 5882

line-o:
I did debug this further last night. I will make further changes to this PR.

line-o:
While debugging the tests I noticed that the extra condition introduced in was never true (in contrast to develop-6.x.x where this special handling of NodeProxy instances is needed). This is why I decided to remove it entirely in 53a9155e37dcf11395f7327085df5e966bf94877

I was also somewhat surprised to learn that this fix - or workaround was already introduced in  48261a8afe1b766a8f1ae12411d8db6739df1ab1

duncdrum (approved):

dizzzz (approved):

PR #6072, opened 2026-03-02
[fix] FLWOR order by: downstream clauses now see tuples in sorted order (Issue #5089)

### LINKED ISSUE #4252: [BUG] "order by" clause stops sorting when one ordering specification returns the empty sequence
**Describe the bug**

In a FLWOR expression, when an `order by` clause contains multiple "ordering specifications" and one of them evaluates to an empty sequence for all members in the tuple, subsequent ordering specifications are disregarded.

**Expected behavior**

Subsequent ordering specifications should be respected.

**To Reproduce**

```Xquery
xquery version "3.1";

    <root>
        <a n="4"/>
        <b n="1"/>
        <c n="2"/>
        <a n="3"/>
    </root>
};

declare
    %test:assertEquals("a3", "a4", "b1", "c2")
function t:ordering-specification-empty-sequence-1st() {
    for $elem in $t:XML/root/*
    order by 
        (),
        $elem/name(), 
        $elem/@n
    return
        $elem/name() || $elem/@n
};

declare
    %test:assertEquals("a3", "a4", "b1", "c2")
function t:ordering-specification-empty-sequence-2nd() {
    for $elem in $t:XML/root/*
    order by 
        $elem/name(), 
        (),
        $elem/@n
    return
        $elem/name() || $elem/@n
};

declare
    %test:assertEquals("a3", "a4", "b1", "c2")
function t:ordering-specification-empty-sequence-last() {
    for $elem in $t:XML/root/*
    order by 
        $elem/name(), 
        $elem/@n,
        ()
    return
        $elem/name() || $elem/@n
};
```

All 3 tests should return the same sequence—sorting first by element name (a b c) and then within identical element names by attribute value (3 before 4): `a3, a4, b1, c2`.

But the 1st test returns the items in document order—i.e., with no ordering applied: `a4, b1, c2, a3`.

And the 2nd test applies the 1st ordering specification—by element name—but fails to apply the 3rd ordering specification—by `@n` value: `a4, a3, b1, c2`.

The final test returns the expected results. 

Here are the raw xqsuite results:

```xml
<testsuite package="http://exist-db.org/xquery/test" timestamp="2022-02-23T12:44:43.534-05:00"
    tests="3" failures="2" errors="0" pending="0" time="PT0.004S">
    <testcase name="ordering-specification-empty-1st" class="t:ordering-specification-empty-1st">
        <failure message="assertEquals failed." type="failure-error-code-1">a3 a4 b1 c2</failure>
        <output>a4 b1 c2 a3</output>
    </testcase>
    <testcase name="ordering-specification-empty-2nd" class="t:ordering-specification-empty-2nd">
        <failure message="assertEquals failed." type="failure-error-code-1">a3 a4 b1 c2</failure>
        <output>a4 a3 b1 c2</output>
    </testcase>
    <testcase name="ordering-specification-empty-3nd" class="t:ordering-specification-empty-3nd"/>
</testsuite>

```

**Screenshots**

n/a

 - OS: macOS 12.2.1

**Additional context**

### LINKED ISSUE #5089: OrderBy clause is not XQuery 3 compliant
Specifically, eXist-db's implementation of the XQuery `order by` clause does not follow the W3C specification. The W3C XQuery 3.1 [Order By Clause](https://www.w3.org/TR/xquery-31/#id-order-by-clause) specification states:

> The purpose of an order by clause is to impose a value-based ordering on the tuples in the tuple stream. The output tuple stream of the order by clause contains the same tuples as its input tuple stream, but the tuples may be in a different order.

This means that any tuples in a tuple stream after an Order By Clause should be in the order defined by the order by clause. In the eXist-db implementation that is not the case. In eXist-db any expressions/clauses within a FLWOR expression that occur after the order by clause, but before the return clause, will not see an ordered tuple stream; instead they see the tuple stream prior to the order by clause. The reason for this is that in eXist-db the operation of executing the `sort` for the Order By clause is always deferred until just before the return clause of the FLWOR expression. In eXist-db this is known as a `postEval` phase, see:
1. [`FLWORClause#postEval(Sequence)`](https://github.com/eXist-db/exist/blob/eXist-6.2.0/exist-core/src/main/java/org/exist/xquery/FLWORClause.java#L95)
2. [`ForExpr#eval(Sequence, Item)`](https://github.com/eXist-db/exist/blob/eXist-6.2.0/exist-core/src/main/java/org/exist/xquery/ForExpr.java#L224)
3. [`OrderByClause#postEval(Sequence)`](https://github.com/eXist-db/exist/blob/eXist-6.2.0/exist-core/src/main/java/org/exist/xquery/OrderByClause.java#L90)

Modifying eXist-db to correctly implement the `Order By Clause` semantics is a large task, and may require a rewrite of all FLWOR expressions. The issue was discovered during development of this PR - https://github.com/eXist-db/exist/pull/4530

On an tangentially related note, we suspect there may be other problems with all other FLWOR expressions in eXist-db that implement `postEval` with regards to compliance with W3C XQuery 3.1.

----
### In Further Detail

Given the XQuery:
```xquery
for $x in ('a', 'b')

  count $index1
  
  let $remainder := $index1 mod 2
  order by $remainder, $index1
  
  count $index2

return
    <x index1="{$index1}" index2="{$index2}">{$x}</x>
```

The result should be computed as:
```xml
<x index1="2" index2="1">b</x>
<x index1="1" index2="2">a</x>
```

However eXist-db produces the *incorrect* result due to its incorrect implementation of the `order by` clause:
```xml
<x index1="2" index2="2">b</x>
<x index1="1" index2="1">a</x>
```

This occurs because eXist-db evaluates the query by effectively executing the following steps:
1. Bind `$x := 'a'`
2. Bind `$index1 := 1`
3. Bind `$remainder := 1`
4. Bind `$index2 := 1`
5. Add `<x index1="1" index2="1">a</x>` to the result sequence.
6. Bind `$x := 'b'`
7. Bind `$index1 := 2`
8. Bind `$remainder := 0`
9. Bind `$index2 := 2`
10. Add `<x index1="2" index2="2">b</x>` to the result sequence.
11. Sort the result sequence based by `$remainder ascending` and then `$index1 ascending`
12. Return the sorted result sequence.

The problem is that eXist-db applies the sorting (i.e. the `order by` clause) after the fact.

I have created a detailed UML Sequence Diagram of the above XQuery execution in eXist-db that shows the exact problem:
eXist-db Count and Order By - UML Sequence Diagram

In case it is helpful for anyone else, the code for the UML Sequence Diagram:
```uml
participant user as "User"
participant xquery as "XQuery"
participant pathexpr as "PathExpr"
participant x as "$x: ForExpr"
participant index1 as "$index1: CountExpr"
participant remainder as "$remainder: LetExpr"
participant orderby as "OrderByClause"
participant index2 as "$index2: CountExpr"
participant element as "ElementConstructor"

activate user

'XQuery Compilation Phase
note over user: XQuery Compilation phase starts here
activate xquery
user -> xquery: compile(...)
xquery -> xquery: analyzeAndOptimizeIfModulesChanged(...)
activate pathexpr
xquery -> pathexpr: analyze(...)
activate x
pathexpr -> x: analyze(...)
activate index1
x -> index1: analyze(...)
activate remainder
index1 -> remainder: analyze(...)
activate orderby
remainder -> orderby: analyze(...)
activate index2
orderby -> index2: analyze(...)
activate element
index2 -> element: analyze(...)
element -> index2
deactivate element
index2 -> orderby
deactivate index2
orderby -> remainder
deactivate orderby
remainder -> index1
deactivate remainder
index1 -> x
deactivate index1
x -> pathexpr
deactivate x
pathexpr -> xquery
deactivate pathexpr
xquery -> user
deactivate xquery

'XQuery Execution Phase
note over user: XQuery Execution phase starts here
activate xquery
user -> xquery: execute(...)
activate pathexpr
xquery -> pathexpr: eval(null)
activate x
pathexpr -> x: eval(null)
x -> x: isOuterFor == true
x -> x: $index1 is FLWORClause == true
activate index1
x -> index1: preEval(('a', 'b'))
index1 -> index1: hasPreviousOrderByDescending() == false
index1 -> index1: calcStartPosition == 0
activate remainder
index1 --> remainder: preEval(('a', 'b'))
activate orderby
remainder --> orderby: preEval(('a', 'b')) 
activate index2
orderby -> index2: preEval(('a', 'b')) 
index2 -> index2: hasPreviousOrderByDescending() == false
index2 -> index2: calcStartPosition == 0
index2 -> orderby
orderby --> remainder
remainder --> index1
index1 -> x

'Process first Item 'a'
x -> x: processItem('a')
x -> index1: eval(null, null)
index1 -> index1: increment/decrement count == 1
index1 -> remainder: eval(null, null)
remainder -> orderby: eval(null, null)
note over orderby, index2, element: Error: No ordering of the Tuple Stream as been performed by `order by` before\n the `count $index2` is evaluated and subsequently used by the `x` Element Constructor
orderby -> index2: eval(null, null)
index2 -> index2: increment/decrement count == 1
activate element
index2 -> element: eval(null, null)
element-> index2: (index1=1 index2=1 x='a')
index2-> orderby: (index1=1 index2=1 x='a')
orderby -> orderby: orderedResult == \n  [\n    (index1=1 index2=1 x='a')\n  ]
note over orderby: orderedResult is not yet sorted!
orderby -> remainder: (index1=1 index2=1 x='a')
remainder -> index1: (index1=1 index2=1 x='a')
index1 -> x: (index1=1 index2=1 x='a')

'Process second Item 'b'
x -> x: processItem('b')
x -> index1: eval(null, null)
index1 -> index1: increment/decrement count == 2
index1 -> remainder: eval(null, null)
note over orderby, index2, element: Error: No ordering of the Tuple Stream as been performed by `order by` before\n the `count $index2` is evaluated and subsequently used by the `x` Element Constructor
remainder -> orderby: eval(null, null)
orderby -> index2: eval(null, null)
index2 -> index2: increment/decrement count == 2
index2 -> element: eval(null, null)
element-> index2: (index1=2 index2=2 x='b')
deactivate element
index2-> orderby: (index1=2 index2=2 x='b')
orderby -> orderby: orderedResult == \n  [\n    (index1=1 index2=1 x='a')\n    (index1=2 index2=2 x='b')\n  ]
note over orderby: orderedResult is not yet sorted!
orderby -> remainder: (index1=2 index2=2 x='b')
remainder -> index1: (index1=2 index2=2 x='b')
index1 -> x: (index1=2 index2=2 x='b')

'Post Eval phase
x -> x: callPostEval == true
note over x: Post Eval Phase starts here...
x --> index1: postEval([\n    (index1=1 index2=1 x='a')\n    (index1=2 index2=2 x='b')\n  ])
index1 --> remainder: postEval([\n    (index1=1 index2=1 x='a')\n    (index1=2 index2=2 x='b')\n  ])
remainder --> orderby: postEval([\n    (index1=1 index2=1 x='a')\n    (index1=2 index2=2 x='b')\n  ])
orderby -> orderby: orderedResult.sort
orderby -> orderby: orderedResult == \n  [\n    (index1=2 index2=2 x='b')\n    (index1=1 index2=1 x='a')\n  ]
note over orderby: orderedResult is now sorted!
note over orderby: Error: The Tuple Stream was sorted after the `count $index2`,\n it should have been sorted before!
orderby -> index2: postEval([\n    (index1=2 index2=2 x='b')\n    (index1=1 index2=1 x='a')\n  ])
index2 -> orderby: [\n    (index1=2 index2=2 x='b')\n    (index1=1 index2=1 x='a')\n  ]
deactivate index2
orderby --> remainder: [\n    (index1=2 index2=2 x='b')\n    (index1=1 index2=1 x='a')\n  ]
deactivate orderby
remainder --> index1: [\n    (index1=2 index2=2 x='b')\n    (index1=1 index2=1 x='a')\n  ]
deactivate remainder
index1 --> x: [\n    (index1=2 index2=2 x='b')\n    (index1=1 index2=1 x='a')\n  ]
deactivate index1
x -> pathexpr: [\n    (index1=2 index2=2 x='b')\n    (index1=1 index2=1 x='a')\n  ]
deactivate x
pathexpr -> xquery: [\n    (index1=2 index2=2 x='b')\n    (index1=1 index2=1 x='a')\n  ]
deactivate pathexpr
note over user: The orderedResult overwrites the result sequence,\n and is returned to the User
xquery -> user: [\n    (index1=2 index2=2 x='b')\n    (index1=1 index2=1 x='a')\n  ]
deactivate xquery

deactivate user
```

joewiz:
Closes eXist-db/exist#5089. Also fixes eXist-db/exist#4252.

### Problem

W3C XQuery 3.1 §3.9.4 requires that after an `order by` clause, every subsequent clause in the same FLWOR expression — `count`, `where`, `let`, and inner `for` — sees tuples **in the sorted order**. eXist-db's previous implementation deferred the sort to `postEval()` but evaluated all downstream clauses eagerly during `eval()`, so they operated on the original, unsorted tuple stream.

This meant, for example:

```xquery
for $x in (5, 3, 1, 4, 2)
order by $x
count $rank          (: rank was assigned in unsorted order — WRONG :)
where $rank le 3
return concat($x, ':', $rank)
```

…produced wrong results. The `count` after `order by` was numbering tuples 1–5 based on the original iteration order, not the sorted order. XQTS `prod-CountClause/count-009` and `prod-OrderByClause/orderBy66` failed for this reason. A related flaw caused incorrect results in 11 `prod-GroupByClause` tests where `group by` was followed by `order by`.

A secondary bug: the `count` clause after a `window` clause counted in reverse (N, N-1, …, 1) because `hasPreviousOrderByDescending()` did not handle the `WINDOW` clause type, causing `TumblingWindowExpr537/538/544` and `SlidingWindowExpr538/544` to fail.

**Issue #4252** is also resolved as a direct consequence: when multiple `order by` sort keys are specified and a leading key evaluates to the empty sequence, subsequent keys were previously ignored. The new tuple-capture-and-replay approach pre-evaluates all sort keys per tuple before sorting, so all key specifications are respected.

### Approach

The fix rewrites `OrderByClause` using a **tuple-capture-and-replay** pattern that mirrors what `GroupByClause` already does for the analogous `group by` problem:

- **`eval()`** captures the full variable state (all in-scope bindings) and pre-evaluated sort keys for each tuple. It returns `EMPTY_SEQUENCE` without invoking any downstream clause.
- **`postEval()`** sorts the collected tuples using the same comparison semantics as before (empty/NaN handling, collator support, descending/ascending modifier, stable sort), then replays each sorted tuple through the downstream clause chain via `returnExpr.eval()`. Downstream clauses now receive tuples in the correct sorted order.

A `Deque<OrderByData>` stack ensures the pattern composes correctly under nested FLWOR expressions.

A small edge case arises when **two `order by` clauses** appear in the same FLWOR (XQTS `orderBy66`): the inner OBC's `eval()` is called from the outer OBC's `postEval()` replay, at which point `rootClause.getStartVariable()` returns a stale `ForExpr` variable that is no longer in the active variable chain. The new package-private helper `XQueryContext.getFirstLocalVariable()` walks backward from `lastVar` to recover the true first active variable, making the variable-capture walk correct in all cases.

With the new replay correctly delivering tuples in sorted order, `CountClause`'s `hasPreviousOrderByDescending()` hack is no longer needed and is removed.

### Commits

| # | Commit | What and why |
|---|--------|--------------|
| 1 | `[fix] WindowExpr` | One-line guard: set `firstVariable` only on the first window, so `rootClause.getStartVariable()` reliably returns the window variable. |
| 2 | `[fix] OrderByClause` | Full tuple-capture-and-replay rewrite; includes `XQueryContext.getFirstLocalVariable()`. |
| 3 | `[refactor] CountClause` | Remove `hasPreviousOrderByDescending()` and the `step` field — now vestigial. |
| 4 | `[test]` | 13 new integration tests in `OrderByCountClauseTest`; `OrderByClauseBenchmark` for performance tracking; removes `@pending` from `count.xqm` XQSuite tests; three new XQSuite tests in `flwor.xql` for issue #4252 (matching the test cases from the bug report). |

### XQTS results

| Test set | Before | After |
|---|---|---|
| `prod-CountClause` | 0 failures | 0 failures ✅ |
| `prod-GroupByClause` | **11 failures** | **0 failures** 🎉 |
| `prod-OrderByClause` | `orderBy66` failing | passes ✅ |
| `prod-WindowClause` | 0 failures | 0 failures ✅ |
| All others | unchanged | unchanged |

> **Note for CI reviewers:** `prod-OrderByClause` has 2 pre-existing failures unrelated to this PR (present on `develop` before these changes). They will show up in CI results but are not regressions introduced here.

### Milestones

> **Note:** Issue #4252 is currently milestoned to `eXist-7.0.1`. If this PR lands in the `eXist-7.0.0` release, that milestone should be updated accordingly.

### Performance

`OrderByClauseBenchmark` can be run with:

```bash
mvn test -pl exist-core -Dtest=OrderByClauseBenchmark \
```

The benchmark is excluded from the normal test suite (the class name does not match Surefire's `*Test` discovery pattern, and an `Assume` guard requires `-Dexist.run.benchmarks=true`).

Results on Intel Core i7-10700K / Java 21, comparing `develop` HEAD (before this PR) against this branch:

| Query | N | Before (avg ms) | After (avg ms) | Δ |
|---|---|---|---|---|
| `for $x order by $x` | 10 | 0.61 | 0.61 | ~0% |
| `for $x order by $x` | 100 | 0.48 | 0.51 | +6% |
| `for $x order by $x` | 1000 | 0.70 | 0.73 | +4% |
| `for $x order by $x count $rank` | 10 | 0.55 | 0.52 | -5% |
| `for $x order by $x count $rank` | 100 | 0.55 | 0.50 | -9% |
| `for $x order by $x count $rank` | 1000 | 0.85 | 0.85 | ~0% |
| `for $x for $y order by …` | 10 | 0.82 | 0.77 | -6% |
| `for $x for $y order by …` | 100 | 1.29 | 1.04 | -20% |
| `for $x for $y order by …` | 1000 | 7.53 | 5.00 | **-34%** |
| `for $x order by $x desc` | 100 | 0.50 | 0.44 | -12% |

**No performance regression.** The simple `for $x order by $x` pattern is within measurement noise (±6%). Multi-variable and descending-sort cases are measurably faster: the new implementation sorts pre-computed `AtomicValue` sort keys rather than full result sequences, which benefits more complex queries.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/CountClause.java
+++ b/exist-core/src/main/java/org/exist/xquery/CountClause.java
@@ -31,7 +31,6 @@ import org.exist.xquery.value.Type;
 import org.exist.xquery.value.ValueSequence;
 
 import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 
 /**
@@ -92,41 +91,9 @@ public class CountClause extends AbstractFLWORClause {
 
     @Override
     public Sequence preEval(final Sequence seq) throws XPathException {
-        // determine whether to count down or up
-        this.step = hasPreviousOrderByDescending() ? -1 : 1;
-
-        // get the count start position
-        if (this.step == 1) {
-            this.count = 0;
-        } else {
-            this.count = seq.getItemCountLong() + 1;
-        }
-
+        this.step = 1;
+        this.count = 0;
         return super.preEval(seq);
-    }
-
-    private boolean hasPreviousOrderByDescending() {
-        FLWORClause prev = getPreviousClause();
-        while (prev != null) {
-            switch (prev.getType()) {
-                case LET, GROUPBY, FOR -> {
-                    return false;
-                }
-                case ORDERBY -> {
-                    return isDescending(((OrderByClause) prev).getOrderSpecs());
-                }
-                default -> prev = prev.getPreviousClause();
-            }
-        }
-        return true;
-    }
-    private boolean isDescending(final List<OrderSpec> orderSpecs) {
-        for (final OrderSpec orderSpec : orderSpecs) {
-            if ((orderSpec.getModifiers() & OrderSpec.DESCENDING_ORDER) == OrderSpec.DESCENDING_ORDER) {
-                return true;
-            }
-        }
-        return false;
     }
 
     @Override
--- a/exist-core/src/main/java/org/exist/xquery/OrderByClause.java
+++ b/exist-core/src/main/java/org/exist/xquery/OrderByClause.java
@@ -23,22 +23,35 @@ package org.exist.xquery;
 
 import org.exist.dom.QName;
 import org.exist.xquery.util.ExpressionDumper;
-import org.exist.xquery.value.Item;
-import org.exist.xquery.value.OrderedValueSequence;
-import org.exist.xquery.value.Sequence;
+import org.exist.xquery.value.*;
 
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Represents an "order by" clause within a FLWOR expression.
+ *
+ * <p>The W3C XQuery 3.1 specification requires that after an order by clause, all subsequent
+ * clauses in the FLWOR expression see tuples in the sorted order. This implementation achieves
+ * that by collecting the full variable state for each tuple during {@link #eval}, sorting the
+ * collected tuples in {@link #postEval}, and then replaying each sorted tuple through the
+ * downstream clause chain — mirroring the pattern used by {@link GroupByClause}.</p>
  */
 public class OrderByClause extends AbstractFLWORClause {
 
     private final List<OrderSpec> orderSpecs;
 
+    /**
+     * The outermost clause in the preceding FLWOR clause chain (e.g. the {@code ForExpr}).
+     * Used in {@link #eval} to locate the first runtime variable of the tuple stream, exactly
+     * as {@link GroupByClause} does via its own {@code rootClause} field.
+     */
+    private FLWORClause rootClause = null;
+
     /*  OrderByClause needs to keep state between calls to eval and postEval. We thus need
         to track state in a stack to avoid overwrites if we're called recursively. */
-    private final Deque<OrderedValueSequence> stack = new ArrayDeque<>();
+    private final Deque<OrderByData> stack = new ArrayDeque<>();
 
     public OrderByClause(final XQueryContext context, final List<OrderSpec> orderSpecs) {
         super(context);
@@ -58,10 +71,20 @@ public class OrderByClause extends AbstractFLWORClause {
     public void analyze(AnalyzeContextInfo contextInfo) throws XPathException {
         contextInfo.setParent(this);
         unordered = (contextInfo.getFlags() & UNORDERED) > 0;
+
+        // Walk the previousClause chain to find the outermost (root) clause, mirroring the
+        // pattern used by GroupByClause.analyze().  rootClause.getStartVariable() gives us the
+        // first runtime variable of the tuple stream during eval().
+        FLWORClause prevClause = getPreviousClause();
+        while (prevClause != null) {
+            rootClause = prevClause;
+            prevClause = prevClause.getPreviousClause();
+        }
+
         final AnalyzeContextInfo newContextInfo = new AnalyzeContextInfo(contextInfo);
         newContextInfo.addFlag(SINGLE_STEP_EXECUTION);
         if (orderSpecs != null) {
-            for (OrderSpec spec: orderSpecs) {
+            for (OrderSpec spec : orderSpecs) {
                 spec.analyze(newContextInfo);
             }
         }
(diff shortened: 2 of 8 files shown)


joewiz:
Sorry for the noise, wrong origin branch.

adamretter:
@joewize An interesting approach! I think the performance benchmarks are as expected, i.e. no change.
However, I would be interested to see what impact this has on memory use. Have you benchmarked the memory use of this before and after on various sizes of data set.

I would expect that this approach of capturing and replaying tuples would have a significant impact on memory use, and  that it would be proportional to the amount of data being sorted. I did consider such an approach in the past when I was working on Windowing, but I rejected it due to my concerns about significantly increased memory use on anything but small data sets.

PR #6073, opened 2026-03-02
[fix] FLWOR order by: downstream clauses now see tuples in sorted order (Issue #5089)

### LINKED ISSUE #4252: [BUG] "order by" clause stops sorting when one ordering specification returns the empty sequence

### LINKED ISSUE #5089: OrderBy clause is not XQuery 3 compliant

joewiz:
Closes eXist-db/exist#5089. Also fixes eXist-db/exist#4252.

### Problem

W3C XQuery 3.1 §3.9.4 requires that after an `order by` clause, every subsequent clause in the same FLWOR expression — `count`, `where`, `let`, and inner `for` — sees tuples **in the sorted order**. eXist-db's previous implementation deferred the sort to `postEval()` but evaluated all downstream clauses eagerly during `eval()`, so they operated on the original, unsorted tuple stream.

This meant, for example:

```xquery
for $x in (5, 3, 1, 4, 2)
order by $x
count $rank          (: rank was assigned in unsorted order — WRONG :)
where $rank le 3
return concat($x, ':', $rank)
```

…produced wrong results. The `count` after `order by` was numbering tuples 1–5 based on the original iteration order, not the sorted order. XQTS `prod-CountClause/count-009` and `prod-OrderByClause/orderBy66` failed for this reason. A related flaw caused incorrect results in 11 `prod-GroupByClause` tests where `group by` was followed by `order by`.

A secondary bug: the `count` clause after a `window` clause counted in reverse (N, N-1, …, 1) because `hasPreviousOrderByDescending()` did not handle the `WINDOW` clause type, causing `TumblingWindowExpr537/538/544` and `SlidingWindowExpr538/544` to fail.

**Issue #4252** is also resolved as a direct consequence: when multiple `order by` sort keys are specified and a leading key evaluates to the empty sequence, subsequent keys were previously ignored. The new tuple-capture-and-replay approach pre-evaluates all sort keys per tuple before sorting, so all key specifications are respected.

### Approach

The fix rewrites `OrderByClause` using a **tuple-capture-and-replay** pattern that mirrors what `GroupByClause` already does for the analogous `group by` problem:

- **`eval()`** captures the full variable state (all in-scope bindings) and pre-evaluated sort keys for each tuple. It returns `EMPTY_SEQUENCE` without invoking any downstream clause.
- **`postEval()`** sorts the collected tuples using the same comparison semantics as before (empty/NaN handling, collator support, descending/ascending modifier, stable sort), then replays each sorted tuple through the downstream clause chain via `returnExpr.eval()`. Downstream clauses now receive tuples in the correct sorted order.

A `Deque<OrderByData>` stack ensures the pattern composes correctly under nested FLWOR expressions.

A small edge case arises when **two `order by` clauses** appear in the same FLWOR (XQTS `orderBy66`): the inner OBC's `eval()` is called from the outer OBC's `postEval()` replay, at which point `rootClause.getStartVariable()` returns a stale `ForExpr` variable that is no longer in the active variable chain. The new package-private helper `XQueryContext.getFirstLocalVariable()` walks backward from `lastVar` to recover the true first active variable, making the variable-capture walk correct in all cases.

With the new replay correctly delivering tuples in sorted order, `CountClause`'s `hasPreviousOrderByDescending()` hack is no longer needed and is removed.

### Commits

| # | Commit | What and why |
|---|--------|--------------|
| 1 | `[fix] WindowExpr` | One-line guard: set `firstVariable` only on the first window, so `rootClause.getStartVariable()` reliably returns the window variable. |
| 2 | `[fix] OrderByClause` | Full tuple-capture-and-replay rewrite; includes `XQueryContext.getFirstLocalVariable()`. |
| 3 | `[refactor] CountClause` | Remove `hasPreviousOrderByDescending()` and the `step` field — now vestigial. |
| 4 | `[test]` | 13 new integration tests in `OrderByCountClauseTest`; `OrderByClauseBenchmark` for performance tracking; removes `@pending` from `count.xqm` XQSuite tests; three new XQSuite tests in `flwor.xql` for issue #4252 (matching the test cases from the bug report). |
| 5 | `[optimize] OrderByClause` | Replace `LinkedHashMap` per tuple with compact parallel arrays (`Sequence[]`, `AtomicValue[]`). Adds `OrderByClauseMemoryBenchmark`. |

### XQTS results

| Test set | Before | After |
|---|---|---|
| `prod-CountClause` | 0 failures | 0 failures ✅ |
| `prod-GroupByClause` | **11 failures** | **0 failures** 🎉 |
| `prod-OrderByClause` | `orderBy66` failing | passes ✅ |
| `prod-WindowClause` | 0 failures | 0 failures ✅ |
| All others | unchanged | unchanged |

> **Note for CI reviewers:** `prod-OrderByClause` has 2 pre-existing failures unrelated to this PR (present on `develop` before these changes). They will show up in CI results but are not regressions introduced here.

### Milestones

> **Note:** Issue #4252 is currently milestoned to `eXist-7.0.1`. If this PR lands in the `eXist-7.0.0` release, that milestone should be updated accordingly.

### Performance

`OrderByClauseBenchmark` can be run with:

```bash
mvn test -pl exist-core -Dtest=OrderByClauseBenchmark \
```

The benchmark is excluded from the normal test suite (the class name does not match Surefire's `*Test` discovery pattern, and an `Assume` guard requires `-Dexist.run.benchmarks=true`).

Results on Intel Core i7-10700K / Java 21, comparing `develop` HEAD (before this PR) against this branch:

| Query | N | Before (avg ms) | After (avg ms) | Δ |
|---|---|---|---|---|
| `for $x order by $x` | 10 | 0.61 | 0.61 | ~0% |
| `for $x order by $x` | 100 | 0.48 | 0.51 | +6% |
| `for $x order by $x` | 1000 | 0.70 | 0.73 | +4% |
| `for $x order by $x count $rank` | 10 | 0.55 | 0.52 | -5% |
| `for $x order by $x count $rank` | 100 | 0.55 | 0.50 | -9% |
| `for $x order by $x count $rank` | 1000 | 0.85 | 0.85 | ~0% |
| `for $x for $y order by …` | 10 | 0.82 | 0.77 | -6% |
| `for $x for $y order by …` | 100 | 1.29 | 1.04 | -20% |
| `for $x for $y order by …` | 1000 | 7.53 | 5.00 | **-34%** |
| `for $x order by $x desc` | 100 | 0.50 | 0.44 | -12% |

**No performance regression.** The simple `for $x order by $x` pattern is within measurement noise (±6%). Multi-variable and descending-sort cases are measurably faster: the new implementation sorts pre-computed `AtomicValue` sort keys rather than full result sequences, which benefits more complex queries.

### Memory

In response to review feedback about memory impact, per-tuple storage was optimized to use compact parallel arrays instead of `LinkedHashMap`. `OrderByClauseMemoryBenchmark` can be run with:

```bash
mvn test -pl exist-core -Dtest=OrderByClauseMemoryBenchmark \
```

**Per-tuple overhead comparison** (V = in-scope variables, K = sort keys):

| Approach | Formula | Per-tuple (V=1, K=1) |
|---|---|---|
| Old (`OrderedValueSequence.Entry`) | ~96 + 8K bytes | ~104 bytes |
| New (`OrderByTuple` + parallel arrays) | ~72 + 8V + 8K bytes | ~88 bytes |

The new approach stores variable bindings (needed for spec-correct behavior) **and** is lighter than the old approach for the common 1–2 variable case:

| Scenario | Old | New | Ratio |
|---|---|---|---|
| 1 variable, 1 key (most common) | 104 B/tuple | 88 B/tuple | **0.8x** ✅ |
| 2 variables, 1 key | 104 B/tuple | 96 B/tuple | **0.9x** ✅ |
| 3 variables, 1 key | 104 B/tuple | 104 B/tuple | **1.0x** |

At N=100,000 tuples (single variable): the new approach uses ~8.4 MB vs the old approach's ~9.9 MB — **1.5 MB less**. All tuple memory is transient (freed after `postEval()`), and variable values are references, not copies.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/OrderByClause.java
+++ b/exist-core/src/main/java/org/exist/xquery/OrderByClause.java
@@ -41,7 +41,10 @@ import java.util.stream.IntStream;
 public class OrderByClause extends AbstractFLWORClause {
 
     // Package-private instrumentation fields for memory benchmarking.
-    // Set in postEval() before sorting; zero-cost when not read.
+    // volatile is required: postEval() writes these in a query-executor worker thread,
+    // while OrderByClauseMemoryBenchmark reads them in the JUnit test thread immediately
+    // after executeQuery() returns. Without volatile there is no happens-before guarantee
+    // for the static field reads even though executeQuery() blocks until completion.
     static volatile int lastTupleCount;
     static volatile int lastVarCount;
 
@@ -55,7 +58,8 @@ public class OrderByClause extends AbstractFLWORClause {
     private FLWORClause rootClause = null;
 
     /*  OrderByClause needs to keep state between calls to eval and postEval. We thus need
-        to track state in a stack to avoid overwrites if we're called recursively. */
+        to track state in a stack to avoid overwrites if we're called recursively.
+        See {@link OrderByData} (inner class below) for what each frame holds. */
     private final Deque<OrderByData> stack = new ArrayDeque<>();
 
     public OrderByClause(final XQueryContext context, final List<OrderSpec> orderSpecs) {
@@ -258,7 +262,11 @@ public class OrderByClause extends AbstractFLWORClause {
             for (final OrderByTuple tuple : data.tuples) {
                 context.proceed(this);
 
-                // Restore variable bindings for this tuple (indexed parallel to variableNames)
+                // Restore variable bindings for this tuple (indexed parallel to variableNames).
+                // tuple.variableValues is always allocated as new Sequence[data.variableNames.length]
+                // (see eval()), so its length equals data.variableTemplates.length — no AOOB risk.
+                assert tuple.variableValues.length == data.variableTemplates.length
+                        : "variableValues length mismatch: " + tuple.variableValues.length + " vs " + data.variableTemplates.length;
                 for (int v = 0; v < data.variableTemplates.length; v++) {
                     data.variableTemplates[v].setValue(tuple.variableValues[v]);
                 }
@@ -288,6 +296,10 @@ public class OrderByClause extends AbstractFLWORClause {
      * </ul>
      */
     private void coerceSortKeys(final OrderByData data) throws XPathException {
+        // data.encounteredPrimitiveTypes is initialised with exactly orderSpecs.size() entries
+        // in OrderByData(numOrderSpecs), so get(t) is always in-bounds — no AOOB risk.
+        assert data.encounteredPrimitiveTypes.size() == orderSpecs.size()
+                : "encounteredPrimitiveTypes size mismatch: " + data.encounteredPrimitiveTypes.size() + " vs " + orderSpecs.size();
         for (int t = 0; t < orderSpecs.size(); t++) {
             final BitSet encounteredTypes = data.encounteredPrimitiveTypes.get(t);
             final int typeCardinality = encounteredTypes.cardinality();
--- a/exist-core/src/main/java/org/exist/xquery/OrderByClause.java
+++ b/exist-core/src/main/java/org/exist/xquery/OrderByClause.java
@@ -23,22 +23,35 @@ package org.exist.xquery;
 
 import org.exist.dom.QName;
 import org.exist.xquery.util.ExpressionDumper;
-import org.exist.xquery.value.Item;
-import org.exist.xquery.value.OrderedValueSequence;
-import org.exist.xquery.value.Sequence;
+import org.exist.xquery.value.*;
 
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Represents an "order by" clause within a FLWOR expression.
+ *
+ * <p>The W3C XQuery 3.1 specification requires that after an order by clause, all subsequent
+ * clauses in the FLWOR expression see tuples in the sorted order. This implementation achieves
+ * that by collecting the full variable state for each tuple during {@link #eval}, sorting the
+ * collected tuples in {@link #postEval}, and then replaying each sorted tuple through the
+ * downstream clause chain — mirroring the pattern used by {@link GroupByClause}.</p>
  */
 public class OrderByClause extends AbstractFLWORClause {
 
     private final List<OrderSpec> orderSpecs;
 
+    /**
+     * The outermost clause in the preceding FLWOR clause chain (e.g. the {@code ForExpr}).
+     * Used in {@link #eval} to locate the first runtime variable of the tuple stream, exactly
+     * as {@link GroupByClause} does via its own {@code rootClause} field.
+     */
+    private FLWORClause rootClause = null;
+
     /*  OrderByClause needs to keep state between calls to eval and postEval. We thus need
         to track state in a stack to avoid overwrites if we're called recursively. */
-    private final Deque<OrderedValueSequence> stack = new ArrayDeque<>();
+    private final Deque<OrderByData> stack = new ArrayDeque<>();
 
     public OrderByClause(final XQueryContext context, final List<OrderSpec> orderSpecs) {
         super(context);
@@ -58,10 +71,20 @@ public class OrderByClause extends AbstractFLWORClause {
     public void analyze(AnalyzeContextInfo contextInfo) throws XPathException {
         contextInfo.setParent(this);
         unordered = (contextInfo.getFlags() & UNORDERED) > 0;
+
+        // Walk the previousClause chain to find the outermost (root) clause, mirroring the
+        // pattern used by GroupByClause.analyze().  rootClause.getStartVariable() gives us the
+        // first runtime variable of the tuple stream during eval().
+        FLWORClause prevClause = getPreviousClause();
+        while (prevClause != null) {
+            rootClause = prevClause;
+            prevClause = prevClause.getPreviousClause();
+        }
+
         final AnalyzeContextInfo newContextInfo = new AnalyzeContextInfo(contextInfo);
         newContextInfo.addFlag(SINGLE_STEP_EXECUTION);
         if (orderSpecs != null) {
-            for (OrderSpec spec: orderSpecs) {
+            for (OrderSpec spec : orderSpecs) {
                 spec.analyze(newContextInfo);
             }
         }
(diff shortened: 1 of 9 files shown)


adamretter:
@joewiz An interesting approach! I think the performance benchmarks are as expected, i.e. no change.
However, I would be interested to see what impact this has on memory use. Have you benchmarked the memory use of this before and after on various sizes of data set.

I would expect that this approach of capturing and replaying tuples would have a significant impact on memory use, and that it would be proportional to the amount of data being sorted. I did consider such an approach in the past when I was working on Windowing, but I rejected it due to my concerns about significantly increased memory use on anything but small data sets.

joewiz:
@adamretter [Claude authored what follows. -Joe] Thanks for flagging the memory concern — it's a fair point, and exactly the kind of thing worth quantifying.

### TL;DR

After optimizing the per-tuple storage to use parallel arrays instead of `LinkedHashMap`, the new approach is actually **lighter than the old `OrderedValueSequence.Entry` approach** for the common case (1–2 in-scope variables), while additionally storing all the variable bindings needed for spec-correct behavior.

### What was done

I created `OrderByClauseMemoryBenchmark` (commit 5, same pattern as the existing `OrderByClauseBenchmark`) and measured per-tuple overhead across 4 query variants × 4 data sizes (N = 100 to 100,000).

The initial implementation used a `LinkedHashMap<QName, Sequence>` per tuple (mirroring `GroupByClause`'s pattern), which had **1.8x–2.7x** the overhead of the old `Entry` class. That prompted a straightforward optimization: replace the `LinkedHashMap` with compact parallel arrays — `Sequence[]` per tuple for variable values, `AtomicValue[]` for sort keys, with the `QName[]` and `LocalVariable[]` stored once per invocation in `OrderByData` and shared across all tuples.

### Results after optimization

**Per-tuple overhead** (V = in-scope variables, K = sort keys):

| Approach | Formula | V=1, K=1 |
|---|---|---|
| Old (`OrderedValueSequence.Entry`) | ~96 + 8K bytes | 104 bytes |
| New (`OrderByTuple` + parallel arrays) | ~72 + 8V + 8K bytes | 88 bytes |

| Scenario | Old | New | Ratio |
|---|---|---|---|
| 1 var, 1 key (most common) | 104 B | 88 B | **0.8x** |
| 2 vars, 1 key | 104 B | 96 B | **0.9x** |
| 3 vars, 1 key | 104 B | 104 B | **1.0x** |

At N=100,000 with a single variable: the new approach uses ~8.4 MB vs the old approach's ~9.9 MB.

### Key mitigating factors

1. **Memory is transient** — all `OrderByTuple` objects are freed after `postEval()` completes
2. **Variable values are references** — `Sequence[]` stores pointers to existing objects, no item data is copied
3. **For queries where the old approach was incorrect** (order-by + count/where/let), there's no valid baseline — the old code didn't capture variable bindings at all, which is why it was wrong
4. The approach mirrors `GroupByClause`'s existing tuple-capture pattern, which has been in production without memory concerns

The benchmark can be reproduced with:
```bash
mvn test -pl exist-core -Dtest=OrderByClauseMemoryBenchmark \
```

All 13 `OrderByCountClauseTest` tests continue to pass.

adamretter:
> for the common case 

I would like to know what you deem to be the "common case" exactly?

With regards to memory concerns...

I am not quite sure we are talking about the same thing! I wasn't asking to reduce the memory of the existing approach. My point was rather, doesn't the approach taken here in this PR require memorisation of all tuple values to be able to sort, e.g. if you want to sort 1 million items, you have to load all 1 million items into memory before you sort them?

joewiz:
[Claude authored what follows. -Joe]

Good questions — let me clarify both points.

### "Common case"

By "common case" I mean a FLWOR expression with 1–2 in-scope variables and 1 sort key, e.g.:

```xquery
for $item in $sequence
order by $item/name
return $item
```

or with a `let`:

```xquery
for $item in $sequence
let $key := $item/name
order by $key
return <entry>{ $item }</entry>
```

This covers the vast majority of real-world `order by` usage. With 3+ in-scope variables the per-tuple overhead breaks even with the old approach, and beyond that it grows by 8 bytes per additional variable.

### Memory and materializing all tuples

You're right that this approach requires all tuples to be collected in memory before sorting — but that was equally true of the previous `OrderedValueSequence` implementation. The old code accumulated every result item into an `Entry[]` array during `eval()`, then sorted in `postEval()`. This is inherent to sorting: you can't produce sorted output without seeing all the input first.

The difference is what's stored per tuple:

- **Old:** `Entry` = the result `Item` + an `AtomicValue[]` of sort keys (~104 bytes for 1 key)
- **New:** `OrderByTuple` = a `Sequence[]` of variable value references + an `AtomicValue[]` of sort keys (~88 bytes for 1 var, 1 key)

The new approach stores variable binding references (pointers to existing `Sequence` objects — no item data is copied) instead of result items, because it needs to replay the full variable state through downstream clauses for correctness. But the total per-tuple footprint is comparable — actually slightly smaller in the 1–2 variable case.

So to directly answer your question: yes, sorting 1 million items requires holding 1 million tuples in memory, but that was already the case before this PR. An external/disk-based sort would address that concern, but that would be a substantial separate enhancement orthogonal to this correctness fix.

duncdrum:
needs a rebase

exist-core/src/main/java/org/exist/xquery/OrderByClause.java
```diff
@@ -23,22 +23,40 @@
...
+ * clauses in the FLWOR expression see tuples in the sorted order. This implementation achieves
+ * that by collecting the full variable state for each tuple during {@link #eval}, sorting the
+ * collected tuples in {@link #postEval}, and then replaying each sorted tuple through the
+ * downstream clause chain — mirroring the pattern used by {@link GroupByClause}.</p>
  */
 public class OrderByClause extends AbstractFLWORClause {
 
+    // Package-private instrumentation fields for memory benchmarking.
+    // Set in postEval() before sorting; zero-cost when not read.
+    static volatile int lastTupleCount;
```
dizzzz, line 48: do we need volatile here?

exist-core/src/main/java/org/exist/xquery/OrderByClause.java
```diff
@@ -23,22 +23,40 @@
...
+     * The outermost clause in the preceding FLWOR clause chain (e.g. the {@code ForExpr}).
+     * Used in {@link #eval} to locate the first runtime variable of the tuple stream, exactly
+     * as {@link GroupByClause} does via its own {@code rootClause} field.
+     */
+    private FLWORClause rootClause = null;
+
     /*  OrderByClause needs to keep state between calls to eval and postEval. We thus need
         to track state in a stack to avoid overwrites if we're called recursively. */
-    private final Deque<OrderedValueSequence> stack = new ArrayDeque<>();
+    private final Deque<OrderByData> stack = new ArrayDeque<>();
```
dizzzz, line 63: OrderByData..... what is the origin?

exist-core/src/test/java/org/exist/xquery/OrderByClauseBenchmark.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    @Test
+    public void benchmark() throws XMLDBException {
+        Assume.assumeTrue(
+                "Benchmark skipped by default. Re-run with -Dexist.run.benchmarks=true",
+                Boolean.getBoolean("exist.run.benchmarks"));
+
+        System.out.println("\n=== OrderByClause benchmark ===");
```
dizzzz, line 90: Is System.out.println() on purpose?

exist-core/src/main/java/org/exist/xquery/OrderByClause.java
```diff
@@ -58,55 +76,312 @@ public ClauseType getType() {
...
+                context.declareVariableBinding(var);
+            }
+
+            // Replay each sorted tuple through the downstream clause chain
+            for (final OrderByTuple tuple : data.tuples) {
+                context.proceed(this);
+
+                // Restore variable bindings for this tuple (indexed parallel to variableNames)
+                for (int v = 0; v < data.variableTemplates.length; v++) {
+                    data.variableTemplates[v].setValue(tuple.variableValues[v]);
```
dizzzz, line 271: can this yield into a AOOB-exception ?

exist-core/src/main/java/org/exist/xquery/OrderByClause.java
```diff
@@ -58,55 +76,312 @@ public ClauseType getType() {
...
+     * <ul>
+     *   <li>xs:string and xs:anyURI mixed → cast all to xs:string</li>
+     *   <li>xs:decimal and xs:float mixed → cast all to xs:float</li>
+     *   <li>xs:decimal, xs:float, and/or xs:double mixed → cast all to xs:double</li>
+     *   <li>Any other mix → XPTY0004 error</li>
+     * </ul>
+     */
+    private void coerceSortKeys(final OrderByData data) throws XPathException {
+        for (int t = 0; t < orderSpecs.size(); t++) {
+            final BitSet encounteredTypes = data.encounteredPrimitiveTypes.get(t);
```
dizzzz, line 304: Out of array exception risk?

dizzzz (approved):

exist-core/src/main/java/org/exist/xquery/OrderByClause.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Yes, `volatile` is needed here. `postEval()` writes these fields in a query-executor worker thread, while `OrderByClauseMemoryBenchmark` reads them in the JUnit test thread immediately after `executeQuery()` returns. Without `volatile`, the JIT has no happens-before guarantee for the static field reads, even though `executeQuery()` blocks until the query completes. Added a comment explaining this in 36456915e1.

exist-core/src/main/java/org/exist/xquery/OrderByClause.java
joewiz: [This response was co-authored with Claude Code. -Joe]

`OrderByData` is a private static inner class defined later in the same file (around line 439 in the current version). It holds all per-invocation state collected during `eval()` and consumed in `postEval()` — the tuple list, sort-key type tracking, variable templates, and the `mark` for scope cleanup. Added a reference to it in the comment on the `stack` field in 36456915e1.

exist-core/src/test/java/org/exist/xquery/OrderByClauseBenchmark.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Yes, intentional — benchmark output goes to stdout by design, mirroring the existing `OrderByClauseBenchmark`. The `System.out` calls are gated behind `-Dexist.run.benchmarks=true` (the `Assume.assumeTrue` guard at the top of the method), so they produce no output in normal test runs.

exist-core/src/main/java/org/exist/xquery/OrderByClause.java
joewiz: [This response was co-authored with Claude Code. -Joe]

No AOOB risk. `tuple.variableValues` is always allocated as `new Sequence[data.variableNames.length]` in `eval()`, and `data.variableNames` and `data.variableTemplates` are always the same length — both are built from the same variable-walk loop on the first `eval()` call. The loop bound here is `data.variableTemplates.length`, so the index is always in-bounds. Added an `assert` to document this invariant explicitly in 36456915e1.

exist-core/src/main/java/org/exist/xquery/OrderByClause.java
joewiz: [This response was co-authored with Claude Code. -Joe]

No AOOB risk. `OrderByData` is always constructed with `new OrderByData(orderSpecs.size())`, which initialises `encounteredPrimitiveTypes` with exactly `orderSpecs.size()` entries. The loop here runs `for (int t = 0; t < orderSpecs.size(); t++)`, so `get(t)` is always in-bounds. Added an `assert` to document this invariant explicitly in 36456915e1.

joewiz:

Rebased against `develop` and addressed all of @dizzzz's inline comments in commit 36456915e1:

- **`volatile` fields**: added comment explaining the cross-thread visibility requirement (written by query worker thread, read by benchmark test thread)
- **`OrderByData` origin**: added a reference in the `stack` field comment pointing to the inner class definition
- **`System.out.println`**: intentional — gated behind `-Dexist.run.benchmarks=true`, no output in normal test runs
- **AOOB at replay loop**: added `assert` documenting that `tuple.variableValues.length == data.variableTemplates.length` is invariant
- **AOOB in `coerceSortKeys`**: added `assert` documenting that `encounteredPrimitiveTypes.size() == orderSpecs.size()` is invariant

joewiz:

Rebased against `develop` and addressed all of @dizzzz's inline comments in commit 36456915e1:

- **`volatile` fields**: added comment explaining the cross-thread visibility requirement (written by query worker thread, read by benchmark test thread)
- **`OrderByData` origin**: added a reference in the `stack` field comment pointing to the inner class definition
- **`System.out.println`**: intentional — gated behind `-Dexist.run.benchmarks=true`, no output in normal test runs
- **AOOB at replay loop**: added `assert` documenting that `tuple.variableValues.length == data.variableTemplates.length` is invariant
- **AOOB in `coerceSortKeys`**: added `assert` documenting that `encounteredPrimitiveTypes.size() == orderSpecs.size()` is invariant

duncdrum (commented):
Lgtm. @joewiz could you take a look at the codacy warnings about the sibgle line declaration and unused variables look actionable. 

We should already have a benchmark profile, from my reading these tests are included when it's active, correct?

We should consider, not necessarily in this PR, to have biweekly benchmark runs on ci. They ll be messy, but it seems better than not executing them at all in the repo.

joewiz:

Addressed the two actionable Codacy warnings in the benchmark classes:

- **`OneDeclarationPerLine`**: split `long sum = 0, min = Long.MAX_VALUE` onto separate lines
- **`UnusedLocalVariable`**: removed unused `expectedV`

The remaining 4 new Codacy issues are unavoidable: `ClassNamingConventions` and `JUnitTestsShouldIncludeAssert` (×2 each) flag the benchmark classes for not following test naming conventions and not containing assertions — both intentional, since these are benchmarks, not tests.

On the benchmark profile: yes, the benchmark tests are gated behind `-Dexist.run.benchmarks=true` (a system property checked in the test classes themselves), so they're excluded from normal test runs. The `micro-benchmarks` Maven profile in `exist-parent/pom.xml` is a separate JMH profile; these JUnit-based benchmarks use their own flag rather than that profile.

The biweekly CI benchmark idea sounds worth pursuing — happy to open a follow-up issue if that would be useful.

duncdrum (approved):

PR #2375, opened 2019-01-09
add testsuite for literals in curly braces

line-o:
### Description:

This adds tests for https://github.com/eXist-db/exist/issues/291

### Type of tests:

xqsuite

PR #6075, opened 2026-03-03
[fix] Fix string literals in attribute value templates

### LINKED ISSUE #291: Problem with XQuery literal construction
The following (legal, according to my and Saxon's understanding) XQuery expressions fails to parse in eXist-db-2.2RC2:

```
<root a="{'a'}"/>
```

The error is: err:XPST0003 unexpected token: ' (while expecting closing tag for element constructor: root)

Other literals do no cause this problem; for instance `<root a="{1}"/>` works without problems.

joewiz:
Closes eXist-db/exist#291.

### Problem

String literals inside enclosed expressions within attribute value templates cause a parse error:

```xquery
<root a="{'a'}"/>
(: err:XPST0003 — unexpected token :)
```

This is a [longstanding issue](https://github.com/eXist-db/exist/issues/291) (opened 2014) affecting any use of quoted strings in attribute enclosed expressions. Numeric literals and variable references work fine (`<root a="{1}"/>`, `<root a="{$x}"/>`), but string literals fail.

### Root cause

eXist-db uses ANTLR 2, which evaluates syntactic predicates by performing lookahead — consuming tokens and caching them. The `attributeEnclosedExpr` rule has a syntactic predicate `( LCURLY RCURLY ) =>` to detect empty enclosed expressions (`{}`). During this lookahead:

1. The lexer produces `LCURLY`, running its action: `inAttributeContent = false`, `inElementContent = false`
2. The lexer tries to produce the next token. With `parseStringLiterals = false`, a quote character (`'`) matches the `APOS` alternative instead of starting a `STRING_LITERAL`
3. `APOS ≠ RCURLY`, so the syntactic predicate fails and ANTLR rewinds the token stream position
4. **But the incorrectly-tokenized `APOS` is already cached** — when the parser takes the non-empty expression alternative and the parser action sets `parseStringLiterals = true`, it's too late. The lexer has already committed to `APOS`.

### Approach

The fix ensures `parseStringLiterals` is set at the **lexer level** during the `LCURLY` token action, before any syntactic predicate can cache incorrectly-tokenized tokens:

1. **`XQuery.g` lexer LCURLY action** (line ~2700): When `inAttributeContent` is true, set `parseStringLiterals = true` immediately. This ensures that during syntactic predicate evaluation, any quote character encountered after `{` is correctly tokenized as a string literal delimiter.

2. **`XQuery.g` parser empty expression handler** (line ~2050): Reset `parseStringLiterals = false` alongside the existing `inAttributeContent = true` restoration. Without this, `parseStringLiterals` would remain `true` after an empty `{}` in an attribute, causing the closing attribute delimiter to be misparsed.

The second fix was discovered during testing: the existing test `<elem attribute="content{}content" />` (from `empty-direct-constructor.xqm`) would fail without it.

### XQTS results

| Test set | Before (develop) | After (this PR) | Notes |
|---|---|---|---|
| `prod-DirElemConstructor` | 11 failures, 1 error | 11 failures, 1 error | No change — all pre-existing |
| `prod-CompAttrConstructor` | 44 failures | 44 failures | No change — all pre-existing |

No regressions introduced.

### Test plan

- [x] 8 new Java tests in `StringLiteralInAttributeConstructorTest` covering: single-quoted strings in double-quoted attributes, double-quoted strings in single-quoted attributes, numeric literals, concat with empty sequence, variable substitution, string constructors, multiple attributes, and string concatenation
- [x] Pre-existing `empty-direct-constructor.xqm` tests continue to pass (including `content{}content` in attributes)
- [x] Full eXist test suite: 6,480 tests, **0 failures**, 129 skipped
- [x] XQTS `prod-DirElemConstructor`, `prod-CompAttrConstructor`: no regressions


line-o (approved):
This looks really good and fixes a long standing shortcoming.

The only thing worth looking at is the implementation of the tests in Java instead of XQuery (XQSuite). Although I am pleased to see that `XQueryCompilationTest` is extended here. The tests are short and to the point because of that.

line-o:
Wouldn't a single test case added to test/src/xquery/literals-in-curlies.xql be enough?

line-o:
```xquery
declare
    %test:assertEquals("<root a='a'/>")
function litcurl:concatStringWithEmptySequence() {
    <root a="{'a'}"/>
};
```

line-o:
The fix is so simple that we might want to backport it to develop-6.x.x

line-o:
https://github.com/eXist-db/exist/pull/2375/changes#diff-cc8f62f3e764d03387006010aa63aaab28805fabd6d4a94c3e96f8dfedd7362fR73-R102

joewiz:

Great call — I wasn't aware of the existing `literals-in-curlies.xql` suite (which you wrote in #2375!). I've replaced the Java test class with changes to that file:

- Activated `literalString()` (removed `%test:pending`, uncommented the expression)
- Added three additional cases: `doubleQuotedStringInSingleQuotedAttribute`, `multipleAttributesWithStringLiterals`, `stringConcatInAttribute`

All 17 tests in the suite pass. Agree on the backport to `develop-6.x.x` — I'll cherry-pick after this merges.

If this approach in de0329f looks good, I can consolidate both commits into a cleaned-up PR.

line-o:
The XQTS test cases that do raise an error but none of the expected ones are something to look at. Either in this PR or in a follow up. Most of them are about an unmatched single curly brace in attribute content (see `K2-DirectConElemNamespace-61` for example).

```xquery
 <e xmlns:p="{"/>
```

should either raise `err:XPST0003`  or `err:XQST0022`

line-o:
The fix for my last comment is out of scope of this PR

duncdrum (approved):

PR #6081, opened 2026-03-03
fix: restore support for declare option exist:timeout "-1" to disable timeout

### LINKED ISSUE #2529: Timeouts can no longer be disabled via option exist:set-timeout
As documented, the default query timeout in eXist could be set to infinite by adding

```xquery
declare option exist:timeout "-1";
```

It looks like support for this was mistakenly dropped in b7cc1140c0397f9c8be68afa8917e2e176c0c2ef and should be restored.

joewiz:
## Summary

- Restores the ability to disable query timeouts via `declare option exist:timeout "-1"` (or `"0"`)
- `setTimeoutFromOption()` was missing the `<= 0` guard that `configureDefaults()` already had, causing negative values to be stored directly as the timeout — since `elapsed > -1` is always true, queries were immediately terminated
- Adds `XQueryWatchDogTest` with 7 unit tests covering positive, zero, negative, invalid, and multi-value timeout options

Fixes #2529

## Test plan

- [x] New `XQueryWatchDogTest` passes (7/7 tests)
- [ ] Verify `declare option exist:timeout "-1";` no longer causes immediate query termination in a running eXist-db instance

### DIFF
--- a/exist-core/src/test/java/org/exist/xquery/XQueryWatchDogTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/XQueryWatchDogTest.java
@@ -46,7 +46,7 @@ public class XQueryWatchDogTest {
 
     private long getTimeout(final XQueryWatchDog watchDog) throws Exception {
         final Field f = XQueryWatchDog.class.getDeclaredField("timeout");
-        f.setAccessible(true);
+        f.trySetAccessible();
         return (long) f.get(watchDog);
     }
 
--- a/exist-core/src/test/java/org/exist/xquery/XQueryWatchDogTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/XQueryWatchDogTest.java
@@ -1,0 +1,104 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery;
+
+import org.exist.Namespaces;
+import org.exist.dom.QName;
+import org.easymock.EasyMock;
+import org.junit.Test;
+
+import java.lang.reflect.Field;
+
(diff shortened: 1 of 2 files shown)


joewiz:
## Ubuntu CI timeout analysis

The `ubuntu-latest unit (Java 21)` job has been timing out after 45 minutes on repeated runs of this PR. After investigation, we suspect this is a pre-existing infrastructure issue unrelated to the PR's changes.

**Why we suspect this PR isn't the cause:**

1. The PR changes are trivially small: a 3-line `if` guard in `XQueryWatchDog.java` and a new unit test file with 7 simple EasyMock-based tests. Nothing that could cause a hang.
2. The previous run of this exact PR (run 22698232712) had ubuntu unit tests **pass in ~16 minutes**.
3. The unit test step runs `mvnd test` across **all modules**, so any pre-existing flaky test anywhere in the project can cause the hang.

**Why only ubuntu is affected:**

The CI matrix excludes Windows and macOS from the `unit` test type — they only run `integration` tests. So only ubuntu would exhibit a unit test timeout.

**Likely culprit:**

Known flaky infrastructure tests like `MoveResourceTest` and `RenameCollectionTest` can hang indefinitely on CI runners. Re-running the failed job should resolve the issue.

We welcome other ideas about what might be causing this timeout.

🤖 Co-authored by [Claude Code](https://claude.com/claude-code)

duncdrum:
We should put a hold on this one, until the open xsuite PRs are in. The hang is a concern, imv.

joewiz:
@duncdrum Sure, makes sense! I've changed this PR to draft to make it clear that we shouldn't merge it in this state. 

The other PR of mine showing ubuntu failures despite repeated runs is https://github.com/eXist-db/exist/pull/6083.

joewiz:
@duncdrum says the hold can be lifted.

line-o (approved):
All looking good.
The first warning from Codacy must be ignored as it is imperative to have access to the timeout field in order to test its value. The only other option would be to expose a getter method in order to read it.

The second warning might be something worth to fix.

joewiz:

@line-o Thanks for the review and guidance on the Codacy warnings!

Actions taken:
1. **`setAccessible()` warning** — ignored as you advised (needed to read the private `timeout` field in tests)
2. **`JUnitTestsShouldIncludeAssert` warning** — fixed by adding `assertEquals(Long.MAX_VALUE, getTimeout(watchDog))` to `proceedDoesNotThrowWhenTimeoutDisabledViaOption`, verifying the timeout value is preserved after `proceed()` completes without throwing

Commit: fa8d9a74d1

exist-core/src/test/java/org/exist/xquery/XQueryWatchDogTest.java
```diff
@@ -0,0 +1,105 @@
...
+    private XQueryWatchDog createWatchDog() {
+        final XQueryContext context = EasyMock.createMockBuilder(XQueryContext.class)
+                .withConstructor()
+                .createMock();
+        return new XQueryWatchDog(context);
+    }
+
+    private long getTimeout(final XQueryWatchDog watchDog) throws Exception {
+        final Field f = XQueryWatchDog.class.getDeclaredField("timeout");
+        f.setAccessible(true);
```
reinhapa: since Java 9 we could use `trySetAccessible()` instead of `setAccessible(true)`

reinhapa (changes requested):

joewiz:

@reinhapa Done — switched to `trySetAccessible()` per your suggestion. Commit 66a9e999ae.

reinhapa (approved):

Issue #3289, opened 2020-03-09
Type error: eXist misreads an XPath node sequence with a predicate as xs:boolean [BUG]


ebeshero:
This issue may be related to https://github.com/eXist-db/exist/issues/2159 (where I first posted about it). I'm seeing a buggy error in which eXist seems to be misreading and XPath node sequence as xs:boolean, with this error: Type error: the sequence cannot be converted into a node set. Item type is xs:boolean. This seems to be a misreading of a certain kind of syntax, which I'll document here. 

### Context:
 I generate the same error when running a query in eXist-dB 4.2.1 and in eXist 5.2.  I don't see the error when running the same query in oXygen version 22 with Saxon-HE XQuery 9.9.1.5. 

### Test File and Query to generate the error
Here is a simple XML test file and a query to generate the error.

1) Here's [a simple XML file](https://github.com/robftg/Rocket-Launches/blob/master/XMLandRNG/sts135.xml) for the query (from my students, with marked data about dates and times of rocket launches): This contains a couple of elements with attribute nodes holding date-time data.

2) Here's an XQuery script to try (change the `doc()` function to wherever you're storing the test file.). 
```
xquery version "3.1";
let $testDoc := doc('/db/testQuery/sts135.xml')
let $launchAndLand := $testDoc//*[@*[name() ! contains(., 'DateTime')]]
return $launchAndLand
```

The error message generated in eXist-db is: 
**exerr:ERROR: Type error: the sequence cannot be converted into a node set. Item type is xs:boolean**

ebeshero:
I've also been testing a little further some combinations in oXygen with Saxon-HE XQuery 9.9.1.5

1) This doesn't work anywhere:
```
//*[@* ! name() ! contains(., 'DateTime')]
```

* oXygen tells me: "Effective boolean value is not defined for a sequence of two or more items starting with a boolean".
* eXist tells me: "Type error: the sequence cannot be converted into a node set. Item type is xs:boolean"


2) This syntax doesn't work in eXist, but does work in oXygen (and this is what I think is a bug): 
```
//*[@*[name() ! contains(., 'DateTime')]]
```

3) This syntax works in both oXygen and eXist:
```
//*[@* ! name()[contains(., 'DateTime')]]
```

also 

```
//*[@*[name()[contains(., 'DateTime')]]]
```

adamretter:
[...]

PR #6083, opened 2026-03-03
Fix fn:not() predicate crashes on empty context and atomic sequences

### LINKED ISSUE #2159: Weird "cannot convert xs:boolean('true') to a node set" message
### What is the problem`
The following XQuery gives an error message that as far as I understand things doesn't make sense:
```xquery
let $col as document-node()* := collection('/db/apps/dashboard')
let $doc as element()* := $col/xyz (: Resolves to an empty sequence :)
return 
  $doc/*[not(self::abc)]
  (: Results in: exerr:ERROR cannot convert xs:boolean('true') to a node set :)
```
### What did you expect
An empty sequence as output

Run the XQuery above

### Context information
- eXist 4.3.1 build 201807241821
- Java version 1.8.0_181
- Operating system W10
- 64 bit
- JAR installed
- Any custom changes in e.g. conf.xml? Data directory moved to outside eXist installation directory

### LINKED ISSUE #2308: not() function failing on boolean sequence
### What is the problem
The `not()` function used in a predicate on a boolean sequence fails... I'm trying to count the number of false or true entries in a sequence of booleans. 

1. Counting the number of false entries:
    1. `count((true(), false(), false())[not(.)])` fails with the message: item of '(true, false, false)...' is not a node, and sequence length > 1
    1. `count((true(), false(), false())[. eq false()])` works
1. Counting the number of true entries:
   1. `count((true(), false(), false())[.])` works

Now 1.i. looks like a bug. In XSLT (Saxon) it simply works.

### What did you expect
That this expression `count((true(), false(), false())[not(.)])` works as expected and as it does in Saxon.

Run the attached XQuery

### Context information

- eXist-db version 4.4.0 201809211803
- Java version 1.8.0_191
- Operating system W10
- 64 bit
- Any custom changes in e.g. conf.xml: Enabled caching module

joewiz:
## Summary

Fixes two long-standing bugs in `fn:not()` that cause unexpected errors when used in predicates:

- **#2159**: `$doc/*[not(self::abc)]` throws "cannot convert xs:boolean('true') to a node set" when `$doc` resolves to empty
- **#2308**: `(true(), false())[not(.)]` throws `err:FORG0006` instead of returning `(false())`

Both stem from optimizations in `FunNot` that assume node-set contexts but encounter empty sets or atomic values. These bugs have been open since 2018.

The `FunNot.getDependencies()` fix for #2308 also fixes **#3289**: `@*[name() ! contains(., 'DateTime')]` throws "the sequence cannot be converted into a node set. Item type is xs:boolean" on persistent (stored) documents. The fix ensures `not(.)` reports `CONTEXT_ITEM` dependency inside predicates, which prevents `Predicate.recomputeExecutionMode()` from pre-evaluating `not(.)` against the full context sequence.

## Approach

Per @wolfgangmm's [concern](https://github.com/eXist-db/exist/issues/2159#issuecomment-423218498): "eXist handles fn:not in a particular way by trying to evaluate it as a set operation. A naive fix would likely result in many expressions becoming slower by an order of magnitude."

These fixes **preserve the set-difference optimization** for all performance-critical patterns (`[not(child)]`, `[not(@attr)]`, `[not(descendant::x)]`, `[not(self::element)]`).

### Fix #2159 — `FunNot.eval()` empty-context guard

When `fn:not()` is used inside a predicate and the context sequence is empty, the set-difference path returned a `BooleanValue` which `Predicate.selectByNodeSet()` cannot consume. Fix: return `EMPTY_SEQUENCE` when in a predicate with empty context (filtering an empty set always yields empty). Outside predicates, the boolean path is preserved. This matches the intent of the original commented-out TODO code.

### Fix #2308 — `FunNot.getDependencies()` targeted `CONTEXT_ITEM`

`LocationStep.getDependencies()` suppresses `CONTEXT_ITEM` for the self axis inside predicates to enable the set-difference optimization. This is correct for node-set predicates but causes `not(.)` on atomic sequences to be pre-evaluated against the full sequence instead of item-by-item, throwing `FORG0006`.

Fix: `FunNot.getDependencies()` adds `CONTEXT_ITEM` to the dependency flags when the argument is the context item expression `.` — specifically, a `LocationStep` with `SELF_AXIS` and a `node()` type test — and the function is inside a predicate. This is targeted and narrow: it only affects `.` (`self::node()`), not typed self-axis steps like `self::element`, forcing `Predicate` to use per-item boolean evaluation for that specific case. All other predicates (`not(child::bar)`, `not(@attr)`, `not(descendant::x)`, `not(self::element)`) are unaffected because their arguments use different axes, different type tests, or are not `LocationStep` instances.

`FunNot.returnsType()` and `LocationStep.getDependencies()` are unchanged from develop.

### Why the set-difference optimization is preserved

The optimization in `FunNot.eval()` runs when the argument returns NODE, context is a persistent set, and argument has no `CONTEXT_ITEM` dependency. After these changes:

| Pattern | Arg depends on CONTEXT_ITEM? | Set-difference? | Notes |
|---|---|---|---|
| `[not(child::bar)]` | No | Yes | Unchanged |
| `[not(@attr)]` | No | Yes | Unchanged |
| `[not(descendant::x)]` | No | Yes | Unchanged |
| `[not(self::abc)]` | No (element type test, not node()) | Yes | Unchanged |
| `[not(.)]` on atomics | Yes (self::node(), inPredicate) | No — boolean | Correct, was erroring before |
| `[not(.)]` on nodes | Yes (self::node(), inPredicate) | No — boolean | Trivial pattern, not perf-sensitive |

The promotion path via `Predicate.recomputeExecutionMode()` is unchanged.

## Performance benchmark

`FunNotBenchmark.java` measures 5 query patterns (100 warmup + 500 measured iterations each) against a 200-item generated XML document with no indexes (structural index only).

Run with:
```
mvn test -pl exist-core -Dtest=FunNotBenchmark \
```

### Results

| Query | develop avg (ms) | PR avg (ms) | develop ops/s | PR ops/s |
|---|---|---|---|---|
| `//item[not(child)]` | 0.277 | 0.270 | 3,609 | 3,698 |
| `//item[not(@attr)]` | 0.277 | 0.263 | 3,608 | 3,804 |
| `//item[not(descendant::x)]` | 0.266 | 0.262 | 3,756 | 3,821 |
| `//item[not(.)]` | 0.261 | 0.247 | 3,838 | 4,044 |
| `//item[not(@id > 100)]` | 0.724 | 0.720 | 1,382 | 1,388 |

**Takeaway**: All queries are within noise margin (~2%). The set-difference optimization is fully preserved for child, attribute, and descendant axis patterns. The `not(.)` pattern (boolean fallback path) also shows no regression. Environment: macOS, Zulu JDK 21, single-threaded, in-process embedded server.

## Test plan

- [x] New `fn-not.xq` XQSuite tests covering empty paths, persistent node set-difference, `not(.)` on integers/strings/booleans/nodes
- [x] Removed `%test:pending` from `boolseq:countNegativesContextItem` in `boolean-sequences.xq`
- [x] New `simple-map-predicate.xq` tests for #3289: `@*[name() ! contains(., 'DateTime')]` on persistent and in-memory documents
- [x] Performance benchmark (`FunNotBenchmark.java`): 5 queries × 500 iterations, no regression vs develop
- [x] Full `CoreTests` suite: 992 tests, 0 failures, 0 errors
- [x] `OptimizerTest`: 6 tests, 0 failures
- [x] `DateTest`: 46 tests, 0 failures

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunNot.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunNot.java
@@ -106,13 +106,16 @@ public class FunNot extends Function {
             (contextSequence == null || contextSequence.isPersistentSet()) &&
             !Dependency.dependsOn(arg, Dependency.CONTEXT_ITEM)) {
 			if (contextSequence == null || contextSequence.isEmpty()) {
-				// TODO: special treatment if the context sequence is empty:
-				// within a predicate, we just return the empty sequence
-				// otherwise evaluate the argument and return a boolean result			    
-//				if (inPredicate && !inWhereClause)
-//                    result = Sequence.EMPTY_SEQUENCE;
-//				else
-                    result = evalBoolean(contextSequence, contextItem, arg);
+				if (inPredicate) {
+					// When used inside a predicate in NODE mode, the result is consumed
+					// as a node set by Predicate.selectByNodeSet(). An empty context
+					// means there are no nodes to filter — the set-difference result
+					// is always empty. Returning a boolean here would crash with
+					// "cannot convert xs:boolean to a node set" (GitHub #2159).
+					result = Sequence.EMPTY_SEQUENCE;
+				} else {
+					result = evalBoolean(contextSequence, contextItem, arg);
+				}
 			} else {
     			result = contextSequence.toNodeSet().copy();
 
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunNot.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunNot.java
@@ -26,8 +26,10 @@ import org.exist.dom.persistent.NodeSet;
 import org.exist.dom.QName;
 import org.exist.xquery.AnalyzeContextInfo;
 import org.exist.xquery.Cardinality;
+import org.exist.xquery.Constants;
 import org.exist.xquery.Dependency;
 import org.exist.xquery.Expression;
+import org.exist.xquery.LocationStep;
 import org.exist.xquery.Function;
 import org.exist.xquery.FunctionSignature;
 import org.exist.xquery.Profiler;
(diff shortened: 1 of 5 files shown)


joewiz:

Closing — superseded by #6207 (v2/xq31-compliance-fixes).

PR #6085, opened 2026-03-03
Normalize annotation values in XQSuite assertEquals for XML comparison

### LINKED ISSUE #4327: [BUG] XQsuite `%test:assertEquals` not correctly parsed as XML
**Describe the bug**
When writing an XQsuite test designed to test some XML element content, the content of the `%test:assertEquals` string is not parsed as XML, as promised by the documentation.

**Expected behavior**
From the XQsuite assertEquals documentation:
> If the sequence returned by the function contains one or more XML elements, they will be normalized (ignorable whitespace is stripped). The assertion argument is then parsed into XML and the two node trees are compared using the deep-equals function.

I would expect the xquery below to return a passed test; it currently fails.

**To Reproduce**

```Xquery
xquery version "3.1";
  

declare option output:indent "yes";

declare 
%test:assertEquals('
  <span type="xml">Success!</span>
')
function local:test-nodes() as element(span) {
  <span type="xml">Success!</span>
};

test:suite(local:test-nodes#0)
```

 - OS: Ubuntu 20.04.4 LTS (GNU/Linux 5.4.0-1058-kvm x86_64)

**Additional context**
- How is eXist-db installed? Docker
- Any custom changes in e.g. `conf.xml`? Nope

joewiz:
## Summary

- Fixes `%test:assertEquals` failing when the annotation value contains whitespace around XML (e.g., newlines before/after `<span>`)
- The actual result was already normalized (whitespace stripped via `test:normalize`) but the expected value from the annotation was not, causing `deep-equal` to fail
- Now both values are normalized before comparison, consistent with the [XQSuite docs](https://exist-db.org/exist/apps/doc/xqsuite): "The assertion argument is then parsed into XML and the two node trees are compared using the deep-equals function"

Closes https://github.com/eXist-db/exist/issues/4327

## Test plan

- [x] Added XQSuite test with whitespace in `%test:assertEquals` annotation around XML element
- [x] All 91 existing XQSuite tests continue to pass

### DIFF
--- a/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
+++ b/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
@@ -935,7 +935,12 @@ declare %private function test:equals($annotation-value as element(value), $resu
             default return
                 $result
     let $value := test:xdm-value-from-annotation-value($annotation-value)
-    let $normValue := test:cast-to-type($value, $result)
+    let $normValue :=
+        typeswitch ($result)
+            case node() return
+                test:cast-to-type($value, $result) => test:normalize()
+            default return
+                test:cast-to-type($value, $result)
     return
         typeswitch ($normResult)
             case node() return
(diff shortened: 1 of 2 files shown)


joewiz:

Closing — superseded by #6207 (v2/xq31-compliance-fixes).

PR #6086, opened 2026-03-03
[bugfix] xml-to-json() validates namespace requirement per spec

### LINKED ISSUE #5543: [BUG] xml-to-json() ignores namespace requirement
**Describe the bug**

The description of the `xml-to-json()` function in the [XPath and XQuery Functions and Operators 3.1 spec](https://www.w3.org/TR/xpath-functions-31/#func-xml-to-json) states that:

> The first argument `$input` is a node; the subtree rooted at this node will typically be the XML representation of a JSON document as defined in [17.4.2 XML Representation of JSON](https://www.w3.org/TR/xpath-functions-31/#json-to-xml-mapping).

That definition says that:

> the phrase "an element named N" is to be interpreted as meaning "an element node whose local name is N and whose namespace URI is http://www.w3.org/2005/xpath-functions".

The `xml-to-json()` function in eXist-db transforms XML map and array markup to JSON even when the namespace is omitted. &lt;oXygen/> raises an error if the XML elements to be transformed are not in the required namespace.

**Expected behavior**

I expect eXist-db to raise an error if the XML elements to be transformed by the `xml-to-json()` function are not in the required namespace.

**To Reproduce**

Create the following XQuery and run in eXide and in the &lt;oXygen/> XQuery debugger. It will succeed in both. Then remove the namespace declaration and rerun. The code will succeed (incorrectly) in eXide and raise an error (correctly) in &lt;oXygen/>:

```xquery
let $doc :=
<array
  key="stooges"
  xmlns="http://www.w3.org/2005/xpath-functions">
  <string>Curly</string>
  <string>Larry</string>
  <string>Moe</string>
</array>
return
  fn:xml-to-json($doc)
```

Build: eXist-6.2.0 (c8fa4958b6d4a50bd0cba7f3e76a150226414187)
Java: 1.8.0_333 (Oracle Corporation)
OS: Mac OS X 12.7.6 (x86_64)

**Additional context**
- How is eXist-db installed? [e.g. JAR installer, DMG, … ]: DMG
- Any custom changes in e.g. `conf.xml`?: No

joewiz:
## Summary
- **Fixes** xml-to-json() to validate that input elements are in the `http://www.w3.org/2005/xpath-functions` namespace, as required by the [XPath and XQuery Functions and Operators 3.1 spec](https://www.w3.org/TR/xpath-functions-31/#func-xml-to-json)
- Raises `FOJS0006` ("Invalid XML representation of JSON") when elements are not in the required namespace
- Updates existing tests to use the correct namespace (via `declare default element namespace`) and adds two new tests for the wrong-namespace error case

Closes #5543

## Test plan
- [x] All 942 existing XQuery3 tests pass (0 failures, 0 errors)
- [x] New tests verify FOJS0006 is raised for elements with no namespace and elements with a wrong namespace
- [x] Existing tests with correct namespace (`generatedFromSchema-1`, `generatedFromSchema-2`, etc.) continue to pass
- [x] XQTS `fn-xml-to-json` test set (144 tests from [w3c/qt3tests](https://github.com/w3c/qt3tests) HEAD): **no regressions, no improvements** — 72 pass, 68 fail, 4 skipped, identical before and after this change. The XQTS tests already use the correct namespace, so the new validation doesn't affect them. The 4 skipped tests require `schemaImport` (unrelated). The 68 pre-existing failures break down as: 46 from `fn:doc()`-loaded content (streaming/element recognition issue), 11 missing FOJS0006 validation (duplicate keys, invalid escaped values), 6 escaping differences, and 5 other.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunXmlToJson.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunXmlToJson.java
@@ -24,6 +24,7 @@ package org.exist.xquery.functions.fn;
 import com.fasterxml.jackson.core.*;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.exist.Namespaces;
 import org.exist.dom.QName;
 import org.exist.xquery.*;
 import org.exist.xquery.functions.map.MapType;
@@ -120,6 +121,12 @@ public class FunXmlToJson extends BasicFunction {
                 switch (status) {
                     case XMLStreamReader.START_ELEMENT:
                         tempStringBuilder.setLength(0);
+                        final String elementNamespaceURI = reader.getNamespaceURI();
+                        if (!Namespaces.XPATH_FUNCTIONS_NS.equals(elementNamespaceURI)) {
+                            throw new XPathException(this, ErrorCodes.FOJS0006,
+                                    "Invalid XML representation of JSON. Element '" + reader.getLocalName()
+                                    + "' is not in the required namespace '" + Namespaces.XPATH_FUNCTIONS_NS + "'.");
+                        }
                         final String elementAttributeEscapedValue = reader.getAttributeValue(null, "escaped");
                         elementValueIsEscaped = "true".equals(elementAttributeEscapedValue);
                         final String elementAttributeEscapedKeyValue = reader.getAttributeValue(null, "escaped-key");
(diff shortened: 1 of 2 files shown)


joewiz:
cc/ @djbpitt

duncdrum (approved):
Namespaces are fun, right?

line-o (approved):

PR #6088, opened 2026-03-03
Fix #2205: context problem with map:get() in xpath predicate

### LINKED ISSUE #2205: Context problem with map:get() in xpath
eXist-db 4.4.0 (build Sept 21)

Code:
```
let $myXml      := doc(xmldb:store('/db/apps', 'ttt.xml', <x><y k1="a" k2="1"/><y k1="a" k2="2"/></x>))/x
let $myMap      := map:new((map:entry('a', '1'), map:entry('b', '2')))

return
    $myXml//y[@k1][@k2 = map:get($myMap, @k1)]
```

Yields: 

_err:XPTY0004 checking function parameter 2 in call map:get($myMap, untyped-value-check[xs:anyAtomicType, attribute::{}k1]): XPTY0004: The actual cardinality for parameter 2 does not match the cardinality declared in the function's signature: map:get($map as map(*), $key as xs:anyAtomicType) item()*. Expected cardinality: exactly one, got 2. [at line 6, column 42)_

Expected: <y k1="a" k2="1"/>

There are two ways to work around this:

Change order of @k2 and map:get() like so:
`    $myXml//y[@k1][map:get($myMap, @k1) = @k2]`

or work from an in memory xml fragment.

So there seems to be a context problem with database nodes and the map function

joewiz:
## Summary

- Fix `GeneralComparison.nodeSetCompare()` else branch to evaluate the right operand per-node using the context chain (`item.getContext().getNode()`) instead of the entire `contextSequence`. This resolves `XPTY0004` errors when predicates like `[@k2 = map:get($map, @k1)]` are evaluated against stored (on-disk) nodes.
- Add regression tests for both `=` (general comparison) and `eq` (value comparison) variants of the bug.
- Add a `GeneralComparisonBenchmark` for measuring predicate evaluation performance (guarded by `-Dexist.run.benchmarks=true`).

### Root cause

In `nodeSetCompare()`, the else branch (used when the context documents contain the node documents) evaluated the right operand with `contextSequence` (the whole node set) and `null` contextItem. For `map:get($map, @k1)`, this caused `@k1` to resolve against all nodes at once instead of per-node, producing multiple values where a single one was expected.

The fix mirrors the if-branch's approach: use `context.getNode().toSequence()` to evaluate the right operand against the correct parent element for each node.

## Performance benchmark

`GeneralComparisonBenchmark` is excluded from the normal test suite (the class name does not match Surefire's `*Test` discovery pattern, and an `Assume` guard requires `-Dexist.run.benchmarks=true`). To run it:

```bash
mvn test -pl exist-core -Dtest=GeneralComparisonBenchmark \
```

Benchmarked with an embedded eXist instance, 3 warmup + 5 measured iterations per query. Two query types exercise the **unchanged** code path (no context chain needed); two exercise the **changed** code path (context-dependent right operand).

| Query | Size | develop (before) | fix branch (after) | Δ |
|---|---|---|---|---|
| `@v = 'v0'` (literal) | 100 | 2.37 ms | 2.43 ms | +3% |
| | 500 | 4.22 ms | 4.61 ms | +9% |
| | 2000 | 8.07 ms | 9.99 ms | +24% |
| `@v = $x` (variable) | 100 | 1.14 ms | 0.90 ms | −21% |
| | 500 | 2.84 ms | 1.78 ms | −37% |
| | 2000 | 6.87 ms | 4.44 ms | −35% |
| `@v = map:get($map, @k)` | 100 | ❌ XPTY0004 | 2.50 ms | ✅ fixed |
| | 500 | ❌ XPTY0004 | 5.35 ms | ✅ fixed |
| | 2000 | ❌ XPTY0004 | 13.60 ms | ✅ fixed |
| `@v = local:get(@k)` | 100 | ❌ XPTY0004 | 2.77 ms | ✅ fixed |
| | 500 | ❌ XPTY0004 | 6.54 ms | ✅ fixed |
| | 2000 | ❌ XPTY0004 | 15.59 ms | ✅ fixed |

**Key takeaways:**
- **No performance regression** on the unchanged code paths. The literal-comparison results are within normal JVM microbenchmark noise; the variable-comparison path actually improved (likely JIT warmup variance).
- The context-dependent queries (`map:get`, `local:get`) **crashed on develop** with `XPTY0004` and now **work correctly** at reasonable latency on the fix branch.
- The fix adds a single `if (ctxItem != null)` branch check per node — negligible overhead.

## Test plan

- [x] `xquery.optimizer.OptimizerTests` — 60/60 passed (includes 2 new regression tests)
- [x] `xquery.dates.DateTests` — 121/121 passed (no regressions)
- [x] `xquery.maps.MapTests` — 124/124 passed (no regressions)
- [x] Full `exist-core` test suite — 6496 tests, 0 failures, 0 errors
- [x] `GeneralComparisonBenchmark` — all 4 query variants pass at 3 data sizes

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
+++ b/exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
@@ -638,13 +638,30 @@ public class GeneralComparison extends BinaryOp implements Optimizable, IndexUse
 
             for( final NodeProxy item : nodes ) {
                 final AtomicValue lv = item.atomize();
-                final Sequence    rs = getRight().eval(contextSequence, null);
+                ContextItem ctxItem = item.getContext();
+                if( ctxItem != null ) {
+                    // use the context node (e.g. the parent element) so that
+                    // the right operand can resolve sibling attributes correctly
+                    do {
+                        final Sequence rs = getRight().eval(ctxItem.getNode().toSequence(), null);
 
-                for( final SequenceIterator i2 = Atomize.atomize(rs).iterate(); i2.hasNext(); ) {
-                    final AtomicValue rv = i2.nextItem().atomize();
+                        for( final SequenceIterator i2 = Atomize.atomize(rs).iterate(); i2.hasNext(); ) {
+                            final AtomicValue rv = i2.nextItem().atomize();
 
-                    if( compareAtomic( collator, lv, rv ) ) {
-                        result.add( item );
+                            if( compareAtomic( collator, lv, rv ) ) {
+                                result.add( item );
+                            }
+                        }
+                    } while( ( ctxItem = ctxItem.getNextDirect() ) != null );
+                } else {
+                    final Sequence rs = getRight().eval(contextSequence, null);
+
+                    for( final SequenceIterator i2 = Atomize.atomize(rs).iterate(); i2.hasNext(); ) {
+                        final AtomicValue rv = i2.nextItem().atomize();
+
+                        if( compareAtomic( collator, lv, rv ) ) {
+                            result.add( item );
+                        }
                     }
                 }
             }
--- a/exist-core/src/test/java/org/exist/xquery/GeneralComparisonBenchmark.java
+++ b/exist-core/src/test/java/org/exist/xquery/GeneralComparisonBenchmark.java
@@ -1,0 +1,175 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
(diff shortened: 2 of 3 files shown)


joewiz:
cc/ @djbpitt

duncdrum (approved):

line-o (approved):
Looks good to me!
We should make sure to have an eye on code formatting when touching a file.
Since this should happen in a separate commit, we do not need to wait with merging this PR).

PR #6089, opened 2026-03-03
[bugfix] Fix fn:lang corrupting XQueryContext during compilation

### LINKED ISSUE #5103: [BUG] Variable declaration in library module is not processed reliably

**Describe the bug**

The bug is reproducible in https://github.com/djbpitt/exist-import-test. Clone the repo, build, install, and load the index page in a browser. When you click on a link, the browser reports:

> err:XPDY0002 undefined value for variable '$re:genres' [at line 16, column 37, source: String/-988904492058005134] In function: re:bgMsName(element(tei:TEI)) [31:35:String/-988904492058005134]

If you reload the page it responds correctly, and continues to do so. That is, the error is raised only the first time you try to use the resource. If you stop and restart eXist-db, the error will again appear on first use of the resource.

**Expected behavior**

I expect the link to load the target result correctly on first access, and not only on subsequent access, as it does now.

One option is to use [xst](https://www.npmjs.com/package/@existdb/xst), and copy and paste the output produced by running `xst info` here:**

- Build: eXist-6.2.0 (c8fa4958b6d4a50bd0cba7f3e76a150226414187)
- Java: 21.0.1 (Eclipse Adoptium)
- OS: Mac OS X 14.0 (aarch64)

I see the same behavior on my Centos 7 server. @line-o was able to reproduce the behavior with the current (as of 2023-10-28) development version of eXist-db.

**Additional context**

On MacOS with DMG. On Centos with JAR installer.

No

**Additional information**

- Originally discussed in Slack at https://exist-db.slack.com/archives/CG2MRUZ35/p1698415596487759 beginning on 2023-10-27.
- If I replace the variable declaration for `$re:genres` with a `let` statement inside the function that uses that variable, there is no error. This is a workaround, rather than a solution, because 1) the variable should be available when declared in the prologue, and not only when declared in a `let` statement, and 2) in Real Life I use the same variable in more than one function, so declaring it in the prologue is DRYer than declaring it separately in each function that uses it.
- @line-o reports that: "For me adding `util:log("info", $re:genres),` to the function body weirdly does fix the issue on first request. A heisen-bug, the moment you debug it it is gone."

joewiz:
## Summary

- Fix `FunLang.analyze()` to compile its internal sub-query in a **copied context** (`context.copyContext()`), preventing `XQuery.compile()` from corrupting the outer `XQueryContext`
- Add regression test: library module with a module-level variable and a function using `fn:lang`, imported from a main query — previously failed with `err:XPDY0002` on first access

Closes https://github.com/eXist-db/exist/issues/5103

## Test plan

- [x] New test `fnLangInLibraryModuleWithVariable` passes
- [x] Full `FunLangTest` suite passes (4/4)
- [x] XQTS `fn-lang` test set passes (36/36, 0 failures)

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunLang.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunLang.java
@@ -99,10 +99,11 @@ public class FunLang extends Function {
 	public void analyze(AnalyzeContextInfo contextInfo) throws XPathException {
 		super.analyze(contextInfo);
 		try {
-			query = context.getBroker().getBrokerPool().getXQueryService().compile(context, queryString);
+			final XQueryContext innerContext = context.copyContext();
+			query = context.getBroker().getBrokerPool().getXQueryService().compile(innerContext, queryString);
 		} catch (final PermissionDeniedException e) {
 			throw new XPathException(this, e);
-		}		
+		}
 	}
 	
 	public Sequence eval(Sequence contextSequence, Item contextItem) throws XPathException {
--- a/exist-core/src/test/java/org/exist/xquery/functions/fn/FunLangTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/functions/fn/FunLangTest.java
@@ -22,14 +22,24 @@
 package org.exist.xquery.functions.fn;
 
 import com.googlecode.junittoolbox.ParallelRunner;
+import org.exist.TestUtils;
 import org.exist.test.ExistXmldbEmbeddedServer;
+import org.exist.xmldb.EXistResource;
+import org.exist.xmldb.XmldbURI;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


joewiz:
cc/ @djbpitt

line-o (approved):
I am curious if this is a more general error class we will find elsewhere as well.

Note: Have a close look at all functions that call compile in their implementation.

joewiz:

@line-o — good instinct to ask. We audited all functions that call `compile()`:

| Call site | Context used | Safe? |
|---|---|---|
| `fn:lang()` (FunLang.java:102) | `context` (shared) | **NO** — this PR fixes it |
| `util:eval()` (Eval.java:474) | `innerContext` (fresh) | Yes |
| `util:compile()` (Compile.java:122) | `new XQueryContext(...)` (fresh) | Yes |
| `fn:transform()` (Transform.java:241) | XSLT compiler (different mechanism) | Yes |

`fn:lang()` was the only function compiling with the shared outer context. All other compile calls use a fresh context or a different compilation mechanism entirely.

reinhapa (approved):

PR #6090, opened 2026-03-03
[bugfix] Reject xs:string as argument to format-date()

### LINKED ISSUE #4425: [BUG] format-date() incorrectly accepts a string as first parameter
**Describe the bug**

`format-date('2022-05-30', '[MNn] [D], [Y]')` returns a formatted date in eXist-db (6.0.1) and raises an error in \<oXygen/\>:

> XPath failed due to: The required item type of the first argument of fn:format-date() is xs:date. The supplied value is of type xs:string

I think \<oXygen/\> is correct here because [the spec](https://www.w3.org/TR/xpath-functions-31/#func-format-date) says that the first argument to the function should be of type `xs:date`.

**Expected behavior**

I expect eXist-db to raise an error if a string is submitted as the first argument to `format-date()`. 

**To Reproduce**

Execute the example above in eXide.

 - OS: macOS 12.4
 - eXist-db version: 6.0.1
 - Java Version: 1.8.0_33

**Additional context**
- Any custom changes in e.g. `conf.xml`? no

joewiz:
## Summary

- **Fix:** `format-date('2022-05-30', '[MNn] [D], [Y]')` now correctly raises `XPTY0004` instead of silently converting the string to a date
- **Root cause:** `DynamicTypeCheck.check()` had commented-out type hierarchy guards on the DATE and DURATION conversion branches, allowing any type to be converted to date/duration types
- **Change:** Uncomment the `Type.subTypeOf(type, requiredType)` guards so only within-family conversions are permitted (e.g., `xs:duration` → `xs:dayTimeDuration`), while `xs:string` → `xs:date` is correctly rejected

Closes #4425

## Test plan

- [x] New XQSuite test `fd:string-instead-of-date()` verifies `XPTY0004` is raised for string argument
- [x] All existing `DateTests` pass (122 tests, 0 failures)
- [x] XQTS conformance: zero regressions across `fn-format-date`, `fn-format-dateTime`, `fn-format-time`, and `fn-months-from-duration`

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/DynamicTypeCheck.java
+++ b/exist-core/src/main/java/org/exist/xquery/DynamicTypeCheck.java
@@ -110,8 +110,7 @@ public class DynamicTypeCheck extends AbstractExpression {
             //Then, if duration, try to refine the type
             //No test on the type hierarchy ; this has to pass :
             //fn:months-from-duration(xs:duration("P1Y2M3DT10H30M"))
-            //TODO : find a way to enforce the test (by making a difference between casting and treating as ?)
-            } else if (Type.subTypeOf(requiredType, Type.DURATION) /*&& Type.subTypeOf(type, requiredType)*/) {
+            } else if (Type.subTypeOf(requiredType, Type.DURATION) && Type.subTypeOf(type, requiredType)) {
                 try {
                     item = item.convertTo(requiredType);
                 //No way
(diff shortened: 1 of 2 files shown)


joewiz:
cc/ @djbpitt

line-o (approved):
Small concise change, thanks!

duncdrum (approved):
nice one

PR #6091, opened 2026-03-03
Fix type checking of 'as' clause in for-binding

### LINKED ISSUE #3553: Error in type-checking with "as" in for-statement [BUG]
**Describe the bug**

```xquery
for $i as xs:string in ('a', 'b', 'c')
return <p>{$i}</p>
```

in eXist-db raises XPTY0004 (type error), continuing “Expected xs:string, got element()“. Saxon does not raise an error. It appears, then, as if an `as` clause that specifies the datatype of the variable declared in a `for` statement is checked against the type of the `return` value at the end of the FLWOR expression, instead of the type of the variable.

**Expected behavior**

I expect the `for` statement in the preceding example to pass type checking.

**To Reproduce**

Execute:

```xquery
for $i as xs:string in ('a', 'b', 'c')
return <p>{$i}</p>
```

in eXist-db, which raises a type error. Execute the same code in Saxon, which runs to completion, without error.

 - OS: macOS 10.15.7
 - eXist-db version: eXist 5.3.0-SNAPSHOT 587dde5f72791201512956125d4d8d0f6cc0a319 20201002114117
 - Java Version: 1.8.0_191

**Additional context**

joewiz:
## Summary

- Fix incorrect type validation of `as` clause in `for` bindings that checked the **return expression result** instead of the **iteration variable value**
- Expressions like `for $i as xs:string in ('a', 'b', 'c') return <p>{$i}</p>` no longer incorrectly raise XPTY0004
- Also fixes the related XQTS test case `hof-046` (function type in for-binding)

## Details

The `as` type declaration in a `for` binding specifies the expected type of each item bound to the iteration variable ([XQuery 3.1 spec §3.12.2](https://www.w3.org/TR/xquery-31/#prod-xquery31-ForBinding)). The previous implementation had two bugs in `ForExpr.java`:

1. **Per-item type checking was skipped** when a `sequenceType` was declared — `var.checkType()` was only called when `sequenceType == null`
2. **Post-loop type checking validated the wrong thing** — after the loop, the code checked `resultSequence` (output of the `return` clause) against the for-binding's type declaration, instead of the binding sequence

The fix ensures `var.checkType()` is always called per iteration item (validating the bound value against the declared type) and removes the incorrect post-loop check on the return expression's result.

Closes #3553

## Test plan

- [x] Added 5 xqsuite tests to `flwor.xql`:
  - `for-as-string-return-element`: string type declaration with element return (the reported bug)
  - `for-as-function`: function type declaration (XQTS hof-046 equivalent)
  - `for-as-type-mismatch`: type mismatch still raises error
  - `for-as-decimal-with-integers`: subtype promotion works
  - `for-as-string-binding`: string binding passes type check
- [x] All 945 XQuery3 tests pass (0 failures)
- [x] XQTS `prod-ForClause`: 188 tests, 167 pass, 21 failures — identical to develop baseline (no regressions)

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ForExpr.java
+++ b/exist-core/src/main/java/org/exist/xquery/ForExpr.java
@@ -189,29 +189,6 @@ public class ForExpr extends BindingExpression {
         }
 
         clearContext(getExpressionId(), in);
-        if (sequenceType != null) {
-            //Type.EMPTY is *not* a subtype of other types ; checking cardinality first
-            //only a check on empty sequence is accurate here
-            if (resultSequence.isEmpty() &&
-                    !sequenceType.getCardinality().isSuperCardinalityOrEqualOf(Cardinality.EMPTY_SEQUENCE))
-                {throw new XPathException(this, ErrorCodes.XPTY0004,
-                    "Invalid cardinality for variable $" + varName + ". Expected " +
-                    sequenceType.getCardinality().getHumanDescription() +
-                    ", got " + Cardinality.EMPTY_SEQUENCE.getHumanDescription());}
-            //TODO : ignore nodes right now ; they are returned as xs:untypedAtomicType
-            if (!Type.subTypeOf(sequenceType.getPrimaryType(), Type.NODE)) {
-                if (!resultSequence.isEmpty() &&
-                        !Type.subTypeOf(resultSequence.getItemType(),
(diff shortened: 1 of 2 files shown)


joewiz:
cc/ @djbpitt

line-o (approved):
LGTM

duncdrum (approved):

PR #6093, opened 2026-03-03
Fix false XPTY0004 from index functions with for-bound variables

### LINKED ISSUE #2204: ngram:contains() bug?
Thank you for reporting your issue!

To be able to better understand you problem or suggestion, please add as much information as possible to this ticket. When there is no or just little information, it will be too difficult to work on it (and hence we will not be able to solve it)

Please fill in the following sections:

I am filtering a list of titles (in a single auxiliary document) to retain only those
that contain a particular substring (using the ngram index). Once I have found
the exact auxiliary titles that contain the target substring, I am using those
exact titles to filter a collection of manuscript descriptions (each in a
separate XML file) to keep only those that have a matching `<title>` element. I'm
using `ngram:contains()`, rather than an explicit equality test, because it gets me 
case-insensitivity and yields the correct results (that is, substring matches don't 
contaminate the results).

```xquery
xquery version "3.1";
declare variable $mss as document-node()+ := collection('/db/repertorium/mss');
declare variable $auxTitles as element(title)* :=
doc('/db/repertorium/aux/titles_cyrillic.xml')//title;
declare variable $target as xs:string :=
request:get-parameter('target','константин');
let $bgTitles := $auxTitles[ngram:contains(*,$target)]/bg
for $title in $bgTitles
where $mss/descendant::tei:title[ngram:contains(.,$title)]
return $title
```

it errors out with:

```
exerr:ERROR XPTY0004: The actual cardinality for parameter 2 does not match the
cardinality declared in the function's signature: ngram:contains($nodes as
node()*, $queryString as xs:string?) node()*. Expected cardinality: zero or one,
got 16. [at line 8, column 52, source: xquery version "3.1"; declare namespace
tei="http://www.tei-c.org/ns/1.0"; declare variable $mss as document-node()+ :=
collection('/db/repertorium/mss'); declare variable $auxTitles as
element(title)* := doc('/db/repertorium/aux/titles_cyrillic.xml')//title;
declare variable $target as xs:string :=
request:get-parameter('target','константин'); let $bgTitles :=
$auxTitles[ngram:contains(*,$target)]/bg for $title in $bgTitles where
$mss/descendant::tei:title[ngram:contains(.,$title)] return $title]
```

If I'm reading this correctly, `ngram:contains()` thinks that `$title` has  a
cardinality not of 1 (the actual cardinality of `$title`), but of 16 (the
cardinality of `$bgTitles`, that is, of the sequence variable in the `for`
statement, rather than of the range variable).

If I change the manuscript filtering to use general equality instead of
`ngram:contains()`, as follows:

```xquery
xquery version "3.1";
declare variable $mss as document-node()+ := collection('/db/repertorium/mss');
declare variable $auxTitles as element(title)* :=
doc('/db/repertorium/aux/titles_cyrillic.xml')//title;
declare variable $target as xs:string :=
request:get-parameter('target','константин');
let $bgTitles := $auxTitles[ngram:contains(*,$target)]/bg
for $title in $bgTitles
where $mss/descendant::tei:title[lower-case(.) = lower-case($title)]
return $title
```

it returns the expected results. 

I expect that the range variable in a `for` statement will always have a cardinality of 1, and that, therefore, `ngram:contains()` will never raise a cardinality error about its second argument when the second argument is the range variable in a `for` statement.

I would be happy to make my data available on request, but if the issue is, in fact, a bug in the implementation of `ngram:contains()`, it should be reproducible with other data. 

### Context information
- eXist-db version + Git Revision hash e.g. eXist-db 3.0 /  acd0c14: 5.0RC4 (same result with 3.4.1, but not with 2.2)
- Java version (e.g. Java8u121): 10.0.2
- Operating system (Windows 7, Linux, MacOs): MacOS High Sierra (10.13.6)
- 32 or 64 bit: 64
- How is eXist-db installed? (JAR installer, DMG, WAR file, clone from GitHub): Homebrew
- Any custom changes in e.g. conf.xmlL: No

joewiz:
## Summary

- Fix `getDependencies()` in all five `Optimizable` index functions (`NGramSearch`, lucene `Query`, lucene `QueryField`, range `Lookup`, range `FieldLookup`) to propagate variable dependencies from all arguments, not just argument 0
- Add XQSuite regression tests for all three index modules (ngram, lucene, range) and a throughput benchmark

Closes #2204

## Problem

When `ngram:contains()` (or `ft:query()`, `ft:query-field()`, `range:eq()`, `range:field-eq()`) is called with a `for`-bound variable as the query argument inside a where clause, it raises a false XPTY0004 error. The original issue report demonstrates this pattern:

```xquery
xquery version "3.1";
declare variable $mss as document-node()+ := collection('/db/repertorium/mss');
declare variable $auxTitles as element(title)* :=
doc('/db/repertorium/aux/titles_cyrillic.xml')//title;
declare variable $target as xs:string :=
request:get-parameter('target','константин');
let $bgTitles := $auxTitles[ngram:contains(*,$target)]/bg
for $title in $bgTitles
where $mss/descendant::tei:title[ngram:contains(.,$title)]
return $title
```

```
exerr:ERROR XPTY0004: The actual cardinality for parameter 2 does not match the
cardinality declared in the function's signature: ngram:contains($nodes as
node()*, $queryString as xs:string?) node()*. Expected cardinality: zero or one,
got 16. [at line 8, column 52]
```

## Root cause

The bug triggers under two specific conditions:

1. **Persistent node set**: The `for` expression iterates over nodes from the database (`in.isPersistentSet()` returns true).
2. **No CONTEXT_ITEM in where clause**: The where expression's path starts from a variable (`$mss`) rather than a function call like `collection()`. The default `Function.getDependencies()` returns `CONTEXT_ITEM`, which would prevent `WhereClause.preEval()` from running. A `VariableReference` does *not* include `CONTEXT_ITEM`, so the preEval optimization fires.

When both conditions are met:

1. `ForExpr.eval()` sets the for-binding variable to the **full input sequence** before calling `WhereClause.preEval()` — this is needed for the bulk where-clause optimization.
2. `WhereClause.preEval()` checks `!Dependency.dependsOn(whereExpr, Dependency.CONTEXT_ITEM)` to decide if it can evaluate the where expression in one shot. Since the path starts from a variable, this check passes.
3. The five `Optimizable` index functions override `getDependencies()` and only check **argument 0's** dependencies. They never propagate `LOCAL_VARS` from argument 1+ (the query string/value argument).
4. So when argument 1 is a for-bound variable (`$title`), the `LOCAL_VARS` dependency is lost, the preEval condition passes, and the where expression is evaluated with `$title` bound to the full sequence.
5. The `DynamicCardinalityCheck` wrapper on `$title` (expected `ZERO_OR_ONE`) sees the full sequence and throws XPTY0004.

Standard functions (`fn:contains`, etc.) are unaffected because `Function.getDependencies()` always returns `CONTEXT_ITEM | CONTEXT_SET`, which causes the preEval check to be skipped.

## Why the fix is in each index function, not in the core

A natural question is whether this should be fixed centrally in `WhereClause.preEval()` or `ForExpr.eval()` rather than in five separate index functions. The answer is no — the dependency information is already lost by the time it reaches the core:

- `WhereClause.preEval()` correctly checks `getDependencies()` on the where expression, and the result propagates up through the expression tree. If `ngram:contains(., $title)` doesn't report `LOCAL_VARS`, no amount of checking in `WhereClause` can recover that information.
- `GeneralComparison` (`=` operator) proves the dependency system works when implemented correctly — it [explicitly propagates](https://github.com/eXist-db/exist/blob/212491410b/exist-core/src/main/java/org/exist/xquery/GeneralComparison.java#L293) `VARS` dependencies from its operands, which is exactly the pattern the `Optimizable` functions should follow.
- The bug is a **`getDependencies()` contract violation** in each function: they opt into bulk evaluation by returning `CONTEXT_SET` without `CONTEXT_ITEM`, but incorrectly report their dependencies by ignoring arguments beyond the first.

## Performance

The `getDependencies()` fix adds a `Dependency.dependsOnVar()` check (a single bitwise AND on a cached int) per non-first argument, called once during query analysis. A throughput benchmark (`NGramContainsBenchmark`) confirms no measurable overhead.

`NGramContainsBenchmark` is excluded from the normal test suite (the class name does not match Surefire's `*Test` discovery pattern, and an `Assume` guard requires `-Dexist.run.benchmarks=true`). To run it:

```bash
mvn test -pl extensions/indexes/ngram -Dtest=NGramContainsBenchmark \
```

Results:

```
query          avg(ms)   min(ms)       ops/s
──────────────────────────────────────────────
literal          0.553     0.361      1808.3
let-var          0.429     0.355      2329.5
for-var          0.525     0.428      1903.8
```

## Test plan

- [x] New XQSuite regression tests for all three index modules (ngram, lucene, range) — each tests for-bound variables in where clauses using module-level collection variables (the pattern that triggers the bug)
- [x] New Java regression test in `CustomIndexTest` for ngram
- [x] **Verified**: all for-variable tests **fail** without the fix and **pass** with it:
  - ngram: XPTY0004 ("Expected cardinality: zero or one, got 3")
  - lucene: XPTY0004 for both `ft:query()` and `ft:query-field()`
  - range: wrong results for `range:eq()` and `range:field-eq()`
- [x] Full test suite passes for all three modules (ngram: 18, lucene: 416, range: 348 — 0 failures)
- [x] XQTS — not directly relevant since these changes are in index extension modules only (XQTS does not exercise ngram/lucene/range index functions)

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Query.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Query.java
@@ -326,6 +326,14 @@ public class Query extends Function implements Optimizable {
         final Expression stringArg = getArgument(0);
         if (Type.subTypeOf(stringArg.returnsType(), Type.NODE) &&
                 !Dependency.dependsOn(stringArg, Dependency.CONTEXT_ITEM)) {
+            // Check if any other argument depends on local/context variables.
+            // If so, the expression cannot be bulk-evaluated during ForExpr preEval
+            // because the variable value changes per iteration. (GH-2204)
+            for (int i = 1; i < getArgumentCount(); i++) {
+                if (Dependency.dependsOnVar(getArgument(i))) {
+                    return Dependency.CONTEXT_SET + Dependency.CONTEXT_ITEM;
+                }
+            }
             return Dependency.CONTEXT_SET;
         } else {
             return Dependency.CONTEXT_SET + Dependency.CONTEXT_ITEM;
--- a/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryField.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryField.java
@@ -167,6 +167,14 @@ public class QueryField extends Query implements Optimizable {
 
     @Override
     public int getDependencies() {
+        // Check if any argument depends on local/context variables.
+        // If so, the expression cannot be bulk-evaluated during ForExpr preEval
+        // because the variable value changes per iteration. (GH-2204)
+        for (int i = 0; i < getArgumentCount(); i++) {
+            if (Dependency.dependsOnVar(getArgument(i))) {
+                return Dependency.CONTEXT_SET + Dependency.CONTEXT_ITEM;
+            }
+        }
         return Dependency.CONTEXT_SET;
     }
 
--- a/extensions/indexes/lucene/src/test/xquery/lucene/for-variable.xql
+++ b/extensions/indexes/lucene/src/test/xquery/lucene/for-variable.xql
@@ -1,0 +1,132 @@
+(:
+ : eXist-db Open Source Native XML Database
+ : Copyright (C) 2001 The eXist-db Authors
+ :
+ : info@exist-db.org
+ : http://www.exist-db.org
+ :
+ : This library is free software; you can redistribute it and/or
+ : modify it under the terms of the GNU Lesser General Public
+ : License as published by the Free Software Foundation; either
+ : version 2.1 of the License, or (at your option) any later version.
+ :
+ : This library is distributed in the hope that it will be useful,
+ : but WITHOUT ANY WARRANTY; without even the implied warranty of
+ : MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ : Lesser General Public License for more details.
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+(:~
+ : Regression tests for GH-2204: ft:query() and ft:query-field() must not
+ : raise XPTY0004 when a for-bound variable is used as the query argument.
+ :
+ : The bug triggers when:
+ : 1. The for-expression iterates over a persistent node set from the database
+ :    (WhereClause.preEval() requires in.isPersistentSet()).
+ : 2. The where clause path starts from a variable (not collection() directly),
+ :    so the where expression does not have a CONTEXT_ITEM dependency (which
+ :    would prevent preEval from running).
+ :)
+module namespace fv="http://exist-db.org/xquery/lucene/for-variable/test";
+
+declare namespace test="http://exist-db.org/xquery/xqsuite";
+
+declare variable $fv:XCONF as element(collection) :=
+    <collection xmlns="http://exist-db.org/collection-config/1.0">
+        <index>
+            <lucene>
+                <analyzer class="org.apache.lucene.analysis.standard.StandardAnalyzer"/>
+                <text qname="description"/>
+                <text field="desc" qname="description"/>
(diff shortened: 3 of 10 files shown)


joewiz:
cc/ @djbpitt

duncdrum (approved):

line-o (approved):
LGTM

PR #6095, opened 2026-03-03
Fix fn:xml-to-json failing on stored XML

### LINKED ISSUE #3989: [BUG] fn:xml-to-json may fail on stored XML representations of JSON
**Describe the bug**
`fn:xml-to-json#1` will sometimes fail to translate a [valid](https://www.w3.org/XML/Group/qtspecs/specifications/xpath-functions-31/html/schema-for-json.xsd) XML representation of JSON, when that XML has been stored in the database. The error returned is FOJS0006, “Invalid XML representation of JSON.”

**Expected behavior**
I expected that the function would take the pseudo-JSON document or outermost node and convert the XML into a JSON string. The same XML can be successfully converted when it’s stored and referenced as an XQuery variable.

**To Reproduce**
```Xquery
xquery version "3.1";

import module namespace jx="http://joewiz.org/ns/xquery/json-xml"
at "https://gist.githubusercontent.com/joewiz/d986da715facaad633db/raw/07f5eb5804673a38491bd4f4e6e26ca2986934cb/json-xml.xqm";
declare namespace err = "http://www.w3.org/2005/xqt-errors";

        <map xmlns="http://www.w3.org/2005/xpath-functions">
            <string key="hello">World</string>
        </map>
    };

    declare
        %test:setUp
    function t:setup() {
        xmldb:store("/db", "test.xml", $t:XML)
    };
    
    declare
        %test:tearDown
    function t:tearDown() {
        xmldb:remove("/db", "test.xml")
    };
    
(:  TESTS  :)
    
    (:~ Turn stored pseudo-JSON XML into JSON. This one fails in v5.3.0. :)
    declare %test:assertTrue function t:exist-doc() {
        fn:xml-to-json(doc("/db/test.xml"))
    };
    
    (:~ Turn in-memory pseudo-JSON into JSON. :)
    declare %test:assertTrue function t:exist-in-memory() {
        fn:xml-to-json($t:XML)
    };
    
    (: Use Joe Wicentowski's shim for xml-to-json() to turn stored pseudo-JSON 
        into JSON. :)
    declare %test:assertTrue function t:shim-doc() {
        jx:xml-to-json(doc("/db/test.xml"))
    };
```

- macOS v11.4
- eXist v5.3.0 via the JAR installer, no configuration
- Java OpenJDK v1.8.0_292

joewiz:
## Summary

- `fn:xml-to-json` threw FOJS0006 ("Invalid XML representation of JSON") when applied to XML stored in the database, but worked correctly on in-memory XML
- Root cause: `EmbeddedXMLStreamReader.getAttributeValue(null, localName)` never matched any attributes because `"".equals(null)` is `false`
- One-line fix aligns `EmbeddedXMLStreamReader` with `InMemoryXMLStreamReader` and the `javax.xml.stream.XMLStreamReader` javadoc

Closes #3989

## What Changed

**`exist-core/src/main/java/org/exist/stax/EmbeddedXMLStreamReader.java`**
- Added null guard for `namespaceURI` parameter in `getAttributeValue()`, matching the pattern already used by `InMemoryXMLStreamReader` and specified by the `XMLStreamReader` javadoc: "If the namespaceURI is null the namespace is not checked for equality"

**`exist-core/src/test/xquery/xquery3/xml-to-json.xql`**
- Added setUp/tearDown to store test XML documents in the database
- Added 3 XQSuite tests: simple map, nested structure, and stored-vs-in-memory equivalence

## Test Plan

- [x] New XQSuite tests pass for stored XML documents
- [x] All 943 existing XQuery3 tests pass (0 failures, 0 errors)

### DIFF
--- a/exist-core/src/main/java/org/exist/stax/EmbeddedXMLStreamReader.java
+++ b/exist-core/src/main/java/org/exist/stax/EmbeddedXMLStreamReader.java
@@ -355,7 +355,7 @@ public class EmbeddedXMLStreamReader implements IEmbeddedXMLStreamReader, Extend
         readAttributes();
         for(int i = 0; i < attributes.getLength(); i++) {
             final org.exist.dom.QName qn = attributes.getQName(i);
-            if(qn.getNamespaceURI().equals(namespaceURI) && qn.getLocalPart().equals(localName)) {
+            if((namespaceURI == null || qn.getNamespaceURI().equals(namespaceURI)) && qn.getLocalPart().equals(localName)) {
                 return attributes.getValue(i);
             }
         }
--- a/exist-core/src/main/java/org/exist/stax/EmbeddedXMLStreamReader.java
+++ b/exist-core/src/main/java/org/exist/stax/EmbeddedXMLStreamReader.java
@@ -355,6 +355,10 @@ public class EmbeddedXMLStreamReader implements IEmbeddedXMLStreamReader, Extend
         readAttributes();
         for(int i = 0; i < attributes.getLength(); i++) {
             final org.exist.dom.QName qn = attributes.getQName(i);
+            // Per XMLStreamReader contract, null namespaceURI means "no namespace" —
+            // match any attribute regardless of its namespace. This is consistent with
+            // the JDK's default XMLStreamReader implementations (e.g., Woodstox, Aalto)
+            // where getAttributeValue(null, localName) matches unqualified attributes.
             if((namespaceURI == null || qn.getNamespaceURI().equals(namespaceURI)) && qn.getLocalPart().equals(localName)) {
                 return attributes.getValue(i);
             }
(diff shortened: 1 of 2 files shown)


duncdrum (approved):

line-o (approved):
I would like to see the explanation added here as a comment so that when I read the line next time I know that this is done in the same way in other places as well.

More generally speaking: should we be more explicit

- treat "" and null as empty namespaces and return the attribute value but
- fail if null is set but the attribute has a namespace set?

joewiz:

Added a comment explaining the null namespace handling and its consistency with standard XMLStreamReader implementations in 435ba2e.

Re: being more explicit about `""` vs `null` — I think the current behavior (treating null as "match any namespace") is the pragmatic choice and matches what callers like `FunXmlToJson` expect. Making it stricter could break existing usage. Open to discussing further if you feel strongly.

line-o:
Thanks, that argument is convincing and makes sense!

Issue #3632, opened 2020-11-30
[feature] Integrate XQuery Test Suite into CI for PR checks


joewiz:
Integrating an automated run https://github.com/eXist-db/exist-xqts-runner into the CI checks performed with every PR will allow the eXist community to take advantage of the 30K+ XQuery tests created by the W3C working group and community. CI should report on the % of passed tests before the PR and with the PR, and should note newly failed or fixed tests. Test results should be retrievable for some time.

line-o:
Running this on each PR (on travis) would eat up our credits in no time. I therefore suggest to create a GitHub action that can be manually triggered on any branch.

joewiz:
@line-o Makes sense!

Also somewhat related to the discussion of CI-based testing: https://github.com/eXist-db/exist/pull/3341.

duncdrum:
see #3766 
so what are the parameters for how we want this to look?

adamretter:
The problem is not plumbing this into the CI. That has never really been the issue.

The problems are:
1. where to upload the JUnit results?
2. how to process those results into summary stats?
3. and where to build and host graphs that show those results and the change over time?

duncdrum:
1. and 3. => GitHub (pages)
2. can you expand on what processing is necessary

joewiz:
Closed by #4661.

PR #6098, opened 2026-03-04
[ci] Add error diagnostics to XQTS artifact retrieval

joewiz:
## Summary

- Improve error diagnostics in the XQTS CI workflow's artifact retrieval steps, so that when a failure occurs, developers can quickly identify the root cause

## Problem

The XQTS workflow compares the current run's test results against a **previous** run's results from the `develop` branch. It does this by fetching the list of `xqts-logs` artifacts via the GitHub API and selecting index `[1]` — skipping `[0]` (the artifact just uploaded by the current run) to get the prior baseline. This means **at least 2 develop-branch artifacts must exist** for the comparison to work.

When fewer than 2 exist (e.g., after artifact expiration, repository migration, or manual cleanup), the step failed with a generic error and no context, making it difficult to diagnose.

## Changes

Added error handling and diagnostics across four failure conditions:

| Condition | Step | Error Message | Additional Diagnostics |
|---|---|---|---|
| Non-200 HTTP status fetching artifact list | Get Previous XQTS Logs Artifacts JSON | `Failed to fetch artifacts list (HTTP <code>)` | Logs `x-ratelimit-*` headers and full response body |
| API response contains `.message` (e.g. rate limit, auth error) | Extract Previous XQTS Logs Artifact URL | `GitHub API error: <message>` | Logs full response body |
| Fewer than 2 develop-branch artifacts | Extract Previous XQTS Logs Artifact URL | `No previous XQTS artifact found for the develop branch (need at least 2 artifacts from develop, found <n>)` | Logs total artifact count and develop-branch count |
| Non-200 HTTP status downloading artifact | Download Previous XQTS Logs Artifact | `Failed to download artifact (HTTP <code>)` | — |

## Test plan

- [ ] Verify CI runs green when 2+ develop-branch artifacts exist
- [ ] Confirm the error message and diagnostics are visible in the GitHub Actions log when the failure condition is triggered

### DIFF
--- a/.github/workflows/ci-xqts.yml
+++ b/.github/workflows/ci-xqts.yml
@@ -49,35 +49,69 @@ jobs:
           name: xqts-logs
           retention-days: 14
           path: /tmp/xqts-output
-      - name: Get Previous XQTS Logs Artifacts JSON
+      - name: Find Previous XQTS Logs Artifact from develop
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          HTTP_CODE=$(curl -s -o /tmp/previous-xqts-logs-artifacts.json -D /tmp/previous-xqts-logs-headers.txt -w "%{http_code}" \
-            -H "Accept: application/vnd.github+json" \
-            -H "Authorization: Bearer ${GITHUB_TOKEN}" \
-            -H "X-GitHub-Api-Version: 2022-11-28" \
-            https://api.github.com/repos/exist-db/exist/actions/artifacts?name=xqts-logs)
-          if [ "$HTTP_CODE" -ne 200 ]; then
-            echo "::error::Failed to fetch artifacts list (HTTP $HTTP_CODE)"
-            grep -i "x-ratelimit" /tmp/previous-xqts-logs-headers.txt || true
-            cat /tmp/previous-xqts-logs-artifacts.json
-            exit 1
-          fi
-      - name: Extract Previous XQTS Logs Artifact URL
-        run: |
-          # Check for API-level errors (rate limiting, auth issues) in the response
-          api_message=$(jq -r '.message // empty' /tmp/previous-xqts-logs-artifacts.json)
-          if [ -n "$api_message" ]; then
-            echo "::error::GitHub API error: $api_message"
-            cat /tmp/previous-xqts-logs-artifacts.json
-            exit 1
-          fi
-          total_count=$(jq -r '.total_count' /tmp/previous-xqts-logs-artifacts.json)
-          echo "Total xqts-logs artifacts found: $total_count"
-          develop_count=$(jq '[.artifacts[] | select(.workflow_run.head_branch == "develop")] | length' /tmp/previous-xqts-logs-artifacts.json)
+          # Paginate through artifacts to find at least 2 from the develop branch.
+          # The GitHub API returns 30 items per page by default; we request 100.
+          # Among those, only a few may be from 'develop', so we paginate until
+          # we find 2 (current + previous) or exhaust all pages.
+          develop_artifacts="[]"
+          page=1
+          max_pages=20
+          total_count=0
+
+          while [ "$page" -le "$max_pages" ]; do
+            HTTP_CODE=$(curl -s -o /tmp/artifacts-page.json -D /tmp/artifacts-headers.txt -w "%{http_code}" \
+              -H "Accept: application/vnd.github+json" \
+              -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+              -H "X-GitHub-Api-Version: 2022-11-28" \
+              "https://api.github.com/repos/exist-db/exist/actions/artifacts?name=xqts-logs&per_page=100&page=${page}")
+            if [ "$HTTP_CODE" -ne 200 ]; then
+              echo "::error::Failed to fetch artifacts list page $page (HTTP $HTTP_CODE)"
+              grep -i "x-ratelimit" /tmp/artifacts-headers.txt || true
+              cat /tmp/artifacts-page.json
+              exit 1
+            fi
+
+            api_message=$(jq -r '.message // empty' /tmp/artifacts-page.json)
+            if [ -n "$api_message" ]; then
+              echo "::error::GitHub API error on page $page: $api_message"
+              cat /tmp/artifacts-page.json
+              exit 1
+            fi
+
+            if [ "$page" -eq 1 ]; then
+              total_count=$(jq -r '.total_count' /tmp/artifacts-page.json)
+              echo "Total xqts-logs artifacts found: $total_count"
+            fi
+
+            page_count=$(jq '.artifacts | length' /tmp/artifacts-page.json)
+            if [ "$page_count" -eq 0 ]; then
+              echo "No more artifacts on page $page, stopping."
+              break
+            fi
+
+            # Extract develop-branch artifacts from this page and append them
+            page_develop=$(jq '[.artifacts[] | select(.workflow_run.head_branch == "develop")]' /tmp/artifacts-page.json)
+            develop_artifacts=$(jq -n --argjson existing "$develop_artifacts" --argjson new "$page_develop" '$existing + $new')
+
+            develop_count=$(echo "$develop_artifacts" | jq 'length')
+            echo "Page $page: found $page_count artifacts, $develop_count develop artifacts so far"
+
+            if [ "$develop_count" -ge 2 ]; then
+              echo "Found enough develop artifacts."
+              break
+            fi
+
+            page=$((page + 1))
+          done
+
+          develop_count=$(echo "$develop_artifacts" | jq 'length')
           echo "Artifacts from develop branch: $develop_count"
-          url=$(jq -r "[.artifacts[] | select(.workflow_run.head_branch == \"develop\")][1].archive_download_url" /tmp/previous-xqts-logs-artifacts.json)
+
+          url=$(echo "$develop_artifacts" | jq -r '.[1].archive_download_url')
           if [ "$url" = "null" ] || [ -z "$url" ]; then
             echo "::error::No previous XQTS artifact found for the develop branch (need at least 2 artifacts from develop, found $develop_count)"
             exit 1
--- a/.github/workflows/ci-xqts.yml
+++ b/.github/workflows/ci-xqts.yml
@@ -50,13 +50,51 @@ jobs:
           retention-days: 14
           path: /tmp/xqts-output
       - name: Get Previous XQTS Logs Artifacts JSON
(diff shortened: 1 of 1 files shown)


joewiz:
The XQTS CI failure here is expected, given the error condition now described in the CI logs:

```
Total xqts-logs artifacts found: 1300
Artifacts from develop branch: 1
Error: No previous XQTS artifact found for the develop branch (need at least 2 artifacts from develop, found 1)
Error: Process completed with exit code 1.
```

Before this PR, only the final line of that message would appear. The first 3 lines now provide a detailed description of the cause of the error.

line-o:
@joewiz I am not convinced that we need two previous runs. The other run, the one we want to compare with is in almost all cases executed on a branch _other_ than `develop`.

line-o:
Please compare with #6099

joewiz:
@line-o I definitely trust your understanding of the CI checks' inner workings more than mine. But for context, Claude's analysis above was:

> The XQTS workflow compares the current run's test results against a previous run's results from the develop branch. It does this by fetching the list of xqts-logs artifacts via the GitHub API and selecting index `[1]` — skipping `[0]` (the artifact just uploaded by the current run) to get the prior baseline. This means at least 2 develop-branch artifacts must exist for the comparison to work.

Also, for context, this CI check was successfully passing before https://github.com/eXist-db/exist/commit/87b766ee09dd3fc23b88e5007ccca5e0f69a30de - and the use of `[1]` has been in this CI check for ~3 years: https://github.com/eXist-db/exist/blame/develop/.github/workflows/ci-xqts.yml#L55.

line-o:
Yeah, there is definitely more to it. But see, for three years now I have been wondering that the results did not feel correct or exact. Now, I see why.
Consider the usual scenario:

- there was a previous run of XQTS which was uploaded to the list of artefacts
- someone opens a PR with changes to the XQuery processor and we are especially keen to see how that affects XQTS test results
- before the comparison the artefacts are uploaded with a head_branch _other than develop_.
- the test results were alwasy compared to the _second_ result _not_ the first one

joewiz:
@line-o Are you sure that "the first one" isn't "the artifact just uploaded by the current run"? Because if it is, we would need to look at "the second one" to find "the previous XQTS run" (that latter quote is from the PR that added this feature, https://github.com/eXist-db/exist/issues/3632). 

I see your point that "the previous XQTS run" may not be from develop (i.e., that the results from all runs from various branches are dumped into a single bucket and could make for a comparison of apples and oranges), but I don't think we'd want to compare the current run to itself.

line-o:
The jq snippet filters all artefacts in the list by their head_branch and this is the branch the action is run in, like `"head_branch": "build/disable-dependency-check",` (see https://api.github.com/repos/exist-db/exist/actions/artifacts?name=xqts-logs).
I made it now extra safe in #6099 by uploading the artefact _after_ the comparison has finished.

line-o:
And https://github.com/eXist-db/exist/actions/runs/22686995593/job/65772506988?pr=6099 is proof that the artefact can be uploaded afterwards.

joewiz:
Also, while I typed my previous response, I was waiting for Claude's answer to a question I posed to it:

> Given the following error message produced by the code in this branch, if there are 1300 artifacts, why does it say, "No previous XQTS artifact found for the develop branch (need at least 2 artifacts from develop, found 1)"? Were the 1300 artifacts not in the `develop` branch but only from other branches? Or is there a flaw in the logic of this error report?
> 
> ```
> Total xqts-logs artifacts found: 1300
> Artifacts from develop branch: 1
> Error: No previous XQTS artifact found for the develop branch (need at least 2 artifacts from develop, found 1)
> Error: Process completed with exit code 1.
> ```

Claude's answer sheds light on the question of results from develop vs. other branches:

> ## Analysis
> 
> The issue is a combination of **both factors you suspected**. Here's what's happening:
> 
> ### 1. The 1300 artifacts span all branches, not just `develop`
> 
> The API call on **line 60** fetches artifacts named `xqts-logs` with no branch filter:
> 
> ```
> https://api.github.com/repos/exist-db/exist/actions/artifacts?name=xqts-logs
> ```
> 
> The `total_count: 1300` in the response is the total across **all branches** (PRs, feature branches, develop, etc.).
> 
> ### 2. The real flaw: pagination
> 
> This is the key bug. The GitHub API **paginates** results — by default it returns only **30 items per page**. The response JSON looks like:
> 
> ```json
> {
>   "total_count": 1300,
>   "artifacts": [ /* only the first 30 artifacts */ ]
> }
> ```
> 
> - **Line 76** reads `.total_count` → reports **1300** (the global total across all pages and all branches)
> - **Line 78** counts items in `.artifacts[]` where `head_branch == "develop"` → but `.artifacts[]` only contains the **first 30** most-recent artifacts, not all 1300
> - **Line 80** tries to grab the **second** develop artifact (`[1]`) from that same small set
> 
> So among the 30 most-recent `xqts-logs` artifacts returned on page 1, only **1** happened to be from the `develop` branch. The other 29 were from PRs or feature branches. Since the code needs index `[1]` (the second develop artifact, to use as the "previous" baseline for comparison), it fails.
> 
> ### Why it needs 2 develop artifacts
> 
> The workflow compares the **current** XQTS run against a **previous** one. The artifacts are sorted newest-first, so `[0]` is the artifact from the *current* run (just uploaded on line 45-51), and `[1]` would be the *previous* run's artifact. That's why it needs at least 2.
> 
> ### The fix
> 
> The API call on line 56-60 should request more results per page to increase the chance of finding 2 develop artifacts. Adding `&per_page=100` to the URL would help, but in a repo with heavy PR activity, even 100 might not be enough. A more robust approach would be to either:
> 
> 1. **Add `&per_page=100`** to the API URL (simple, may suffice)
> 2. **Implement pagination** to fetch additional pages if needed
> 3. **Filter server-side** by adding `&head_branch=develop` to the API query (but note: the artifacts API doesn't natively support branch filtering, so this isn't available)
> 
> Would you like me to implement a fix?

This suggests that the reason we started to see XQTS CI "failures" here is that I submitted a whole bunch of PRs and pushed the "window" of recent CI runs past the first page of results. 

The improved reporting in this PR will at least help us diagnose such issues better. But as it is, it wouldn't fix the problem I caused by submitting so many PRs. I'll have it start work on the "pagination" fix in my fork and see if it manages to locate the artifact that it's currently unable to.

line-o:
@joewiz yes handling of pagination is an important thing to consider
I am actually _very_ happy that you created the edge case that _finally_ made me see the off by one error that kept nagging me.
Let us also not forget about the relatively brief time window that these artefacts will be retained (90 days).

And please also consider looking for alternatives other than even more complex command line invocations.
Have you considered using the `gh` command line tool to achieve some of the necessary work?

line-o:
I would like to see that we merge #6099 now and then improve based on that.

joewiz:
@line-o Great idea to switch to `gh`. I've been using Claude Code's "on the web" edition today on this, and its container doesn't have the `gh` binary. But when I'm back on Claude Code (on my Mac), I'll have it update this PR to use `gh` instead. For now, the most recent commit added above adds pagination. From its PR to my fork:

> The previous code fetched only the first page (30 items) of artifacts from the GitHub API but reported total_count (spanning all pages). When heavy PR activity pushed develop artifacts beyond the first page, the workflow failed with "need at least 2 artifacts from develop, found 1".
>
> Now paginates with per_page=100 (up to 20 pages) and stops as soon as 2 develop-branch artifacts are found.

I'll also clean up this PR to remove the merge commit. Ultimately, the pagination thing is the fix to the error we "suddenly" began seeing, so we could do away with the first commit here, though I think the additional diagnostic info that it deposits in the CI logs if it fails to obtain a previous XQTS artifact could be useful. I imagine this won't be the last time we encounter such a problem with this particular XQTS CI routine, and this diagnostic info could help.

As with the other Claude Code PRs, I have no pride with the "generated" code (though I did prompt it, evaluate its results, steer it along toward what you see here - this wasn't purely on "autopilot"), so if you would prefer a different approach, we can just close this. I won't be offended in the least.

line-o:
I think your exploration did find additional areas we have to look at. The core issue that spoiled our comparisons is fixed in #6099
The off-by-one error I initially reported is still existing in this PR.

joewiz:
@line-o Is there really an off-by-one error? The logic of the code for the last 3 years, as I understand it, is that the 1st artifact (`[0]`) will be from *this* CI run, and the 2nd artifact (`[1]`) will be from the *previous* CI run in develop. 

We agree that the real problem surfaced by my flood of PRs is that the 2nd artifact wasn't in the first page of artifacts. And the most recent commit from Claude added pagination - which according to the latest CI run was [successful](https://github.com/eXist-db/exist/actions/runs/22687888338/job/65775615786?pr=6098#step:12:74).

line-o:
@joewiz yes, I stand by my analysis. `[1]` was wrong since the moment it was introduced. It _may_ have been introduced as a _band aid_ for runs that are comparing a run on the develop branch. But the actual solution to this problem is the reordering of steps.

line-o:
The only reasonable order of events as I see it:

1. run test
2. load most recent test result from develop
3. compare test results
4. upload artefact

line-o:
@joewiz If you think I am wrong here please tell me where.

joewiz:
@line-o Let me try to tease out your analysis of the current state. In https://github.com/eXist-db/exist/pull/6099#issuecomment-4000037411 you wrote:

> The first one is newer, the list is sorted by created date in descending order.

That also reflects my understanding and is what I meant when I wrote:

>  the 1st artifact ([0]) will be from this CI run, and the 2nd artifact ([1]) will be from the previous CI run in develop.

... though I was wrong to say "in develop." I should've said "from across all branches."

In obtaining the 2 pieces of information we need - (1) the current run and (2) the most recent run from develop for comparison - the order in which we obtain them doesn't affect the outcome, right? Getting 1 then 2 or 2 then 1 still allows us to compare the data.

The real problem we encountered in these failures since yesterday is that we were failing to grab (2) - the most recent run for comparison. 

Right?

line-o:
The workflow reads the first page of stored artefacts from GitHub and filters that list by

- branch name == develop
- and then selects the _second_ of them

The list is ordered "newest first".

The most recent run on develop is therefore _never_ selected. 

And this is regardless in which branch the current run is taking place.

**Also important to note:** While the logs / the test result of the _current_ run is uploaded as an artefact, the comparison uses the locally stored logs and compares them against the _one_ result from develop selected in the later step (the second newest).

joewiz:
@line-o Reading your last comment makes it clear to me that my earlier claim that "the 1st artifact ([0]) will be from this CI run" was completely wrong. And now I see why grabbing the most recent previous run before performing the current CI run is important. Because if we didn't, the "most recent previous run" could be in a different position, depending on whether the current CI run itself is in develop or a non-develop branch. Sorry for my misunderstanding! I agree that your approach of changing the order of operations is the core fix to the underlying problem.

PR #6099, opened 2026-03-04
[ci] improve xqts workflow

line-o:
- if only one XQTS run on develop was retained the jq script returned null
- the command line invocations were hard to read and are now split into multiple lines
- temporary files have file endings according to their contents and more meaningful names
- the names of the steps try to be as brief and clear as possible

PR #6105, opened 2026-03-06
[ci] use gh CLI to find previous XQTS baseline on develop

joewiz:
## Summary

- Replaces `curl` + `jq` artifact lookup with `gh run list` and `gh run download` to find the previous XQTS baseline on develop, as [suggested by @line-o](https://github.com/eXist-db/exist/pull/6098#issuecomment-4000068788)
- Fixes the pagination bug where the develop artifact was pushed off the first page of results (default 30 per page), causing all XQTS CI runs to fail with `Could not resolve host: null`
- Builds on #6099

## What Changed

**`.github/workflows/ci-xqts.yml`**
- Replaced 4 steps (list artifacts via curl, extract URL via jq, download zip via curl, unzip) with 2 steps using `gh run list` and `gh run download`
- `gh run list --branch develop --status success` filters server-side, eliminating the pagination problem entirely
- `gh run download` handles authentication, download, and extraction in one step
- Added `--status success` filter to ensure comparison is only against runs that completed successfully
- Added explicit error handling with `::error::` annotation if no previous develop run is found

Note: `gh` CLI [v2.87.3 is pre-installed](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) on `ubuntu-latest` runners.

## Test Plan

- [x] Verify the XQTS CI job passes on this PR (the job itself will exercise the new baseline lookup logic)

### DIFF
--- a/.github/workflows/ci-xqts.yml
+++ b/.github/workflows/ci-xqts.yml
@@ -42,30 +42,35 @@ jobs:
           name: exist-xqts-runner-hprof
           retention-days: 1
           path: /tmp/*.hprof.zst
-      - name: Get List of Previous XQTS Logs
-        run: |
-          curl \
-          -H "Accept: application/vnd.github+json" \
-          -H "X-GitHub-Api-Version: 2022-11-28" \
-          https://api.github.com/repos/exist-db/exist/actions/artifacts?name=xqts-logs \
-          > /tmp/artifacts-list.json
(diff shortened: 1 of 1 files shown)


line-o (approved):
Cool! And just in time as it seems.

reinhapa (approved):

PR #6103, opened 2026-03-06
Add more XQuery 3.1 keywords to reservedKeywords rule

joewiz:
## Summary

- Add 10 missing keywords to the `reservedKeywords` grammar rule so they can be used as NCNames (variable names, function names, etc.)
- Keywords added: `ascending`, `descending`, `greatest`, `least`, `satisfies`, `schema-attribute`, `castable`, `idiv`, `processing-instruction`, `allowing`
- Without these entries, using any of them as a variable name (e.g., `let $ascending := 1`) causes an XPST0003 parse error

## Context

The `reservedKeywords` rule in `XQuery.g` lists keywords that the parser recognizes as valid NCNames. Keywords used in grammar rules but not listed here cannot be used as identifiers, which violates the XQuery specification's rules for contextual keywords.

Continues the pattern from 1dc5913 (added `case`) and addresses the class of issues described in #2381.

## What Changed

| File | Change |
|------|--------|
| `exist-core/.../parser/XQuery.g` | Added 10 keywords to `reservedKeywords` rule |
| `exist-core/.../xquery/ReservedKeywordsAsNCNamesTest.java` | New test class with 9 tests covering each keyword |

## XQTS Results (W3C XQTS 3.1, full suite)

| Branch | Tests | Pass | Fail | Error | Rate |
|--------|-------|------|------|-------|------|
| develop (baseline) | 53546 | 48152 | 5094 | 300 | 89.9% |
| this PR | 53546 | 48134 | 5108 | 304 | 89.9% |

**Improvements** (4 tests newly passing):
- `xquery10keywords`, `xquery10keywords2`, `xquery10keywords3` (prod-LetClause) — keywords as variable names
- `K2-ForExprWithout-38` (prod-ForClause) — `castable` as variable name

**Apparent regressions** (7 tests) are XQTS runner flakiness — axis tests returning `actual='0'` (truncated results from resource contention), not caused by this change. Confirmed by cross-checking against the DSLASH branch where the same tests pass.

## Test Plan

- [x] New JUnit tests pass: `ReservedKeywordsAsNCNamesTest` (9 tests)
- [x] Full exist-core test suite passes: 6503 tests, 0 failures, 0 errors
- [x] Full W3C XQTS 3.1: no real regressions, 4 improvements

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6080, opened 2026-03-03
Fix // followed by reverse axis step being misinterpreted

joewiz:
## Summary

- Fix `//` (DSLASH) followed by a reverse axis step (`preceding::`, `ancestor::`, `ancestor-or-self::`, `parent::`, `preceding-sibling::`) being misinterpreted by the XQuery tree walker
- The tree walker was overwriting the reverse axis with `DESCENDANT_SELF_AXIS`, so e.g. `$node//preceding::node()` behaved like `$node/descendant-or-self::node()` instead of the correct `$node/descendant-or-self::node()/preceding::node()`
- Fix both the `DSLASH` and `ABSOLUTE_DSLASH` handlers to detect reverse axes and insert an explicit `descendant-or-self::node()` step before the reverse axis step, rather than merging them
- Add tests covering `//` with all reverse axes

Closes #691

## Test plan

- [x] New `dslashWithReverseAxis` test in `XPathQueryTest` passes (both local and remote parametrized variants)
- [x] All 150 existing `XPathQueryTest` tests pass with no regressions
- [x] XQTS `prod-AxisStep.*` test suites (345 tests) show zero regressions compared to baseline

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #691, opened 2015-06-24
The // pseudo-operator followed by an axis step is misinterpreted


jensopetersen:
In the following, preceding-1, preceding-2 and preceding-3 should return the same, but preceding-3 returns the same as descendant-or-self.

	let $node :=
	<x>
		<y/>
		<a>a</a>
		<b n="1">b</b>
		<c>c</c>
		<z/>
	</x>
	(:let $node := doc('/db/x.xml'):)
	let $node := $node//b

	return 
		<preceding>
			<preceding-1>{for $node in $node/preceding::node() return <node>{$node}</node>}</preceding-1>
			<preceding-2>{for $node in $node/descendant-or-self::node()/preceding::node() return <node>{$node}</node>}</preceding-2>
			<preceding-3>{for $node in $node//preceding::node() return <node>{$node}</node>}</preceding-3>
			<descendant-or-self>{for $node in $node//node() return <node>{$node}</node>}</descendant-or-self>
		</preceding>

PR #6106, opened 2026-03-06
Fix // with parenthesized and non-LocationStep expressions

joewiz:
## Summary

- Fix the abbreviated descendant-or-self syntax (`//`) when followed by non-LocationStep expressions like `//(@attr)`, `//(a | b)`, and `//$var`
- These expressions previously returned 0 results because `setPrimaryAxis()` is a no-op on non-LocationStep classes

## Root Cause

In `XQueryTree.g`, the DSLASH handler has two branches:

1. **LocationStep** (e.g., `//child::x`): merges the descendant-or-self axis into the step — works correctly
2. **Non-LocationStep** (e.g., `//(@x)`, `//(a | b)`): called `setPrimaryAxis(Constants.DESCENDANT_SELF_AXIS)`, which is defined as an empty method body in `AbstractExpression` — effectively a no-op

## Fix

Split the non-LocationStep branch into three cases:

- **`VariableReference`**: retain `setPrimaryAxis` + wrap in `SimpleStep` (existing behavior, works correctly)
- **`FilteredExpression`**: retain `setPrimaryAxis` + mark abbreviated (preserves Lucene index optimizer, which reads the primary axis from inner PathExprs to determine search strategy)
- **All other non-LocationStep expressions** (e.g., `PathExpr` wrapping parenthesized expressions): insert an explicit `descendant-or-self::node()` LocationStep before the expression, correctly expanding `//expr` to `/descendant-or-self::node()/expr` per spec. `setPrimaryAxis` must NOT be called here because it propagates into inner PathExpr children and corrupts their axes (e.g., overwriting an attribute axis in `//(@x)`). The inserted step must NOT be marked as abbreviated, because `LocationStep.analyze()` rewrites abbreviated descendant-or-self steps to descendant-only, which would skip direct children.

## What Changed

| File | Change |
|------|--------|
| `exist-core/.../parser/XQueryTree.g` | Split non-LocationStep DSLASH handling into three branches: VariableReference, FilteredExpression, and explicit step insertion for all others |
| `exist-core/.../xquery/DescendantOrSelfWithNonLocationStepTest.java` | New test class with 4 tests covering parenthesized attributes, element unions, and axis combinations |

## Spec Reference

[XQuery 3.1 §3.3.5 Abbreviated Syntax](https://www.w3.org/TR/xquery-31/#abbrev): "The abbreviated syntax `//` is short for `/descendant-or-self::node()/`"

## Related

- PR #6080 and issue #691 (fix `//` followed by reverse axis steps — same handler, different branch)
- `K2-Steps-31` (XQTS) fails on both `develop` and this PR due to a pre-existing dedup bug in `PathExpr` when `/` is followed by a function call — will be addressed in a follow-on PR

## XQTS Results (W3C XQTS 3.1, full suite)

| Branch | Tests | Pass | Fail | Error | Rate |
|--------|-------|------|------|-------|------|
| develop (baseline) | 53546 | 48152 | 5094 | 300 | 89.9% |
| this PR | 53546 | 48124 | 5116 | 306 | 89.9% |

**Improvements** (2 tests newly passing):
- `ParenthesizedExpr-8` (prod-ParenthesizedExpr) — `//(@attr)` now works
- `hof-046` (prod-NamedFunctionRef) — higher-order function with `//`

Apparent regressions (9 tests) are pre-existing keyword-as-variable-name parse errors (`ascending`, `castable`, `processing-instruction`), not caused by this change. These are addressed by PR #6103.

## Test Plan

- [x] New JUnit tests pass: `DescendantOrSelfWithNonLocationStepTest` (4 tests)
- [x] Full exist-core test suite passes: 6498 tests, 0 failures, 0 errors
- [x] Full W3C XQTS 3.1: no real regressions, 2 improvements

### DIFF
--- a/exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g
+++ b/exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g
@@ -2989,14 +2989,22 @@ throws PermissionDeniedException, EXistException, XPathException
                         rs.setAbbreviated(true);
                     }
 
-                } else {
+                } else if (rightStep instanceof VariableReference) {
                     rightStep.setPrimaryAxis(Constants.DESCENDANT_SELF_AXIS);
-                    if(rightStep instanceof VariableReference) {
-                        rightStep = new SimpleStep(context, Constants.DESCENDANT_SELF_AXIS, rightStep);
-                        path.replaceLastExpression(rightStep);
-                    } else if (rightStep instanceof FilteredExpression)
-                        ((FilteredExpression)rightStep).setAbbreviated(true);
-
+                    rightStep = new SimpleStep(context, Constants.DESCENDANT_SELF_AXIS, rightStep);
+                    path.replaceLastExpression(rightStep);
+                } else if (rightStep instanceof FilteredExpression) {
+                    rightStep.setPrimaryAxis(Constants.DESCENDANT_SELF_AXIS);
+                    ((FilteredExpression)rightStep).setAbbreviated(true);
+                } else {
+                    // For other non-LocationStep expressions (e.g., PathExpr wrapping
+                    // parenthesized expressions like //(@x) or //(a | b)), insert an
+                    // explicit descendant-or-self::node() step. We must NOT call
+                    // setPrimaryAxis here because it would corrupt inner axes (e.g.,
+                    // overwriting an attribute axis in //(@x)).
+                    LocationStep descStep = new LocationStep(context, Constants.DESCENDANT_SELF_AXIS, new TypeTest(Type.NODE));
+                    path.replaceLastExpression(descStep);
+                    path.add(rightStep);
                 }
             }
         )?
--- a/exist-core/src/test/java/org/exist/xquery/DescendantOrSelfWithNonLocationStepTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/DescendantOrSelfWithNonLocationStepTest.java
@@ -1,0 +1,100 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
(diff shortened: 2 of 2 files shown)


line-o:
The failure of `K2-Steps-31` seems significant

line-o (commented):
This looks like a very valuable addition.

line-o:
Can the commits be squashed into one?

joewiz:

Good eye — K2-Steps-31 does test `$root//local:function(.)`, which per [§3.3.5](https://www.w3.org/TR/xquery-31/#abbrev) expands to `$root/descendant-or-self::node()/local:function(.)`. This PR correctly handles that expansion (inserting an explicit `descendant-or-self::node()` step).

The remaining failure is a **pre-existing dedup bug** — identical on `develop`. Per [§3.3.1.1](https://www.w3.org/TR/xquery-31/#id-path-operator), the `/` operator must eliminate duplicate nodes by identity. Both descendant-or-self nodes (`<root>` and `<c/>`) cause `local:function(.)` to return the same `$root` element, but eXist returns 2 copies instead of deduplicating to 1.

The root cause is in `PathExpr.java:301` — `removeDuplicates()` is only called when `getLastExpression() instanceof Step`, which excludes function calls (`Function extends PathExpr`). This guard was added in 2009 to prevent over-aggressive dedup in for-loops ([bug #2880394](https://sourceforge.net/tracker/?func=detail&aid=2880394&group_id=17691&atid=117691)).

I'll file a separate issue for the path expression dedup bug. It affects any `/function-call()` path where multiple context nodes produce the same result node.

joewiz:

Will squash into a single commit.

joewiz:
Squashed - thanks for spotting the need for cleanup!

joewiz:
> The failure of `K2-Steps-31` seems significant

Fixed in https://github.com/eXist-db/exist/pull/6110

line-o (approved):
LGTM

duncdrum (approved):

PR #6110, opened 2026-03-07
Fix duplicate node elimination for function calls in path expressions

joewiz:
## Summary

- Fix the path operator `/` to correctly eliminate duplicate nodes when the right-hand step is a function call or other non-`Step` expression
- Per [XPath 3.1 §3.3.1.1](https://www.w3.org/TR/xpath-31/#id-path-operator), `/` must combine results, eliminate duplicates by node identity, and return in document order — regardless of the expression type of each step
- Fix NPE in `NodeImpl.compareTo` when sorting in-memory nodes from independent document contexts (e.g., standalone attribute or document constructors)

## Root Cause

In `PathExpr.eval()`, duplicate elimination was guarded by:

```java
if (steps.size() > 1 && getLastExpression() instanceof Step)
    result.removeDuplicates();
```

This only triggered when the last expression was a `Step` (LocationStep, RootNode, SimpleStep). Function calls (`Function extends PathExpr`), filtered expressions, and other PostfixExpr types were excluded, so paths like `$root/*/local:getroot(.)` could return duplicate nodes.

This guard was added in 2009 ([SF #2880394](https://sourceforge.net/tracker/?func=detail&aid=2880394&group_id=17691&atid=117691)) to prevent over-aggressive dedup on FLWOR results in PathExprs used as generic expression containers.

## Fix

Replace the `instanceof Step` heuristic with a `hasSlash` boolean flag that the grammar tree walker sets when it encounters `SLASH`, `DSLASH`, `ABSOLUTE_SLASH`, or `ABSOLUTE_DSLASH` tokens. This correctly distinguishes genuine path expressions from PathExprs that are merely expression containers.

Additionally:

1. **PathExpr.eval**: Guard `removeDuplicates()` with `!result.isEmpty() && Type.subTypeOf(result.getItemType(), Type.NODE)` so it is only called when the result actually contains nodes, not atomic values. This prevents unnecessary processing and avoids issues with paths like `$x/(...)/3` where the final step returns atomic values.

2. **NodeImpl.compareTo**: Handle null `document` references defensively. `DocumentImpl` nodes always have `document=null` (by design — `super(expression, null, 0)` in the constructor), so comparing two `DocumentImpl` instances from different constructor contexts would NPE at `document.docId`. This is a latent bug exposed when `removeDuplicates` sorts sequences containing nodes from independent constructor documents.

| Approach | Correctness | Performance | 2009 regression risk |
|---|---|---|---|
| Old: `getLastExpression() instanceof Step` | Misses function calls | O(1) | None (too conservative) |
| New: `hasSlash` flag + node-type guard | Correct by construction | O(1) | None (flag not set for FLWOR containers) |

## What Changed

| File | Change |
|------|--------|
| `exist-core/.../xquery/PathExpr.java` | Add `hasSlash` flag with setter; replace `instanceof Step` guard with flag check + node-type guard |
| `exist-core/.../parser/XQueryTree.g` | Call `path.setHasSlash()` at `SLASH`, `DSLASH`, `ABSOLUTE_SLASH`, `ABSOLUTE_DSLASH` |
| `exist-core/.../dom/memtree/NodeImpl.java` | Fix `compareTo` to handle null `document` field (DocumentImpl nodes) |
| `exist-core/.../xquery/PathExprDedupTest.java` | New test class: 7 tests covering dedup for function calls in paths, preservation of FLWOR duplicates, and K2-Axes-48/49 regression tests |

## Spec Reference

[XPath 3.1 §3.3.1.1 Path operator (/)](https://www.w3.org/TR/xpath-31/#id-path-operator):
> If every evaluation of E2 returns a (possibly empty) sequence of nodes, these sequences are combined, and duplicate nodes are eliminated based on node identity. The resulting node sequence is returned in document order.

## Related

- PR #6106 (fix `//` with non-LocationStep expressions — complements this fix; together they fully resolve XQTS K2-Steps-31)

## XQTS Results (W3C XQTS 3.1, prod-AxisStep)

K2-Axes-48 and K2-Axes-49 both **pass** (previously NPE).

## Test Plan

- [x] New JUnit tests pass: `PathExprDedupTest` (7 tests, including K2-Axes-48/49)
- [x] `XQueryTest` suite: 99 tests, 0 failures, 0 errors
- [x] XQTS prod-AxisStep: K2-Axes-48 ✅, K2-Axes-49 ✅

### DIFF
--- a/exist-core/src/test/java/org/exist/xquery/PathExprDedupTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/PathExprDedupTest.java
@@ -59,10 +59,10 @@ public class PathExprDedupTest {
      */
     @Test
     public void functionCallInPathDedup() throws XMLDBException {
-        final String result = query(
-                "declare variable $root := <root><c/></root>;\n" +
-                "declare function local:function($arg) { $root[$arg] };\n" +
-                "count($root/descendant-or-self::node()/local:function(.))");
+        final String result = query("""
+                declare variable $root := <root><c/></root>;
+                declare function local:function($arg) { $root[$arg] };
+                count($root/descendant-or-self::node()/local:function(.))""");
         assertEquals("1", result);
     }
 
@@ -71,10 +71,10 @@ public class PathExprDedupTest {
      */
     @Test
     public void functionReturnsConstantNodeDedup() throws XMLDBException {
-        final String result = query(
-                "declare variable $root := <root><a/><b/></root>;\n" +
-                "declare function local:getroot($x) { $root };\n" +
-                "count($root/*/local:getroot(.))");
+        final String result = query("""
+                declare variable $root := <root><a/><b/></root>;
+                declare function local:getroot($x) { $root };
+                count($root/*/local:getroot(.))""");
         assertEquals("1", result);
     }
 
@@ -84,9 +84,9 @@ public class PathExprDedupTest {
      */
     @Test
     public void forLoopPreservesDuplicates() throws XMLDBException {
-        final String result = query(
-                "declare variable $root := <root/>;\n" +
-                "count(for $x in (1, 2, 3) return $root)");
+        final String result = query("""
+                declare variable $root := <root/>;
+                count(for $x in (1, 2, 3) return $root)""");
         assertEquals("3", result);
     }
 
@@ -95,9 +95,9 @@ public class PathExprDedupTest {
      */
     @Test
     public void globalVarForLoopPreservesDuplicates() throws XMLDBException {
-        final String result = query(
-                "declare variable $data := <item/>;\n" +
-                "count(for $i in 1 to 5 return $data)");
+        final String result = query("""
+                declare variable $data := <item/>;
+                count(for $i in 1 to 5 return $data)""");
         assertEquals("5", result);
     }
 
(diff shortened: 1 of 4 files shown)


line-o:
I think this is a valuable addition to the codebase but the to failures in XQTS are alarming (`K2-Axes-48`, `K2-Axes-49`).
They both fail with a null pointer exception even though they are valid queries and quite simple ones, too.

### K2-Axes-48

```xquery
declare variable $myVar := <e/>;
$myVar/(<a/>, <b/>, <?d ?>, <!-- e-->, attribute name {}, document {()})/3
```

should return  `3 3 3 3 3 3`

### K2-Axes-49

```xquery
declare variable $myVar := <e/>;
$myVar/(<a/>, <b/>, <?d ?>, <!-- e-->, attribute name {}, document {()})/number()
```

should return `NaN NaN NaN NaN NaN`

I do know that exist-db's processor does struggle to with static items as the last path step. But introduction a new NPE is not acceptable.

joewiz:

Thanks for catching this, @line-o — you're absolutely right that introducing an NPE is not acceptable. I've pushed a fix in f342031.

**Root cause:** The `hasSlash`-based `removeDuplicates()` call was being applied to intermediate step results containing in-memory nodes from independent constructor documents. Specifically, `document {()}` creates a `DocumentImpl` whose `document` field is `null` (by design — `super(expression, null, 0)` in the constructor). When `sortInDocumentOrder()` tried to compare these nodes via `NodeImpl.compareTo()`, it hit `document.docId` without a null check.

**Two fixes:**
1. **PathExpr**: Guard `removeDuplicates()` with a node-type check so it skips atomic results entirely (paths ending in `/3` or `/number()` don't need node deduplication)
2. **NodeImpl.compareTo**: Handle null `document` defensively — this was a latent bug that could affect any code path sorting in-memory nodes from different constructor contexts

Both K2-Axes-48 and K2-Axes-49 now pass, along with all existing dedup tests and the full XQueryTest suite.

joewiz:
I can squash if the fix looks good.

exist-core/src/test/java/org/exist/xquery/PathExprDedupTest.java
```diff
@@ -0,0 +1,147 @@
...
+                "declare variable $root := <root><c/></root>;\n" +
+                "declare function local:function($arg) { $root[$arg] };\n" +
+                "count($root/descendant-or-self::node()/local:function(.))");
+        assertEquals("1", result);
+    }
+
+    /**
+     * Simpler case: child::* / function that always returns the same node.
+     */
+    @Test
```
duncdrum, line 72: I would prefer java 15 style text blocks, for better legibility throughout

duncdrum (commented):
looks good overall, one comment

joewiz:

Good call, @duncdrum — converted all XQuery test strings to Java 15 text blocks in 0ae5b26.

line-o (approved):

duncdrum (approved):

PR #6121, opened 2026-03-10
[enhancement] Add 'work/' directory to .gitignore

### LINKED ISSUE #6114: Ignore the works folder
The folder is created when running the xqts submodule

duncdrum:
- so xqts runs do not fill up the local repo with tracked modifications

close  #6114

- fix `template.bat` so it plays nice with the recently modified `.gitattributes`


line-o:
> so xqts runs do not fill up the local repo

Did you mean: "So that, after an XQTS run, the downloaded testfiles are ignored." ?

line-o:
> fix template.bat so it plays nice with the recently modified .gitattributes

What do you mean with playing nicely? What happened now in conjunction with the gitattributes?

duncdrum:
gittatributes were modified in the caching PR, so we don't store every jar twice with LF and CRLF., the CRLF bat file was modified before, but the PR wasn't rebased. So there was a template.bat file upon checkout in the list of modified files, that was impossible to get rid of without committing a renormalised version.

line-o:
@duncdrum could you just rephrase the first commit to something like

```
[ignore] Add 'work/' directory to .gitignore

Downloaded XQTS test cases are otherwise show up as untracked changes. 
```

duncdrum:
@line-o done

line-o:
@duncdrum It looks to me as there is an extra merge commit that sneaked in. Can you get rid of it?

dizzzz (changes requested):
there seems to be a mix of commits

line-o:
I will force-push the changes. The current state of develop is unbearable

PR #6126, opened 2026-03-12
Update idea run configurations

line-o:
### Description:

With this PR applied the runConfigurations match the paths when packaged so that
admin client and Jetty servers can be run from within IDEA.


dizzzz (approved):

line-o:
We really should find a way to have the path set by variables so that they will always match without us having to manually do that.

dizzzz (changes requested):
Please close, as I filed another PR for this

line-o:
This targets 6.x.x @dizzzz I will reopen this PR if we do not backport #6134

PR #6128, opened 2026-03-12
6121 redone

### LINKED ISSUE #6114: Ignore the works folder

line-o:
Same changes without the extra commit
completely redone as I was unable to checkout the original PR

supersedes #6121
closes #6114


duncdrum (approved):

dizzzz:
what about setting in .gitattributes

```
# Force Windows line endings for one specific file
path/to/yourfile.txt text eol=crlf

# Default for everything else: Linux line endings
* text eol=lf
```

and fix all for once and forever with

```
git add --renormalize .
git commit -m "Apply .gitattributes normalization"
```

dizzzz (commented):
made a suggestion

line-o:
As far as I understand this has already happened. The bat file was initially committed with LF and is now normalized in all our branches which makes rebasing impossible

duncdrum:
yes the `.gitattributes` are correct and identifying the file as `CRLF`, the problem is that because we didn't rebase and merge, the file never went through git's normalisation routine. 

the necessary step is `git add --renormalize .` which will change CRLF into the kind of CRLF git history understands. That's what I did in my PR.

line-o:
And is done here. I didn't even have to do this. The file just converts whenever I work with my IDE. This is why it is such a problem for me.

duncdrum:
@line-o not just for you

PR #6132, opened 2026-03-13
[fix] re-allow switch on xs:anyURI

### LINKED ISSUE #6131: [BUG] switch on xs:anyURI matches xs:string
**Describe the bug**

xs:anyURI can be promoted to xs:string for comparisons.
While this works in the latest released version 6.4.1 and current develop HEAD

```xquery
      xs:anyURI("test") eq "test"
```

This expression only returns true in 6.4.1 and false on develop.  

```xquery
switch(xs:anyURI("test"))
        case "test" return true()
        default return false()
```

That is a regression that needs to be fixed. Both Saxon and BaseX do handle that in the same way - for XQuery 3.1 and later.

**Expected behavior**

The xs:anyURI in a switch to be matched against string literals.

**To Reproduce**

```Xquery
xquery version "3.1";

declare
    %test:assertTrue
function t:test() {
    switch(xs:anyURI("test"))
        case "test"return true()
        default return false()
};

declare
    %test:assertTrue
function t:test() {
      xs:anyURI("test") eq "test"
};
```

Build: eXist-7.0.0-SNAPSHOT (d18dbfd9592763f49a0074fe53c550be7295c04f)
Java: 21.0.10 (Eclipse Adoptium)
OS: Linux 6.12.54-linuxkit (aarch64)

**Additional context**
- 
- How is eXist-db installed? Docker image existdb/existdb:latest

line-o:
fixes #6131

Added some XQSuite tests to ensure functionality.

### DIFF
--- a/exist-core/src/test/xquery/xquery3/switch.xqm
+++ b/exist-core/src/test/xquery/xquery3/switch.xqm
@@ -1,0 +1,78 @@
+(:
+ : eXist-db Open Source Native XML Database
+ : Copyright (C) 2001 The eXist-db Authors
+ :
+ : info@exist-db.org
+ : http://www.exist-db.org
+ :
+ : This library is free software; you can redistribute it and/or
+ : modify it under the terms of the GNU Lesser General Public
+ : License as published by the Free Software Foundation; either
+ : version 2.1 of the License, or (at your option) any later version.
+ :
+ : This library is distributed in the hope that it will be useful,
+ : but WITHOUT ANY WARRANTY; without even the implied warranty of
+ : MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ : Lesser General Public License for more details.
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+(:~
+ : ensure that different types can be matched to string literals in switch-statements
+ : @see https://github.com/eXist-db/exist/issues/6131
+ :)
+module namespace tsw="http://exist-db.org/xquery/test/switch";
+
+declare namespace test="http://exist-db.org/xquery/xqsuite";
+
+declare
+    %test:assertTrue
+function tsw:element() {
+    switch(<node>test</node>)
+        case "test" return true()
+        default return false()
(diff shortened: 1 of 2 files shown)


duncdrum (approved):

line-o:
With this PR applied an additional 16 XQTS cases pass

```       
<testcase name="fn-subsequence-mix-args-005"/>
<testcase name="fn-insert-before-mix-args-005"/>
<testcase name="constSeq-16"/>
<testcase name="op-concatenate-mix-args-005"/>
<testcase name="op-concatenate-mix-args-006"/>
<testcase name="op-concatenate-mix-args-007"/>
<testcase name="fn-remove-mix-args-006"/>
<testcase name="fn-remove-mix-args-005"/>
<testcase name="fn-remove-mix-args-007"/>
<testcase name="fn-transform-14"/>
<testcase name="analyzeString-028"/>
<testcase name="ns-queries-results-q1"/>
<testcase name="fo-test-fn-namespace-uri-for-prefix-002"/>
<testcase name="fo-test-fn-namespace-uri-for-prefix-003"/>
<testcase name="fo-test-fn-namespace-uri-for-prefix-004"/>
<testcase name="fo-test-fn-namespace-uri-for-prefix-005"/>
<testcase name="fo-test-fn-namespace-uri-for-prefix-001"/>
```

line-o:
Thanks @wolfgangmm for reporting this bug!

exist-core/src/test/xquery/xquery3/switch.xqm
```diff
@@ -0,0 +1,78 @@
...
+
+(:~
+ : ensure that different types can be matched to string literals in switch-statements
+ : @see https://github.com/eXist-db/exist/issues/6131
+ :)
+module namespace tsw="http://exist-db.org/xquery/test/switch";
+
+declare namespace test="http://exist-db.org/xquery/xqsuite";
+
+declare
```
dizzzz, line 32: It might be nice to have some tests where 'false' will be returned?

line-o: I understand where you are coming from but the tests are _not_ about checking switch statement in general.
I wanted to check for the regression found by @wolfgangmm today
If you really think there is a test missing I should add please tell me exactly what should be added.

dizzzz: I don't really have an answer yet. My point is that the results are all 'true', so in evil thought it would be possible if false() could never return?

PR #6134, opened 2026-03-14
[ci]  Make sure that upon version bump no paths need to be changed

### LINKED ISSUE #6127: Find a way to set the target path in IDEA runConfigurations automatically
At the moment we have to change the runConfigurations manually every time we change a version number in any develop branch.

see #6126 

The runConfigurations _do_ support placeholders. Maybe this can be done for the target path as well.

Or, alternatively, we should output the packages /artifacts into a static path.

```
exist-distribution/target/exist-distribution-7.0.0-SNAPSHOT-dir
```
Do we really need to have the _directory_ to include the current version number? 

```
exist-distribution/target/exist-distribution-dir
```
Could be an alternative path. 
I believe the actual package _should_ still have the version in its name.

dizzzz:
Make sure that upon version bump no paths need to be changed in a bunch of files.

fixes https://github.com/eXist-db/exist/issues/6127 
supersedes https://github.com/eXist-db/exist/pull/6126

### DIFF
--- a/.idea/runConfigurations/Java_Admin_Client.xml
+++ b/.idea/runConfigurations/Java_Admin_Client.xml
@@ -5,7 +5,7 @@
     <option name="MAIN_CLASS_NAME" value="org.exist.start.Main" />
     <module name="exist-distribution" />
     <option name="PROGRAM_PARAMETERS" value="client" />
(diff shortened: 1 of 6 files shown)


line-o:
I am all for this. 👌
-- snip --

line-o:
The GHA workflows might need changes. It turns out it's the exist-docker module

dizzzz:
@line-o oops missed that part. fixed.

line-o (approved):
Excellent. Small change in build tooling, big change for us developers!
We could back port this even.

duncdrum (approved):

PR #6140, opened 2026-03-15
Implement fn:invisible-xml() using Markup Blitz

joewiz:
## Summary

- Implements `fn:invisible-xml($grammar, $options?)` per the [XQuery 4.0 spec](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-invisible-xml)
- Compiles an Invisible XML grammar and returns a function that parses input strings into XML documents
- Uses the [Markup Blitz](https://github.com/GuntherRademacher/markup-blitz) library (204 KB, zero transitive deps, Apache 2.0, 100% ixml spec compliance)

## What Changed

| File | Change |
|------|--------|
| `exist-parent/pom.xml` | Add `markup-blitz:1.10` to `<dependencyManagement>` |
| `exist-core/pom.xml` | Add `markup-blitz` dependency |
| `ErrorCodes.java` | Add `FOIX0001` (invalid grammar) and `FOIX0002` (parse error) |
| `FnInvisibleXml.java` | New function implementation |
| `FnModule.java` | Register 1-arity and 2-arity signatures |
| `fnInvisibleXml.xqm` | 7 XQSuite tests |

## Features

- **String grammars**: ixml notation compiled via `Blitz.generate()`
- **XML grammars**: Element node grammars compiled via `Blitz.generateFromXml()`
- **Empty sequence**: Uses default ixml grammar via `Blitz.ixmlGrammar()`
- **`fail-on-error` option**: When `false` (default per spec), parse failures return XML with `ixml:state="failed"` attribute; when `true`, raises `FOIX0002`
- **Reusable parser**: Grammar is compiled once; returned function can parse multiple inputs

## Note: `Blitz.generateFromXml()` thread-safety

`Blitz.generateFromXml()` is not thread-safe — concurrent calls fail with "Failed to parse XML gramamr input" (~50% failure rate). This was confirmed with a standalone test outside eXist-db. `Blitz.generate()` (string grammars) is not affected. The implementation wraps `generateFromXml()` calls in a `synchronized` block as a workaround.

## XQTS Results

| Test Set | Total | Pass | Fail | Rate |
|----------|-------|------|------|------|
| fn-invisible-xml | 12 | 12 | 0 | 100% |

## Test Plan

- [x] XQSuite: 7 tests (date parsing, reuse, error handling, XML grammar, complex XML grammar)
- [x] XQTS QT4: 12/12 on fn-invisible-xml test set
- [x] Full XQSuite regression: 1139 tests, 0 failures

## Credits

Markup Blitz integration pattern informed by BaseX's implementation.


exist-core/src/main/java/org/exist/util/Collations.java
```diff
@@ -346,7 +346,24 @@ public static boolean equals(@Nullable final Collator collator, final String s1,
      */
     public static int compare(@Nullable final Collator collator, final String s1,final  String s2) {
         if (collator == null) {
-            return s1 == null ? (s2 == null ? 0 : -1) : s1.compareTo(s2);
+            if (s1 == null) {
+                return s2 == null ? 0 : -1;
+            }
+            // Compare by Unicode codepoints, not UTF-16 code units.
```
dizzzz, line 352: could this give a performance penalty?

would be than an option  to have a separate PR. lso,,, UTF-16 is not so common, let alone I am not sure if exist encodes all in UTF-8 internally.

joewiz:

Closing this PR — the fn:invisible-xml() implementation has been squashed into the `feature/xquery-4.0-parser` branch (commit 2c9572fa44), which already contains all the XQuery 4.0 functions work. This avoids an extra branch in the dependency tree.

PR #6144, opened 2026-03-16
Upgrade Jetty 11.0.25 to 12.0.16 (EE10)

joewiz:
## Summary
Upgrades Jetty from 11.0.25 (EOL since Jan 2025) to 12.0.16 with EE10 (Jakarta EE 10, Servlet 6.0).

## Commits

| # | Commit | Scope |
|---|--------|-------|
| 1 | **Dependencies** | Jetty 12, Servlet 6.0, ee10 artifacts, Milton servlet fix |
| 2 | **XML config** | Handler model, deployer, security, compliance, class paths |
| 3 | **Java code** | JettyStart, WebAppContext, Servlet 6.0 API changes |
| 4 | **Test updates** | web.xml schemas, XSD files, @Ignore HTTP/0.9+1.0, test API |

### Jetty 12
- Servlet artifacts → `org.eclipse.jetty.ee10` namespace; JAAS → `jetty-security`
- `HandlerCollection` → `Handler$Sequence`; `PathResource` → `ResourceFactory`
- Inline `addHandler` deployment replaces `ContextProvider` scanner
- `UriCompliance` → `UNSAFE` for non-ASCII path support
- `webdefault.xml`: `DefaultServlet`/`IntrospectorCleaner` → ee10 packages

### Servlet 6.0
- Added: `getServletConnection()`, `getRequestId()`, `getProtocolRequestId()`
- Removed: `setStatus(int, String)`, `isRequestedSessionIdFromUrl()`, `getRealPath(String)`
- All `web.xml` descriptors → `version="6.0"`; distribution XSDs → Jakarta EE 10

### Milton servlet compatibility
Milton 1.8.1.3-jakarta5 calls `HttpServletResponse.setStatus(int, String)` which was removed in Servlet 6.0. This PR temporarily uses the `com.evolvedbinary.thirdparty.com.ettrema:milton-servlet:1.8.1.3-jakarta-ee10` variant which fixes this one-line call. 

## Test Results

| Module | Tests | Result |
|--------|-------|--------|
| exist-core | ~6490 | All pass (4 HTTP/0.9+1.0 @Ignored) |
| RestXQ | 20 | All pass |
| File Module | 63 | All pass |
| Persistent Login | — | All pass |
| WebDAV | 13 | All pass |

## CI Status

All checks green — unit tests, integration tests (macOS/ubuntu/windows), XQTS, container image, license check, Codacy.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

closes #5462

Issue #5827, opened 2025-08-04
[BUG] Revise log files


dizzzz:
There are a few items that should be addressed:
- Jetty (now using slf4j) should not write into exist.log
- All tests seem to have a full blown log file.

line-o:
@dizzzz I think at least the first item - Jetty logs - is already addressed. Could you expand what is meant with the second one?

duncdrum:
@dizzzz I m not sure I understand the issue about the tests, can you elaborate?

joewiz:
[This response was co-authored with Claude Code. -Joe]

Quick clarification from a digression on #6145, where @duncdrum flagged that `data/0000000000.log` grows continuously while the DB is idle.

That file is *not* a logger output — it's the storage engine's write-ahead transaction journal, written by [`org.exist.storage.journal.Journal`](https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/storage/journal/Journal.java). Filenames follow `%010x.log` (see `Journal.getFileName` around line 780), with `0000000000.log`, `0000000001.log`, ... rolling in sequence as the buffer fills or checkpoints land. It is governed by `conf.xml` recovery settings (`recovery.enabled`, `recovery.sync-on-commit`, journal size threshold), not by `log4j2.xml`.

So when revising "log files" under this issue, I'd suggest scoping the journal separately — its lifecycle (durability, checkpoint cadence, rotation) is a storage-engine concern rather than a logging-config concern, and the right fix for "grows without bound on idle" probably lives in checkpoint scheduling or an idle-checkpoint trigger.

Happy to open a dedicated issue if that helps keep this one focused on the slf4j/jetty/test-log items @dizzzz originally raised.

PR #6145, opened 2026-03-17
Upgrade Jetty 11→12 (EE10, Jakarta Servlet 6.0) and add WebSocket module with streaming query evaluation

### LINKED ISSUE #5462: [feature] update to jetty12
This would bring us (at least) a better cookie implementation with jakarta.servlet-api version 6

joewiz:
## Summary

This PR does two things:

1. **Jetty 11→12 upgrade**: Upgrades from Jetty 11.0.25 to 12.0.16 with EE10 environment (Jakarta Servlet 6.0), including the full `javax.servlet` → `jakarta.servlet` migration.

2. **WebSocket infrastructure**: Adds two WebSocket endpoints to exist-core:
   - **`/ws`** — Channel-based pub/sub messaging (ported from monex) with `_monitor` query lifecycle broadcasting
   - **`/ws/eval`** — Real-time streaming XQuery evaluation with cancellation, progress reporting, and timing breakdown

These were originally two separate PRs (#6213 for Jetty 12, this PR for WebSockets). The community call on 2026-04-13 agreed to consolidate them here since the WebSocket work depends on Jetty 12's Jakarta WebSocket API.

## Jetty 12 Upgrade

- Jetty 11 → 12 with EE10 (Jakarta Servlet 6.0)
- `javax.servlet` → `jakarta.servlet` across all modules
- `e.printStackTrace()` → proper `logger.fatal/error` in JettyStart
- Removed unhelpful backwards-compatibility comments (per reviewer request from @reinhapa and @dizzzz)
- `setStatus(int, String)` reason parameter dropped (Servlet 6.0 removes it; `sendError` not safe for RESTXQ — would commit response and prevent custom body serialization)
- Portal webapp at `/` restored — direct WebAppContext handler (the Jetty 11 DeploymentManager/ContextProvider pattern doesn't work in Jetty 12 EE10)
- Supersedes PR #6144 and closed PR #6213 — incorporates all reviewer feedback
- closes #5462 

## `/ws` — Channel Pub/Sub

Ported from monex's `RemoteConsoleEndpoint` into a general-purpose WebSocket module.

```
XQuery ws:send("channel", $data)
    → WebSocketModule → WebSocketAdapter → WebSocketEndpoint → Browser
```

Clients connect, subscribe to channels via `{"channel": "name"}`, and receive JSON messages. A 500ms heartbeat ping keeps connections alive through proxies.

### `_monitor` Channel (new)

Admin clients can subscribe to the `_monitor` channel to receive real-time query lifecycle events:

- **Event-driven**: `started`, `progress`, `completed`, `error`, `cancelled` events from `/ws/eval` queries
- **Periodic snapshots**: All running queries (including REST/XQueryServlet) via ProcessMonitor, broadcast every 1 second

### XQuery Functions

| Function | Signature | Purpose |
|----------|-----------|---------| 
| `ws:log` | `($items as item()*) as empty-sequence()` | Log to "default" channel |
| `ws:log` | `($channel, $items)` | Log to specific channel |
| `ws:send` | `($channel, $items)` | Send JSON to channel |
| `ws:broadcast` | `($items as item()*)` | Send to ALL clients |
| `ws:channel-count` | `($channel) as xs:integer` | Count subscribers |

## `/ws/eval` — Streaming Query Evaluation (new)

A WebSocket endpoint for real-time XQuery evaluation — the XQuery equivalent of a modern database's interactive query console protocol.

### Protocol (JSON over WebSocket)

**Client → Server:**

| Action | Purpose |
|--------|---------|
| `eval` | Execute query with optional streaming, variables, serialization, timeout |
| `cancel` | Cancel a running query by ID |
| `compile` | Compile-check query without execution (syntax validation) |
| `admin-cancel` | DBA-only: kill any running query via ProcessMonitor |

**Server → Client:**

| Type | Purpose |
|------|---------|
| `progress` | Phase updates: parsing → compiling → evaluating → serializing |
| `result` | Serialized data chunks with `more`/done indicators and timing |
| `error` | XPath error with code, message, line, column |
| `cancelled` | Confirmation with items-produced count |
| `compile` | Success/failure with diagnostics array |

### Key Features

- **Streaming**: Configurable chunk size (default 1000 items). Large results stream incrementally — the client sees data as it's produced.
- **Cancellation**: `cancel` action sets a flag on `XQueryWatchDog`. Works mid-evaluation and mid-serialization.
- **Timing breakdown**: Parse, compile, evaluate, serialize phases measured independently.
- **Authentication**: Basic auth on WebSocket handshake, falls back to guest.
- **Admin-cancel**: DBA users can kill any running query (same as `system:kill-running-xquery`).
- **Progress reporting**: Phase transitions broadcast to both the eval client and `_monitor` subscribers.

### Example

```javascript
const ws = new WebSocket('ws://localhost:8080/exist/ws/eval');

ws.send(JSON.stringify({
    action: 'eval',
    id: 'q-1',
    query: 'for $i in 1 to 1000000 return $i',
    "chunk-size": 1000,
    serialization: { method: 'adaptive' }
}));

ws.onmessage = (event) => {
    const msg = JSON.parse(event.data);
    if (msg.type === 'result') appendToOutput(msg.data);
    if (msg.type === 'progress') updateStatusBar(msg.phase);
    if (!msg.more) showTiming(msg.timing);
};

// Cancel
ws.send(JSON.stringify({ action: 'cancel', id: 'q-1' }));
```

## New Classes

| Class | Package | Purpose |
|-------|---------|---------|
| `WebSocketEndpoint` | `o.e.xquery.functions.websocket` | `@ServerEndpoint("/ws")` — channel pub/sub, heartbeat |
| `WebSocketModule` | `o.e.xquery.functions.websocket` | XQuery module (`ws:` namespace) |
| `ConsoleCompatModule` | `o.e.xquery.functions.websocket` | Backward-compatible `console:` namespace |
| `EvalWebSocketEndpoint` | `o.e.http.ws` | `@ServerEndpoint("/ws/eval")` — streaming eval |
| `EvalSession` | `o.e.http.ws` | Per-connection state, query cancellation |
| `QueryExecutor` | `o.e.http.ws` | XQuery execution with streaming, progress, timing |
| `EvalProtocol` | `o.e.http.ws` | JSON message parsing and generation |
| `QueryMonitorBroadcaster` | `o.e.http.ws` | Bridges eval lifecycle → `_monitor` channel |

## Backward Compatibility

The `console:` namespace (`http://exist-db.org/xquery/console`) is provided by `ConsoleCompatModule`, which delegates to `WebSocketModule`. Existing monex XQuery code using `console:log()` works unchanged.

## Test plan

**Jetty 12:**
- [x] All 50 modules compile with Jakarta Servlet 6.0
- [x] HTTP endpoints respond correctly (verified on next-v2 Docker)
- [x] Portal at / serves redirect to /exist

**WebSocket pub/sub (3 tests):**
- [x] Connect and receive heartbeat ping
- [x] Subscribe to channel and receive message
- [x] Channel count reflects subscribers

**Eval endpoint (20 tests):**
- [x] Simple eval (`1 + 1` → `"2"`)
- [x] Variables (external variable binding with type casting)
- [x] Compile error (syntax error returns error message)
- [x] Compile action (success and failure with diagnostics)
- [x] Streaming results (500 items, chunk-size 100 → 5 chunks)
- [x] Large result streaming (100K items, chunk-size 1000 → 100 chunks)
- [x] Cancellation (long query cancelled mid-flight)
- [x] Rapid cancel (cancel before compilation completes)
- [x] Max execution time (timeout enforcement via watchdog)
- [x] Timing (result includes parse/compile/eval/serialize breakdown)
- [x] Progress reporting (phase messages during execution)
- [x] Serialization options (adaptive method)
- [x] Binary result handling (xs:base64Binary)
- [x] Map/array serialization (adaptive output)
- [x] Module-load-path resolution (import module from /db)
- [x] Admin-cancel permission check (non-DBA gets denied)
- [x] Monitor channel (\_monitor receives query lifecycle events)
- [x] Connection cleanup (disconnect mid-query releases resources)
- [x] Error recovery (invalid JSON doesn't kill connection)
- [x] Invalid message / missing query handling

**Total: 23 WebSocket integration tests + Jetty 12 migration**

## What This Enables

- **eXide**: Stream results live, cancel runaway queries, see timing breakdown
- **Sandbox**: Notebook cells with progress bars and incremental output
- **monex**: Live query dashboard via `_monitor` channel
- **VS Code**: "Run Query" with streaming results and cancel button
- **xst CLI**: `xst eval --stream` prints results as they arrive


dizzzz:
shall be pulled in after #6144 as there is overlap? That makes reviewing a bit more simple.

joewiz:
@dizzzz Yes, you're right, this PR depends on Jetty 12.

dizzzz (approved):

exist-core/src/main/java/org/exist/http/ws/EvalSession.java
```diff
@@ -0,0 +1,86 @@
...
+    public void unregisterQuery(final String queryId) {
+        runningQueries.remove(queryId);
+    }
+
+    /**
+     * Cancel a running query by its ID.
+     *
+     * @return true if the query was found and cancelled
+     */
+    public boolean cancelQuery(final String queryId) {
```
duncdrum, line 68: I m a bit confused about the cancel endpoint. The dba version makes sense. But are we going to have a cancel endpoint where `guest` can cancel queries? Is that advisable? What is the intended use for that endpoint?

exist-core/src/main/java/org/exist/http/ws/EvalWebSocketEndpoint.java
```diff
@@ -0,0 +1,322 @@
...
+                    } catch (final AuthenticationException e) {
+                        LOG.warn("WebSocket eval authentication failed: {}", e.getMessage());
+                    } catch (final EXistException e) {
+                        LOG.error("Failed to get BrokerPool for WebSocket auth: {}", e.getMessage());
+                    }
+                }
+            }
+
+            // Fall back to guest if no credentials provided
+            if (subject == null) {
```
duncdrum, line 114: S.a. I don't even need to provide `guest`?

exist-core/src/main/java/org/exist/http/ws/EvalWebSocketEndpoint.java
```diff
@@ -0,0 +1,322 @@
...
+            if (subject != null) {
+                sec.getUserProperties().put(USER_PROPERTY, subject);
+            }
+        }
+    }
+
+    @OnOpen
+    public void onOpen(final Session session, final EndpointConfig config) {
+        session.setMaxIdleTimeout(0); // no idle timeout for eval sessions
+        session.setMaxTextMessageBufferSize(10 * 1024 * 1024); // 10MB for large queries
```
duncdrum, line 132: I have the feeling we might want to log when this buffer size is reached. In either debug or trace?

exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -603,7 +603,7 @@ public void run() {
                             // make sure to stop the timer thread!
                             timer.cancel();
                         } catch (final Exception e) {
-                            e.printStackTrace();
+                            logger.error("An error occurred in the shutdown scheduler: {}", e.getMessage(), e);
```
duncdrum, line 606: Why remove the stacktrace here?

exist-jetty-config/src/main/resources/org/exist/jetty/etc/jetty-logging.xml
duncdrum, line 1: Where do we want the jetty logging to go in the new hotness? This looks like it's sticking close to where things were before, no?

extensions/webdav/pom.xml
```diff
@@ -84,7 +84,7 @@
         </dependency>
 
         <dependency>
-            <groupId>org.exist-db.thirdparty.com.ettrema</groupId>
+            <groupId>com.evolvedbinary.thirdparty.com.ettrema</groupId>
```
duncdrum, line 87: Huh? Why the change in namespace?

duncdrum (commented):
Good stuff. `cancel` sounds dangerous but I might be misunderstanding how it works. 

Curious about logging. Should we tie jetties logging closer into exists? Do we need a separate ws log? Do we want to stick with `request.log`

@joewiz did you check the docs? I have the feeling some pages might need an update?

exist-core/src/main/java/org/exist/http/ws/EvalSession.java
line-o: Cancelling a query is - at least that is my understanding - a necessary tool to end requests server-side. So, no `guest` should never be able to call it. But it might happen automatically for requests initiated by guest.

extensions/webdav/pom.xml
line-o: I believe that evolved binary has a newer version of the ettrema library.
If you want to check this was discussed in one of the previous PRs this one supersedes.

duncdrum: https://github.com/eXist-db/exist/pull/6144#issuecomment-4138524377

exist-core/src/main/java/org/exist/jetty/JettyStart.java
reinhapa: @duncdrum seems to be logged, see second parameter.

exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -644,6 +644,7 @@ public synchronized boolean isStarted() {
             try {
                 wait();
             } catch (final InterruptedException e) {
+                // nop
```
reinhapa: Should a catched `InterruptedException` not trigger a `Thread.currentThread().interrupt()` ?

joewiz:

Thanks for catching these, @reinhapa. Fixed both `InterruptedException` catches in `JettyStart` to call `Thread.currentThread().interrupt()` — the interrupt status was being silently swallowed.

Pushed in 586452506e.

exist-jetty-config/src/main/resources/org/exist/jetty/etc/jetty-logging.xml
line-o: I would assume that is a good thing if there is not a better alternative.

dizzzz:
DW: rebased with develop

exist-jetty-config/src/main/resources/org/exist/jetty/etc/jetty-logging.xml
dizzzz: the file is now called jetty.log, which is different than before.
I'd not mind if the jetty log ends up in a jetty specific sub directory

dizzzz:
todo~: favicon is now jetty's, we have one for eXistDB

image

joewiz:
@dizzzz I had replaced Jetty's favicon with eXist's locally but somehow didn't commit it. Fix on the way.

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/adapters/HttpServletResponseAdapter.java
```diff
@@ -51,7 +51,9 @@ public void setHeader(final String name, final String value) {
 
     @Override
     public void setStatus(final HttpStatus status, final String reason) {
```
reinhapa, line 53: As we now ignore the `reason`, we should check the actual invocation point within our code and adjust this as well. Maybe we log the reason if any instead, or log it within this method...

reinhapa (changes requested):

extensions/exquery/restxq/src/main/java/org/exist/extensions/exquery/restxq/impl/adapters/HttpServletResponseAdapter.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Good point. The Servlet 6.0 spec dropped the reason parameter from `sendError(int, String)` — the string is now treated as the error page body, not the HTTP reason phrase. But callers throughout our codebase still pass meaningful reason strings.

I've updated `HttpServletResponseAdapter.setStatus()` to log the reason at DEBUG level when one is provided, so the information isn't lost silently:

```java
if (reason != null && !reason.isEmpty()) {
    LOG.debug("HTTP {} — reason (not sent per Servlet 6.0): {}", status.getStatus(), reason);
}
```

This preserves diagnostic information without trying to set a reason phrase that Jetty 12 would ignore. Commit: 0865a967d0 on next-v3.

exist-core/src/main/java/org/exist/http/ws/EvalSession.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Great catch — you're right that `guest` should not be able to cancel queries. As @line-o noted, cancellation is a server-side resource management tool, not something anonymous users should invoke.

The current design has two cancel paths:
- **DBA cancel** (`/ws/eval-dba`): Authenticated DBA can cancel any running query by ID. This is the admin tool.
- **Session cancel** (`/ws/eval`): A connected user can cancel their own running query. The cancel applies only to queries started on that same WebSocket session.

The session cancel is safe because it's scoped to the caller's own queries — you can only cancel what you started. But the default `guest` fallback in `EvalWebSocketEndpoint` should be removed. If no authentication is provided, the connection should be rejected rather than falling back to guest. Will update.

exist-jetty-config/src/main/resources/org/exist/jetty/etc/jetty-logging.xml
joewiz: [This response was co-authored with Claude Code. -Joe]

Yes, keeping it close to the existing location is intentional. The file is now `jetty.log` (vs the old name) but still in eXist's log directory. @dizzzz suggested a jetty-specific subdirectory which could be cleaner — happy to move it to `logs/jetty/` if there's consensus. The main thing is that it's separate from eXist's own log4j logs so they don't interleave.

extensions/webdav/pom.xml
joewiz: [This response was co-authored with Claude Code. -Joe]

This is the `javax.servlet` → `jakarta.servlet` migration that comes with Jetty 12 (EE10). The WebDAV module's servlet dependencies moved from `javax.servlet:javax.servlet-api` to `jakarta.servlet:jakarta.servlet-api`. As @line-o noted, this was discussed in PR #6144 — it's a mechanical consequence of the Jakarta EE 10 transition.

exist-core/src/main/java/org/exist/http/ws/EvalWebSocketEndpoint.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Correct — the guest fallback has been removed. If no valid credentials are provided in the WebSocket handshake, the connection is now rejected with a `VIOLATED_POLICY` close reason. Authentication is required for all WebSocket eval sessions.

exist-core/src/main/java/org/exist/http/ws/EvalWebSocketEndpoint.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Good idea. I've extracted the buffer size to a `MAX_TEXT_MESSAGE_SIZE` constant and added specific logging in `onError` that detects when a message exceeds the 10MB limit:

```java
if (error.getMessage() != null && error.getMessage().contains("Text message size")) {
    LOG.warn("WebSocket message exceeds {}MB buffer limit: {}", MAX_TEXT_MESSAGE_SIZE / (1024 * 1024), error.getMessage());
}
```

This way oversized messages get a clear warning rather than a generic error. Commit: a9ada646b4 on next-v3.

reinhapa (approved):

exist-core/src/main/java/org/exist/http/ws/EvalWebSocketEndpoint.java
```diff
@@ -0,0 +1,322 @@
...
+            default:
+                try {
+                    session.getBasicRemote().sendText(
+                            EvalProtocol.errorMessage(msg.id, null,
+                                    "Unknown action: " + msg.action, 0, 0, null));
+                } catch (final IOException e) {
+                    LOG.debug("Failed to send unknown action error: {}", e.getMessage());
+                }
+                break;
+        }
```
reinhapa: Maybe turn this into new switch-as-expression style as this would be more compact.

exist-core/src/test/java/org/exist/http/ws/EvalWebSocketEndpointTest.java
```diff
@@ -0,0 +1,1288 @@
...
+                        result.put(field, "OBJECT");
+                        parser.skipChildren();
+                        break;
+                    case START_ARRAY:
+                        result.put(field, "ARRAY");
+                        parser.skipChildren();
+                        break;
+                    default:
+                        break;
+                }
```
reinhapa, line 176: Maybe turn this into new switch-as-expression style as this would be more compact.

joewiz:

Thanks @reinhapa — addressed in e749728b28:
- Converted switch statement in `EvalWebSocketEndpoint.onMessage()` to arrow-style switch
- Converted switch statement in `EvalWebSocketEndpointTest.parseJson()` to arrow-style switch

joewiz:

@dizzzz Thanks for the thorough review! All five fixes applied in 93bc4e87aa:

1. **NPE guard in HttpServletRequestWrapper** — null check for `getContentType()` in both the constructor and `toString()`
2. **Status code propagation in flushError** — `response.setStatus(wrappedResponse.getStatus())` before writing error data
3. **EvalWebSocketEndpoint resource leak** — `shutdown()` + lazy `getExecutorService()` pattern for the query executor
4. **WebSocketEndpoint resource leak** — `shutdown()` for heartbeat and monitor `ScheduledExecutorService` instances
5. **Lifecycle integration** — both shutdowns called from `JettyStart.lifeCycleStopped()`
6. **sendAll made static** — removes per-broadcast instantiation

Good catches, especially the `flushError` status propagation — that would have caused error responses to return 200.

duncdrum (changes requested):
before we merge this, I would like to reconsider how we handle logging in `jetty-logging.xml` and (more pressing to me) `jetty-requestlog.xml` 

the config files live in `etc/jetty/...` so having `logs/jetty/...` makes it nicely symmetrical. 

ultimately I would prefer if `log4j2.xml` becomes the only file required for configuring logging for an instance, we already have some pipes from jetty, but to properly consolidate requires some more thought and work. I've added the item to the community call agenda. 

We also need to update the docs for the new jetty hotness.

joewiz:
From yesterday's call, we decided to pursue:

* All logging should be centrally defined from log4j2.xml
* Jetty logs should go into the logs/jetty subfolder
* There should be an option to write all logs to stdout, which will be enabled by default in containers
* A stretch goal would be to offer a configuration that suppresses request logs entirely - with particular benefit to GDPR applications

I'll seek Claude's assistance with adding this to the PR.

duncdrum (changes requested):
off to a great start. 

### From CI

1. there is a test failure
```Error: core] [ERROR] Failures: 
Error: core] [ERROR]   WebSocketEndpointTest.channelCountReflectsSubscribers:163 One subscriber expected:<1> but was:<0>
[exist-core] [INFO] 
Error: core] [ERROR] Tests run: 6638, Failures: 1, Errors: 0, Skipped: 110
```

2. the docker deploy step is failing. We'll need to adjust the deployment check in the bats test, and also the copy logs on failed run steps, with the new logging that won't work anymore 

### Testing on `next-v3`

1. I think we should be GDPR friendly by default. To encourage privacy friendly practices and not risk users inadvertently getting into trouble.
   -  Even if there is no consensus on the above, we should be GDPR friendly in the container, as the `healthcheck` command will just pollute the logs more than necessary. If we could filter the health check requests from the request logs that would be an option, but not logging seems preferable to me. 

### not jetty but logging related
1.  `0000000.log` in `exist/data` 

This is concerning, it is continuously written to, without the db doing anything, and seems to grow without limit. After a few minutes of idling it's already larger than anything else in the data folder. 

as a log file, it really should work like the other logs. I couldn't inspect the contents of the file yet, as it's locked. Requires more digging, and probably a more involved fix. 

Image

joewiz:

Thanks @duncdrum. Addressed the test failure in c5afd8d9ff.

**Root cause:** `WebSocketEndpointTest.channelCountReflectsSubscribers` asserted the server-side channel count 100 ms after the client's `CountDownLatch` fired. The latch counts down once `session.getBasicRemote().sendText()` returns on the client, which only guarantees the subscription message was buffered/sent — it does not wait for the server's `@OnMessage` handler (running on a different thread) to update the static `sessions` map. On a loaded CI runner, 100 ms was insufficient, producing the `One subscriber expected:<1> but was:<0>` failure.

**Fix:** replaced the fixed `Thread.sleep(100)` with a polling helper `awaitChannelCount(channel, expected, timeoutMillis)` that polls the channel count every 25 ms up to a 2 s deadline. Applied the same poll on cleanup so a leaked session cannot affect subsequent tests sharing the static `sessions` map.

Verified locally with 3 consecutive runs of `WebSocketEndpointTest` — all green.

Will follow up on the docker deploy / bats / log-copy changes and the request-log GDPR / `0000000.log` items in separate commits.

duncdrum:
on the `00000.log` I compared `:next-v3` to  `:latest` on the latter after ~1h of idling light usage, the file is still at ~10MB, on `:next-v3` its at 60MB and growing within minutes.  So I think it's good that we are catching this now, but I doubt it's directly related to this PR alone.

duncdrum:
see #5827

joewiz:

Thanks for the careful review @duncdrum. Pushed two follow-ups (c5afd8d9ff and 6e37bf8709) addressing your April 29 / May 2 items:

**Done in this PR**

1. **Flaky `WebSocketEndpointTest.channelCountReflectsSubscribers`** — the test was checking the server-side channel count 100 ms after the client's `CountDownLatch` fired, but `session.getBasicRemote().sendText()` only guarantees the message is buffered, not that the server's `@OnMessage` handler has run. Replaced the fixed sleep with a polling helper (up to 2 s) and applied the same poll on cleanup. 3 consecutive local runs pass.
2. **CI "Copy logs on failure" step** — the `docker cp .../logs/exist.log` line couldn't work after the log4j2 consolidation (containers no longer write on-disk rolling files). Replaced with `docker logs exist-ci > exist.log 2>&1` so the artifact still captures the failure context.
3. **Bats test 6 ("logs contain repo.log output")** — root cause was the container console pattern using `%logger{36}` (logger name) instead of `%F` (file name). Switched the container pattern to match the standard `log4j2.xml` (`%d{DATE} [%t] %-5p (%F [%M]:%L) - %m %n`), so `Deployment.java` shows up in `docker logs` again. All 6 connect-spec bats tests pass locally against the rebuilt image.
4. **GDPR-friendly default in the container** — agreed that not logging is preferable to filtering. Defaulted `EXIST_REQUEST_LOG_LEVEL=off` in `log4j2-container.xml` only; bare-metal installs still default to `info`. Operators who want access logs in containers can opt back in via `docker run -e EXIST_REQUEST_LOG_LEVEL=info ...`. Verified locally — zero NCSA entries in `docker logs` after a `curl` and several healthcheck cycles.

**Deferred**

5. **`0000000.log` growth** — agreed this is out of scope for this PR; tracking under #5827 as you suggested. (Will add a quick clarifying comment on that issue: this file is the storage engine's write-ahead journal under `org.exist.storage.journal`, written by `Journal.java`, not a logger output — so the fix path is different from the rest of #5827's "revise log files" scope.)

Re-running CI now.

exist-core/src/main/java/org/exist/xquery/functions/websocket/WebSocketEndpoint.java
```diff
@@ -0,0 +1,217 @@
...
+                        gen.writeStringField(entry.getKey(), s);
+                    } else if (val instanceof Integer i) {
+                        gen.writeNumberField(entry.getKey(), i);
+                    } else if (val instanceof Boolean b) {
+                        gen.writeBooleanField(entry.getKey(), b);
+                    } else if (val == null) {
+                        gen.writeNullField(entry.getKey());
+                    } else {
+                        gen.writeStringField(entry.getKey(), val.toString());
+                    }
```
reinhapa: Consider converting this to a switch expression

exist-core/src/test/java/org/exist/xquery/functions/websocket/WebSocketEndpointTest.java
```diff
@@ -0,0 +1,185 @@
...
+                    @Override
+                    public void onMessage(final String message) {
+                        if (!subscribed && "ping".equals(message)) {
+                            // after first ping, subscribe to a channel
+                            try {
+                                session.getBasicRemote().sendText("{\"channel\": \"test-channel\"}");
+                                subscribed = true;
+                                subscribedLatch.countDown();
+                            } catch (IOException e) {
+                                throw new RuntimeException(e);
```
reinhapa: Throw a `UncheckedIOException` here

reinhapa (changes requested):

joewiz:

@reinhapa thanks for the review. Both points addressed in 961881f6e7:

1. **`WebSocketEndpoint.sendAll(String, Map)`** — converted the `if/else if (val instanceof ...)` chain into a switch expression with pattern matching, including a dedicated `case null` arm for the null field path.
2. **`WebSocketEndpointTest`** — replaced both `throw new RuntimeException(e)` wrappers around `IOException` with `throw new UncheckedIOException(e)`, which is the appropriate JDK type for this case.

Verified with `WebSocketEndpointTest` (3/3 pass) on the rebuilt module.

reinhapa (approved):

line-o:
Rebase please! Also there are three codacy warnings that would be good to look at.

joewiz:

@line-o Rebased onto current `develop` (`b917e1ab1d`). The only conflicts were in the POMs from the original Jetty 11→12 commit (`exist-core/pom.xml` removed the `jetty-jaas` + `mina-core` block — JAAS now ships in `jetty-security` under Jetty 12 — and `exist-parent/pom.xml` had a `jetty.version` overlap and a `log4j.version` overlap, where I kept develop's newer `log4j 2.26.0`). The remaining 18 commits applied cleanly.

The three Codacy PMD warnings are addressed in `bb122c4a42`:

- **`EvalProtocol.ClientMessage`** (`ExcessiveParameterList`, 10 params): grouped `streaming` + `chunkSize` into a new `StreamConfig` record.
- **`QueryExecutor.reportTerminationOrError` / `reportError`** (`ExcessiveParameterList`, 10 / 9 params): grouped the four error fields (`xpe`, `message`, `line`, `column`) into a private `ErrorInfo` record with named factories (`ErrorInfo.of(xpe)`, `ErrorInfo.of(msg)`, `ErrorInfo.of(msg, line, col)`).
- **`EvalWebSocketEndpointTest`** (`AvoidThrowingRawExceptionTypes`): replaced bare `throw new RuntimeException(e)` for caught `IOException` with the standard `UncheckedIOException(e)`. Codacy only flagged one delta-new instance, but the pattern was identical across all 24 sites in the test, so I applied it uniformly.

Local `codacy-cli analyze --tool pmd` reports zero violations on the three files. Local `mvn test` green for `*WebSocket*`, `XQuery3Tests` (1011), `XPathQueryTest` (150), `XQueryURLRewriteTest`, `extensions/webdav` (16), and `extensions/exquery/restxq` (20). CI rerunning.

exist-core/src/main/java/org/exist/http/ws/EvalProtocol.java
```diff
@@ -0,0 +1,341 @@
...
+     * Streaming configuration for a query response.
+     */
+    public record StreamConfig(boolean enabled, int chunkSize) {
+        public static final StreamConfig DEFAULT = new StreamConfig(true, DEFAULT_CHUNK_SIZE);
+    }
+
+    /**
+     * Parsed client request message.
+     */
+    public static final class ClientMessage {
```
reinhapa: Looks like also a record candidate to me..

exist-core/src/main/java/org/exist/http/ws/EvalProtocol.java
```diff
@@ -0,0 +1,341 @@
...
+        int chunkSize = DEFAULT_CHUNK_SIZE;
+
+        try (final JsonParser parser = JSON_FACTORY.createParser(json)) {
+            if (parser.nextToken() != JsonToken.START_OBJECT) {
+                throw new IOException("Expected JSON object");
+            }
+            while (parser.nextToken() != JsonToken.END_OBJECT) {
+                final String field = parser.currentName();
+                parser.nextToken();
+                switch (field) {
```
reinhapa, line 125: Use switch expression

reinhapa (changes requested):

exist-core/src/main/java/org/exist/http/ws/EvalProtocol.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `ea144ed2ee` — converted `ClientMessage` to a `record` and updated all callers in `EvalWebSocketEndpoint` and `QueryExecutor` to use record-accessor syntax. Drops ~25 lines of constructor boilerplate.

exist-core/src/main/java/org/exist/http/ws/EvalProtocol.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `ea144ed2ee` — converted to arrow form. Each JSON-field case is now a single labeled-arrow statement, dropping the `break;` boilerplate.

Kept it as an arrow-form switch *statement* rather than a switch expression because each case mutates a different local variable; refactoring to one switch expression per assigned variable would multiply parser passes (and arrow form is already the modernization Patrick has been asking for elsewhere).

dizzzz:
LGTM

exist-distribution/src/main/config/log4j2.xml
```diff
@@ -3,10 +3,14 @@
...
     <Properties>
         <Property name="logs">${log4j:configParentLocation}/../logs</Property>
+        <Property name="jetty.logs">${logs}/jetty</Property>
         <Property name="rollover.max.size">10MB</Property>
         <Property name="rollover.max">14</Property>
         <Property name="rollover.file.pattern">%d{yyyyMMddHHmmss}</Property>
         <Property name="exist.file.pattern">%d [%t] %-5p (%F [%M]:%L) - %m %n</Property>
+        <Property name="jetty.file.pattern">%d{ISO8601} [%t] %-5p %logger{36} - %m%n</Property>
+        <!-- GDPR: set EXIST_REQUEST_LOG_LEVEL=OFF to disable client IP logging -->
+        <Property name="exist.request.log.level">${env:EXIST_REQUEST_LOG_LEVEL:-info}</Property>
```
duncdrum: I m still in favour of setting this off by default

exist-core/src/test/java/org/exist/http/ws/EvalWebSocketEndpointTest.java
```diff
@@ -0,0 +1,1279 @@
...
+
+/**
+ * Integration tests for the /ws/eval WebSocket endpoint.
+ */
+public class EvalWebSocketEndpointTest {
+
+    private static final JsonFactory JSON_FACTORY = new JsonFactory();
+
+    private static final String TEST_COLLECTION = "/db/ws-eval-test";
+    private static final String TEST_MODULE = "module namespace test = 'http://exist-db.org/test';\n" +
```
duncdrum: use text block

duncdrum (changes requested):
two minor changes, logging defaults and code style. This is approaching the point where we should just merge and observe the fallout. 

You might want to rebase once more.

joewiz:

Addressed both of today's review items + rebased onto current develop (`a1a56f915f`):

- **`exist-distribution/src/main/config/log4j2.xml:13`** — `exist.request.log.level` default flipped from `info` to `off`. Operators who want request logs now opt in via `EXIST_REQUEST_LOG_LEVEL=info`. Reduces healthcheck-induced log noise in the container and avoids logging client IPs by default per the GDPR posture you raised in the testing notes.
- **`exist-core/src/test/java/org/exist/http/ws/EvalWebSocketEndpointTest.java:71`** — `TEST_MODULE` converted from multi-line `+`-concat to a Java 21 text block.

128-commit rebase came through clean; force-pushed. CI re-runs.

(Re the separate `0000000.log` continuously-growing concern in your testing notes — that's clearly a real issue but a different scope than the WebSocket/Jetty 12 changes here; happy to file a separate issue for follow-up so we can land #6145 without that hanging over it.)

duncdrum:
Great yes please open a new issue about the 0000.log

duncdrum (approved):

duncdrum:
@line-o @reinhapa good to go from
my side.

reinhapa (approved):

line-o:
Well then, let's see this in action!

Issue #6012, opened 2026-01-16
Allow returning 0 values in ft:facets#3 and ft:facets#4 functions (for consistency with similar functions)


daliboris:
**Is your feature request related to a problem? Please describe.**
The proposed feature will ensure consistency when calling functions to return range, field and facet values for node set.

The `range:index-keys-for-field#3`, `range:index-keys-for-field#4`, `util:index-keys#4`, `util:index-keys#5`, and `ft:index-keys-for-field#4`  functions accept `$max-number-returned as xs:int?` parameter defined as `The maximum number of returned keys` (or `The maximum number of keys to return`).

The `ft:facets#3` and  `ft:facets#4` functions accept `$count as xs:integer?` parameter defined as `The number of facet labels to be returned.`

While `range`, `util` and `ft` index keys functions return an empty set of items when `$max-number-returned` parameter is set to `0`, `ft:facet` functions throw `topN must be > 0 (got: 0)` error with following trace.

```text
at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:618) ~[exist.uber.jar:6.4.0]
at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127) ~[exist.uber.jar:6.4.0]
at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:235) ~[exist.uber.jar:6.4.0]
at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:1624) ~[exist.uber.jar:6.4.0]
at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:233) ~[exist.uber.jar:6.4.0]
at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1440) ~[exist.uber.jar:6.4.0]
at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:188) ~[exist.uber.jar:6.4.0]
at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:505) ~[exist.uber.jar:6.4.0]
at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1594) ~[exist.uber.jar:6.4.0]
at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:186) ~[exist.uber.jar:6.4.0]
[...]

Issue #4881, opened 2023-04-19
[BUG] matches is not being handled by the range module


marmoure:
While working on issue #3620, I noticed that `matches` is not being handled by the Range Index module. 

For example, these queries are being handled by the Range Index module:
```xquery
collection("/db/test")//foo[@bar eq "baz"]
collection("/db/test")//foo[contains(@bar, "b")]
collection("/db/test")//foo[starts-with(@bar, "b")]
```
The exception is the following query:
```
collection("/db/test")//foo[matches(@bar, "^b")]
```
Which is not handled by the Range Index, and whilst debugging this, I was unable to trigger any breakpoints within the Range Index module.
The code for `matches` is here: https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/xquery/functions/fn/FunMatches.java

See below for the full test:
```xquery
xquery version "3.1";

let $data := 
    <root>
        <foo bar="baz"/>
    </root>

let $old-range-xconf := 
    <collection xmlns="http://exist-db.org/collection-config/1.0">
        <index xmlns:xs="http://www.w3.org/2001/XMLSchema">
            <create qname="@bar" type="xs:string"/>
        </index>
    </collection>

let $new-range-xconf := 
    <collection xmlns="http://exist-db.org/collection-config/1.0">
        <index xmlns:xs="http://www.w3.org/2001/XMLSchema">
            <range>
                <create qname="@bar" type="xs:string"/>
            </range>
        </index>
    </collection>
    
let $test-col := xmldb:create-collection("/db", "test")
let $xconf-col :=   xmldb:create-collection("/db/system/config/db", "test")
let $conf := for $col in ("old-range-index", "new-range-index", "no-range-index")
    return
        (
            xmldb:create-collection("/db/test", $col),
            xmldb:create-collection("/db/system/config/db/test", $col),
            xmldb:store("/db/test/" || $col, "test.xml", $data)
        )
let $store-xconf-old :=    xmldb:store("/db/system/config/db/test/old-range-index", "collection.xconf", $old-range-xconf)
let $store-xconf-new :=    xmldb:store("/db/system/config/db/test/new-range-index", "collection.xconf", $new-range-xconf)
let $reinedx :=    xmldb:reindex("/db/test")

(:let $hits := collection("/db/test")//foo[@bar eq "baz"]:)
let $hits := collection("/db/test")//foo[matches(@bar, "^b")]
[...]

Issue #4835, opened 2023-03-27
Only last matched item in the parent element is highlighted


daliboris:
**Describe the bug**

If an element contains multiple found items, after calling `util:expand()` function, only the last matched item in the the parent element is expanded/highlighted.

**Expected behavior**
I would expect a function like `util:expand` to highlight **all** matches in full text hits.

**To Reproduce**

```Xquery
xquery version "3.1";
module namespace t="http://exist-db.org/xquery/test";
declare namespace test="http://exist-db.org/xquery/xqsuite";
declare namespace tei="http://www.tei-c.org/ns/1.0";
declare namespace exist = "http://exist.sourceforge.net/NS/exist";

declare variable $t:XML-multiple-nested := document { 
 <root>
  <div>
   <p>Letter</p>
   <p>LETTER</p>
  </div>
  <div>
   <p>letter</p>
   <p>leTTer</p>
   <div>
    <p>LeTtEr</p>
   </div>
  </div>
 </root>
};

declare variable $t:XML-two-in-one := document { 
 <root>
   <p>Letter and letter</p>
 </root>
};

declare variable $t:XML-two-in-two := document { 
 <root>
   <p>Letter and letter</p>
   <p>LETTER and leTTer</p>
 </root>
};

declare variable $t:XML-three-in-three := document { 
 <root>
   <p>Letter</p>
   <p>LETTER</p>
   <p>letter</p>
 </root>
};

declare variable $t:XML-three-different-elements := document { 
 <root>
   <p>Letter</p>
   <h1>LETTER</h1>
   <h2>letter</h2>
 </root>
};



declare variable $t:xconf :=
    <collection xmlns="http://exist-db.org/collection-config/1.0">
    <index xmlns:xs="http://www.w3.org/2001/XMLSchema">
        <fulltext default="none" attributes="false"/>
        <lucene>
         <text qname="p" />
         <text qname="h1" />
         <text qname="h2" />
        </lucene>
    </index>
</collection>;

   declare variable $t:testCol := xmldb:create-collection("/db", "test");
   declare variable $t:indexCol := xmldb:create-collection("/db/system/config/db", "test");


declare
    %test:setUp
function t:setup() {
        (
            xmldb:store($t:testCol, "test-multiple-nested.xml", $t:XML-multiple-nested),
            xmldb:store($t:testCol, "test-two-in-one.xml", $t:XML-two-in-one),
[...]

Issue #4539, opened 2022-09-05
[feature] type parameter in ft:field#3 should treat empty sequence as "return untyped value"


line-o:
**Is your feature request related to a problem? Please describe.**

It can be handy for dynamic queries to be able to parameterise the type parameter of calls to `ft:field`.
But there is no way to do that with the current implementation that requires the third parameter to be of type `xs:string` and
no way to signal that the raw value should be returned instead. 

**Describe the solution you'd like**

```xquery
ft:field($node, "myfield", ())
```

Should return the same untyped field values as

```xquery
ft:field($node, "myfield")
```

**Describe alternatives you've considered**

Introducing a keyword like `raw` might be an option. But then again this needs to be memoized by users and it does not simplify the implementation in any way.
Of course, this limitation can be worked around by wrapping calls to `ft:field`.

PR #6146, opened 2026-03-17
[enhancement] update Lucene to 10

duncdrum:
### Summary
- Upgrades eXist full-text and range indexing to Lucene 10.3.0.
- Adds **vector KNN search** (Lucene `KnnFloatVector*`) plus **embedding and storage** (`vector.dbx`, ONNX and HTTP providers, XQuery `vector:*` module).
- Adds regression coverage and fixes for historic Lucene/range issues (e.g. **#4389**, **#4539**, **#4835**, **#3977**, **#4074**, **#4881**, **#6012**) and vector IT/XQSuite coverage.

### Status / tests
- Lucene JUnit + XQSuite: **~686 tests, 0 failures** (some pending/ignored around attr-in-predicate context propagation).
- Range index: **349/349 passing**.
- Vector IT: **13 passing** (plus expanded XQuery vector tests).

### Notes / known gaps
- Remaining work and pending tests  (e.g. predicate-context propagation for `ft:query(., ...)` in some attribute-in-predicate patterns).
- java's vector incubator module is not available on JRE's 21 - 25, its available in JDKs
- regex traversal from `ft:matches` to Lucene falls back to full scan when translation fails. 
- `xmldb:reindex('db/' , 'fulltext')` describes the scope, `fulltext` is no longer treated as file name. `fulltext.xml`, `fulltext.txt`, etc.  still are. 

### Test plan
- [x] `mvn test -pl extensions/indexes/lucene -DfailIfNoTests=false -Denforcer.skip -Ddependency-check.skip -Dlicense.skip -Dmdep.analyze.skip`
- [x] `mvn -pl extensions/indexes/range -am test -Denforcer.skip -Ddependency-check.skip -Dlicense.skip -Dmdep.analyze.skip`
- [x] `mvn verify` (optionally with `-Ponnx-model` for embedding IT)

---


windauer (approved):

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/XMLToQuery.java
```diff
@@ -113,7 +130,7 @@ private Query phraseQuery(String field, Element node, Analyzer analyzer) throws
         for (int i = 0; i < termList.getLength(); i++) {
             Element elem = (Element) termList.item(i);
             String text = getText(elem);
-            if (text.indexOf('?') > -1 || text.indexOf('*') > 0) {
+            if (text.indexOf('?') > -1 || text.indexOf('*') > -1) {
```
line-o, line 138: Note to self: check leading wildcards

adamretter:
Would you be interested in review of this @duncdrum ?

duncdrum:
@adamretter certainly, feedback is always welcome.

adamretter:
@duncdrum cool! Let me know when it's out of draft status and ready to review and I'll take a look.

duncdrum:
@adamretter looking forward to it. Fire away. 

We discussed how to deal with the simultaneous large changes from open PR, and decided for a speedy merge so we can see how they interact. 

Initial testing shows this to be pretty stable, so I consider this a good merge candidate to evaluate the overall design. While feedback on any aspect is welcome, my main concerns for this PR are: 
- stability between old Fulltext in JRE, vs new hotness with JDK, incubator module, etc. Old indexes should continue to work as before within a similar performance bracket.
- e2e integration of new vector features in different contexts.
- Index changes in the context of other architectural changes.

beyond what's mentioned in the OP and inline, I have plans for additional convenience functions for the API, and a dedicated round of performance improvements.

.github/ISSUE_TEMPLATE/bug_report.md
line-o, line 1: In a perfect world this would have been a separate commit

.github/workflows/ci-test.yml
```diff
@@ -90,15 +90,22 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: ${{ steps.install-mvnd.outputs.mvnd-dir }}/mvnd -V -B --no-transfer-progress -T 1C compile test-compile -DtrimStackTrace=false -D'dependency-check.skip' -D'license.skip' --projects '!exist-installer'
+        run: ${{ steps.install-mvnd.outputs.mvnd-dir }}/mvnd -V -B --no-transfer-progress -T 1C compile test-compile -Ponnx-model -DtrimStackTrace=false -D'dependency-check.skip' -D'license.skip' --projects '!exist-installer' 
       - name: Maven Unit Tests
+        id: unit-tests
         if: matrix.test-type == 'unit'
         timeout-minutes: 45
         run: ${{ steps.install-mvnd.outputs.mvnd-dir }}/mvnd -V -B --no-transfer-progress test -DtrimStackTrace=false -D'dependency-check.skip' -D'license.skip' -D'mvnd.maxLostKeepAlive=6000'
+      - name: Dependency tree on unit test failure
```
line-o, line 100: Is this necessary to stay in or was it just used to debug while developing?

.github/workflows/ci-test.yml
```diff
@@ -128,3 +135,4 @@ jobs:
             **/hs_err_pid*.log
             **/target/surefire-reports/*
             **/target/failsafe-reports/*
+            dependency-tree-lucene.txt
```
line-o, line 139: see above

exist-core/src/main/java/org/exist/indexing/ReindexScope.java
```diff
@@ -0,0 +1,67 @@
...
+     * Parse mode string from XQuery. Returns {@link #ALL} for null, empty, or invalid values.
+     */
+    public static ReindexScope fromString(final String mode) {
+        if (mode == null || mode.isEmpty()) {
+            return ALL;
+        }
+        switch (mode.toLowerCase()) {
+            case "all":
+                return ALL;
+            case "fulltext":
```
line-o, line 59: I wonder if the range index is a special reindex scope that is either part of all or always active.
Can you elaborate on that?

exist-core/src/main/java/org/exist/storage/vector/VectorStore.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Gets a vector by document path and node id.
+     *
+     * @param docPath document URI (e.g. /db/collection/doc.xml)
+     * @param nodeId  node id within the document
+     * @return float32 bytes, or null if not found
+     */
+    @Nullable
+    byte[] get(String docPath, org.exist.numbering.NodeId nodeId) throws IOException;
```
line-o, line 45: I would want to see an XQuery function to read the vector stored for a node.

exist-versioning-release.md
```diff
@@ -333,7 +333,7 @@ project.version.numeric = 3.0.4
 Our goal is to prevent such a divergence in versioning from creeping back into eXist and to ensure a clean versioning system to serve eXist through its future development and release cycles.
 
 ### The New Way
-eXist now has a single version number, forumlated according to the precise principles of Semantic Versioning, captured in a single property:
```
line-o, line 336: If possible I would want to see this excluded from the PR

tools/backrest/src/eXist-settings.sh
```diff
@@ -71,13 +71,17 @@ restore_library_path() {
 
 set_client_java_options() {
     if [ -z "${CLIENT_JAVA_OPTIONS}" ]; then
+	# FIXME: Add --add-modules jdk.incubator.vector for optimal Lucene 10 SIMD
```
line-o, line 74: As far as I know this file is unused. We might want to remove tools/backrest entirely until someone here can justify us keeping it.

tools/backrest/src/backup.bat
```diff
@@ -47,6 +47,8 @@ echo home directory of eXist BackRest.
 goto :eof
 
 :gotExistHome
+REM FIXME: Add --add-modules jdk.incubator.vector for optimal Lucene 10 SIMD
```
line-o, line 50: see below

extensions/vector/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <!-- Profile gpu: mvn -P gpu ... uses onnxruntime_gpu for CUDA (requires NVIDIA GPU + CUDA) -->
+    <profiles>
+        <profile>
+            <id>gpu</id>
```
line-o, line 99: Have you tested this? Do we want to maintain this already? Is this an experimental feature?

extensions/vector/src/main/java/org/exist/vector/VectorEmbeddingService.java
```diff
@@ -0,0 +1,161 @@
...
+      @Nullable final String pathOrUrl,
+      final int dimension,
+      @Nullable final String apiKey) {
+    final ModelRegistry.Resolved resolved = ModelRegistry.getInstance().resolve(modelId, pathOrUrl, dimension);
+    final String path = resolved.path;
+    final int dim = resolved.dimension;
+    if (HttpVectorProvider.isHttpApiUrl(path)) {
+      final boolean useCache = apiKey == null || apiKey.isEmpty();
+      if (useCache) {
+        final String cacheKey = "http:" + modelId + ":" + path;
```
line-o: cachekey construction could be a separate method (duplicated a few lines later)

extensions/vector/src/main/java/org/exist/vector/VectorModelConstants.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Centralized model IDs and dimensions for built-in embedding models (ONNX and HTTP API).
+ * Used by Embed, EmbedBatch, Models, and HttpVectorProvider to avoid drift.
+ */
+public final class VectorModelConstants {
+
+  /** OpenAI embedding models. */
+  private static final Map<String, Integer> OPENAI_DIMENSIONS = Map.of(
+      "text-embedding-ada-002", 1536,
```
line-o, line 35: Is there a path to extend this list of models by users or does this require new releases of exist?

extensions/vector/src/main/java/org/exist/xquery/modules/vector/Embed.java
```diff
@@ -0,0 +1,132 @@
...
+    final String text = args[0].getStringValue();
+    final String model = requireModel(args[1]);
+    final String pathOrUrl = optionalArg(args, 2);
+    final String apiKey = optionalArg(args, 3);
+    final int dimension = VectorModelConstants.getDefaultDimension(model);
+
+    try {
+      final VectorEmbeddingProvider provider = VectorEmbeddingService.getInstance().getProvider(model, pathOrUrl, dimension, apiKey);
+      if (provider == null) {
+        throw new XPathException(this, "Failed to load embedding model: " + model);
```
line-o: I would like to see all XPathExceptions to have ErrorCode (either one of the built-ins or a module specific one). Because otherwise it will be hard to catch them and also distinguish error states

extensions/vector/src/main/java/org/exist/xquery/modules/vector/EmbedBatch.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+
+  private ArrayType embedAll(final VectorEmbeddingProvider provider, final Sequence textsSeq) throws XPathException {
+    final List<Sequence> resultArrays = new ArrayList<>();
+    for (final SequenceIterator it = textsSeq.iterate(); it.hasNext(); ) {
+      final String text = it.nextItem().getStringValue();
+      final float[] vec = provider.embed(text, true);
+      if (vec == null || vec.length == 0) {
+        throw new XPathException(this, "Embedding returned empty result for text: "
```
line-o: ErrorCode would be better (see above)

extensions/vector/src/main/java/org/exist/xquery/modules/vector/EmbedBatch.java
```diff
@@ -0,0 +1,144 @@
...
+      "Texts to embed (e.g. document titles for batch indexing).");
+  private static final FunctionParameterSequenceType FS_PARAM_MODEL = param("model", Type.STRING,
+      "Model identifier (e.g. 'all-MiniLM-L6-v2').");
+  private static final FunctionParameterSequenceType FS_PARAM_MODEL_PATH = optParam("model-path", Type.STRING,
+      "Optional path to model directory. If omitted, resolves via onnx-models/{model} or conf.xml registry.");
+  private static final FunctionParameterSequenceType FS_PARAM_API_KEY = optParam("api-key", Type.STRING,
+      "Optional API key for HTTP APIs (OpenAI, Cohere). Alternative to OPENAI_API_KEY/COHERE_API_KEY env vars.");
+
+  static final FunctionSignature[] signatures = {
+      functionSignature("embed-batch",
```
line-o, line 64: I would prefer "batch-embed" as the name of this function.

extensions/indexes/lucene/src/test/xquery/lucene/index-keys.xql
```diff
@@ -162,6 +162,20 @@ function idxk:tearDown() {
...
 
+(:~
+ : util:index-keys on path-based Lucene index (match="//pPath") should return terms.
+ : getQNamesFromNodes or scanIndex must resolve path-based Lucene indexes from context.
+ : New test for Lucene 10; counterpart: idxk:index-scan-path-lucene.
+ :)
+declare
+    %test:assertTrue
+function idxk:index-scan-path-lucene-has-terms() {
+    let $callback := util:function(xs:QName("idxk:term-callback"), 2),
```
line-o: I prefer the Xquey standard way of referencing functions here `let $callback := idx:term-callback#2`

extensions/indexes/lucene/src/test/xquery/lucene/ft-search-binary.xqm
```diff
@@ -0,0 +1,313 @@
...
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+(:~
+ : XQSuite tests for ft:index and ft:search on binary/non-XML data.
+ : Consolidates binary.xql, plain-store.xql, plain-ft-functions.xql.
```
line-o, line 26: I think it would be safe to add a note that this does _not_ test binary fields

extensions/indexes/lucene/src/test/xquery/lucene/ft-attr-context.xql
```diff
@@ -137,17 +137,51 @@ declare %test:assertTrue function ftac:query-path-element-context() {
...
+(:~
+ : FIXME: attr-in-predicate path — same semantic as query-qname-attr-in-predicate but
+ : uses /*[descendant-or-self::p/@att1[ft:query(., ...)]] which triggers "context is
+ : missing for node". LuceneHitCollector returnAncestor copies null context from storedNode
+ : onto parentNode. Fix: preserve parentNode context when storedNode has none.
+ :)
+declare
+    %test:pending("FIXME: context is missing for node; attr-in-predicate path evaluation")
+    %test:assertTrue
+function ftac:query-qname-attr-in-predicate-broken() {
```
line-o, line 171: This is a regression?

extensions/indexes/lucene/src/test/resources/log4j2.xml
```diff
@@ -29,6 +29,7 @@
         </Console>
     </Appenders>
     <Loggers>
+        <Logger name="org.exist.storage.NativeBroker" level="INFO"/>
```
line-o, line 32: Was this needed for debugging or should this stay in? If it should stay I would like to learn why.

extensions/indexes/lucene/src/test/java/xquery/lucene/VectorSearchEmbeddingTest.java
```diff
@@ -0,0 +1,113 @@
...
+ * (profile sets exist.home to project.basedir and downloads model to target/onnx-models/all-MiniLM-L6-v2).
+ */
+public class VectorSearchEmbeddingTest {
+
+    @ClassRule
+    public static ExistEmbeddedServer existEmbeddedServer = new ExistEmbeddedServer(true, true);
+
+    private static final String COLLECTION = "/db/lucene-test-vector-embedding-local";
+    private static final String COLLECTION_NAME = "lucene-test-vector-embedding-local";
+    private static final String XCONF =
```
line-o: I find the new [Java text blocks](https://openjdk.org/jeps/378) much better readable than these `"" + ""`
I am aware that this is the form we have a ton of but I recently started changing that. Since, these are new additions we should think if it makes sense to do this from the start.

extensions/indexes/lucene/src/test/java/org/exist/indexing/lucene/LuceneMatchListenerTest.java
```diff
@@ -301,8 +306,135 @@ public void inlineNodes_whenNotIndenting() throws EXistException, PermissionDeni
...
+            // Batch util:expand (all 5 in one call)
+            Sequence batchExpand = xquery.execute(broker,
+                "let $doc := doc('" + TestConstants.TEST_COLLECTION_URI + "/test_matches.xml') " +
+                "let $hits := $doc//p[ft:query(., 'letter')] " +
+                "return count(util:expand($hits)//exist:match)", null);
+            int batchCount = batchExpand.itemAt(0).toJavaObject(Integer.class).intValue();
+
+            // Diagnostic output when assertions fail; run with -Dtest=LuceneMatchListenerTest#issue4835_diagnostic
+            if (totalExpand != 5 || batchCount != 5) {
+                System.err.println("[#4835 diagnostic] let binding: " + hitsLet.getItemCount() + " hits, " + withMatchesLet + " with matches");
```
line-o, line 368: Why does this test print directly to stderr?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryVector.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.Set;
+
+import static org.exist.xquery.FunctionDSL.optManyParam;
+import static org.exist.xquery.FunctionDSL.optParam;
+import static org.exist.xquery.FunctionDSL.param;
+import static org.exist.xquery.FunctionDSL.returnsOptMany;
+import static org.exist.xquery.modules.lucene.LuceneModule.functionSignature;
+
+/**
+ * ft:query-vector(nodes, vector, k?, options?) — KNN vector search with node context.
```
line-o, line 48: I think we should discuss if ft is the right namespace for vector queries. It is not using the full text per-se.
I see `vector:query` as an alternative option

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryFieldVector.java
```diff
@@ -0,0 +1,160 @@
...
+import javax.annotation.Nullable;
+import java.io.IOException;
+
+import static org.exist.xquery.FunctionDSL.optParam;
+import static org.exist.xquery.FunctionDSL.param;
+import static org.exist.xquery.FunctionDSL.returnsOptMany;
+import static org.exist.xquery.modules.lucene.LuceneModule.functionSignature;
+
+/**
+ * ft:query-field-vector(field, vector, k?, options?) — KNN vector search by field name.
```
line-o, line 42: same here: We should discuss if ft is the correct namespace for this function

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneUtil.java
```diff
@@ -51,9 +51,15 @@
...
     public static final String FIELD_NODE_ID = "nodeId";
+    /** DocValues-only field for nodeId; avoids conflict with indexed Field of same name (LUCENE-6019). */
+    public static final String FIELD_NODE_ID_DV = "nodeId_dv";
 
     public static final String FIELD_DOC_ID = "docId";
     public static final String FIELD_DOC_URI = "docUri";
+    /** Per-document boost for attribute/config-based scoring (Lucene 10: index-time boost via FloatDocValues) */
+    public static final String FIELD_BOOST = "_boost";
+    /** Identifies which index config created the Lucene doc; used to filter when multiple indexes share a collection. */
+    public static final String FIELD_INDEX_TYPE = "_idx";
```
line-o, line 62: Note to self: Is this used?

line-o (approved):
Lots todo still but nothing so serious that we should not pull that in.

If we decide to split some of the unrelated things out or there are other changes to the history then we can wait but if not we will have to patch later.

I forgot to mention here that the third parameter in `ft:query-vector` is not being taken into account. Whatever value is given all matches are returned.

And huge props to you @duncdrum to take this on by yourself. Looking forward to engage more on the further development.

.github/workflows/ci-test.yml
duncdrum: yes necessary as long as we haven't hit `develop` and ci replaces the cache with updated deps

.github/ISSUE_TEMPLATE/bug_report.md
duncdrum: we live in different perfect worlds

exist-core/src/main/java/org/exist/indexing/ReindexScope.java
duncdrum: its part of fulltext, so it runs with all and fulltext.

exist-core/src/main/java/org/exist/storage/vector/VectorStore.java
duncdrum: sounds like a nice addition to the Xquery api provided here

extensions/vector/pom.xml
duncdrum: we are simply exposing the flag from onnxruntime, which is tested upstream and not experimental. I have not yet run CUDA tests in particular.

extensions/vector/src/main/java/org/exist/vector/VectorModelConstants.java
duncdrum: see the docs, you can load any onnx compatible model

extensions/indexes/lucene/src/test/xquery/lucene/ft-attr-context.xql
duncdrum: see op

tools/backrest/src/eXist-settings.sh
duncdrum: If these are unused removing them is out of scope for this PR

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryVector.java
duncdrum: we did discuss the names of the functions, and how they map to preexisting functions weeks ago.

extensions/indexes/lucene/src/test/java/xquery/lucene/VectorSearchEmbeddingTest.java
duncdrum: agreed, I already have a commit refactoring text blocks in the whole codebase. Ready for the next PR

extensions/indexes/lucene/src/test/java/org/exist/indexing/lucene/LuceneMatchListenerTest.java
duncdrum: faster iteration when debugging this, changing it now is probably the right call.

duncdrum:
> I forgot to mention here that the third parameter in ft:query-vector is not being taken into account. Whatever value is given all matches are returned.

This sounds like real bug that I missed, and more worryingly the test suite as well. How did you test this? 

It's possible that all matches are above the threshold for different values irrespective of k value. I need to check but I don't think `k=O` is a thing.

extensions/indexes/lucene/src/test/resources/log4j2.xml
duncdrum: Stay in possibly even after rc1. Generally speaking the logging is still very rough, and set up for more debugging work to come. 

I intend to clean it up once we are closer to cutting a full release. Maybe one of the release candidates. 

Logging native broker comes into play for almost any debugging work.  This helps to make the embedded instance db that is used during test less of a black box.

adamretter:
@duncdrum I should be able to review after Easter. So I can set my expectations, is this AI generated code, or code written manually by yourself?

duncdrum:
@adamretter a wild mix of both.

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
```diff
@@ -167,6 +169,8 @@ public void flush() {
             case REMOVE_BINARY:
             	removePlainTextIndexes();
             	break;
+            default:
```
reinhapa: do we want to have a default or should we add all missing cases in oder to get a compile error on a future change?

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
```diff
@@ -207,6 +211,8 @@ public void setMode(ReindexMode mode) {
             case REMOVE_SOME_NODES:
                 nodesToRemove = new TreeSet<>();
                 break;
```
reinhapa, line 213: See above comment...

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/XMLToQuery.java
```diff
@@ -242,57 +273,164 @@ private int getSlop(Element node) throws XPathException {
...
+                case '\\':
+                    sb.append("\\\\");
+                    break;
+                case '.': case '+': case '(': case ')': case '[': case ']':
+                case '{': case '}': case '^': case '$': case '|':
+                    sb.append('\\').append(c);
+                    break;
+                default:
+                    sb.append(c);
+            }
```
reinhapa, line 309: We could write this more compact using the new switch: 

```suggestion
            switch (c) {
                case '*' -> sb.append(".*");
                case '?' -> sb.append(".");
                case '\\' -> sb.append("\\\\");
                case '.', '+', '(', ')', '[', ']', '{', '}', '^', '$', '|' -> sb.append('\\').append(c);
                default -> sb.append(c);
            }
```

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Query.java
```diff
@@ -235,7 +243,28 @@ public NodeSet preSelect(final Sequence contextSequence, final boolean useContex
...
+                            if (qn != null) {
+                                seen.add(qn);
+                            }
+                        }
+                        if (!seen.isEmpty()) {
+                            qnames = new ArrayList<>(seen);
+                        }
+                    }
+                } catch (final IOException ignored) {
+                }
```
reinhapa, line 266: Should we log this at least on debug instead of completely ignore it? Maybe with like `logger.debug("Ignoring error", e)` or something like this?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Query.java
```diff
@@ -280,7 +321,25 @@ public Sequence eval(Sequence contextSequence, @Nullable final Item contextItem)
...
+                                final QName qn = proxy.getQName();
+                                if (qn != null) {
+                                    seen.add(qn);
+                                }
+                            }
+                            if (!seen.isEmpty()) {
+                                qnames = new ArrayList<>(seen);
+                            }
+                        }
+                    } catch (final IOException ignored) {
```
reinhapa: Same as above

extensions/indexes/range/src/main/java/org/exist/indexing/range/RangeIndexWorker.java
reinhapa, line 1: New switches may be more compact... Or should this be addressed later?

extensions/vector/src/main/java/org/exist/vector/HttpVectorProvider.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+    try {
+      switch (apiType) {
+        case OPENAI:
+          return embedOpenAI(text);
+        case COHERE:
+          return embedCohere(text, forQuery);
+        default:
+          return null;
+      }
```
reinhapa, line 106: Even more compact using switch as a function:
```suggestion
      return switch (apiType) {
        case OPENAI -> embedOpenAI(text);
        case COHERE -> embedCohere(text, forQuery);
        default-> null;
      };
```

extensions/vector/src/main/java/org/exist/vector/VectorModelConstants.java
```diff
@@ -0,0 +1,107 @@
...
+      Map.entry("text-embedding-3-small", 1536),
+      Map.entry("text-embedding-3-large", 3072),
+      Map.entry("embed-english-v3.0", 1024),
+      Map.entry("embed-multilingual-v3.0", 1024),
+      Map.entry("embed-english-light-v3.0", 384),
+      Map.entry("embed-multilingual-light-v3.0", 384)
+  );
+
+  /** Built-in model IDs (for vector:models when not in registry). */
+  public static final String[] KNOWN_MODEL_IDS = {
```
reinhapa: In order to improve security and immutability would a List created with`List.of()` be preferable as even though the actual constant is final, the entries can be modified externally.

extensions/indexes/lucene/src/test/java/org/exist/indexing/lucene/LuceneMatchListenerTest.java
duncdrum: actually looking more closely, no stderr is the right location. This is separate from normal logging, only happens during targeted test runs, when they fail, and doesn't require wiring log4j into the testfile. Knowing how hard it was to track this, I also don't want to remove these, in case somebody else is stumped by what's happening down the road.

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneUtil.java
duncdrum: yes two places inside LuceneIndexWorker e.g. `filterByIndexType`

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java
duncdrum: good point, I think we can make the intent clearer.

extensions/vector/src/main/java/org/exist/vector/HttpVectorProvider.java
duncdrum: already part of see #6226

duncdrum:
@line-o @reinhapa thank you for the reviews. I addressed as much as I could in here. The linked follow up PR is my preferred option for further tweaks or features like additional xquery functions, or the potential bug concerning k-value handling @line-o memtioned. 

I hope that this is merge ready now, so we can unblock our backlog.

exist-core/src/main/java/org/exist/storage/NativeBroker.java
```diff
@@ -591,6 +591,10 @@ public void backupToArchive(final RawDataBackup backup) throws IOException, EXis
         }
         pool.getSymbols().backupToArchive(backup);
         pool.getBlobStore().backupToArchive(backup);
+        final org.exist.storage.vector.VectorStore vectorStore = pool.getVectorStore();
+        if (vectorStore != null) {
+            vectorStore.backupToArchive(backup);
+        }
         pool.getIndexManager().backupToArchive(backup);
         //TODO backup counters
         //TODO USE zip64 or tar to create snapshots larger then 4Gb
```
dizzzz, line 600: zip64 is default since javaXYZ / 64bit

dizzzz (approved):
Nice piece of work!

extensions/indexes/vector-it/src/test/java/xquery/lucene/VectorSearchEmbeddingIT.java
reinhapa, line 1: Maybe a textblock would be more readable?

reinhapa (approved):
As far as I can judge, it looks good so far. 

We might need to look into the constant arrays in a second pass an convert those to imutable lists from a security perspective, as the individual positions are not imutable even though tha array itself is.

dizzzz:
@reinhapa I agree that this gives us a good baseline to learn and understand. A good basis to further tune and tweak and improve. AT least we can get 'running hours' and feedback

extensions/indexes/vector-it/src/test/java/xquery/lucene/VectorSearchEmbeddingIT.java
duncdrum: good catch, I think my moderne pass skipped IT test files.

PR #6153, opened 2026-03-20
[test] Re-enable 9 skipped tests that now pass

joewiz:
## Summary

Audited all 129 skipped tests (`@Ignore`, `%test:pending`, surefire excludes) across the codebase.

### Tests re-enabled (8)

| Test | Skip Type | Why It Was Skipped | Why It Passes Now |
|---|---|---|---|
| `CollectionConfigurationTest#rangeIndex1` | `@Ignore` | No reason given | Range index query works |
| `CollectionConfigurationTest#rangeIndexOverAttributes` | `@Ignore` | No reason given | Range index on attributes works |
| `XPathQueryTest#ids_memtree` | `@Ignore` | "Not yet supported in eXist" | `fn:id` on memtree now works |
| `AdditionalJingXsdRngTest#repeatTests` | `@Ignore` | "Looks good, but memory issue" | 1000 iterations in 7.5s, no OOM |
| `arr:get-invalid-type` | `%test:pending` | "eXist-db does not correctly detect the type error" | `XPTY0004` now raised for non-integer array index |
| `fd:format-dateTime-ZN-zero` | `%test:pending` | "[ZN] is not yet supported" | `[ZN]` works for UTC offset |
| `fd:format-dateTime-ZN-NY-negative-single-digit` | `%test:pending` | "[ZN] is not yet supported" | `[ZN]` works with IANA timezone ID |
| `syncmod:prunes-a-file-with-after` | `%test:pending` | No reason given | Passes cleanly |

### Annotation cleanup

All 34 remaining bare `@Ignore` annotations now have descriptive reasons, so future developers don't need to investigate why each test is skipped. GitHub issues filed for undocumented failures:
- #6156 — Whitespace suppress-type not implemented (17 tests)
- #6157 — ResourceSet intersection wrong count
- #6158 — Removing root collection doesn't remove documents
- #6159 — Context item not propagated into util:eval()
- #6160 — Adjacent text nodes not merged after XUpdate append

### Related PRs (split out per reviewer request)
- #6162 — [bugfix] Fix negative double/float comparison in value index
- #6163 — [bugfix] Validate json-to-xml escape option type (FOJS0005)

## Phase 2 cross-reference

A second audit pass (#6164) found 3 additional tests that could be re-enabled after bug fixes. Of the 120 remaining skipped tests, only 2 were fixed by changes on the `next` integration branch — the rest require targeted bug fixes or are legitimately deferred (e.g., XML Schema, known platform issues).

## CI Status

[...]

PR #6162, opened 2026-03-21
[bugfix] Fix negative double/float comparison in value index

### LINKED ISSUE #6161: Negative double/float values compare incorrectly in value index
**Fixed in this PR** — filed for tracking.

The value index serialization for \`xs:double\` and \`xs:float\` used \`bits ^ SIGN_BIT\` for all values regardless of sign. This only produces correct unsigned byte ordering for positive values. For negative IEEE 754 values, flipping only the sign bit leaves the magnitude bits in the wrong order (more-negative values sort after less-negative values).

**Fix:** Applied the standard IEEE 754 sortable encoding:
- Positive: flip sign bit (so positives sort after negatives)
- Negative: flip all bits (to invert magnitude ordering)

**Tests:** \`ValueIndexFactoryTest#negativeNumbersComparison\`, \`negativeNumbersComparison2\`

**Note:** Existing range indexes containing negative \`xs:double\` or \`xs:float\` values will need to be reindexed after this fix.

Fixed in #6153.

joewiz:
## Summary

The value index serialization for `xs:double` and `xs:float` used `bits ^ SIGN_BIT` for all IEEE 754 values regardless of sign. This only produces correct unsigned byte ordering for positive values. For negative values, the magnitude bits must also be flipped.

Applied the standard IEEE 754 sortable encoding:
- Positive values: flip sign bit (so positives sort after negatives)
- Negative values: flip all bits (to invert magnitude ordering)

Two previously `@Ignore`'d tests (`negativeNumbersComparison`, `negativeNumbersComparison2`) now pass and have been re-enabled, plus two new tests added for round-trip verification and comprehensive ordering.

**Note:** existing range indexes containing negative `xs:double` or `xs:float` values will need to be reindexed after this fix.

Closes #6161

Split out from #6153 per reviewer request.

## Test plan

- [x] `ValueIndexFactoryTest` — 6 tests, 0 failures
- [x] Full `exist-core` test suite — 0 regressions

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/ValueIndexFactory.java
+++ b/exist-core/src/main/java/org/exist/storage/ValueIndexFactory.java
@@ -93,15 +93,19 @@ public class ValueIndexFactory {
         }
         /* xs:double */
         else if (type == Type.DOUBLE) {
-            final long bits = ByteConversion.byteToLong(data, start +
-                    (ValueIndexFactory.LENGTH_VALUE_TYPE)) ^ 0x8000000000000000L;
+            long bits = ByteConversion.byteToLong(data, start +
+                    (ValueIndexFactory.LENGTH_VALUE_TYPE));
+            // Reverse the sortable encoding: if high bit is set, it was positive (flip sign bit);
+            // if high bit is clear, it was negative (flip all bits)
+            bits = bits < 0 ? bits ^ 0x8000000000000000L : ~bits;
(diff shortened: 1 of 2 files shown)


duncdrum (approved):

duncdrum:
I ll cross check with the lucene-update branch

reinhapa (approved):

duncdrum (approved):

PR #6164, opened 2026-03-22
[bugfix] Fix 3 skipped test failures: ResourceSet, util:eval, XUpdate

### LINKED ISSUE #6157: ResourceSet.intersection() returns wrong count
`ResourceSetTest#intersection1` is skipped with `@Ignore`. The test expects the intersection of two ResourceSets to return 87 results, but it returns 1.

**Test:** `exist-core/src/test/java/org/exist/xmldb/ResourceSetTest.java#intersection1`

**Steps:** Query two XPath result sets against `shakes.xsl` and compute their intersection.

**Expected:** 87 results  
**Actual:** 1 result

Found during the skipped tests audit in #6153.

### LINKED ISSUE #6158: Removing root collection does not remove contained documents
`RemoveRootCollectionTest#removeRootCollectionWithDocument` is skipped with `@Ignore`. After adding a document to the root collection and then removing the root collection, the document count is 1 instead of the expected 0.

**Test:** `exist-core/src/test/java/org/exist/storage/RemoveRootCollectionTest.java#removeRootCollectionWithDocument`

**Expected:** Document count 0 after removing root collection  
**Actual:** Document count 1

Found during the skipped tests audit in #6153.

### LINKED ISSUE #6159: Context item not propagated into util:eval()
`XQueryFunctionsTest#utilEval1` is skipped with `@Ignore`. The query `<a><b/></a>/util:eval('*')` raises XPDY0002 (undefined context sequence) instead of returning `<b/>`.

The context item from the path expression is not propagated into the dynamically evaluated expression in `util:eval()`.

**Test:** `exist-core/src/test/java/org/exist/xquery/XQueryFunctionsTest.java#utilEval1`

**Expected:** `<b/>`  
**Actual:** `XPDY0002: Undefined context sequence for 'child::element()'`

Found during the skipped tests audit in #6153.

### LINKED ISSUE #6160: Adjacent text nodes not merged after XUpdate append
`XQueryTest#xupdateWithAdjacentTextNodes` is skipped with `@Ignore`. After appending text node `yyy` to `<test>aaa</test>` via XUpdate, the result contains 2 text nodes instead of the expected 1 merged text node.

Per the XDM spec, adjacent text nodes should be merged.

**Test:** `exist-core/src/test/java/org/exist/xquery/XQueryTest.java#xupdateWithAdjacentTextNodes`

**Expected:** 1 text node (`aaayyy`)  
**Actual:** 2 text nodes (`aaa`, `yyy`)

Found during the skipped tests audit in #6153.

joewiz:
## Summary

Fixes 3 long-standing bugs uncovered by the skipped tests audit (#6153), re-enabling 3 previously `@Ignore`'d tests.

Closes https://github.com/eXist-db/exist/issues/6157
Closes https://github.com/eXist-db/exist/issues/6159
Closes https://github.com/eXist-db/exist/issues/6160

### 1. ResourceSet intersection returns wrong count (#6157)

`ResourceSetHelper.intersection()` keyed resources by document filename, so all nodes from the same document collapsed into a single entry. Rewrote to use a composite identity key (`documentId#nodeId`) for `LocalXMLResource` instances backed by `NodeProxy`.

Also fixed the test expectation: the intersection of `position() >= 5` and `position() <= 10` over 18 `xsl:template` children is 6, not 87.

### 2. Context item not propagated into util:eval() (#6159)

When `util:eval` is called as a step in a path expression (e.g., `<a><b/></a>/util:eval('*')`), the context item was not passed to the dynamically evaluated expression. Added a narrowly-scoped check: when no explicit context was provided and a non-empty context sequence exists from the calling expression, use it.

### 3. Adjacent text nodes not merged after XUpdate append (#6160)

After `Append.process()` creates new text nodes via `appendChildren()`, adjacent text nodes were left as siblings instead of being merged per the XDM spec. Added a `mergeAdjacentTextNodes()` post-processing step that walks the element's children and merges adjacent text pairs.

### Not fixed: #6158 (RemoveRootCollection)

Investigated and determined to be **by-design behavior** — the root collection `/db` is not truly removable. [Commented on the issue](https://github.com/eXist-db/exist/issues/6158#issuecomment-4106450426) with analysis.

## Test plan

- [x] `ResourceSetTest#intersection1` — passes (was returning 1, now returns 6)
- [x] `XQueryFunctionsTest#utilEval1` — passes (was XPDY0002, now returns `<b/>`)
- [x] `XQueryTest#xupdateWithAdjacentTextNodes` — passes (was 2 text nodes, now 1)
- [x] Full `exist-core` test suite: 6533 tests, 0 failures

## CI Status

All checks green — unit tests, integration tests (macOS/ubuntu/windows), XQTS, container image, license check, Codacy.

### DIFF
--- a/exist-core/src/main/java/org/exist/xmldb/ResourceSetHelper.java
+++ b/exist-core/src/main/java/org/exist/xmldb/ResourceSetHelper.java
@@ -21,40 +21,79 @@
  */
 package org.exist.xmldb;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.Set;
 
+import org.exist.dom.persistent.NodeProxy;
 import org.xmldb.api.base.Resource;
 import org.xmldb.api.base.ResourceSet;
 import org.xmldb.api.base.XMLDBException;
 
 /**
  * @author jmv
- *
  */
 public class ResourceSetHelper {
 
+    /**
+     * Computes the intersection of two resource sets.
+     *
+     * <p>For {@link LocalXMLResource} instances backed by persistent nodes,
+     * intersection is determined by node identity (document ID + node ID).
+     * For other resources, the resource ID ({@link Resource#getId()}) is used
+     * as the identity key.</p>
+     *
+     * @param s1 the first resource set
+     * @param s2 the second resource set
+     * @return a new resource set containing only resources present in both sets
+     * @throws XMLDBException if an error occurs accessing the resources
+     */
     public static ResourceSet intersection(final ResourceSet s1, final ResourceSet s2) throws XMLDBException {
-        final Map<String, Resource> m1 = new MapResourceSet(s1).getResourcesMap();
-        final Map<String, Resource> m2 = new MapResourceSet(s2).getResourcesMap();
-        final Set<String> set1 = m1.keySet();
-        final Set<String> set2 = m2.keySet();
-        set1.retainAll(set2);
-        final Map<String, Resource> m = new HashMap<>();
-        for (String key : set1) {
-            final Resource resource = m1.get(key);
-            m.put(key, resource);
+        // Build a map from identity key to resource for the first set
+        final Map<String, Resource> m1 = buildIdentityMap(s1);
+
+        // Build the intersection: for each resource in s2, check if it's also in s1.
+        // Use the node identity key (not res.getId()) as the MapResourceSet key
+        // to avoid collapsing multiple nodes from the same document into one entry.
+        final Map<String, Resource> result = new LinkedHashMap<>();
+        for (long i = 0; i < s2.getSize(); i++) {
+            final Resource res = s2.getResource(i);
+            final String key = getIdentityKey(res);
+            if (m1.containsKey(key) && !result.containsKey(key)) {
+                result.put(key, m1.get(key));
+            }
         }
-        final MapResourceSet res = new MapResourceSet(m);
-        /*
-         VectorResourceSet res = new VectorResourceSet(); 
-         Collection c1 = new VectorResourceSet(s1).getResources();
-         res.getResources().addAll( c1 );
-         Collection c2 = new VectorResourceSet(s2).getResources();
-         res.getResources().retainAll(c2);
-         return res;
-         */
-        return res;
+
+        return new MapResourceSet(result);
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


duncdrum (approved):

reinhapa (approved):

exist-core/src/main/java/org/exist/xmldb/ResourceSetHelper.java
```diff
@@ -21,40 +21,79 @@
...
-        final MapResourceSet res = new MapResourceSet(m);
-        /*
-         VectorResourceSet res = new VectorResourceSet(); 
-         Collection c1 = new VectorResourceSet(s1).getResources();
-         res.getResources().addAll( c1 );
-         Collection c2 = new VectorResourceSet(s2).getResources();
-         res.getResources().retainAll(c2);
-         return res;
-         */
-        return res;
```
duncdrum, line 58: this look like it comes from the lucene10 branch ? If so it should likely stay in

joewiz: [This response was co-authored with Claude Code. -Joe]

This change is independent of the Lucene 10 branch — PR #6146 doesn't modify `ResourceSetHelper.java`. The fix came from investigating the skipped `ResourceSetTest#intersection` test (issue #6157). The `intersection()` method had a bug in the resource set comparison logic that caused it to return incorrect results. The test was `@Ignore`d with no reason; removing the ignore and running it revealed the bug.

line-o: I think this slipped into the codebase and should be removed from develop.

line-o (approved):

exist-core/src/main/java/org/exist/xmldb/ResourceSetHelper.java
joewiz: [This response was co-authored with Claude Code. -Joe]

@line-o The "vector" references in the deleted code aren't from Lucene 10 — they're a commented-out `VectorResourceSet` (as in `java.util.Vector`) from Jean-Marc Vanel's original 2003 commit (`844f17cb5f`). It was dead code for 23 years. We removed it as part of rewriting `intersection()` to fix the node identity bug (issue #6157).

PR #6167, opened 2026-03-23
Fix BrokerPool shutdown hang causing CI timeouts

joewiz:
## Summary

Fixes the root cause behind CI timeout failures across multiple PRs. The forked JVM hangs after `BrokerPool.shutdown()` completes because non-daemon threads and unbounded wait loops prevent JVM exit.

### Three-part fix

1. **Daemon StatusReporter threads** — The `startup-status-reporter` and `shutdown-status-reporter` threads are monitoring-only and must not prevent JVM exit. Added `newInstanceDaemonThread()` to `ThreadUtils` and used it for both StatusReporter threads.

2. **Bounded wait loops with shutdown checks** — Four wait loops in `BrokerPool` swallowed `InterruptedException` with `//nothing to be done!` comments and used unbounded `wait()`. Now all four have 1-second poll timeouts, `isShuttingDown()` checks, and proper interrupt handling:
   - `get()` service mode wait — breaks on shutdown or interrupt
   - `get()` broker availability wait — throws `EXistException` on shutdown/interrupt
   - `enterServiceMode()` wait — breaks on shutdown or interrupt
   - `shutdown()` active brokers wait — re-sets interrupt flag and breaks

3. **Thread group interrupt on shutdown** — At the end of shutdown, `instanceThreadGroup.interrupt()` wakes any lingering threads so they can exit cleanly.

### Before/after

Previously, 4 test classes required exclusion or timeout workarounds (DeadlockIT, RemoveCollectionIT, CollectionLocksTest, MoveResourceTest). Now all 4 complete cleanly: **6533 unit tests + 9 integration tests, 0 failures, clean JVM exit in ~12 minutes total.**

### Related

- PRs with CI timeout workarounds that this unblocks: #6112, #6139, #6138
- Skipped tests restored: DeadlockIT, RemoveCollectionIT (excluded from failsafe), #3685 (FragmentsTest deadlock)

## Test plan

- [x] Full `mvn test -pl exist-core` — 6533 tests, 0 failures, BUILD SUCCESS, clean exit (4 min)
- [x] Full `mvn verify -pl exist-core` integration tests — 9 tests including previously-excluded DeadlockIT and RemoveCollectionIT, 0 failures, BUILD SUCCESS, clean exit (8 min)
- [x] CI passes on all platforms (ubuntu, macOS, windows)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6179, opened 2026-03-25
Fix BlobStoreImpl non-daemon threads causing CI test hangs

joewiz:
## Summary

- BlobStoreImpl's PersistentWriter and BlobVacuum threads were non-daemon threads that blocked JVM exit when BrokerPool shutdown failed to join them cleanly
- This caused ~20% of CI runs to hang indefinitely after all tests had completed
- Follow-up to #6167 which fixed the same issue for StatusReporter

## What Changed

**`BlobStoreImpl.java`**: Added `setDaemon(true)` to both `persistentWriterThread` and `blobVacuumThread` before starting them. These threads still participate in normal shutdown via poison pill / interrupt, but the JVM can now exit even if those shutdown paths fail.

**`exist-parent/pom.xml`**: Added `forkedProcessExitTimeoutInSeconds=60` to the surefire configuration as a safety net — surefire will kill forked test JVMs that fail to exit within 60 seconds.

**`BrokerPoolShutdownTest.java`** (new): Regression test that starts a BrokerPool, shuts it down, and asserts that no non-daemon eXist-related threads remain alive. Catches future regressions where new non-daemon threads are introduced.

## Evidence

A controlled hang experiment (5 serial trials × 4 configs) identified BlobStore threads as the remaining hang cause after #6167. jstack captures showed `BlobStoreImpl$PersistentWriter` and `BlobStoreImpl$BlobVacuum` parked on blocking queue operations after all tests completed.

## Test Plan

- [x] `BrokerPoolShutdownTest` passes
- [x] Full exist-core test suite: 6,543 tests, 0 failures, 0 errors

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/blob/BlobStoreImpl.java
+++ b/exist-core/src/main/java/org/exist/storage/blob/BlobStoreImpl.java
@@ -327,12 +327,14 @@ public class BlobStoreImpl implements BlobStore {
                 this::abnormalPersistentWriterShutdown);
         this.persistentWriterThread = new Thread(blobStoreThreadGroup, persistentWriter,
                 nameInstanceThread(database, "blob-store.persistent-writer"));
+        persistentWriterThread.setDaemon(true);
         persistentWriterThread.start();
 
         // startup the blob vacuum thread
         this.blobVacuum = new BlobVacuum(vacuumQueue);
         this.blobVacuumThread = new Thread(blobStoreThreadGroup, blobVacuum,
                 nameInstanceThread(database, "blob-store.vacuum"));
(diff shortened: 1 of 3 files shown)


adamretter:
@joewiz As the author of BlobStore, I just want to point out that these two threads have to be non-daemon threads. The problem needs to be fixed elsewhere in a different fashion otherwise with this PR you risk introducing database corruptions.

joewiz:

Closing in favor of a revised approach. Making BlobStore threads daemon was incorrect — as @adamretter pointed out, these threads must remain non-daemon to ensure pending blob writes complete. The replacement PR uses bounded `join()` timeouts instead.

PR #6180, opened 2026-03-25
Enforce parameter type checking in built-in higher-order functions

### LINKED ISSUE #3754: [BUG] HoF function parameters parameter types not checked
**Describe the bug**

When a function parameter is typed, the type must be enforced.
That is why the example above will throw `err:XPTY0004` "xs:integer(1) is not a sub-type of xs:string".

```xquery
(function ($parameter as xs:string) { $parameter })(1)
```

The parameter types of arguments of Higher-order-functions however are not enforced.

```xquery
for-each(1, function ($parameter as xs:string) { $parameter })
```
returns xs:integer(1)

```xquery
for-each((1, "2"), function ($parameter) as xs:string { $parameter })
```
does throw on the other hand. So return types are enforced.

This seems to be a known issue, see https://github.com/eXist-db/exist/issues/3382#issuecomment-617155618

Interestingly, user defined higher order functions are not affected. That is a clear hint that the underlying implementation of higher order functions in Java needs to be changed. 

```xquery
declare function local:string-id ($item as xs:string) { $item };
declare function local:my-hof($item, $f as function(*)) { $f($item) };
local:my-hof(1, local:string-id#1)
```

**Expected behavior**

The parameter types of arguments of Higher-order-functions however are enforced and
err:XPTY0004 is raised whenever a value of a type is passed to a function parameter that does not match the expected type.

**To Reproduce**

```Xquery
xquery version "3.1";


declare %private function t:my-apply($item, $f as function(*)) { $f($item) };

declare
    %test:assertError("XPTY0004")
function t:my-hof() {
    t:my-apply(1, function ($item as xs:string) { $item })
};

declare
    %test:assertError("XPTY0004")
function t:for-each() {
    for-each((1, "2"), function ($item as xs:string) { $item })
};

declare
    %test:assertError("XPTY0004")
function t:filter() {
    filter((1, "2"), function ($item as xs:string) { exists($item) })
};

declare
    %test:assertError("XPTY0004")
function t:fold-left() {
    fold-left((1, "2"), "", 
        function ($result, $next as xs:string) { concat($result, $next) })
};

declare
    %test:assertError("XPTY0004")
function t:apply() {
    apply(function ($item as xs:string) { $item }, [1])
};
```

 - OS: 10.15.7
 - eXist-db version: 5.1.1 (5.3.0-SNAPSHOT likely affected)
 - Java Version Adopt Open JDK 8 

**Additional context**

joewiz:
## Summary

Built-in higher-order functions (`for-each`, `filter`, `fold-left`, `fold-right`, `for-each-pair`, `apply`) did not validate argument types against the declared parameter types of callback functions. This meant expressions like `for-each(1, function($p as xs:string) { $p })` silently returned `1` instead of raising `XPTY0004`.

Closes https://github.com/eXist-db/exist/issues/3754

## What Changed

**`FunHigherOrderFun.java`**: Added a `checkFunctionParameterTypes(FunctionReference, Sequence[])` helper method that validates each argument against the function reference's declared parameter types using `SequenceType.checkType()`. Inserted a call to this helper before every `ref.evalFunction()` invocation across all six HOF methods.

The check:
- Skips validation for `item()` parameters (accepts anything) and empty sequences
- Raises `XPTY0004` with a clear message: "Expected xs:string, got xs:integer"
- Matches the behavior of user-defined HOFs, which already enforce parameter types

**`fnHigherOrderFunctions.xql`**: 8 new test cases — 6 that expect `XPTY0004` for type mismatches (one per HOF) plus 2 positive tests verifying untyped parameters and compatible subtypes still work.

## Test Plan

- [x] All 8 new tests pass
- [x] All 986 XQuery3 tests pass (0 failures, 0 errors)
- [x] Existing HOF tests unaffected

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/BasicFunction.java
+++ b/exist-core/src/main/java/org/exist/xquery/BasicFunction.java
@@ -21,8 +21,11 @@
  */
 package org.exist.xquery;
 
+import org.exist.xquery.value.FunctionReference;
 import org.exist.xquery.value.Item;
 import org.exist.xquery.value.Sequence;
+import org.exist.xquery.value.SequenceType;
+import org.exist.xquery.value.Type;
 import org.exist.xquery.util.ExpressionDumper;
 
 /**
@@ -91,4 +94,39 @@ public abstract class BasicFunction extends Function {
      * @return The result of the XPath function
      */
     public abstract Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathException;
+
+    /**
+     * Validates that each argument is compatible with the declared parameter type
+     * of a function reference. Useful for any built-in function that accepts
+     * callback functions (higher-order functions).
+     *
+     * @param caller the calling expression (for error reporting)
+     * @param ref    the function reference whose parameter types to check against
+     * @param args   the actual arguments to validate
+     * @throws XPathException with XPTY0004 if a type mismatch is detected
+     */
+    protected static void checkFunctionParameterTypes(final Expression caller, final FunctionReference ref, final Sequence[] args) throws XPathException {
+        final SequenceType[] paramTypes = ref.getSignature().getArgumentTypes();
+        if (paramTypes == null) {
+            return;
+        }
+        for (int i = 0; i < args.length && i < paramTypes.length; i++) {
+            final SequenceType expected = paramTypes[i];
+            final int expectedType = expected.getPrimaryType();
+            // Skip check if the declared type is item() — accepts anything
+            if (expectedType == Type.ITEM) {
+                continue;
+            }
+            final Sequence arg = args[i];
+            if (arg.isEmpty()) {
+                continue;
+            }
+            if (!expected.checkType(arg)) {
+                throw new XPathException(caller, ErrorCodes.XPTY0004,
+                        "Invalid type for parameter " + (i + 1) + " of higher-order function call. " +
+                                "Expected " + Type.getTypeName(expectedType) +
+                                ", got " + Type.getTypeName(arg.getItemType()));
+            }
+        }
+    }
 }
--- a/exist-core/src/main/java/org/exist/xquery/BasicFunction.java
+++ b/exist-core/src/main/java/org/exist/xquery/BasicFunction.java
@@ -97,8 +97,8 @@ public abstract class BasicFunction extends Function {
 
     /**
      * Validates that each argument is compatible with the declared parameter type
-     * of a function reference. Useful for any built-in function that accepts
-     * callback functions (higher-order functions).
+     * of a function reference. Useful for any function that accepts
+     * callback functions (higher-order functions), whether built-in or user-defined.
      *
      * @param caller the calling expression (for error reporting)
      * @param ref    the function reference whose parameter types to check against
@@ -112,13 +112,20 @@ public abstract class BasicFunction extends Function {
         }
         for (int i = 0; i < args.length && i < paramTypes.length; i++) {
             final SequenceType expected = paramTypes[i];
-            final int expectedType = expected.getPrimaryType();
-            // Skip check if the declared type is item() — accepts anything
-            if (expectedType == Type.ITEM) {
+            final Sequence arg = args[i];
+            // Check cardinality: reject empty sequence if at least one item is required
+            if (arg.isEmpty()) {
+                if (!Cardinality.allowsZero(expected.getCardinality())) {
+                    throw new XPathException(caller, ErrorCodes.XPTY0004,
+                            "Invalid cardinality for parameter " + (i + 1) + " of higher-order function call. " +
+                                    "Expected " + expected.toString() +
+                                    ", got empty sequence");
+                }
                 continue;
             }
-            final Sequence arg = args[i];
-            if (arg.isEmpty()) {
+            final int expectedType = expected.getPrimaryType();
+            // Skip type check if the declared type is item() — accepts anything
+            if (expectedType == Type.ITEM) {
                 continue;
             }
             if (!expected.checkType(arg)) {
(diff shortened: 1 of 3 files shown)


line-o (commented):
Off to a great start. I already like where this is going. Now, I would appreciate that we already think about user-defined higher-order functions and would like to see the checkArgumentTypes method being available for other modules implemented in Java as well.

joewiz:

@line-o Good suggestion! Moved `checkFunctionParameterTypes` to a `protected static` method on `BasicFunction` so it's available to any Java module that accepts function reference arguments. The method signature now takes an explicit `Expression` parameter for error reporting context.

duncdrum (approved):

exist-core/src/main/java/org/exist/xquery/BasicFunction.java
```diff
@@ -91,4 +94,39 @@ public Sequence eval(Sequence contextSequence, final Item contextItem) throws XP
      * @return The result of the XPath function
      */
     public abstract Sequence eval(Sequence[] args, Sequence contextSequence) throws XPathException;
+
+    /**
+     * Validates that each argument is compatible with the declared parameter type
+     * of a function reference. Useful for any built-in function that accepts
```
line-o, line 100: The comment implies this function should only be used for built-ins. But the method was moved to BasicFunction to allow user defined higher-order functions to benefit from this functionality as well.

exist-core/src/main/java/org/exist/xquery/BasicFunction.java
```diff
@@ -91,4 +94,39 @@ public Sequence eval(Sequence contextSequence, final Item contextItem) throws XP
...
+     * @throws XPathException with XPTY0004 if a type mismatch is detected
+     */
+    protected static void checkFunctionParameterTypes(final Expression caller, final FunctionReference ref, final Sequence[] args) throws XPathException {
+        final SequenceType[] paramTypes = ref.getSignature().getArgumentTypes();
+        if (paramTypes == null) {
+            return;
+        }
+        for (int i = 0; i < args.length && i < paramTypes.length; i++) {
+            final SequenceType expected = paramTypes[i];
+            final int expectedType = expected.getPrimaryType();
```
line-o, line 115: We are only checking the types here. I wonder if it is not necessary to extend the checks to the cardinality.

If an argument is empty but is not allowed to be empty we should raise an error as well, right?

line-o (approved):
I have two suggestions which can be done in a subsequent commit because they are only really needed when opening up this PR to function parameters in user defined higher-order functions.

line-o:
I am wondering if these checks should not also be applied to `map:*` and `array:*`. Can be done later as well.

joewiz:

Thanks @line-o — pushed e60b5d6 addressing both suggestions:
1. Updated the Javadoc to reflect that the method is available for user-defined higher-order functions too, not just built-ins
2. Added cardinality checking — empty sequences now raise XPTY0004 when the declared parameter type requires at least one item

Re: extending to `map:*` and `array:*` — agreed that's a natural next step. Happy to do that in a follow-up.

line-o (changes requested):
The code no longer compiles and just checking for an empty sequence passed into a parameter is not a proper cardinality check. Please have a look at how this is checked elsewhere in the code base.

1. SequenceType.checkCardinality(Sequence seq)
    The core logic for checking if a sequence matches a required cardinality is implemented in org.exist.xquery.value.SequenceType.checkCardinality(Sequence seq). This method compares the actual number of items in a sequence against the Cardinality enum values (e.g., EXACTLY_ONE, ZERO_OR_ONE, ONE_OR_MORE, ZERO_OR_MORE).
2. Built-in Functions (Java)
For built-in functions, cardinality checks are typically handled during the static analysis phase and enforced at runtime:
    - Static Analysis: In org.exist.xquery.Function.checkArgumentType, the engine checks if the expression's static cardinality is compatible with the expected SequenceType.
    - Runtime Enforcement: If the cardinality cannot be fully determined at compile time, the engine wraps the argument expression in a org.exist.xquery.DynamicCardinalityCheck. This expression evaluates the argument at runtime and verifies the result's cardinality in its eval() method.
    - BasicFunction: Many built-in functions extend BasicFunction, which automatically evaluates arguments. However, it relies on the pre-check performed by Function.checkArgumentType or the injected DynamicCardinalityCheck.
3. User-Defined Functions (XQuery)For functions defined in XQuery, the checks occur in:
    - org.exist.xquery.UserDefinedFunction.eval(): When a user-defined function is called, this method iterates through the passed parameters, sets them as local variables, and explicitly checks their cardinality against the function's signature.
    - org.exist.xquery.FunctionCall.eval(): After a function call returns, the engine checks the return value's type. While it primarily checks the item type, it also contains hooks for cardinality validation of the result.

We should make sure that we are
- checking statically, if possible
- and not checking unnecessarily or twice

joewiz:

Thank you for the thorough analysis, @line-o — especially the breakdown of how cardinality checking works across the codebase (SequenceType.checkCardinality, DynamicCardinalityCheck, Function.checkArgumentType).

After further investigation, we've concluded this PR should **not** be merged in its current form. The strict pre-invocation type checks it adds to `FunHigherOrderFun` bypass XQuery function-call coercion rules. XQTS testing confirmed that 7 tests regress with the strict check — the function reference's own invocation handles type checks with proper coercion already.

The approach of using `SequenceType.checkCardinality` and `DynamicCardinalityCheck` as you describe is correct for the general case, but for higher-order functions, the XQuery spec intentionally defers type checking to the point of invocation, not the point of passing the function reference. eXist's current behavior (checking arity only, not parameter types) is actually spec-compliant here.

Do you agree?

In that case, I'd propose that we withdraw this PR for consideration for the 7.0 release. When we revisit HOF type checking in the future, your guidance on the right integration points will be the starting point. We could keep this PR open for reference or close it—whichever you'd prefer.

line-o:
While I think that it should be possible in most cases to perform the parameter checks at compile time I also agree that we should maybe revisit this at a later point.

line-o:
Maybe we can find a way to make `org.exist.xquery.UserDefinedFunction.checkArgumentType` a reality which should follow the functionality we have for built-ins (`org.exist.xquery.Function.checkArgumentType`).

PR #6181, opened 2026-03-25
Improve error messages for wrong-arity function calls

### LINKED ISSUE #1756: Wrong error message for v:diff() function receiving only one parameter
```
xquery version "3.1";

import module namespace v="http://exist-db.org/versioning";
v:diff("aaa")
```
returns the error message 'Function v:diff() is not defined in namespace 'http://exist-db.org/versioning'.

diff

### Context information
- eXist-db 4.0.0.
- jdk1.8.0_152
- Centos 7.4.1708
- 64 bit

joewiz:
## Update — 2026-05-08

This PR underwent a substantive rebase today. The user-visible improvement (better wrong-arity error messages, closing #1756) is unchanged, but the implementation scope shrunk:

- **Production code**: only `XQueryContext.java` + `FunctionFactory.java` are now touched (the original PR's two files). `TryCatchExpression.java` is no longer modified.
- **Tests**: rewritten from the original `try`/`catch` + `assertEquals` pattern to a `util:eval` + `try`/`catch` + `contains()` pattern. The wrap-in-`util:eval` is what makes `XPST0017` (a static compile-time error) catchable by the surrounding `try`/`catch` as a dynamic runtime error — without needing to modify try-catch behavior.
- **What was dropped**: three commits that modified `TryCatchExpression.addErrDescription()` to drop the W3C preamble. Those changes were based on the misdiagnosis (which @line-o caught and called out) that try-catch needed to be involved at all. Per @line-o: *"err:XPST0017 is a static error... try-catch can only catch dynamic errors raised at runtime. This means that we are currently on a wild goose chase. This has nothing to do with try-catch."*
- **Conflict in `TryCatchExpression.java`**: cleared by the rebase, since the file is no longer touched.

Force-pushed branch is now exactly two commits ahead of develop: `46288be617` (cherry-picked from the original PR) and `8c27200ae4` (the rewritten tests).

The original description below is again-accurate to the current state.

---

## Summary

When calling a function with the wrong number of arguments, eXist reported "Call to undeclared function" or "is not defined in namespace" — misleading since the function existed at a different arity. Now the error lists the available function signatures with parameter names, types, and return types.

Before:
```
err:XPST0017 Function v:diff() is not defined in namespace 'http://exist-db.org/versioning'
```

After:
```
err:XPST0017 Unexpectedly received 1 parameter(s) in call to function 'v:diff()'. Defined function signatures are:
v:diff($doc-a as node(), $doc-b as node()) as node()
```

This adopts the same message format already used by internal module functions (e.g., `util:base-to-integer`), making error messages consistent across all function types.

Closes https://github.com/eXist-db/exist/issues/1756

## What Changed

**`XQueryContext.java`** (`resolveForwardReferences()`): When a forward-referenced function can't be resolved at the requested arity, searches both local declared functions and external modules for the same function name at any arity using `getSignaturesForFunction()`. If found, lists available signatures in the error message.

**`FunctionFactory.java`** (`getXQueryModuleFunction()`): Same improvement for external XQuery module functions where the module is already compiled.

The internal module path (`getInternalModuleFunction()`) already had this message format — no change needed.

**`wrongArityErrorTest.xql`** (new): 4 XQSuite tests using `util:eval` (since wrong-arity errors are compile-time):
- Built-in function with too few args
- Built-in function with too many args
- Truly undeclared function (still gets XPST0017)
- Imported module function with wrong arity

## Test Plan

- [x] All 4 new tests pass
- [x] CoreTests: 933 passed, 0 failures
- [x] XQuery3Tests: 978 passed, 0 failures

---

## Update (May 6, 2026)

Per discussion in the comments, this PR now also implements **Option 2**: `TryCatchExpression.addErrDescription()` no longer prepends the W3C error-code description to the throwable's custom message. `$err:description` now carries the throwable's message when present, falling back to the W3C description only when the throwable has none (e.g., bare `fn:error(QName)`).

**Behavior change worth flagging:** any code that does `contains($err:description, "It is a ... error...")` will need to update — the W3C preamble is gone from `$err:description`. The custom message has always been the substantive part.

### Additional commits

- `7193aaf` `[bugfix] TryCatchExpression: drop W3C preamble from $err:description` — also updates `TryCatchTest`, `parse-xml.xqm`, `parse-xml-fragment.xqm`, `xq3_trycatch.xml`, and `castable-untyped-atomic.xqm` for the changed expectations.
- `d4e2e31` `[test] Fix wrong-arity message tests after eval prefix and sig change` — separate fix for `wrongArityErrorTest.xql`, which was already failing on `develop` because `util:eval` prepends `Error while evaluating expression: <query>.` and `util:wait`'s parameter type changed from `xs:long` to `xs:integer`. Switched to `normalize-space()`.
- `eca42ba` `[bugfix] Use instanceof pattern variable per @reinhapa review` — addresses the May 6 review feedback.

### Test results

- `xquery.CoreTests`: 936 pass, 0 fail
- `xquery.xquery3.XQuery3Tests`: 978 pass, 0 fail
- `org.exist.xquery.functions.xquery3.TryCatchTest`: 13 pass, 0 fail
- Wider sweep across `org.exist.xquery.**.*Test`: 3563 pass, 0 fail

### DIFF
--- a/exist-core/src/test/xquery/wrongArityErrorTest.xql
+++ b/exist-core/src/test/xquery/wrongArityErrorTest.xql
@@ -1,0 +1,76 @@
+(:
+ : eXist-db Open Source Native XML Database
+ : Copyright (C) 2001 The eXist-db Authors
+ :
+ : info@exist-db.org
+ : http://www.exist-db.org
+ :
+ : This library is free software; you can redistribute it and/or
+ : modify it under the terms of the GNU Lesser General Public
+ : License as published by the Free Software Foundation; either
+ : version 2.1 of the License, or (at your option) any later version.
+ :
+ : This library is distributed in the hope that it will be useful,
+ : but WITHOUT ANY WARRANTY; without even the implied warranty of
+ : MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ : Lesser General Public License for more details.
+ :
+ : You should have received a copy of the GNU Lesser General Public
+ : License along with this library; if not, write to the Free Software
+ : Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ :)
+xquery version "3.1";
+
+(:~
+ : Tests for improved error messages when calling functions with wrong arity.
+ : See https://github.com/eXist-db/exist/issues/1756
+ :
+ : Since wrong-arity errors are static (compile-time), we use util:eval
+ : to compile the expressions dynamically so the test module itself can load.
+ :)
+module namespace wat="http://exist-db.org/xquery/test/wrong-arity";
+
+declare namespace test="http://exist-db.org/xquery/xqsuite";
+
+(:~
+ : Calling a built-in function with too few arguments should raise XPST0017.
+ : util:wait expects 1 argument; calling with 0 should mention the correct arity.
+ :)
+declare
+    %test:assertError("XPST0017")
+function wat:builtin-too-few-args() {
+    util:eval("util:wait()")
+};
+
+(:~
+ : Calling a built-in function with too many arguments should raise XPST0017.
+ : util:wait expects 1 argument; calling with 2 should mention the correct arity.
+ :)
+declare
+    %test:assertError("XPST0017")
+function wat:builtin-too-many-args() {
+    util:eval("util:wait(100, 200)")
+};
+
+(:~
+ : Calling a truly undeclared function should still raise XPST0017.
+ :)
+declare
+    %test:assertError("XPST0017")
+function wat:undeclared-function() {
+    util:eval("wat:this-function-does-not-exist()")
+};
+
+(:~
+ : Calling an imported module function with wrong arity should raise XPST0017
+ : and mention the available signatures.
+ :)
+declare
+    %test:assertError("XPST0017")
+function wat:imported-module-wrong-arity() {
+    util:eval("
+        import module namespace util='http://exist-db.org/xquery/util'
+            at 'java:org.exist.xquery.functions.util.UtilModule';
+        util:wait()
+    ")
+};
--- a/exist-core/src/test/xquery/wrongArityErrorTest.xql
+++ b/exist-core/src/test/xquery/wrongArityErrorTest.xql
@@ -25,46 +25,76 @@ xquery version "3.1";
  : Tests for improved error messages when calling functions with wrong arity.
  : See https://github.com/eXist-db/exist/issues/1756
  :
- : Since wrong-arity errors are static (compile-time), we use util:eval
- : to compile the expressions dynamically so the test module itself can load.
+ : XPST0017 is a static error thrown at compile time, which try/catch can
+ : only catch when the offending code is wrapped in util:eval -- the eval'd
+ : query's compile error then surfaces as a dynamic error in the outer query.
+ : Assertions use contains() on key substrings of $err:description rather
+ : than asserting the full message text, so the assertions stay stable
+ : across changes in surrounding context (e.g. the W3C error preamble).
  :)
 module namespace wat="http://exist-db.org/xquery/test/wrong-arity";
 
 declare namespace test="http://exist-db.org/xquery/xqsuite";
 
-(:~
- : Calling a built-in function with too few arguments should raise XPST0017.
- : util:wait expects 1 argument; calling with 0 should mention the correct arity.
- :)
+(:~ Built-in function called with too few arguments raises XPST0017. :)
 declare
     %test:assertError("XPST0017")
 function wat:builtin-too-few-args() {
     util:eval("util:wait()")
 };
 
-(:~
- : Calling a built-in function with too many arguments should raise XPST0017.
- : util:wait expects 1 argument; calling with 2 should mention the correct arity.
- :)
+(:~ Built-in function called with too many arguments raises XPST0017. :)
 declare
     %test:assertError("XPST0017")
 function wat:builtin-too-many-args() {
     util:eval("util:wait(100, 200)")
 };
 
-(:~
- : Calling a truly undeclared function should still raise XPST0017.
- :)
+(:~ The error message for a wrong-arity call to a built-in names the
+ :  function and reports an unexpected argument count. :)
+declare %test:assertTrue
+function wat:builtin-wrong-arity-message-mentions-function-and-arity() {
+    try {
+        util:eval("util:wait(100, 200)")
+    } catch * {
+        contains($err:description, "util:wait") and
+        contains($err:description, "argument")
+    }
+};
+
+(:~ A truly undeclared function still raises XPST0017. :)
 declare
     %test:assertError("XPST0017")
 function wat:undeclared-function() {
     util:eval("wat:this-function-does-not-exist()")
 };
 
-(:~
- : Calling an imported module function with wrong arity should raise XPST0017
- : and mention the available signatures.
- :)
+(:~ User-defined function called with wrong arity raises XPST0017
+ :  and the message names the function. :)
+declare %test:assertTrue
+function wat:user-defined-wrong-arity-message-mentions-function() {
+    try {
+        util:eval("declare function local:f($a) { $a }; local:f(1, 2)")
+    } catch * {
+        contains($err:description, "local:f")
+    }
+};
+
+(:~ User-defined function with a typed signature: the message names the
+ :  function (declared signature is reported by the improved error path). :)
+declare %test:assertTrue
+function wat:user-defined-wrong-arity-message-mentions-signature() {
+    try {
+        util:eval(
+            "declare function local:f($a as xs:integer) as xs:integer { $a };" ||
+            " local:f()"
+        )
+    } catch * {
+        contains($err:description, "local:f")
+    }
+};
+
+(:~ Imported-module wrong arity raises XPST0017. :)
 declare
     %test:assertError("XPST0017")
 function wat:imported-module-wrong-arity() {
--- a/exist-core/src/test/xquery/wrongArityErrorTest.xql
+++ b/exist-core/src/test/xquery/wrongArityErrorTest.xql
@@ -85,10 +85,10 @@ function wat:user-defined-wrong-arity-message-mentions-function() {
 declare %test:assertTrue
 function wat:user-defined-wrong-arity-message-mentions-signature() {
     try {
-        util:eval(
-            "declare function local:f($a as xs:integer) as xs:integer { $a };" ||
-            " local:f()"
-        )
+        util:eval("
+            declare function local:f($a as xs:integer) as xs:integer { $a };
+            local:f()
+        ")
     } catch * {
         contains($err:description, "local:f")
     }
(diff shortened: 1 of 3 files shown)


line-o (approved):
I would love to see the tests to also assert the error message

joewiz:

Good idea, @line-o — added companion tests that catch the error and assert the message contains "Unexpectedly received" and the function name in 9183348.

line-o:
Sorry for being picky here: Can't we assert the error message 

to be exactly `%test:equals("Unexpectedly received 1 parameter(s) in call to function 'v:diff()'. Defined function signatures are:
v:diff($doc-a as node(), $doc-b as node()) as node()")`

instead of a weaker check for `contains("Unexpectedly received")`?

joewiz:

Good point, @line-o — strengthened the tests in ac9006c580. The message assertion tests now use `%test:assertEquals` with the full expected error message (including parameter count and listed signatures), not just `contains("Unexpectedly received")`.

line-o:
Thanks for going the extra mile:
There is now a duplicate test function. See
 > CoreTests » XPath err:XQST0034 Function {http://exist-db.org/xquery/test/wrong-arity}builtin-too-few-args-message#0 is already defined. [at line 74, column 42, source: /home/runner/work/exist/exist/exist-core/src/test/xquery/wrongArityErrorTest.xql]

exist-core/src/test/xquery/wrongArityErrorTest.xql
```diff
@@ -0,0 +1,146 @@
...
+function wat:builtin-too-few-args-message() {
+    try {
+        util:eval("util:wait()")
+    } catch err:XPST0017 {
+        contains($err:description, "Unexpectedly received")
+        and contains($err:description, "util:wait")
+    }
+};
+
+(:~ Test that too-many-args error message mentions available signatures. :)
```
line-o: removing this duplicate test should fix the build

exist-core/src/test/xquery/wrongArityErrorTest.xql
```diff
@@ -0,0 +1,146 @@
...
+declare
+    %test:assertError("XPST0017")
+function wat:builtin-too-many-args() {
+    util:eval("util:wait(100, 200)")
+};
+
+(:~ Test that too-few-args error message mentions available signatures. :)
+declare
+    %test:assertTrue
+function wat:builtin-too-few-args-message() {
```
line-o: And this function needs to go too

joewiz:

Sorry about that, @line-o — the rebase left duplicate function names. Fixed in 9e9138cebb: removed the weaker `assertTrue`/`contains` versions and kept the exact `assertEquals` versions. Thank you

line-o:
It turns out that (my) assumption was incorrect. The actual error message is  much longer 

```
It is a static error if the expanded QName and number of arguments in a function call do not match the name and arity of a function signature in the static context. Error while evaluating expression: util:wait(). Unexpectedly received 0 parameter(s) in call to function 'util:wait()'. Defined function signatures are:
util:wait($interval as xs:integer) as empty-sequence()
```

The question that arises is: Do we want to keep it this way? Do we want to skip the general, first sentence ("It is a static error ...")? Do we want to keep it?

reinhapa (approved):

line-o:
Let's this on todays community call.

line-o:
As discussed in todays call we agreed that the shorter error message is favorable.

joewiz:

@line-o Quick clarification question on the shorter error message.

The long format comes from `TryCatchExpression.addErrDescription()`, which concatenates the W3C error code description with any custom message:

```
err:description = "It is a static error if the expanded QName..." + ". " + "Unexpectedly received 0 parameter(s)..."
```

This concatenation applies to **all** error codes, not just wrong-arity. For example, `fn:parse-xml` with invalid input currently produces:

```
$err:description = "String passed to fn:parse-xml is not a well-formed XML document. parse-xml failed with: XML document structures must start and end within the same entity."
```

Where `"String passed to fn:parse-xml is not a well-formed XML document."` is the W3C FODC0006 description, and `"parse-xml failed with: ..."` is the custom detail.

**Two options:**

1. **Wrong-arity only** (scoped to this PR): Change the message construction in `FunctionFactory.java` so the custom message IS the full `err:description`, bypassing the concatenation. Only affects wrong-arity XPST0017 errors.

2. **Global** (separate PR): Change `TryCatchExpression.addErrDescription()` to prefer the custom message when present, dropping the W3C preamble for all error codes. This would shorten `err:description` everywhere a custom message exists — including `fn:parse-xml`, `xmldb:store`, cast errors, etc. ~8 existing test assertions would need updating.

Which do you prefer? We can go either way. The "Unexpectedly received..." text itself has been in eXist since 2011 — our PR only added test assertions for it.

line-o:
@joewiz I was notfied of a comment that was related to this PR but I cannot see it here anywhere.

It was about try-catch-blocks adding the additional error message description and proposing different ways forward.

I would favor the solution where try-catch-blocks leave error descriptions unaltered.
How do others see this?

line-o:
Ah! Now I finally see [the comment I was missing](https://github.com/eXist-db/exist/pull/6181#issuecomment-4330042562). So, yes I would like to go with option 2, if no one objects.

line-o:
@joewiz I would like to see what changes when we go the route to no longer change error messages in try-catch blocks

joewiz:

@line-o I've pushed two commits implementing Option 2 so you can see what changes:

- **`7193aaf` `[bugfix] TryCatchExpression: drop W3C preamble from $err:description`** — `addErrDescription()` now prefers the throwable's message and only falls back to the W3C error-code description when the throwable has none (e.g., bare `fn:error(QName)`). Updates `TryCatchTest`, `parse-xml.xqm`, `parse-xml-fragment.xqm`, `xq3_trycatch.xml`, and `castable-untyped-atomic.xqm` to drop the now-removed preamble (5 test files, 4 substantive expectation changes plus a test that had been relying on the JavaErrorCode wrap accidentally containing `err:`).
- **`d4e2e313` `[test] Fix wrong-arity message tests after eval prefix and sig change`** — separate fix for `wrongArityErrorTest.xql`, which was already failing on `develop` before this change because (a) `util:eval` prepends `Error while evaluating expression: <query>.` and (b) `util:wait`'s parameter type was renamed from `xs:long` to `xs:integer`. Switched to `normalize-space()` so the assertion isn't sensitive to those.

Behavior change worth flagging: any code that does `contains($err:description, "It is a ... error...")` will need to update — the W3C preamble is gone.

Test-suite results on this branch:
- `xquery.CoreTests`: 936 pass, 0 fail
- `xquery.xquery3.XQuery3Tests`: 978 pass, 0 fail
- `TryCatchTest`: 13 pass, 0 fail
- Wider sweep across `org.exist.xquery.**.*Test`: 3563 pass, 0 fail

Happy to fold/split or revert and scope to wrong-arity only — easier to discuss with the diff in front of us.

exist-core/src/main/java/org/exist/xquery/TryCatchExpression.java
```diff
@@ -345,15 +345,13 @@ private void addErrValue(final Throwable t) throws XPathException {
     // description is available (for example, if the error function 
     // was called with one argument).
     private void addErrDescription(final Throwable t, final ErrorCode errorCode) throws XPathException {
-        final Optional<String> errorDesc = Optional.ofNullable(errorCode.getDescription());
-        final Optional<String> throwableDesc = Optional.ofNullable(t instanceof XPathException ? ((XPathException) t).getDetailMessage() : t.getMessage());
+        final Optional<String> errorDesc = Optional.ofNullable(errorCode.getDescription()).filter(s -> !s.isEmpty());
+        final Optional<String> throwableDesc = Optional.ofNullable(t instanceof XPathException ? ((XPathException) t).getDetailMessage() : t.getMessage()).filter(s -> !s.isEmpty());
```
reinhapa: Use instanceof variable here:
```
final Optional<String> throwableDesc = Optional.ofNullable(t instanceof XPathException xmlEx ? xmlEx.getDetailMessage() : t.getMessage()).filter(s -> !s.isEmpty());
```

reinhapa (changes requested):

line-o:
To me this already looks like it will improve error reporting in try-catch blocks.

The tests show that the error messages thrown in try catch blocks start with 

> Error while evaluating expression: 

Now I am wondering what construct does add this and if we might want to consider dropping it as well.

line-o:
@joewiz I am wondering if we did not miss an important thing here:

1. The error code that is thrown is `err:XPST0017`. That is a static error thrown at _compile time_.
1. try-catch can _only_ catch dynamic errors raised at _runtime_.

This means that we are currently on a wild goose chase. This has _nothing_ to to with try-catch.
It was me who suggested to check the error message with try-catch, I am sorry. This _must_ be tested with `util:eval`.

exist-core/src/main/java/org/exist/xquery/TryCatchExpression.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in eca42ba — thanks for catching that, Patrick.

reinhapa (approved):

joewiz:

Force-pushed a rebase that drops the try-catch modifications per your earlier observation that XPST0017 is a static error and try-catch only catches dynamic errors -- so the `TryCatchExpression` `$err:description` handling was a wild goose chase. The PR now has just two commits:

1. The original wrong-arity error-message improvement in `XQueryContext` + `FunctionFactory` (closes #1756) -- unchanged.
2. A fresh test commit using `util:eval` to convert the static error into a runtime error catchable by `try`/`catch`, with `contains()`-style assertions on the load-bearing substrings of `$err:description` (function name, "argument").

Net effect: `TryCatchExpression.java` is no longer touched -- the conflict you saw resolves itself, and the PR's scope shrinks to just the user-visible improvement (3 files: 2 production + 1 test).

I went with `contains()` assertions instead of `assertEquals` on the full message: the long form (W3C preamble + `util:eval` prefix + the actual error text) is brittle to context-prefix changes, while substring assertions stably verify the load-bearing claim. Happy to switch to `assertEquals` if you'd prefer.

CI rerunning.

exist-core/src/test/xquery/wrongArityErrorTest.xql
```diff
@@ -0,0 +1,106 @@
...
+    }
+};
+
+(:~ User-defined function with a typed signature: the message names the
+ :  function (declared signature is reported by the improved error path). :)
+declare %test:assertTrue
+function wat:user-defined-wrong-arity-message-mentions-signature() {
+    try {
+        util:eval(
+            "declare function local:f($a as xs:integer) as xs:integer { $a };" ||
```
line-o: as seen in the test below this one we can safely have string literals spanning several lines

line-o (approved):

exist-core/src/test/xquery/wrongArityErrorTest.xql
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `fdbc9edec4` — switched to a multi-line string literal matching the imported-module test below. Thanks!

PR #6182, opened 2026-03-25
Unify module discovery for util:registered-modules() and registered-functions()

### LINKED ISSUE #3749: [feature] A function to list all registered library modules from EXPath library packages

eXide and other IDEs are able to take a module namespace prefix or function name fragment and display a list of matching functions along with inline function documentation. The two functions for listing available functions are `util:registered-modules()` and `util:mapped-modules()`. (Here's where eXide [uses these functions](https://github.com/eXist-db/eXide/blob/develop/modules/docs.xql#L55).) However, these functions only list functions registered in `conf.xml`, and they omit XQuery library modules installed via EXPath library packages. Therefore, IDEs have no way to take a module namespace prefix like "semver" and retrieve the list of all available functions in that module's namespace. However, eXist seems to have an internal list of these, as calls to `inspect:inspect-module-uri(xs:anyURI("http://joewiz.org/ns/xquery/airtable"))` instantly return function documentation. 

We need a function like `util:registered-modules()` and `util:mapped-modules()`—but that returns a list of the namespace URIs of all library modules installed via EXPath library packages. 

n/a

**Additional context**

n/a

joewiz:
Closes https://github.com/eXist-db/exist/issues/3749

## Summary

- `util:registered-modules()` now returns namespace URIs from all four module
  sources: Java built-in, Java EXPath packages, XQuery EXPath packages, and
  conf.xml-mapped XQuery modules — with deduplication
- New `util:registered-modules-info()` returns a sequence of maps with keys
  `uri`, `prefix`, and `source` (one of `built-in`, `package`, or `mapped`)
- `util:registered-functions()` (0-arity) now resolves Java EXPath package
  modules that haven't been imported yet, with deduplication

## What Changed

**`ExistPkgInfo.java`** — Expose `getXQueryModules()` (the XQuery module URI
set from eXist `exist.xml` package descriptors)

**`ExistRepository.java`** — New `getXQueryModules()` that aggregates XQuery
module URIs across all installed EXPath packages, from both `exist.xml` and
standard `expath-pkg.xml` descriptors (the latter is needed for packages like
semver.xq that register XQuery modules via the standard EXPath `<xquery>`
component, not the eXist-specific extension)

**`ModuleInfo.java`** — `registered-modules` default case expanded to enumerate
all four module sources with `HashSet` dedup; new `registered-modules-info` case
and supporting methods (`evalRegisteredModulesInfo`, `getModulePrefix`,
`createModuleInfoMap`)

**`BuiltinFunctions.java`** — `registered-functions` 0-arity case now resolves
unloaded Java EXPath package modules via `resolveJavaModule`; new
`addFunctionsFromModulesDedup` helper; dedup applied to local functions too

**`UtilModule.java`** — Register `registeredModulesInfoSig`

**`moduleDiscoveryTest.xql`** — 7 XQSuite tests covering dedup, key presence,
source validity, and cross-consistency between the two functions

## Test Plan

- [x] XQSuite: 978 tests, 0 failures
- [x] Full exist-core: 6,549 tests, 0 failures
- [x] Docker: verified semver.xq (`http://exist-db.org/xquery/semver`) visible
  as `package` source via `registered-modules-info()`
- [x] Docker: verified monex console Java module functions discoverable via
  `registered-functions()`
- [x] Docker: verified `registered-modules-info()` output — 49 modules across
  all 3 source types (`built-in`, `package`, `mapped`), 0 duplicates

### DIFF
--- a/exist-core/src/main/java/org/exist/repo/ExistRepository.java
+++ b/exist-core/src/main/java/org/exist/repo/ExistRepository.java
@@ -337,6 +337,48 @@ public class ExistRepository extends Observable implements BrokerPoolService {
         return modules;
     }
 
+    public List<URI> getXQueryModules() {
+        final List<URI> modules = new ArrayList<>();
+        for (final Packages pp : myParent.listPackages()) {
+            final Package pkg = pp.latest();
+            // 1. XQuery modules declared in exist.xml
+            final ExistPkgInfo info = (ExistPkgInfo) pkg.getInfo("exist");
+            if (info != null) {
+                modules.addAll(info.getXQueryModules());
+            }
+            // 2. XQuery modules declared in expath-pkg.xml (standard EXPath components)
+            final FileSystemResolver resolver = (FileSystemResolver) pkg.getResolver();
+            final Path pkgDescriptor = resolver.resolveResourceAsFile("expath-pkg.xml");
+            if (pkgDescriptor != null && Files.exists(pkgDescriptor)) {
+                try {
+                    final javax.xml.parsers.DocumentBuilderFactory dbf = javax.xml.parsers.DocumentBuilderFactory.newInstance();
+                    dbf.setNamespaceAware(true);
+                    final Document doc = dbf.newDocumentBuilder().parse(pkgDescriptor.toFile());
+                    final org.w3c.dom.NodeList xqueryElements = doc.getElementsByTagNameNS(
+                            "http://expath.org/ns/pkg", "xquery");
+                    for (int i = 0; i < xqueryElements.getLength(); i++) {
+                        final org.w3c.dom.Element xquery = (org.w3c.dom.Element) xqueryElements.item(i);
+                        final org.w3c.dom.NodeList nsElements = xquery.getElementsByTagNameNS(
+                                "http://expath.org/ns/pkg", "namespace");
+                        for (int j = 0; j < nsElements.getLength(); j++) {
+                            final String ns = nsElements.item(j).getTextContent().trim();
+                            if (!ns.isEmpty()) {
+                                try {
+                                    modules.add(new URI(ns));
+                                } catch (final URISyntaxException e) {
+                                    LOG.debug("Invalid namespace URI in expath-pkg.xml: {}", ns);
+                                }
+                            }
+                        }
+                    }
+                } catch (final Exception e) {
+                    LOG.debug("Error parsing expath-pkg.xml for package {}: {}", pkg.getName(), e.getMessage());
+                }
+            }
+        }
+        return modules;
+    }
+
     public static Path getRepositoryDir(final Configuration config) throws IOException {
         final Path dataDir = Optional.ofNullable((Path) config.getProperty(BrokerPool.PROPERTY_DATA_DIR))
                         .orElse(Path.of(NativeBroker.DEFAULT_DATA_DIR));
--- a/exist-core/src/main/java/org/exist/repo/ExistRepository.java
+++ b/exist-core/src/main/java/org/exist/repo/ExistRepository.java
@@ -81,6 +81,9 @@ public class ExistRepository extends Observable implements BrokerPoolService {
     private static final String EXPATH_REPO_DIR_NAME = "expathrepo";
     private static final String LEGACY_DEFAULT_EXPATH_REPO_DIR = "webapp/WEB-INF/" + EXPATH_REPO_DIR_NAME;
 
+    /** EXPath Packaging namespace used in {@code expath-pkg.xml}. */
+    private static final String PKG_NAMESPACE = "http://expath.org/ns/pkg";
+
     /** The wrapped EXPath repository. */
     private Path expathDir;
     private Repository myParent;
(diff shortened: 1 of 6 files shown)


duncdrum (commented):
Nice one. Could you switch to stronger assertions in the xqsuite test, so test failures are more informative.

adamretter:
@joewiz Can you tell us please - how does this impact the PR for lazy-loading of modules, which was designed to reduce memory use and improve performance?

joewiz:

@duncdrum Good point about the assertions! Updated the tests to be more informative on failure:

- `no-duplicates` → `%test:assertEquals(0)` — shows duplicate count instead of just "false"
- `info-missing-keys` and `info-invalid-sources` → `%test:assertEmpty` — returns the offending map entries on failure
- `info-uris-match-registered` → split into `registered-not-in-info` and `info-not-in-registered`, both `%test:assertEmpty` — shows the specific mismatched URIs
- `util-is-registered` → `%test:assertEquals("true")`
- `util-is-built-in` → `%test:assertEquals("built-in")`
- `mapped-modules-in-registered` — kept `%test:assertTrue` since it's a universal quantifier with no single expected value

joewiz:

We investigated performance concerns related to this PR and the broader question of module loading costs raised in #1848 and #4481. We ran JMH benchmarks measuring startup, memory, per-query overhead, and module discovery cost across minimal (4 module) and full (14+ module) configurations.

Key findings:

| Metric | Result |
|--------|--------|
| Startup cost per module | ~0.9 ms (extrapolates to ~40 ms for 44 modules — < 1% of BrokerPool startup) |
| Memory overhead per module | ~0 KB (modules share instances) |
| Per-query cost | Identical regardless of module count — no per-query compilation tax |
| XQueryContext creation | Scales at ~260 ns/module (~10 µs for 44 modules — negligible vs ~80 µs compilation) |
| Cold-call latency (first import) | +100 µs over warm call |
| `registered-modules()` | ~600 µs (no module loading, just URI collection) |
| `registered-functions()` 0-arity | ~1.2 ms for 3,328 functions |

We're pleased to find that eXist's module loading performance is quite good. Concerns about the cost of eagerly loading modules may not be a significant factor on the current codebase. The per-query overhead described in #1848 does not manifest in our benchmarks — query compilation and execution times are flat regardless of how many modules are configured.

Full benchmark data and methodology: `joewiz/benchmark/module-loading` branch, `MODULE-LOADING-REPORT.md`.

adamretter:
@joewiz Sorry to report the benchmarks for modules are flawed as they are not measuring the correct part of the code to achieve what you are trying to investigate Claude is looking in the wrong place again.

joewiz:

@adamretter You were right that the benchmarks weren't measuring the correct path. We dug deeper and found:

**Modules ARE re-instantiated per XQueryContext via reflection.** Every `new XQueryContext()` calls `loadDefaults()` which iterates all configured modules and calls `Class.getConstructor().newInstance()` + `prepare()` for each — zero sharing between contexts.

Updated measurements:

| Metric | Value |
|--------|-------|
| Per-module instantiation cost | ~623 ns (560 ns constructor + 63 ns prepare) |
| 14 test modules per context | ~8.7 µs |
| 44 production modules (extrapolated) | ~27 µs |
| Typical compilation time | 80–170 µs |
| Module instantiation as % of compilation | ~15–30% |

So the per-query module loading cost is real — confirmed at the code level. It's modest relative to compilation but would accumulate at high throughput.

That said, this PR (`registered-modules()` / `registered-functions()`) doesn't affect the per-query path. These discovery functions are called explicitly by developers and admin tools, not per query. The `registered-modules()` change only collects namespace URIs from already-loaded sources — no module instantiation. The `registered-functions()` 0-arity expansion resolves unloaded Java EXPath package modules, but only when the function is explicitly called.

Lazy loading (#4481) would eliminate the per-context instantiation cost for unused modules — a real optimization worth ~27 µs/query for a full production config. Our PR is compatible with that approach.

Updated benchmark data: `joewiz/benchmark/module-loading` branch, `MODULE-LOADING-REPORT.md`.

duncdrum (approved):

exist-core/src/main/java/org/exist/xquery/functions/util/ModuleInfo.java
```diff
@@ -250,6 +290,81 @@ public Sequence eval(Sequence[] args, Sequence contextSequence)
...
+		}
+		return "";
+	}
+
+	/**
+	 * Create a map with keys "uri", "prefix", and "source".
+	 */
+	private MapType createModuleInfoMap(final String uri, final String prefix, final String source) throws XPathException {
+		MapType map = new MapType(this, context);
+		map = (MapType) map.put(new StringValue(this, "uri"), new StringValue(this, uri));
```
line-o: I assume - I hope - there is a more elegant way to initialize a MapType

joewiz:

@line-o — good point, there is! `MapType.add(key, value)` is the mutable variant that avoids creating a new map for each entry. Fixed in 173ebb9819 — uses `add()` instead of the `put()` + cast chain.

exist-core/src/main/java/org/exist/repo/ExistRepository.java
```diff
@@ -334,6 +334,48 @@ public List<URI> getJavaModules() {
...
+            final Package pkg = pp.latest();
+            // 1. XQuery modules declared in exist.xml
+            final ExistPkgInfo info = (ExistPkgInfo) pkg.getInfo("exist");
+            if (info != null) {
+                modules.addAll(info.getXQueryModules());
+            }
+            // 2. XQuery modules declared in expath-pkg.xml (standard EXPath components)
+            final FileSystemResolver resolver = (FileSystemResolver) pkg.getResolver();
+            final Path pkgDescriptor = resolver.resolveResourceAsFile("expath-pkg.xml");
+            if (pkgDescriptor != null && Files.exists(pkgDescriptor)) {
```
line-o: Please invert the if-condition and `continue` when there is no `pkgDescriptor`. This will get rid of one level of indentation for the common case.

exist-core/src/main/java/org/exist/repo/ExistRepository.java
```diff
@@ -334,6 +334,48 @@ public List<URI> getJavaModules() {
...
+                try {
+                    final javax.xml.parsers.DocumentBuilderFactory dbf = javax.xml.parsers.DocumentBuilderFactory.newInstance();
+                    dbf.setNamespaceAware(true);
+                    final Document doc = dbf.newDocumentBuilder().parse(pkgDescriptor.toFile());
+                    final org.w3c.dom.NodeList xqueryElements = doc.getElementsByTagNameNS(
+                            "http://expath.org/ns/pkg", "xquery");
```
line-o: The namespace URI "http://expath.org/ns/pkg" should be a constant

exist-core/src/main/java/org/exist/repo/ExistRepository.java
```diff
@@ -334,6 +334,48 @@ public List<URI> getJavaModules() {
...
+            // 1. XQuery modules declared in exist.xml
+            final ExistPkgInfo info = (ExistPkgInfo) pkg.getInfo("exist");
+            if (info != null) {
+                modules.addAll(info.getXQueryModules());
+            }
+            // 2. XQuery modules declared in expath-pkg.xml (standard EXPath components)
+            final FileSystemResolver resolver = (FileSystemResolver) pkg.getResolver();
+            final Path pkgDescriptor = resolver.resolveResourceAsFile("expath-pkg.xml");
+            if (pkgDescriptor != null && Files.exists(pkgDescriptor)) {
+                try {
```
line-o: Consider extracting a method out of the entire try block

exist-core/src/main/java/org/exist/xquery/functions/util/BuiltinFunctions.java
```diff
@@ -151,6 +175,18 @@ private void addFunctionsFromModules(final ValueSequence resultSeq, final Module
 		}
 	}
 
+	private void addFunctionsFromModulesDedup(final ValueSequence resultSeq, final Module[] modules, final Set<QName> seen) {
```
line-o: consider renaming this function to `addDistinctFunctionsFromModules`

exist-core/src/main/java/org/exist/xquery/functions/util/ModuleInfo.java
```diff
@@ -228,17 +242,43 @@ public Sequence eval(Sequence[] args, Sequence contextSequence)
                     context.popDocumentContext();
                 }
             }
+            case "registered-modules-info" -> evalRegisteredModulesInfo();
             case null, default -> {
+                // registered-modules: return the union of all module namespace URIs
```
line-o: The default case should be extracted into a method call

duncdrum:
@joewiz  could you address the review comments?

joewiz:
@duncdrum Thanks for flagging these for me - I'm not sure how I missed them! Responses coming shortly.

joewiz:

Sorry for the delay @line-o — addressed all 5 in `29a840ddfa`:

| # | File:line | Resolution |
|---|---|---|
| 1 | `ExistRepository.java:349` (invert if + continue) | `getExpathPkgXQueryModules` extraction now early-returns `List.of()` when the descriptor is missing; the inner empty-namespace check uses `continue` to flatten the common case. |
| 2 | `ExistRepository.java:355` (namespace URI as constant) | Added `private static final String PKG_NAMESPACE = "http://expath.org/ns/pkg";` matching the existing convention used in `AppRestoreUtils` + `Deployment`. |
| 3 | `ExistRepository.java:350` (extract try-block into method) | New private `getExpathPkgXQueryModules(Package)` carries the descriptor parsing; `getXQueryModules()` now reads as four lines per package (info → addAll → expath-pkg → addAll). |
| 4 | `BuiltinFunctions.java:178` (rename `addFunctionsFromModulesDedup` → `addDistinctFunctionsFromModules`) | Done; all 3 sites updated. |
| 5 | `ModuleInfo.java:247` (extract default case into a method) | New private `evalRegisteredModules()` matches the existing `evalRegisteredModulesInfo()` pattern. The switch reads uniformly as one-line arrows now. |

No behavior changes. `xquery.util.UtilTests` passes 69/69 locally (covers the `moduleDiscoveryTest.xql` assertions on `registered-modules()`). Force-pushed and rebased onto current develop.

exist-core/src/main/java/org/exist/xquery/functions/util/BuiltinFunctions.java
```diff
@@ -57,7 +60,7 @@ public class BuiltinFunctions extends BasicFunction {
 			new FunctionSignature(
 					new QName("registered-functions", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
 					"Returns a sequence containing the QNames of all functions " +
-							"currently known to the system, including functions in imported and built-in modules.",
+							"currently known to the system, including functions in built-in modules and Java modules from installed EXPath packages.",
```
line-o: I am stumbling over the changed function description

> Returns a sequence containing the QNames of all functions currently known to the system, including functions in imported and built-in modules.

Changed to

> Returns a sequence containing the QNames of all functions currently known to the system, including functions in built-in modules and Java modules from installed EXPath packages.

I thought that this function would now _finally_ return all functions that are in the current context. And this would _include_ imported modules _regardless_ of their implementation.

line-o (commented):
Thanks for going the extra mile.
One last question

exist-core/src/main/java/org/exist/xquery/functions/util/BuiltinFunctions.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Good catch @line-o — the new docstring undersells what the function actually returns. Reading the `eval()` body for the 0-arg overload:

```java
// 1. Every module currently in the context (built-in + imported XQuery + imported Java)
final Module[] modules = StreamSupport.stream(context.getModules()...).toArray(...);
addDistinctFunctionsFromModules(resultSeq, modules, seen);

// 2. Java EXPath package modules not yet loaded into the context
if (context.getRepository().isPresent()) {
    for (final URI uri : repo.getJavaModules()) { ... }
}

// 3. Functions declared in the current main module
for (final Iterator<UserDefinedFunction> i = context.localFunctions(); i.hasNext(); ) { ... }
```

`context.getModules()` returns every module registered in the current `XQueryContext`, which includes any XQuery library module the running query has imported (since `import module namespace ... at "..."` registers the imported module in the context). So your expectation is right: this function DOES return all functions in the current context regardless of implementation language, plus EXPath-installed Java modules that haven't been imported yet, plus locally declared functions in the main module.

Pushed a docstring fix as `d41ac006`:

> Returns a sequence containing the QNames of all functions currently visible in the query context, including functions declared in the main module, functions from modules imported by the current query (XQuery library modules and Java modules), functions from eXist's built-in modules, and functions from Java modules in installed EXPath packages that have not yet been loaded into the context.

line-o (approved):

duncdrum (approved):

PR #6175, opened 2026-03-25
Fix ExecutorService thread leak in TransactionTestDSL

joewiz:
## Summary

`TransactionTestDSL.execute()` creates two `ExecutorService` instances but only shuts them down in the `catch` block (error path). On the happy path, both executors leak — their non-daemon threads accumulate across test classes in the same surefire fork.

The fix moves `shutdownNow()` to a `finally` block.

## How it was found

Thread dump captured during a full `mvn test -pl exist-core` run showed 10 leaked `db.exist.transaction-test-dsl.transaction-*-schedule` threads from 5 `ConcurrentTransactionsTest` invocations, all in `WAITING (parking)` state at `ThreadPoolExecutor.getTask()`. These non-daemon threads prevented clean JVM exit after all tests completed.

## What changed

**`TransactionTestDSL.java`** — Moved `t1ExecutorService.shutdownNow()` and `t2ExecutorService.shutdownNow()` from `catch` to `finally` so executor cleanup happens on both success and error paths.

## Related

- Companion PR: #6167 (BrokerPool shutdown hang fix — daemon StatusReporter threads + bounded wait loops)
- Together with #6167, this resolves CI timeout issues caused by JVM exit hangs (thread leak + shutdown hang)
- **Remaining hang**: `MoveResourceTest` still triggers mid-test `MultiLock` contention on concurrent collection move + query. This is real lock contention in the engine, not a test infrastructure issue. PR #6112 (preclaiming locks) resolves it — full test suite completes in 3:26 with #6112 vs hanging without it.

## Test plan

- [x] Full `mvn test -pl exist-core` — 6533 tests, 0 failures, BUILD SUCCESS
- [ ] CI passes on all platforms

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6183, opened 2026-03-26
Add bounded timeouts to BlobStore shutdown join() calls

joewiz:
## Summary

`BlobStoreImpl.normalClose()` and `abnormalPersistentWriterShutdown()` called `Thread.join()` with no timeout on the PersistentWriter and BlobVacuum threads. If either thread failed to terminate promptly, the shutdown would hang indefinitely — causing CI test suite timeouts when the surefire fork couldn't exit.

This supersedes #6179, which incorrectly made BlobStore threads daemon threads. As @adamretter (author of BlobStore) correctly pointed out, these threads must remain non-daemon to ensure pending blob writes complete. The fix is bounded `join()` timeouts during shutdown — giving threads 30s to drain before proceeding.

## What changed

**`BlobStoreImpl.java`** — Added 30-second bounded timeouts to all three `Thread.join()` calls:

- `normalClose()` — PersistentWriter: `join(30_000)`, then `interrupt()` + `join(5_000)` if still alive
- `normalClose()` — BlobVacuum: `join(30_000)` with warning if still alive
- `abnormalPersistentWriterShutdown()` — BlobVacuum: `join(30_000)` with warning if still alive

## Why this is safe

The existing shutdown mechanisms are correct and unchanged:
1. PersistentWriter receives a `POISON_PILL` via the persist queue → drains remaining items and exits
2. BlobVacuum receives `interrupt()` → wakes from `PriorityBlockingQueue.take()` and exits

In practice, both threads respond within milliseconds. The 30s timeout is a safety net — it only fires if something goes wrong, ensuring shutdown completes instead of hanging forever.

## Test results

- 27/27 BlobStore-specific tests pass
- 3/3 full test suite trials: BUILD SUCCESS, clean exit (3:26–4:23)

## Test plan

- [x] `BlobStoreImplTest` — 27 tests, 0 failures
- [x] Full `mvn test -pl exist-core` — 3 trials, all pass, no hangs
- [ ] CI passes on all platforms

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/blob/BlobStoreImpl.java
+++ b/exist-core/src/main/java/org/exist/storage/blob/BlobStoreImpl.java
@@ -447,13 +447,21 @@ public class BlobStoreImpl implements BlobStore {
             if (persistentWriter != null) {
                 persistQueue.put(PersistentWriter.POISON_PILL);
             }
-            persistentWriterThread.join();
+            persistentWriterThread.join(30_000);
+            if (persistentWriterThread.isAlive()) {
+                LOG.warn("BlobStore PersistentWriter did not terminate within 30s, interrupting");
+                persistentWriterThread.interrupt();
+                persistentWriterThread.join(5_000);
+            }
 
             // shutdown the vacuum
             if (blobVacuum != null) {
                 blobVacuumThread.interrupt();
             }
-            blobVacuumThread.join();
+            blobVacuumThread.join(30_000);
+            if (blobVacuumThread.isAlive()) {
+                LOG.warn("BlobStore BlobVacuum did not terminate within 30s");
+            }
         } catch (final InterruptedException e) {
             // Restore the interrupted status
             Thread.currentThread().interrupt();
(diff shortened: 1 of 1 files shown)


adamretter:
@joewiz Unfortunately, in practice this has the same corruption issue as your previous approach in #6179, it just goes abount achieving that corruption in a different manner.

joewiz:

Closing this PR. Adam is right that the bounded timeout has the same fundamental risk as the daemon thread approach — if the timeout fires and `closeBlobStore()` proceeds while the writer is mid-write, data corruption is possible.

The CI hang symptom that motivated this fix is now addressed through other means:
- #6167 (merged) and #6175 (merged) fixed the dominant shutdown hangs — develop baseline is clean in our controlled experiment (5/5 trials, 0 hangs)
- #6186 adds surefire fork timeouts as a safety net, ensuring CI never burns 45+ minutes on a hung fork

The underlying BlobStore shutdown behavior (non-daemon threads that can prevent JVM exit if they hang) remains as-is. It's the correct design for data integrity — these threads must complete their work before the JVM exits.

PR #6184, opened 2026-03-26
[feature] Add repo:resource-available() function

### LINKED ISSUE #3904: [feature] Add a new function repo:resource-available

With https://github.com/eXist-db/exist/pull/3896, calls to `repo:get-resource` now throw an error when the 2nd argument, `$resource`, does not exist in the package (as stated in [the function's documentation](http://www.exist-db.org/exist/apps/fundocs/view.html?uri=http://exist-db.org/xquery/repo&location=java:org.exist.xquery.modules.expathrepo.ExpathPackageModule&details=true#get-resource.2)).

We now need a `repo:resource-available` function to check if a resource exists before we use `repo:get-resource` to get it. The purpose is analogous to `fn:doc-available` for `fn:doc`, and `util:binary-resource-available` for `util:binary-resource`.

We need a new function that checks if a resource is available and returns `true()` or `false()` accordingly:

```xquery
declare function repo:resource-available($pkgName as xs:string, $resource as xs:string) as xs:boolean
```

In the meantime, a try-catch expression will fix affected code, but it's just a short term workaround.

**Additional context**

The motivation is to help in the fix to https://github.com/eXist-db/existdb-packageservice/issues/19.

joewiz:
## Summary

Add `repo:resource-available($pkgName, $resource)` that returns `true` if the specified resource exists in the named EXPath package, `false` otherwise. This is the companion to `repo:get-resource()`, analogous to `fn:doc-available()` for `fn:doc()`.

## Motivation

Since PR #3896, `repo:get-resource()` throws `FODC0002` when the resource doesn't exist. There's currently no way to check if a resource exists without catching the error. Tools like exist-api need to probe for resources like `exist.xml`, `repo.xml`, etc. in installed packages.

## What Changed

- **`ResourceAvailable.java`** (new): Implementation modeled on `GetResource.java`. Uses `pkg.getResolver().resolveResource(path)` and catches `Storage.NotExistException` to return false. Closes the `StreamSource` input stream immediately since only existence is needed.
- **`ExpathPackageModule.java`**: Register the new function.
- **`resource-available.xql`** (new): 3 XQSuite tests covering missing resource, missing package, and existing resource scenarios.

## Test plan

- [x] `mvn test -pl extensions/modules/expathrepo -am` — 8 tests pass (5 existing + 3 new)
- [x] CI green

Closes https://github.com/eXist-db/exist/issues/3904

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/extensions/modules/expathrepo/src/main/java/org/exist/xquery/modules/expathrepo/ResourceAvailable.java
+++ b/extensions/modules/expathrepo/src/main/java/org/exist/xquery/modules/expathrepo/ResourceAvailable.java
@@ -1,0 +1,100 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.modules.expathrepo;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Optional;
+
+import javax.xml.transform.stream.StreamSource;
+
+import org.exist.dom.QName;
+import org.exist.repo.ExistRepository;
+import org.exist.xquery.BasicFunction;
+import org.exist.xquery.Cardinality;
+import org.exist.xquery.FunctionSignature;
+import org.exist.xquery.XPathException;
+import org.exist.xquery.XQueryContext;
+import org.exist.xquery.value.*;
+import org.expath.pkg.repo.Package;
+import org.expath.pkg.repo.PackageException;
+import org.expath.pkg.repo.Packages;
+import org.expath.pkg.repo.Storage;
+
+/**
+ * repo:resource-available($pkgName, $resource) as xs:boolean
+ *
+ * Returns true if the specified resource exists in the named EXPath package,
+ * false otherwise. Companion to repo:get-resource(), analogous to
+ * fn:doc-available() for fn:doc().
+ */
+public class ResourceAvailable extends BasicFunction {
+
+    public final static FunctionSignature signature =
+        new FunctionSignature(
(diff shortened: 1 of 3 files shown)


extensions/modules/expathrepo/src/test/xquery/modules/expathrepo/resource-available.xql
```diff
@@ -0,0 +1,61 @@
...
+
+(: If any packages are installed, their expath-pkg.xml should exist :)
+declare
+    %test:assertTrue
+function ra:existing-resource-if-packages-installed() {
+    let $pkgs := repo:list()
+    return
+        if (exists($pkgs))
+        then repo:resource-available($pkgs[1], "expath-pkg.xml")
+        else true() (: vacuously true if no packages installed :)
```
duncdrum: This seems to defeat the purpose of the test.

joewiz:

Good catch, @duncdrum — the vacuous `true()` fallback defeated the purpose. Rewrote the test to install a temporary package (following the `deployment.xql` pattern with setUp/tearDown), then check `resource-available` against a real package resource. All 3 tests now exercise the function properly.

duncdrum (approved):

extensions/modules/expathrepo/src/main/java/org/exist/xquery/modules/expathrepo/ResourceAvailable.java
```diff
@@ -0,0 +1,100 @@
...
+                    if (pkg.getName().equals(pkgName)) {
+                        try {
+                            // resolveResource opens a stream — close it immediately
+                            final StreamSource source = (StreamSource) pkg.getResolver().resolveResource(path);
+                            final InputStream is = source.getInputStream();
+                            if (is != null) {
+                                is.close();
+                            }
+                            return BooleanValue.TRUE;
+                        } catch (final Storage.NotExistException e) {
```
line-o, line 86: Can the catch-statements be combined? One statement handling both errors since they both just return FALSE
```java
} catch (final Storage.NotExistException | IOException e) {
```

extensions/modules/expathrepo/src/main/java/org/exist/xquery/modules/expathrepo/ResourceAvailable.java
```diff
@@ -0,0 +1,100 @@
...
+        super(context, signature);
+    }
+
+    @Override
+    public Sequence eval(final Sequence[] args, final Sequence contextSequence) throws XPathException {
+        final String pkgName = args[0].getStringValue();
+        final String path = args[1].getStringValue();
+
+        final Optional<ExistRepository> repo = context.getRepository();
+        if (repo.isPresent()) {
```
line-o, line 73: Please invert the condition here 

```java
if (!repo.isPresent()) { return BooleanValue.FALSE }
// the happy path here
```

line-o (changes requested):
I found two minor issues in the implementation that I would like to see if they can be improved.

joewiz:

Good suggestions, @line-o. Pushed 7f1e741 addressing both:
1. Inverted the `repo.isPresent()` guard to early-return, de-nesting the happy path
2. Combined the two catch blocks into `catch (final Storage.NotExistException | IOException e)`

line-o (approved):

PR #6185, opened 2026-03-26
[ci] Fix composite action secrets access for GitHub Packages auth

joewiz:
## Summary

- Fix `maven-github-settings` composite action to accept the GitHub token as an input parameter instead of accessing `secrets.GITHUB_TOKEN` directly (which doesn't work in composite actions)
- Update all three calling workflows (`ci-test`, `ci-snapshots`, `ci-xqts`) to pass the token via `with`

## Problem

Composite actions [cannot access the `secrets` context](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action#using-inputs-and-outputs-in-a-composite-action). The `${{ secrets.GITHUB_TOKEN }}` reference in the action silently resolves to an empty string, breaking GitHub Packages authentication for Maven dependency resolution.

Appears to have been broken since commit dd71b1afbf (Feb 13, 2026).

## Changes

| File | Change |
|------|--------|
| `.github/actions/maven-github-settings/action.yml` | Add `inputs.token` parameter; use `${{ inputs.token }}` instead of `${{ secrets.GITHUB_TOKEN }}` |
| `.github/workflows/ci-test.yml` | Pass `token: ${{ secrets.GITHUB_TOKEN }}` to the action |
| `.github/workflows/ci-snapshots.yml` | Pass `token: ${{ secrets.GITHUB_TOKEN }}` to the action |
| `.github/workflows/ci-xqts.yml` | Pass `token: ${{ secrets.GITHUB_TOKEN }}` to the action |

## Test plan

- [x] CI workflows should now authenticate correctly with GitHub Packages

### DIFF
--- a/.github/actions/maven-github-settings/action.yml
+++ b/.github/actions/maven-github-settings/action.yml
@@ -2,6 +2,10 @@
 # Required for resolving from maven.pkg.github.com/eXist-db/exist and exist-xqts-runner.
 name: Maven GitHub Packages settings
 description: Create settings.xml with github and github-xqts-runner servers
+inputs:
+  token:
+    description: 'GitHub token for package authentication'
+    required: true
 runs:
   using: 'composite'
   steps:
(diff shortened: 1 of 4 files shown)


duncdrum (approved):
I m happy to move to a more idiomatic pattern, but out of curiosity: where do you see the failure with the composite action? The failing xqts pull during build is passing, snapshots are deployed? 

My understanding is that `GITHUB_TOKEN` should be available to all workflows including composite

line-o:
@duncdrum here for instance https://github.com/eXist-db/exist/actions/runs/23578638838/job/68656676280

line-o (approved):

PR #6186, opened 2026-03-26
[ci] Add surefire fork timeouts to prevent CI hangs

joewiz:
## Summary

Two CI reliability improvements:

1. **Surefire fork timeouts**: When a test like `DeadlockIT` or `MoveResourceTest` hangs during CI, the surefire/failsafe forked JVM waits indefinitely — the only protection is the GitHub Actions step timeout at 45 minutes. This burns CI minutes and blocks PR merges. Adding surefire fork timeouts kills hung test JVMs after 10 minutes instead of 45.

2. **Maven Build wagon timeouts + offline license check**: Several Maven repositories (`repo.exist-db.org`, `repo.evolvedbinary.com`) cause stalled or failed HTTP connections from CI runners (CLOSE_WAIT and 504 errors respectively). This causes the Maven Build step to hang silently for 25+ minutes until the 30-minute step timeout fires.

### `exist-parent/pom.xml`

Added to both `maven-surefire-plugin` and `maven-failsafe-plugin` configuration:

```xml
<forkedProcessTimeoutInSeconds>600</forkedProcessTimeoutInSeconds>
<forkedProcessExitTimeoutInSeconds>60</forkedProcessExitTimeoutInSeconds>
```

- **`forkedProcessTimeoutInSeconds=600`**: Kills the forked JVM after 10 minutes. Clean test runs complete in ~3.5 minutes, so this only fires on hung tests.
- **`forkedProcessExitTimeoutInSeconds=60`**: Gives the fork 60 seconds to flush results before force-kill.

### `.github/workflows/ci-test.yml`

Three changes:

1. **Maven Build: wagon timeout via MAVEN_OPTS.** Added wagon connection/read timeouts (10s/30s) to the Maven Build step's `MAVEN_OPTS`. Setting these in `MAVEN_OPTS` (not as `-D` flags on the mvnd command line) ensures they are passed as JVM system properties to the Maven daemon — which is where wagon reads them. Causes stalled connections to fail quickly instead of hanging for 25+ minutes.

2. **License check: run offline.** Added `--offline` to the `license:check` step. The license job always restores the Maven cache from `develop` (which has all required artifacts), so no network access is needed. Eliminates 7+ minute timeouts from the same repo issues. Result: ~20s.

3. **Integration test step timeout: 45 → 30 minutes.** With surefire killing hung forks at 10 minutes, 30 minutes is ample for the full integration suite.

## Evidence the fix works

An earlier iteration of this PR proved the surefire timeout infrastructure works:

| Metric | Before (no timeout) | After (this PR) |
|--------|---------------------|-----------------|
| `DeadlockIT` | Hung for 44 min, killed by step timeout | Completed in 522s (under 600s limit) |
| Windows integration total | 49 min (step timeout) | **14 min 55s** |
| Tests lost | Fork killed mid-run, results lost | **All 9 tests passed**, 0 failures |

## Why these values

- Clean test suite runs complete in ~3.5 min (local) and ~16 min (CI with all modules)
- Hung tests (`DeadlockIT`, `MoveResourceTest`) run indefinitely without intervention
- 10 minutes (600s) is generous enough to never kill a slow-but-legitimate test run
- 30s wagon read timeout: legitimate artifact downloads complete in seconds; CLOSE_WAIT connections produce no data

## Test plan

- [ ] License check completes in ~20s (offline, no repo network calls)
- [ ] Maven Build completes without repo timeout hangs
- [ ] Integration tests: hung forks killed after 10 min, not 45
- [ ] No regressions: all clean tests still pass within the 600s fork timeout

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -17,7 +17,7 @@ jobs:
           distribution: liberica
           java-version: ${{ env.DEV_JDK }}
       - uses: ./.github/actions/maven-cache
-      - run: mvn -V -B --no-transfer-progress license:check
+      - run: mvn -V -B --no-transfer-progress --offline license:check
         timeout-minutes: 10
   dependencies:
     name: Dependency checks
@@ -88,6 +88,11 @@ jobs:
         timeout-minutes: 30
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          MAVEN_OPTS: >-
+            -DtrimStackTrace=false
+            -D'maven.resolver.transport=wagon'
+            -D'maven.wagon.http.connectionTimeout=10000'
+            -D'maven.wagon.http.readTimeout=30000'
         run: ${{ steps.install-mvnd.outputs.mvnd-dir }}/mvnd -V -B --no-transfer-progress -T 1C compile test-compile -DtrimStackTrace=false -D'dependency-check.skip' -D'license.skip' --projects '!exist-installer'
       - name: Maven Unit Tests
         if: matrix.test-type == 'unit'
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -95,7 +95,7 @@ jobs:
         run: ${{ steps.install-mvnd.outputs.mvnd-dir }}/mvnd -V -B --no-transfer-progress test -DtrimStackTrace=false -D'dependency-check.skip' -D'license.skip' -D'mvnd.maxLostKeepAlive=6000'
       - name: Maven Integration Tests
         if: matrix.test-type == 'integration'
-        timeout-minutes: 45
+        timeout-minutes: 30
         run: ${{ steps.install-mvnd.outputs.mvnd-dir }}/mvnd -V -B --no-transfer-progress verify -DskipUnitTests=true -DtrimStackTrace=false -D'dependency-check.skip' -D'license.skip' -D'mvnd.maxLostKeepAlive=6000'
       - name: Javadoc (ubuntu unit only)
         if: matrix.os == 'ubuntu-latest' && matrix.test-type == 'unit'
--- a/exist-parent/pom.xml
+++ b/exist-parent/pom.xml
@@ -842,6 +842,10 @@
                               However it can make it hard to diagnose problems if tests leak state; If you experience
                               such a problem you may want to set it to `false` whilst debugging -->
                         <reuseForks>true</reuseForks>
+                        <!-- Kill forked JVMs that hang (e.g. DeadlockIT, MoveResourceTest).
+                             10 minutes is generous; clean runs complete in ~3.5 minutes. -->
+                        <forkedProcessTimeoutInSeconds>600</forkedProcessTimeoutInSeconds>
+                        <forkedProcessExitTimeoutInSeconds>60</forkedProcessExitTimeoutInSeconds>
                         <argLine>@{jacocoArgLine} --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.ref=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED -Dfile.encoding=${project.build.sourceEncoding}</argLine>
                         <systemPropertyVariables>
                             <user.country>UK</user.country>
@@ -863,6 +867,10 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-failsafe-plugin</artifactId>
                     <version>3.5.5</version>
+                    <configuration>
+                        <forkedProcessTimeoutInSeconds>600</forkedProcessTimeoutInSeconds>
+                        <forkedProcessExitTimeoutInSeconds>60</forkedProcessExitTimeoutInSeconds>
+                    </configuration>
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>


dizzzz (approved):

.github/workflows/ci-test.yml
```diff
@@ -90,14 +88,19 @@ jobs:
         timeout-minutes: 30
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          MAVEN_OPTS: >-
+            -DtrimStackTrace=false
```
duncdrum, line 92: I like centralising these options, but we are doing it only half-heartedly e.g. `-DtrimStackTrace=false` is appears in the env and on every invocation. We should make this DRY. There are more flags similarly repeated

.github/workflows/ci-test.yml
```diff
@@ -90,14 +88,19 @@ jobs:
...
+            -D'maven.resolver.transport=wagon'
```
duncdrum, line 93: Do we want wagon at all? Its no longer the default https://maven.apache.org/guides/mini/guide-resolver-transport.html for a reason

exist-parent/pom.xml
```diff
@@ -842,6 +842,10 @@
                               However it can make it hard to diagnose problems if tests leak state; If you experience
                               such a problem you may want to set it to `false` whilst debugging -->
                         <reuseForks>true</reuseForks>
+                        <!-- Kill forked JVMs that hang (e.g. DeadlockIT, MoveResourceTest).
+                             10 minutes is generous; clean runs complete in ~3.5 minutes. -->
+                        <forkedProcessTimeoutInSeconds>600</forkedProcessTimeoutInSeconds>
```
duncdrum, line 847: The idea with #6045 was to move slow tests into the IT phase. Unit test should be faster, so I would expect the timeouts here to be lower compared to failsafe tests. If there are slow tests left in surefire, I would rather move them to IT. 
Does surefire even fork a jvm? I thought this was a failsafe thing?

.github/workflows/ci-test.yml
```diff
@@ -90,14 +88,19 @@ jobs:
...
+            -D'maven.wagon.http.readTimeout=30000'
         run: ${{ steps.install-mvnd.outputs.mvnd-dir }}/mvnd -V -B --no-transfer-progress -T 1C compile test-compile -DtrimStackTrace=false -D'dependency-check.skip' -D'license.skip' --projects '!exist-installer'
       - name: Maven Unit Tests
         if: matrix.test-type == 'unit'
         timeout-minutes: 45
         run: ${{ steps.install-mvnd.outputs.mvnd-dir }}/mvnd -V -B --no-transfer-progress test -DtrimStackTrace=false -D'dependency-check.skip' -D'license.skip' -D'mvnd.maxLostKeepAlive=6000'
       - name: Maven Integration Tests
         if: matrix.test-type == 'integration'
-        timeout-minutes: 45
+        timeout-minutes: 30
```
duncdrum, line 103: I've seen 37min successful runs on CI. I would suggest to be more generous with the CI timeout until we can establish a new baseline of successful CI runs, and then lower the workflow timeouts accordingly.

duncdrum (changes requested):
I m not quite sure why we keep running into these build issues on CI. Maven central, repo.exist-db and Github are up.  It's quite unlikely that all three are suddenly struck with outages.

.github/workflows/ci-test.yml
joewiz: [This response was co-authored with Claude Code. -Joe]

Agreed — this is a real smell and worth cleaning up properly. I'll address this as part of reworking the transport question; once we settle on native vs wagon the flag set will change anyway, and that's a good moment to centralize everything.

.github/workflows/ci-test.yml
joewiz: [This response was co-authored with Claude Code. -Joe]

Good question, and the git history has an answer: `-Dmaven.resolver.transport=wagon` was added by Dannes on 2023-03-02 with commit message "testing" — almost certainly a temporary workaround for instability in Maven 3.9.0's native transport when it first became the default. It was never revisited. mvnd 1.0.3 bundles Maven 3.9.11, so whatever prompted the workaround is long fixed.

I'd like to try removing it (and the wagon-specific timeout flags) and replacing with Aether native timeout flags (`aether.connector.basic.connectTimeout`, `aether.connector.basic.requestTimeout`). Will open a test branch and report back.

exist-parent/pom.xml
joewiz: [This response was co-authored with Claude Code. -Joe]

Surefire does fork here — `forkCount=2C` with `reuseForks=true` in exist-parent/pom.xml, so `forkedProcessTimeoutInSeconds` is applicable. That said, your point about lower timeouts for unit tests is well taken, and moving slow stragglers to the IT phase (#6045) is the right long-term answer. For now I'll remove the `forkedProcessTimeoutInSeconds` from surefire and keep it only in failsafe, where it has the clearest justification.

.github/workflows/ci-test.yml
joewiz: [This response was co-authored with Claude Code. -Joe]

Good to know — reverting to 45 minutes. The 30-minute reduction was premature given you've seen legitimate 37-minute runs.

joewiz:

Fair challenge. What we observed locally was CLOSE_WAIT on TCP connections to `repo.exist-db.org` — the server accepts the connection then sends FIN, leaving the client socket in CLOSE_WAIT and Maven's HTTP client waiting for data that never arrives. And CI logs showed HTTP 504 from `repo.evolvedbinary.com`. Neither means the services are "down" in the traditional sense — it's more likely a connection-lifecycle mismatch between wagon's HTTP client and how those servers handle idle connections.

But your question about wagon (see inline comment) led to something more interesting: the `-Dmaven.resolver.transport=wagon` flag in global MAVEN_OPTS was added by Dannes in March 2023 with commit message "testing" — likely a workaround for early Maven 3.9.0 native transport instability when it first became the default. mvnd 1.0.3 now bundles Maven 3.9.11. That flag is almost certainly stale, and removing it (letting native transport handle everything, with Aether timeout flags) is probably the right fix rather than working around wagon's behavior. Opening a test branch now.

joewiz:
Superseded by #6224.

PR #6192, opened 2026-03-28
[feature] Support file: URIs in fn:collection() for filesystem directory querying

joewiz:
## Summary

- Extend `fn:collection()` to support `file:` URIs, enabling queries over filesystem directories
- `fn:collection("file:///path/to/dir")` scans for `*.xml` files by default
- `fn:collection("file:///path/to/dir?select=*.xhtml")` supports glob filtering (Saxon convention)
- DBA-only access for `file:` URIs (security boundary, consistent with `fn:doc()`)
- Non-parseable files silently skipped; non-existent directories throw FODC0002

## Motivation

BaseX and Saxon both support `collection("file:/path/to/dir")` for querying filesystem directories. eXist's `fn:doc()` already supports `file:` URIs but `fn:collection()` was limited to database collections only. The W3C spec says `fn:collection()` is implementation-defined, making this a conformant extension.

## Test plan

- [x] XQSuite test: non-existent directory throws FODC0002
- [ ] Manual: `collection("file:///tmp/test-xml/")` returns XML documents
- [ ] Manual: `?select=*.xhtml` glob filtering works
- [ ] Manual: non-DBA user gets permission denied

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/CollectionQueryParameters.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/CollectionQueryParameters.java
@@ -1,0 +1,237 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.functions.fn;
+
+import org.exist.xquery.ErrorCodes;
+import org.exist.xquery.Expression;
+import org.exist.xquery.XPathException;
+
+import javax.annotation.Nullable;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Saxon-style query string parameters shared by {@link FunUriCollection}
+ * (database collections) and {@link ExtCollection} (file: URI collections).
+ *
+ * <p>Supports four parameters:</p>
+ * <ul>
+ *   <li>{@code select} — glob pattern matching resource names (file: only)</li>
+ *   <li>{@code match} — regex matching resource names</li>
+ *   <li>{@code content-type} — MIME type filter (database: distinguishes XML/binary/collection;
+ *       file: distinguishes XML/binary documents)</li>
+ *   <li>{@code stable} — {@code yes}/{@code no} for deterministic ordering and caching</li>
+ * </ul>
+ *
+ * <p>Each consumer specifies which subset of parameters it accepts via the {@code allowedKeys}
+ * argument to {@link #parse(String, Set, Expression)}. Unknown or invalid keys/values raise
+ * {@link ErrorCodes#FODC0004}.</p>
+ */
+public final class CollectionQueryParameters {
+
+    public static final String KEY_SELECT = "select";
+    public static final String KEY_MATCH = "match";
+    public static final String KEY_CONTENT_TYPE = "content-type";
+    public static final String KEY_STABLE = "stable";
+
+    public static final String VALUE_CONTENT_TYPE_DOCUMENT = "application/vnd.existdb.document";
+    public static final String VALUE_CONTENT_TYPE_DOCUMENT_BINARY = "application/vnd.existdb.document+binary";
+    public static final String VALUE_CONTENT_TYPE_DOCUMENT_XML = "application/vnd.existdb.document+xml";
+    public static final String VALUE_CONTENT_TYPE_SUBCOLLECTION = "application/vnd.existdb.collection";
+    public static final String[] VALUE_CONTENT_TYPES = {
+            VALUE_CONTENT_TYPE_DOCUMENT,
+            VALUE_CONTENT_TYPE_DOCUMENT_BINARY,
+            VALUE_CONTENT_TYPE_DOCUMENT_XML,
+            VALUE_CONTENT_TYPE_SUBCOLLECTION
+    };
+
+    public static final String VALUE_STABLE_NO = "no";
+    public static final String VALUE_STABLE_YES = "yes";
+    public static final String[] VALUE_STABLES = {
+            VALUE_STABLE_NO,
+            VALUE_STABLE_YES
+    };
+
+    /** Keys accepted by fn:uri-collection (no select). */
+    public static final Set<String> URI_COLLECTION_KEYS = Set.of(KEY_MATCH, KEY_CONTENT_TYPE, KEY_STABLE);
+
+    /** Keys accepted by fn:collection() with file: URIs (includes select). */
+    public static final Set<String> FILE_COLLECTION_KEYS = Set.of(KEY_SELECT, KEY_MATCH, KEY_CONTENT_TYPE, KEY_STABLE);
+
+    @Nullable private final String select;
+    @Nullable private final String match;
+    @Nullable private final String contentType;
+    private final boolean stable;
+    private final Map<String, String> rawMap;
+
+    private CollectionQueryParameters(@Nullable final String select,
+                                      @Nullable final String match,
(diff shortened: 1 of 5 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
```diff
@@ -120,6 +132,10 @@ private void getCollectionItems(final URI collectionUri, final Sequence items) t
         if (dynamicCollection != null) {
             items.addAll(dynamicCollection);
 
+        } else if (collectionUri.getScheme() != null && collectionUri.getScheme().equals("file")) {
```
reinhapa:
```suggestion
        } else if ("file".equals(collectionUri.getScheme()) {
```

This would be even a simpler check...

reinhapa (changes requested):

joewiz:

Good catch, @reinhapa — simplified to `"file".equals(collectionUri.getScheme())` which handles the null case cleanly. Pushed.

adamretter:
A few thoughts:

1. This is really just a shortcut for doing the exact equivalent using the File extension module.

2. `fn:collection-uri`  already supports Saxon query string syntax that I added for `match`, `content-type`, and `stable` - it would be nice to see this unified with `fn:collection` i.e. both should support those options plus also `select` if that's one you would like to add.

3. There are serious security implications around using `file://`. Can I suggest that you add some sensible controls around that in the same manner (or better) that we did for the File extension module please? Any time you provide a route to access the filesystem, you open up a security hole. I already demonstrated that several of the online XQuery and XSLT fiddles can be used to read /etc/passwd (and worse) from the hosted systems. eXide which is included by default is not different to a fiddle - not to mention the REST API etc that allows remote query execution by guest and/or non-DBA users..

line-o:
> fn:collection-uri already supports Saxon query string syntax that I added for match, content-type, and stable - it would be nice to see this unified with fn:collection i.e. both should support those options plus also select if that's one you would like to add.

I would want to avoid putting in resources into supporting non-standard extensions from Saxon.

I would be in favour of proposing that XQuery 4.0 will get `fn:collection#2` with options maps similar to `fn:doc#2`  and `fn:doc-available#2` see https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-doc

exist-core/src/main/java/org/exist/xquery/functions/fn/CollectionQueryParameters.java
```diff
@@ -0,0 +1,237 @@
...
+                        String.format("Unexpected query string \"%s\".", entry));
+            }
+
+            if (key.equals(KEY_CONTENT_TYPE)) {
+                if (Arrays.stream(VALUE_CONTENT_TYPES).noneMatch(v -> v.equals(value))) {
+                    throw new XPathException(caller, ErrorCodes.FODC0004,
+                            String.format("Invalid query-string value \"%s\".", entry));
+                }
+            } else if (key.equals(KEY_STABLE)) {
+                if (Arrays.stream(VALUE_STABLES).noneMatch(v -> v.equals(value))) {
```
reinhapa: Join this if statements

exist-core/src/main/java/org/exist/xquery/functions/fn/CollectionQueryParameters.java
```diff
@@ -0,0 +1,237 @@
...
+    public static final String KEY_SELECT = "select";
+    public static final String KEY_MATCH = "match";
+    public static final String KEY_CONTENT_TYPE = "content-type";
+    public static final String KEY_STABLE = "stable";
+
+    public static final String VALUE_CONTENT_TYPE_DOCUMENT = "application/vnd.existdb.document";
+    public static final String VALUE_CONTENT_TYPE_DOCUMENT_BINARY = "application/vnd.existdb.document+binary";
+    public static final String VALUE_CONTENT_TYPE_DOCUMENT_XML = "application/vnd.existdb.document+xml";
+    public static final String VALUE_CONTENT_TYPE_SUBCOLLECTION = "application/vnd.existdb.collection";
+    public static final String[] VALUE_CONTENT_TYPES = {
```
reinhapa: Consider using `Set.of` to have a real immutable collection.

exist-core/src/main/java/org/exist/xquery/functions/fn/CollectionQueryParameters.java
```diff
@@ -0,0 +1,237 @@
...
+            VALUE_CONTENT_TYPE_DOCUMENT,
+            VALUE_CONTENT_TYPE_DOCUMENT_BINARY,
+            VALUE_CONTENT_TYPE_DOCUMENT_XML,
+            VALUE_CONTENT_TYPE_SUBCOLLECTION
+    };
+
+    public static final String VALUE_STABLE_NO = "no";
+    public static final String VALUE_STABLE_YES = "yes";
+    public static final String[] VALUE_STABLES = {
```
reinhapa: Consider using `Set.of` to have a real immutable collection.

exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
```diff
@@ -147,6 +191,96 @@ private void getCollectionItems(final URI collectionUri, final Sequence items) t
...
+     *   <li>{@code content-type=...} — MIME filter; for file: URIs:
+     *     {@code application/vnd.existdb.document+xml} or {@code application/vnd.existdb.document}
+     *     selects XML files (the only kind fn:collection returns)</li>
+     *   <li>{@code stable=yes|no} — when {@code yes} (default), files are returned in alphabetical order</li>
+     * </ul>
+     * <p>
+     * Only DBA users can access the file system directly.
+     * </p>
+     */
+    private void getFileCollectionItems(final URI collectionUri, final String rawQueryString, final Sequence items) throws XPathException {
```
reinhapa, line 211: Addresss Codacy issue: _The method 'getFileCollectionItems(URI, String, Sequence)' has an NPath complexity of 17280, current threshold is 200_

reinhapa (changes requested):

line-o:
I would like to postone this PR to have more time to think about the consequences

joewiz:

Thanks @reinhapa — addressed in 0e86b7cd25.

**Converted to `Set.of(...)`** (`CollectionQueryParameters.java:62, :71`): `VALUE_CONTENT_TYPES` and `VALUE_STABLES` are now `Set<String>` declared with `Set.of(...)`, matching the same immutability pattern already used for `URI_COLLECTION_KEYS` and `FILE_COLLECTION_KEYS` in the same file. Dropped the now-unused `java.util.Arrays` import.

**Joined nested if statements** (`CollectionQueryParameters.java:168`): `checkQueryStringMap` now does a single combined check per case (`if (key.equals(KEY_CONTENT_TYPE) && !VALUE_CONTENT_TYPES.contains(value))`). The new `Set.contains(...)` is both clearer and faster than the previous `Arrays.stream(...).noneMatch(...)`.

**Already addressed** (`ExtCollection.java:135`): your earlier suggestion (March) to use `"file".equals(collectionUri.getScheme())` is now in the branch at line 163 — no further action needed.

**Not addressed** (`ExtCollection.java:211` NPath complexity 17280 on `getFileCollectionItems`): leaving this one untouched for the same reason @line-o gave on #6344 yesterday — refactoring a pre-existing high-NPath method mid-review-wave raises the chance of conflicts with other pipeline PRs touching `ExtCollection`. Happy to revisit as a focused follow-up PR after the current release wave clears.

Verified: build clean, `XQuery3Tests` 1027/1027 + `CollectionFileUriTest` 12/12 pass.

reinhapa (approved):

duncdrum:
@joewiz one codacy flag is still in. Could you try to lower that functions npath score

joewiz:

Done in eede5ee1c0 — and there turned out to be two NPath methods rather than one, so I lowered both.

**`ExtCollection.getFileCollectionItems`** — NPath 17280, and genuinely new in this PR. Now an orchestrator over `requireDbaAccessForFileUris`, `collectCandidateFiles`, `isSelectableFile` and `addParsedDocuments`, split along the phase boundaries the method already had. One ordering detail I deliberately preserved: the `select` glob and `match` regex are still compiled *before* the directory-existence check, so an invalid regex combined with a missing directory still surfaces the regex error first rather than "Directory does not exist".

**`FunUriCollection.eval`** — NPath 1307. This one is pre-existing: it measures 2619 on `develop`, and this PR had already halved it by extracting `CollectionQueryParameters`. It's now an orchestrator over `collectUris` and `toUriSequence`, with the single-assignment `result` variable replaced by early returns for the two short-circuit cases (empty argument, stable cache hit) — both of which already bypassed the result cache, so returning early is equivalent. I kept the distinction between the empty sequence (an explicit `match` that selects nothing) and an empty `ValueSequence` (no `match`, no results), since those two aren't interchangeable once cached.

No behavior change in either. PMD now reports **no findings at all** on the three files this PR touches, and both methods are under the threshold of 200.

Verification: `CollectionFileUriTest` 12/12, and the XQuery3 XQSuite run that covers `fnUriCollection.xqm` 1027/1027 (1 pre-existing skip).

PR #6208, opened 2026-04-06
Add query profiling functions: util:time, util:memory, util:track, util:explain, util:profile

joewiz:
## Summary

Adds 5 new XQuery functions for profiling and debugging query performance. Includes XQSuite tests.

## What Changed

- `util:time($expr)` — measure execution time
- `util:memory($expr)` — measure memory allocation
- `util:track($expr)` — combined time + memory tracking
- `util:explain($expr)` — show query plan
- `util:profile($expr)` — detailed profiling with index usage

## Tests

- exist-core: 6,575 run, 0 failures, 0 errors
- Codacy: 0 new issues (6 fixed)

## Test plan

- [x] exist-core unit tests pass (6,575 run, 0 failures)
- [x] Profiling functions return valid data on sample queries

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ExpressionVisitor.java
+++ b/exist-core/src/main/java/org/exist/xquery/ExpressionVisitor.java
@@ -107,4 +107,26 @@ public interface ExpressionVisitor {
     void visitSimpleMapOperator(OpSimpleMap simpleMap);
 
     void visitWindowExpression(WindowExpr windowExpr);
+
+    /**
+     * Visit a map-constructor expression ({@code map { k1 : v1, k2 : v2 }}).
+     * The visitor controls element emission: the call site (i.e.
+     * {@link org.exist.xquery.functions.map.MapExpr#accept}) is expected
+     * to delegate here rather than calling {@code super.accept} + recursing
+     * manually, so that emitters like
+     * {@link org.exist.xquery.functions.util.QueryPlanSerializer} can wrap
+     * the entries inside the map element rather than emit them as siblings.
+     */
+    default void visitMapExpr(final org.exist.xquery.functions.map.MapExpr mapExpr) {
+        // Default: preserve the prior traversal semantics of
+        // MapExpr.accept (a generic visit() followed by walking through
+        // every key and value expression). Visitors that care about
+        // structure (e.g. emitters wrapping entries in a parent element)
+        // override this method.
+        visit(mapExpr);
+        for (int i = 0; i < mapExpr.getMappingCount(); i++) {
+            mapExpr.getMappingKey(i).accept(this);
+            mapExpr.getMappingValue(i).accept(this);
+        }
+    }
 }
--- a/exist-core/src/main/java/org/exist/xquery/ExpressionVisitor.java
+++ b/exist-core/src/main/java/org/exist/xquery/ExpressionVisitor.java
@@ -21,6 +21,8 @@
  */
 package org.exist.xquery;
 
+import org.exist.xquery.functions.map.MapExpr;
+
 /**
  * Defines a visitor to be used for traversing and analyzing the
  * expression tree.
@@ -111,13 +113,13 @@ public interface ExpressionVisitor {
     /**
      * Visit a map-constructor expression ({@code map { k1 : v1, k2 : v2 }}).
      * The visitor controls element emission: the call site (i.e.
-     * {@link org.exist.xquery.functions.map.MapExpr#accept}) is expected
+     * {@link MapExpr#accept}) is expected
      * to delegate here rather than calling {@code super.accept} + recursing
      * manually, so that emitters like
      * {@link org.exist.xquery.functions.util.QueryPlanSerializer} can wrap
      * the entries inside the map element rather than emit them as siblings.
      */
-    default void visitMapExpr(final org.exist.xquery.functions.map.MapExpr mapExpr) {
+    default void visitMapExpr(final MapExpr mapExpr) {
         // Default: preserve the prior traversal semantics of
         // MapExpr.accept (a generic visit() followed by walking through
         // every key and value expression). Visitors that care about
--- a/exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
@@ -53,10 +53,19 @@ import static javax.xml.XMLConstants.XML_NS_PREFIX;
  * util:explain('for $x in 1 to 10 where $x > 5 return $x * 2')
  * </pre>
  *
- * Returns an XML representation of the expression tree showing prolog
+ * <p>Returns an XML representation of the expression tree showing prolog
  * declarations (namespace declarations, module imports) followed by the
  * post-optimization expression body (FLWOR clauses, path expressions,
- * function calls, comparisons, etc.).
+ * function calls, comparisons, etc.).</p>
+ *
+ * <p>Inspired by Saxon's explain capabilities (which group output by XQuery
+ * construct rather than by AST class name) and the general "show me how the
+ * engine reads my query" idea behind PostgreSQL's EXPLAIN ANALYZE. BaseX
+ * also ships a {@code query:plan} function with a different output style
+ * (element names mirror internal Java class names: {@code GFLWOR},
+ * {@code IterPath}, {@code MapEntry}); eXist's util:explain uses XQuery
+ * construct names ({@code <for>}, {@code <path>}, {@code <map>/<entry>}) to
+ * stay close to how the developer thinks about the query.</p>
  */
 public class FunExplain extends BasicFunction {
 
--- a/exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
@@ -27,6 +27,7 @@ import org.exist.dom.memtree.DocumentImpl;
 import org.exist.dom.memtree.MemTreeBuilder;
 import org.exist.source.Source;
 import org.exist.xquery.*;
+import org.exist.xquery.Module;
 import org.exist.xquery.parser.XQueryLexer;
 import org.exist.xquery.parser.XQueryParser;
 import org.exist.xquery.parser.XQueryTreeParser;
@@ -199,8 +200,8 @@ public class FunExplain extends BasicFunction {
             builder.endElement();
         }
 
-        for (final Iterator<org.exist.xquery.Module> it = pContext.getRootModules(); it.hasNext(); ) {
-            final org.exist.xquery.Module module = it.next();
+        for (final Iterator<Module> it = pContext.getRootModules(); it.hasNext(); ) {
+            final Module module = it.next();
             if (module == null || module.isInternalModule()) {
                 continue;
             }
--- a/exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
@@ -1,0 +1,150 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.functions.util;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.exist.dom.QName;
+import org.exist.dom.memtree.DocumentImpl;
+import org.exist.dom.memtree.MemTreeBuilder;
+import org.exist.xquery.*;
+import org.exist.xquery.parser.XQueryLexer;
+import org.exist.xquery.parser.XQueryParser;
+import org.exist.xquery.parser.XQueryTreeParser;
+import org.exist.xquery.value.*;
+
+import antlr.collections.AST;
+
+import java.io.StringReader;
+
+/**
+ * Returns the compiled expression tree of an XQuery expression as XML.
+ * This is the core query visibility function — shows what the optimizer produces.
+ *
+ * <pre>
+ * util:explain('for $x in 1 to 10 where $x > 5 return $x * 2')
+ * </pre>
+ *
+ * Returns an XML representation of the expression tree showing FLWOR clauses,
+ * path expressions, function calls, comparisons, etc.
+ */
+public class FunExplain extends BasicFunction {
+
+    private static final Logger LOG = LogManager.getLogger(FunExplain.class);
+
+    public static final FunctionSignature[] signatures = {
+            new FunctionSignature(
+                    new QName("explain", UtilModule.NAMESPACE_URI, UtilModule.PREFIX),
+                    "Compiles the given XQuery expression and returns its expression tree as XML. " +
+                            "Shows the post-optimization query plan.",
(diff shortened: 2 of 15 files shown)


joewiz:

**CI state**: 7/9 checks pass. The 2 remaining failures (windows and macOS integration) are pre-existing test hangs unrelated to this PR — the surefire fork timeout fires and the CI job reports FAILURE. This PR adds only new code; no existing code is modified.

**Dependencies**: None. This PR is Wave 1 — it can be reviewed and merged independently.

line-o:
I tested this briefly today and have to initial findings:

- `util:time` and `util:memory` cannot be combined as they will each compile and run the query immediately
- passing in queries as strings feels awkward but maybe the simplest solution
- I am not sure if those measuring functions should return the results instead of the time and memory used together
- `util:explain` does not include AST nodes like namespace declarations and module imports
- `util:explain` fails with a cryptic error when the query fails to parse (I think...)

It would good to add intended use-case, three example queries and their expected outcomes here for human testing and review. This will help us come to an informed decision.

exist-core/src/main/java/org/exist/xquery/functions/util/FunIndexReport.java
```diff
@@ -0,0 +1,141 @@
...
+            // Execute to trigger index usage
+            path.eval(null, null);
+
+            // Serialize the per-query profiler stats as the index report
+            final PerformanceStats stats = profiler.getPerformanceStats();
+            context.pushDocumentContext();
+            try {
+                final MemTreeBuilder builder = context.getDocumentBuilder();
+                if (stats instanceof PerformanceStatsImpl) {
+                    ((PerformanceStatsImpl) stats).serialize(builder);
```
reinhapa: Use instanceof cast variable

exist-core/src/main/java/org/exist/xquery/functions/util/FunProfile.java
```diff
@@ -0,0 +1,217 @@
...
+     * BrokerPool stats) to ensure per-query isolation under concurrent load.
+     */
+    private Sequence serializeProfilerStats(final XQueryContext pContext) throws XPathException {
+        // Read from the per-query profiler's stats, not the global BrokerPool stats
+        final PerformanceStats stats = pContext.getProfiler().getPerformanceStats();
+        context.pushDocumentContext();
+        try {
+            final MemTreeBuilder builder = context.getDocumentBuilder();
+            if (stats instanceof PerformanceStatsImpl) {
+                ((PerformanceStatsImpl) stats).serialize(builder);
```
reinhapa: Use instanceof cast variable

reinhapa (changes requested):

joewiz:

@line-o thanks for the testing — going point-by-point. Pushed 219398b6cc to address all five.

**1. `util:time` + `util:memory` cannot be combined.** This was a real bug, not a design tradeoff. All three (`util:time`, `util:memory`, `util:track`) extended `BasicFunction`, so by the time `eval()` ran the argument expression had already been fully evaluated and the timer captured essentially zero. Switching them to `extends Function` with explicit `getArgument(0).eval(...)` inside the measurement block makes them measure real work, and as a result composition now works naturally:

```xquery
util:time(util:memory(collection("/db/data")//book[year > 2020]))
```

The outer `util:time` measurement spans the inner `util:memory` measurement, which spans the actual query evaluation. The label argument (when supplied) is still evaluated eagerly, outside the timing block, so it doesn't contaminate the measurement.

**2. Passing queries as strings feels awkward.** Agreed for the *expression* profilers — `util:time`, `util:memory`, `util:track` all take an `item()*` expression directly (no string), and after the fix above they actually work that way. `util:explain` and `util:profile` still take strings because they need to compile the query in a fresh isolated context (without polluting the caller's static namespace/variable bindings or the live profiler stats). I think that's the right tradeoff: lazy expression for ad-hoc instrumentation, string for full plan/profile inspection. Happy to add a function-arg variant of `util:explain` as a follow-up if you'd find it useful, but I don't want to widen the API in this PR.

**3. Should the measuring functions return the result instead of time/memory?** They already do — that was the original intent and the tests cover it (`prof:time-returns-result`, `prof:memory-returns-result`). They're pass-through wrappers that log the metric and return the value unchanged. If you want the metric *as a value*, use `util:track` (returns `map { time, memory, value, label? }`) or `util:profile` (returns `map { result, time, memory, plan, stats }`).

**4. `util:explain` does not include namespace declarations / module imports.** Fixed. The output now wraps the existing expression tree in `<body>` and prepends a `<prolog>` element:

```xml
<explain>
  <prolog>
    <namespace prefix="foo" uri="http://example.com/foo"/>
    <module-import prefix="foo" uri="http://example.com/foo" location="..."/>
  </prolog>
  <body>...</body>
</explain>
```

Built-in prefixes (`xs`, `fn`, `xsi`, `xml`, `local`, `err`, `exist`, `exerr`, `java`, `dbgp`, `xdt`) are filtered out — only what the user declared in the prolog appears. Internal modules are filtered out of `<module-import>` too.

**5. `util:explain` cryptic parse-error.** Fixed. The previous code caught `Exception` broadly and rewrapped everything as `"Parse error: " + e.getMessage()`, which obscured useful detail. The new code catches `RecognitionException` / `TokenStreamException` separately and surfaces `err:XPST0003` with the underlying parser message preserved (e.g. `"unexpected token: $x"`). New regression test `prof:explain-parse-error-clear` asserts the error code.

@reinhapa — also addressed your two `instanceof` cast comments by switching `FunIndexReport.java:121` and `FunProfile.java:206` to Java 21 pattern variables.

**Test plan.** All 102 tests in `xquery.util.UtilTests` pass (was 95, +7 new tests covering composition, prolog dumping, and the parse-error path). Full `xquery.xquery3.XQuery3Tests` suite is unchanged (978 pass, 9 skipped — same as develop). Will run the broader exist-core suite and post here if anything regresses.

exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
```diff
@@ -97,50 +124,105 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
...
+                continue;
+            }
+            final AttributesImpl atts = new AttributesImpl();
+            atts.addAttribute("", "prefix", "prefix", "CDATA", prefix);
+            atts.addAttribute("", "uri", "uri", "CDATA", entry.getValue());
+            builder.startElement("", "namespace", "namespace", atts);
+            builder.endElement();
+        }
+
+        for (final Iterator<org.exist.xquery.Module> it = pContext.getRootModules(); it.hasNext(); ) {
```
reinhapa: Address Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.Module' due to existing import 'org.exist.xquery.*'_

exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
```diff
@@ -97,50 +124,105 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
...
+            final org.exist.xquery.Module module = it.next();
```
reinhapa: Address Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.Module' due to existing import 'org.exist.xquery.*'_

reinhapa (changes requested):

exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `6838ebb0ac` — added an explicit `import org.exist.xquery.Module;` so the call sites can drop the qualifier (also at line 203). Note that the wildcard `import org.exist.xquery.*` alone does not let Codacy collapse this — `java.lang.Module` (from JPMS) is implicitly visible and the wildcard does not override it, so the FQN was actually load-bearing. The explicit import gives precedence over `java.lang.*` and keeps the call sites unqualified.

exist-core/src/main/java/org/exist/xquery/functions/util/FunExplain.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Same fix as line 202 — `6838ebb0ac` makes both call sites unqualified via the explicit import.

duncdrum (approved):
LGTM @reinhapa what do you think?

reinhapa (approved):

line-o:
I queried `util:explain("for $i at $p in //p[@id] return map { $p : $i }")` which returns in its body:

```xml
<body>
    <for variable="$i" line="1" column="6">
        <in>
            <path line="1" column="26">
                <expression type="RootNode"/>
                <step axis="descendant" test="{}p" line="1" column="19">
                    <predicate line="1" column="21">
                        <step axis="attribute" test="{}id" line="1" column="21"/>
                    </predicate>
                </step>
            </path>
        </in>
        <expression type="MapExpr" line="1" column="33"/>
        <variable name="$p" line="1" column="40"/>
        <variable name="$i" line="1" column="45"/>
    </for>
</body>
```

I would expect the last expression to contain both variables as children otherwise it is hard to understand what this is doing. Also they should be wrapped in key and value or something similar.

Maybe the first expression needs to do the same

```xml
<expression type="RootNode">
    <step axis="descendant" test="{}p" line="1" column="19">
        <predicate line="1" column="21">
            <step axis="attribute" test="{}id" line="1" column="21"/>
        </predicate>
    </step>
</expression>

File: exist-core/src/main/java/org/exist/xquery/functions/util/QueryPlanSerializer.java [hunk 5]
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    // === Helper methods ===
+
+    private void startElement(final String name) {
+        builder.startElement("", name, name, null);
+    }
+
+    private void startElement(final String name, final String[][] attrs) {
+        final org.xml.sax.helpers.AttributesImpl atts = new org.xml.sax.helpers.AttributesImpl();
```
reinhapa: Use import and simple Class name instead

exist-core/src/main/java/org/exist/xquery/functions/util/QueryPlanSerializer.java
```diff
@@ -0,0 +1,400 @@
...
+            expression.getExpression(i).accept(this);
+        }
+        endElement();
+    }
+
+    @Override
+    public void visitLocationStep(final LocationStep locationStep) {
+        startElement("step", mergeAttrs(
+                new String[][]{
+                        {"axis", org.exist.xquery.Constants.AXISSPECIFIERS[locationStep.getAxis()]},
```
reinhapa: Use static import

exist-core/src/main/java/org/exist/xquery/ExpressionVisitor.java
```diff
@@ -107,4 +107,26 @@ public interface ExpressionVisitor {
...
+    /**
+     * Visit a map-constructor expression ({@code map { k1 : v1, k2 : v2 }}).
+     * The visitor controls element emission: the call site (i.e.
+     * {@link org.exist.xquery.functions.map.MapExpr#accept}) is expected
+     * to delegate here rather than calling {@code super.accept} + recursing
+     * manually, so that emitters like
+     * {@link org.exist.xquery.functions.util.QueryPlanSerializer} can wrap
+     * the entries inside the map element rather than emit them as siblings.
+     */
+    default void visitMapExpr(final org.exist.xquery.functions.map.MapExpr mapExpr) {
```
reinhapa: Address Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.functions.map.MapExpr' due to existing same package import 'org.exist.xquery.*'_

reinhapa (changes requested):

joewiz:

@line-o addressed in `fd193f3b35` with a comprehensive approach (tests first, then implementation) rather than a one-off MapExpr patch. Three coordinated changes:

**1. `QueryPlanSerializer.visit(Expression)` fallback** — the default used to emit `<expression type="X"/>` and immediately close, so any expression walking children inside its own `accept()` rendered them as siblings of a self-closed parent. Made the default recurse via the standard `getSubExpressionCount` / `getSubExpression` contract. Specific `visit*()` methods still override entirely; this is purely a safety net.

**2. `visitMapExpr` hook on the visitor + MapExpr.accept delegation** — added a `visitMapExpr(MapExpr)` method to `ExpressionVisitor` (default preserves prior traversal). `MapExpr.accept` now delegates to that hook (matching `ForExpr`/`LetExpr`'s pattern: one visitor call, visitor owns element lifecycle). Exposed `getMappingCount` / `getMappingKey` / `getMappingValue` so structural visitors can iterate without reflection.

**3. `QueryPlanSerializer.visitMapExpr`** — emits `<map>` with `<entry><key>...</key><value>...</value></entry>` children, directly addressing your "should be wrapped in key and value or something similar" point.

For `RootNode`: it's a true AST leaf (the descendant step is a sibling, not a child, in PathExpr's step list), so the bare `<expression type="RootNode"/>` form is structurally accurate. The downstream descendant step + predicate render correctly under `<path>` as siblings — that's faithful to the AST shape.

**Tests** (new file `QueryPlanSerializerTest`, 8 cases covering MapExpr, RootNode-path, conditional, try-catch, function-call, simple-map, variable-reference, and a generic "no MapExpr-shape container self-closes" assertion).

**Drive-by**: `MapExpr.eval` was reassigning its `contextSequence` parameter (PMD `AvoidReassigningParameters`). Rewritten as an `effectiveContext` local — pre-existing but in scope per the strengthened "address all Codacy on touched files" rule.

**Results**: 1134 tests pass across `QueryPlanSerializerTest`, `XPathQueryTest`, `XQuery3Tests`, `MapsTests`. Codacy clean on all touched files.

Sample output for `for $i at $p in //p[@id] return map { $p : $i }`:

```xml
<for variable="$i" ...>
    <in>
        <path ...>
            <expression type="RootNode" ...>
            <step axis="descendant" test="{}p" ...>
                <predicate ...>
                    <step axis="attribute" test="{}id" .../>
                </predicate>
            </step>
        </path>
    </in>
    <map line="1" column="33">
        <entry>
            <key><variable name="$p" .../></key>
            <value><variable name="$i" .../></value>
        </entry>
    </map>
    <variable name="$p" .../>
    <variable name="$i" .../>
</for>
```

(The two trailing `<variable>` elements are emitted by `visitForExpression` for the loop's own positional + iteration variables — that's the existing for-loop rendering and unchanged. Let me know if you'd like those re-shaped too.)

While I had `util:explain` open, I also updated the FunExplain javadoc to credit the original design influences per the tasking that produced this PR — Saxon's explain output (which uses XQuery-construct element names rather than AST-class names) and PostgreSQL's EXPLAIN ANALYZE for the general visibility idea. BaseX has a `query:plan()` function that follows a different convention (element names mirror internal Java class names: `<GFLWOR>`, `<IterPath>`, `<MapEntry>`, `<VarRef>`); eXist's `util:explain` is deliberately the construct-name style. Happy to revisit that design choice if you'd prefer BaseX-style faithful AST mirroring, but for now the docstring documents the intentional difference.

exist-core/src/main/java/org/exist/xquery/ExpressionVisitor.java
duncdrum: @joewiz could you take another look

exist-core/src/main/java/org/exist/xquery/functions/util/QueryPlanSerializer.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in 1348ea27eb — imported `org.xml.sax.helpers.AttributesImpl` and use the simple class name here.

exist-core/src/main/java/org/exist/xquery/functions/util/QueryPlanSerializer.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in 1348ea27eb — added `import static org.exist.xquery.Constants.AXISSPECIFIERS;` and use `AXISSPECIFIERS` directly.

exist-core/src/main/java/org/exist/xquery/ExpressionVisitor.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in 1348ea27eb — added `import org.exist.xquery.functions.map.MapExpr;` and use the simple `MapExpr` name in the signature (and the javadoc link). Thanks for the nudge, @duncdrum — these three from the 06-01 round had slipped through; all addressed now. @reinhapa ready for another look when you have a moment (Codacy clean, exist-core compiles).

duncdrum (approved):

reinhapa (approved):

joewiz:
Thanks, all! As promised to @line-o on yesterday's Community Call, I've got an action item to provide a walkthrough of the functions to help everyone evaluate the functions and their outputs.

PR #6138, opened 2026-03-14
Improve W3C serialization compliance across all output methods

joewiz:
## Summary

Improve eXist-db's compliance with the [W3C XSLT and XQuery Serialization 3.1](https://www.w3.org/TR/xslt-xquery-serialization-31/) specification across all output methods (JSON, adaptive, XML, text, HTML, XHTML) and fix `fn:xml-to-json` for element node inputs.

**Depends on:** #6139 (XQuery 4.0 parser) — rebased on Parser; merge after it lands.

> **CI Note:** This branch inherits CI timeouts from #6139 (BrokerPool shutdown hangs in `DeadlockIT`, `RemoveCollectionIT`, `serialize-node`). These are pre-existing issues unrelated to serialization changes and will resolve when merged in order (#6139 first, then this PR).

## What Changed

| File | Changes |
|------|---------|
| `JSONSerializer.java` | Enable forward-slash escaping; handle INF/NaN/negative-zero per QT4; fix inverted `allow-duplicate-names`; add SERE0022 duplicate-key detection |
| `XQuerySerializer.java` | Add item-separator support for XML/text methods; restore backwards-compat routing for single element/document nodes in JSON method (RESTXQ) |
| `JSON.java` | Add option type validation for `liberal` (boolean) and `duplicates` (string) |
| `AdaptiveWriter.java` | Remove `map` prefix (`{...}` not `map{...}`); fix double INF/NaN to use text not Unicode symbols |
| `XMLWriter.java` | Standalone declaration output; CDATA section output for `cdata-section-elements`; CR and LINE SEPARATOR character reference escaping; `&{` attribute escaping hook |
| `IndentingXMLWriter.java` | `suppress-indentation` parameter |
| `Option.java` | Allow `Q{namespace}local` URIQualifiedName in `declare option` |
| `AbstractSerializer.java` | Default `html-version` to 5.0; `output:version` → `html-version` mapping |
| `XHTMLWriter.java` | `include-content-type` meta tag (first child of `<head>`); boolean attribute minimization; XHTML content-type uses `http-equiv` form |
| `HTML5Writer.java` | HTML5 processing instruction format |
| `XHTML5Writer.java` | DOCTYPE PUBLIC/SYSTEM support |
| `FunSerialize.java` | SEPM0009 parameter validation |
| `FunXmlToJson.java` | DOM traversal rewrite; map key validation and duplicate detection |
[...]

PR #6219, opened 2026-04-06
Improve W3C serialization compliance across all output methods

joewiz:
## Summary

Fixes and improvements to XML, HTML, XHTML, JSON, text, and adaptive serialization per W3C Serialization 3.1.

## What Changed

- **XMLWriter**: Fixed empty namespace undeclaration handling, xml prefix skip
- **XHTMLWriter**: Self-closing meta tags in XHTML mode (fixes URL rewrite pipeline regression)
- **JSONSerializer**: escape-solidus default to "no" (spec-compliant)
- **HTML5Writer**: Proper void element handling
- **AdaptiveWriter**: Map serialization format
- **parameter-document**: Serialization parameter document support
- **normalization-form**: Unicode normalization parameter
- **SVG/MathML**: Namespace prefix normalization for XHTML5

## Tests

- exist-core: 6,564 tests, 7 failures, 0 errors, 115 skipped
  - 7 failures are pre-existing namespace undeclaration bugs on `develop` (not regressions from this branch)
  - 0 new failures introduced
- URLRewriteViewPipelineTest: 2/2 pass
- Codacy: 0 new issues (4 fixed)

## Supersedes

- PR #6138

## Test plan

- [x] exist-core unit tests pass (7 pre-existing failures, 0 new failures)
- [x] URL rewrite pipeline test passes (eXide regression — 2/2)
- [x] HTML5 serialization tests pass
- [x] JSON serialization tests pass

### DIFF
--- a/exist-core/src/main/java/org/exist/util/serializer/HTML5Writer.java
+++ b/exist-core/src/main/java/org/exist/util/serializer/HTML5Writer.java
@@ -118,6 +118,13 @@ public class HTML5Writer extends XHTML5Writer {
         BOOLEAN_ATTRIBUTE_NAMES.add("willValidate");
     }
 
+    private static final ObjectSet<String> BOOLEAN_ATTRIBUTE_NAMES_LOWER = new ObjectOpenHashSet<>(BOOLEAN_ATTRIBUTE_NAMES.size());
+    static {
+        for (final String n : BOOLEAN_ATTRIBUTE_NAMES) {
+            BOOLEAN_ATTRIBUTE_NAMES_LOWER.add(n.toLowerCase(java.util.Locale.ROOT));
+        }
+    }
+
     private static final ObjectSet<String> EMPTY_TAGS = new ObjectOpenHashSet<>(31);
     static {
         EMPTY_TAGS.add("area");
@@ -186,18 +193,20 @@ public class HTML5Writer extends XHTML5Writer {
     public void attribute(String qname, CharSequence value) throws TransformerException {
         // Strip prefix for the meta-dedup redundancy check
         final int colon = qname.indexOf(':');
-        noteMetaAttribute(colon < 0 ? qname : qname.substring(colon + 1), value);
+        final String localName = colon < 0 ? qname : qname.substring(colon + 1);
+        noteMetaAttribute(localName, value);
+        final CharSequence effectiveValue = maybeEscapeUriHtml5(localName, value);
         try {
             if(!tagIsOpen) {
-                characters(value);
+                characters(effectiveValue);
                 return;
             }
             final Writer writer = getWriter();
             writer.write(' ');
             writer.write(qname);
-            if (!(BOOLEAN_ATTRIBUTE_NAMES.contains(qname) && qname.contentEquals(value))) {
+            if (!isBooleanAttributeMatch(qname, effectiveValue)) {
                 writer.write("=\"");
-                writeChars(value, true);
+                writeChars(effectiveValue, true);
                 writer.write('"');
             }
         } catch(final IOException ioe) {
@@ -208,9 +217,11 @@ public class HTML5Writer extends XHTML5Writer {
     @Override
     public void attribute(QName qname, CharSequence value) throws TransformerException {
         noteMetaAttribute(qname.getLocalPart(), value);
+        final String localPart = qname.getLocalPart();
+        final CharSequence effectiveValue = maybeEscapeUriHtml5(localPart, value);
         try {
             if(!tagIsOpen) {
-                characters(value);
+                characters(effectiveValue);
                 return;
                 // throw new TransformerException("Found an attribute outside an
                 // element");
@@ -221,11 +232,10 @@ public class HTML5Writer extends XHTML5Writer {
                 writer.write(qname.getPrefix());
                 writer.write(':');
             }
-            final String localPart = qname.getLocalPart();
             writer.write(localPart);
-            if (!(BOOLEAN_ATTRIBUTE_NAMES.contains(localPart) && localPart.contentEquals(value))) {
+            if (!isBooleanAttributeMatch(localPart, effectiveValue)) {
                 writer.write("=\"");
-                writeChars(value, true);
+                writeChars(effectiveValue, true);
                 writer.write('"');
             }
         } catch(final IOException ioe) {
@@ -233,9 +243,53 @@ public class HTML5Writer extends XHTML5Writer {
         }
     }
 
+    /**
+     * URI-attribute escaping for the HTML5 writer. Mirrors
+     * {@link XHTMLWriter#shouldEscapeUriAttribute(String, String)} but unwraps
+     * the prefixed form of {@link #currentTag} so the (element, attribute)
+     * lookup uses local names only.
+     */
+    private CharSequence maybeEscapeUriHtml5(final String attrLocal, final CharSequence value) {
+        if (currentTag == null) {
+            return value;
+        }
+        final String elementLocal = currentTag.contains(":")
+                ? currentTag.substring(currentTag.indexOf(':') + 1)
+                : currentTag;
+        if (!shouldEscapeUriAttribute(elementLocal, attrLocal)) {
+            return value;
+        }
+        return escapeUriAttribute(value);
+    }
+
+    /**
+     * HTML5 boolean attribute minimization: emit just the bare name when the
+     * value is empty or matches the attribute name case-insensitively
+     * (per W3C XSLT/XQuery Serialization 3.1, section 7.2.2).
+     */
+    private static boolean isBooleanAttributeMatch(final String name, final CharSequence value) {
+        if (!BOOLEAN_ATTRIBUTE_NAMES_LOWER.contains(name.toLowerCase(java.util.Locale.ROOT))) {
+            return false;
+        }
+        if (value == null || value.length() == 0) {
+            return true;
+        }
+        return name.equalsIgnoreCase(value.toString());
+    }
+
     @Override
     public void namespace(String prefix, String nsURI) throws TransformerException {
-        // no namespaces allowed in HTML5
+        // HTML5 elements never carry an explicit xmlns since the parser puts
+        // them in the HTML namespace implicitly. Foreign content (anything
+        // outside the XHTML namespace, e.g. SVG, MathML, custom XML) keeps
+        // its namespace declarations so the receiver can re-parse it as XML.
+        if (nsURI == null || nsURI.isEmpty()) {
+            return;
+        }
+        if (org.exist.Namespaces.XHTML_NS.equals(nsURI)) {
+            return;
+        }
+        super.namespace(prefix, nsURI);
     }
 
     @Override
@@ -246,6 +300,11 @@ public class HTML5Writer extends XHTML5Writer {
                 if (isEmpty) {
                     if (isEmptyTag(currentTag)) {
                         w.write('>');
+                    } else if (isForeignContent()) {
+                        // Foreign content (SVG, MathML, custom XML namespace)
+                        // embedded in HTML5 is serialized with XML self-close
+                        // syntax so the receiver can re-parse it as XML.
+                        w.write("/>");
                     } else {
                         // Coalesce ">", "</", tag, ">" into 2 writer calls instead of 4
                         w.write("></");
@@ -260,6 +319,17 @@ public class HTML5Writer extends XHTML5Writer {
         } catch (final IOException ioe) {
             throw new TransformerException(ioe.getMessage(), ioe);
         }
+    }
+
+    /**
+     * The current element is "foreign content" when its namespace is neither
+     * the XHTML namespace nor the empty (no-namespace) HTML namespace; that
+     * is the trigger for XML-style self-closing per HTML5's foreign-content
+     * serialization rule.
+     */
+    private boolean isForeignContent() {
+        final String ns = currentElementNamespaceURI();
+        return ns != null && !ns.isEmpty() && !org.exist.Namespaces.XHTML_NS.equals(ns);
     }
 
     @Override
--- a/exist-core/src/main/java/org/exist/util/serializer/HTML5Writer.java
+++ b/exist-core/src/main/java/org/exist/util/serializer/HTML5Writer.java
@@ -248,18 +248,23 @@ public class HTML5Writer extends XHTML5Writer {
     }
 
     @Override
-    public void processingInstruction(String target, String data) throws TransformerException {
+    public void processingInstruction(final String target, final String data) throws TransformerException {
+        // QT4 PR2372: HTML5 has no PI syntax, so the serializer renders
+        // processing instructions as comments of the form `<!--?target data?-->`,
+        // matching the HTML5 parser's coercion of `<?...?>` content.
         try {
-            closeStartTag(false);
+            if (tagIsOpen) {
+                closeStartTag(false);
+            }
             final Writer writer = getWriter();
-            writer.write("<?");
+            writer.write("<!--?");
             writer.write(target);
             if (data != null && !data.isEmpty()) {
                 writer.write(' ');
                 writer.write(data);
             }
-            writer.write('>');
-        } catch (IOException e) {
+            writer.write("?-->");
+        } catch (final IOException e) {
             throw new TransformerException(e.getMessage(), e);
         }
     }
--- a/exist-core/src/main/java/org/exist/util/serializer/HTML5Writer.java
+++ b/exist-core/src/main/java/org/exist/util/serializer/HTML5Writer.java
@@ -247,10 +247,43 @@ public class HTML5Writer extends XHTML5Writer {
     }
 
     @Override
+    public void processingInstruction(String target, String data) throws TransformerException {
+        try {
+            closeStartTag(false);
+            final Writer writer = getWriter();
+            writer.write("<?");
+            writer.write(target);
+            if (data != null && !data.isEmpty()) {
+                writer.write(' ');
+                writer.write(data);
+            }
+            writer.write('>');
+        } catch (IOException e) {
+            throw new TransformerException(e.getMessage(), e);
+        }
+    }
+
+    @Override
     protected boolean needsEscape(char ch) {
         if (RAW_TEXT_ELEMENTS.contains(currentTag)) {
             return false;
         }
         return super.needsEscape(ch);
     }
+
+    @Override
+    protected boolean needsEscape(final char ch, final boolean inAttribute) {
+        // In raw text elements (script, style), suppress escaping for TEXT content only.
+        // Attribute values must always be escaped, even on raw text elements.
+        if (!inAttribute && RAW_TEXT_ELEMENTS.contains(currentTag)) {
+            return false;
+        }
+        // For attributes, always return true (bypass the 1-arg override
+        // which returns false for all script/style content)
+        if (inAttribute) {
+            return true;
+        }
+        return super.needsEscape(ch, inAttribute);
+    }
+
 }
--- a/exist-core/src/main/java/org/exist/util/serializer/HTML5Writer.java
+++ b/exist-core/src/main/java/org/exist/util/serializer/HTML5Writer.java
@@ -382,10 +382,7 @@ public class HTML5Writer extends XHTML5Writer {
         // Mirror the per-char rule above: TEXT content inside script/style is
         // raw text and never needs escaping. Lets writeChars() bulk-stream
         // the entire block in one Writer.write() call.
-        if (!inAttribute && RAW_TEXT_ELEMENTS.contains(currentTag)) {
-            return false;
-        }
-        return true;
+        return inAttribute || !RAW_TEXT_ELEMENTS.contains(currentTag);
     }
 
 }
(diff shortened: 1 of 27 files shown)


joewiz:

**CI state**: 5/9 checks pass. Of the 4 failures:
- 1 (ubuntu unit) is a surefire fork timeout — a test hung and the process was killed. This is the known pre-existing infrastructure hang, not a failure in the serialization code. The unit test run completed 4,923 tests with 0 failures before the timeout fired.
- 1 (W3C XQuery Test Suite) likely timed out for the same reason
- 2 are pre-existing integration test hangs (windows, macOS)

**Dependencies**: Wave 5 — independent, can merge in any order after Wave 4.

duncdrum:
needs a rebase

joewiz:

Rebased onto develop (force-pushed). The only conflict was in `exist-core/pom.xml` — develop added `--add-modules jdk.incubator.vector --enable-native-access=ALL-UNNAMED` to the failsafe argLine while this branch had added the `forkedProcessTimeoutInSeconds` and `excludes` config; resolved by combining both.

duncdrum:
there is a test failure: 
```
Error: core] [ERROR] Failures: 
Error: core] [ERROR]   XQuery failure: xml-to-json.xql:359 xml-to-json-wrong-namespace
	at (xml-to-json.xql:359) expected:<[Error: FOJS0006]> but was:<[map{"value":"{}"}]>
Error: core] [ERROR]   XQuery failure: xml-to-json.xql:366 xml-to-json-wrong-namespace-non-empty
	at (xml-to-json.xql:366) expected:<[Error: FOJS0006]> but was:<[map{"value":"{}"}]>
[exist-core] [INFO] 
Error: core] [ERROR] Tests run: 5005, Failures: 2, Errors: 0, Skipped: 68
```
 the 15 codacy issues all seem actionable to me. could you take another look please

joewiz:

Thanks @duncdrum — both issues addressed in 70120853b9:

**2 test failures fixed**: `fn:xml-to-json()` was missing namespace validation in the DOM-based code path. Elements not in `http://www.w3.org/2005/xpath-functions` now raise FOJS0006 as spec'd. The older stream-based method already had this check; the newer DOM method was missing it.

**15 Codacy PMD issues fixed**:
- `FunXmlToJson.java` (4): moved field to top, added default to switch, replaced FQNs
- `XHTMLWriter.java` (5): moved 4 field declarations to top, replaced FQN
- `XQueryContext.java` (5): replaced FQNs with imports
- `HTML5FragmentTest.java` (1): removed unused variable

Test results: 6,605 tests, 0 failures.

exist-core/src/main/java/org/exist/xquery/util/SerializerUtils.java
reinhapa: Convert to switch expression and eliminate use of `value` as it is not used outside of the switch itself

exist-core/src/main/java/org/exist/xquery/Option.java
```diff
@@ -60,7 +60,9 @@ public Option(QName qname, String contents)  throws XPathException {
     }
 	
 	public Option(final Expression expression, QName qname, String contents)  throws XPathException {
-		if (qname.getPrefix() == null || qname.getPrefix().isEmpty())
+		// Options must be in a namespace: either via prefix or via URIQualifiedName Q{uri}local
+		if ((qname.getPrefix() == null || qname.getPrefix().isEmpty())
+				&& (qname.getNamespaceURI() == null || qname.getNamespaceURI().isEmpty()))
```
reinhapa: May introduce local variables for `qname.getPrefix()` and `qname.getNamespaceURI()` for improved readability.

exist-core/src/test/java/org/exist/http/urlrewrite/URLRewriteViewPipelineTest.java
reinhapa, line 1: Using text blocks for XQ values would improve readability

exist-core/src/test/java/org/exist/util/serializer/HtmlSerializerBenchmark.java
reinhapa, line 1: For benchmarks we may want to take a look into JMH (Java Microbenchmark Harness)

exist-core/src/main/java/org/exist/util/serializer/TEXTWriter.java
```diff
@@ -211,8 +211,28 @@ protected void writeChars(final CharSequence s, final boolean inAttribute) throw
...
+            final char[] buf = new char[len];
+            sb.getChars(start, end, buf, 0);
+            writer.write(buf, 0, len);
+        } else {
+            final char[] buf = new char[len];
+            for (int i = 0; i < len; i++) {
+                buf[i] = ch.charAt(start + i);
+            }
+            writer.write(buf, 0, len);
         }
```
reinhapa: Maybe use switch expression

```suggestion
        switch (ch) {
            case String stringValue -> writer.write(stringValue, start, len);
            case StringBuilder stringBuilder -> {
                final char[] buf = new char[len];
                stringBuilder.getChars(start, end, buf, 0);
                writer.write(buf, 0, len);
            }
            case StringBuffer stringBuffer -> {
                final char[] buf = new char[len];
                stringBuffer.getChars(start, end, buf, 0);
                writer.write(buf, 0, len);
            }
            default -> {
                final char[] buf = new char[len];
                for (int i = start; i < end; i++) {
                    buf[i]=ch.charAt(i);
                }
                writer.write(buf, 0, len);
            }
        }
```

reinhapa (changes requested):

exist-core/src/test/java/org/exist/util/serializer/HtmlSerializerBenchmark.java
duncdrum: we already have microbenchmarks wired up afaik

joewiz:

@reinhapa All review items addressed in 06fe273825:

1. **SerializerUtils.java:638** — Converted to switch expression with arrow syntax; eliminated the outer `value` variable; merged STRING/DECIMAL/INTEGER cases into a single arm.
2. **Option.java:65** — Extracted `prefix` and `namespaceURI` local variables for readability.
3. **URLRewriteViewPipelineTest.java** — Converted all XQ string concatenation to Java text blocks.
4. **TEXTWriter.java:236** — Converted instanceof chain to Java 21 pattern-matching switch (`case String s ->`, `case CharSlice cs ->`, etc.).

Re: @duncdrum's note about JMH — agreed, the existing microbenchmark infrastructure is available. The `HtmlSerializerBenchmark` uses a JUnit-based approach for simplicity (runs as part of the test suite), but a JMH version would give more stable numbers. Not blocking for this PR.

All 30 serializer tests pass (HtmlSerializerBenchmark, SerializerPoolTest, DOMSerializerTest, XIncludeSerializerTest, SerializationTest).

exist-core/src/main/java/org/exist/util/serializer/XMLWriter.java
```diff
@@ -653,9 +1025,42 @@ protected void writeChars(final CharSequence s, final boolean inAttribute) throw
...
+            writer.write(charBuffer, 0, len);
+        } else {
+            // Generic CharSequence — copy then bulk-write
+            final char[] buf = ensureCharBuffer(len);
+            for (int i = 0; i < len; i++) {
+                buf[i] = ch.charAt(start + i);
+            }
+            writer.write(buf, 0, len);
+        }
+    }
```
reinhapa: This method looks the similar to the one in `TEXTWriter` may this could be pulled up to the `XMLWriter` and be re-used here instead of duplicating the logic

exist-core/src/main/java/org/exist/util/serializer/XMLWriter.java
```diff
@@ -653,9 +1025,42 @@ protected void writeChars(final CharSequence s, final boolean inAttribute) throw
...
+        if (charBuffer == null || charBuffer.length < minLen) {
+            // Grow to a power-of-two-ish size to amortize re-allocation
+            int n = charBuffer == null ? 256 : charBuffer.length;
+            while (n < minLen) {
+                n <<= 1;
+            }
+            charBuffer = new char[n];
         }
+        return charBuffer;
     }
```
reinhapa, line 1049: what about concurrency issues as we create/use an instance field here...

reinhapa (changes requested):

joewiz:

@reinhapa Thanks for the careful review. Addressing both points:

1. **writeCharSeq duplication** — You're right, `TEXTWriter extends XMLWriter`, so the method can be pulled up. `TEXTWriter.writeCharSeq()` is identical to `XMLWriter.writeCharSeq()` — it was duplicated during the optimization work. Will move it to `XMLWriter` only and let `TEXTWriter` inherit.

2. **charBuffer concurrency** — The `charBuffer` field is safe because `XMLWriter` instances are borrowed from `SerializerPool` (backed by Commons Pool2 `GenericKeyedObjectPool`). The pool guarantees exclusive access: each thread borrows its own instance, uses it, and returns it. No two threads share the same `XMLWriter` instance concurrently. The instance field avoids re-allocating the buffer on every `writeCharSeq` call — it grows once and is reused across serializations within the same borrow cycle. Will add a comment explaining this.

joewiz:

Both review points addressed in 4bcc33895a:

1. **Deduplicated `writeCharSeq`**: Removed `TEXTWriter.writeCharSeq()` and promoted `XMLWriter.writeCharSeq()` from `private` to `protected` so the subclass inherits it. Bonus: the inherited version uses the pooled `charBuffer` instead of allocating a fresh `char[]` per call, so this is also a small allocation win on the text-output path.

2. **Documented `charBuffer` thread-safety**: Added a javadoc on the field explaining that exclusive access is guaranteed by `SerializerPool` (Commons Pool2), so the unsynchronised field is safe by construction.

All target serialization tests still pass (1022 tests, 0 failures): `HtmlSerializerBenchmark`, `SerializerPoolTest`, `DOMSerializerTest`, `XIncludeSerializerTest`, `SerializationTest`, `XQuery3Tests`.

exist-core/src/main/java/org/exist/util/serializer/XMLWriter.java
```diff
@@ -589,36 +885,96 @@ protected void writeDoctype(final String rootElement) throws TransformerExceptio
...
+            writeCharSeq(text, pos, i);
             // writer.write(s.subSequence(pos, i).toString());
             
             if (i >= len) {
                 return;
             }
             
-            if(needsEscape(ch)) {
+            if(needsEscape(ch, inAttribute)) {
                 switch(ch) {
```
reinhapa, line 978: may use switch expression here

exist-core/src/main/java/org/exist/util/serializer/XMLWriter.java
```diff
@@ -652,12 +1013,52 @@ protected void writeChars(final CharSequence s, final boolean inAttribute) throw
...
+            sb.getChars(start, end, ensureCharBuffer(len), 0);
+            writer.write(charBuffer, 0, len);
+        } else {
+            // Generic CharSequence — copy then bulk-write
+            final char[] buf = ensureCharBuffer(len);
+            for (int i = 0; i < len; i++) {
+                buf[i] = ch.charAt(start + i);
+            }
+            writer.write(buf, 0, len);
         }
```
reinhapa: could be rewritten to a switch expression

reinhapa (changes requested):

reinhapa (approved):

exist-core/src/main/java/org/exist/util/serializer/HTML5Writer.java
reinhapa: Address "Avoid unnecessary if..then..else statements when returning booleans" comment of Codacy

exist-core/src/main/java/org/exist/util/serializer/XHTMLWriter.java
reinhapa, line 1: Address "Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes." Codacy issues

exist-core/src/main/java/org/exist/util/serializer/XHTMLWriter.java
```diff
@@ -339,11 +569,61 @@ private boolean isHtml5Version() {
...
+     *   <li>doctype-system absent, html method, doctype-public set: emit DOCTYPE PUBLIC</li>
+     *   <li>doctype-system absent, html-version &ge; 5: emit {@code <!DOCTYPE html>}</li>
+     *   <li>otherwise: no DOCTYPE</li>
+     * </ul>
+     *
+     * Only emitted when the root element is {@code html} (case-insensitive); for
+     * fragments rooted on any other element the DOCTYPE is suppressed.
+     */
+    @Override
+    protected void writeDoctype(final String rootElement) throws TransformerException {
```
reinhapa, line 587: Adress issue "The method 'writeDoctype(String)' has an NPath complexity of 320, current threshold is 200" of Codacy

reinhapa (changes requested):

joewiz:

@reinhapa Addressed all 3 Codacy items in e190bcd:

1. **HTML5Writer.needsEscaping**: collapsed the if/return into a single boolean expression.
2. **XHTMLWriter** field ordering: hoisted all fields (HEX, EMPTY_TAGS, INLINE_TAGS, the meta-buffering state, SVG_NS/MATHML_NS, RAW_TEXT_ELEMENTS_HTML, etc.) above the methods/constructors so the class passes `FieldDeclarationsShouldBeAtStartOfClass`.
3. **XHTMLWriter.writeDoctype** NPath complexity (320 → under 200): extracted `isHtmlRoot`, `getDoctypeProperty`, and `emitHtmlDoctype` helpers.

While I was at it, Codacy also flagged two more findings on the touched files which I cleaned up in the same commit:

- `XHTMLWriter.maybeEscapeUri` had a redundant nested null guard (`CollapsibleIfStatements`); the `!isHtmlMethod()` leg never triggered an early return, so the gate was effectively just the `currentTag` null check. Collapsed to a single `if`.
- `XHTMLWriter.shouldUseCdataSections` had another `SimplifyBooleanReturns` opportunity in the html-method branch — simplified.

Re-ran `codacy-cli analyze --tool pmd` against both files: clean. All 31 HTML/XHTML serializer JUnit tests still pass.

reinhapa (approved):

joewiz:

Rebased `v2/serialization-compliance` onto current `develop` (`4f09d0acccd3162ff6de95ba7add6c64b825cd89`). New tip: `8431f2869f`.

**Reasoning**: a previous cherry-pick attempt to extract this branch into a fresh develop-based branch hit conflicts in `FunSerialize.java` and `JSONSerializer.java` that needed v2-branch-history context for clean resolution. Rebasing the source first restores that context and resolves cleanly per-commit.

**Two commits dropped** as already-applied on develop (same patch content):
- `01d71e23ab` "W3C-compliant JSON output method for nodes and solidus" -- already on develop as `5c407e54f3`
- `09242b6628` "Auto-detect legacy XML-to-JSON convention via json: prefix" -- already on develop as `e247d9aa3a`

26 commits remain (down from 28).

**Conflict resolutions** (where develop had shipped overlapping or alternative fixes):

- `FunSerialize.java` (was: `ab9d6d446a` SERE0024-specific catch): kept develop's broader `SER[EPM]\d{4}` regex from `e5abb798b9`. The regex superset covers the SERE0024 case already.
- `JSONSerializer.java` (was: `b3dc16e1b6` SERE0022 manual duplicate-detection rewrite + JSON-lines): kept the JSON-lines feature (net-new, 3.1-mandatory); dropped v2's manual duplicate-detection rewrite in favour of develop's `STRICT_DUPLICATE_DETECTION` Jackson-flag approach from `c94ee102d6` (spec-aligned). Removed the now-dead `seenKeys` HashSet and `allowDuplicateNames` field. Imports cleaned up.
- `XQuerySerializer.java` (was: `0064090270` if/else style): kept develop's `useLegacySerializer` early-return pattern but pointed it at `serializeXMLDirect` (the new helper that this commit introduces) so the legacy JSON path bypasses item-separator routing.
- `FunXmlToJson.java` (was: `70120853b9` logger style): kept develop's explicit `getLogger(FunXmlToJson.class)` plus v2's new `JSON_ELEMENT_NAMES` constant.

**Verification**:
- `mvn install -pl exist-core -am -DskipTests` -- BUILD SUCCESS
- Targeted serialization JUnit gate (`-Dtest='*Serialize*,*Output*,*Json*,*Xml2Json*,*Json2Xml*,*Format*,*Adaptive*'`) -- 27/27 pass, 0 failures, 0 errors
- Full `mvn test -pl exist-core` lock-script gate -- aborted: stale-lock preemption by a parallel local session at the 30-minute mark introduced BrokerPool "Database is read-only" contention failures unrelated to serialization. CI will provide a clean run.
- XQTS HEAD smoke test deferred: this branch carries Saxon 9.9, while the bundled XQTS runner JAR uses Saxon 12 (per `project_xqts_runner_saxon_dependency.md`). XQTS verification will happen at next-v3 integration time, where Saxon 12 is present.

CI on the new tip should run green. Once review is complete, the planned extraction to a 7.0-bound PR off develop (per the 2026-05-10 v2/* extraction audit) becomes straightforward -- the cherry-picks should now be clean.

joewiz:

Per the 2026-05-10 v2/* extraction audit, the W3C-mandated serialization-compliance work in this branch is XQ 3.1-mandatory and has no dependency on Saxon 12, the new parser, or XQ 4.0 grammar extensions. All 26 commits cherry-pick cleanly onto current `develop` (`4f09d0acccd`). The 7.0-bound extraction is now open as #6346.

Verified deltas against the 2026-05-10 XQ 3.1 baseline: +86 passes on serialization-mandated test sets (fn-format-date +35, fn-format-dateTime +15, fn-format-time +9, fn-format-number +8, misc-Serialization +12, fn-parse-json +7), plus spillover gains on element-constructor / namespace test sets driven by the same changes. Full diff is byte-identical to this branch except for one follow-up `[test]` commit correcting two stale `serialize.xql` assertions that expected the legacy XHTML `<meta http-equiv="Content-Type">` form instead of the HTML5 short `<meta charset>` form actually emitted by `XHTMLWriter.writeContentTypeMeta()` since the dedup work in `[bugfix] Suppress duplicate Content-Type/charset meta in HTML/XHTML head`.

CSV serialization was deliberately excluded from the extraction — it's not in any W3C serialization spec (eXist borrows it from BaseX) and the audit confirmed no CSV files appear in #6219's diff. If post-7.0 we want to keep CSV as a distinct eXist extension, it can land separately.

Recommend closing this PR as superseded once #6346 merges.

exist-core/src/main/java/org/exist/util/serializer/XQuerySerializer.java
reinhapa, line 67: convert to switch expression

exist-core/src/main/java/org/exist/util/serializer/XQuerySerializer.java
```diff
@@ -72,17 +73,167 @@ public void serialize(final Sequence sequence, final int start, final int howman
...
+            if (type == Type.MAP_ITEM || type == Type.FUNCTION) {
+                throw new SAXException("err:SENR0001 Cannot serialize a " +
+                        Type.getTypeName(type) + " with the XML or text output method");
+            }
+        }
+    }
+
+    private boolean isCanonical() {
+        final String v = outputProperties.getProperty(EXistOutputKeys.CANONICAL);
+        return "yes".equals(v) || "true".equals(v) || "1".equals(v);
```
reinhapa, line 115: Should we not compare `yes` / `true`  case insensitive?

exist-core/src/main/java/org/exist/util/serializer/XQuerySerializer.java
reinhapa, line 1: Use imports instead of canonical class names....

exist-core/src/main/java/org/exist/xquery/functions/fn/FunSerialize.java
```diff
@@ -141,6 +146,60 @@ private static boolean isSerializationParametersElement(final Item item) {
...
+     * Check if a serialization boolean parameter value is true.
+     * W3C Serialization 3.1 accepts "yes", "true", "1" (with optional whitespace) as true.
+     */
+    private static boolean isBooleanTrue(final String value) {
+        if (value == null) {
+            return false;
+        }
+        final String trimmed = value.trim();
+        return "yes".equals(trimmed) || "true".equals(trimmed) || "1".equals(trimmed);
+    }
```
reinhapa, line 159: Looks like code duplication of `org.exist.util.serializer.XQuerySerializer` line 115 (test part) would make sense to extract this method to an utility class as static method and search for similar occurrences. If mainly needed within functions it would make sense to move it to the `org.exist.xquery.Function` class.

reinhapa (changes requested):

joewiz:

Closing as superseded: PR #6346 (the 7.0-bound extraction of this work onto current `develop`) was merged on 2026-05-31. Per the audit posted upthread, #6346 contained the same 26 commits as this branch (byte-identical apart from a single `[test]` follow-up correcting two stale `serialize.xql` assertions for the HTML5 short `<meta charset>` form), plus a 3-commit follow-up wave on the same branch addressing review feedback (XQ4 strip per @line-o, Juri's review fixes, the orphan-fork-timeout revert).

CSV serialization was deliberately excluded from #6346 — not in any W3C serialization spec, eXist borrows it from BaseX, and the extraction audit confirmed no CSV files appeared in this PR's diff. If we want to keep CSV as a distinct eXist extension post-7.0, it can land separately.

Thanks @reinhapa for the Codacy and review passes that got incorporated into the extracted PR.

Issue #2182, opened 2018-10-01
Fragment selection skips namespace declaration for xsi:type value


ahenket:
When you select a piece of XML , all relevant namespace declarations are copied into the fragment, except when the @xsi:type value has one. The net result is now an invalid fragment.

Tested with eXist-db 2.2 and 4.4.0 on macOS Mojave, you may run the following code to reproduce:

```
xquery version "3.0";

let $xml    :=
    <MCCI_IN200101 xmlns="urn:hl7-org:v3" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:hl7nl="urn:hl7-nl:v3">
        <effectiveTime xsi:type="hl7nl:PIVL_TS"/>
    </MCCI_IN200101>

return
    $xml/*
```

Actual result:

```<effectiveTime xmlns="urn:hl7-org:v3" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="hl7nl:PIVL_TS"/>```

Expected result:
```<effectiveTime xmlns="urn:hl7-org:v3" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:hl7nl="urn:hl7-nl:v3" xsi:type="hl7nl:PIVL_TS"/>```

joewiz:
I'm genuinely unsure where in the spec the feature that would yield the expected outcome is stated or implied. Is it the static typing feature, https://www.w3.org/TR/xquery-31/#id-static-typing-feature?

For comparison:

- BaseX returns the same actual result as eXist.
- Saxon returns the expected result.

duncdrum:
I m very much leaning to Saxon being right, and both eXist and BaseX being wrong. @adamretter thoughts?

joewiz:
@ahenket Thank you for that reference. I actually meant "where in the *XQuery* spec". I'd like to be able to point to a section in the XQuery specification that says that processors must respect this. There are a number of schema- and type- related features that are labeled "optional" here: https://www.w3.org/TR/xquery-31/#id-conform-optional-features. And eXist's documentation says which optional features it doesn't support: http://exist-db.org/exist/apps/doc/xquery.xml#unsupported-features. 

Basically, I'm wondering whether this is a bug (failure to implement a required feature) or just an omission of an optional feature. 

To know how to classify this, we'll need to find the place in the XQuery specification that addresses this.

ahenket:
[...]

PR #6222, opened 2026-04-08
[bugfix] declare copy-namespaces no-inherit breaking element constructors

joewiz:
## Summary

`declare copy-namespaces preserve, no-inherit` was incorrectly applying the `no-inherit` flag during element constructor evaluation, causing two distinct bugs:

1. **Name resolution failure**: `getURIForPrefix()` skips `inheritedInScopeNamespaces` when `inheritNamespaces=false`, so a child constructor `<b/>` inside `<e xmlns="http://example.com/">` could not see the inherited default namespace, and a prefixed name like `<ns:c/>` whose prefix was declared on an enclosing element threw `XPTY0004`.

2. **`in-scope-prefixes()` returning incomplete results**: Nested direct constructors like `<e3>{<e2>{<e1/>}</e2>}</e3>` — calling `in-scope-prefixes(<e1>)` only returned `<e1>`'s own prefix, missing `namespace2` and `namespace3` declared by the enclosing constructors.

Per XQuery 3.1 §3.9.3.4, `no-inherit` governs how namespaces propagate from *copied source nodes* (existing XDM nodes placed into constructors via `{$var}`) — it must not affect namespace resolution for element constructors themselves.

Also fixes the real-world regression in eXist-db/exist#2182 where `declare copy-namespaces preserve, no-inherit` caused XPath path steps over constructed elements to silently return empty sequences.

Companion to PR #6219 (serializer `xmlns=""` fix).

## What Changed

**`ElementConstructor.java` — Fix A (name resolution)**

Temporarily restores `inheritNamespaces=true` while resolving the element's QName via `QName.parse()`, so that prefix-to-URI lookups consult the inherited namespace context. Restored in a `finally` block.

**`FunInScopePrefixes.java` — Fix B (ancestor traversal)**

Always traverse the in-memory node's ancestor chain when collecting namespace prefixes, removing the coarse `inheritNamespaces()` switch that previously blocked traversal for all `no-inherit` queries. Updated cleanup to remove all entries with an empty URI (namespace undeclarations).

**`EnclosedExpr.java` — Fix C (no-inherit copy semantics)**

When `no-inherit` is active and an enclosed expression `{...}` places a *pre-existing* element node (a variable reference) into the outer element, a new `NoInheritCopyReceiver` intercepts the copy event stream and injects namespace undeclarations (`xmlns:prefix=""`) onto the root of the copy for every ancestor namespace binding not already declared by that root. This neutralizes ancestor traversal for those nodes so `in-scope-prefixes()` returns only the node's own context.

Pre-existing nodes are distinguished from direct constructors by capturing the `MemTreeBuilder` allocated during the enclosed expression's evaluation (via new `peekDocumentBuilder()`) and checking whether each result node belongs to that builder's document.

**`XQueryContext.java` / `ModuleContext.java`**: added `peekDocumentBuilder()` to inspect the current builder without disturbing the context stack.

## Spec References

- [XQuery 3.1 §3.9.3.4 — Copy Namespaces Declaration](https://www.w3.org/TR/xquery-31/#id-copy-namespaces-declaration)

## XQTS Before/After (prod-CopyNamespacesDecl, XQ 3.1)

| Test | Before | After |
|------|--------|-------|
| K2-CopyNamespacesProlog-4 | FAIL | **PASS** |
| K2-CopyNamespacesProlog-5 | FAIL | **PASS** |
| K2-CopyNamespacesProlog-9 (direct constructor half) | FAIL | **PASS** |
| copynamespace-2 | FAIL | **PASS** |
| K2-CopyNamespacesProlog-9 (variable copy half) | PASS | PASS ✓ |
| copynamespace-3 (no-inherit, variable copy) | PASS | PASS ✓ |
| copynamespace-5 (no-inherit, variable copy + FLWOR sort) | PASS | PASS ✓ |

35/35 passing in the prod-CopyNamespacesDecl test set.

## CI Changes (second commit)

This PR also cherry-picks CI improvements from PR #6186 and extends them:

- **`exist-parent/pom.xml`**: Added `forkedProcessTimeoutInSeconds=600` to surefire and failsafe plugins, preventing hung test JVMs from blocking CI indefinitely (e.g. `DeadlockIT`, `MoveResourceTest`).
- **`.github/workflows/ci-test.yml`**: Added `--offline` to the `license:check` step (eliminates 7+ min timeouts from unreachable repos; the develop cache has all needed artifacts). Added wagon HTTP timeout flags via step-level `MAVEN_OPTS` to prevent Maven Build hangs caused by stalled connections to `repo.exist-db.org` and `repo.evolvedbinary.com`.

## Test Plan

- [x] XQTS `prod-CopyNamespacesDecl`: 35/35 pass (4 were failing before, none regressed)
- [x] JUnit full suite: 6,542 tests, 0 failures, 0 errors
- [ ] Manual: `declare copy-namespaces preserve, no-inherit; <e xmlns="http://example.com/">{ <b/> }</e>` → `<b>` in `http://example.com/` (implied by K2-CopyNamespacesProlog-4 ✓)
- [ ] Manual: eXist-db/exist#2182 (@daliboris) — `$items/t:item/cc:collection` returns result instead of empty sequence (implied by copynamespace-2 + Fix A ✓)

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ElementConstructor.java
+++ b/exist-core/src/main/java/org/exist/xquery/ElementConstructor.java
@@ -294,12 +294,6 @@ public class ElementConstructor extends NodeConstructor {
                         throw e;
                     }
 
-                    //Use the default namespace if specified
-                    /*
-                     if (qn.getPrefix() == null && context.inScopeNamespaces.get("xmlns") != null) {
-                         qn.setNamespaceURI((String)context.inScopeNamespaces.get("xmlns"));
-                     }
-                     */
                     if (qn.getPrefix() == null && context.getInScopeNamespace(XMLConstants.DEFAULT_NS_PREFIX) != null) {
                         qn = new QName(qn.getLocalPart(), context.getInScopeNamespace(XMLConstants.DEFAULT_NS_PREFIX), qn.getPrefix());
                     }
--- a/exist-core/src/main/java/org/exist/xquery/ElementConstructor.java
+++ b/exist-core/src/main/java/org/exist/xquery/ElementConstructor.java
@@ -275,24 +275,38 @@ public class ElementConstructor extends NodeConstructor {
             if (qnitem instanceof QNameValue) {
                 qn = ((QNameValue) qnitem).getQName();
             } else {
-                //Do we have the same result than Atomize there ? -pb
-                try {
-                    qn = QName.parse(context, qnitem.getStringValue());
-                } catch (final QName.IllegalQNameException e) {
-                    throw new XPathException(this, ErrorCodes.XPTY0004, "'" + qnitem.getStringValue() + "' is not a valid element name");
-                } catch (final XPathException e) {
-                    e.setLocation(getLine(), getColumn(), getSource());
-                    throw e;
+                // Element constructors must resolve namespace prefixes using the full
+                // inherited namespace context, regardless of declare copy-namespaces no-inherit.
+                // The no-inherit option governs how namespaces propagate from copied source
+                // nodes, not how constructor names are resolved (XQuery 3.1 §3.9.3.4).
+                final boolean savedInherit = context.inheritNamespaces();
+                if (!savedInherit) {
+                    context.setInheritNamespaces(true);
                 }
+                try {
+                    //Do we have the same result than Atomize there ? -pb
+                    try {
+                        qn = QName.parse(context, qnitem.getStringValue());
+                    } catch (final QName.IllegalQNameException e) {
+                        throw new XPathException(this, ErrorCodes.XPTY0004, "'" + qnitem.getStringValue() + "' is not a valid element name");
+                    } catch (final XPathException e) {
+                        e.setLocation(getLine(), getColumn(), getSource());
+                        throw e;
+                    }
 
-                //Use the default namespace if specified
-                /*
-                 if (qn.getPrefix() == null && context.inScopeNamespaces.get("xmlns") != null) {
-                     qn.setNamespaceURI((String)context.inScopeNamespaces.get("xmlns"));
-                 }
-                 */
-                if (qn.getPrefix() == null && context.getInScopeNamespace(XMLConstants.DEFAULT_NS_PREFIX) != null) {
-                    qn = new QName(qn.getLocalPart(), context.getInScopeNamespace(XMLConstants.DEFAULT_NS_PREFIX), qn.getPrefix());
+                    //Use the default namespace if specified
+                    /*
+                     if (qn.getPrefix() == null && context.inScopeNamespaces.get("xmlns") != null) {
+                         qn.setNamespaceURI((String)context.inScopeNamespaces.get("xmlns"));
+                     }
+                     */
+                    if (qn.getPrefix() == null && context.getInScopeNamespace(XMLConstants.DEFAULT_NS_PREFIX) != null) {
+                        qn = new QName(qn.getLocalPart(), context.getInScopeNamespace(XMLConstants.DEFAULT_NS_PREFIX), qn.getPrefix());
+                    }
+                } finally {
+                    if (!savedInherit) {
+                        context.setInheritNamespaces(false);
+                    }
                 }
             }
 
--- a/exist-core/src/main/java/org/exist/xquery/EnclosedExpr.java
+++ b/exist-core/src/main/java/org/exist/xquery/EnclosedExpr.java
@@ -84,15 +84,22 @@ public class EnclosedExpr extends PathExpr {
         Sequence result;
         context.enterEnclosedExpr();
         try {
+            // Check copy-namespaces mode before evaluation so we can lazily capture
+            // innerBuilder only when no-inherit is active. In the default (inherit) case
+            // this avoids the peekDocumentBuilder() call entirely.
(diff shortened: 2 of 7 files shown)


exist-core/src/main/java/org/exist/xquery/ElementConstructor.java
```diff
@@ -275,24 +275,38 @@ public Sequence eval(final Sequence contextSequence, final Item contextItem) thr
...
-                 }
-                 */
-                if (qn.getPrefix() == null && context.getInScopeNamespace(XMLConstants.DEFAULT_NS_PREFIX) != null) {
-                    qn = new QName(qn.getLocalPart(), context.getInScopeNamespace(XMLConstants.DEFAULT_NS_PREFIX), qn.getPrefix());
+                    //Use the default namespace if specified
+                    /*
+                     if (qn.getPrefix() == null && context.inScopeNamespaces.get("xmlns") != null) {
+                         qn.setNamespaceURI((String)context.inScopeNamespaces.get("xmlns"));
+                     }
+                     */
```
duncdrum: dead code?

duncdrum (commented):
more namespace fun, one comment in file.

exist-core/src/main/java/org/exist/xquery/ElementConstructor.java
line-o: Looks like it, I would be in favour of removing this.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -969,6 +969,18 @@ public String getInheritedNamespace(final String prefix) {
...
 
+    public Map<String, String> getAllInheritedNamespaces() {
+        return inheritedInScopeNamespaces;
+    }
+
+    public Map<String, String> getInScopeNamespaces() {
+        return inScopeNamespaces;
+    }
+
+    public MemTreeBuilder peekDocumentBuilder() {
```
line-o: Why is this method named _peek_DocumentBuilder?
This looks like a standard getter to me

line-o (commented):
Do these extra checks have a significant impact on element construction performance even if copy-namespaces no-inherit is not set?

exist-core/src/main/java/org/exist/xquery/EnclosedExpr.java
```diff
@@ -130,7 +161,25 @@ public Sequence eval(Sequence contextSequence, Item contextItem) throws XPathExc
...
-                            next.copyTo(context.getBroker(), receiver);
+                            // When copy-namespaces no-inherit is active, pre-existing element nodes
+                            // (i.e. nodes from variable references, not direct constructors) must have
+                            // namespace undeclarations injected so that ancestor-traversal in
+                            // in-scope-prefixes() is neutralized for inherited namespace bindings.
+                            if (noInherit && ancestorNS != null
+                                    && next.getType() == Type.ELEMENT
+                                    && next instanceof NodeImpl) {
+                                final NodeImpl nodeImpl = (NodeImpl) next;
+                                final boolean isPreExisting = (innerBuilder == null)
```
line-o, line 178: I would prefer to populate innerBuilder lazily on first access instead of always regardless if it will ever be used.
Then it also becomes clear that only its document is needed to check for preExisting nodes.

joewiz:

Good catches, addressed in 2d9e248c0a:

- Removed the stale commented-out block in `ElementConstructor` (dead since the in-scope namespace API was modernized)
- Renamed `peekDocumentBuilder()` → `getCurrentDocumentBuilder()` in `XQueryContext`, `ModuleContext`, and the single call site in `EnclosedExpr`
- Made `innerBuilder` capture lazy: moved `noInherit` computation before the try block so `getCurrentDocumentBuilder()` is only called when `copy-namespaces no-inherit` is active — the hot path (default inherit mode) now has zero overhead from this method

On the performance question: with the lazy fix, the overhead in the default (inherit) case is exactly one `context.inheritNamespaces()` call per `EnclosedExpr` evaluation — just a field read on `XQueryContext`. The per-item check at line 168 (`noInherit && ancestorNS != null`) short-circuits immediately on the first boolean. I don't think this warrants a microbenchmark, but happy to add one if you'd like to see the numbers.

line-o (approved):
LGTM!

I would have preferred an even more lazy approach were the document of the innerBuilder would have been initialized on first use.
Because even if `noinherit` is true it is not 100% sure we need it.

joewiz:

Thanks for the review and approval, @line-o! Noted on the lazier initialization — the current approach already makes `innerBuilder` capture conditional on `noInherit`, so the overhead in the default (inherit) case is just a boolean check. Making the document initialization even lazier (on first use within the loop) is possible but would add complexity to the per-item code path. Happy to revisit if profiling shows it matters.

duncdrum (changes requested):
thanks @joewiz changes look good to me, given all the samples in #2182 could you add a XQSuite test to guard against regressions. It's easier to track, than XQTS failures.

joewiz:

@duncdrum Good idea — will add an XQSuite test covering the key scenarios from #2182 (inherit vs no-inherit with nested element constructors, namespace undeclaration). XQSuite tests are easier to maintain and give quicker feedback than XQTS-only coverage.

duncdrum (approved):

PR #6235, opened 2026-04-13
[bugfix] Make zip test with illegal filenames pass consistently.

dizzzz:
Modified unzip-tests.xql to relax the error expectation for uz: fnUzipCp437ContentWrongEncoding: Updated the %test: assertError regex to include invalid LOC header, which is the standard error message thrown by java.util.zip.ZipInputStream (specifically ZipException) in Java 17+ and 21 when encountering malformed encoding in ZIP

Somehow this issue irregularly appeared for local builds java21/25 only.


duncdrum (approved):

line-o (approved):
I believe I have seen that too.

PR #6226, opened 2026-04-10
[refactor] java 21 syntax

duncdrum:
extends #6146 

minor version bump to lucene and matching ICU release.

use a number of open rewrite recipes for syntax updates, mainly 

```
mod run . --recipe org.openrewrite.java.migrate.UpgradeToJava21
```

These are applied in the whole codebase, not just within files directly affected by the lucene update. I planned this as a separate PR, to facilitate review.

PR #6239, opened 2026-04-14
Syntax updates

duncdrum:
Large Changeset, supersedes #6226 

use a number of open rewrite recipes for syntax updates, mainly 

```
mod run . --recipe org.openrewrite.java.migrate.UpgradeToJava21
```

These are applied in the whole codebase, not just within files directly affected by the lucene update. I planned this as a separate PR, to facilitate review.


duncdrum:
P.S.: the larger number of  codacy issues, is to be expected. This PR liberally modifies legacy files, so all old issues resurface.

dizzzz (approved):
LGTM

exist-core/src/main/java/org/exist/client/security/AccessControlEntryDialog.java
```diff
@@ -298,16 +300,12 @@ private void btnCloseActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRS
...
-            case GROUP:
-                cmbUsername.setEnabled(false);
-                cmbGroupName.setEnabled(true);
-                break;
+        if (aceTarget == ACLPermission.ACE_TARGET.USER) {
+            cmbGroupName.setEnabled(false);
+            cmbUsername.setEnabled(true);
+        } else if (aceTarget == ACLPermission.ACE_TARGET.GROUP) {
+            cmbUsername.setEnabled(false);
+            cmbGroupName.setEnabled(true);
```
reinhapa, line 308: All fine and good as long we have only two options. Using this way, we will silently miss a potential implementation space if an enum value is added later...

exist-core/src/main/java/org/exist/client/security/AutoCompletion.java
```diff
@@ -76,18 +76,12 @@ public void keyPressed(final KeyEvent e) {
...
-                    case KeyEvent.VK_DELETE:
-                        e.consume();
-                        comboBox.getToolkit().beep();
-                        break;
+                if (e.getKeyCode() == KeyEvent.VK_BACK_SPACE) {
+                    hitBackspace = true;
+                    hitBackspaceOnSelection = editor.getSelectionStart() != editor.getSelectionEnd();
+                } else if (e.getKeyCode() == KeyEvent.VK_DELETE) {
+                    e.consume();
+                    comboBox.getToolkit().beep();
```
reinhapa, line 84: I know the the default "behavior" for only two options is to go with `if-else-if` constructs but for my taste the new switch expression seem easier to read...

exist-core/src/main/java/org/exist/client/security/UmaskSpinnerModel.java
reinhapa, line 1: the change from octal to decimal makes the code less readable in this special case of umask

exist-core/src/main/java/org/exist/client/security/UserManagerDialog.java
```diff
@@ -437,8 +440,8 @@ private void tblUsersMouseClicked(java.awt.event.MouseEvent evt) {//GEN-FIRST:ev
         final boolean userSelected = tblUsers.getSelectedRow() > -1;
         final String selectedUsername = getSelectedUsername();
         
-        boolean canModify = userSelected && !selectedUsername.equals(SecurityManager.SYSTEM);
-        boolean canDelete = userSelected && !(selectedUsername.equals(SecurityManager.SYSTEM) || selectedUsername.equals(SecurityManager.DBA_USER) || selectedUsername.equals(SecurityManager.GUEST_USER));
+        boolean canModify = userSelected && !SecurityManager.SYSTEM.equals(selectedUsername);
+        boolean canDelete = userSelected && !(SecurityManager.SYSTEM.equals(selectedUsername) || SecurityManager.DBA_USER.equals(selectedUsername) || SecurityManager.GUEST_USER.equals(selectedUsername));
```
reinhapa, line 444: Maybe using static imports would make it even more readable

exist-core/src/main/java/org/exist/client/security/UserManagerDialog.java
```diff
@@ -480,7 +483,7 @@ private void tblGroupsMouseClicked(java.awt.event.MouseEvent evt) {//GEN-FIRST:e
         final boolean groupSelected = tblGroups.getSelectedRow() > -1;
         final String selectedGroup = getSelectedGroup();
         
-        boolean canDelete = groupSelected && !(selectedGroup.equals(SecurityManager.DBA_GROUP) || selectedGroup.equals(SecurityManager.GUEST_GROUP));
+        boolean canDelete = groupSelected && !(SecurityManager.DBA_GROUP.equals(selectedGroup) || SecurityManager.GUEST_GROUP.equals(selectedGroup));
```
reinhapa, line 486: Maybe using static imports would make it even more readable

exist-core/src/main/java/org/exist/client/tristatecheckbox/TristateCheckBox.java
reinhapa, line 63:
```suggestion
switch (state) {
case SELECTED -> setSelected(true);
case INDETERMINATE, DESELECTED -> putClientProperty("SelectionState", state);
default -> throw new IllegalArgumentException();
}
```
Why not use new switch as expression?

exist-core/src/main/java/org/exist/client/ClientFrame.java
```diff
@@ -659,11 +659,11 @@ private synchronized void type(final KeyEvent e) {
...
-        if (cmd.equals(CUT)) {
+        if (CUT.equals(cmd)) {
             shell.cut();
-        } else if (cmd.equals(COPY)) {
+        } else if (COPY.equals(cmd)) {
             shell.copy();
-        } else if (cmd.equals(PASTE)) {
+        } else if (PASTE.equals(cmd)) {
             shell.paste();
         }
```
reinhapa, line 668: Using switch-as-expression seems to be better here...

exist-core/src/main/java/org/exist/collections/triggers/XQueryStartupTrigger.java
```diff
@@ -164,9 +164,9 @@ private boolean isPermissionsOK(Collection collection) {
 
         Permission perms = collection.getPermissions();
 
-        return (perms.getOwner().getName().equals(SecurityManager.SYSTEM)
-                && perms.getGroup().getName().equals(SecurityManager.DBA_GROUP)
-                && perms.getMode() == Permission.DEFAULT_SYSTEM_SECURITY_COLLECTION_PERM);
+        return SecurityManager.SYSTEM.equals(perms.getOwner().getName())
+                && SecurityManager.DBA_GROUP.equals(perms.getGroup().getName())
```
reinhapa, line 168: Maybe use static imports here?

exist-core/src/main/java/org/exist/collections/IndexInfo.java
```diff
@@ -31,11 +31,7 @@
 import org.exist.storage.txn.Txn;
 import org.exist.util.serializer.DOMStreamer;
 import org.exist.xmldb.XmldbURI;
-import org.xml.sax.ContentHandler;
-import org.xml.sax.EntityResolver;
-import org.xml.sax.InputSource;
-import org.xml.sax.SAXException;
-import org.xml.sax.XMLReader;
+import org.xml.sax.*;
```
reinhapa, line 34: Do we now use single or star imports?

exist-core/src/main/java/org/exist/collections/MutableCollection.java
```diff
@@ -705,9 +703,6 @@ public LockedDocument getDocumentWithLock(final DBBroker broker, final XmldbURI
 
             final DocumentImpl doc = documents.get(name.lastSegmentString());
 
-            // NOTE: early release of Collection lock inline with Asymmetrical Locking scheme
-            collectionLock.close();
-
```
reinhapa, line 710: hmm, this seems to be removed prematurely

exist-core/src/main/java/org/exist/dom/memtree/DOMIndexer.java
reinhapa, line 1: See `Unnecessary use of fully qualified name` Codacy checks introduced with the change...

exist-core/src/main/java/org/exist/http/servlets/RedirectorServlet.java
```diff
@@ -107,18 +107,19 @@
...
 
-    public final static String DEFAULT_USER = "guest";
-    public final static String DEFAULT_PASS = "guest";
-    public final static XmldbURI DEFAULT_URI = XmldbURI.EMBEDDED_SERVER_URI.append(XmldbURI.ROOT_COLLECTION_URI);
+    public static final String DEFAULT_USER = "guest";
+    public static final String DEFAULT_PASS = "guest";
+    public static final XmldbURI DEFAULT_URI = XmldbURI.EMBEDDED_SERVER_URI.append(XmldbURI.ROOT_COLLECTION_URI);
 
-    private String user = null;
-    private String password = null;
```
reinhapa, line 119: Maybe use move the fields `user` and `password` to local variables as proposed by the Codacy check.

exist-core/src/main/java/org/exist/launcher/SplashScreen.java
```diff
@@ -21,30 +21,31 @@
...
 public class SplashScreen extends JFrame implements Observer, Comparable {
 
+    @Serial
     private static final long serialVersionUID = -8449133653386075548L;
 
-    private JLabel statusLabel;
-    private JLabel versionLabel;
-    private Launcher launcher;
+    private final JLabel statusLabel;
+    private final JLabel versionLabel;
```
reinhapa, line 47: Maybe convert the field `versionLabel` to variable as suggested by the Codacy

exist-core/src/main/java/org/exist/management/client/JMXClient.java
```diff
@@ -30,15 +30,7 @@
...
-import javax.management.Attribute;
-import javax.management.AttributeList;
-import javax.management.AttributeNotFoundException;
-import javax.management.InstanceNotFoundException;
-import javax.management.MBeanException;
-import javax.management.MBeanServerConnection;
-import javax.management.MalformedObjectNameException;
-import javax.management.ObjectName;
-import javax.management.ReflectionException;
+import javax.management.*;
```
reinhapa, line 33: maybe revert start import...

exist-core/src/main/java/org/exist/management/impl/JMXAgent.java
```diff
@@ -28,14 +28,7 @@
...
 
-import javax.management.InstanceAlreadyExistsException;
-import javax.management.InstanceNotFoundException;
-import javax.management.MBeanRegistrationException;
-import javax.management.MBeanServer;
-import javax.management.MBeanServerFactory;
-import javax.management.MalformedObjectNameException;
-import javax.management.NotCompliantMBeanException;
-import javax.management.ObjectName;
+import javax.management.*;
```
reinhapa, line 31: senseless  change...

exist-core/src/main/java/org/exist/security/internal/GroupImpl.java
```diff
@@ -30,6 +29,8 @@
 import org.exist.security.Account;
 import org.exist.storage.DBBroker;
 
+import java.util.List;
+
```
reinhapa: No actual change

exist-core/src/main/java/org/exist/security/Permission.java
reinhapa, line 1: Octal value definitions make more sense in the permission context..

exist-core/src/main/java/org/exist/security/Session.java
```diff
@@ -82,6 +82,6 @@ public Subject getSubject() {
 	}
 
 	public boolean isValid() {
-		return (System.currentTimeMillis() - lastUse <= 30*1000); //30 seconds
+		return System.currentTimeMillis() - lastUse <= 30*1000; //30 seconds
```
reinhapa: Could use underscore instead do calculation on init/compile
```suggestion
		return System.currentTimeMillis() - lastUse <= 30_000; //30 seconds
```

exist-core/src/main/java/org/exist/security/UnixStylePermission.java
```diff
@@ -433,13 +433,7 @@ public boolean validate(final Subject user, final int mode) {
...
             }
         }
-
-        //check other
-        if((mode & (vector & 7)) == mode) {
-            return true;
-        }
-
-        return false;
+        return (mode & (vector & 7)) == mode;
```
reinhapa, line 437: may be leave commit in...
```suggestion
        //check other
        return (mode & (vector & 7)) == mode;
```

exist-core/src/main/java/org/exist/storage/dom/RemoveEmptyPageLoggable.java
```diff
@@ -21,14 +21,14 @@
...
-import java.nio.ByteBuffer;
-
 import org.exist.storage.DBBroker;
 import org.exist.storage.NativeBroker;
 import org.exist.storage.journal.AbstractLoggable;
 import org.exist.storage.journal.LogException;
 import org.exist.storage.txn.Txn;
 
+import java.nio.ByteBuffer;
+
```
reinhapa: no functional change here

exist-core/src/main/java/org/exist/util/serializer/MicroXmlWriter.java
```diff
@@ -55,14 +55,9 @@ private CharSequence removeRestrictedChars(final CharSequence charSeq) {
...
-
-                case '>':
-                    //ignore char
-                    break;
-
-                default:
-                    buf.append(charSeq.charAt(i));
+            if (charSeq.charAt(i) == '>') {
+            } else {
+                buf.append(charSeq.charAt(i));
```
reinhapa: Maybe even simpler by doing this:
```suggestion
            if (charSeq.charAt(i) != '>') {
                buf.append(charSeq.charAt(i));
```

exist-core/src/test/java/org/exist/collections/CollectionRemovalTest.java
```diff
@@ -184,8 +183,8 @@ public void initDB() throws EXistException, PermissionDeniedException, IOExcepti
         try(final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()));
             final Txn transaction = transact.beginTransaction()) {
 
-            final int worldReadable = 0744;
-            final int worldForbidden = 0700;
+            final int worldReadable = 484;
+            final int worldForbidden = 448;
```
reinhapa, line 187: Using octal values makes more sense here

exist-core/src/test/java/org/exist/security/internal/aider/UnixStylePermissionAiderTest.java
reinhapa, line 1: using octal values is more readable as decimal values in this case...

exist-core/src/test/java/org/exist/security/SimpleACLPermissionTest.java
reinhapa, line 1: using octal values is more readable as decimal values in this case...

exist-core/src/test/java/org/exist/security/UnixStylePermissionTest.java
reinhapa, line 1: using octal values is more readable as decimal values in this case...

exist-core/src/test/java/org/exist/security/XMLDBSecurityTest.java
reinhapa, line 1: using octal values is more readable as decimal values in this case...

exist-core/src/test/java/org/exist/storage/CopyCollectionTest.java
reinhapa, line 1: using octal values is more readable as decimal values in this case...

exist-core/src/test/java/org/exist/storage/CopyResourceTest.java
reinhapa, line 1: using octal values is more readable as decimal values in this case...

exist-core/src/test/java/org/exist/storage/StartupLockingTest.java
reinhapa, line 1: Needs reformatting around the enhanced `while` loops where `continue;` was added

exist-core/src/test/java/org/exist/storage/NativeBrokerLockingTest.java
reinhapa, line 1: Needs reformatting around the enhanced `while` loops where `continue;` was added

exist-core/src/test/java/org/exist/storage/StoreResourceTest.java
reinhapa, line 1: using octal values is more readable as decimal values in this case...

exist-core/src/test/java/org/exist/xquery/functions/securitymanager/AccountManagementFunctionRemoveAccountTest.java
```diff
@@ -81,16 +81,15 @@ private Sequence xqueryRemoveAccount(final String username, final Optional<Subje
...
         }
     }
 
     private static void extractPermissionDenied(final Runnable3E<XPathException, PermissionDeniedException, EXistException> runnable) throws XPathException, PermissionDeniedException, EXistException {
         try {
             runnable.run();
         } catch (final XPathException e) {
-            if (e.getCause() != null && e.getCause() instanceof PermissionDeniedException) {
+            if (e.getCause() instanceof PermissionDeniedException) {
                 throw (PermissionDeniedException)e.getCause();
```
reinhapa:
```suggestion
            if (e.getCause() instanceof PermissionDeniedException permissionDeniedException) {
throw permissionDeniedException.getCause();
```

exist-core/src/test/java/org/exist/xquery/functions/securitymanager/GroupManagementFunctionRemoveGroupTest.java
```diff
@@ -307,16 +307,15 @@ private Sequence xqueryRemoveGroup(final String groupname) throws EXistException
...
         }
     }
 
     private static void extractPermissionDenied(final Runnable3E<XPathException, PermissionDeniedException, EXistException> runnable) throws XPathException, PermissionDeniedException, EXistException {
         try {
             runnable.run();
         } catch (final XPathException e) {
-            if (e.getCause() != null && e.getCause() instanceof PermissionDeniedException) {
+            if (e.getCause() instanceof PermissionDeniedException) {
                 throw (PermissionDeniedException)e.getCause();
```
reinhapa:
```suggestion
            if (e.getCause() instanceof PermissionDeniedException permissionDeniedException) {
throw permissionDeniedException.getCause();
```

exist-core/src/test/java/org/exist/xquery/functions/securitymanager/GroupMembershipFunctionRemoveGroupMemberTest.java
```diff
@@ -174,7 +172,7 @@ private static void extractPermissionDenied(final Runnable3E<XPathException, Per
         try {
             runnable.run();
         } catch (final XPathException e) {
-            if (e.getCause() != null && e.getCause() instanceof PermissionDeniedException) {
+            if (e.getCause() instanceof PermissionDeniedException) {
                 throw (PermissionDeniedException)e.getCause();
```
reinhapa:
```suggestion
            if (e.getCause() instanceof PermissionDeniedException permissionDeniedException) {
throw permissionDeniedException.getCause();
```

exist-core/src/test/java/org/exist/xquery/functions/securitymanager/PermissionsFunctionChmodTest.java
```diff
@@ -344,7 +343,7 @@ private static void extractPermissionDenied(final Runnable4E<AuthenticationExcep
         try {
             runnable.run();
         } catch (final XPathException e) {
-            if (e.getCause() != null && e.getCause() instanceof PermissionDeniedException) {
+            if (e.getCause() instanceof PermissionDeniedException) {
                 throw (PermissionDeniedException)e.getCause();
```
reinhapa:
```suggestion
            if (e.getCause() instanceof PermissionDeniedException permissionDeniedException) {
throw permissionDeniedException.getCause();
```

exist-core/src/test/java/org/exist/xquery/functions/securitymanager/PermissionsFunctionChownTest.java
```diff
@@ -1719,7 +1718,7 @@ private static void extractPermissionDenied(final Runnable4E<AuthenticationExcep
         try {
             runnable.run();
         } catch (final XPathException e) {
-            if (e.getCause() != null && e.getCause() instanceof PermissionDeniedException) {
+            if (e.getCause() instanceof PermissionDeniedException) {
                 throw (PermissionDeniedException)e.getCause();
```
reinhapa:
```suggestion
            if (e.getCause() instanceof PermissionDeniedException permissionDeniedException) {
throw permissionDeniedException.getCause();
```

extensions/exquery/restxq/src/test/java/org/exist/extensions/exquery/restxq/impl/adapters/DomEnhancingNodeProxyAdapterTest.java
```diff
@@ -81,9 +81,6 @@ public static void setup() throws EXistException, PermissionDeniedException, IOE
 
             collection.storeDocument(transaction, broker, TEST_DOC_URI, new StringInputSource(TEST_DOC), MimeType.XML_TYPE);
 
-            // async release of collection lock
-            collection.close();
-
```
reinhapa, line 86: Maybe this was removed unintentionally?

reinhapa (changes requested):
There are a couple of points that need some rework:

- Still a lot of static String constants that could be converted to text blocks in order to be better readable. 
- We need to define how he handle imports in general, as there are some places, where single imports were replaced with start imports and vice-versa.
- There are static code validation errors, that can be solved by moving the constructor  in relation of the field definition or using local variables instead of fields...

exist-core/src/main/java/org/exist/client/security/AutoCompletion.java
duncdrum: s.b. we could tweak that setting, but it strikes me as a personal preference. Not sure what others think. One problem with j21 minimal switches (which I like) is that the codacy cli (on Mac) cannot parse them properly yet, so you loose analysis locally, and have to wait for CI to tell you about errors in the file.

duncdrum:
@reinhapa thank you for looking into this. The idea was to get a sense of the capabilities of open rewrite and to see existing recipes in action against our code base. 

The points you raise are valid. However we should clarify what our expectations are. From the time I spent with this tool I can say that the collaborative and iterative tooling for refactorings works well in principle, and is something I want to further explore and make use of (e.g. for migrating apps from exist v6 to v7). 

When it comes to the output of any specific recipe run, there are a number of questions. (each commit above is a different recipe sometimes run repeatedly). The point is less that it promises to catch-em-all in a single run, but to quickly perform safe ones. It's not yet clear to me when the tool determines a textblock refactor to be safe. It clearly skips some. Generally, I don't mind that some are missed, as long as the ones which are found are actually implemented  properly. 

If the java21 recipe  does not meet our expectations we have in principle three options.  
- we can try to figure out why and send a PR to update the recipe. 
- We can create private or public recipes that do what we want, 
- in some cases we can tweak the parameters when invoking a recipe.

The question for me really is about, is this recipe worth applying or not? To me it looks to do more good than bad overall. 

as for a policy for imports, agreed. Let's discuss this in a community call, and depending on outcome we can see if we want a refractor at all, and the direction. I had to rebase this multiple times it's also possible that that introduces some regressions with star imports. The `common static` recipe was the least convincing and most error prone. 

From a practical point of view rewrite recipes compete with other modes of enforcing code style. It's certainly not the only or best option for everything. But I see the value of having reproducible shareable and composable refactor recipes doing the heavy lifting. I'm more curious to see if I'm the only one feeling that way, or if this is something we want to explore as a group.

I m happy to drop commits from this PR, to make it a smaller change set. The gitignore and formatting commits however are important to actually use the tool without error. The pom changes seems also good. Would that be a way forward, any of the others you think are safe to pull in?

exist-core/src/main/java/org/exist/client/security/UmaskSpinnerModel.java
duncdrum: agreed, this is something we should try to filter from the recipe

exist-core/src/main/java/org/exist/client/security/AutoCompletion.java
line-o: If Codacy on desktop can parse them or not is much less relevant than the points @reinhapa raised.
I am in favour of _switching_ to the construct that will give us static checks if all options were handled wherever possible.

duncdrum: my comment was about switch statements with less then two switches. @line-o happy to review your PR.

duncdrum:
in case anybody is curious, fresh run of java21 recipe against `develop`: 

- [org.openrewrite.table.RecipeRunStats.csv](https://github.com/user-attachments/files/26781130/org.openrewrite.table.RecipeRunStats.csv)
- [org.openrewrite.table.SourcesFileResults.csv](https://github.com/user-attachments/files/26781140/org.openrewrite.table.SourcesFileResults.csv)

exist-core/src/main/java/org/exist/client/security/AutoCompletion.java
line-o: Less than two would be one or zero. I think when switching on enums the _current_ number of options is less relevant.

exist-core/src/main/java/org/exist/client/security/AccessControlEntryDialog.java
duncdrum: I don't see it. The poorly formed switch without a default doesn't anticipate anything other than `group` or `user` . The if-else-if is functionally identical, removes linter noise, and is shorter. If we ever get trustees other than group or users, this code section will have to be modified one way or the other.

exist-core/src/main/java/org/exist/client/security/AutoCompletion.java
duncdrum: @line-o you're missing the point. I'm evaluating recipes. If the output of a recipes, will lead into an infinite cycle because our manual or automated review undoes the output that is exactly the relevant information I'm looking for. 

I seem to remember you arguing for switch statements with defaults in the past. But thanks for the hints about counting.

exist-core/src/main/java/org/exist/client/security/UserManagerDialog.java
duncdrum: It would,  but that applies to both before and after. So essentially something for a future refactor.

exist-core/src/main/java/org/exist/client/security/UserManagerDialog.java
duncdrum: s.a

exist-core/src/main/java/org/exist/client/tristatecheckbox/TristateCheckBox.java
duncdrum: certainly nicer.  Would make a good extension to the j21 recipe .

exist-core/src/main/java/org/exist/client/ClientFrame.java
duncdrum: `CUT`, `COPY`, `PASTE` are initialised as methods, so `case CUT -> …` is not allowed. The recipe was correct in not rewriting it, this is cleaner.

exist-core/src/main/java/org/exist/collections/IndexInfo.java
duncdrum: I don't know but it would be good to come to an agreement. The second recipe seems to follow a different pattern than the others. Given all the noise I had to undo before committing, I m leaning on taking it out alltogether.

exist-core/src/main/java/org/exist/collections/MutableCollection.java
duncdrum: agreed seems fishy

exist-core/src/main/java/org/exist/management/client/JMXClient.java
duncdrum: s.a. the static error recipe is not for us, I m going to drop the whole commit.

exist-core/src/main/java/org/exist/dom/memtree/DOMIndexer.java
duncdrum: dropping

exist-core/src/main/java/org/exist/launcher/SplashScreen.java
duncdrum: good catch, but not something I would have expected the recipe to do

exist-core/src/main/java/org/exist/management/impl/JMXAgent.java
duncdrum: yup

exist-core/src/main/java/org/exist/security/Permission.java
duncdrum: yup

exist-core/src/main/java/org/exist/security/UnixStylePermission.java
duncdrum: battle for another day, sticking with the original

exist-core/src/test/java/org/exist/storage/StartupLockingTest.java
duncdrum: good catch, weird that the formatting got mangled in the first place 🤔

extensions/exquery/restxq/src/test/java/org/exist/extensions/exquery/restxq/impl/adapters/DomEnhancingNodeProxyAdapterTest.java
duncdrum: yup static analysis

duncdrum:
ok  I want through all the comments and suggestions. The main takeaway for me is that the java21 changes are fundamentally sound. Which is good, as they are part of the upcoming java25 changes we want to tackle in the not so distant future. 

The static analysis recipe is not a good match for our code. Not even going to try to adjust it to do what we want.

the instanceOf recipe is similarly sound, but has room for improvement. I think sequence matters here. It should have run before the java21 updates. Good to know. 

I m going to split the work from this PR into smaller chunks.

line-o:
@duncdrum can this PR be closed now that the PRs that you split this one up are merged?

duncdrum:
@line-o not yet I m not sure where the review comments will go. I have plans for them. Will close when done

line-o:
@duncdrum I am surprised to see this is still open. I assume we should close it and rather run the scripts again on a fresh checkout. Correct?

duncdrum:
it is a draft, I still need the review comments, I will close it once someone had time to rerun the scripts and that PR is merged

PR #6244, opened 2026-04-15
Add Citation.cff and incorporate updates into maven build

### LINKED ISSUE #6243: [feature] add CITATION.cff to help our academic users
[GitHub Docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) 

Academia and citations have a way of catching up with you it seems. I think we should add a citation file. We have many academic users, and frankly I just stumbled upon both the requirement to cite exist-db, and a recent citation from an not-to-be-named fellow Dev that ..well let's just say could be better

duncdrum:
close #6243

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -11,13 +11,13 @@ identifiers:
   - type: url
     value: >-
       https://github.com/eXist-db/exist/releases/tag/eXist-6.4.1
-    description: The GitHub release URL of latest stable release
+    description: The URL of the latest GitHub release
 repository-code: 'https://github.com/eXist-db/exist'
 url: 'https://exist-db.org/'
 repository-artifact: 'https://central.sonatype.com/artifact/org.exist-db/exist'
 abstract: >-
   eXist-db is a high-performance open source native XML
-  database—a NoSQL document database and application
+  database — a NoSQL document database and application
   platform built entirely around XML technologies. 
 keywords:
   - xquery
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,32 @@
+# This CITATION.cff file was generated with cffinit.
+# Visit https://bit.ly/cffinit to generate yours today!
+
+cff-version: 1.2.0
+title: eXist-db
+message: Native XML Database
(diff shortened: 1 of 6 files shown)


CITATION.cff
```diff
@@ -0,0 +1,29 @@
...
+title: eXist-db
+message: Native XML Database
+type: software
+authors:
+  - name: The eXist-db authors
+identifiers:
+  - type: url
+    value: >-
+      https://github.com/eXist-db/exist/releases/tag/eXist-6.4.1
+    description: The GitHub release URL of latest stable release
```
line-o: How about:

The URL of the latest stable GitHub release

CITATION.cff
```diff
@@ -0,0 +1,29 @@
...
+repository-code: 'https://github.com/eXist-db/exist'
+url: 'https://exist-db.org/'
+repository-artifact: 'https://central.sonatype.com/artifact/org.exist-db/exist'
+abstract: >-
+  eXist-db is a high-performance open source native XML
+  database—a NoSQL document database and application
```
line-o: `database—a` looks weird to me

Whitespace around the em-dash at least maybe just resort to a full stop.

line-o (commented):
I like the idea of updating the citation file automatically. I did not read the documentation yet, does it need to be branch based? As currently it will be 6.4.1 even though one might cite 7.0.0-SNAPSHOT code, right?

duncdrum:
Not really branch based. The target for citations styles is usually a versioned release, ideally this file would have been generated with the release, and will only change with the next release. But bootstrapping has to happen at some point. The Github UI support for the file, will kick in once it appears on the default branch. 

Folks are free to build references from this for whatever fork branch commit etc. This is making the software as a whole citeable. It's not supposed to replace or overwrite code references in computer science dissertations. 

What fields would you suggest to add or modify? Take a look at the docs and let me know.

line-o (approved):
+1

joewiz (approved):

Issue #49, opened 2013-09-19
Spatial module doesn't build


gtnewton:
Environment: ubuntu 13.04 64-bit, OpenJDK 7, latest exist from git

Steps to reproduce:
1) Activate spatial module in $build-dir/extensions/indexes/local.build.properties
2) ./build.sh download-additional-jars
3) ./build.sh all

The end of the compile looks like this

```
compile-src:
     [echo] Compiling sources 'index-spatial'
    [mkdir] Created dir: /home/gtn/exist/extensions/indexes/spatial/build/classes
    [javac] Compiling 10 source files to /home/gtn/exist/extensions/indexes/spatial/build/classes
    [javac] warning: [options] bootstrap class path not set in conjunction with -source 1.6
    [javac] /home/gtn/exist/extensions/indexes/spatial/src/org/exist/indexing/spatial/GMLHSQLIndexWorker.java:61: error: GMLHSQLIndexWorker is not abstract and does not override abstract method getQueryRewriter(XQueryContext) in IndexWorker
    [javac] public class GMLHSQLIndexWorker extends AbstractGMLJDBCIndexWorker {
    [javac]        ^
    [javac] Note: Some input files use or override a deprecated API.
    [javac] Note: Recompile with -Xlint:deprecation for details.
    [javac] 1 error
    [javac] 1 warning

BUILD FAILED
/home/gtn/exist/build/scripts/build-impl.xml:415: The following error occurred while executing this line:
/home/gtn/exist/build/scripts/extensions-build.xml:31: The following error occurred while executing this line:
/home/gtn/exist/build/scripts/extensions-common.xml:137: The following error occurred while executing this line:
/home/gtn/exist/build/scripts/extensions-common.xml:199: The following error occurred while executing this line:
/home/gtn/exist/build/scripts/extensions-common.xml:152: The following error occurred while executing this line:
/home/gtn/exist/build/scripts/extensions-common.xml:163: Compile failed; see the compiler error output for details.
```

Also, note that the fetch url in $build-dir/extensions/indexes/spatial/extension.xml pointing to geoapi-nogenerics-2.1.1.jar (line 19) gets a 404. The file is available at the osgeo.org location used elsewhere in the file.

joewiz:
Close as a dupe of the more recently verified/labeled/decided on #306?

PR #6247, opened 2026-04-18
Add built-in package management REST API

joewiz:
## Summary

Add a built-in package management REST API to eXist-db, eliminating the need for the `existdb-packageservice` XAR package. The servlet handles package CRUD operations — list, get, install (from registry or upload), remove (with dependency checking), update-check, and icon serving.

## What Changed

All new code is in the **`exist-services`** Maven module (not `exist-core`), per the architectural principle that exist-core should contain only the core database engine.

### New module: `exist-services/`
- `PackageManagementServlet.java` — HTTP servlet at `/api/packages` for package CRUD
- `PackageService.java` — business logic for package operations (install, remove, dependency resolution)
- `RepoPackageLoader.java` — downloads packages from public/private registries

### Configuration
- `web.xml` — servlet registration with multipart config for XAR uploads
- `controller-config.xml` — URL pattern forwarding for `/api/packages`
- `pom.xml` — new `exist-services` module in reactor and `exist-distribution` dependency

## Motivation

The `existdb-packageservice` XAR package has a fundamental problem: it can't upgrade itself. The XQuery code executing the upgrade gets deleted mid-operation when the package is removed. By moving package management into the platform (compiled Java), this self-upgrade problem is eliminated.

## Test Plan

- [ ] `mvn install -pl exist-services -am -DskipTests` compiles cleanly
- [ ] Dashboard package management UI works (Cypress: `packages.cy.js`)
- [ ] `xst package install/remove/list` works against the built-in API
- [ ] public-repo app works (Cypress: all specs pass)

### DIFF
--- a/exist-core/src/main/java/org/exist/http/urlrewrite/Forward.java
+++ b/exist-core/src/main/java/org/exist/http/urlrewrite/Forward.java
@@ -21,6 +21,8 @@
  */
 package org.exist.http.urlrewrite;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.exist.http.servlets.HttpResponseWrapper;
 import org.w3c.dom.Element;
 
@@ -31,6 +33,8 @@ import jakarta.servlet.http.HttpServletResponse;
 import java.io.IOException;
 
 public abstract class Forward extends URLRewrite {
+
+    private static final Logger LOG = LogManager.getLogger(Forward.class);
 
     protected Forward(final Element config, final String uri) {
         super(config, uri);
@@ -44,7 +48,9 @@ public abstract class Forward extends URLRewrite {
     public void doRewrite(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
         final RequestDispatcher dispatcher = getRequestDispatcher(request);
         if (dispatcher == null) {
-            throw new ServletException("Failed to initialize request dispatcher to forward request to " + uri);
+            LOG.warn("No servlet registered for forward target '{}' — is the required module on the classpath?", uri);
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, "Servlet not available: " + uri);
+            return;
         }
         setHeaders(new HttpResponseWrapper(response));
         dispatcher.forward(request, response);
--- a/exist-core/src/main/java/org/exist/repo/RepoPackageLoader.java
+++ b/exist-core/src/main/java/org/exist/repo/RepoPackageLoader.java
@@ -1,0 +1,88 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.repo;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.exist.SystemProperties;
+import org.exist.util.io.TemporaryFileManager;
+import org.expath.pkg.repo.XarFileSource;
+import org.expath.pkg.repo.XarSource;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URI;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+
+/**
+ * Loads packages from a remote public repository (e.g. https://exist-db.org/exist/apps/public-repo).
+ * Used by {@link Deployment} to resolve package dependencies during installation,
+ * and by {@link PackageService} for direct package installation from a registry.
+ */
+public record RepoPackageLoader(String repoURL) implements PackageLoader {
+
+    private static final Logger LOG = LogManager.getLogger(RepoPackageLoader.class);
+
+    private static final int CONNECT_TIMEOUT = 15_000;
+    private static final int READ_TIMEOUT = 15_000;
+
+    @Override
+    public XarSource load(final String name, final Version version) throws IOException {
+        String pkgURL = repoURL + "?name=" + URLEncoder.encode(name, StandardCharsets.UTF_8) +
+                "&processor=" + SystemProperties.getInstance().getSystemProperty("product-version", "2.2.0");
+        if (version != null) {
+            if (version.getMin() != null) {
+                pkgURL += "&semver-min=" + version.getMin();
+            }
+            if (version.getMax() != null) {
+                pkgURL += "&semver-max=" + version.getMax();
+            }
+            if (version.getSemVer() != null) {
+                pkgURL += "&semver=" + version.getSemVer();
+            }
+            if (version.getVersion() != null) {
+                pkgURL += "&version=" + URLEncoder.encode(version.getVersion(), StandardCharsets.UTF_8);
+            }
+        }
+        LOG.info("Retrieving package from {}", pkgURL);
+        final HttpURLConnection connection = (HttpURLConnection) URI.create(pkgURL).toURL().openConnection();
+        connection.setConnectTimeout(CONNECT_TIMEOUT);
+        connection.setReadTimeout(READ_TIMEOUT);
+        connection.setRequestMethod("GET");
+        connection.setRequestProperty("User-Agent", "eXist-db Package Manager");
+        connection.connect();
+
+        try (final InputStream is = connection.getInputStream()) {
+            final TemporaryFileManager temporaryFileManager = TemporaryFileManager.getInstance();
+            final Path outFile = temporaryFileManager.getTemporaryFile();
+            Files.copy(is, outFile, StandardCopyOption.REPLACE_EXISTING);
+            return new XarFileSource(outFile);
+        } catch (final IOException e) {
+            throw new IOException("Failed to download package from " + pkgURL + ": " + e.getMessage(), e);
+        }
+    }
+}
--- a/exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
+++ b/exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
@@ -1,0 +1,432 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.http.servlets;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import jakarta.servlet.http.Part;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.exist.EXistException;
+import org.exist.repo.PackageService;
+import org.exist.security.Subject;
+import org.exist.storage.DBBroker;
+import org.exist.storage.txn.Txn;
+import org.expath.pkg.repo.PackageException;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Built-in REST API for package management.
+ * <p>
+ * This servlet provides package CRUD operations (list, get, install, remove,
+ * update-check) as a built-in Java endpoint that does not depend on any XAR
+ * package. This ensures package management remains available even during
+ * package upgrades, solving the self-upgrade problem that affects XQuery-based
+ * package management implementations.
+ * </p>
+ * <p>
+ * Endpoints:
+ * <ul>
+ *   <li>GET /api/packages — list all installed packages</li>
+ *   <li>GET /api/packages/{name} — get single package details</li>
+ *   <li>GET /api/packages/{name}/icon — get package icon</li>
+ *   <li>POST /api/packages/install — install from registry (JSON) or upload (multipart)</li>
+ *   <li>POST /api/packages/update-check — check for available updates</li>
+ *   <li>DELETE /api/packages/{name} — remove package</li>
+ * </ul>
+ * </p>
+ */
+public class PackageManagementServlet extends AbstractExistHttpServlet {
+
+    private static final Logger LOG = LogManager.getLogger(PackageManagementServlet.class);
+
+    private final PackageService packageService = new PackageService();
+
+    @Override
+    public Logger getLog() {
+        return LOG;
+    }
+
+    @Override
+    protected void doGet(final HttpServletRequest request, final HttpServletResponse response)
+            throws ServletException, IOException {
+        final Subject user = authenticate(request, response);
+        if (user == null) {
+            return;
+        }
+
+        final String pathInfo = normalizePath(request.getPathInfo());
+
+        try {
+            if (pathInfo == null || pathInfo.isEmpty() || pathInfo.equals("/")) {
+                // GET /api/packages — list all
+                handleListPackages(response);
+            } else if (pathInfo.endsWith("/icon")) {
+                // GET /api/packages/{name}/icon
+                final String name = pathInfo.substring(1, pathInfo.length() - "/icon".length());
+                handleGetIcon(response, name);
+            } else {
+                // GET /api/packages/{name}
+                final String name = pathInfo.substring(1);
+                handleGetPackage(response, name);
+            }
+        } catch (final PackageException e) {
+            writeError(response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+                    "PACKAGE_ERROR", e.getMessage());
+        }
+    }
+
+    @Override
+    protected void doPost(final HttpServletRequest request, final HttpServletResponse response)
+            throws ServletException, IOException {
+        final Subject user = authenticate(request, response);
+        if (user == null) {
+            return;
+        }
+        if (!user.hasDbaRole()) {
+            writeError(response, HttpServletResponse.SC_FORBIDDEN,
+                    "FORBIDDEN", "DBA role required for package installation");
+            return;
+        }
+
+        final String pathInfo = normalizePath(request.getPathInfo());
+
+        try {
+            if ("/install".equals(pathInfo)) {
+                handleInstall(request, response, user);
+            } else if ("/update-check".equals(pathInfo)) {
+                handleUpdateCheck(request, response);
+            } else {
+                writeError(response, HttpServletResponse.SC_BAD_REQUEST,
+                        "BAD_REQUEST", "Unknown POST endpoint: " + pathInfo);
+            }
+        } catch (final PackageException e) {
+            writeError(response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+                    "PACKAGE_ERROR", e.getMessage());
+        } catch (final EXistException e) {
+            writeError(response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+                    "DATABASE_ERROR", e.getMessage());
+        }
+    }
+
+    @Override
+    protected void doDelete(final HttpServletRequest request, final HttpServletResponse response)
+            throws ServletException, IOException {
+        final Subject user = authenticate(request, response);
+        if (user == null) {
+            return;
+        }
+        if (!user.hasDbaRole()) {
+            writeError(response, HttpServletResponse.SC_FORBIDDEN,
+                    "FORBIDDEN", "DBA role required for package removal");
+            return;
+        }
+
+        final String pathInfo = normalizePath(request.getPathInfo());
+        if (pathInfo == null || pathInfo.length() <= 1) {
+            writeError(response, HttpServletResponse.SC_BAD_REQUEST,
+                    "BAD_REQUEST", "Package name required");
+            return;
+        }
+
+        final String nameOrAbbrev = pathInfo.substring(1);
+
+        try {
+            // Check for dependents first
+            final List<String> dependents = packageService.findDependents(getPool(), nameOrAbbrev);
+            final boolean force = "true".equals(request.getParameter("force"));
+            if (!dependents.isEmpty() && !force) {
+                response.setStatus(HttpServletResponse.SC_CONFLICT);
+                response.setContentType("application/json");
+                try (final OutputStream os = response.getOutputStream();
+                     final JsonGenerator gen = createJsonGenerator(os)) {
+                    gen.writeStartObject();
+                    gen.writeStringField("status", "error");
+                    gen.writeStringField("code", "HAS_DEPENDENTS");
+                    gen.writeStringField("message", "Cannot remove: other packages depend on " + nameOrAbbrev);
+                    gen.writeArrayFieldStart("dependents");
+                    for (final String dep : dependents) {
+                        gen.writeString(dep);
+                    }
+                    gen.writeEndArray();
+                    gen.writeEndObject();
+                }
+                return;
+            }
+
+            try (final DBBroker broker = getPool().get(Optional.of(user));
+                 final Txn transaction = getPool().getTransactionManager().beginTransaction()) {
+                final Map<String, Object> result = packageService.removePackage(broker, transaction, nameOrAbbrev);
+                transaction.commit();
+                writeJsonMap(response, HttpServletResponse.SC_OK, result);
+            } catch (final EXistException e) {
+                writeError(response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+                        "DATABASE_ERROR", e.getMessage());
+            }
+        } catch (final PackageException e) {
+            if (e.getMessage() != null && e.getMessage().contains("not found")) {
+                writeError(response, HttpServletResponse.SC_NOT_FOUND,
+                        "PACKAGE_NOT_FOUND", e.getMessage());
+            } else {
+                writeError(response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+                        "PACKAGE_ERROR", e.getMessage());
+            }
+        }
+    }
+
+    @Override
+    protected void doOptions(final HttpServletRequest request, final HttpServletResponse response)
+            throws ServletException, IOException {
+        response.setHeader("Access-Control-Allow-Origin", "*");
+        response.setHeader("Access-Control-Allow-Methods", "GET, POST, DELETE, OPTIONS");
+        response.setHeader("Access-Control-Allow-Headers", "Content-Type, Authorization");
+        response.setStatus(HttpServletResponse.SC_OK);
+    }
+
+    // --- Handler methods ---
+
+    private void handleListPackages(final HttpServletResponse response)
+            throws IOException, PackageException {
+        final List<Map<String, Object>> packages = packageService.listPackages(getPool());
+        response.setContentType("application/json");
+        response.setStatus(HttpServletResponse.SC_OK);
+        try (final OutputStream os = response.getOutputStream();
+             final JsonGenerator gen = createJsonGenerator(os)) {
+            gen.writeStartArray();
+            for (final Map<String, Object> pkg : packages) {
+                writeMapAsJson(gen, pkg);
+            }
+            gen.writeEndArray();
+        }
+    }
+
+    private void handleGetPackage(final HttpServletResponse response, final String nameOrAbbrev)
+            throws IOException, PackageException {
+        final Map<String, Object> pkg = packageService.getPackage(getPool(), nameOrAbbrev);
+        if (pkg == null) {
+            writeError(response, HttpServletResponse.SC_NOT_FOUND,
+                    "PACKAGE_NOT_FOUND", "Package not found: " + nameOrAbbrev);
+            return;
+        }
+        writeJsonMap(response, HttpServletResponse.SC_OK, pkg);
+    }
+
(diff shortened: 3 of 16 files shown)


adamretter:
May I suggest that this should go into its own separate module. The exist-core module was created with the specific purpose of being slimmed down rather than made bigger, the previous plans was to remove everything from exist-core apart from core database engine.

line-o:
@joewiz I am intrigued by the approach you chose here. I assume that this does integrate with your other proposals to extend and unify the REST API exposed by exist-db.

I do see testfailures.

Since xst is not using the packagemanager/exist-packageservice at all it can be removed from the testplan.
This is not implying at all that xst could not benefit from a good API around packages. But ultimately this would need to be implemented in node-exist and xst and gulp-exist would both improve.

joewiz:

@line-o — test failures fixed. Three root causes:

1. **License check**: `exist-services/pom.xml` was missing the LGPL header
2. **Unit test failures** (URLRewritingTest, ControllerTest, JmxRemoteTest all 503): `PackageManagementServlet` was registered in `web.xml` with `load-on-startup`, causing Jetty to fail when the servlet class wasn't on exist-core's test classpath. Fix: moved servlet registration from `web.xml` to a `web-fragment.xml` in the exist-services JAR — only loaded when the JAR is present
3. **Integration test** (dependency:analyze): exist-services used 4 transitive deps without declaring them. Now declared explicitly.

Also made `Forward.java` return 404 gracefully instead of throwing when a target servlet is missing (defensive improvement).

6,559 tests, 0 failures.

Re: xst — agreed, the test plan reference was misleading. Removed.

Re: @adamretter's module suggestion — that's a good point about keeping exist-core slim. The exist-services module in this PR is now separate from exist-core.

line-o:
@joewiz should this be added to existdb-openapi ?

duncdrum:
my main concern is that since we will install dashboard on way or the other, we ll have two package services running? Should this be part of a larger group of PRs?

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
```diff
@@ -0,0 +1,430 @@
...
+                        gen.writeString(String.valueOf(item));
+                    }
+                }
+                gen.writeEndArray();
+            } else if (value instanceof Map<?, ?> nested) {
+                gen.writeFieldName(key);
+                writeMapAsJson(gen, (Map<String, Object>) nested);
+            } else {
+                gen.writeStringField(key, String.valueOf(value));
+            }
```
reinhapa: This could be changed to a switch expression

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
```diff
@@ -0,0 +1,430 @@
...
+        }
+        gen.writeEndObject();
+    }
+
+    private String normalizePath(final String pathInfo) {
+        if (pathInfo == null) {
+            return null;
+        }
+        // URL-decode the path
+        return java.net.URLDecoder.decode(pathInfo, java.nio.charset.StandardCharsets.UTF_8);
```
reinhapa: Use imports instead of canonical class names

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
```diff
@@ -0,0 +1,430 @@
...
+            writeError(response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+                    "PACKAGE_ERROR", e.getMessage());
+        } catch (final EXistException e) {
+            writeError(response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+                    "DATABASE_ERROR", e.getMessage());
+        }
+    }
+
+    @Override
+    protected void doDelete(final HttpServletRequest request, final HttpServletResponse response)
```
reinhapa, line 140: Address Codacy issue: _The method 'doDelete(HttpServletRequest, HttpServletResponse)' has an NPath complexity of 216, current threshold is 200_

exist-services/src/main/java/org/exist/repo/PackageService.java
```diff
@@ -0,0 +1,474 @@
...
+
+/**
+ * Business logic for package management operations.
+ * Used by {@link org.exist.http.servlets.PackageManagementServlet} to provide
+ * a built-in REST API for package CRUD that does not depend on any XAR package.
+ */
+public class PackageService {
+
+    private static final Logger LOG = LogManager.getLogger(PackageService.class);
+
```
reinhapa, line 58: remove this empty line

exist-services/src/main/java/org/exist/repo/RepoPackageLoader.java
```diff
@@ -0,0 +1,90 @@
...
+ * Loads packages from a remote public repository (e.g. https://exist-db.org/exist/apps/public-repo).
+ * Used by {@link Deployment} to resolve package dependencies during installation,
+ * and by {@link PackageService} for direct package installation from a registry.
+ *
+ * @param repoURL the base URL of the public package repository
+ */
+public record RepoPackageLoader(String repoURL) implements PackageLoader {
+
+    private static final Logger LOG = LogManager.getLogger(RepoPackageLoader.class);
+
```
reinhapa, line 50: remove empty line

reinhapa (changes requested):

joewiz:
[This text was co-authored with Claude Code. -Joe]

This comment was prompted by @line-o and @duncdrum's questions above:

> should this be added to existdb-openapi ?

> since we will install dashboard on way or the other, we ll have two package services running?

## Package management: why it needs a home in core, and how it converges with existdb-openapi

**The motivation first — why not just leave package management to a package?** Because a package can't safely manage itself. existdb-openapi's `modules/packages.xqm` runs `repo:remove` / `repo:install-and-deploy` from XQuery stored *in the existdb-openapi XAR*. The moment you use that handler to upgrade or remove existdb-openapi itself (or anything it depends on), it deletes its own executing code mid-call — `remove` even clears the XQuery pool. So there must be at least one package-management entry point that is **not itself a package**: compiled, in core, able to manage any package including the package layer itself. That bootstrap case is the irreducible reason this PR exists.

**So why two surfaces, and aren't they redundant?** They're complementary, divided by *self-management capability* rather than by feature:

- **Core `/api/packages` (this PR)** — a compiled-Java `PackageService` behind `PackageManagementServlet` (new `exist-services` module), at server-root. The bootstrap-safe platform floor: always present, survives any package being removed, and the only surface that can safely upgrade/remove existdb-openapi itself.
- **existdb-openapi `/api/packages` (XQuery/Roaster, + existdb-openapi#49)** — the app-level surface for the common case (installing/removing *other* packages), consistent with the rest of the openapi API that clients (Oxygen plugin, eXide, vscode-existdb, the notebook kernel) already speak.

The operation sets are 1:1 (list / get / icon / install-from-registry / install-from-upload / remove / update-check) and there's no path collision (server-root vs app-scoped), so nothing is blocked today — the question is how they should relate.

### Direction: one engine in core; no duplicate logic

The thing to avoid is *two implementations* of package management drifting apart. So core owns the engine and the bootstrap-safe entry point — that part is non-negotiable for the self-upgrade case. What's genuinely open (see below) is the **end state of the app surface**: after convergence, should existdb-openapi drop its package endpoints in favor of the single core surface, or keep them as a **thin delegator** to the core engine for API-surface uniformity? Either way there is exactly one implementation. Our lean is the single core engine as the canonical implementation; whether the app keeps a delegating front door is the softer, uniformity-driven call.

This isn't a stub standing in for a real implementation: #6247 already authenticates and enforces `hasDbaRole()` (403) on install and remove, and `installFromUpload` already parses `multipart/`. So this is convergence/migration, not new capability.

### The work: three gates + sequencing

1. **Contract alignment.** The two surfaces differ in details that would silently break a consumer re-pointing between them — clearest example, the multipart field name: existdb-openapi#49 takes the file as `file` (Roaster's form-data shape), while this servlet reads `request.getPart("xar")`. These, plus the JSON shapes for list/get/icon and the error envelope, need reconciling so `/api/packages` is a drop-in for what the openapi consumers expect.
2. **OpenAPI description of `/api/packages`** so consumers keep the documented, generated-client-friendly contract they had under Roaster (see core-input item 2).
3. **Consumer migration.** Inventory and re-point everything that calls the openapi package endpoints (or `repo:*` directly): the Oxygen plugin, eXide, vscode-existdb, the notebook kernel, and the `existdb-packageservice` / Dashboard app.

**Sequence — do not retire existdb-openapi's package endpoints (or close existdb-openapi#49) first.** Land #6247 → migrate consumers to `/api/packages` → *then* settle the app-surface end state and retire existdb-openapi#49's standalone XQuery install path in favor of the core engine. Closing #49 before `/api/packages` is consumer-ready would open a regression window on the Oxygen plugin's "Install `.xar`…" feature. If that feature is wanted *in* the next beta and #6247 can't be merged-and-migrated in time, existdb-openapi#49 can ship as a documented interim and the openapi endpoints be removed (or converted to delegation) as a fast-follow.

### Needs core-team input (not settled between the strategy threads)

1. **App-surface end state — retire or delegate?** After convergence, should existdb-openapi drop its package endpoints (single core surface) or keep them as a thin delegator to the core engine? This is the "why two?" decision; the bootstrap mechanism in core is required either way, but the app-surface choice is a product/architecture call. (If we want delegation, that's the optional follow-up of exposing `PackageService` as an XQuery function module — useful for in-process callers that are *not* self-managing, but never the self-management path.)
2. **Should `/api/packages` be served/described via the built-in OpenAPI-routing direction (#6248), and is that direction wanted in core?** Gate 2 quietly reframes this PR from "a servlet" into "the first consumer of a built-in OpenAPI-routing layer in eXist core" — an architecture-of-core decision the team should buy into. Refinement we'd suggest: keep the requirement as "the endpoint must be OpenAPI-described," but **don't block #6247 on the full #6248 framework** — an interim hand-written description is acceptable if #6248 lags, so package-management convergence isn't held hostage to a larger piece.
3. **Security review of the new root servlet.** A server-root `/api/packages` that ships CORS headers and performs dba-gated install/remove is a security-sensitive core addition and wants a core security reviewer's eyes — CORS scope, CSRF for any cookie-auth path, and dba enforcement on every mutating op (and ideally the read ops too).

Feedback on the direction, and especially the three items above, would be much appreciated before we wire up the contract alignment and consumer migration.

joewiz:

Thanks @reinhapa — all addressed in 2f38194a1e:

- **Switch expression** — `writeMapAsJson` now uses pattern-matching `switch` instead of the `instanceof` if/else chains (outer dispatch and the nested list-item loop).
- **Canonical class names** — added imports for `URLDecoder` and `StandardCharsets`; the inline `java.net.`/`java.nio.charset.` references are gone.
- **`doDelete` NPath (216)** — now under the threshold, by extracting `writeDependentsConflict()` and `writePackageError()`.
- **Empty lines** — removed the blank between the type declaration and the `LOG` field in both `PackageService` and `RepoPackageLoader`.

While in `PackageService` I also brought `readExpathMetadata` under the NPath threshold (it was 262 — same file, not separately flagged) by extracting `readDependency()` / `putAttributeIfPresent()`, so Codacy stays clean across the changed files. `exist-services` compiles and Codacy PMD is clean on all four changed files.

Separately: this branch is now well behind `develop`. I've kept it on its original base for now so the review diff stays focused, but I'm happy to rebase onto current `develop` whenever you'd like it — just say the word (or if you'd rather review first and rebase right before merge, that works too).

duncdrum:
@joewiz needs a rebase

joewiz:
@duncdrum Ok, rebased.

duncdrum:
Thank you @joewiz two failures on CI.

joewiz:

Thanks @duncdrum — both were real, and both are fixed in 70f1d4a2a9.

**Native XSD checks.** The rebase brought the branch under the schema governance gate, and adding the `/api/packages` forward to the canonical `controller-config.xml` counts as a template change: `Template ... changed but @version is still "1.1.1"`. Bumped `controller-config.xsd` to 1.1.2 and kept the `schemaVersion` attribute on both canonical instances in step. I made it a PATCH rather than a MINOR because only the template's content changed — `forward` was already allowed by the grammar, so no new element or attribute is introduced and existing instance documents stay valid. Shout if you'd rather see MINOR.

**Windows integration.** This one turned out to be a latent bug in the branch rather than rebase fallout: `exist-services/pom.xml` pinned its parent to the literal `7.0.0-SNAPSHOT` instead of `${revision}` like every other module, so Maven failed while scanning projects with `Non-resolvable parent POM ... 'parent.relativePath' points at wrong local POM`. Linux never caught it because `exist-parent` was already in the local repository; building `exist-services` against an empty local repo reproduces it exactly, and switching to `${revision}` clears it.

Verified locally before pushing: `prepare-governance-context.sh` + `mvn -N xml:transform@schema-governance` against current `develop` passes, and `mvn validate` is clean across the reactor including the canonical-instance validation against the bumped schema.

One other thing worth flagging from the rebase, since it isn't visible in the diff: `develop` now generates each module's test `controller-config.xml` from the canonical template via `controller-config-fixture.xsl`, so this branch's old hand-edits to `exist-core/src/test/resources/standalone-webapp/WEB-INF/controller-config.xml` are gone rather than re-applied. All six fixtures use explicit `keep-forwards` allow-lists, so `/api/packages` is filtered out of the generated test configs and there's no test-side impact.

duncdrum:
Thank you @joewiz. About those handrolled edits, if we don't need them in the test suite, leaving them out makes sense. If we should add them back in, the way is via xsl.

duncdrum (commented):
Nice work standing up the built-in package API! Left a handful of inline notes — mostly a few unguarded edge cases that can throw uncaught exceptions, two temp-file leaks, a missing-tests flag, and some design/duplication points worth a look.

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
```diff
@@ -0,0 +1,442 @@
...
+        gen.writeEndObject();
+    }
+
+    private String normalizePath(final String pathInfo) {
+        if (pathInfo == null) {
+            return null;
+        }
+        // URL-decode the path
+        return URLDecoder.decode(pathInfo, StandardCharsets.UTF_8);
+    }
```
duncdrum, line 441: `request.getPathInfo()` is already URL-decoded by the servlet container, so decoding it again here double-decodes valid paths (a literal `+` becomes a space) and throws an uncaught `IllegalArgumentException` for a bare `%` not followed by two hex digits — e.g. `GET /api/packages/my%pkg` would 500 instead of returning a clean 4xx, since only `PackageException` is caught in the callers.

```suggestion
    private String normalizePath(final String pathInfo) {
        // getPathInfo() is already URL-decoded by the servlet container — decoding
        // it again double-decodes valid paths and throws on a bare '%'.
        return pathInfo;
    }
```

If you go this route, `java.net.URLDecoder` becomes an unused import and can be dropped too.

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
```diff
@@ -0,0 +1,442 @@
...
+        final String pathInfo = normalizePath(request.getPathInfo());
+
+        try {
+            if (pathInfo == null || pathInfo.isEmpty() || pathInfo.equals("/")) {
+                // GET /api/packages — list all
+                handleListPackages(response);
+            } else if (pathInfo.endsWith("/icon")) {
+                // GET /api/packages/{name}/icon
+                final String name = pathInfo.substring(1, pathInfo.length() - "/icon".length());
+                handleGetIcon(response, name);
```
duncdrum, line 94: `GET /api/packages/icon` (no package name) makes `pathInfo` exactly `"/icon"`, so `substring(1, 5 - 5)` becomes `substring(1, 0)` and throws `StringIndexOutOfBoundsException` — uncaught, so it surfaces as a 500 instead of a 400.

```suggestion
            } else if (pathInfo.endsWith("/icon")) {
                // GET /api/packages/{name}/icon
                final String name = pathInfo.substring(1, pathInfo.length() - "/icon".length());
                if (name.isEmpty()) {
                    writeError(response, HttpServletResponse.SC_BAD_REQUEST,
                            "BAD_REQUEST", "Package name is required");
                } else {
                    handleGetIcon(response, name);
                }
```

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
```diff
@@ -0,0 +1,442 @@
...
+            }
+            try (final InputStream is = xarPart.getInputStream();
+                 final DBBroker broker = getPool().get(Optional.of(user));
+                 final Txn transaction = getPool().getTransactionManager().beginTransaction()) {
+                final Map<String, Object> result = packageService.installFromUpload(
+                        broker, transaction, is);
+                transaction.commit();
+                writeJsonMap(response, HttpServletResponse.SC_OK, result);
+            }
+        } catch (final ServletException e) {
```
duncdrum, line 331: Jetty/Tomcat throw `IllegalStateException` (not `ServletException`) from `getPart()` when an upload exceeds the configured `max-file-size`/`max-request-size`. That's not caught here or by `doPost`'s other catches, so a too-large XAR upload 500s instead of getting a clean 400.

```suggestion
        } catch (final ServletException | IllegalStateException e) {
```

exist-core/src/main/java/org/exist/http/urlrewrite/Forward.java
```diff
@@ -44,7 +48,9 @@ protected Forward(final URLRewrite other) {
     public void doRewrite(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
         final RequestDispatcher dispatcher = getRequestDispatcher(request);
         if (dispatcher == null) {
-            throw new ServletException("Failed to initialize request dispatcher to forward request to " + uri);
+            LOG.warn("No servlet registered for forward target '{}' — is the required module on the classpath?", uri);
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, "Servlet not available: " + uri);
+            return;
         }
```
duncdrum, line 54: Heads up: `Forward` is the shared base class for *every* `<forward>` entry in `controller-config.xml`, not just the new package API route — so this change to log-and-404 instead of throwing softens failures app-wide. A typo'd servlet name in `/rest`, `/webdav`, etc. would previously surface as a loud 500/`ServletException` (visible to monitoring); now it silently degrades to a per-request 404 with only a WARN log line, which could make a broken deployment look like ordinary "not found" traffic.

If the goal is just to let the package API's forward target be optional (module not on the classpath), might be worth scoping the graceful-404 behavior to that specific subclass instead of changing the shared base class's contract for all forwards.

exist-services/src/main/java/org/exist/repo/PackageService.java
```diff
@@ -0,0 +1,474 @@
...
+        final int startQuote = json.indexOf('"', colonIdx + 1);
+        if (startQuote < 0) {
+            return null;
+        }
+        final int endQuote = json.indexOf('"', startQuote + 1);
+        if (endQuote < 0) {
+            return null;
+        }
+        return json.substring(startQuote + 1, endQuote);
+    }
```
duncdrum, line 473: This hand-rolled substring scan has no awareness of JSON nesting or escaping, so a body like `{"other":{"name":"nested"},"url":"..."}` or a value containing an escaped quote can make it find the wrong/truncated occurrence of a key and silently return an incorrect value instead of failing loudly. `jackson-core` is already a dependency in this module (used for `JsonGenerator` in the servlet) — its streaming `JsonParser` would avoid this class of bug cheaply:

```suggestion
    public static String extractJsonStringValue(final String json, final String key) {
        try (final com.fasterxml.jackson.core.JsonParser parser =
                     new com.fasterxml.jackson.core.JsonFactory().createParser(json)) {
            if (parser.nextToken() != com.fasterxml.jackson.core.JsonToken.START_OBJECT) {
                return null;
            }
            while (parser.nextToken() == com.fasterxml.jackson.core.JsonToken.FIELD_NAME) {
                final String fieldName = parser.currentName();
                parser.nextToken();
                if (fieldName.equals(key) && parser.currentToken() == com.fasterxml.jackson.core.JsonToken.VALUE_STRING) {
                    return parser.getText();
                }
                parser.skipChildren();
            }
            return null;
        } catch (final java.io.IOException e) {
            return null;
        }
    }
```

(Used fully-qualified names to keep this a self-contained suggestion — worth switching to proper imports if you take it.)

exist-services/src/main/java/org/exist/repo/PackageService.java
```diff
@@ -0,0 +1,474 @@
...
+        // Read package name from the XAR descriptor
+        final Optional<org.exist.dom.memtree.DocumentImpl> descriptor = deployment.getDescriptor(broker, xarSource);
+        final String name = descriptor.map(d -> d.getDocumentElement().getAttribute("name")).orElse("unknown");
+
+        final Map<String, Object> result = new LinkedHashMap<>();
+        result.put("status", "ok");
+        result.put("name", name);
+        target.ifPresent(t -> result.put("target", t));
+        return result;
+    }
```
duncdrum, line 179: `tempFile` is never released back to `TemporaryFileManager` — only `deleteOnExit()` protects it (registered internally by the manager), which the manager's own javadoc documents as unreliable. Each upload leaks up to the XAR's size on disk until restart.

```suggestion
    public Map<String, Object> installFromUpload(final DBBroker broker, final Txn transaction,
                                                  final InputStream xarStream) throws PackageException, IOException {
        // Save to temporary file
        final TemporaryFileManager tempManager = TemporaryFileManager.getInstance();
        final Path tempFile = tempManager.getTemporaryFile();
        try {
            Files.copy(xarStream, tempFile, StandardCopyOption.REPLACE_EXISTING);

            final XarFileSource xarSource = new XarFileSource(tempFile);
            final Deployment deployment = new Deployment();
            final Optional<String> target = deployment.installAndDeploy(broker, transaction, xarSource, null);

            // Read package name from the XAR descriptor
            final Optional<org.exist.dom.memtree.DocumentImpl> descriptor = deployment.getDescriptor(broker, xarSource);
            final String name = descriptor.map(d -> d.getDocumentElement().getAttribute("name")).orElse("unknown");

            final Map<String, Object> result = new LinkedHashMap<>();
            result.put("status", "ok");
            result.put("name", name);
            target.ifPresent(t -> result.put("target", t));
            return result;
        } finally {
            tempManager.returnTemporaryFile(tempFile);
        }
    }
```

exist-services/src/main/java/org/exist/repo/RepoPackageLoader.java
```diff
@@ -0,0 +1,89 @@
...
+        connection.connect();
+
+        try (final InputStream is = connection.getInputStream()) {
+            final TemporaryFileManager temporaryFileManager = TemporaryFileManager.getInstance();
+            final Path outFile = temporaryFileManager.getTemporaryFile();
+            Files.copy(is, outFile, StandardCopyOption.REPLACE_EXISTING);
+            return new XarFileSource(outFile);
+        } catch (final IOException e) {
+            throw new IOException("Failed to download package from " + pkgURL + ": " + e.getMessage(), e);
+        }
```
duncdrum, line 87: Same temp-file leak as `installFromUpload` below — `outFile` is never returned to `TemporaryFileManager`, relying only on `deleteOnExit()`. Every registry install and every dependency resolved during install (`PackageService.installFromRegistry` / `Deployment`) leaves a downloaded XAR on disk indefinitely.

Unlike the upload case, the fix isn't fully local: `load()` hands the `XarFileSource` (wrapping this temp file) back to the caller, which may still be reading from it during `installAndDeploy`. Cleanest fix is probably for the caller (`PackageService.installFromRegistry`) to track and release the temp path once `deployment.installAndDeploy(...)` returns, mirroring the `try/finally` suggested for `installFromUpload`.

exist-services/src/main/java/org/exist/repo/RepoPackageLoader.java
```diff
@@ -0,0 +1,89 @@
...
+import java.nio.file.StandardCopyOption;
+
+/**
+ * Loads packages from a remote public repository (e.g. https://exist-db.org/exist/apps/public-repo).
+ * Used by {@link Deployment} to resolve package dependencies during installation,
+ * and by {@link PackageService} for direct package installation from a registry.
+ *
+ * @param repoURL the base URL of the public package repository
+ */
+public record RepoPackageLoader(String repoURL) implements PackageLoader {
```
duncdrum, line 48: This duplicates the existing private inner `RepoPackageLoader` record in `extensions/modules/expathrepo/.../Deploy.java` almost verbatim (same query-URL construction, same download-to-temp-file logic) rather than extracting a shared implementation. They've already drifted slightly — different `User-Agent` string and exception cause-chaining — so a future fix to one (e.g. the temp-file leak above, or the query parameter format) won't automatically propagate to the other. Worth considering extracting one shared implementation that both `Deploy` and this module use.

exist-services/src/main/java/org/exist/repo/PackageService.java
```diff
@@ -0,0 +1,474 @@
...
+        if (maybeRepo.isEmpty()) {
+            return Collections.emptyList();
+        }
+        final Repository repo = maybeRepo.get().getParentRepo();
+        final String findUrl = registryUrl + "/find";
+        final String processorVersion = org.exist.SystemProperties.getInstance()
+                .getSystemProperty("product-version", "7.0.0");
+
+        final List<Map<String, Object>> updates = new ArrayList<>();
+        for (final Packages packages : repo.listPackages()) {
```
duncdrum, line 268: `checkUpdates()` queries the registry for every installed package sequentially (10s connect + 10s read timeout each), with no concurrency and no overall bound on total latency. An instance with many packages and a slow/unresponsive registry can block `POST /api/packages/update-check` for up to N × ~20s. Worth considering a small bounded `ExecutorService` (or `HttpClient`'s async API) to fan these out, especially since each iteration is independent.

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
```diff
@@ -0,0 +1,442 @@
...
+ * <ul>
+ *   <li>GET /api/packages — list all installed packages</li>
+ *   <li>GET /api/packages/{name} — get single package details</li>
+ *   <li>GET /api/packages/{name}/icon — get package icon</li>
+ *   <li>POST /api/packages/install — install from registry (JSON) or upload (multipart)</li>
+ *   <li>POST /api/packages/update-check — check for available updates</li>
+ *   <li>DELETE /api/packages/{name} — remove package</li>
+ * </ul>
+ */
+public class PackageManagementServlet extends AbstractExistHttpServlet {
```
duncdrum, line 66: Flagging here too (see review summary): this new servlet — and the ~6 endpoints it exposes with specific status codes and JSON bodies — has no test coverage anywhere in the diff. Per this repo's guidance, HTTP/transport-layer behavior like this should get a `RESTServiceTest`-style integration test so regressions in status codes or JSON shape get caught.

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Thanks — the two failures are real, but the decode itself turns out to be needed. The servlet is reached through the controller's `<forward>`, and the forwarded request's path info comes from `XQueryURLRewrite.RequestWrapper`, which takes it from the raw `getRequestURI()`. So it arrives still percent-encoded. Package names are usually URIs (`/api/packages/http%3A%2F%2Fexist-db.org%2Fapps%2Fdashboard`), so dropping the decode would break lookups by name.

What fc228593b8 changes instead: only percent escapes are decoded, so `+` stays a literal plus, and a malformed escape such as a bare `%` now answers `400` instead of escaping as a 500.

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Fixed as you suggested: an empty name answers `400`.

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Fixed, with two adjustments. It answers `413 Payload Too Large`, which says more than a 400. And the `IllegalStateException` is caught around `getPart` only: a catch around the whole install would also turn an unrelated `IllegalStateException` from the installation into "upload too large".

exist-core/src/main/java/org/exist/http/urlrewrite/Forward.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Agreed, and it reached further than `controller-config.xml`: `PathForward` also handles `<exist:forward servlet="…">` from `controller.xq`, so app URL rewriting was affected too. Now a forward answers 404 for a missing servlet only when it is marked `optional="true"`. Every other forward fails loudly as before. The package API's forward carries the attribute, and `controller-config.xsd` declares it (that's what the 1.1.2 schema bump now covers). `PathForwardOptionalTest` covers both cases.

exist-services/src/main/java/org/exist/repo/PackageService.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Replaced with Jackson's streaming parser, as you suggested, with imports and a shared `JsonFactory`. It reads top-level members only. `PackageServiceJsonTest` covers a nested object with the same key, escaped quotes, and missing, non-string and malformed input.

exist-services/src/main/java/org/exist/repo/PackageService.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Fixed with a `try`/`finally` that returns the file to `TemporaryFileManager` once the installation has read it.

exist-core/src/main/java/org/exist/repo/RepoPackageLoader.java
```diff
@@ -0,0 +1,89 @@
...
+        connection.connect();
+
+        try (final InputStream is = connection.getInputStream()) {
+            final TemporaryFileManager temporaryFileManager = TemporaryFileManager.getInstance();
+            final Path outFile = temporaryFileManager.getTemporaryFile();
+            Files.copy(is, outFile, StandardCopyOption.REPLACE_EXISTING);
+            return new XarFileSource(outFile);
+        } catch (final IOException e) {
+            throw new IOException("Failed to download package from " + pkgURL + ": " + e.getMessage(), e);
+        }
```
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Fixed in the loader itself rather than in each caller: `RepoPackageLoader` is now `AutoCloseable`. It records every file it downloads, including dependencies resolved during `installAndDeploy`, and `close()` returns them all. Both callers use it in a try-with-resources that spans the installation, so the files stay available as long as they're read.

exist-services/src/main/java/org/exist/repo/RepoPackageLoader.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Done. There's now one `RepoPackageLoader`, in exist-core's `org.exist.repo` next to `Deployment` and `PackageLoader`, and `repo:install-and-deploy` uses it instead of its private copy. That brings `Deploy` the temp-file fix and the exception cause-chaining too. On the User-Agent: I kept "eXist-db Package Manager" over the old Firefox 3.5 string; the public repo answers both the same way.

exist-services/src/main/java/org/exist/repo/PackageService.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Done: the lookups run four at a time on a small fixed pool, with a 60-second bound on the whole check. Lookups still running at the deadline are cancelled, and the check reports what it has.

exist-services/src/main/java/org/exist/http/servlets/PackageManagementServlet.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Agreed, it needs HTTP-level tests. This round adds unit tests for the pieces I changed (`PathForwardOptionalTest` in exist-core, `PackageServiceJsonTest` as exist-services' first test). A `RESTServiceTest`-style test of the endpoints needs the Jetty test harness exist-core uses, set up in exist-services, which has had none so far. I'll add that in a follow-up commit on this PR, covering status codes and JSON shapes for each endpoint, including the 400/413 cases above.

duncdrum (approved):

PR #6248, opened 2026-04-18
PoC: Built-in OpenAPI routing with controller.json

joewiz:
## Summary

Add built-in OpenAPI routing to eXist-db via `controller.json` + `api.json`. Apps define REST endpoints declaratively in an OpenAPI 3.0 spec; eXist automatically routes requests to XQuery handler functions by `operationId`. This eliminates the need for Roaster and the `controller.xql` + `api.xq` + `$lookup` boilerplate.

## What Changed

All new code is in the **`exist-services`** Maven module (not `exist-core`), per the architectural principle that exist-core should contain only the core database engine.

### New classes in `exist-services/`
- `OpenApiServiceRegistry.java` — in-memory routing table with path pattern matching
- `OpenApiServlet.java` — dispatches matched routes to XQuery handlers
- `OpenApiStartupTrigger.java` — scans `/db/apps/` at startup for `controller.json`
- `OpenApiTrigger.java` — collection trigger for auto-discovery on storage events

### Configuration
- `web.xml` — OpenApiServlet registration
- `controller-config.xml` — URL pattern forwarding for `/openapi/`
- `conf.xml` — startup trigger registration

### Demo app
- `exist-core/src/test/resources/openapi-demo/` — three endpoints, no Roaster, no controller.xql

## Architecture

```
controller.json (in app collection)
    └── references api.json (OpenAPI 3.0 spec)
            └── operationId: "prefix:functionName"
                    └── OpenApiServlet resolves to XQuery module + function
                            └── Handler returns map → servlet serializes as JSON
```

## Relationship to PR #6247

This PR builds on #6247 (`feature/builtin-package-api`). Both share the `exist-services` module.

## Test Plan

- [ ] `mvn install -pl exist-services -am -DskipTests` compiles cleanly
- [ ] Dashboard tests pass without Roaster installed
- [ ] Apps using `controller.json` route correctly via OpenApiServlet


adamretter:
May I suggest that this should go into its own module. The exist-core module was created with the specific purpose of being slimmed down rather than made bigger, the previous plans was to remove everything from exist-core apart from core database engine.

exist-services/src/main/java/org/exist/http/openapi/OpenApiServlet.java
```diff
@@ -0,0 +1,382 @@
...
+ *   <li>If not found: return false to let the filter chain continue (falls through to XQueryURLRewrite)</li>
+ * </ol>
+ * </p>
+ */
+public class OpenApiServlet extends AbstractExistHttpServlet {
+
+    private static final Logger LOG = LogManager.getLogger(OpenApiServlet.class);
+
+    @Override
+    public Logger getLog() {
```
dizzzz, line 72: not sure why ?

joewiz: Hey Dannes! Claude provided what I hope is an answer to your question:

> AbstractExistHttpServlet.getLog() is abstract — every subclass must provide its logger instance. This is the standard pattern used by all eXist servlets (e.g., RESTServer, XQueryServlet). The OpenApiServlet extends AbstractExistHttpServlet to get broker pool access, authentication handling, and request lifecycle management for free.

Does that answer your question? Happy to take another stab.

exist-services/src/main/java/org/exist/repo/RepoPackageLoader.java
```diff
@@ -0,0 +1,88 @@
...
+            }
+            if (version.getSemVer() != null) {
+                pkgURL += "&semver=" + version.getSemVer();
+            }
+            if (version.getVersion() != null) {
+                pkgURL += "&version=" + URLEncoder.encode(version.getVersion(), StandardCharsets.UTF_8);
+            }
+        }
+        LOG.info("Retrieving package from {}", pkgURL);
+        final HttpURLConnection connection = (HttpURLConnection) URI.create(pkgURL).toURL().openConnection();
```
dizzzz: I am not sture if HttpURLConnection is the way to go.  E.g. there are situations where a (corporate) proxy needs to be used. The Java9+ HttpClient is probably a better, more modern solution. Maybe for a future PR?

joewiz:

Folded the switch in here rather than as a follow-up — pushed as `5959ce8c01`. `RepoPackageLoader` now uses `java.net.http.HttpClient` with `ProxySelector.getDefault()` and `followRedirects(NORMAL)`, so deployments behind a corporate proxy get the right behaviour out of the box via the standard `-Dhttp.proxyHost` / `-Dhttps.proxyHost` / `-Djava.net.useSystemProxies=true` hooks.

Two semantic differences worth flagging:

- The `load()` interface only declares `IOException`, but `HttpClient.send` throws `InterruptedException` too — caught and rethrown wrapped in `IOException` with `Thread.currentThread().interrupt()` to preserve the flag.
- `HttpClient` does *not* throw on non-2xx responses the way `HttpURLConnection.getInputStream()` did, so I added an explicit `statusCode` check before streaming the body.

Happy path is unchanged: GET, 15s connect + request timeouts, `User-Agent: eXist-db Package Manager`, body piped into `TemporaryFileManager.getTemporaryFile()`.

PR #6254, opened 2026-04-22
Fix DefaultExpressionVisitor not traversing into comparisons, logical ops, casts, and filters

joewiz:
## Summary

`DefaultExpressionVisitor` silently skips child expressions inside 5 expression types. Any code that relies on the visitor pattern to traverse a complete expression tree — LSP hover, code analysis, optimization — will miss function calls, variable references, and other expressions nested inside these constructs.

## The bug

`BasicExpressionVisitor` defines no-op implementations for several `visit*` methods. `DefaultExpressionVisitor` overrides most of them to traverse children (e.g., `visitUnionExpr` traverses `left` and `right`), but misses these five:

| Method | Expression | What's skipped |
|--------|-----------|----------------|
| `visitGeneralComparison` | `=`, `!=`, `<`, `>`, `<=`, `>=` | Left and right operands |
| `visitAndExpr` | `and` | Left and right operands |
| `visitOrExpr` | `or` | Left and right operands |
| `visitCastExpr` | `cast as` | The expression being cast |
| `visitFilteredExpr` | `expr[predicate]` | The base expression and predicates |

## Reproducer

Compile any XQuery expression containing a function call inside a comparison, then traverse it with `DefaultExpressionVisitor`:

```java
String xquery = "upper-case('test') = 'TEST'";

// Parse, compile, analyze
PathExpr path = new PathExpr(context);
XQueryTreeParser astParser = new XQueryTreeParser(context);
astParser.xpath(new XQueryParser(new XQueryLexer(context,
    new StringReader(xquery))).xpath(), path);
path.analyze(new AnalyzeContextInfo());

// Traverse — should find upper-case()
FunctionNameCollector collector = new FunctionNameCollector();
path.accept(collector);

// EXPECTED: ["upper-case"]
// ACTUAL:   [] — the visitor never enters the GeneralComparison
```

The same pattern fails for `and`, `or`, `cast as`, and filter predicates.

## Fix

Override all five methods in `DefaultExpressionVisitor` to traverse their children, matching the pattern already used for `visitUnionExpr` and `visitIntersectionExpr`:

```java
@Override
public void visitGeneralComparison(GeneralComparison comparison) {
    comparison.getLeft().accept(this);
    comparison.getRight().accept(this);
}
```

## Verified

- Without fix: **7/7 tests fail** on `develop`
- With fix: **7/7 tests pass**

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/DefaultExpressionVisitor.java
+++ b/exist-core/src/main/java/org/exist/xquery/DefaultExpressionVisitor.java
@@ -162,6 +162,37 @@ public class DefaultExpressionVisitor extends BasicExpressionVisitor {
     }
 
     @Override
+    public void visitGeneralComparison(GeneralComparison comparison) {
+        comparison.getLeft().accept(this);
+        comparison.getRight().accept(this);
+    }
+
+    @Override
+    public void visitAndExpr(OpAnd and) {
+        and.getLeft().accept(this);
+        and.getRight().accept(this);
+    }
+
+    @Override
+    public void visitOrExpr(OpOr or) {
+        or.getLeft().accept(this);
+        or.getRight().accept(this);
+    }
+
+    @Override
+    public void visitCastExpr(CastExpression expression) {
+        expression.getInnerExpression().accept(this);
+    }
+
+    @Override
+    public void visitFilteredExpr(FilteredExpression filtered) {
+        filtered.getExpression().accept(this);
+        for (final Predicate pred : filtered.getPredicates()) {
+            pred.accept(this);
+        }
+    }
+
+    @Override
     public void visitUnionExpr(Union union) {
         union.left.accept(this);
         union.right.accept(this);
--- a/exist-core/src/main/java/org/exist/xquery/DefaultExpressionVisitor.java
+++ b/exist-core/src/main/java/org/exist/xquery/DefaultExpressionVisitor.java
@@ -186,8 +186,9 @@ public class DefaultExpressionVisitor extends BasicExpressionVisitor {
 
     @Override
     public void visitFilteredExpr(FilteredExpression filtered) {
-        if (filtered.getExpression() != null) {
-            filtered.getExpression().accept(this);
+        @Nullable final Expression expr = filtered.getExpression();
+        if (expr != null) {
+            expr.accept(this);
         }
         for (final Predicate pred : filtered.getPredicates()) {
             pred.accept(this);
--- a/exist-core/src/main/java/org/exist/xquery/DefaultExpressionVisitor.java
+++ b/exist-core/src/main/java/org/exist/xquery/DefaultExpressionVisitor.java
@@ -186,7 +186,9 @@ public class DefaultExpressionVisitor extends BasicExpressionVisitor {
 
     @Override
     public void visitFilteredExpr(FilteredExpression filtered) {
-        filtered.getExpression().accept(this);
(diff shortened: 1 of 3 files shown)


reinhapa:
There is a failing test:
```
[exist-expathrepo] [INFO] Running xquery.modules.expathrepo.ExpathRepoTests
Error: expathrepo] [ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.760 s <<< FAILURE! -- in xquery.modules.expathrepo.ExpathRepoTests
Error: expathrepo] [ERROR] xquery.modules.expathrepo.ExpathRepoTests -- Time elapsed: 0.760 s <<< ERROR!
java.lang.NullPointerException: Cannot invoke "org.exist.xquery.Expression.accept(org.exist.xquery.ExpressionVisitor)" because the return value of "org.exist.xquery.FilteredExpression.getExpression()" is null
	at org.exist.xquery.DefaultExpressionVisitor.visitFilteredExpr(DefaultExpressionVisitor.java:189)
	at org.exist.xquery.FilteredExpression.accept(FilteredExpression.java:224)
	at org.exist.xquery.DefaultExpressionVisitor.visitBuiltinFunction(DefaultExpressionVisitor.java:51)
```

joewiz:

Good catch, @reinhapa — `FilteredExpression.getExpression()` can return null (e.g., during the ExpathRepo test suite). Added a null guard in e19dcebbf1. The visitor now safely skips the base expression when it's null while still traversing the predicates.

exist-core/src/main/java/org/exist/xquery/DefaultExpressionVisitor.java
```diff
@@ -186,7 +186,9 @@ public void visitCastExpr(CastExpression expression) {
 
     @Override
     public void visitFilteredExpr(FilteredExpression filtered) {
-        filtered.getExpression().accept(this);
+        if (filtered.getExpression() != null) {
+            filtered.getExpression().accept(this);
```
reinhapa: Maybe it make sense to extract `filtered.getExpression()` to a local variable.

joewiz:

@reinhapa Good suggestion — extracted to a `@Nullable` local variable in 3f6e3efd9f. Avoids calling the getter twice and makes the null check clearer.

line-o:
We still have failiing tests

> [ERROR]   XQueryTest.predicatePositionLast_1665213:2043 » ClassCast class org.exist.xquery.FilteredExpression cannot be cast to class org.exist.xquery.LocationStep (org.exist.xquery.FilteredExpression and org.exist.xquery.LocationStep are in unnamed module of loader 'app')

joewiz:

Good catch, @line-o — the `ClassCastException` in `predicatePositionLast_1665213` was caused by the Optimizer's `visitAndExpr` assuming `predicate.getParent()` is always a `LocationStep`. Our visitor change now traverses into `FilteredExpression`'s predicates (which it didn't before), so the Optimizer encounters predicates whose parent is a `FilteredExpression`, not a `LocationStep`.

Fixed in 08266fcc59: added an `instanceof LocationStep` guard before the cast. The predicate-splitting optimization is correctly skipped when the predicate belongs to a non-LocationStep parent.

104 tests pass (including `predicatePositionLast_1665213` and all 5 visitor traversal tests).

line-o (approved):
Thanks for this important fix.

reinhapa (approved):

PR #6256, opened 2026-04-24
[feature] Improve XQTS QT4 compliance: codepoints, error codes

joewiz:
## Summary

- `fn:codepoints-to-string`: Accept C0 control characters (U+0001-U+001F) per XML 1.1 / XQ 4.0 while still rejecting Unicode noncharacters. Fixes 87+ fn-graphemes tests that use codepoint 1.
- `ErrorCodes`: Add FODT0005 (missing date/time component) and FODT0006 (invalid date/time component value).

## Test plan

- [ ] Run `fn-codepoints-to-string` XQTS test set — should improve by 87+ tests
- [ ] Verify FODT0005/FODT0006 error codes compile cleanly

Note: regex `\b`/`\B` word boundary and `fn:build-dateTime` fixes from the original next-v3 commit depend on Saxon 12 / v2/ branches and will land with those PRs.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ErrorCodes.java
+++ b/exist-core/src/main/java/org/exist/xquery/ErrorCodes.java
@@ -161,6 +161,10 @@ public class ErrorCodes {
     public static final ErrorCode FODT0001 = new W3CErrorCode("FODT0001", "Overflow/underflow in date/time operation.");
     public static final ErrorCode FODT0002 = new W3CErrorCode("FODT0002", "Overflow/underflow in duration operation.");
     public static final ErrorCode FODT0003 = new W3CErrorCode("FODT0003", "Invalid timezone value.");
+    // --- XQuery 4.0 fn:build-dateTime error codes ---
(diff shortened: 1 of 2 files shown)


reinhapa (approved):

line-o:
@joewiz The PR title mentiosn QT4 so I assume this PR should wait until we get to the other PRs with that feature set. Correct?
I have trouble understanding why the errorcodes are part of this commit. I believe they should be added to the respective PR that adds XQuery 4 functions.

joewiz:

@line-o — good point. This PR was cherry-picked from next-v3 and only the develop-compatible parts were included (codepoints fix + error codes). You're right that the error codes are premature — they belong with the XQ4 function PRs that define the functions using those codes.

I'll split this: keep the `fn:codepoints-to-string` fix (which is a 3.1 compliance fix) and move the error codes to their respective v2/ PRs. Or if you prefer, I can close this PR and let the fixes land with the v2/ branches.

PR #6259, opened 2026-04-25
[feature] Support map atomization in XQuery 4.0

joewiz:
## Summary

- Add version-gated map atomization for XQuery 4.0 (maps are atomizable in XQ4 but not in XQ 3.1)
- Resolves ~103 XQTS QT4 test failures caused by `err:FOTY0013` when atomizing map results
- XQ 3.1 behavior unchanged: maps still raise FOTY0013

### `AbstractMapType.java`
- `isXq4Atomizable()` — version gate: returns `true` when `context.getXQueryVersion() >= 40`
- `atomizeValues()` — iterates all map entries, atomizes each value, returns flat `Sequence`
- `atomize()` override — for XQ4 single-entry maps returns the atomized value; for XQ 3.1 throws FOTY0013

### `Atomize.java`
- Updated static `atomize(Sequence)` to detect XQ4 maps and expand them via `atomizeValues()` before individual item atomization — same pattern as `ArrayType.flatten()` for arrays

### `ConcatExpr.java`
- The FOTY0013 guard for function items now has a carve-out: XQ4 maps are atomized and their string values concatenated

## Spec References

- [XQuery 4.0 Data Model — Maps](https://qt4cg.org/specifications/xpath-datamodel-40/xpath-datamodel-40.html#dt-map): Maps are no longer function items; they have a typed value (the concatenation of the typed values of their entries' values)
- [XQuery 3.1 — FOTY0013](https://www.w3.org/TR/xquery-31/#dt-type-error): Raised when atomizing a function item other than an array

## Test Plan

- [x] Build compiles on develop (no XQ4 parser = dormant code path, safe to merge independently)
- [x] XQSuite tests (1218): zero new regressions (7 pre-existing HOF/xml-to-json failures)
- [x] XPathQueryTest (150): all pass
- [x] Full exist-core test suite (7835): zero new regressions
- [x] XQ 3.1 `maps.xqm:no-atomization` still expects error (maps not atomizable in 3.1)
- [x] XQ 3.1 `cast-constructor.xqm:atomize-map-fails` still expects FOTY0013
- [ ] XQTS QT4 runner verification (requires `next-v3` integration with XQ4 parser)

**Note:** On develop, the `isXq4Atomizable()` check always returns `false` since the XQ4 parser support (`v2/xq4-parser-extensions`) is not yet merged. The code path activates when integrated into `next-v3`.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/Atomize.java
+++ b/exist-core/src/main/java/org/exist/xquery/Atomize.java
@@ -23,6 +23,7 @@ package org.exist.xquery;
 
 import org.exist.dom.persistent.DocumentSet;
 import org.exist.xquery.functions.array.ArrayType;
+import org.exist.xquery.functions.map.AbstractMapType;
 import org.exist.xquery.util.ExpressionDumper;
 import org.exist.xquery.value.Item;
 import org.exist.xquery.value.Sequence;
@@ -75,15 +76,25 @@ public class Atomize extends AbstractExpression {
         if (input.isEmpty())
             {return Sequence.EMPTY_SEQUENCE;}
         input = ArrayType.flatten(input);
-        if (input.hasOne()) {return
-            input.itemAt(0).atomize();
+        if (input.hasOne()) {
+            final Item single = input.itemAt(0);
+            // XQ4: maps are atomizable — expand to their values before atomizing
+            if (single instanceof AbstractMapType && ((AbstractMapType) single).isXq4Atomizable()) {
+                return ((AbstractMapType) single).atomizeValues();
+            }
+            return single.atomize();
         }
 
         Item next;
         final ValueSequence result = new ValueSequence();
         for(final SequenceIterator i = input.iterate(); i.hasNext(); ) {
             next = i.nextItem();
-            result.add(next.atomize());
+            // XQ4: maps are atomizable — expand to their values before atomizing
+            if (next instanceof AbstractMapType && ((AbstractMapType) next).isXq4Atomizable()) {
+                result.addAll(((AbstractMapType) next).atomizeValues());
+            } else {
+                result.add(next.atomize());
+            }
         }
         return result;
     }
--- a/exist-core/src/main/java/org/exist/xquery/Atomize.java
+++ b/exist-core/src/main/java/org/exist/xquery/Atomize.java
@@ -79,8 +79,8 @@ public class Atomize extends AbstractExpression {
         if (input.hasOne()) {
             final Item single = input.itemAt(0);
             // XQ4: maps are atomizable — expand to their values before atomizing
-            if (single instanceof AbstractMapType && ((AbstractMapType) single).isXq4Atomizable()) {
-                return ((AbstractMapType) single).atomizeValues();
+            if (single instanceof AbstractMapType mapType && mapType.isXq4Atomizable()) {
+                return mapType.atomizeValues();
             }
             return single.atomize();
         }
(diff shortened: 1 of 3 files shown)


exist-core/src/main/java/org/exist/xquery/Atomize.java
```diff
@@ -75,15 +76,25 @@ public static Sequence atomize(Sequence input) throws XPathException {
...
             {return Sequence.EMPTY_SEQUENCE;}
         input = ArrayType.flatten(input);
-        if (input.hasOne()) {return
-            input.itemAt(0).atomize();
+        if (input.hasOne()) {
+            final Item single = input.itemAt(0);
+            // XQ4: maps are atomizable — expand to their values before atomizing
+            if (single instanceof AbstractMapType && ((AbstractMapType) single).isXq4Atomizable()) {
+                return ((AbstractMapType) single).atomizeValues();
+            }
```
reinhapa: Use direct cast using local variable here:
```suggestion
            if (single instanceof AbstractMapType mapType && mapType.isXq4Atomizable()) {
                return mapType.atomizeValues();
            }
```

exist-core/src/main/java/org/exist/xquery/Atomize.java
```diff
@@ -75,15 +76,25 @@ public static Sequence atomize(Sequence input) throws XPathException {
...
         }
 
         Item next;
         final ValueSequence result = new ValueSequence();
         for(final SequenceIterator i = input.iterate(); i.hasNext(); ) {
             next = i.nextItem();
-            result.add(next.atomize());
+            // XQ4: maps are atomizable — expand to their values before atomizing
+            if (next instanceof AbstractMapType && ((AbstractMapType) next).isXq4Atomizable()) {
+                result.addAll(((AbstractMapType) next).atomizeValues());
```
reinhapa: Same as above
```suggestion
            if (next instanceof AbstractMapType mapType && mapType.isXq4Atomizable()) {
                result.addAll(mapType.atomizeValues());
```

exist-core/src/main/java/org/exist/xquery/ConcatExpr.java
```diff
@@ -65,8 +66,17 @@ public Sequence eval(Sequence contextSequence, Item contextItem)
             final Sequence seq = step.eval(contextSequence, contextItem);
             for (final SequenceIterator i = seq.iterate(); i.hasNext(); ) {
                 final Item item = i.nextItem();
-                if (Type.subTypeOf(item.getType(), Type.FUNCTION))
-                    {throw new XPathException(this, ErrorCodes.FOTY0013, "Got a function item as operand in string concatenation");}
+                if (Type.subTypeOf(item.getType(), Type.FUNCTION)) {
+                    // XQ4: maps are no longer function items for atomization purposes
+                    if (item instanceof AbstractMapType && ((AbstractMapType) item).isXq4Atomizable()) {
+                        final Sequence atomized = ((AbstractMapType) item).atomizeValues();
```
reinhapa: Same here
```suggestion
                    if (item instanceof AbstractMapType mapType && mapType.isXq4Atomizable()) {
                        final Sequence atomized = mapType.atomizeValues();
```

reinhapa (changes requested):

joewiz:

Thanks @reinhapa — addressed in b7c258cf94:
- Converted all 3 `instanceof` + cast patterns to Java 16+ pattern-matching `instanceof` in `Atomize.java` (2 locations) and `ConcatExpr.java` (1 location)

PR #6260, opened 2026-04-25
[feature] XQuery 4.0 type coercion rules

joewiz:
## Summary

- Implement XQuery 4.0 coercion rules from spec §3.4.1, version-gated behind `xquery version "4.0"`
- Add implicit casting: `xs:string → xs:anyURI`, `xs:hexBinary ↔ xs:base64Binary`, bidirectional numeric
- Add relabeling for derived atomic types (e.g., `xs:integer` 3 accepted as `xs:positiveInteger`)
- Accept `xquery version "4.0"` in the ANTLR tree walker

## What Changed

**`XQueryTree.g`** — Accept `xquery version "4.0"`, setting context version to 40.

**`DynamicTypeCheck.java`** — New `xq4ImplicitCast()` method handles runtime coercion per spec §3.4.1:
- Item 4 (Implicit Casting table): `xs:string→xs:anyURI`, `xs:hexBinary↔xs:base64Binary`, any numeric → any numeric
- Item 6 (Relabeling): derived type acceptance when datum is in value space
- `isXQ4ImplicitCast()` and `isXQ4Relabeling()` are `static` for reuse by the static check

**`Function.java`** — Bypass static type check for XQ4 implicit casts and relabeling (line 305).

**`UntypedValueCheck.java`** — Version-gate the `xs:integer → xs:positiveInteger` rejection; allow in XQ4 per §3.4.1 item 6.

**`xq4-type-promotion.xql`** — 8 XQSuite tests (XQ 4.0 module) covering all coercion rules.

**`xq4-type-promotion-gating.xql`** — 1 version-gating test (XQ 3.1 module) confirming `xs:string→xs:anyURI` is correctly rejected.

## Spec References

- [XQuery 4.0 §3.4.1 Item Coercion Rules](https://qt4cg.org/specifications/xquery-40/#id-item-coercion-rules) — Implicit Casting table (item 4) and Relabeling (item 6)
- QT4CG issues: #117 (relabeling), #980 (bidirectional numeric), #130/#480 (binary coercion)

## Design Decisions

- **No changes to `Type.java`** — the type hierarchy is correct; coercion operates at the function argument level via `DynamicTypeCheck`, exactly as the spec intends
- **Version-gated** — all new rules only activate when `xquery version "4.0"` is declared
- The 76 date/time XQTS failures (`xs:time is not a sub-type of xs:dateTime`) are NOT type promotion — they require function signature changes for `fn:*-from-dateTime` (QT4 PR1481), tracked separately

## Test Plan

- [x] 9 XQSuite tests pass (string→anyURI, binary coercion, numeric coercion, relabeling, version gating)
- [x] Full exist-core test suite: 6601 tests, 0 failures, 0 regressions
- [ ] XQTS QT4 run to verify `function-call-promotion-*` tests pass

### DIFF
--- a/exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g
+++ b/exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g
@@ -451,14 +451,16 @@ throws PermissionDeniedException, EXistException, XPathException
             v:VERSION_DECL
             {
                 final String version = v.getText();
-                if (version.equals("3.1")) {
+                if (version.equals("4.0")) {
+                    context.setXQueryVersion(40);
+                } else if (version.equals("3.1")) {
                     context.setXQueryVersion(31);
                 } else if (version.equals("3.0")) {
                     context.setXQueryVersion(30);
                 } else if (version.equals("1.0")) {
                     context.setXQueryVersion(10);
                 } else {
-                    throw new XPathException(v, ErrorCodes.XQST0031, "Wrong XQuery version: require 1.0, 3.0 or 3.1");
+                    throw new XPathException(v, ErrorCodes.XQST0031, "Wrong XQuery version: require 1.0, 3.0, 3.1, or 4.0");
                 }
             }
             ( enc:STRING_LITERAL )?
--- a/exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g
+++ b/exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g
@@ -451,17 +451,13 @@ throws PermissionDeniedException, EXistException, XPathException
             v:VERSION_DECL
             {
                 final String version = v.getText();
-                if (version.equals("4.0")) {
-                    context.setXQueryVersion(40);
-                } else if (version.equals("3.1")) {
-                    context.setXQueryVersion(31);
-                } else if (version.equals("3.0")) {
-                    context.setXQueryVersion(30);
-                } else if (version.equals("1.0")) {
-                    context.setXQueryVersion(10);
-                } else {
-                    throw new XPathException(v, ErrorCodes.XQST0031, "Wrong XQuery version: require 1.0, 3.0, 3.1, or 4.0");
-                }
+                context.setXQueryVersion(switch (version) {
+                    case "4.0" -> 40;
+                    case "3.1" -> 31;
+                    case "3.0" -> 30;
+                    case "1.0" -> 10;
+                    default -> throw new XPathException(v, ErrorCodes.XQST0031, "Wrong XQuery version: require 1.0, 3.0, 3.1, or 4.0");
+                });
             }
             ( enc:STRING_LITERAL )?
             {
--- a/exist-core/src/main/java/org/exist/xquery/UntypedValueCheck.java
+++ b/exist-core/src/main/java/org/exist/xquery/UntypedValueCheck.java
@@ -114,7 +114,10 @@ public class UntypedValueCheck extends AbstractExpression {
                 if (Type.subTypeOf(item.getType(), requiredType)) {
                     return item;
                 }
-                if (item.getType() == Type.INTEGER && requiredType == Type.POSITIVE_INTEGER) {
+                // In XQuery 3.1, reject integer→positiveInteger conversion.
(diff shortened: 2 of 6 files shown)


exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g
```diff
@@ -267,14 +267,16 @@ throws PermissionDeniedException, EXistException, XPathException
...
+                } else if (version.equals("3.1")) {
                     context.setXQueryVersion(31);
                 } else if (version.equals("3.0")) {
                     context.setXQueryVersion(30);
                 } else if (version.equals("1.0")) {
                     context.setXQueryVersion(10);
                 } else {
-                    throw new XPathException(v, ErrorCodes.XQST0031, "Wrong XQuery version: require 1.0, 3.0 or 3.1");
+                    throw new XPathException(v, ErrorCodes.XQST0031, "Wrong XQuery version: require 1.0, 3.0, 3.1, or 4.0");
                 }
```
reinhapa: maybe convert this to a switch-as-expression syntax?

joewiz:

Thanks @reinhapa — addressed in 826ef2d682:
- Converted the XQuery version if-else chain in `XQueryTree.g` to a switch-as-expression

exist-core/src/main/java/org/exist/xquery/UntypedValueCheck.java
```diff
@@ -114,7 +114,10 @@ private Item convert(Item item) throws XPathException {
                 if (Type.subTypeOf(item.getType(), requiredType)) {
                     return item;
                 }
-                if (item.getType() == Type.INTEGER && requiredType == Type.POSITIVE_INTEGER) {
+                // In XQuery 3.1, reject integer→positiveInteger conversion.
```
dizzzz, line 117: is this a regression?

dizzzz: nope

dizzzz (approved):
LGTM

dizzzz:
@reinhapa any new thoughts?

reinhapa (approved):

joewiz:

Picked this one back up. Two updates, and a scope question for you all.

**Rebased and green.** The branch was ~930 commits behind, and the earlier Windows CI failure turned out to be purely the 45-minute integration-test timeout (no test failures) — which is long gone on current develop (that job now finishes in ~24 min). Rebased onto develop (clean, no conflicts), the full suite passes, and CI is green.

**A scope finding worth your input.** I ran the actual QT4 `function-call-promotion` tests against the branch. **5 of 9 pass; 4 fail:**

- **Pass:** `403`, `404`, `405` (string ↔ anyURI) and `408`, `409` (binary promotion of a function *result*).
- **Fail (`XPTY0004`):** `401`, `402`, `406`, `407`.

The four failures aren't in the argument coercion this PR implements — `408`/`409` confirm hexBinary/base64Binary promotion works at the call boundary. They fail because each test body then *compares* the coerced value, e.g. `$z = xs:base64Binary(...)` with `$z` now an `xs:hexBinary`. Making that comparison succeed requires applying §3.4.1 implicit casting to **value-comparison operands** as well — a distinct slice of the spec that this PR (scoped to function-call/return coercion) does not touch. In XQuery 3.1 that comparison is correctly an `XPTY0004`, and everything here is gated behind `xquery version "4.0"`, so 3.1 behavior is unchanged either way.

**The question:** how would you like to handle the comparison-coercion gap?

1. Treat this PR as a correct, self-contained increment (function-call/return coercion), merge it when XQ4 work is in scope, and do **value-comparison coercion as a follow-up PR** — with the four QT4 cases tracked against that follow-up.
2. Or expand *this* PR to also cover comparison-operand coercion, aiming for the full 9/9 before it merges.

We'd lean toward (1) — it keeps a reviewable, spec-correct increment moving and separates a genuinely different mechanism — but this is squarely your call, especially given XQ4 is parked behind 7.0 for now. Happy to open a tracking issue for the comparison-coercion follow-up if that's useful.

dizzzz:
I agree with (1) "as is"

duncdrum:
same here 3.1 as is

PR #6264, opened 2026-04-27
[optimize] XQuery local-variable resolution and dead cache removal

joewiz:
## Summary

Two independent XQuery engine optimizations with no regressions across XQuery3Tests (1227 tests, identical baseline failures), QT4 fn-* / HOF / XQUF sets, and app-XMark.

### 1. `XQueryContext.resolveLocalVariable` — O(N) → O(1)

The previous implementation walked a linked list of `LocalVariable`s backward from `lastVar`, stopping at the current `contextStack` mark. For a function body of N parameters analyzing N variable references this is O(N²). For the 123,456-arity reference in QT4 `misc-HigherOrderFunctions/hof-025`, this lookup alone cost 22.6s.

A `{QName → LocalVariable}` HashMap now points to the most recently declared binding for each name. Each `LocalVariable` carries:

- `prevSameName` — the binding shadowed by this declaration; used by `popLocalVariables` to unwind the chain when the variable goes out of scope.
- `markedUnder` — the `contextStack` frame this variable was declared under. A binding is in scope iff `markedUnder == contextStack.peek()`.

The HashMap is shared by reference with context copies (`copyFields`, `updateContext`) alongside `lastVar` and `contextStack`, so module and forked contexts see the same in-scope variables.

### 2. `UserDefinedFunction.addVariable` — O(N) → O(1)

The duplicate-parameter check used `parameters.contains`, an O(N) ArrayList scan, making parameter-list construction O(N²). For hof-025 this loop alone cost ~8s. A HashSet<QName> alongside the ordered List gives O(1) duplicate detection while preserving insertion order.

### 3. `LocationStep`: remove dead `CachedResult` write

The `CachedResult` write inside `LocationStep.eval` was a leftover from a removed code path. Its result was never read. Removed.

## XQTS impact (QT4 vs `qt4-rd-2026-04-26` baseline)

| Test set | Before | After | Speedup |
|----------|--------|-------|---------|
| `misc-HigherOrderFunctions` | 36.81s | 1.49s | 24x (-4 failures) |
| `hof-025` alone | 33.81s | 0.25s | 135x |
| `app-XMark` | 29.64s | 13.13s | 2.3x (FLWOR var-resolution) |
| `fn-fold-left` | 6.23s | 3.73s | 1.7x |
| `fn-fold-right` | 3.45s | 1.09s | 3.2x |
| `upd-*` (16 sets) | — | — | 1.5x–4.7x, no regressions |

XQuery3Tests JUnit: 1227 tests, identical failure set.

## Note on a third optimization

A companion optimization for `fn:divide-decimals` exists in our integration branches but depends on `FnDivideDecimals.java`, which lives on `v2/xq4-core-functions`. It will be reapplied here once that branch merges.

## Test plan

- [x] `mvn install -pl exist-core -am -DskipTests` passes
- [x] `mvn test -pl exist-core -Dtest=xquery.xquery3.XQuery3Tests` — 1227 tests, baseline failures only
- [ ] CI green

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -2501,22 +2501,24 @@ public class XQueryContext implements BinaryValueManager, Context {
      * @param resultSeq the result sequence
      */
     public void popLocalVariables(@Nullable final LocalVariable var, @Nullable final Sequence resultSeq) {
-        if (var != null) {
-            // Restore localVariableLookup bindings in REVERSE-of-declaration order
-            // so each var's prevSameName chain unwinds correctly. (Walking forward
-            // would leave the map pointing at a popped variable when multiple vars
-            // share a name within the popped scope.)
-            for (LocalVariable cursor = lastVar; cursor != null && cursor != var; cursor = cursor.before) {
-                if (localVariableLookup.get(cursor.getQName()) == cursor) {
-                    if (cursor.prevSameName != null) {
-                        localVariableLookup.put(cursor.getQName(), cursor.prevSameName);
-                    } else {
-                        localVariableLookup.remove(cursor.getQName());
-                    }
+        // Restore localVariableLookup bindings in REVERSE-of-declaration order
+        // so each var's prevSameName chain unwinds correctly. (Walking forward
+        // would leave the map pointing at a popped variable when multiple vars
+        // share a name within the popped scope.) Runs whether or not {@code var}
+        // is null: when {@code var == null}, every binding from {@code lastVar}
+        // back to the start is being popped, so the whole chain must unwind.
+        for (LocalVariable cursor = lastVar; cursor != null && cursor != var; cursor = cursor.before) {
+            if (localVariableLookup.get(cursor.getQName()) == cursor) {
+                if (cursor.prevSameName != null) {
+                    localVariableLookup.put(cursor.getQName(), cursor.prevSameName);
+                } else {
+                    localVariableLookup.remove(cursor.getQName());
                 }
-                cursor.prevSameName = null;
             }
+            cursor.prevSameName = null;
+        }
 
+        if (var != null) {
             // clear all variables registered after var. they should be out of scope.
             LocalVariable outOfScope = var.after;
             while (outOfScope != null) {
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -188,11 +188,13 @@ public class XQueryContext implements BinaryValueManager, Context {
     // The last element in the linked list of local in-scope variables
     private LocalVariable lastVar = null;
 
-    // O(1) lookup table from QName to the most-recently-declared LocalVariable
-    // with that name. Maintained alongside the {@link #lastVar} linked list:
-    // declareVariableBinding adds, popLocalVariables restores the prevSameName
-    // chain. Visibility is enforced in resolveLocalVariable by comparing
-    // {@link LocalVariable#markedUnder} to {@code contextStack.peek()}.
+    /**
+     * O(1) lookup table from QName to the most-recently-declared LocalVariable
+     * with that name. Maintained alongside the {@link #lastVar} linked list:
+     * declareVariableBinding adds, popLocalVariables restores the prevSameName
+     * chain. Visibility is enforced in resolveLocalVariable by comparing
+     * {@link LocalVariable#markedUnder} to {@code contextStack.peek()}.
+     */
     // Shared by reference with copies of this context (see copyFields,
     // updateContext) just like {@link #contextStack} and {@link #lastVar}.
     private Map<QName, LocalVariable> localVariableLookup = new HashMap<>();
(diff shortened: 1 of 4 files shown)


duncdrum:
```
Error: core] [ERROR]   XQueryTest.precedence:641 XQuery: xquery version "1.0";
declare namespace blah="blah";
declare variable $blah:param := "value-1";
let $blah:param := "value-2"
(:: FLWOR expressions have a higher precedence than the comma operator ::)
return $blah:param, $blah:param  expected:<value-[1]> but was:<value-[2]>
[exist-core] [INFO] 
Error: core] [ERROR] Tests run: 6592, Failures: 1, Errors: 0, Skipped: 106
```

joewiz:

@duncdrum Thanks for catching this. The HashMap variable resolution has a scoping bug with namespaced variables — investigating now. The optimization works correctly on the `next-v3` integration branch (where it was originally developed and tested), but the cherry-pick to `develop` may have a different interaction with the variable resolution path. Will push a fix shortly.

joewiz:

@duncdrum Update on investigation: confirmed the HashMap variable resolution has a scoping edge case with namespaced global-vs-local variable shadowing on `develop`. The same optimization works correctly on `next-v3` (which has XQ4 expression changes that affect scoping).

The specific issue: when `let $blah:param := "value-2" return $blah:param, $blah:param` is evaluated, the second `$blah:param` (after the comma, outside the FLWOR scope) should resolve to the global `declare variable $blah:param := "value-1"`. The HashMap fails to restore the correct binding after the let scope pops.

Working on a fix — likely a fallback to the linked-list walk when the HashMap visibility check produces different results than the original path. Will push shortly.

joewiz:

@duncdrum Fixed in 28c3eb1491. Root cause: when `markLocalVariables(false)` returns null (top-level let with no prior locals), the HashMap unwind loop in `popLocalVariables` was inside the `if (var != null)` guard and never ran. The popped local stayed in the lookup map, and the visibility check passed spuriously because both `markedUnder` and `contextStack.peek()` were null.

Fix: moved the unwind loop outside the null guard. The `cursor != var` termination condition handles both cases correctly.

`XQueryTest.precedence` now passes. Full test suite: 1,233 tests, 0 failures, 0 errors. HOF O(1) fast path preserved.

duncdrum (approved):

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -2024,16 +2038,19 @@ Variable resolveGlobalVariable(final QName qname) {
...
-                return var;
-            }
+        // O(1) fast path. The linked-list walk previously here is O(N) per
+        // call and O(N²) when a body of N variables is analyzed.
+        final LocalVariable var = localVariableLookup.get(qname);
+        if (var == null) {
+            return null;
         }
-        return null;
+        // Visibility: var is visible iff it was declared under the current
```
line-o: typo: iff -> if

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -2484,6 +2501,23 @@ public void popLocalVariables(@Nullable final LocalVariable var) {
      * @param resultSeq the result sequence
      */
     public void popLocalVariables(@Nullable final LocalVariable var, @Nullable final Sequence resultSeq) {
+        // Restore localVariableLookup bindings in REVERSE-of-declaration order
```
line-o: I would prefer the longer paragraph of explanation to be moved to this methods JavaDoc and to be shortened a tad.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
```diff
@@ -188,6 +188,15 @@ public class XQueryContext implements BinaryValueManager, Context {
     // The last element in the linked list of local in-scope variables
     private LocalVariable lastVar = null;
 
+    // O(1) lookup table from QName to the most-recently-declared LocalVariable
```
line-o: I would prefer the comment to be a block comment instead of repeated line comments.

line-o (commented):
Very nice!
I found some minor issues which I would like to be addressed.

line-o:
@joewiz To me this is one of the really important PRs to land. We can change the comments later on.

joewiz:
@line-o Sorry for missing your comments here from a few days ago! Acting on them now...

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `49cb746041` — fixed.

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `49cb746041` — moved the explanation up into `popLocalVariables`'s JavaDoc as a `<p>` paragraph and shortened it (REVERSE-of-declaration-order rationale kept; the rest was redundant with the body).

exist-core/src/main/java/org/exist/xquery/XQueryContext.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `49cb746041` — converted to a JavaDoc block comment. The `{@link ...}` references are JavaDoc-tagged so the block form makes them resolve correctly.

Apologies for the delay on these — they slipped through our review-sweep. Thanks also for the today follow-up framing; ready for re-review whenever you have a moment.

line-o (approved):
There is still a _very_  minor issue with comments :)
Blockcomments followed by line comments.
I'll let it slip and will just fix it myself.

PR #6246, opened 2026-04-17
Dp mod refactor1

duncdrum:
nothing the hasn't been in  #6239

ensure rewrite  can run properly, some formatting fixes, and consistent ordering of poms.

PR #6276, opened 2026-04-29
[WIP] maven 4 migration

duncdrum:
update build tooling for maven 4. For now use mvnw for clean separation of v3 and v4. Once v4 is no longer RC we can consider its removal.

simplify releasish profiles, make installer generation off by default, and avoid long chains of disabled profiles for day to day operations. 

start with releases config for CI.

needs #6246


dizzzz:
I think it is just a bit too early for v7

PR #4578, opened 2022-09-28
Preserve XML Declaration

adamretter:
Requires Java 17 support to be merged first - https://github.com/eXist-db/exist/pull/4574 (and so at this time it contains all commits of that PR also).

This Pull Request causes an XML Declaration to be preserved if it is present. This means that it will be stored into the database if a document that is stored has an XML Declaration, and it will be retrieved at serialisation time.

This has been implemented and tested for the following APIs:
* HTTP RESTServer
* XML-RPC
* XML:DB Local and Remote
* WebDAV

As part of the document, the XML Declaration is now visible and modifiable from:
* eXist-db Java Admin Client
* Oxygen XML Editor
* eXide

This change necessitates changing the on-disk storage format of the `collections.dbx`, and so increments the file version (which is checked at startup).

* This required me to also release new versions of Apache Xerces 2 which supports the extended SAX API of Java 14+ - https://repo1.maven.org/maven2/org/exist-db/thirdparty/xerces/xercesImpl/2.12.2/ (see: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8230814)

PR #6277, opened 2026-04-29
[fix] remove omit-original-xml-declaration flag

line-o:
### Description:

The custom serialization flag `omit-original-xml-declaration` is removed. The standard `omit-xml-declaration` is the only parameter that decides whether an XML declaration is serialized:

- `omit-xml-declaration=no`: the persisted XML declaration of the document is used if present, otherwise a default one is generated
- `omit-xml-declaration=yes` (the configured default for REST, XML:DB, XML-RPC and WebDAV): no XML declaration, even if one is persisted

The flag added with #4578 took precedence over `omit-xml-declaration`, so persisted declarations were serialized even with `omit-xml-declaration=yes`. That broke the tests of node-exist and xst. The flag has only shipped in the 7.0.0 pre-releases (alpha, beta1–beta3).

As agreed on the community call of 18 May 2026 ([comment](https://github.com/eXist-db/exist/pull/6277#issuecomment-4480608690)):

- the flag is removed entirely
- backups set `omit-xml-declaration=no`, so every XML resource in a backup has an XML declaration, and therefore has one after a restore

> [!IMPORTANT]
> **Decision taken in this PR: WebDAV keeps `omit-xml-declaration=yes`**
>
> `webdav.properties` sets `omit-xml-declaration=yes`. Without the removed flag, **a persisted XML declaration is no longer returned by a WebDAV GET**, so it does not survive an edit round trip through a WebDAV client unless `omit-xml-declaration=no` is configured. We keep `yes` to stay consistent with REST and XML:DB, and with the 6.x behaviour. Switching the WebDAV default to `no` would round-trip persisted declarations, but would also add a declaration to every document served over WebDAV, including documents that never had one. `WebDavRoundTripTest.getXmlDeclDefault` was updated to match.

#### What changed

- **Flag removal**: `EXistOutputKeys`, `Serializer`, `Configuration`, `RESTServer` / `RESTServerParameter` (`_omit-original-xml-declaration`), `RpcConnection`, `LocalCollection`, `InteractiveClient`, `AbstractSerializer`, `MicroXmlWriter`, `conf.xml`, `client.properties`, both `webdav.properties`
- **`NativeSerializer`** always forwards the persisted declaration to the receiver (resolves the nested-`if` review comment)
- **`XMLWriter.writeDeclaration`** is the single decision point: a declaration is written if `omit-xml-declaration` is false or `standalone` is set explicitly (Serialization 3.1). Fields come from the serialization parameters, falling back to the persisted declaration, then to the defaults
- **`Backup`** also sets `omit-xml-declaration=no` on the collection it reads, so remote (XML-RPC) backups keep the persisted declaration instead of a generated default
- **`SystemExport`** now forwards the persisted declaration, which it previously dropped
- **`conf.xsd`**: attribute removed, version `2.6.0` → `2.7.0`. This is a minor bump, following the precedent of fcb5d34756; strictly, the schema policy classes an attribute removal as MAJOR. Happy to change that.
- Rebased onto current `develop`. The conflicts were in the decomposed `RESTServer.doGet` and the W3C-compliance rework of `XMLWriter.writeDeclaration`

### Reference:

Regression introduced with #4578 (855922a1)

### Type of tests:

JUnit:

- `XMLDBBackupTest` (local/remote): the backup always contains an XML declaration; a persisted declaration including `standalone` survives backup and restore
- `SystemExportImportTest`: a persisted declaration including `standalone` survives export and import
- `SerializationTest` (XML:DB local/remote), `RESTServiceTest`, `WebDavRoundTripTest`: default / `yes` / `no` against a document with a persisted declaration

Locally, all passing:

- exist-core: `XMLDBBackupTest`, `SystemExportImportTest`, `RestoreAppsTest`, `SerializationTest`, `RESTServiceTest`, `XmlRpcTest`, `ResourceTest`, `ContentAsDOMTest`, `DocTypeTest`, `PersistentDomTest`, `XMLUtilTest`, `ExtDocTest`, `org.exist.util.serializer.*Test`, `SchemaVersionSyncTest`, `xquery.xquery3.XQuery3Tests`: 1267 tests, 0 failures
- webdav: `WebDavRoundTripTest`: 6 tests, 0 failures
- schema governance (`mvn -N xml:transform@schema-governance`): passes

https://claude.ai/code/session_01ASiMAehNpncHLvKSS2oDnt

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/serializers/NativeSerializer.java
+++ b/exist-core/src/main/java/org/exist/storage/serializers/NativeSerializer.java
@@ -44,6 +44,7 @@ import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
+import javax.xml.transform.OutputKeys;
 import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
@@ -118,7 +119,7 @@ public class NativeSerializer extends Serializer {
         }
 
         if (doc.getXmlDeclaration() != null){
-            if ("no".equals(getProperty(EXistOutputKeys.OMIT_ORIGINAL_XML_DECLARATION, "no"))) {
+            if ("no".equals(getProperty(OutputKeys.OMIT_XML_DECLARATION, "yes"))) {
                 final XMLDeclarationImpl xmlDecl = doc.getXmlDeclaration();
                 receiver.declaration(xmlDecl.getVersion(), xmlDecl.getEncoding(), xmlDecl.getStandalone());
             }
--- a/exist-core/src/main/java/org/exist/client/InteractiveClient.java
+++ b/exist-core/src/main/java/org/exist/client/InteractiveClient.java
@@ -92,7 +92,6 @@ import se.softhouse.jargo.ArgumentException;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.time.ZoneOffset.UTC;
 import static javax.xml.transform.OutputKeys.OMIT_XML_DECLARATION;
-import static org.exist.storage.serializers.EXistOutputKeys.OMIT_ORIGINAL_XML_DECLARATION;
 import static org.exist.storage.serializers.EXistOutputKeys.OUTPUT_DOCTYPE;
 import static org.xmldb.api.base.ResourceType.XML_RESOURCE;
 
@@ -150,7 +149,6 @@ public class InteractiveClient {
         DEFAULT_PROPERTIES.setProperty(EDITOR, EDIT_CMD);
         DEFAULT_PROPERTIES.setProperty(INDENT, "true");
         DEFAULT_PROPERTIES.setProperty(OMIT_XML_DECLARATION, "no");
-        DEFAULT_PROPERTIES.setProperty(OMIT_ORIGINAL_XML_DECLARATION, "no");
         DEFAULT_PROPERTIES.setProperty(OUTPUT_DOCTYPE, "true");
         DEFAULT_PROPERTIES.setProperty(ENCODING, ENCODING_DEFAULT.name());
         DEFAULT_PROPERTIES.setProperty(COLORS, "false");
--- a/exist-core/src/main/java/org/exist/http/RESTServer.java
+++ b/exist-core/src/main/java/org/exist/http/RESTServer.java
@@ -128,7 +128,6 @@ public class RESTServer {
     public final static Properties defaultOutputKeysProperties = new Properties();
 
     static {
-        defaultOutputKeysProperties.setProperty(EXistOutputKeys.OMIT_ORIGINAL_XML_DECLARATION, "no");
         defaultOutputKeysProperties.setProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
         defaultOutputKeysProperties.setProperty(OutputKeys.INDENT, "yes");
         defaultOutputKeysProperties.setProperty(OutputKeys.MEDIA_TYPE,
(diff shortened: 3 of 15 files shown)


exist-core/src/main/java/org/exist/storage/serializers/NativeSerializer.java
```diff
@@ -118,7 +119,7 @@ protected void serializeToReceiver(DocumentImpl doc, boolean generateDocEvent) t
         }
 
         if (doc.getXmlDeclaration() != null){
-            if ("no".equals(getProperty(EXistOutputKeys.OMIT_ORIGINAL_XML_DECLARATION, "no"))) {
+            if ("no".equals(getProperty(OutputKeys.OMIT_XML_DECLARATION, "yes"))) {
```
reinhapa, line 122: Join this `if` statements as suggested by Codacy

reinhapa (changes requested):

duncdrum:
test failures 

```
Error: core] [ERROR] Tests run: 8, Failures: 4, Errors: 0, Skipped: 0, Time elapsed: 19.33 s <<< FAILURE! -- in org.exist.backup.XMLDBBackupTest
Error: core] [ERROR] org.exist.backup.XMLDBBackupTest.backupRestoreWithXmlDecl[local (classic)] -- Time elapsed: 0.116 s <<< FAILURE!
java.lang.AssertionError: Content should contain XML declaration
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.exist.backup.XMLDBBackupTest.backupRestoreWithXmlDecl(XMLDBBackupTest.java:173)
Error: core] [ERROR] org.exist.backup.XMLDBBackupTest.backupRestoreWithXmlDecl[remote (classic)] -- Time elapsed: 3.974 s <<< FAILURE!
java.lang.AssertionError: Content should contain XML declaration
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.exist.backup.XMLDBBackupTest.backupRestoreWithXmlDecl(XMLDBBackupTest.java:173)
Error: core] [ERROR] org.exist.backup.XMLDBBackupTest.backupRestoreWithXmlDecl[local (dedup)] -- Time elapsed: 0.102 s <<< FAILURE!
java.lang.AssertionError: Content should contain XML declaration
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.exist.backup.XMLDBBackupTest.backupRestoreWithXmlDecl(XMLDBBackupTest.java:173)
Error: core] [ERROR] org.exist.backup.XMLDBBackupTest.backupRestoreWithXmlDecl[remote (dedup)] -- Time elapsed: 3.518 s <<< FAILURE!
java.lang.AssertionError: Content should contain XML declaration
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.exist.backup.XMLDBBackupTest.backupRestoreWithXmlDecl(XMLDBBackupTest.java:173)
```

line-o:
@duncdrum yes this is expected until we come to a solution for this issue 

> The only remaining issue is how to handle stored declarations in backups. They should be part of the backup if - and only if - there is one stored.

I should have marked this as a draft perhaps.

line-o:
The consensus when discusssing this topic on the community call on 18th May was to 

- remove the custom serialization flag entirely
- when doing backups set "omit-xml-declaration" flag to false which will mean that all xml resources will have a xml declaration from then on

exist-core/src/main/java/org/exist/storage/serializers/NativeSerializer.java
line-o: Addressed in 85a7e4f7ed: with the flag removed there is no condition left here. `NativeSerializer` always forwards the persisted declaration, and `XMLWriter.writeDeclaration` alone decides based on `omit-xml-declaration` (or an explicit `standalone`), so the nested `if` is gone.

reinhapa (approved):

duncdrum:
@line-o [docs article on webdav](https://github.com/eXist-db/documentation/blob/master/src/main/xar-resources/data/webdav/webdav.xml) the section with `xml:id="serialisation"` seems like the natural place.

schema/conf.xsd
```diff
@@ -1057,16 +1057,9 @@
...
-                            </xs:annotation>
-                        </xs:attribute>
-                        <xs:attribute name="omit-original-xml-declaration" type="yes_no" default="no">
-                            <xs:annotation>
-                                <xs:documentation>
-                                    Should the original persisted XML Declaration for a document be serialized
-                                    if it is present?
+                                    Should the XML Declaration be omitted when serializing a document?
+                                    If not omitted, the persisted XML Declaration of the document is
+                                    used if present, otherwise a default XML Declaration is generated.
```
duncdrum: we should mention what the default is here ie. `1.0` and `utf-8` I guess

duncdrum (commented):
one nit pick

schema/conf.xsd
line-o: If it is that.

line-o: Confirmed, it is that: `XMLWriter` generates the default declaration with version `1.0` and encoding `UTF-8` (`DEFAULT_XML_VERSION` / `DEFAULT_XML_ENCODING`), without `standalone`. The `version`, `encoding` and `standalone` serialization parameters take precedence, also over a persisted declaration. Documented in conf.xsd and conf.xml in 9795e876a6.

duncdrum (approved):

PR #6283, opened 2026-04-30
[bugfix] Harden defaults against unauthenticated REST/XUpdate abuse

joewiz:
## Summary

Addresses six findings (3 Critical, 3 High) from a security surface
audit of eXist-db's shipped defaults and input handling. The threat
model is an unauthenticated attacker reachable on the HTTP port.

| ID | Severity | Issue | Fix |
|----|----------|-------|-----|
| C1 | Critical | `admin` account ships with empty password | Log SEVERE warning at startup if admin is reachable with empty password |
| C2 | Critical | Shipped web.xml allows anonymous XQuery / XUpdate submission via REST | Change defaults to `authenticated` |
| C3 | Critical | `SessionManager` ids are sequential ints + cache not subject-scoped | `SecureRandom` long ids, cache entries bound to subject |
| H1 | High | `XUpdateProcessor` parses XML without XXE hardening | Disallow DOCTYPE; disable external entities/parameter entities; secure-processing on; XInclude off |
| H3 | High | REST `hide-error-messages=false` by default; `path` echoed unencoded into error envelope | Default to `true`; `XMLUtil.encodeAttrMarkup(path)` |
| H4 | High | `JMXServlet.hasSecretToken` would accept an empty stored token paired with an empty request token | Refuse blank stored tokens |

### `exist-core/src/main/java/org/exist/security/internal/RealmImpl.java` (C1)
After `createAdminAndGuestIfNotExist`, attempt to authenticate `admin`
with the empty default password. On success, log a SEVERE warning
guiding the operator to set a real password. Failure is the expected
case (real password set) and is silent.

### `exist-jetty-config/src/main/resources/webapp/WEB-INF/web.xml` (C2, H3)
- ` <param-name>xquery-submission</param-name> `: `enabled` -> `authenticated`
- ` <param-name>xupdate-submission</param-name> `: `enabled` -> `authenticated`
- ` <param-name>hide-error-messages</param-name> `: `false` -> `true` (with comment explaining the dev-mode override)

### `exist-core/src/main/java/org/exist/http/SessionManager.java` (C3)
- Cache key changed from `Integer` to `Long`.
- Ids are drawn from `SecureRandom.nextLong() & Long.MAX_VALUE` and
  retried on collision.
- `QueryResult` now stores the `subjectId` of the caller that created
  the entry. `get()` and `release()` refuse access from any other
  subject (logged as a warning).
- `add(subjectId, query, sequence)` / `get(subjectId, query, sessionId)` / `release(subjectId, sessionId)` are the new signatures.

**Trade-off.** Combined with the move to `SecureRandom` long ids, binding cache entries to the creating subject closes the window where an attacker who guessed (now: brute-forced 2^63) a session id could read or release another caller's paginated result set. The side effect is that if the creating subject ends without calling `release` (browser closes, network drop, RESTXQ caller crashes), the entry can no longer be evicted by id from a different subject — only the 2-minute timeout reaper can reclaim it. That is intentional, and acceptable in the current REST surface because the entire pagination GET sequence is driven by the same caller. There is no admin-scoped "evict by id" code path today; if one is ever added, that code can pass `RealmImpl.SYSTEM_ACCOUNT_ID` (or a privilege check) into `release` to bypass the subject match.

### `exist-core/src/main/java/org/exist/http/RESTServer.java` (C3, H3)
- `Integer.parseInt` / `Integer.toString` replaced with `Long`-form
  for session ids. Header is set with `setHeader` rather than
  `setIntHeader`. Wire format remains a numeric string.
- `release` path now wraps `parseLong` in a `try/catch` and surfaces a
  `BadRequestException` instead of throwing `NumberFormatException` to
  the servlet container.
- All three `sessionManager.{add,get,release}` calls pass
  `broker.getCurrentSubject().getId()`.
- `writeXPathException` now writes `XMLUtil.encodeAttrMarkup(path)`
  rather than the raw path string.

### `exist-core/src/main/java/org/exist/xupdate/XUpdateProcessor.java` (H1)
The `DocumentBuilderFactory` used for XUpdate parsing now sets:

- `http://apache.org/xml/features/disallow-doctype-decl=true`
- `http://xml.org/sax/features/external-general-entities=false`
- `http://xml.org/sax/features/external-parameter-entities=false`
- `http://apache.org/xml/features/nonvalidating/load-external-dtd=false`
- `XMLConstants.FEATURE_SECURE_PROCESSING=true`
- `setXIncludeAware(false)`
- `setExpandEntityReferences(false)`

This matches the pattern already used in
`org.exist.util.Configuration.parse`. (`ACCESS_EXTERNAL_DTD` /
`ACCESS_EXTERNAL_SCHEMA` properties are intentionally not set on this
factory because the bundled Xerces build does not recognise them via
`setAttribute`; disallowing DOCTYPE outright covers the same surface.)

### `exist-core/src/main/java/org/exist/management/client/JMXServlet.java` (H4)
`hasSecretToken` now short-circuits to `false` when the stored token
is null, empty, or whitespace-only.

### Follow-up commit `59d406f1f0`

Two fixes after Duncan's review:

- **`exist-core/src/main/java/org/exist/http/RESTServer.java`** -- the `AUTHENTICATED_USERS_ONLY` gate (added in `ee7a1392293`, 2017) was checking `currentSubject.getId() == RealmImpl.GUEST_GROUP_ID`. By coincidence, `GUEST_GROUP_ID` and `ADMIN_ACCOUNT_ID` both equal `1048574` -- account ids and group ids live in different namespaces, but the constants collide. The intended check was "reject the **guest account**"; the actual effect was the inverse: admin received 403 and guest sailed through. The bug was invisible while the shipped default was `xquery-submission=enabled` (no gate runs at all). C2 of this PR flips the default to `authenticated`, which makes the gate reachable. Fix: compare against `GUEST_ACCOUNT_ID`.
- **`exist-docker/src/test/bats/01-connect-spec.bats`** -- the "logs are error free" test greps for the literal token `ERROR`. C1's SEVERE warning is emitted via `LOG.error(...)` and the shipped image starts with admin's empty default password, so the warning fires every startup. Excluded the specific `SECURITY WARNING` line from the grep.

## Out of scope (deferred)

- **H2 (SSRF allowlist for `fn:doc` / `fn:unparsed-text` / `fn:json-doc` / `httpclient:*`)** -- requires a config-schema design for the allow/deny list (RFC1918, link-local, multicast, loopback, plus operator overrides) and is a substantively bigger change than the rest of this PR.
- **M1-M7 (defense in depth / latent)** -- not addressed here.

## Test Plan

- [x] `mvn install -pl exist-core -am -DskipTests` -- compiles cleanly.
- [x] `mvn test -Ddependency-check.skip=true -Dlicense.skip=true -Ddocker=false` -- exist-core + every extension module builds and tests pass. (One unrelated reactor-level failure in `eXist-db W3C XQTS` from an inability to download `exist-xqts-runner_2.13:2.0.0-SNAPSHOT` from the Maven mirrors -- network/auth issue, not a code regression.)
- [x] Spot-checked the directly-affected suites in isolation:
  - `RESTServiceTest` (43 tests, exercises `SessionManager`)
  - `XUpdateTest`, `RemoveAppendTest`, `StressTest` (XUpdate XXE path)
  - `SecurityManagerTest`, `BackupRestoreSecurityPrincipalsTest`, `XqueryApiTest`, `UnixStylePermissionTest`
  - `XPathQueryTest` (148 tests, parameterised local + remote XML-RPC)
  - All pass with 0 failures, 0 errors.

### DIFF
--- a/exist-core/src/main/java/org/exist/http/SessionManager.java
+++ b/exist-core/src/main/java/org/exist/http/SessionManager.java
@@ -117,7 +117,9 @@ public class SessionManager {
 
     /**
      * Release a cached query result. Only the original subject may
-     * release the entry.
+     * release the entry. Privileged callers (DBAs) should use
+     * {@link #releaseAny(long)} to bypass the subject-match check, e.g.
+     * to reclaim entries whose creators ended without releasing them.
      */
     public void release(final int subjectId, final long sessionId) {
         if (sessionId < 0) {
@@ -133,5 +135,36 @@ public class SessionManager {
             return;
         }
         cache.invalidate(sessionId);
+    }
+
+    /**
+     * Privileged release of a single cache entry, bypassing the
+     * subject-match check. Intended for DBA-driven cleanup of entries
+     * whose creating subject ended without calling {@link #release}
+     * (e.g. browser closed, network drop, RESTXQ caller crashed).
+     * Callers MUST verify DBA privileges before invoking.
+     */
+    public void releaseAny(final long sessionId) {
+        if (sessionId < 0) {
+            return;
+        }
+        cache.invalidate(sessionId);
+    }
+
+    /**
+     * Privileged eviction of every cached entry, bypassing the
+     * subject-match check. Intended as a DBA-accessible "force reaper"
+     * to recover from cases where many entries are stranded by failed
+     * client sessions and would otherwise wait for the per-entry
+     * 2-minute timeout. Callers MUST verify DBA privileges before
+     * invoking.
+     *
+     * @return the approximate number of entries evicted
+     */
+    public long releaseAll() {
+        final long size = cache.estimatedSize();
+        cache.invalidateAll();
+        cache.cleanUp();
+        return size;
     }
 }
--- a/exist-core/src/main/java/org/exist/http/SessionManager.java
+++ b/exist-core/src/main/java/org/exist/http/SessionManager.java
@@ -28,8 +28,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.exist.xquery.value.Sequence;
 
+import java.security.SecureRandom;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * @author <a href="mailto:adam@evolvedbinary.com">Adam Retter</a>
@@ -40,14 +40,16 @@ public class SessionManager {
     private static final Logger LOG = LogManager.getLogger(SessionManager.class);
     private static final long TIMEOUT = 120_000;  // ms (e.g. 2 minutes)
 
-    private final AtomicInteger sessionIdCounter = new AtomicInteger();
-    private final Cache<Integer, QueryResult> cache;
+    private final SecureRandom random = new SecureRandom();
+    private final Cache<Long, QueryResult> cache;
 
     private static class QueryResult {
+        final int subjectId;
         final String query;
         final Sequence sequence;
 
-        private QueryResult(final String query, final Sequence sequence) {
+        private QueryResult(final int subjectId, final String query, final Sequence sequence) {
+            this.subjectId = subjectId;
             this.query = query;
             this.sequence = sequence;
         }
@@ -62,19 +64,46 @@ public class SessionManager {
         cache = cacheBuilder.build();
     }
 
-    public int add(final String query, final Sequence sequence) {
-        final int sessionId = sessionIdCounter.getAndIncrement();
-        cache.put(sessionId, new QueryResult(query, sequence));
+    /**
+     * Cache a query result and return a cryptographically random,
+     * non-negative session id.
+     *
+     * @param subjectId the id of the subject that owns this cache entry;
+     *                  only callers with the same subject id can later
+     *                  retrieve or release it
+     * @param query     the original query text (used as a secondary
+     *                  consistency check)
+     * @param sequence  the result sequence
+     * @return a non-negative session id; collisions with existing live
+     *         entries are retried
+     */
+    public long add(final int subjectId, final String query, final Sequence sequence) {
+        long sessionId;
+        do {
+            sessionId = random.nextLong() & Long.MAX_VALUE;
+        } while (cache.getIfPresent(sessionId) != null);
+        cache.put(sessionId, new QueryResult(subjectId, query, sequence));
         return sessionId;
     }
 
-    public Sequence get(final String query, final int sessionId) {
-        if (sessionId < 0 || sessionId >= sessionIdCounter.get()) {
-            return null; // out of scope
+    /**
+     * Look up a cached query result. Returns null when no entry exists,
+     * the entry has expired, the query text differs, or the requesting
+     * subject does not match the subject that created the entry.
+     */
+    public Sequence get(final int subjectId, final String query, final long sessionId) {
+        if (sessionId < 0) {
+            return null;
         }
 
         final QueryResult cached = cache.getIfPresent(sessionId);
         if (cached == null) {
+            return null;
+        }
+
+        if (cached.subjectId != subjectId) {
+            LOG.warn("Session {} requested by subject {} but owned by subject {}; refusing.",
+                    sessionId, subjectId, cached.subjectId);
             return null;
         }
 
@@ -86,9 +115,22 @@ public class SessionManager {
         }
     }
 
-    public void release(final int sessionId) {
-        if (sessionId < 0 || sessionId >= sessionIdCounter.get()) {
-            return; // out of scope
+    /**
+     * Release a cached query result. Only the original subject may
+     * release the entry.
+     */
+    public void release(final int subjectId, final long sessionId) {
+        if (sessionId < 0) {
+            return;
+        }
+        final QueryResult cached = cache.getIfPresent(sessionId);
+        if (cached == null) {
+            return;
+        }
+        if (cached.subjectId != subjectId) {
+            LOG.warn("Session {} release requested by subject {} but owned by subject {}; refusing.",
+                    sessionId, subjectId, cached.subjectId);
+            return;
         }
         cache.invalidate(sessionId);
     }
--- a/exist-core/src/main/java/org/exist/http/SessionManager.java
+++ b/exist-core/src/main/java/org/exist/http/SessionManager.java
@@ -39,8 +39,8 @@ public class SessionManager {
 
     private static final Logger LOG = LogManager.getLogger(SessionManager.class);
     private static final long TIMEOUT = 120_000;  // ms (e.g. 2 minutes)
+    private static final SecureRandom RANDOM = new SecureRandom();
 
-    private final SecureRandom random = new SecureRandom();
     private final Cache<Long, QueryResult> cache;
 
     private static class QueryResult {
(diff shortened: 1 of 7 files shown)


duncdrum (changes requested):
good stuff, I m not 100% sure I understand the consequences of tying cache releases to subjectIDs. 

The failing container tests are a concern, and all related to the contents of the PR.

dizzzz (approved):
LGTM

joewiz:

@duncdrum thanks for the review. Pushed `59d406f1f0` to address both points.

### Failing container tests (`Test and Publish Container Images`)

I reproduced the failures locally against `existdb/existdb:bugfix-sec` and traced them:

- **Test 5 ("logs are error free")** trips on `01-connect-spec.bats` greping `docker logs` for the literal token `ERROR`. C1's SEVERE warning is emitted via `LOG.error(...)` from `RealmImpl.warnIfAdminPasswordIsEmpty`, and the shipped image starts with admin's default empty password, so the warning fires every startup. **The warning is the feature** -- it's exactly the unsafe state we want operators to see -- so I narrowed the bats grep to exclude the specific `SECURITY WARNING` line, leaving the test guarding against any other unexpected `ERROR`-level output.

- **Tests 15 and 17 (POST query as `admin:nimda`)** were failing in CI but actually pass locally on a clean `bats` run against the same image. While digging into why CI rejected them, I uncovered a pre-existing latent bug that **C2 of this PR exposed for the first time**:

  `RESTServer.java`'s `AUTHENTICATED_USERS_ONLY` gate (added in `ee7a1392293`, 2017) was checking `currentSubject.getId() == RealmImpl.GUEST_GROUP_ID`. By coincidence, `GUEST_GROUP_ID` and `ADMIN_ACCOUNT_ID` share the same numeric value `1048574` -- account ids and group ids live in different namespaces, but the constants collide. The intended check was "reject the **guest account**"; the actual effect was the inverse: **admin** received 403 and **guest** sailed through. The bug was invisible while the shipped default was `xquery-submission=enabled` (no gate runs at all). C2 flips the default to `authenticated`, which makes the gate reachable -- and on a CI host where any pagination state happens to slip in, admin's POST gets gated as guest. Fix: compare against `GUEST_ACCOUNT_ID`.

Local bats run on the rebuilt image: 18/19 pass. Test 18 ("PUT xq should succeed and trigger auto-register") fails locally on macOS but passes in CI -- it also fails on a clean `develop` build locally, so it's a pre-existing macOS-specific RESTXQ auto-register flake unrelated to this PR.

### Subject-scoped cache (C3)

You're right to flag that this needs a clearer trade-off. Concretely:

- **Intent:** combined with the move to `SecureRandom` long ids, binding cache entries to the creating subject's id closes a window where an attacker who guessed (now: brute-forced 2^63) a session id could observe or release another caller's paginated result set. Defense in depth on top of the random ids.

- **Side effect:** if the creating subject ends without calling `release` (browser closes, network drop, RESTXQ caller crashes), the entry can no longer be evicted by id from a different subject -- only the 2-minute timeout reaper can reclaim it. That's intentional, and acceptable in the current REST surface because the entire pagination GET sequence is driven by the same caller. There is no admin-scoped "evict by id" code path today; if we ever add one, that code can pass `RealmImpl.SYSTEM_ACCOUNT_ID` (or a privilege check) into `release` to bypass the subject match.

I'll add a paragraph to the PR description summarising this.

exist-core/src/main/java/org/exist/http/SessionManager.java
```diff
@@ -40,14 +40,16 @@ public class SessionManager {
     private static final Logger LOG = LogManager.getLogger(SessionManager.class);
     private static final long TIMEOUT = 120_000;  // ms (e.g. 2 minutes)
 
-    private final AtomicInteger sessionIdCounter = new AtomicInteger();
-    private final Cache<Integer, QueryResult> cache;
+    private final SecureRandom random = new SecureRandom();
```
reinhapa: Maybe use a static variable as `SecureRandom` should be thread save

exist-core/src/main/java/org/exist/management/client/JMXServlet.java
```diff
@@ -262,6 +262,11 @@ boolean isFromLocalHost(final HttpServletRequest request) {
      * @return TRUE if request contains correct value for token, else FALSE
      */
     boolean hasSecretToken(final HttpServletRequest request, final String token) {
+        if (StringUtils.isBlank(token)) {
```
reinhapa: If this is the only place, where we use the `StringUtils`, we could also use this:
```suggestion
        if (token == null||token.isBlank()) {
```

joewiz:

@reinhapa thanks for the review. Both points addressed in c436f8ac6c:

1. **`SessionManager`** — promoted `SecureRandom` from a per-instance field to `private static final RANDOM`; it's thread-safe, so a single shared instance is correct.
2. **`JMXServlet`** — dropped the `org.apache.commons.lang3.StringUtils` import. Both call sites (`isBlank` in `hasSecretToken`, `isNotBlank` in `writeXmlData`'s timeout-parameter handling) now use native `String.isBlank()` with explicit null checks.

Verified with `RESTServiceTest` (43/43 pass) on the rebuilt module.

reinhapa (approved):

duncdrum:
@joewiz 

> C1's SEVERE warning is emitted via `LOG.error(…)`  

I'm  not trying to start a philosophical debate, but booting with the OOTB configuration shouldn't result in an `ERROR` . Make the warning strongly worded, include mortally horrifying ascii art if need be, but it should remain `WARN` imv.

> Fix: compare against `GUEST_ACCOUNT_ID`.

sounds good

> Test 18 ("PUT xq should succeed and trigger auto-register") fails locally on macOS but passes in CI -- it also fails on a clean develop build locally, so it's a pre-existing macOS-specific RESTXQ auto-register flake unrelated to this PR.

It is not flake. I doubt its is OS specific as well. The test has been running and confirmed manually (I put it in) many times going back to ~ exist 5 if not 4. If auto-register for restxq enabled instance isn't working, we need to figure out why. 

>  the entry can no longer be evicted by id from a different subject -- only the 2-minute timeout reaper can reclaim it. 

hmm that sounds worrisome to me. it sounds like we are one network flake away from collecting a bunch of unclearable 
caches. Wouldn't it be safer to have a dba accessible force switch to execute the reaper?

joewiz:

@duncdrum thanks for pushing back -- pushed `77ec9a1f4d` covering all three points.

**1. Demoted to `WARN`.** Agreed; OOTB shouldn't produce `ERROR`. `RealmImpl.warnIfAdminPasswordIsEmpty()` now logs at `WARN` with a hard-to-miss banner so it still stands out in operator logs. Also reverted the bats grep narrowing in `01-connect-spec.bats` since the WARN-level message no longer trips an `ERROR` grep.

**2. `GUEST_ACCOUNT_ID` fix** -- already in `59d406f1f0`, no change.

**3. PUT xq auto-register.** You were right -- it's not flake and not macOS-specific. Rebuilt the image clean and ran the full bats suite locally: **19/19 pass, including this test.** My earlier diagnosis was wrong; I was almost certainly testing against a stale image. Apologies for the misdirection. `rest:resource-functions()` correctly returns the registered `host:function1` after the PUT of `forgot.xqm`.

**4. DBA force-reaper.** Added. `SessionManager` now exposes `releaseAny(long)` and `releaseAll()` as privileged paths that bypass the subject-match check. The existing GET `?_release=` parameter handler in `RESTServer` accepts:
- a numeric session id -- DBA may release any entry; non-DBA still subject-restricted by `release()`
- the literal `all` (or `*`) -- DBA-only force reaper; 403 to non-DBA; logged at INFO with username + count

Verified end-to-end against the rebuilt image:
```
X-Session-Id: 9062459834385983712               # cached query
_release=all (no creds)        => HTTP 401
_release=all (admin:nimda)     => HTTP 200
INFO  RESTServer: DBA 'admin' force-released all cached query results (1 entries).
```

`RESTServiceTest` 43/43 pass.

duncdrum (approved):

PR #6294, opened 2026-05-04
[bugfix] fix resource leak in SQL module 

dizzzz:
Fix for resource leak in SQL module

### DIFF
--- a/extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/ExecuteFunction.java
+++ b/extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/ExecuteFunction.java
@@ -185,8 +185,8 @@ public class ExecuteFunction extends BasicFunction {
 
         // get the Connection
         final long connectionUID = ((IntegerValue) args[0].itemAt(0)).getLong();
-        final Connection con = SQLModule.retrieveConnection(context, connectionUID);
-        if (con == null) {
+        final Connection connection = SQLModule.retrieveConnection(context, connectionUID);
+        if (connection == null) {
             throw new XPathException(this, "No such SQL Connection");
         }
 
@@ -194,7 +194,7 @@ public class ExecuteFunction extends BasicFunction {
 
         //setup the SQL statement
         String sql = null;
-        Statement stmt = null;
+        Statement statement = null;
 
         try {
             final boolean makeNodeFromColumnName;
@@ -207,7 +207,7 @@ public class ExecuteFunction extends BasicFunction {
 
                 // get the static SQL statement
                 sql = args[1].getStringValue();
-                stmt = con.createStatement();
+                statement = connection.createStatement();
                 makeNodeFromColumnName = ((BooleanValue) args[2].itemAt(0)).effectiveBooleanValue();
                 if (args.length == 5) {
                     namespaceUri = args[3].itemAt(0).getStringValue();
@@ -219,7 +219,7 @@ public class ExecuteFunction extends BasicFunction {
                 }
 
                 //execute the static SQL statement
-                executeResult = stmt.execute(sql);
+                executeResult = statement.execute(sql);
 
             } else if (args.length == 4 || args.length == 6) {
                     //get the prepared statement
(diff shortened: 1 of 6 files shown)


duncdrum (approved):
LGTM, one codacy flag

extensions/modules/sql/src/main/java/org/exist/xquery/modules/sql/ExecuteFunction.java
```diff
@@ -222,33 +222,36 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
...
-                makeNodeFromColumnName = ((BooleanValue) args[3].itemAt(0)).effectiveBooleanValue();
-                if (args.length == 6) {
-                    namespaceUri = args[4].itemAt(0).getStringValue();
-                    namespacePrefix = args[5].itemAt(0).getStringValue();
-                } else {
-                    // The default namespace for result elements.
-                    namespaceUri = NAMESPACE_URI;
-                    namespacePrefix = PREFIX;
-                }
+                    if (stmt.getConnection() != con) {
```
reinhapa: Do we need this identity equals here or would be using `equals` be sufficient?

line-o: I suspect that we _need_ reference comparison here.

reinhapa: hmm, suspect or knowledge... seems fishy to me....

line-o: The question really boils down to if it is OK for the SQL connection to be just equal by its contents / hashcode or does it have to be the _same_ connection object.

line-o: On the other hand we are dealing with a connection pool. So it is actually likely that we have more than one connection (object) for the same server that a prepared statement could use.
So with this in mind I am more inclined to say we should switch to 

```
if (!con.equals(stmt.getConnection())) { ... throw ... }
```

dizzzz: It is a check to verify if the connection used in the statement is the same of the actual connection.

The interesting thing is... on line 210 it reads

```
stmt = connection.createStatement();
```

so what is the check for? well... for cached/stored connections! So we need to be sure that the right connection is used.

dizzzz: with a bit of help....

The connection comparison stmt.getConnection() != connection is a security and correctness guard.
When using prepared statements in eXist-db's SQL module, the caller provides two separate handles: a connection-handle (long) and a statement-handle (long). These are looked up independently from the module's stored maps — there is no inherent linkage between them at the XQuery level.

Why the check is needed
1.
Correctness: A PreparedStatement is bound to the Connection that created it. It was compiled/planned for that specific connection's session and transaction context. Executing it via a different connection object is undefined behavior in JDBC and will typically cause a driver-level error or silent misbehavior.
2.
Security: Without this check, if a caller (accidentally or maliciously) passes a connection-handle that belongs to a different session than the one used to create the prepared statement, the code would attempt to mix them. The check enforces that both handles actually belong together.
3.
Fail-fast with a clear message: Rather than getting a cryptic JDBC driver exception deep in execution, the check throws a clear XPathException immediately: "SQL Connection does not match that used for creating the PreparedStatement".

Why != (reference equality) is correct here
Connection objects are not value types — two distinct Connection instances represent two distinct database sessions even if they connect to the same database with the same credentials. Reference equality (!=) is exactly the right comparison here: it checks that the same object (i.e., the same live session) is being used for both the statement and the execution call.

line-o (approved):
👍🏼

line-o:
@dizzzz do you plan to backport this?

reinhapa (approved):

dizzzz:
Yes but only a simplified one

dizzzz:
Ready to pull @line-o

PR #6295, opened 2026-05-04
[fix] reinstate batch operations for ngram:contains

line-o:
### Description:

This PR reverts parts of #6286 and #6093
Merging this PR means we need to reopen #2204

`ngram:contains#2` would fail to use batch operation mode when in a predicate.

The easiest way to reproduce it is in function documentation search.

### Reference:

reopen #2204

### Type of tests:

None added. @joewiz  was so kind to create benchmarks and share analysis in #6296

### DIFF
--- a/extensions/indexes/ngram/src/main/java/org/exist/xquery/modules/ngram/NGramSearch.java
+++ b/extensions/indexes/ngram/src/main/java/org/exist/xquery/modules/ngram/NGramSearch.java
@@ -566,22 +566,11 @@ public class NGramSearch extends Function implements Optimizable {
     @Override
 	public int getDependencies() {
         final Expression stringArg = getArgument(0);
-        if (Type.subTypeOf(stringArg.returnsType(), Type.NODE)
-            && !Dependency.dependsOn(stringArg, Dependency.CONTEXT_ITEM)) {
-            // Check if any other argument depends on a local iteration variable.
-            // If so, the expression cannot be bulk-evaluated during ForExpr preEval
-            // because the variable value changes per iteration. (GH-2204)
-            // Only LOCAL_VARS (same for/let scope) prevent bulk evaluation;
-            // CONTEXT_VARS (outer scope) are static and safe to preselect.
-            for (int i = 1; i < getArgumentCount(); i++) {
-                if (Dependency.dependsOnLocalVar(getArgument(i))) {
-                    return Dependency.CONTEXT_SET + Dependency.CONTEXT_ITEM;
-                }
-            }
-            return Dependency.CONTEXT_SET;
-        } else {
+        if (!Type.subTypeOf(stringArg.returnsType(), Type.NODE)
+                || Dependency.dependsOn(stringArg, Dependency.CONTEXT_ITEM)) {
             return Dependency.CONTEXT_SET + Dependency.CONTEXT_ITEM;
         }
+        return Dependency.CONTEXT_SET;
     }
 
     @Override
--- a/extensions/indexes/ngram/src/test/java/org/exist/indexing/ngram/CustomIndexTest.java
+++ b/extensions/indexes/ngram/src/test/java/org/exist/indexing/ngram/CustomIndexTest.java
@@ -543,6 +543,7 @@ public class CustomIndexTest {
      * must not raise XPTY0004.
      */
     @Test
+    @Ignore("skip until better solution is found")
     public void ngramContainsWithForVariable() throws PermissionDeniedException, XPathException, EXistException {
         final BrokerPool pool = existEmbeddedServer.getBrokerPool();
         try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()))) {
--- a/extensions/indexes/ngram/src/test/xquery/ngram/for-variable.xql
+++ b/extensions/indexes/ngram/src/test/xquery/ngram/for-variable.xql
@@ -100,6 +100,7 @@ function fv:tearDown() {
  : collection variable. This is the pattern from the original bug report.
  :)
 declare
+    %test:pending
     %test:assertEquals("Chair", "Table")
 function fv:for-variable-in-where-clause() {
     for $q in doc($fv:COLLECTION || "/queries.xml")//q


reinhapa (approved):

line-o:
I think this can be closed. Will retest now that #6300 was merged

line-o:
No, performance for ngram:contains is still not good enough. But I will limit this to only ngram

joewiz:

@line-o Reproduced your finding -- and the diagnosis is **not what either of us expected**. The residual slowness has nothing to do with `ngram:contains` or its `getDependencies()` guard. Reverting the guard barely moves the numbers (within noise on the affected workload).

I built a synthetic xqdoc-shaped corpus matching the function-documentation app's data (200 modules x 30 functions = 6,000 functions, ngram-indexed on `xqdoc:description`/`name`/`signature`/`param`/`return`, mirroring the conf the app installs). Then ran the **exact** query shapes from `function-documentation/.build/modules/app.xqm` plus narrowing variants to isolate the cause.

### The result table (develop @ 675a2ddcb5, post-#6300)

```
                                                          time   preSel  eval  optimize-firings  notes
(7) //xqdoc:description[ngram:contains(., 'X')]            3ms      1     1          1
(4) //xqdoc:function[ngram:contains(comment/desc, 'X')]    3ms      1     1          1
(5) let $a := //xqdoc:xqdoc return $a//xqdoc:function[..]  3ms      1     1          1
(3) //xqdoc:function[..] | //xqdoc:module[..]              4ms      2     1          2     <-- equivalent semantics, fast
(1) ACTUAL fundocs shape: //(xqdoc:function[..] | xqdoc:module[..])
                                                          37ms      2     1          2     <-- ~10x slower
(8) search-everywhere (7-branch union)                    35ms      7     1          7
```

`preSel` and `eval` show that `ngram:contains` is **already** taking the bulk-evaluation path in the slow variants -- preSelect fires correctly, eval uses the cached result, no per-tuple fallback. So the gap isn't in ngram at all.

### Stripping out the predicate confirms it

Same shapes, no predicates whatsoever:

```
                                                          time   notes
(10) //xqdoc:function | //xqdoc:module                     3ms     count=6200
 (9) //(xqdoc:function | xqdoc:module)                   165ms     count=6200, ~50x slower
```

**No ngram, no optimizer, no predicates** -- and the gap is still there at 50x. The path engine evaluates the union-step `//(A | B)` by materialising the full descendant axis and applying the parenthesized expression as a generic step, instead of dispatching each branch through the structural index by qname.

### Even more striking: parens on a single name test trigger the same path

```
//xqdoc:function          (no parens)                3ms
//(xqdoc:function)        (parens, single name)    104ms     <-- 35x slower
//*                       (full descendant scan)    22ms
```

The `//*` full descendant scan is **5x faster** than `//(xqdoc:function)` because the engine has a fast path for wildcard descendant. But parens around a name test defeat that too. The parens are the trigger, not the union.

### What's actually happening

`//(X)` parses as `descendant-or-self::node() / FilteredExpression(LocationStep(X))`. The `Optimizer.visitFilteredExpr` (`exist-core/src/main/java/org/exist/xquery/Optimizer.java:159-186`) already has logic to lift a single `LocationStep` out of a `FilteredExpression` -- but only when the FilteredExpression has predicates to lift onto it (line 170: `if (optimizable)`). When the parenthesized expression has no predicates of its own (which it doesn't in any of the slow variants), the simplification is skipped.

For the union case, there's no simplification at all -- a `UnionExpr` of `LocationStep`s inside a FilteredExpression is treated as a generic step expression at runtime.

### Two paths forward

**1. Fix the function-documentation app.** Rewrite each `app:search-*` function from `$app:data//(A[..] | B[..])` to `$app:data//A[..] | $app:data//B[..]`. Single-day PR, immediate ~10x speedup, no eXist changes needed. Probably the right next step regardless.

**2. Fix eXist's optimizer.** Extend `visitFilteredExpr` to also simplify the no-predicate case (drop the parens, keep the inner `LocationStep`) and the union-of-LocationSteps case (distribute the parent `//` axis over each branch). That's the proper fix and would close this for everyone, not just function-documentation.

Either way, **the ngram guard isn't load-bearing here** -- I'd suggest closing this PR (or pivoting it to mark `for-variable-in-where-clause` `%test:pending` only, with a note that GH-2204's perf concern is now elsewhere) and opening a separate fix for the parenthesized-step issue. Happy to draft the eXist optimizer fix as a small surgical PR similar to #6300 if that helps.

Reproducer / diagnostic class is in my local `feature/index-jmh-benchmarks` worktree as `FundocsSearchDiagnostic` -- I'll fold it into the JMH module on PR #6296 so this becomes a permanent regression-test target.

line-o:
Please go for the real fix in the optimizer.
We have had to hand-optimize for as long as I work with exist and it is tedious work. Not just for maintainers but all users. The benefits warrant investing resources here.

joewiz:

@line-o Posted #6303 with the optimizer-side fix for the simple-parens half of this. Numbers on the synthetic xqdoc corpus:

```
//xqdoc:function          3 ms     (baseline, no parens)
//(xqdoc:function)      104 ms     (the bug, before fix)
//(xqdoc:function)       37 ms     (after #6303, ~3x speedup)
```

Two things I want to flag explicitly because they shaped the patch's scope:

1. **The "exact class check" was load-bearing.** A first version using `step instanceof PathExpr` failed `XQuery3Tests.bang.precedence1` because `(-.)` is a `UnaryExpr` (which extends `PathExpr`) containing a `.` `LocationStep`, and a generic instanceof would unwrap the `.`, dropping the unary minus. Many other operators (`BinaryOp`, `OpNumeric`, `ConcatExpr`, `EnclosedExpr`, `LogicalOp`, `RangeExpression`, ...) inherit from `PathExpr` for the same reason. The patch uses `step.getClass() == PathExpr.class` to match only the parser's generic parens-wrapper, never the semantic subclasses.

2. **The union case `//(A | B)` is not in this PR.** Distributing the parent path over union branches requires either a `PathExpr.replaceAllSteps`-style API or rewriting the outer `PathExpr` at its parent (which is often not a `RewritableExpression`). I tried the in-place rewrite and it was getting hairy enough to deserve its own scope. For the function-documentation app the workaround is to rewrite the queries from `$app:data//(A | B)` to `$app:data//A | $app:data//B` -- I'll open a PR against `function-documentation` separately. The eXist-side union-distribution remains as future work.

Closing this PR (#6295) makes sense whenever you're ready -- the diagnosis turned out to point away from the ngram guard.

PR #6296, opened 2026-05-05
[test] Add exist-indexes-jmh benchmark module (FLWOR/index where-clause shapes)

joewiz:
## Summary

A new `exist-indexes-jmh` Maven module that benchmarks four shapes of `ngram:contains()` usage inside FLWOR expressions. Motivated by the discussion on #6295 (which reverts #6093 and #6286 in favour of fixing [GH-2204](https://github.com/eXist-db/exist/issues/2204) at the FLWOR / where-clause level rather than per-index): without numbers it's hard to tell whether any future where-clause optimization is actually moving the needle, and harder still to tell whether a future change accidentally breaks the bulk-probe fast path.

## What changed

- New module `exist-indexes-jmh/` registered in the `micro-benchmarks` profile of the parent pom.
- `NgramWhereClauseBenchmark` covers four shapes:
  - `shapeALiteral` &mdash; `//LINE[ngram:contains(., 'Denmark')]`
  - `shapeALetVar` &mdash; `let $q := 'Denmark' return //LINE[ngram:contains(., $q)]`
  - `shapeBForVarPredicate` &mdash; `for $q in $TERMS return //LINE[ngram:contains(., $q)]` (the rewrite target)
  - `shapeBForVarWhere` &mdash; `for $q in $TERMS where //LINE[ngram:contains(., $q)] return $q` (the regression case from GH-2204)
- Hamlet (~270 KB) is vendored from `exist-samples/.../shakespeare/` under its existing public-domain notice. The collection-config registers an ngram index on `LINE` and `SPEAKER`.
- Run via the `bin/run-bench.sh` wrapper, which materialises the runtime classpath from `target/classpath.txt` and invokes `org.openjdk.jmh.Main`. Shading into an uberjar collides with Saxon's `ServiceLoader`, which is why we don't.
- The module declares the JMH annotation processor in `<annotationProcessorPaths>` so `META-INF/BenchmarkList` actually gets generated &mdash; the parent pom's `annotationProcessorPaths` is set to only `jsr305`, which silently shadows the JMH processor. The existing `exist-core-jmh` is affected by the same issue (its shaded jar has no `BenchmarkList`); a follow-up could apply the same fix there.

## Baseline numbers (this branch, 1 warmup / 2 measurement / 1 fork)

```
Benchmark                                        Mode  Cnt    Score   Error  Units
NgramWhereClauseBenchmark.shapeALiteral          avgt    2    0.117          ms/op
NgramWhereClauseBenchmark.shapeALetVar           avgt    2    0.142          ms/op
NgramWhereClauseBenchmark.shapeBForVarPredicate  avgt    2    0.358          ms/op
NgramWhereClauseBenchmark.shapeBForVarWhere      avgt    2  850.163          ms/op
```

Shape A is flat (~0.12 ms): literal and let-bound paths are equivalent &mdash; both single bulk probes. `shapeBForVarPredicate` is ~3x Shape A, exactly matching the expected "five invariant bulk probes." `shapeBForVarWhere` is **roughly 6,000x slower** than the predicate form, because the where-clause path collapses into per-tuple EBV evaluation. Closing that gap is the goal Lever 1 (BaseX-equivalent `toPredicate` rewrite at compile time) is aimed at.

## Roadmap (out of scope for this PR)

- Range-index benchmarks (`range:eq`, `range:field-eq`)
- Lucene benchmark (`ft:query`)
- XMark-scaled corpus via on-demand `xmlgen`
- CI nightly job posting numbers to a tracking issue

## Test plan

- [x] `mvn install -pl exist-indexes-jmh -am` succeeds
- [x] `mvn license:check -pl exist-indexes-jmh` passes
- [x] `codacy-cli analyze --tool pmd` on the new file returns no findings
- [x] `bin/run-bench.sh NgramWhereClauseBenchmark -wi 1 -i 2 -f 1` runs all four shapes to completion and produces the numbers above
- [x] Smoke check inside `@Setup` verifies the corpus + ngram index produce non-zero hits for the literal query before any benchmark fires &mdash; a misconfigured index would now fail loudly instead of silently reporting "fast empty results"

## References

- #6295 &mdash; reverts #6093 and #6286, prompts this work
- [GH-2204](https://github.com/eXist-db/exist/issues/2204) &mdash; the underlying for-bound where-clause regression
- #6207 &mdash; the IN_PREDICATE / LOCAL_VARS fix that protected Shape A but not Shape B


line-o:
#2204 is just part of the bigger issue of batch operations on indexed nodes in predicates with changing search terms.

I would like to see us having the same benchmarks for ngram, range and full-text indices.

joewiz:

@line-o, I've extended the PR with three new benchmark classes:

- `RangeEqWhereClauseBenchmark` — `range:eq` against a plain range index on `SPEAKER` (the path #6093 originally guarded)
- `RangeFieldEqWhereClauseBenchmark` — `range:field-eq` against a combined range index with a `speaker` field on `SPEECH` (also covered by #6093; different `getDependencies()` path)
- `LuceneWhereClauseBenchmark` — `ft:query` against a lucene index on `LINE` (the path #6286 originally guarded)

Same Hamlet corpus, same four shapes per class. README has a consolidated 4-class × 4-shape baseline.

**A surprise from the numbers** worth flagging before any FLWOR-level rewrite is designed:

```
Benchmark                                               Mode  Cnt    Score   Units
NgramWhereClauseBenchmark.shapeBForVarPredicate         avgt    2    0.345  ms/op
NgramWhereClauseBenchmark.shapeBForVarWhere             avgt    2  928.391  ms/op   <-- ~2700x
RangeEqWhereClauseBenchmark.shapeBForVarPredicate       avgt    2    0.479  ms/op
RangeEqWhereClauseBenchmark.shapeBForVarWhere           avgt    2    0.488  ms/op   <-- flat
RangeFieldEqWhereClauseBenchmark.shapeBForVarPredicate  avgt    2    0.213  ms/op
RangeFieldEqWhereClauseBenchmark.shapeBForVarWhere      avgt    2    0.206  ms/op   <-- flat
LuceneWhereClauseBenchmark.shapeBForVarPredicate        avgt    2    1.172  ms/op
LuceneWhereClauseBenchmark.shapeBForVarWhere            avgt    2    1.114  ms/op   <-- flat
```

On develop today (post-#6295 revert), the where-clause regression appears **only on ngram**. Range, range-field, and lucene's where-clause shapes are all within noise of their predicate shapes. Their `getDependencies()` paths handle the loop-variable case correctly without the per-index guards.

That suggests one or more of:

1. GH-2204 might be ngram-specific, and the right fix could be smaller-scoped than a generic FLWOR rewrite — worth investigating what ngram does differently from `range:eq` and `ft:query` in `getDependencies()` / index-rewrite analysis.
2. The per-index guards reverted by #6093 / #6286 may have been guarding a case that those indexes didn't actually regress on (i.e., the guards were defensive rather than load-bearing).
3. A FLWOR-level Lever 1 rewrite would still be valuable as a generalization, but the urgency is different than if all four indexes were equally affected.

Two cross-checks I haven't run yet that would sharpen this:

- **Pre-revert baseline** — re-run on a develop tip from before #6295 merged, with #6093 and #6286 both still in place. Confirms the guards weren't quietly slowing Shape A on any index.
- **Bigger corpus** — Hamlet is small enough that the Shape B per-tuple penalty on lucene/range may be hidden in microsecond noise; XMark `-f 1` (~110 MB) would surface it if it's there.

Both are noted in the README's roadmap; happy to chase either as follow-ups before the FLWOR rewrite work begins, or in parallel with it.

joewiz:

@line-o, big follow-up. First, an important correction to my previous comment: I'd characterised what we were seeing as "a regression that appears only on ngram." That framing was wrong &mdash; what we were actually measuring was the *cost* of the correctness guards from #6093 and #6286 (both already merged into develop), and that cost is only visible on ngram. The other indexes' guards aren't load-bearing in any way the Hamlet corpus surfaces. Sorry for the muddle.

The latest commit (f65ec388ec) expands the benchmark in three ways:

1. **Three plays.** `macbeth.xml` and `r_and_j.xml` are now vendored alongside `hamlet.xml`; the @Setup loads all three. ~660 KB total.
2. **`@Param int termCount` of {5, 50, 100}** on all four benchmark classes. The term lists cycle through the per-class base list to fill the requested count.
3. **A pre-#6295 vs. post-#6295 comparison run**, captured in the README. I cherry-picked your `b064cddafc` onto a temp branch, rebuilt exist-core + extensions, and reran. The state of the PR's tracked code is unchanged &mdash; only the bench output reflects the comparison.

### The headline number

| Benchmark / shape (termCount=100) | develop | +#6295 | Delta |
|---|---|---|---|
| Ngram shapeBForVarWhere | **35,190 ms** | **58.7 ms** | **~600x faster** |
| Ngram shapeBForVarPredicate | 10.57 ms | 8.33 ms | ~unchanged |
| RangeEq shapeBForVarWhere | 14.93 ms | 14.70 ms | unchanged |
| RangeFieldEq shapeBForVarWhere | 2.23 ms | 2.13 ms | unchanged |
| Lucene shapeBForVarWhere | 45.94 ms | 45.33 ms | unchanged |

So #6295 buys a real, ngram-only perf win &mdash; about 600x on the where-clause shape at termCount=100. Range and lucene's numbers are within noise of each other before and after. Term-list scaling is linear across every shape, in both states; nothing quadratic anywhere.

### Three things to read from this

1. **#6295 is a meaningful perf win on ngram, not a no-op.** The 600x figure is robust across all three termCount values (5, 50, 100). This is exactly the win you wrote the PR for, and the benchmark now demonstrates it numerically.
2. **The guards on range and lucene weren't doing visible work.** Reverting them costs zero perf on this corpus; they were defensive against a bug those indexes don't currently exhibit.
3. **The trade #6295 makes is squarely on ngram.** Faster, but at the cost of the `for-variable-in-where-clause` correctness test which gets marked `%test:pending`. The other indexes' tests stay correct because their bulk path was already correct on Shape B.

### What this implies for the FLWOR Lever 1 design

Two paths that this data supports:

- **Land #6295 now and design Lever 1 against the ngram bug specifically** &mdash; the test asymmetry is real, range/lucene don't need anything done at the FLWOR layer to keep Shape B correct, and a ngram-targeted fix would close the loop without dragging the other index modules into the rewrite.
- **Hold #6295 until Lever 1 is ready** &mdash; lose the 600x ngram win in the meantime, but ship the trade as a single coherent change.

Either is defensible; the data doesn't pick. What it does say is that any FLWOR-level rewrite needs to *only* fix ngram-style per-tuple-EBV and *not* perturb the bulk paths of the other three index APIs &mdash; the benchmark's `shapeALetVar` row across all four classes is the regression-test we can run against any future patch.

Full numbers are in `exist-indexes-jmh/README.md`. Happy to chase any next angle &mdash; XMark scaling, or a deeper dive into why ngram's path differs from the others, or anything else.

line-o:
In light of the findings here I would propose to drop #6295 - or at least narrow it down to just ngram as the other paths do not have the performance regression but they fix real issue there.
What do others think?

joewiz:

@line-o, follow-up: I dug into ngram vs range/lucene to find the actual asymmetry, and it turned out to be a much smaller and more interesting bug than any of us were sketching. **It's not in any of the index code -- it's a missing visitor override on `WhereClause` itself.** Five lines in `exist-core/src/main/java/org/exist/xquery/WhereClause.java`.

PR #6300 has the fix and the full diagnostic walkthrough.

The short version: `WhereClause` is the only FLWOR clause type missing an `accept(ExpressionVisitor)` override, so visitors -- including the Optimizer pass that wraps indexable `LocationStep` predicates in `#exist:optimize#` pragmas -- never descended into the where expression. Predicates inside a where clause therefore never got the bulk-evaluation rewrite, and `ngram:contains` fell through to per-tuple EBV evaluation. On Hamlet+Macbeth+R&J that's ~47,460 calls per outer iteration; with the fix, it's the expected 15.

For the GH-2204 reproducer at termCount=100 the where-shape goes from **35 seconds to 59 ms** -- roughly the same shape as the predicate form, with full correctness preserved (no `%test:pending` markers needed).

Implications:

* **#6295's revert is no longer needed in any form.** The perf win it was reaching for arrives without the correctness regression. Both `%test:pending` markers can be dropped.
* **The per-index guards in #6093 and #6286 become redundant** but harmless. They were defending against the missed-optimization case this fix removes. They can stay or go; a separate cleanup PR can remove them later if desired.
* **The `exist-indexes-jmh` benchmark in this PR now has a permanent role**: it's the regression-test harness for verifying that any future FLWOR / index-optimization change keeps the bulk-evaluation path intact across all four index APIs. Worth landing whether or not you also land #6300.

I left this PR's existing four commits untouched and posted #6300 as a separate, surgical fix targeting `develop`.

line-o:
Now that #6300 was merged I retested on a local build of the fundocs app. The search was still  _a lot slower_ than before the changes to ngram:contains. It also _seems_ to be slower without the revert in #6295 but even with that we do not get back to the speed before.

line-o (approved):
I would like to see this merged and added to our toolbox

reinhapa (approved):

line-o:
There is an additional test we should wait for before merging this

duncdrum (commented):
I need a bit more time here

line-o:
@duncdrum I would prefer that the added subproject exist-index-jmh would be kept and your benchmarks would become part of it.

adamretter:
xmlgen can be replaced with it's successor there we built in Java as part of @alanpaxton nosqlbench paper

line-o:
@adamretter Thanks for the suggestion. Do you have a link to the project?

line-o:
@duncdrum can this be merged now?

duncdrum:
@line-o I ll mark I my review once I m through.

duncdrum (changes requested):
After spending some time with the PR, there are a few things I'm not happy about. 
1. I dislike the new maven reactor layout, we shouldn't split micro-benchmarks over multiple top level reactors. Maintainability wise it set's us on the wrong path, and will just lead to outdated copies of e.g. `con.xml` to accumulate, and unnecessary  compile work when building. I would either see a `jmh-parent` or top-level `benchmarks` reactor with subproject as the better choice. There is some design decision we should discuss in a call. Because this PR merges cleanly with other benchmark PRs, let's keep this point for a follow up PR (the single stress test might find a new home there too…) 
2. I don't like the shell script. It's OS specific, and adds an unecessary dependency. Use `exec-maven-plugin`, with `${path.separator}`, and adjust the readme if staying with fat jar.  in either case java and maven all the way. 
3. `core-benchmark` uses a shaded jar we should stick with one approach across benchmarks (shaded or fat) with `ServicesResourceTransformer` as we do in the docker pom we can avoid the Saxon problem mentioned in the pom. Which is where this surfaced, but I doubt it's actually specific to Saxon. Shaded will likely require additional fiddling to set up properly. The basic question shaded or fat is a style question I have no preference for benchmarks. 
4. Like in the docker reactor use `xml-maven-plugin` and XSLT to get the `conf.xml` we want / need for benchmarks, this avoid the drift potential we introduce here.
5. The PR mentions GC but no benchmark actually profiles it, use JMH's `GCProfiler` or `StackProfiler` to do so. 

I do like the improved style of these new benchmarks, and we should update `core-benchmarks` to be inline in a follow up PR.

I m agree that we need to run benchmarks on CI and get some kind of regression Dashboard (also follow up)

joewiz:

Thanks @duncdrum — point-by-point, and parking the work until you confirm the choices below before I push anything:

**1. Reactor layout.** Agree on the design concern; happy to take it to a call. Also relevant: PR #6324 (JMH benchmarks for `following::`/`preceding::` axes) landed in `exist-core-jmh` rather than `exist-indexes-jmh` last night specifically because this PR was still open — it has the same "another top-level micro-benchmark reactor" smell. Whatever layout we settle on (`jmh-parent` vs `benchmarks` reactor) should fold #6324 into it too. Punting to a follow-up PR per your suggestion.

**2. Shell script.** Will switch to `exec-maven-plugin` with `${path.separator}` and drop the OS-specific script.

**3. Shaded vs fat.** Leaning toward **fat with `ServicesResourceTransformer`**, on the grounds that (a) the current PR is already fat so the diff stays smaller, (b) `ServicesResourceTransformer` directly addresses the Saxon `META-INF/services` collision the pom comment flags, and (c) consistency-with-`core-benchmark` is a separate question that can land in the follow-up reactor PR. If you'd rather we match `core-benchmark`'s shaded shape now (instead of converging in the follow-up), that's a one-line change to call out.

**4. `xml-maven-plugin` + XSLT for `conf.xml`.** Agreed — will mirror the `exist-docker` pattern (`exist-docker/src/main/xslt/` is the precedent) so the convention is consistent across reactors.

**5. GC profiling.** Will add `GCProfiler` to one or two of the index-walk benchmarks. `StackProfiler` is more situational; happy to add it on a specific benchmark if you want.

On the "improved style of these new benchmarks" — happy to fold a `core-benchmarks` style refresh into the follow-up PR alongside the reactor layout call.

**Holding pushes on this PR until you sign off** on the shaded-vs-fat call (item 3). Items 2/4/5 are mechanical and ready to apply as soon as you confirm. Item 1 stays as a follow-up regardless.

duncdrum:
I m ok with fat, and we can switch core to fat in the follow up. Thanks for taking care of the other options, @eXist-db/core if anybody has reasons to stay with shaded, let's hear it.

joewiz:

@duncdrum cleanup applied per your 2026-05-10 review (commit 8b5400790a):

- **Item 2** (drop shell script): `bin/run-bench.sh` is gone; benchmarks now run via `exec-maven-plugin`. README rewritten end-to-end:
    ```
    mvn exec:exec -pl exist-indexes-jmh
    mvn exec:exec -pl exist-indexes-jmh -Dbenchmark.args="NgramWhereClauseBenchmark.shapeA.* -wi 5 -i 10 -f 3 -prof gc"
    ```
    The uber-jar is also directly runnable via `java -jar` once built.
- **Item 3** (shaded vs fat): switched to a fat uber-jar built by `maven-shade-plugin` with `ServicesResourceTransformer` (your endorsement at 13:53), `org.openjdk.jmh.Main` as the manifest entry point. Confirmed Saxon's `META-INF/services/javax.xml.transform.TransformerFactory` is preserved end-to-end (Lucene benchmark runs cleanly through `BrokerPool` boot, which exercises Saxon). The original Saxon comment in the pom is gone, and the now-redundant `maven-dependency-plugin > build-classpath` execution went with it.
- **Item 4** (`xml-maven-plugin` + XSLT for `conf.xml`): mirrored the `exist-docker` log4j2.xml pattern. Source is `extensions/indexes/indexes-integration-tests/src/test/resources-filtered/conf.xml`; the XSLT (`exist-indexes-jmh/src/main/xslt/conf-jmh.xslt`) adds the plain `range-index` storage module + corresponding XQuery module on top. Generated into `target/generated-resources/conf/conf.xml`, then resource-filtered for `${basedir}` resolution. The local `src/main/resources/conf.xml` is deleted.
- **Item 5** (GC profiling): `-prof gc` is now in the default `benchmark.args`, so the standard `mvn exec:exec` invocation prints `gc.alloc.rate`, `gc.alloc.rate.norm`, `gc.count`, and `gc.time` rows alongside the headline numbers. Verified locally on `NgramWhereClauseBenchmark.shapeALiteral`. README documents `-prof stack` as a follow-up option for specific benchmarks.

**Item 1** (reactor layout) stays as a follow-up PR per your suggestion. PR #6324 (the JMH benchmarks for `following::`/`preceding::` axes that landed in `exist-core-jmh`) should fold into whatever layout the call settles on.

Local verification: `mvn install -pl exist-indexes-jmh -am -DskipTests` is green; quick smoke runs of `NgramWhereClauseBenchmark.shapeALiteral`, `LuceneWhereClauseBenchmark.shapeALiteral`, and `RangeEqWhereClauseBenchmark.shapeALiteral` all complete with sensible numbers and the GC profiler attached. Codacy PMD on the unchanged Java sources is clean.

Ready for re-review when you have a moment.

duncdrum (approved):

Issue #2817, opened 2019-06-15
The results of 'where' clause in FLWOR are different in memory and database.


jesselet:
### What is the problem
<pre><code>
for $i in $data1//item, $j in $data2//item
where $i/name = $j/name
return $i/name
</code></pre>
The results of 'where' clause in FLWOR are different in memory and database.
### What did you expect
The results of 'where' clause in FLWOR are the same in memory and database.
### Describe how to reproduce or add a test
<pre><code>
xquery version "3.0";
let $col := "/db/apps/test"
let $data1 :=
&lt;data>
    &lt;item>&lt;name>1&lt;/name>&lt;/item>
    &lt;item>&lt;name>2&lt;/name>&lt;/item>
&lt;/data>
let $data2 :=
&lt;data>
    &lt;item>&lt;name>2&lt;/name>&lt;/item>
    &lt;item>&lt;name>3&lt;/name>&lt;/item>
&lt;/data>
let $test := function ($data1, $data2) {
    for $i in $data1//item, $j in $data2//item
    where $i/name = $j/name
    return $i/name
}
return (
    try {
        $test($data1, $data2)
    } catch * { "test1: " || $err:description },
    (:in eXist 2.2, 3.6.1, 4.6.1, 4.7.0: OK:)

    try {
        xmldb:create-collection("", $col),
        let $data1 := doc(xmldb:store($col, "data1.xml", $data1))
        let $data2 := doc(xmldb:store($col, "data2.xml", $data2))
        return $test($data1, $data2)
    } catch * { "test2: " || $err:description }
    (:in eXist 2.2: OK:)
    (:in eXist 3.6.1, 4.6.1, 4.7.0: EXCEPTION: variable '$j' is not set:)
)
</code></pre>
### Context information
- OS: Windows 7
- Java: jre1.8.0_202
- eXist-db: 2.2, 3.6.1, 4.6.1, 4.7.0

duncdrum:
yup confirmed, and looks like a bug that has been there for a while, ouch. 

tested on `5.0.0-RC8-SNAPSHOT` via docker

line-o:
I would also see this as a bug.
It can be worked around with:
```
    let $j := $data2//item
    return
    for $i in $data1//item
    where $i/name = $j/name
    return $i/name
```

line-o:
The equivalent form described in the [spec](https://www.w3.org/TR/xquery-31/#id-xquery-for-clause) does also fail:
```
    for $i in $data1//item
[...]

Issue #2283, opened 2018-11-08
Additional criteria in where clause breaks KWIC when also using a full-text index


shanemcgarry:
### What is the problem
> I am attempting to use the KWIC module to give me information on full text searches that are run against my exist instance. This all works fine but occasionally I also want to add additional filters on other, non-full text fields. While the search itself works in this instance, the KWIC module returns nothing. I can reproduce this same behaviour using the eXist-db Shakespeare demo. For example:

**Standard xQuery**
```
xquery version "3.1";
import module namespace kwic="http://exist-db.org/xquery/kwic";
for $x in doc("db/apps/demo/data/hamlet.xml")//PLAY/ACT/SCENE
where $x//SPEECH[ft:query(.,"alas mad lady")] 
order by ft:score($x) descending
return
<page>
<data>{$x}</data>
<match>
{kwic:summarize($x, <config width="40"/>)}
</match>
</page>
```

> This query works as expected. However if I add additional criteria to the where clause (e.g. $x//SPEAKER = 'HAMLET', kwic:summarize no longer returns results, even though there are results in the query itself (and there are still full-text matches). For example:

**xQuery with Additional Where Criteria**
```
xquery version "3.1";
import module namespace kwic="http://exist-db.org/xquery/kwic";
for $x in doc("db/apps/demo/data/hamlet.xml")//PLAY/ACT/SCENE
where $x//SPEECH[ft:query(.,"alas mad lady")] and $x//SPEAKER = 'HAMLET'
order by ft:score($x) descending
return
<page>
<data>{$x}</data>
<match>
{kwic:summarize($x, <config width="40"/>)}
</match>
</page>
```

> This query returns results in the <data> node but the <match> node is now empty (despite valid full-text matches existing and thus KWIC should provide a summary).

### What did you expect
> Regardless of the criteria supplied in the where clause, I should be see a kwic:summarize if there are matching keywords from the full-text criteria.

### Describe how to reproduce or add a test
> Steps to reproduce:

1. Open the public eXide application (http://exist-db.org/exist/apps/eXide/index.html)
2. Run the "Standard xQuery" from above. Verify results in <match> node exist.
3. Run the "xQuery with Additional Where Criteria" from above. Verify <match> node is now empty.

### Context information
> My local environment is configured as follows:

> Please always add the following information
- eXist-db version + Git Revision hash e.g. eXist-db 3.0 /  acd0c14 : 4.4.0
- Java version (e.g. Java8u121): 1.8.0_181
[...]

Issue #1423, opened 2017-04-24
undefined local variable in WHERE-clause


opax:
### What is the problem

In some cases, a local variable bound in a FLWOR expression is not readable in the WHERE clause.

An example is this query:
```
    for $doc in xmldb:xcollection( "/db/tmp" )
    let $dn := "/db/tmp" || util:document-name($doc)
    where (doc-available($dn))
    return true()
```

### What did you expect

The query should return a sequence of true()-values for each document in `/db/tmp`. 
Instead, the following error is reported:
```
err:XPDY0002 variable '$dn' is not set.
```

Some equivalent expressions in the WHERE-clause  do not cause the error, e.g. `where (doc-available($dn) or false())` works as expected.

### Describe how to reproduce or add a test

A XQSuite unit test is attached.

### Context information

- eXist 3.1.1 / fe05cf7 / build date 20170322200706 
- java version "1.8.0_91"
- MacOS El Capitan
- 64 bit
- default configuration

opax:
FWIW, the same query worked in eXist-db 2.2.

joewiz:
Confirmed that the test suite shows the error is still present in 3.5.0:

```xml
<testsuites>
    <testsuite package="http://exist-db.org/xquery/test/issue1423"
        timestamp="2017-10-03T17:22:56.816-04:00" failures="0" pending="0" tests="3" time="PT0.098S">
        <testcase name="failing_with_XPDY0002" class="t:failing_with_XPDY0002">
            <error type="err:XPDY0002"
                message="It is a dynamic error if evaluation of an expression relies on some part of the dynamic context that has not been assigned a value. variable '$dn' is not set."
            />
        </testcase>
        <testcase name="thisAlsoWorks" class="t:thisAlsoWorks"/>
        <testcase name="thisWorks" class="t:thisWorks"/>
    </testsuite>
</testsuites>
```

The key bit of the error:

> It is a dynamic error if evaluation of an expression relies on some part of the dynamic context that has not been assigned a value. variable '`$dn`' is not set.

joewiz:
@duncdrum These tests are not in the repo, as far as I know, so we need another PR with pending tests.

line-o:
exist 5.3.0 is still affected

Issue #990, opened 2016-05-03
3.0 RC1: Incorrect tuple mapping to where clause?


median1st:
In some cases the WHERE clause in a simple FOR-WHERE-RETURN construct gets the variable mapped to the whole sequence iterated by the FOR clause instead to a single member at a time. This changes to the correct behavior once the construct is expanded of a do-nothing LET:

[test.xml.zip](https://github.com/eXist-db/exist/files/246945/test.xml.zip)
Create a simple document in an empty collection (/db/amdr/workspace in this case). An example is attached to the issue (test.xml.zip).
Try evaluating the following code:

```
declare function local:test($node as node()*, $filter as function(node()) as xs:boolean ) as node()*
{
    for $p in $node//f
    where $filter($p)
    return $p
};

let $test := collection('/db/amdr/workspace')/test//b
 return local:test($test, function ($i as node()) as xs:boolean { true() })

```
This results in error

```
exerr:ERROR XPTY0004: The actual cardinality for parameter 1 does not match the cardinality declared in the function's signature: ($i as node()) xs:boolean. Expected cardinality: exactly one, got 14. [at line 6, column 20, source: String]
In function:
	(node()) [11:35:String]
	local:test(node()*, function) [11:8:String]

```
Now add a dummy LET clause above the failing where clause so that the function looks like:
```
declare function local:test($node as node()*, $filter as function(node()) as xs:boolean ) as node()*
{
    for $p in $node//f
    let $z := $p
    where $filter($p)
    return $p
};
```
Note that the $z variable is completely useless here, but the behavior changes - now the $p variable is mapped to one sequence member at a time and the function no more complains about incorrect cardinality and all 14 elements are returned.
This does not happen when a literal XML element is used instead of addressing collection contents.

duncdrum:
@joewiz could you triage this please, thx.

duncdrum:
@joewiz ping is this still an issue, reproducible?

joewiz:
I can confirm that this issue still affects 5.0.0-RC8-SNAPSHOT (develop).

joewiz:
This issue still affects eXist 5.3.0-SNAPSHOT (current develop HEAD).

PR #6300, opened 2026-05-06
[bugfix] WhereClause: descend into where-expression during AST visitor walks (closes GH-2204)

### LINKED ISSUE #2204: ngram:contains() bug?

### LINKED ISSUE #5352: [BUG] where clauses do not appear to use indexes as they should, query consequently very slow

**Describe the bug**
An xquery query using `where` clauses should be rewritten to use defined attribute indexes, but is not. It is consequently very slow.

I run this query against a large database (the auction site generated by the xmlgen tool) [xmlgen](https://github.com/eliben/xmlgen)
```
xquery version "3.1";

for $p in /site/people/person
    let $a :=
        for $t in /site/closed_auctions/closed_auction
        where $t/buyer/@person = $p/@id
        return $t
    let $c := count($a)
    where $c > 0
    return <item person="{$p/name/text()}" id="{$p/@id}">{count ($a)}</item>
```
I have created and loaded the following indexes (sample from collection.xconf)
```
<index>
        <!-- Range indexes -->
        <range>
            <create qname="@id" type="xs:string"/>
            <create qname="@person" type="xs:string"/>
        </range>
    </index>
```

**Expected behavior**
The performance should be much quicker than it is, and Monex should report the configured attribute index queries as being used.

By rewriting the query I can get the same result with the expected performance
and index usage
```
xquery version "3.1";

for $p in /site/people/person
    let $a := /site/closed_auctions/closed_auction[buyer/@person = $p/@id]
    let $c := count($a)
    where $c gt 0
    return <item person="{$p/name/text()}" id="{$p/@id}">{count($a)}</item>
```
To quote @adamretter :

Basically, the Query Rewriter in eXist-db is very bad at re-writing where clauses to use indexes as an optimisation, it is better at re-writing predicates. So in general, when working with eXist-db, I advise people to use predicates instead of where clauses.

**To Reproduce**
- Generate the default auction database with xmlgen `./xmlgen-unix -f 1 -o file-auction.xml`
- Install the above indexes in collection.xconf
- load `file-auction.xml` from eXide using `xmldb:store-files-from-pattern`
- paste the above query in eXide

One option is to use [xst](https://www.npmjs.com/package/@existdb/xst), and copy and paste the output produced by running `xst info` here:**

 - Build: eXist-7.0.0-SNAPSHOT
 - Java: openjdk version "21.0.2" 2024-01-16
 - OS: Mac OS X 14.5

**Additional context**
- How is eXist-db installed? built from source (evolvedbinary/exist HEAD)

joewiz:
## Summary

`WhereClause` was missing an `accept(ExpressionVisitor)` override. The other FLWOR clause classes (`ForExpr`, `LetExpr`, `WindowExpr`, `OrderByClause`, `GroupByClause`) all override `accept` to dispatch to their corresponding `visitXxx` methods; `WhereClause` was the lone sibling falling through to `AbstractExpression.accept` &rarr; the generic `visit()` &rarr; `processWrappers()` &mdash; which only descends into `Atomize`, `DynamicCardinalityCheck`, `DynamicNameCheck`, `DynamicTypeCheck`, `UntypedValueCheck`, `PathExpr`. `WhereClause` is none of those, so the visitor stopped there.

Concretely: the Optimizer pass that wraps optimizable `LocationStep` predicates in `#exist:optimize#` pragmas never reached predicates inside a `where` clause. Indexed predicate functions therefore missed the bulk-evaluation path and fell through to per-tuple EBV evaluation &mdash; with `ngram:contains`, that's roughly N nodes &times; M outer iterations of `eval()` calls.

## The fix

Five lines in `exist-core/src/main/java/org/exist/xquery/WhereClause.java`:

```java
@Override
public void accept(final ExpressionVisitor visitor) {
    visitor.visitWhereClause(this);
}
```

`DefaultExpressionVisitor.visitWhereClause` already exists and descends into `whereExpr` and `returnExpr` correctly &mdash; only the entry point was missing.

## Diagnostic numbers

Reproducer (the GH-2204 shape, three Shakespeare plays, 5 search terms):

```xquery
for $q in ('Denmark', 'England', 'Norway', 'Polonius', 'France')
where collection('/db/test')//LINE[ngram:contains(., $q)]
return $q
```

Counters added during diagnosis (not in this PR):

| | LocationSteps visited | Optimize.eval | preSelect | NGramSearch.eval | uncached | time |
|---|---|---|---|---|---|---|
| before | **0** | **0** | **0** | **47,460** | **47,460** | **1882 ms** |
| after | 2 | 15 | 15 | 77 | 0 | **6 ms** |

That's a **~313x speedup** at termCount=5; **~600x** at termCount=100. Same correct results in both cases.

## Benchmark evidence

The full benchmark methodology and JMH numbers across all four index APIs (`ngram:contains`, `range:eq`, `range:field-eq`, `ft:query`) live in `exist-indexes-jmh` &mdash; see #6296. The benchmark module is independently useful as the regression-test harness for any future FLWOR / index-optimization work; the numbers there directly motivated this fix.

## Test plan

- [x] `extensions/indexes/ngram` &mdash; 39 tests pass (6 skipped, 0 failures)
- [x] `extensions/indexes/range` &mdash; 422 tests pass (3 skipped, 0 failures)
- [x] `extensions/indexes/lucene` &mdash; 659 tests pass (40 skipped, 0 failures)
- [x] `exist-core` `XQuery3Tests` + `XPathQueryTest` + `OptimizerTest` &mdash; 1140 tests pass (5 skipped, 0 failures)
- [x] The existing `for-variable-in-where-clause` and `for-variable-in-field-eq` regression tests added in #6093 / #6286 still pass with this fix
- [x] Codacy PMD: no new findings on the changed file
- [x] License check passes

## Impact on related PRs

- **#6295** (Juri's "reinstate batch operations" revert) is no longer needed in any form. The perf win it was reaching for arrives via this fix without the correctness regression that revert introduced. The two `%test:pending` markers in #6295 can be dropped because the underlying tests pass against this branch.
- **#6093 and #6286** added per-index `dependsOnLocalVar` / `dependsOnVar` guards in `range:eq`, `range:field-eq`, `ngram:contains`, and `ft:query`. With this fix, those guards become redundant &mdash; they were defending against the missed-optimization case this PR removes. The guards still work and tests still pass with them in place; a follow-up cleanup PR can remove them if desired, but it isn't required.

## Spec / issue references

- Closes https://github.com/eXist-db/exist/issues/2204
Closes https://github.com/eXist-db/exist/issues/5352
- See PR #6296 for the JMH benchmark module that surfaces the numbers and serves as the regression harness for any future FLWOR / index work

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/WhereClause.java
+++ b/exist-core/src/main/java/org/exist/xquery/WhereClause.java
@@ -152,6 +152,11 @@ public class WhereClause extends AbstractFLWORClause {
     }
 
     @Override
+    public void accept(final ExpressionVisitor visitor) {
+        visitor.visitWhereClause(this);
+    }
+
+    @Override
     public void dump(ExpressionDumper dumper) {
         dumper.display("where", whereExpr.getLine());
         dumper.startIndent();


joewiz:

A follow-up datapoint that may interest the maintainers, and which ties this PR to two more places in our existing record:

**Issue [#5352](https://github.com/eXist-db/exist/issues/5352)** is the same bug. The reporter's example uses `where $t/buyer/@person = $p/@id` -- a `GeneralComparison` between two indexed attributes -- and quotes @adamretter: "the Query Rewriter in eXist-db is very bad at re-writing `where` clauses to use indexes as an optimisation, it is better at re-writing predicates." That's the structural diagnosis, three years before this PR. With the visitor descending properly into the where expression, `GeneralComparison` predicates inside `where` clauses get the same `#exist:optimize#` rewrite that they get in predicate position. I've added `Closes #5352` to the PR body.

**Wolfgang's [tuning docs](https://exist-db.org/exist/apps/doc/tuning#prefer-xpath-predicates-over-where-expressions)** advise "prefer XPath predicates over `where` expressions" with the rationale that "the for ... where expression forces the query engine into a step-by-step iteration over the input sequence ... any possible optimizations are lost." That advice was correct prose for what the engine actually did; the underlying cause was this missing visitor override.

To verify the fix's reach extends beyond explicit indexed function calls, I added a `GeneralComparisonWhereClauseBenchmark` to PR #6296 that exercises Wolfgang's exact "predicate vs. where" case (`//SPEAKER[. = $term]` with a range index on `SPEAKER`). Numbers on develop @ b917e1ab1d:

```
                              develop      with #6300       speedup
shapeBForVarWhere @ 5         14.4 ms        9.4 ms          ~1.5x
shapeBForVarWhere @ 100      264.1 ms      177.2 ms          ~1.5x
```

Smaller speedup than the ngram case (~600x) because `GeneralComparison`'s un-optimized fallback is "iterate nodes, do string equality" -- O(n) per outer iteration but cheap per node, whereas ngram's un-optimized fallback was a per-node ngram-index probe (much more expensive per call). Same structural fix; absolute payoff scales with how expensive the missed bulk-eval was.

Post-#6300, the where-shape lands within noise of the predicate-shape for indexed `GeneralComparison` predicates -- closing the gap Wolfgang's docs warn about. The advice remains correct for two scenarios: (a) where-clauses without an indexable predicate (the FLWOR semantics still iterate per-tuple -- there's no `Optimizable` for the pragma to bind to), and (b) joins between documents where `where` can't be replaced by a predicate at all. Worth a documentation edit at some point; I'll leave that decision to the maintainers.

A side finding the new benchmark surfaces but this PR doesn't address: the `shapeALetVar` form (`let $q := 'HAMLET' return //SPEAKER[. = $q]`) runs **~28x slower than the literal form** even though both should be a single bulk index probe. Same numbers pre- and post- #6300, so it's a separate, pre-existing missed-optimization -- the optimizer doesn't fold a let-bound term as invariant for `GeneralComparison` the way it does for explicit indexed function calls (`range:eq($nodes, $q)`). Probably worth its own future PR.

joewiz:

Triaged the five other open `where clause` issues against this PR. Here's what's covered and what isn't, verified by running each issue's reproducer (or its closest equivalent) on both `develop` alone and `develop + this PR`:

| Issue | Status with #6300 |
|---|---|
| [#2204](https://github.com/eXist-db/exist/issues/2204) -- ngram:contains in for-bound where | **Closed** (already in PR body) |
| [#5352](https://github.com/eXist-db/exist/issues/5352) -- where clause not using indexes (Adam Retter's diagnosis) | **Closed** (already in PR body) |
| [#1423](https://github.com/eXist-db/exist/issues/1423) -- undefined local variable in WHERE-clause | **Already fixed in develop** prior to this PR. The verbatim reproducer (`for $doc in xmldb:xcollection("/db") let $dn := ... where (doc-available($dn)) ...`) runs without `XPDY0002` on both branches. Whatever closed it, it was earlier than #6300. Could be closed if a maintainer wants to verify against the original setup. |
| [#990](https://github.com/eXist-db/exist/issues/990) -- incorrect tuple mapping in for-where with function reference | **Not affected.** Still fails on both branches with `XPTY0004: actual cardinality ... got 3` (the reported symptom). The bug looks like a separate FLWOR-tuple-binding issue in higher-order function calls, not visitor descent. |
| [#2817](https://github.com/eXist-db/exist/issues/2817) -- cross-document join `where $i/name = $j/name` differs in memory vs. database | **Not affected.** Still fails on both branches with `Variable '$j' is not declared` -- a static-analysis variable-scope bug specific to the second `for` clause, unrelated to where-clause optimization. |
| [#2283](https://github.com/eXist-db/exist/issues/2283) -- additional where criteria breaks KWIC matches | **Likely not affected.** Didn't run the full reproducer (KWIC requires the demo Hamlet collection) but the symptom -- KWIC's `<match>` becomes empty when `and $x//SPEAKER = 'HAMLET'` is added to a where containing `ft:query` -- is about how lucene-attached match metadata propagates to ancestor nodes during predicate evaluation. The visitor-descent fix doesn't change that path. Worth a separate investigation. |

So the headline impact of this PR on the open `where clause` issue backlog is: closes **#2204 and #5352 directly**, doesn't change **#990, #2817, #2283** (they're separate bugs each meriting its own fix), and **#1423** was apparently fixed at some earlier point and could be closed independently.

To the broader documentation point @line-o / @adamretter / others may want to consider: the long-standing advice "prefer XPath predicates over where" was a workaround for what turns out to be a 5-line missing `accept` override. With #6300 in place, the where-vs-predicate gap closes for indexed cases, and the doc could be revised to scope the advice more narrowly to:

1. Where clauses without an indexable predicate (per-tuple FLWOR semantics still apply).
2. Cross-document joins where `where` is structurally necessary.

I'll leave the doc edit decision to the maintainers.

reinhapa (approved):

duncdrum (approved):

line-o (approved):
Very interesting results!

PR #6301, opened 2026-05-06
[bugfix] Reject huge dayTimeDuration values in date/time arithmetic

### LINKED ISSUE #5045: [BUG] Adding a large number of days to a date appears to never finish
The following query seems to run forever:

```xquery
fn:current-date() + xs:dayTimeDuration("P1712073600000D")
```

In comparison:
* BaseX - returns an answer almost instantly
* Saxon - returns and error instantly: `E value is too large to be expressed in microseconds`.

The root of the problem is not within eXist-db itself. eXist-db is simply relying on a 3rd-party implementation of the Java interface `javax.xml.datatype.XMLGregorianCalendar`. Various implementations are available, but eXist-db is using Xerces 2.12 for this.

As the JRE embeds a version of Xerces, we can actually just reproduce the problem in pure Java:

```java
import javax.xml.datatype.DatatypeConfigurationException;
import javax.xml.datatype.DatatypeFactory;
import javax.xml.datatype.Duration;
import javax.xml.datatype.XMLGregorianCalendar;
import java.util.GregorianCalendar;

public class Main {

    public static void main(final String args[]) throws DatatypeConfigurationException {

        final DatatypeFactory factory = DatatypeFactory.newInstance();
        System.out.println("Using implementation from: " + factory.getClass().getPackageName());

        // get an XML calendar for `now`
        final GregorianCalendar nowCalendar = new GregorianCalendar();
        final XMLGregorianCalendar xmlCalendar = factory.newXMLGregorianCalendar(nowCalendar);

        // create a very large duration
        final Duration duration = factory.newDuration("P1712073600000D");

        // add the duration to the XML calendar
        xmlCalendar.add(duration);

        // print the new date from the XML calendar... do we ever get to this point?
        System.out.print(xmlCalendar);
    }
}
```

Tested on Java 17, the above never appears to finish.

Next, I need to debug this in detail to figure out if it is just very slow (due to the large duration value), or if it will never complete. More to come...

joewiz:
## Summary

Closes #5045.

`xs:date + xs:dayTimeDuration("P1712073600000D")` (and similar for `xs:dateTime` / `xs:time`) used to lock up the JVM thread for over an hour of CPU. The XMLGregorianCalendar.add() implementation underneath iterates per day, so very large day-time durations cause the call to take seconds, then minutes, then hours.

This PR adds a pre-validation guard in `OrderedDurationValue.addDurationToDate()` that rejects day-time magnitudes exceeding 1,000,000 Gregorian years with `FODT0001`, before delegating to `gc.add()`. The arithmetic engine itself is unchanged for any in-range input, preserving its tested correctness for premodern, BCE, and year-zero-adjacent dates.

## Empirical motivation

`XMLGregorianCalendar.add(Duration)` latency on Zulu 21 / Apple M-series, measured before this PR:

| Duration | dayTime portion | Latency |
|---|---|---|
| `P1000000D` | 2.7k years | 10 ms |
| `P100000000D` | 270k years | 222 ms |
| `P1000000000D` | 2.7M years | 1.0 s |
| `P10000000000D` | 27M years | 12 s |
| `P100000000000D` | 273M years | 119 s |
| `P1712073600000D` (the reported case) | 4.7B years | ~30+ min |

`xs:yearMonthDuration` is **not** affected — Xerces handles those in O(1) (sub-ms even at 1.2 billion months). Only the dayTime portion of a duration triggers the slow path, so the guard is conditioned on `secondsValueSigned()` magnitude.

## What changed

| File | Change |
|---|---|
| `exist-core/src/main/java/org/exist/xquery/value/DurationValue.java` | New constants `DATE_ARITH_DAYS_LIMIT` (1,000,000 Gregorian years) and `MAX_DATE_ARITH_SECONDS`. New `checkDateArithMagnitude(BigDecimal)` helper that raises `FODT0001` if the duration's day-time magnitude exceeds the threshold. |
| `exist-core/src/main/java/org/exist/xquery/value/OrderedDurationValue.java` | One-line guard call (`checkDateArithMagnitude(secondsValueSigned())`) in `addDurationToDate()`, between the input year check and `gc.add(duration)`. |
| `exist-core/src/test/java/org/exist/xquery/value/DateTimeTest.java` | Three JUnit regression cases: huge-duration plus/minus must throw `FODT0001` in <1 s; 100,000-year duration must still succeed. |
| `exist-core/src/test/xquery/xquery3/dateTimeOverflow.xqm` | Four XQSuite cases covering the reported reproducer at the XQuery level for `xs:date`, `xs:dateTime`, and `xs:time`. |

Threshold rationale: 1,000,000 Gregorian years caps worst-case latency at ~365 ms on the test hardware while permitting any realistic date/time arithmetic. The threshold is a hard-coded constant; per XQuery 3.1 §10.1.1 the duration value-space is implementation-defined.

## Spec references

- XQuery 3.1 [§10.1.1 Limits and Precision](https://www.w3.org/TR/xpath-functions-31/#datetime-limits) — duration value-space is implementation-defined; FODT0001 covers date/time overflow.
- XPath 3.1 functions [op:add-dayTimeDuration-to-date](https://www.w3.org/TR/xpath-functions-31/#func-add-dayTimeDuration-to-date) — implementations are not required to handle dates outside 0001–9999.

## Test plan

- [x] `mvn test -pl exist-core -Dtest='Date*Test,Time*Test,Duration*Test,XPathQueryTest'` — 376 tests pass (4 pre-existing skips)
- [x] `mvn test -pl exist-core -Dtest='xquery.xquery3.XQuery3Tests'` — 1015 tests pass, including the four new GH-5045 cases in `dateTimeOverflow.xqm`
- [x] PMD via `codacy-cli analyze --tool pmd` on changed files — no new findings (pre-existing findings unchanged)
- [x] `mvn license:check -pl exist-core -o` — clean
- [x] Built with `mvn install -pl exist-core -am -DskipTests`

### DIFF
--- a/exist-core/src/test/java/org/exist/xquery/value/DateTimeTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/value/DateTimeTest.java
@@ -24,13 +24,16 @@ package org.exist.xquery.value;
 import com.googlecode.junittoolbox.ParallelRunner;
 import org.exist.xquery.Constants;
 import org.exist.xquery.Constants.Comparison;
+import org.exist.xquery.ErrorCodes;
 import org.exist.xquery.XPathException;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 /**
  *	note: some of these tests rely on local timezone override to -05:00, done in super.setUp()
@@ -425,5 +428,50 @@ public class DateTimeTest extends AbstractTimeRelatedTestCase {
 		final DurationValue d = new DayTimeDurationValue("P3DT1H15M");
 		final AbstractDateTimeValue r = new DateTimeValue("2000-10-27T09:57:00");
 		assertDateEquals(r, t.minus(d));
+	}
+
+	// Issue GH-5045: huge xs:dayTimeDuration values used to make
+	// XMLGregorianCalendar.add() iterate per-day, locking up CPU for hours.
+	// The pre-validation guard must reject them with FODT0001 fast.
+
+	@Test
+	public void plus_hugeDayTimeDuration_rejectsFast() throws XPathException {
+		final AbstractDateTimeValue t = new DateTimeValue("2026-05-05T12:00:00Z");
+		final DurationValue d = new DayTimeDurationValue("P1712073600000D");
+		final long start = System.nanoTime();
+		try {
+			t.plus(d);
+			fail("Expected FODT0001 for excessively large duration");
+		} catch (final XPathException ex) {
+			assertSame(ErrorCodes.FODT0001, ex.getErrorCode());
+		}
+		final long elapsedMs = (System.nanoTime() - start) / 1_000_000L;
+		assertTrue("Guard must reject in <1s; took " + elapsedMs + "ms", elapsedMs < 1000);
+	}
+
+	@Test
+	public void minus_hugeDayTimeDuration_rejectsFast() throws XPathException {
+		final AbstractDateTimeValue t = new DateTimeValue("2026-05-05T12:00:00Z");
+		final DurationValue d = new DayTimeDurationValue("P1712073600000D");
+		final long start = System.nanoTime();
+		try {
+			t.minus(d);
+			fail("Expected FODT0001 for excessively large duration");
+		} catch (final XPathException ex) {
+			assertSame(ErrorCodes.FODT0001, ex.getErrorCode());
+		}
+		final long elapsedMs = (System.nanoTime() - start) / 1_000_000L;
+		assertTrue("Guard must reject in <1s; took " + elapsedMs + "ms", elapsedMs < 1000);
+	}
+
+	@Test
+	public void plus_largeButPermittedDayTimeDuration_works() throws XPathException {
+		// 100,000 years -- well under the 1,000,000-year guard, must succeed.
+		final AbstractDateTimeValue t = new DateTimeValue("2026-05-05T12:00:00Z");
+		final DurationValue d = new DayTimeDurationValue("P36524250D");
+		final ComputableValue r = t.plus(d);
+		// Year arithmetic: 2026 + 100000 = 102026 (modulo Gregorian leap-year drift).
+		assertTrue("expected year ~102026, got " + r,
+				r.getStringValue().startsWith("102026") || r.getStringValue().startsWith("102025"));
 	}
 }
--- a/exist-core/src/test/java/org/exist/xquery/value/DateTimeTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/value/DateTimeTest.java
@@ -435,7 +435,7 @@ public class DateTimeTest extends AbstractTimeRelatedTestCase {
 	// The pre-validation guard must reject them with FODT0001 fast.
 
 	@Test
-	public void plus_hugeDayTimeDuration_rejectsFast() throws XPathException {
+	public void plusHugeDayTimeDurationRejectsFast() throws XPathException {
 		final AbstractDateTimeValue t = new DateTimeValue("2026-05-05T12:00:00Z");
 		final DurationValue d = new DayTimeDurationValue("P1712073600000D");
 		final long start = System.nanoTime();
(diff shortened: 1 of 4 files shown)


exist-core/src/test/java/org/exist/xquery/value/DateTimeTest.java
```diff
@@ -426,4 +429,49 @@ public void minus3() throws XPathException {
 		final AbstractDateTimeValue r = new DateTimeValue("2000-10-27T09:57:00");
 		assertDateEquals(r, t.minus(d));
 	}
+
+	// Issue GH-5045: huge xs:dayTimeDuration values used to make
+	// XMLGregorianCalendar.add() iterate per-day, locking up CPU for hours.
+	// The pre-validation guard must reject them with FODT0001 fast.
+
+	@Test
+	public void plus_hugeDayTimeDuration_rejectsFast() throws XPathException {
```
reinhapa: Address Codacy issue `The JUnit 4 test method name 'plus_hugeDayTimeDuration_rejectsFast' doesn't match '[a-z][a-zA-Z0-9]*'`

exist-core/src/test/java/org/exist/xquery/value/DateTimeTest.java
```diff
@@ -426,4 +429,49 @@ public void minus3() throws XPathException {
...
+			fail("Expected FODT0001 for excessively large duration");
+		} catch (final XPathException ex) {
+			assertSame(ErrorCodes.FODT0001, ex.getErrorCode());
+		}
+		final long elapsedMs = (System.nanoTime() - start) / 1_000_000L;
+		assertTrue("Guard must reject in <1s; took " + elapsedMs + "ms", elapsedMs < 1000);
+	}
+
+	@Test
+	public void minus_hugeDayTimeDuration_rejectsFast() throws XPathException {
```
reinhapa: Address Codacy issue `The JUnit 4 test method name 'plus_hugeDayTimeDuration_rejectsFast' doesn't match '[a-z][a-zA-Z0-9]*'`

exist-core/src/test/java/org/exist/xquery/value/DateTimeTest.java
```diff
@@ -426,4 +429,49 @@ public void minus3() throws XPathException {
...
+			fail("Expected FODT0001 for excessively large duration");
+		} catch (final XPathException ex) {
+			assertSame(ErrorCodes.FODT0001, ex.getErrorCode());
+		}
+		final long elapsedMs = (System.nanoTime() - start) / 1_000_000L;
+		assertTrue("Guard must reject in <1s; took " + elapsedMs + "ms", elapsedMs < 1000);
+	}
+
+	@Test
+	public void plus_largeButPermittedDayTimeDuration_works() throws XPathException {
```
reinhapa: Address Codacy issue `The JUnit 4 test method name 'plus_hugeDayTimeDuration_rejectsFast' doesn't match '[a-z][a-zA-Z0-9]*'`

reinhapa (changes requested):

line-o (commented):
I am in favour of this change and this limitation should be documented.

joewiz:

Done — renamed the three GH-5045 test methods to camelCase (`plusHugeDayTimeDurationRejectsFast`, `minusHugeDayTimeDurationRejectsFast`, `plusLargeButPermittedDayTimeDurationWorks`). CI rerunning.

line-o:
@joewiz could you resolve the conflict?

joewiz:

Thanks for the review, Juri. I've rebased this branch onto current `develop` to clear the conflict rather than carrying a merge commit.

The conflict was in `DurationValue.java`, in `equals()`. Since this PR was opened, `develop` replaced the old field-by-field `equals()` with a **value-based** one — it compares signed total months and signed total seconds, with a matching `hashCode()` so equal-valued `xs:duration` / `xs:yearMonthDuration` / `xs:dayTimeDuration` instances land in the same `op:same-key` bucket. This PR's `[refactor]` commit had independently rewritten the *old* `equals()` cosmetically (the PMD `SimplifyBooleanReturns` cleanup, short-circuit boolean form).

Resolution: I kept `develop`'s value-based `equals()`/`hashCode()` and dropped this PR's cosmetic `equals()` change — develop's version is semantically richer and supersedes it. I trimmed the PMD commit message accordingly (the `SimplifyBooleanReturns` bullet now lists only `durationsAreEqual()`, since `equals()` is no longer touched here). The rest of the PMD refactor and the FODT0001 bugfix itself applied cleanly on top of `develop`.

duncdrum (approved):

line-o (approved):

reinhapa (approved):

Issue #3114, opened 2019-11-15
Attribute range index produces incorrect results when used with ft:query.


paulmer:
### What is the problem
An XPath expression of the form

//record[ft:query(.//field[@attr="value"], <phrase>text</phrase>)

ignores the [@attr="value"] predicate if a range index is defined on @attr.  The result of the expression is equivalent to:

//record[ft:query(.//field, <phrase>text</phrase>)

If the range index is removed, the xpath works as expected.

### What did you expect

See above.

### Describe how to reproduce or add a test
> Describe how we can can reproduce the problem.

The attached test demonstrates the difference in behavior when a range index is and is not present.

[rft-test.zip](https://github.com/eXist-db/exist/files/3852818/rft-test.zip)

### Context information
Exist 5.1.0 snapshot built from source.  Confirmed on both Linux and Mac platforms, Java 1.8

joewiz:
@paulmer As a workaround, could you try this?

```xquery
//field[@attr="value"][ft:query(., <phrase>text</phrase>)]/ancestor::record
```

paulmer:
@joewiz That actually works, as does pretty much any variation I've found that doesn't include the '@attr' expression within the ft:query call, but alas, my real use cases are much more complex with the entire XPath expression being assembled from a number of pieces of configuration that can vary.  I suppose I'm being stubborn, but I'd rather tackle the bug than try to code around it.  I was kind of hoping someone would think "Aha! I know what's causing that.", but no such luck.

joewiz:
@paulmer My guess is that the predicate is being blown away in the query optimization surrounding ft:query - somewhere here: https://github.com/eXist-db/exist/blob/develop/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Query.java#L160-L177.

dizzzz:
you could try switching off the optimisation AFAIK;

```
declare option exist:optimize "enable=yes|no";
```

paulmer:
[...]

PR #6309, opened 2026-05-07
[bugfix] range:eq: per-item BOOLEAN-mode evaluation drops value comparison after preSelect

### LINKED ISSUE #6307: range:eq drops value comparison when value comparand is a for-bound variable
## Summary

When `@<attr>` has a configured Lucene-backed range index AND the value comparand is a **for-bound** variable, the predicate `//*[@<attr> = $v]` returns *every* element under the context per outer for-iter (one row per outer iter * scale), not just elements where `@<attr> = $v`. The literal-key form `//*[@<attr> = "X"]` is unaffected.

This is a high-severity correctness bug — it silently breaks every `for $key in $keys return //x[@indexed = $key]` join pattern over a corpus where the indexed attribute has a `<range>` block configured.

## Reproducer

Collection config:

```xml
<collection xmlns="http://exist-db.org/collection-config/1.0"
            xmlns:xs="http://www.w3.org/2001/XMLSchema"
            xmlns:xml="http://www.w3.org/XML/1998/namespace">
  <index>
    <range>
      <create qname="@xml:id" type="xs:string"/>
    </range>
  </index>
</collection>
```

Document (in the configured collection):

```xml
<root xmlns:xml="http://www.w3.org/XML/1998/namespace">
  <item xml:id="id0">v0</item>
  <item xml:id="id1">v1</item>
  ... (100 items total) ...
</root>
```

Query:

```xquery
for \$v in ("id0","id1","id2","id3","id4")
return collection("/db/test")//*[@xml:id = \$v]
```

**Expected**: 5 hits (one matching `<item>` per outer iter).
**Actual on develop**: 505 hits (`5 keys * (100 items + 1 root)` — the entire `*` set per outer iter).

The literal-key control `collection("/db/test")//*[@xml:id = "id0"]` returns 1 hit (correct).

## Mechanism

In `Lookup.eval`'s preselect-result branch (`extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/Lookup.java:446`), the path expression is evaluated against `contextSequence` (the full preloaded set) rather than against the per-item `effectiveContextSequence` (declared just above at line 392):

```java
result = getArgument(0).eval(contextSequence, null).toNodeSet();
```

When the predicate is dispatched in BOOLEAN mode (per-item), `contextSequence` stays the full 101-element preloaded set on every per-item call, so the inner step (`@xml:id`) returns the same non-empty NodeSet every call (the `@xml:id` of the single value-matching preloaded element). `effectiveBooleanValue` is `true` for every context item, so every element passes the predicate. The `\$v` binding is silently discarded.

The literal-key shape is unaffected because it stays in NODE-mode `selectByNodeSet`.

## Related

- #2204 / `d9b2d4d3c2 [bugfix] Fix false XPTY0004 from Optimizable index functions with for-bound variables` — fixed the WhereClause path. Same family but different code path.
- `718d5b02ed [bugfix] Fix index optimization regression for static outer variables` (on `bugfix/ngram-predicate-perf`) — sibling fix for ngram/ft. Different code path again.
- This issue is the third sibling: it lives in the predicate-Optimize-pragma path, specifically in `Lookup.eval`'s preselect-result branch.

## Sibling-bug audit

The other Optimizable index functions have been audited and do **not** reproduce this bug:

- `FieldLookup` (range:field-eq) — different selectAncestorDescendant path; targeted regression test passes.
- `NGramSearch` — reassigns `contextSequence = contextItem.toSequence()` at the top of `eval` (line 242-243), achieving the same effect as using `effectiveContextSequence`.
- Lucene `Query.eval` — same top-of-eval reassignment (line 293-295), plus a separate predicate-presence guard from #3114.
- Lucene `QueryField.eval` — same module pattern as `Query.eval`.

`Lookup.java` is the outlier: it declares `effectiveContextSequence` and uses it correctly in the fresh-eval branch (line 411, 417, 422) but the preselect-result branch at line 446 was missed.

## Fix

A one-line change in `Lookup.eval`'s preselect-result branch — substitute `effectiveContextSequence` and `contextItem` for `contextSequence` and `null`. PR forthcoming with regression test.

joewiz:
## Summary

Fixes a high-severity correctness bug in `range:eq` (and the `[@indexed = $v]` predicate it backs): when the value comparand is a **for-bound** variable AND the indexed attribute has a `<range>` block configured, every element under the context passes the predicate per outer for-iter, not just the matching ones. The literal-key form (`[@indexed = "X"]`) is unaffected.

Closes #6307.

## What changed

**`extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/Lookup.java`** — one-line fix in `eval`'s preselect-result branch (line 446). Substitutes `effectiveContextSequence` and `contextItem` for `contextSequence` and `null` in the `getArgument(0).eval(...)` call. The fresh-eval branch in the same method already uses `effectiveContextSequence`; the preselect-result branch had been missing it.

**`extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java`** — new JUnit regression test, runs by default. Two scenarios (legacy auto-index only; legacy + range index configured), each asserting both the literal-key and for-bound shapes return the expected hit counts. Fails on `develop` (505 hits where 5 expected); passes with this fix.

## Mechanism

In BOOLEAN-mode per-item predicate evaluation, the inner `Lookup.eval` is called once per context node with `contextItem` set. The pre-fix code evaluated the path expression against `contextSequence` (the full 101-element preloaded set), so every per-item call returned the same non-empty NodeSet (the `@xml:id` of the single value-matching preloaded element). `effectiveBooleanValue` was true for every element, silently dropping the `$v` binding.

The fix uses `effectiveContextSequence` (= `contextItem.toSequence()` when `contextItem` is set; otherwise = `contextSequence`). For the literal-key NODE-mode path, `contextItem == null` so the change is a no-op. For the for-bound BOOLEAN-mode path, the per-item context is now correctly threaded through.

The for-bound trace (captured via instrumentation, then reverted) shows 1 `preSelect` + 1 NODE-mode `eval` + 101 BOOLEAN-mode `eval(contextItem=set)` calls per outer iter; pre-fix every per-item call returned 1 item, post-fix only the matching item's per-item call returns non-empty.

## Sibling-bug audit

The other Optimizable index functions were audited and do **not** reproduce the same defect:

- **`FieldLookup.eval`** (range:field-eq) — different `selectAncestorDescendant` pattern; targeted regression test passes.
- **`NGramSearch.eval`** (ngram:contains) — reassigns `contextSequence = contextItem.toSequence()` at the top of `eval` (line 242-243), achieving the same effect as using `effectiveContextSequence`.
- **`Query.eval`** (lucene:query) — same top-of-eval reassignment (line 293-295), plus the #3114 predicate-presence guard.
- **`QueryField.eval`** (lucene:query-field) — same module pattern as `Query.eval`.

`Lookup.java` was the outlier. Fix is therefore scoped to `Lookup.java`.

## Related

- #2204 / `d9b2d4d3c2` (in develop) — fixed the WhereClause path. Same family, different code path.
- `718d5b02ed` on `bugfix/ngram-predicate-perf` — sibling fix for ngram/ft. Different code path again.
- This PR is the third sibling: predicate-Optimize-pragma path, `Lookup.eval` preselect-result branch.

## Test plan

- [x] Regression test `ForBoundXmlIdRegressionTest` fails on `develop`, passes with this commit.
- [x] Full range-index test suite (`mvn test -pl extensions/indexes/range`): **424 tests, 0 failures, 0 errors, 3 pre-existing skips**. Includes the existing `for-variable.xql` GH-2204 coverage.
- [ ] CI green on this PR (will appear after submit).
- [ ] No QT4 / XQTS 3.1 regressions vs `develop` (the rewrite-to-Lookup path is gated on a `<range>` block being configured, and the change is a no-op for the literal-key NODE-mode path).

## Risk

Low. The change matches the pattern used in the same method's fresh-eval branch (lines 411, 417, 422). For NODE-mode (literal-key) paths the change is a no-op because `effectiveContextSequence == contextSequence` when `contextItem == null`. The 422 existing range-index tests (XQSuite + JUnit) all pass, including configured-range tests, field tests, and for-variable-from-#2204 tests.

### DIFF
--- a/extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/Lookup.java
+++ b/extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/Lookup.java
@@ -442,11 +442,10 @@ public class Lookup extends Function implements Optimizable, IndexUseReporter {
 //            LOG.info("eval plain took " + (System.currentTimeMillis() - start));
         } else {
             contextStep.setPreloadedData(preselectResult.getDocumentSet(), preselectResult);
-            // Use effectiveContextSequence (the per-item context when contextItem is set) so
-            // BOOLEAN-mode per-item predicate evaluation actually checks the per-item node.
-            // With contextSequence (the full preloaded set), every per-item call returned the
-            // same non-empty NodeSet, silently dropping the value comparison from a for-bound
-            // variable: //x[@indexed = $v] over-returned by ~scale factor for each outer iter.
+            // Use effectiveContextSequence instead of contextSequence here. With contextSequence
+            // (the full preloaded set), a lookup inside a predicate would "see" the entire nodeSet
+            // for each iteration in a query like //x[@indexed-attribute = $v] and return duplicate
+            // hits.
             result = getArgument(0).eval(effectiveContextSequence, contextItem).toNodeSet();
         }
         return result;
--- a/extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/Lookup.java
+++ b/extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/Lookup.java
@@ -441,10 +441,13 @@ public class Lookup extends Function implements Optimizable, IndexUseReporter {
             }
 //            LOG.info("eval plain took " + (System.currentTimeMillis() - start));
         } else {
-//            long start = System.currentTimeMillis();
             contextStep.setPreloadedData(preselectResult.getDocumentSet(), preselectResult);
-            result = getArgument(0).eval(contextSequence, null).toNodeSet();
-            //LOG.info("eval took " + (System.currentTimeMillis() - start));
+            // Use effectiveContextSequence (the per-item context when contextItem is set) so
+            // BOOLEAN-mode per-item predicate evaluation actually checks the per-item node.
+            // With contextSequence (the full preloaded set), every per-item call returned the
+            // same non-empty NodeSet, silently dropping the value comparison from a for-bound
+            // variable: //x[@indexed = $v] over-returned by ~scale factor for each outer iter.
+            result = getArgument(0).eval(effectiveContextSequence, contextItem).toNodeSet();
         }
         return result;
     }
--- a/extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java
+++ b/extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java
@@ -122,12 +122,11 @@ public class ForBoundXmlIdRegressionTest {
      * @return {@code [literalHits, forBoundHits]}
      */
     private long[] runScenario(final XmldbURI col, final int nKeys) throws Exception {
-        final String keysExpr = "(for $i in 0 to %d return \"id\" || $i)".formatted(nKeys - 1);
         final String literal = """
                 collection("%s")//*[@xml:id = "id0"]""".formatted(col);
         final String forBound = """
-                for $v in %s
-                return collection("%s")//*[@xml:id = $v]""".formatted(keysExpr, col);
+                for $v in (for $i in 0 to %d return "id" || $i)
+                return collection("%s")//*[@xml:id = $v]""".formatted(nKeys - 1, col);
 
         final BrokerPool pool = existEmbeddedServer.getBrokerPool();
         try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()))) {
--- a/extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java
+++ b/extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java
@@ -37,6 +37,8 @@ import org.exist.xquery.XPathException;
 import org.exist.xquery.XQuery;
 import org.exist.xquery.XQueryContext;
 import org.exist.xquery.value.Sequence;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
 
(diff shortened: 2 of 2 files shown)


extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java
```diff
@@ -0,0 +1,168 @@
...
+     * @return {@code [literalHits, forBoundHits]}
+     */
+    private long[] runScenario(final String tag, final String configXml, final int nKeys) throws Exception {
+        final BrokerPool pool = existEmbeddedServer.getBrokerPool();
+        final XmldbURI col = XmldbURI.ROOT_COLLECTION_URI.append("forbound-xmlid-" + tag);
+        try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()))) {
+            createCollectionAndStore(pool, broker, col, configXml);
+        }
+
+        final String keysExpr = IntStream.range(0, nKeys)
```
line-o: Unnecessary complicated code to construct a sequence of keys for testing.

extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * @return {@code [literalHits, forBoundHits]}
+     */
+    private long[] runScenario(final String tag, final String configXml, final int nKeys) throws Exception {
+        final BrokerPool pool = existEmbeddedServer.getBrokerPool();
+        final XmldbURI col = XmldbURI.ROOT_COLLECTION_URI.append("forbound-xmlid-" + tag);
+        try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()))) {
+            createCollectionAndStore(pool, broker, col, configXml);
```
line-o: Shouldn't creating collections and adding collection configurations be better part of a setup method?

extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        final String keysExpr = IntStream.range(0, nKeys)
+                .mapToObj(i -> "\"id" + i + "\"")
+                .collect(Collectors.joining(",", "(", ")"));
+
+        final String literal = "collection(\"" + col + "\")//*[@xml:id = \"id0\"]";
+        final String forBound = "for $v in " + keysExpr
```
line-o: Below a formatted textblock is used. This can be done here as well.

joewiz:

Thanks for the review — addressed all three comments:

- Collection creation + xconf configuration moved to a `@BeforeClass` setup method (with matching `@AfterClass` teardown), so the `@Test` bodies are focused on running the query under test.
- Simplified the key-sequence construction: replaced the `IntStream`/`Collectors` pipeline with a plain XQuery `for $i in 0 to N return "id" || $i`.
- Converted the literal / for-bound query strings to Java text blocks via `String.formatted`, matching the convention already used elsewhere in the file.

CI rerunning.

extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/Lookup.java
```diff
@@ -441,10 +441,13 @@ public Sequence eval(Sequence contextSequence, Item contextItem) throws XPathExc
             }
 //            LOG.info("eval plain took " + (System.currentTimeMillis() - start));
         } else {
-//            long start = System.currentTimeMillis();
             contextStep.setPreloadedData(preselectResult.getDocumentSet(), preselectResult);
-            result = getArgument(0).eval(contextSequence, null).toNodeSet();
-            //LOG.info("eval took " + (System.currentTimeMillis() - start));
+            // Use effectiveContextSequence (the per-item context when contextItem is set) so
```
line-o: This comment could be shortened to 

```
// Use effectiveContextSequence instead of contextSequence here. With contextSequence
// (the full preloaded set), a lookup inside a predicate would "see" the entire nodeSet for each
// iteration in a query like //x[@indexed-attribute = $v] and return duplicate hits

File: extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java [hunk 5]
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * @return {@code [literalHits, forBoundHits]}
+     */
+    private long[] runScenario(final XmldbURI col, final int nKeys) throws Exception {
+        final String keysExpr = "(for $i in 0 to %d return \"id\" || $i)".formatted(nKeys - 1);
+        final String literal = """
+                collection("%s")//*[@xml:id = "id0"]""".formatted(col);
+        final String forBound = """
+                for $v in %s
+                return collection("%s")//*[@xml:id = $v]""".formatted(keysExpr, col);
```
line-o: I think inlining `keysExpr` into the `forBound` string would help to quicker understand what is being tested.

line-o (approved):
I have some suggestions. Not blocking the merge.

extensions/indexes/range/src/main/java/org/exist/xquery/modules/range/Lookup.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `6c371a1e22` — applied the suggestion as-written. Thanks!

extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/ForBoundXmlIdRegressionTest.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `6c371a1e22` — applied the suggestion as-written. Thanks!

duncdrum (approved):

PR #6312, opened 2026-05-07
[bugfix] Optimizer: skip exist:optimize wrap for constant-false predicates

### LINKED ISSUE #3918: [BUG] Strange performance issue - 10-10,000x slower
**Describe the bug**

A query that should return an empty sequence almost instantly (.008s) takes 10x longer (.08s) on a fresh database, and 10,000x longer (80s) on a large database:

```xquery
collection("/db/apps")//foo[bar = "baz"][contains("abc", "123")]
```

Removing either predicate causes the query to return almost instantly. 

The order of the predicates doesn't matter. 

Using the `no-index` pragma to prevent eXist from using any index causes the query to perform as expected, even on a large database:

```xquery
(# exist:no-index #) {
    collection("/db/apps")//foo[bar = "baz"][contains("abc", "123")]
}
```

**Expected behavior**

I expect this query which should return 0 results to return almost instantly, since (a) there are no `<foo>` elements in the database, (b) there are no `<bar>` elements, and (c) the `contains` function here should always evaluate to `false()`.

**To Reproduce**

See the queries above. The large database I described contains a full [hsg-project system](https://github.com/HistoryAtState/hsg-project/wiki/setup) and many additional collections - with an eXist data directory weighing in at 20 GB.

 - OS: macOS 11.4
 - eXist-db version: Tested with fresh installations of 4.7.1, 5.2.0, and 5.3.0-SNAPSHOT, as well as large database of eXist 5.3.0-SNAPSHOT e371efd9987a9a2f4414839c7bf1dbc20107b6d1 20210604033555
 - Java Version: OpenJDK 1.8.0_292 (liberica-jdk-full)

**Additional context**
- How is eXist-db installed? Installed stable versions via DMG, snapshot via source

joewiz:
## Summary

Fixes [#3918](https://github.com/eXist-db/exist/issues/3918): a compound predicate of the form `path[indexed-pred][contains(literal, literal)]` runs 10-10,000x slower than the same query without the constant predicate. The `(# exist:no-index #)` workaround restored normal speed in the issue, proving the slowdown comes from the optimizer wrapping the step in `(# exist:optimize #)` and driving an index pre-select for a query whose result is statically empty.

## Diagnosis

Instrumenting `Optimizer.visitLocationStep` confirmed the wrap fires on the bad shape. The reproducer's two predicates report:

| predicate | `getDependencies()` |
|---|---|
| `[child::bar = "999"]` | `[CONTEXT_SET]` |
| `[contains("abc", "123")]` | `[CONTEXT_SET \| CONTEXT_ITEM]` |

The constant predicate wrongly reports `CONTEXT_ITEM` because `Function.getDependencies()` returns `CONTEXT_ITEM | CONTEXT_SET` unconditionally. The dependency bitmask is therefore unreliable for builtin functions on literal arguments — a structural check is required.

## What changed

- **`Optimizer.java`** — new `ContextFreeChecker` (subclass of `DefaultExpressionVisitor`) walks the predicate's expression tree and flags it as context-dependent on encountering `LocationStep`, `VariableReference`, `FunctionCall`, `UserDefinedFunction`, FLWOR bindings (`for`/`let`/window), or any builtin implementing `Optimizable`. The `Optimizable` exclusion matters: `range:eq`, `ft:query-field` and friends look literal-only but rely on the wrap for their index pre-select, so the gate must never strip a wrap whose whole purpose is to feed an Optimizable predicate.
- **`Optimizer.foldsToFalse(Predicate)`** — when the structural check passes, calls `pred.preprocess()` to evaluate the predicate at compile time. Returns `true` only if the value's effective-boolean is false. Wrapped in try/catch so any failure reverts to the no-fold default.
- **`Optimizer.hasConstantFalsePredicate(...)`** — single-pass check used by `visitLocationStep` and `visitFilteredExpr` before wrapping.
- **`Optimizer.shouldWrapWithOptimizePragma(...)`** — extracted helper combining `findOptimizable` with the new gate. Keeps `visitLocationStep`'s NPath complexity below the PMD threshold of 200.
- **`ContextFreePredicateRegressionTest`** — seven cases covering the issue reproducer, `[false()]`, `[1 = 0]`, `[true()]` (must NOT strip wrap), the bare indexed-predicate baseline, and a variable-referencing predicate (must NOT trigger gate).

## Test plan

- [x] `ContextFreePredicateRegressionTest` — 7 pass
- [x] `exist-core` full module test suite — 6676 pass (105 skipped, 0 failures)
- [x] `extensions/indexes/lucene` — 659 pass (40 skipped, 0 failures)
- [x] `extensions/indexes/range` — 428 pass (3 skipped, 0 failures)
- [x] Codacy PMD on `Optimizer.java`: only pre-existing findings (`reverseAxis` unused, empty method body), no new warnings introduced
- [x] `mvn license:check -pl exist-core` clean

## Notes on scope

The audit's hypothesised fix (skip when `CONTEXT_ITEM` bit is unset) does not work because the conservative default in `Function.getDependencies()` always sets that bit. The structural visitor + compile-time fold is the smallest reliable signal short of M2 generic constant folding. That broader work remains a follow-up; this PR ships the narrow case that closes #3918.

Closes https://github.com/eXist-db/exist/issues/3918

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/Optimizer.java
+++ b/exist-core/src/main/java/org/exist/xquery/Optimizer.java
@@ -24,6 +24,7 @@ package org.exist.xquery;
 import org.exist.storage.DBBroker;
 import org.exist.xquery.functions.array.ArrayConstructor;
 import org.exist.xquery.pragmas.Optimize;
+import org.exist.xquery.value.Sequence;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.exist.xquery.util.ExpressionDumper;
@@ -93,25 +94,8 @@ public class Optimizer extends DefaultExpressionVisitor {
             LOG.warn("Exception called while rewriting location step: {}", e.getMessage(), e);
         }
 
-        boolean optimize = false;
-        // only location steps with predicates can be optimized:
         @Nullable final Predicate[] preds = locationStep.getPredicates();
-        if (preds != null) {
-            // walk through the predicates attached to the current location step.
-            // try to find a predicate containing an expression which is an instance
-            // of Optimizable.
-            for (final Predicate pred : preds) {
-                pred.accept(findOptimizable);
-                @Nullable final Optimizable[] list = findOptimizable.getOptimizables();
-                if (canOptimize(list)) {
-                    optimize = true;
-                }
-                findOptimizable.reset();
-                if (optimize) {
-                    break;
-                }
-            }
-        }
+        final boolean optimize = shouldWrapWithOptimizePragma(preds);
 
         final Expression parent = locationStep.getParentExpression();
 
@@ -135,7 +119,7 @@ public class Optimizer extends DefaultExpressionVisitor {
                 }
                 extension.addPragma(new Optimize(extension, context, Optimize.OPTIMIZE_PRAGMA, null, false));
                 extension.setExpression(locationStep);
-                
+
                 // Replace the old expression with the pragma
                 path.replace(locationStep, extension);
 
@@ -234,7 +218,7 @@ public class Optimizer extends DefaultExpressionVisitor {
 
         // check if there are any predicates which could be optimized
         final List<Predicate> preds = filtered.getPredicates();
-        final boolean optimize = hasOptimizable(preds);
+        final boolean optimize = hasOptimizable(preds) && !hasConstantFalsePredicate(preds);
         if (optimize) {
             // we found at least one Optimizable. Rewrite the whole expression and
             // enclose it in an (#exist:optimize#) pragma.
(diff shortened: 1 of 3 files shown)


exist-core/src/main/java/org/exist/xquery/Optimizer.java
```diff
@@ -283,6 +267,138 @@ private boolean hasOptimizable(final List<Predicate> preds) {
...
+        }
+    }
+
+    /**
+     * Marks an expression tree as context-dependent if it references the
+     * context node, a variable, a user-defined function, or a FLWOR binding.
+     * Walks builtin function arguments and operator operands so that nested
+     * occurrences are caught.
+     */
+    private static final class ContextFreeChecker extends DefaultExpressionVisitor {
```
line-o: The naming clashes for me with context free grammars. Maybe we can find a more fitting one?

`ContextDependencyChecker` for example.

exist-core/src/test/java/org/exist/xquery/ContextFreePredicateRegressionTest.java
```diff
@@ -0,0 +1,209 @@
...
+ * <p>The fix in {@link Optimizer} skips the optimize-pragma wrap when one
+ * predicate is structurally context-free and folds to effective-boolean
+ * false at compile time. The tests below assert correctness for several
+ * shapes of constant predicate; the wrap-skip behavior itself is asserted
+ * indirectly by comparing optimized and {@code (# exist:no-index #)}
+ * results, which must match.
+ *
+ * @see <a href="https://github.com/eXist-db/exist/issues/3918">GH-3918</a>
+ */
+public class ContextFreePredicateRegressionTest {
```
line-o: The same clash in meaning as before. I would prefer this to be renamed to `ContextDependencyRegressionTest`

line-o (commented):
I think this is a leap in the right direction.

I would like us to take a moment to think about the naming (see comments) but other than that this will make some special cases perform much better.

Having written this, what happens when the predicate will evaluate to false at runtime?

```xquery
for $needle in ('a', 'b', 'c')
return //foo[bar = '5'][contains($needle, 'c')]
```

We would have to consider _unfolding_ the above into

```xquery
let $needle := 'a'
return //foo[bar = '5'][contains($needle, 'c')], (: evaluates to an empty sequence :)
let $needle := 'b'
return //foo[bar = '5'][contains($needle, 'c')], (: evaluates to an empty sequence :)
let $needle := 'c',
return //foo[bar = '5'][contains($needle, 'c')]
```

Ah, I see that [constant propagation and folding](https://en.wikipedia.org/wiki/Constant_folding) was explicitly out of scope for this PR

> short of M2 generic constant folding

line-o:
I am still wondering: User-defined functions _cannot_ depend on the context item as this is defined _absent_ by the XQuery spec.
That means that if there is no expression _passed in_ to the function as a parameter the entire predicate _must_ be context set _and_ context item independent.

joewiz:

Done — renamed `ContextFreeChecker` → `ContextDependencyChecker` and `ContextFreePredicateRegressionTest` → `ContextDependencyRegressionTest`. Agreed the original name collided with the formal-language sense of "context-free"; the new names are clearer.

While addressing the rename I also looked at a related concern: no-arg context-dependent builtins like `fn:position()`, `fn:last()`, `fn:name()`, and `fn:local-name()` were structurally classified as "context-free" by the visitor because they have no subexpressions to flag. The runtime `preprocess()` safety net masks this (those builtins throw `XPDY0002` on null context, the catch returns `false`, the wrap fires) but the structural classification was technically wrong, which felt fragile for any future builtin that might silently misevaluate.

The visitor now consults `Function.getDependencies()` for zero-arg builtins specifically. A no-arg builtin has no input besides the dynamic context, so any declared `CONTEXT_*` bit is real. `fn:true()` / `fn:false()` / `fn:current-dateTime()` declare `NO_DEPENDENCY` and stay context-free; `fn:position()` / `fn:last()` declare `CONTEXT_*` and are correctly flagged. The conservative `Function` default (`CONTEXT_ITEM | CONTEXT_SET`) covers `fn:name()` / `fn:local-name()` correctly too. New regression tests for position / last / name / local-name confirm per-context evaluation remains correct.

Regarding your runtime-folding-of-FLWOR-bound-variables thought experiment: yes, that would be generic constant folding / propagation (the audit's M2 line), and it's explicitly out of scope here. Worth a separate follow-up but won't try to fold it into this PR.

Both commits pushed; CI rerunning.

line-o:
Agreed.
For the future in XPath, XSLT and XQuery the term `context independent` is preferrable to `context-free` as this is also used within the specification of those langauges.

line-o:
I had another deeper look and wonder if we are trading massive performance gains for a very special case with a measurable impact on all other queries. 
How expensive is a call to `Optimizer.hasConstantFalsePredicate` for a query with 1 to 3 predicates where each is calling a function?

Do we have performance benchmark that we can use to falsify my assumption?

joewiz:

@line-o — fair concern, measured it. tl;dr the gate is in the noise.

**Setup.** Microbenchmark that compiles each query 200x warmup + 2,000x measured, instrumenting `hasConstantFalsePredicate` and `findOptimizable.accept(pred)` with `System.nanoTime()` counters. `findOptimizable` is the natural baseline -- it's the per-predicate AST walk that already existed before this PR, with the same shape (visitor over the predicate tree) as the new `ContextDependencyChecker`. Java 21 / Zulu 21.38, embedded server, single-threaded. Instrumentation was added locally and reverted before push.

**Per-call gate cost + end-to-end overhead:**

| query | compile (µs) | gate calls | gate avg (ns) | gate total (ns) | findopt total (ns) | gate % of compile | preprocess() invocations |
|---|---:|---:|---:|---:|---:|---:|---:|
| `//foo[bar = 'x']` | 145 | 2000 | **489** | 977,670 | 459,065 | **0.34%** | 0 |
| `//foo[bar = 'x'][@id = 'y']` | 150 | 2000 | 199 | 397,624 | 113,859 | 0.13% | 0 |
| `//foo[contains(bar,'x')][starts-with(@id,'y')][string-length(baz) > 3]` | 324 | 2000 | 339 | 678,644 | 116,108 | 0.10% | 0 |
| `//foo[contains('hello', 'world')]` (literal-only fn) | 119 | **0** | -- | 0 | 154,941 | 0.00% | 0 |
| `//doc/section[para[contains(@type,'note')]]/title[. = 'X']` | 250 | 4000 | 48 | 191,300 | 202,563 | 0.04% | 0 |
| `for $x in //foo[bar = 'x'] return $x/@id` | 152 | 2000 | 44 | 88,896 | 54,482 | 0.03% | 0 |
| `//foo/bar/baz` (no predicates) | 88 | 0 | -- | 0 | 0 | 0.00% | 0 |
| `//foo[contains('a','b')][starts-with('c','d')][string-length('e')=1]` (3 literal-only fn preds) | 240 | **0** | -- | 0 | 135,745 | 0.00% | 0 |

**What the numbers say.**

1. **Per-call cost is sub-microsecond, 44–489 ns.** Tasking threshold for "defend" was <2 µs; we're 4–45x under it. Worst case is the simplest query (single predicate with a single nested LocationStep), where the visitor walks a few nodes before hitting the LocationStep early-exit. Multi-predicate queries are *cheaper per call* on average because the second-and-later predicates short-circuit faster.
2. **End-to-end overhead is 0.03–0.34% of compile time** -- well under the 5% "defend" threshold.
3. **`hasConstantFalsePredicate` never runs when the wrap wouldn't have fired anyway.** Q4 / Q7 / Q8 have zero gate calls. The gate is short-circuited by `optimizable && !hasConstantFalsePredicate(...)` -- the `&&` skips the gate when nothing was index-eligible. For the literal-only-fn-pred case (Q4, Q8) `findOptimizable` correctly reports nothing optimizable, and the gate is skipped. So queries with no Optimizable predicates pay zero gate cost.
4. **`preprocess()` was never invoked across 16,000 query compilations.** Realistic predicates touch the context node (a `LocationStep`, `@id`, `.`, etc.), so the structural check in `ContextDependencyChecker` bails fast and the expensive compile-time-evaluate path stays cold. The preprocess() cost only materialises for the GH-3918 shape itself -- predicates whose *entire* AST is literal-or-builtin-of-literal -- and even then the gate's whole purpose is to avoid the much more expensive index-pre-select-then-discard the wrap would have triggered.
5. **vs. the baseline:** on Q1 the gate adds ~978 µs across 2,000 compilations, against a 145 µs/compile budget = ~0.34% added. `findOptimizable` runs on the same predicates and contributes ~459 µs total -- so the gate adds about 2x the per-call cost of the AST walk that already existed, both still vanishingly small relative to compile time.

**Conclusion.** Defending the current design. The gate's per-call cost is sub-microsecond and only fires when the wrap was about to be added anyway. On compilation-bound queries it's 0.03–0.34% overhead, on actual queries (where compile time is amortised over execution) it's effectively zero.

**Why this matches the structural argument:**

- The gate is short-circuited by `optimizable && !hasConstantFalsePredicate(...)`. If no predicate is index-eligible, the gate doesn't run.
- `ContextDependencyChecker` is a `DefaultExpressionVisitor` with an early-return guard at every visit method. It bails on the first `LocationStep`/`VariableReference`/`FunctionCall`/`ForExpr`/`LetExpr`/`WindowExpr` it sees -- which is the very first thing it sees in any realistic predicate.
- `pred.preprocess()` only runs when the structural check passes -- i.e., the predicate is literally constant. In the 16,000 compilations measured here, that path was cold every time.

Re: naming -- agreed, will rename to `ContextIndependent*` in a separate follow-up PR per your "for the future" comment, since that touches files that aren't part of the GH-3918 fix.

Bench harness was a temporary `Optimizer` instrumentation patch + a JUnit `OptimizerGateMicrobenchmark` test. Both reverted; nothing pushed beyond the existing two commits.

line-o (approved):
Thanks for checkin my concerns.

reinhapa (approved):

PR #6318, opened 2026-05-08
[optimize] util:expand: cache Lucene terms per query, skip empty-termMap scans

joewiz:
## Summary

util:expand with match-highlighting on full-text hits has been substantially slower than util:expand with highlighting off. The reporter's bench (#5738) showed `~1052 ms` vs `~2.7 ms` for a single hit, with the gap widening to several seconds for batch `util:expand($hits)` over a wildcard-matched corpus.

This PR caches the rewritten Lucene terms per query so batch util:expand reuses term-rewrite work, and short-circuits scanMatches when the configured-field exclusion (PR #3467) would produce an empty termMap.

Refs https://github.com/eXist-db/exist/issues/5738

## Scope of this PR

This PR delivers a measurable speedup on two paths:
- **Batch `util:expand($hits)` over wildcard-matched corpora**: ~4.6× speedup at 5000 TEI-entry scale (1441 ms → 312 ms) by caching rewritten Lucene terms per query identity, so the term-rewrite cost is paid once per query instead of once per hit.
- **Named-field queries (e.g., `lemma:Aachen`)**: scanMatches is short-circuited entirely when PR #3467's configured-field exclusion produces an empty termMap. For this query shape the highlighting path is effectively eliminated.

The reporter's BenchmarkDotNet harness in the issue thread documented a ~480× ratio between `ExpandOnly` (highlighting on) and `ExpandSkipXIncludeHighlightNone` at scale. This PR's 4.6× speedup on a synthetic wildcard reproducer doesn't fully close that ratio. Two factors:

1. **Different query shape**: the synthetic reproducer here uses a `lemma:a*` wildcard (term-rewrite-bound). The reporter's `lemma:Aachen` named-field query is a different bottleneck — but one that this PR's empty-termMap short-circuit *should* eliminate for that specific shape. Validation against the reporter's actual corpus (50MB single doc, ~50M / ~200M XML collections) is pending.
2. **Residual per-call overhead**: the reporter's environment (eXist 6.4.0 in Docker, HTTP) shows much higher per-call overhead than this isolated JUnit reproducer. The cache wins should compound there since each `reset()` no longer reopens the IndexReader, but the absolute baseline is higher.

So this PR is `Refs #5738` rather than `Closes` — leaves the issue open until the reporter (or another user with a comparable corpus) confirms the gap is closed at deployment scale. If the named-field-query path with empty termMap covers the reporter's actual use case, the issue can be closed in a follow-up. If a wildcard-over-paragraphs workload remains slow, the architectural next step is opt-in SAX-serialize-and-compare (analogous to xmldiff:compare's path) — bigger refactor, separate design pass.

## Diagnosis

`LuceneMatchListener.reset()` runs once per node passed through the serializer chain. Two costs were repeated unnecessarily on every reset():

1. `getTerms()` reopened the IndexReader and re-rewrote each query's terms, even when the prior `reset()` had just done the same work. For wildcard and prefix queries on large indexes, term enumeration dominates.
2. `scanMatches()` walked the entry's XML stream and tokenized its text **even when termMap was empty** — the typical case for named-field queries like `lemma:Aachen` after PR #3467's configured-field exclusion. Every token lookup returned null but the work was done anyway.

## What changed

`extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneMatchListener.java`:

- Cache rewritten `rawTerms` per Lucene `Query` identity in a bounded LRU (max 32 entries). Lucene's `Query.equals` is content-based, so semantically equal queries share a cache slot. Batch `util:expand($hits)` where every hit shares one ft:query now pays the term-rewrite cost once, not per hit.
- After `getTerms()` builds `termMap`, short-circuit `reset()` and `startElement()` when `termMap` is empty. No token can match, `nodesWithMatch` stays empty, and `characters()` always passes through unchanged — so SAX events flow without any deep scanning work.
- Refactored `getTerms()` into three smaller methods: `getTerms()`, `collectUniqueLuceneQueries()`, `buildTermMap()`.

`extensions/indexes/lucene/src/test/java/org/exist/indexing/lucene/UtilExpandHighlightingPerformanceTest.java` (new):

5 tests:
- `namedFieldQueryProducesNoHighlights` — verifies PR #3467 behavior is preserved
- `highlightingStillWorksForImplicitFieldQuery` — correctness check that exist:match wrappers still appear
- `batchExpandMatchesPerHitForLoop` — `util:expand($hits)` matches the per-hit `for $h in $hits return util:expand($h)` count
- `singleHitHighlightingNotOrderOfMagnitudeSlower` — ratio < 20x for highlighted vs unhighlighted single-hit expand
- `batchWildcardExpandRatioUnderThreshold` — ratio < 15x for batch wildcard expand

## Empirical numbers

Isolated reproducer (5000 TEI-style entries, lemma:a* wildcard returns ~2500 hits, single-doc):

| Variant | Before | After | Speedup |
|---|---|---|---|
| `util:expand($hits)` batch over ~2500 wildcard hits | ~1441 ms | ~312 ms | ~4.6x |
| Single-hit `util:expand(subsequence($hits,1,1))` | ~15 ms | ~15 ms | (already fast in this env) |
| Named-field `lemma:Aachen` style queries (any size) | scanMatches always runs | scanMatches skipped when termMap empty | depends on entry size |

The reporter's bench environment (50MB single doc, eXist 6.4.0 in Docker, HTTP) shows much higher per-call overhead than this isolated JUnit reproducer; the relative wins should compound there since the cache eliminates repeated `IndexReader` acquisition + term enumeration on every `reset()`.

## Why this matters

KWIC-style match highlighting is a default UX path for any eXist app that does full-text search. The slowdown is on the order of hundreds of × for batch expand. Real-world impact on application owners is large — "search results render second-by-second instead of all-at-once" is the kind of perceptible bug that drives users to other engines. The inventory's mid-tier numeric score for #5738 doesn't weight "appears in every search-result page" prominently enough; this issue affects the median deployed eXist app, not just edge cases.

## Test plan

- [x] `mvn test -pl extensions/indexes/lucene` — 661 tests pass (no regressions)
- [x] `mvn test -pl extensions/indexes/lucene -Dtest=UtilExpandHighlightingPerformanceTest` — 5 new tests pass
- [x] `mvn test -pl extensions/indexes/lucene -Dtest=LuceneMatchListenerTest,LuceneIndexTest,SerializeAttrMatchesTest` — pass
- [x] `mvn test -pl extensions/indexes/ngram` — 39 tests pass
- [x] `mvn test -pl exist-core -Dtest=ExpandTest` — 5 tests pass
- [x] `codacy-cli analyze --tool pmd` on touched file — only the pre-existing `scanMatches` NPath finding (unchanged)
- [x] `mvn license:check -pl extensions/indexes/lucene` — clean
- [ ] CI green on PR


extensions/indexes/lucene/src/test/java/org/exist/indexing/lucene/UtilExpandHighlightingPerformanceTest.java
duncdrum, line 1: I don't see a good reason to keep this TEI flavoured within a testing context, it just adds boilerplate namespace declaratations, prefixes etc making the code less legible.

extensions/indexes/lucene/src/test/java/org/exist/indexing/lucene/UtilExpandHighlightingPerformanceTest.java
```diff
@@ -0,0 +1,306 @@
...
+                    "return count(util:expand($hits)//exist:match)", null);
+            assertEquals(1, batch.getItemCount());
+            final int batchCount = batch.itemAt(0).toJavaObject(Integer.class);
+
+            assertEquals("Batch util:expand should produce the same exist:match count as the per-hit for-loop",
+                    forLoopCount, batchCount);
+            assertTrue("Should produce at least one exist:match for 'aword42'; got " + batchCount,
+                    batchCount >= 1);
+        }
+    }
```
duncdrum: the test cases up to here, seem to partially duplicate existing xqsuite tests. I don't see the advantage of having these in java form, why did you choose surefire over XQsuite?

extensions/indexes/lucene/src/test/java/org/exist/indexing/lucene/UtilExpandHighlightingPerformanceTest.java
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * Performance threshold: util:expand on a single hit with highlighting on must not
+     * be orders of magnitude slower than the same call with highlighting off. Pre-fix,
+     * the ratio was 50x-500x on the reporter's data; post-fix it is ~1-3x. The 20x
+     * threshold leaves comfortable headroom for noisy CI runners.
+     */
+    @Test
```
duncdrum: these test on the other hand, should be jmh benchmarks, no?

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneMatchListener.java
```diff
@@ -71,6 +71,20 @@ public class LuceneMatchListener extends AbstractMatchListener {
...
+     * queries share an entry) and is bounded to avoid unbounded growth across long-lived
+     * brokers. */
+    private static final int QUERY_TERM_CACHE_MAX = 32;
+    private final LinkedHashMap<Query, Map<Object, Query>> queryTermCache =
+            new LinkedHashMap<>(QUERY_TERM_CACHE_MAX, 0.75f, true) {
+                @Override
+                protected boolean removeEldestEntry(final Map.Entry<Query, Map<Object, Query>> eldest) {
+                    return size() > QUERY_TERM_CACHE_MAX;
+                }
+            };
```
reinhapa: Avoid writing poor man cache logic here. We should use a dedicated cache with configurable size as provided by the already used `Caffeine` cache.

reinhapa (changes requested):

duncdrum (commented):
The fix seems sound and desirable, some questions about the test design though.

joewiz:

Re your test-design questions, @duncdrum:

**#2 (XQSuite vs surefire)**: agree XQSuite is the established eXist pattern for XQuery-behavior tests. I went with surefire here because the load-bearing claim is "this fix is faster AND produces equivalent output", and mixing correctness assertions with timing assertions in one cohesive class made the tradeoff visible. XQSuite doesn't natively express timing thresholds.

If you'd prefer, I can split the correctness portion into XQSuite (`.xqm` with `%test:assertEquals`) and keep the perf-threshold piece in surefire — that's the cleaner shape, just adds two files and splits tests across the boundary. Happy to do that as a follow-up commit on this PR if you'd rather not merge until then; otherwise as a follow-up PR after this one lands.

**#3 (JMH benchmarks)**: agree JMH is the right tool for perf benchmarking — proper warmup, statistical handling, etc. The blocker is that `exist-indexes-jmh` isn't on develop yet (it's PR #6296, currently open). Once that lands, the perf-threshold tests in this PR should move to JMH benchmarks in that module; happy to file a follow-up issue tracking that move.

**#1 (TEI noise)**: addressed in cde2465dcd — corpus is now a minimal `<dict>/<entry>` shape with no namespace ceremony.

I've also pushed 35e90d160c to swap the LRU for a Caffeine cache per @reinhapa's review.

Let me know your preference on splitting #2 into XQSuite + surefire — happy to do it before merge if that's the cleaner exit.

extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneMatchListener.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in 35e90d160c — switched to Caffeine cache with the same 32-entry maximum size. Behavior unchanged; gain configurable bounds, stats, and proper concurrency vs the hand-rolled LRU. Thanks.

reinhapa (approved):

duncdrum:
@joewiz yes please split the spec. Add xqsuite where the query shape is not already covered by existing tests. 

Please open an issue with the jmh migration. While there is no index-benchmark reactor here, there is a core-benchmark reactor. 

The reactor is less important to me then the proper test type.

duncdrum:
@joewiz can you rebase and do the split?

joewiz:

Apologies for the long delay @duncdrum — your review prompts from May 8 (and the re-ping this morning) didn't surface until now. All three points addressed in `b4458f8e68`:

| # | Ask | Resolution |
|---|---|---|
| 1 | Drop TEI-flavoured namespace boilerplate | Done — corpus now uses a generic `dict/entry` shape with no namespaces, just the minimal vocabulary the test needs (`entry`, `form`, `orth`, `sense`, `def`). |
| 2 | Migrate correctness tests to xqsuite | Done — new `extensions/indexes/lucene/src/test/xquery/lucene/ft-expand-highlight.xqm` carrying the 3 retained correctness tests, runnable via the existing `xquery.lucene.LuceneTests` launcher. `UtilExpandHighlightingPerformanceTest.java` deleted entirely. |
| 3 | Perf-ratio tests → JMH benchmarks | Removed from this PR. Filed [#6387](https://github.com/eXist-db/exist/issues/6387) for the migration: needs an `extensions/indexes` benchmark reactor (mirror of `exist-core-jmh/`), parameterised with a Lucene-indexed corpus fixture, with `@Benchmark`s for the singleHit + batch shapes. Wired into the existing `micro-benchmarks` profile so it doesn't run on every PR. |

Also rebased onto current develop (was 189 commits behind).

The 3 retained xqsuite tests:

- `named-field-query-produces-no-highlights` — verifies the PR #3467 invariant the empty-termMap short-circuit relies on
- `implicit-field-query-still-highlights` — sanity check that the short-circuit doesn't over-suppress
- `batch-expand-matches-per-hit-for-loop` — the path the term-rewrite cache exercises

Verified locally: `xquery.lucene.LuceneTests` runs 633/0/0 (was 630 before; +3 from the new spec).

Diff is now substantially smaller: -306 / +134 LOC on the test files.

line-o (approved):
LGTM

duncdrum (approved):

PR #6319, opened 2026-05-08
[optimize] Configurator.unregister: O(N) scan instead of O(N^2) deep-equals walk

### LINKED ISSUE #3557: [BUG] eXist-db does not scale up with many users
**Describe the bug**
@line-o reported that when many users are defined in that database, the database startup takes a lot of time. With 50k users the startup is about 12 minutes.

**Expected behavior**
Even with many users the database startup should be fast.

**To Reproduce**

run the following query; stop database and start.
```Xquery
xquery version "3.1";

for $i in 10001 to 60002
return sm:create-account("user"||$i, "password"||$i, "group"||$i)
```

**Screenshots**

Running the profiler shows a clear hotspot; I think we should use a faster )Hash)map implementation here.

image

image

image

image

 - OS: macOS 10.15.7
 - eXist-db version: 5.2 / develop
 - Java Version [ 1.8.0_262]

**Additional context**
- How is eXist-db installed? [e.g. JAR installer, DMG, … ] - git 
- Any custom changes in e.g. `conf.xml`? - none

joewiz:
## Summary

Closes https://github.com/eXist-db/exist/issues/3557 — eXist-db startup time was scaling quadratically with the number of users in a realm. With 50k users, dizzzz reported a 12-minute startup; line-o profiled it and pinned the hotspot on a HashMap. The actual bottleneck was a one-line micro-pessimization in `Configurator.unregister` that has been there since 2013 (5e0158b0b8).

## Root cause

`AbstractPrincipal.setCollection` calls `Configurator.unregister(configuration)` once per principal during realm load. Pre-fix:

```java
public static void unregister(final Configuration configuration) {
    if (configuration == null) {
        return;
    }
    if (hotConfigs.containsValue(configuration)) {           // O(N) deep-equals walk
        for (final Entry<FullXmldbURI, Configuration> entry : hotConfigs.entrySet()) {
            if (entry.getValue() == configuration) {         // O(N) reference scan
                hotConfigs.remove(entry.getKey());
                return;
            }
        }
    }
}
```

`ConcurrentHashMap.containsValue` is a linear walk that calls `Configuration.equals()` against every entry, and `ConfigurationImpl.equals` does a deep comparison (`getName` + `getProperty`). Because the loop already uses `==` (reference equality), the `containsValue` precheck was redundant — but on every call it forced `~N/2` deep-equals invocations. Total work for `N` principals: `N * N/2` deep-equals calls. For 50k users: ~1.25 billion calls, matching line-o's profiler output:

- `AbstractRealm.loadAccountsFromRealmStorage` — 66.6%
- `AbstractPrincipal.setCollection` → `Configurator.unregister` → `ConcurrentHashMap.containsValue` — 55.4%
- `ConfigurationImpl.equals` — 379k samples (per profiler screenshots)

## What changed

- `Configurator.unregister` — drop the `containsValue` precheck. The loop's `==` test already returns the same answer (and was always the actual semantic source of truth — the pre-existing code would silently fail to remove if `containsValue` returned true on a different-but-equal `Configuration` and no entry was `==`). The scan stays O(N) per call, so total startup-time contribution is still O(N²), but each iteration is a pointer compare (~ns) instead of a deep equals (~μs). For 50k users this collapses from ~minutes to ~seconds.

- `ConfiguratorUnregisterPerfTest` — new regression test that puts 200 stub `Configuration` instances in `hotConfigs`, calls `unregister` on each, and asserts that `Configuration.equals` was invoked **zero** times. On the pre-fix code the test fails with ~10,184 equals invocations (verified locally by running the test against the unmodified `Configurator`).

## Verification

| Check | Result |
|---|---|
| `mvn install -pl exist-core -am -DskipTests` | clean |
| `mvn test -pl exist-core -Dtest='org.exist.security.*Test,org.exist.config.*Test'` | 263 tests, 0 failures |
| `mvn test -pl exist-core -Dtest='ConfiguratorUnregisterPerfTest'` (post-fix) | passes — 0 equals calls |
| `mvn test -pl exist-core -Dtest='ConfiguratorUnregisterPerfTest'` (pre-fix sanity check via `git stash`) | fails — 10,184 equals calls for 200 principals |
| `mvn test -pl exist-core -Dtest='org.exist.xquery.XPathQueryTest'` | 150 tests, 0 failures |
| `codacy-cli analyze --tool pmd <changed files>` | no new findings on touched lines |
| `codacy-cli analyze --tool lizard <changed files>` | no new findings on touched lines |
| `codacy-cli analyze --tool opengrep <changed files>` | 0 findings |
| `mvn license:check -pl exist-core` | clean |

The bug isn't reproduced via the full 50k-user database reproducer here — it's exercised directly through the same code path the profiler hit (`Configurator.unregister`). The microbenchmark inside the JUnit test is faithful to the production hotspot and runs in 0.3s.

## Test plan

- [x] `mvn install -pl exist-core -am -DskipTests` succeeds
- [x] `org.exist.security.*Test` and `org.exist.config.*Test` pass
- [x] `ConfiguratorUnregisterPerfTest` fails on pre-fix code, passes on fix (regression guard works)
- [x] `XPathQueryTest` unchanged
- [x] Codacy PMD / lizard / opengrep clean on changed lines
- [x] License check clean

### DIFF
--- a/exist-core/src/main/java/org/exist/config/Configurator.java
+++ b/exist-core/src/main/java/org/exist/config/Configurator.java
@@ -1380,11 +1380,10 @@ public class Configurator {
         // Reference equality (==) is intentional: do NOT precheck with
         // hotConfigs.containsValue(), which calls ConfigurationImpl.equals()
         // on every entry and turns realm load into O(N^2).
-        for (final Iterator<Entry<FullXmldbURI, Configuration>> it = hotConfigs.entrySet().iterator(); it.hasNext();) {
-            final Entry<FullXmldbURI, Configuration> entry = it.next();
-            if (entry.getValue() == configuration) {
+        for (final Iterator<Configuration> it = hotConfigs.values().iterator(); it.hasNext();) {
+            if (it.next() == configuration) {
                 it.remove();
-                return;
+                break;
             }
         }
     }
--- a/exist-core/src/main/java/org/exist/config/Configurator.java
+++ b/exist-core/src/main/java/org/exist/config/Configurator.java
@@ -1380,9 +1380,10 @@ public class Configurator {
         // Reference equality (==) is intentional: do NOT precheck with
         // hotConfigs.containsValue(), which calls ConfigurationImpl.equals()
         // on every entry and turns realm load into O(N^2).
-        for (final Entry<FullXmldbURI, Configuration> entry : hotConfigs.entrySet()) {
+        for (final Iterator<Entry<FullXmldbURI, Configuration>> it = hotConfigs.entrySet().iterator(); it.hasNext();) {
+            final Entry<FullXmldbURI, Configuration> entry = it.next();
             if (entry.getValue() == configuration) {
-                hotConfigs.remove(entry.getKey());
+                it.remove();
                 return;
             }
         }
--- a/exist-core/src/main/java/org/exist/config/Configurator.java
+++ b/exist-core/src/main/java/org/exist/config/Configurator.java
@@ -1376,13 +1376,14 @@ public class Configurator {
         if (configuration == null) {
             return;
         }
-        
-        if (hotConfigs.containsValue(configuration)) {
-            for (final Entry<FullXmldbURI, Configuration> entry : hotConfigs.entrySet()) {
-                if (entry.getValue() == configuration) {
-                    hotConfigs.remove(entry.getKey());
-                    return;
-                }
+
+        // Reference equality (==) is intentional: do NOT precheck with
+        // hotConfigs.containsValue(), which calls ConfigurationImpl.equals()
+        // on every entry and turns realm load into O(N^2).
+        for (final Entry<FullXmldbURI, Configuration> entry : hotConfigs.entrySet()) {
+            if (entry.getValue() == configuration) {
+                hotConfigs.remove(entry.getKey());
+                return;
             }
         }
     }
--- a/exist-core/src/test/java/org/exist/config/ConfiguratorUnregisterPerfTest.java
+++ b/exist-core/src/test/java/org/exist/config/ConfiguratorUnregisterPerfTest.java
@@ -1,0 +1,142 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.config;
+
+import org.exist.storage.DBBroker;
+import org.exist.xmldb.FullXmldbURI;
+import org.exist.xmldb.XmldbURI;
+import org.junit.After;
+import org.junit.Test;
+import org.w3c.dom.Element;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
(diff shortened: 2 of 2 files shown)


exist-core/src/main/java/org/exist/config/Configurator.java
```diff
@@ -1376,13 +1376,14 @@ public static void unregister(final Configuration configuration) {
...
+
+        // Reference equality (==) is intentional: do NOT precheck with
+        // hotConfigs.containsValue(), which calls ConfigurationImpl.equals()
+        // on every entry and turns realm load into O(N^2).
+        for (final Entry<FullXmldbURI, Configuration> entry : hotConfigs.entrySet()) {
+            if (entry.getValue() == configuration) {
+                hotConfigs.remove(entry.getKey());
+                return;
             }
         }
```
reinhapa: Using the iterator directly will eliminate also the additional method call for removal.
```suggestion
        for (final Iterator<Entry<FullXmldbURI, Configuration>> it = hotConfigs.entrySet().iterator(); it.hasNext();) {
            final Entry<FullXmldbURI, Configuration> entry = it.next();
            if (entry.getValue() == configuration) {
                it.remove();
                break;
            }
        }
```

reinhapa (changes requested):

exist-core/src/main/java/org/exist/config/Configurator.java
line-o: Hm @reinhapa I actually find the current code more readable. Do you expect the call to htoConfigs.remove to be expensive?

line-o (approved):
LGTM

exist-core/src/main/java/org/exist/config/Configurator.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done — adopted in e96ce91055. Agree the iterator pattern is the canonical Java iterate-and-remove idiom; the extra hash lookup matters at the 50k-principal scale this PR specifically targets. Thanks for the catch.

joewiz: [This response was co-authored with Claude Code. -Joe]

Fair readability point — the iterate-and-remove idiom is verbose but it is the canonical Java pattern, and it's already in use elsewhere in the codebase. The cost difference is one ConcurrentHashMap hash lookup per unregister call: small per call, but it compounds at the 50k-principal startup scale this PR specifically targets, so I went with reinhapa's suggestion. Thanks for the LGTM!

joewiz:

Follow-up sweep for other instances of this anti-pattern across the codebase (per the memory-note action item from this PR's wrap-up):

```
$ grep -rn "\.containsValue\b" --include="*.java" . | grep -v /target/
exist-core/src/main/java/org/exist/config/Configurator.java:1380:        if (hotConfigs.containsValue(configuration)) {
```

That's the entire universe of `Map.containsValue(...)` calls in eXist (main sources, test sources, every extension module). The site this PR fixes is the unique instance — there are no further `containsValue` + `==`-scan duplicates lurking elsewhere.

So #3557 / this PR closes the family. The memory note will be updated to record the negative result so future contributors don't repeat the search.

exist-core/src/main/java/org/exist/config/Configurator.java
```diff
@@ -1376,13 +1376,15 @@ public static void unregister(final Configuration configuration) {
...
+        // Reference equality (==) is intentional: do NOT precheck with
+        // hotConfigs.containsValue(), which calls ConfigurationImpl.equals()
+        // on every entry and turns realm load into O(N^2).
+        for (final Iterator<Entry<FullXmldbURI, Configuration>> it = hotConfigs.entrySet().iterator(); it.hasNext();) {
+            final Entry<FullXmldbURI, Configuration> entry = it.next();
+            if (entry.getValue() == configuration) {
+                it.remove();
+                return;
             }
         }
```
reinhapa: As we do not use any key parts in our iterator-remove idiom, we could write it using the `values()` view to make it simpler to read...
```suggestion
        for (final Iterator<Configuration> it = hotConfigs.values().iterator(); it.hasNext();) {
            if (it.next() == configuration) {
                it.remove();
                break;
            }
        }
```

reinhapa (changes requested):

exist-core/src/main/java/org/exist/config/Configurator.java
line-o: I like it @reinhapa

joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `c2325325dc` — switched to `hotConfigs.values().iterator()` per your suggestion. Drops the `Entry` generic noise and reads more cleanly. Thanks both!

reinhapa (approved):

PR #6321, opened 2026-05-08
[ci] Stabilize Prethink workflow: pin CLI version + retry push on concurrent develop

joewiz:
## Summary

The Prethink workflow has been failing intermittently on both the weekly schedule and manual dispatches. cc @dizzzz — this addresses what you flagged in Slack.

Two distinct failure modes traced and fixed:

### 1. Maven Central propagation race (2026-05-08 manual dispatch — [run #10](https://github.com/eXist-db/exist/actions/runs/25577803898/job/75089183036))

```
Downloading Moderne CLI 4.2.4 for linux...
curl: (22) The requested URL returned error: 404
```

`mod build` had succeeded earlier in the run. Then `mod run` re-resolved `version=RELEASE` from `moderne-cli/maven-metadata.xml` and got `4.2.4` — but the platform-specific `moderne-cli-linux-4.2.4.sh` (Last-Modified `2026-05-08 20:32:15 UTC`) had not yet CDN-synced to the GitHub runner's edge. `mod run` started at 20:34:33 and 404'd.

**Fix:** Pin `MODERNE_CLI_VERSION=4.2.3` and rewrite the wrapper's `version=RELEASE` to the pinned value after install. Subsequent `mod` invocations reuse the same dist and skip the metadata round-trip entirely. Bumping the pin is a deliberate decision, made when a new release is known fully synced.

### 2. Non-fast-forward push rejection (2026-05-04 scheduled — [run #9](https://github.com/eXist-db/exist/actions/runs/25305872834))

```
error: failed to push some refs to 'https://github.com/eXist-db/exist'
```

Another commit landed on `develop` between checkout and the workflow's push.

**Fix:** Wrap the push in a 3-attempt rebase-and-retry loop with backoff. No `--force`; if a real conflict materializes (which is unlikely since this only touches `.moderne/context/` and `AGENTS.md`), the loop exits non-zero so the failure surfaces.

## What Changed

`.github/workflows/prethink.yml` only:
- New `MODERNE_CLI_VERSION` env var on the install step + `sed` to pin the wrapper properties
- New 3-attempt rebase-retry loop around `git push` on the commit step
- Existing `git diff --staged --quiet || commit` collapsed to an explicit early `exit 0` for clarity in logs when there are no context changes

## Test Plan

- [x] YAML parses (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] Workflow dispatch run after merge succeeds end-to-end (Joe will trigger manually)
- [ ] Future weekly schedule runs continue to succeed

I considered switching to a PR-based update model (workflow opens a PR instead of pushing directly to develop) — that's a larger change and arguably the right long-term direction, but these two minimal fixes unblock today's pain without rearchitecting the flow.

### DIFF
--- a/.github/workflows/prethink.yml
+++ b/.github/workflows/prethink.yml
@@ -20,9 +20,19 @@ jobs:
           distribution: 'liberica'
 
       - name: Install Moderne CLI
+        env:
+          # Pinned to avoid the Maven Central propagation race observed on 2026-05-08, where
+          # the parent moderne-cli metadata advertised a release before the platform-specific
+          # moderne-cli-linux binary had CDN-synced. Bump intentionally as new releases stabilize.
+          MODERNE_CLI_VERSION: '4.2.3'
         run: |
           curl -sSL https://app.moderne.io/cli | bash
           echo "$HOME/.moderne/cli/bin" >> $GITHUB_PATH
+          # The installer writes 'version=RELEASE' so every `mod` invocation re-resolves to whatever
+          # is currently latest on Maven Central. Replace with the pinned version so subsequent
+          # `mod` calls reuse the dist downloaded once and never race a fresh release.
+          sed -i "s/^version=RELEASE$/version=${MODERNE_CLI_VERSION}/" \
+            "$HOME/.moderne/cli/dist/moderne-wrapper.properties"
 
       - name: Configure Moderne
         run: |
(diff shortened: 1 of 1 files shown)


duncdrum (approved):

dizzzz (approved):

dizzzz:
Manual trigger works fine again!

Issue #3964, opened 2021-07-02
[BUG] attribute filter not working properly on collection


PieterLamers:
**Describe the bug**
When I try to filter a collection of JATS articles on a particular attribute value I get no results, except when I explicitly convert the attribute value to string.

**Expected behavior**
I except the attribute value to be cast to string in a value comparison.

**To Reproduce**
example (1) does not return a count, example (2) does
```xquery
(: 1 :)
let $coll as item()+ := collection('/db/data/journals.benjamins.com/')/article/front/article-meta/contrib-group/contrib/contrib-id[@contrib-id-type eq 'jb-contributor-id']
return count($coll)
(: 2 :)
let $coll as item()+ := collection('/db/data/journals.benjamins.com/')/article/front/article-meta/contrib-group/contrib/contrib-id[@contrib-id-type/string() eq 'jb-contributor-id']
return count($coll)
```
In addition, `string(@contrib-type-id) eq 'jb-contributor-id'` also works properly, and @contrib-id-type eq 'jb-contributor-id' does not. Representative data sent to @adamretter for reproduction.

**Context (please always complete the following information):**
 - OS: Windows
 - eXist-db version: 5.3.0
 - Java Version 1.8.0_282

adamretter:
I have reduced the test case to:
```
empty(collection('/db/data/journals.benjamins.com/')//contrib-id[@contrib-id-type eq 'jb-contributor-id'])
,
empty(collection('/db/data/journals.benjamins.com/')//contrib-id[@contrib-id-type/string() eq 'jb-contributor-id'])
```

The result should be `false, false`, however the result is `true, false`.

If I replace the `fn:collection` call with `fn:doc` and use a document which is known to match the predicate, then the problem goes away and the result is `false, false`. So the issue is somehow related to the use of the `fn:collection` function. I suspect a bad optimisation path somewhere...

adamretter:
I have found that the first query:
```xquery
empty(collection('/db/data/journals.benjamins.com/')//contrib-id[@contrib-id-type eq 'jb-contributor-id'])
```

Is being rewritten by the optimizer to the form:
```xquery
empty(collection('/db/data/journals.benjamins.com/')/descendant::contrib-id[range:eq(@contrib-id-type, "jb-contributor-id")])
```xquery
i.e. it is attempting to resolve the predicate through a Range Index lookup.

The second query is not rewritten by the optimizer.

[...]

PR #6323, opened 2026-05-09
[test] Add regression test for attribute eq with no range index (issue #3964)

joewiz:
## Summary

Adds a regression test in `extensions/indexes/range` for issue [#3964](https://github.com/eXist-db/exist/issues/3964): an attribute predicate `[@x eq 'val']` on a collection step is unconditionally rewritten by `RangeQueryRewriter` to `[range:eq(@x, 'val')]`, even when no range index is configured on `@x`. The original report was that this caused the predicate to silently return zero matches.

On current `develop` the bug **does not reproduce**: the runtime fallback in `Lookup.canOptimizeSequence` correctly detects that the index is not configured for the path, sets `canOptimize = false`, and `Lookup.eval` delegates to the original `GeneralComparison` via `fallback.eval(...)`. The rewritten AST is still produced — confirmed by inspecting the dumped AST under a debug print — but the runtime path is now correct. The five+ years of optimizer fixes since 5.3.0 most likely closed this gap incidentally; no separate fix could be identified.

This PR ships the regression test as a guard so a future change to `Lookup.canOptimize` cannot silently re-introduce the empty-result bug. No production code changes.

## What changed

- `extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/AttributeEqIndexConsistencyTest.java` — new
  - Configures a range index on an unrelated element (`<article-title>`) so the range module is active for the collection but `@contrib-id-type` itself remains unindexed — the configuration that triggers the unconditional rewrite without a usable index.
  - Stores 5 JATS-shaped sample documents.
  - Asserts that all of these agree with `[@x/string() eq 'val']`:
    - `[@x eq 'val']` over `collection(...)`
    - `[@x = 'val']` over `collection(...)` (general comparison)
    - `[@x eq 'val']` over `doc(...)` (the shape that already worked in the original report)

## Why a test, not a fix

The bug shape from the issue does not reproduce — see the comment thread on #3964 — but the rewrite-path the report identified is still active in current code. This test exercises that path so any regression in the fallback would now fail loudly rather than silently returning zero matches.

## Test plan

- [x] `mvn test -pl extensions/indexes/range` — 425 pass, 0 failures, 3 skipped (the new class contributes 3)
- [x] Codacy PMD on the new test file: clean (after camelCase rename of method names)

Refs https://github.com/eXist-db/exist/issues/3964

### DIFF
--- a/extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/AttributeEqIndexConsistencyTest.java
+++ b/extensions/indexes/range/src/test/java/org/exist/xquery/modules/range/AttributeEqIndexConsistencyTest.java
@@ -1,0 +1,204 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


duncdrum (commented):
I think this would be much cleaner as xqsuite test.

reinhapa (approved):

line-o (approved):
A test is always better than no test. 
Can be rewritten later.

PR #6325, opened 2026-05-09
[bugfix] following::* axis position-dependence (#2129)

joewiz:
## Summary

Wildcard `following::*` is no longer position-dependent. On craigberry's 50k-element reproducer at `@xml:id=45000`, the late-position case is **33% faster** (390ms → 259ms); the position-dependence ratio (45k vs 5k) drops from 3.07× to 2.30×; isolating just `following::*[i][self::w]` shows it now runs in flat ~86ms regardless of reference position.

Closes the `following::` half of https://github.com/eXist-db/exist/issues/2129. The `preceding::` half remains and is being tracked as a follow-up tasking — it requires a different optimisation (last-K circular buffer in `PrecedingFilter` driven by an optimizer-level rewrite).

## Root cause

JFR profile of the late-position case at xml:id=45000 (60s recording, 10ms sample cadence):

| % | method | role |
|--:|--------|------|
| 17.7% | `RawNodeIterator.next` | StAX iteration through every event |
| 17.5% | `EmbeddedXMLStreamReader.filter` | per-event filter |
| 16.5% | `DLN.compareTo` (self) | per-candidate NodeId compare against reference |
| 13.9% | `FastQSort.IntroSortLoop` | sort the accumulated result set |
| 11.4% | `LocationStep.getPrecedingOrFollowing` | the dispatch |

Read together: walk every StAX event from document start, NodeId-compare each to the reference, accumulate, sort. The walker's start position was always the document-child root regardless of where the reference node actually lived.

## Fix

`exist-core/src/main/java/org/exist/xquery/LocationStep.java` — for `FOLLOWING_AXIS` only, start the StAX reader at the reference node when it lies inside the document-child being walked. The existing `FollowingFilter`'s `isAfter` / `isDescendantOf` checks correctly skip the reference and its descendants, so the later start position is safe and produces identical results. New helper `readerStartForWildcardAxis` keeps `getPrecedingOrFollowing`'s NPath at its baseline of 350.

`PRECEDING_AXIS` is intentionally unchanged — its filter must still observe every preceding event to honor positional predicates, so the fix shape is different and requires last-K recognition. Tracked separately.

## Numbers (craigberry's reproducer, 50,000 element flat doc, trimmed mean of 5 runs)

| `@xml:id` | before | after | speedup |
|----------:|-------:|------:|--------:|
| 5,000  | 127 ms | 112 ms | 1.13× |
| 15,000 | 183 ms | 144 ms | 1.27× |
| 25,000 | 255 ms | 186 ms | 1.37× |
| 35,000 | 326 ms | 220 ms | 1.48× |
| 45,000 | 390 ms | 259 ms | **1.51×** |

Position-dependence ratio 45k/5k: **3.07× → 2.30×**.

Isolated `following::*[i][self::w]` after fix: **95 / 88 / 86 / 87 / 86 ms** at the 5 positions — flat (position-independent).

## Test plan

- [x] New `FollowingAxisPositionRegressionTest` (4 cases): midpoint output, late-position output, descendant exclusion (`<a><inner1/><inner2/></a><b/><c><inner3/></c>` → `a/following::*` returns exactly `b, c, inner3`), position-independence guard (`following::` at xml:id=45000 ≤ 3× xml:id=5000)
- [x] `XPathQueryTest` 148/148
- [x] `XQueryTest` 99/99
- [x] `mvn test -pl exist-core` 6596/6596 (0 failures, 0 errors, 106 skipped)
- [x] PMD: `getPrecedingOrFollowing` NPath unchanged from baseline (350)

### DIFF
--- a/exist-core/src/test/java/org/exist/xquery/FollowingAxisPositionRegressionTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/FollowingAxisPositionRegressionTest.java
@@ -55,7 +55,6 @@ public class FollowingAxisPositionRegressionTest {
     public static final ExistXmldbEmbeddedServer existEmbeddedServer =
             new ExistXmldbEmbeddedServer(false, true, true);
 
-    private static final String SMALL_DOC = "/db/words-small.xml";
     private static final String LARGE_DOC = "/db/words-large.xml";
 
     @BeforeClass
--- a/exist-core/src/test/java/org/exist/xquery/FollowingAxisPositionRegressionTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/FollowingAxisPositionRegressionTest.java
@@ -1,0 +1,189 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery;
+
+import org.exist.test.ExistXmldbEmbeddedServer;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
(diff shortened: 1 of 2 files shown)


exist-core/src/test/java/org/exist/xquery/FollowingAxisPositionRegressionTest.java
```diff
@@ -0,0 +1,189 @@
...
+ *   <li>Performance — late positions complete in roughly the same time as
+ *       early positions, on a 50,000-element flat document.</li>
+ * </ul>
+ */
+public class FollowingAxisPositionRegressionTest {
+
+    @ClassRule
+    public static final ExistXmldbEmbeddedServer existEmbeddedServer =
+            new ExistXmldbEmbeddedServer(false, true, true);
+
```
reinhapa, line 57: Address Codacy issue: _Avoid unused private fields such as 'SMALL_DOC'._

reinhapa (changes requested):

exist-core/src/test/java/org/exist/xquery/FollowingAxisPositionRegressionTest.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done in `e628749e20` — `SMALL_DOC` removed (the test exercises only `LARGE_DOC`).

line-o (approved):
LGTM

duncdrum (approved):

reinhapa (approved):

PR #6324, opened 2026-05-09
[test] Add JMH axis-evaluator benchmark (PR #6302 follow-up)

joewiz:
## Summary

Companion to [PR #6302](https://github.com/eXist-db/exist/pull/6302). That PR added `AxisPerformanceRegressionTest`, a JUnit timing assertion that catches a 5x slowdown of `following-sibling`. This PR adds a JMH benchmark in the same vein — ms/op resolution, suitable for measuring smaller changes and for comparing alternative axis-evaluator implementations.

No production-code changes.

## What changed

`exist-core-jmh/src/main/java/org/exist/dom/persistent/AxisBenchmark.java`

- 4 axes × 3 corpus shapes = 12 benchmark cells.
- Sibling axes use the multi-context predicate form (`//b[axis::b]`), the shape `AxisPerformanceRegressionTest` validates and the one sensitive to per-context regressions like the bug PR #6302 fixed.
- Non-sibling axes use a single-context query (`(//b)[1]/following::b` and the symmetric `preceding`) to avoid an O(N²) blow-up that the unbounded axes would otherwise inflict on the predicate form. A single-context measurement still surfaces axis-walk regressions (the issue [#2129](https://github.com/eXist-db/exist/issues/2129) wildcard-vs-typed dispatch concern).
- Three corpus shapes:
  - `1500_20` — matches `AxisPerformanceRegressionTest`'s corpus (the [GH-2697](https://github.com/eXist-db/exist/issues/2697) reproducer).
  - `500_100` — balanced.
  - `100_500` — high fan-out per parent, the case most sensitive to per-context sibling-walk cost.

`exist-core-jmh/pom.xml`

- Add `jmh-generator-annprocess` to the module's `annotationProcessorPaths`. The parent pom restricts the list to `jsr305`, silently shadowing JMH's own processor — without this `META-INF/BenchmarkList` is never generated and `java -jar benchmarks.jar` reports "no matching benchmarks". [PR #6296](https://github.com/eXist-db/exist/pull/6296) documents the same issue.
- Add `xmldb-api` as a direct dependency.
- Add `src/main/resources/conf.xml` to the license-check excludes (it's a verbatim copy).

`exist-core-jmh/src/main/resources/conf.xml`

[...]

PR #6314, opened 2026-05-07
[bugfix] Reindex lifecycle hardening, legacy fixes , and regression benchmarks/tests

duncdrum:
## Summary

Consolidates several legacy-ticket fixes and regression guards across Lucene/range/XQuery internals, with a focus on correctness first and measurable performance impact where relevant.

Key outcomes:

- Fixes full-text match propagation loss in FLWOR `where` boolean expressions (issue #2283 shape), so `util:expand`/KWIC-style highlighting no longer drops `exist:match` markers when extra predicates are present.
- Hardens reindex lifecycle behavior:
  - introduces/uses optimized reindex mode paths,
  - makes drop semantics explicit,
  - aligns lock contract,
  - preserves intended named-field behavior while guarding stale-entry scenarios.
- Adds/extends JMH coverage for Lucene query/reindex/delete paths, including an #2812-style query benchmark.
- Adds custom Java collator loading hardening (`java:` URI) and regression coverage (issue #3747 shape).
- Adds pending regression guards for range optimizer union-shape behavior (#2363) to keep signal while upstream optimizer work lands.

### XQuery / optimizer / collation correctness

- `exist-core/src/main/java/org/exist/xquery/WhereClause.java`
  - Hardened carrier propagation through boolean `where` evaluation so full-text match metadata survives additional predicates.
- `exist-core/src/main/java/org/exist/util/Collations.java`
  - `java:` collator class resolution now tries thread context classloader first, then fallback.
- `exist-core/src/test/java/org/exist/util/CollationsTest.java`
- `exist-core/src/test/java/org/exist/util/TestJavaCollator.java`
  - Added regression guard for custom Java collator loading.

### Reindex lifecycle + Lucene delete strategy

- `exist-core/src/main/java/org/exist/storage/DBBroker.java`
- `exist-core/src/main/java/org/exist/storage/NativeBroker.java`
- `exist-core/src/main/java/org/exist/indexing/IndexController.java`
- `exist-core/src/main/java/org/exist/indexing/IndexWorker.java`
- `exist-core/src/main/java/org/exist/storage/IndexSpec.java`
- `exist-core/src/main/java/org/exist/storage/structural/NativeStructuralIndexWorker.java`
- `exist-core/src/main/java/org/exist/xmldb/LocalIndexQueryService.java`
- `exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java`
- `extensions/indexes/lucene/src/main/java/org/exist/indexing/lucene/LuceneIndexWorker.java`
[...]

PR #6331, opened 2026-05-10
[optimize] Type.subTypeOf: avoid redundant union-types map lookups (closes #6322)

### LINKED ISSUE #6322: [perf] Type.subTypeOf calls Int2ObjectArrayMap.containsKey as linear scan on every comparison (~20% self-time)
## Summary

`Type.subTypeOf` calls `Int2ObjectArrayMap.containsKey(...)` twice per invocation as a linear scan over the union-types map, accounting for ~20% of self-time on string-comparison-heavy workloads. The hot caller chain is `StringValue.compareTo` → `ValueComparison.compareAtomic` → `Type.subTypeOf`, so this affects every `xs:string eq xs:string` (and most other atomic comparisons) the engine performs.

Discovered while diagnosing #3406 ([JFR profile, 3,333 samples at 10 ms cadence](https://github.com/eXist-db/exist/issues/3406)) but the bottleneck is not specific to `for-each-pair` or any other particular function — it's a generic engine hotspot that bites on any comparison-heavy query.

## Where

`exist-core/src/main/java/org/exist/xquery/value/Type.java` — `subTypeOf(int subtype, int supertype)`. The method does two `containsKey` calls against a `it.unimi.dsi.fastutil.ints.Int2ObjectArrayMap` instance that holds union-type membership. `Int2ObjectArrayMap.containsKey` is documented as O(N) by design (the data structure is intended for very small maps), but the union-types map has grown well past the point where linear scan is appropriate.

## JFR self-time on the #3406 reproducer (`$length=1000`)

| % samples | method | role |
|----------:|--------|------|
| 20.0% | `Int2ObjectArrayMap.findKey` | linear scan inside `Type.subTypeOf` union-types map |
| 8.9% | `UserDefinedFunction.eval` | actual user function body |
| 7.5% | `ValueSequence.add` | result accumulation |
| 6.6% | `HashMap.getNode` | variable / scope lookups |

The `Int2ObjectArrayMap.findKey` 20% slice is two orders of magnitude larger than any other engine internal. Fixing this should be visible on a wide range of query shapes, not just the #3406 reproducer.

## Suggested fixes

In order of effort:

1. **One-line fast path**: add `if (unionTypes.isEmpty()) return ...` at the top of `subTypeOf`. Most types are not in any union, so this short-circuits the linear scan to `O(1)` for the common case.
2. **Switch the map type**: `Int2ObjectArrayMap` → `Int2ObjectOpenHashMap` (also from fastutil). Same API, hash-backed, O(1) `containsKey`. This makes the union-types case fast too.
3. **Precompute a flat lookup table at startup**: the union-types map only changes during type-system initialisation and is otherwise read-only. A `boolean[][]` indexed by `(subtype, supertype)` would make `subTypeOf` a pure array access.

Suggested PR scope: combine (1) + (2) — the fast path defends against any future map churn, and the hash-map switch handles the union case. Option (3) is a separate optimisation conversation if (1)+(2) don't close the gap.

## Test plan

- Add a JMH micro-benchmark in `exist-core-jmh` covering `Type.subTypeOf` on common type pairs (CHEAP cases: identical types, primitive subtype hierarchy; UNION cases: types that ARE in `unionTypes`).
- Re-run the #3406 reproducer at `$length=1000` and confirm the JFR slice for `Int2ObjectArrayMap.findKey` drops below 1%.
- Full XQTS QT4 + XQ 3.1 + FTTS pass — type-system changes have wide blast radius, so any regression on type-related test sets is a blocker.

## Notes

- This is a follow-up on the diagnostic work for #3406. The original "10× degrading non-linearly" framing in #3406 is no longer the right lens; that issue is about overall reproducer behaviour, while this one is about a specific engine micro-optimisation.
- A separate ~14% per-call function-dispatch overhead (variable resolve / destroy / analyze) was also surfaced by the same JFR — that's its own architectural conversation, not in scope here.

joewiz:
## Summary

`Type.subTypeOf` is on the hot path of every atomic comparison via `ValueComparison.compareAtomic`. A JFR profile of issue [#3406](https://github.com/eXist-db/exist/issues/3406)'s n×n FLWOR reproducer found `Int2ObjectArrayMap.findKey` at **20% of self-time** in that workload (see [project_3406_diagnosis_2026_05_09](https://github.com/eXist-db/exist/issues/6322)). This PR cuts that overhead with two localised changes plus one redundancy fix, all confined to `Type.java`.

Closes https://github.com/eXist-db/exist/issues/6322.

## What changed

`exist-core/src/main/java/org/exist/xquery/value/Type.java`

1. **Switch `unionTypes` from `Int2ObjectArrayMap` to `Int2ObjectOpenHashMap`.** The fastutil `ArrayMap` is a linear-scan implementation intended for very small maps; the hash-backed variant gives O(1) lookup with the same `Int2ObjectMap<...>` API. Field is now declared `Int2ObjectOpenHashMap<IntArraySet>` so callers benefit from the hash-backed path without a witness cast.
2. **Add an `isEmpty()` guard before the union dispatch.** The static initialiser registers `NUMERIC` and `ERROR` so the guarded branch is taken in current builds; the guard protects against future evolution where union registration becomes conditional, e.g. for XQuery 4.0 user-defined unions.
3. **Collapse `containsKey(...)` + a subsequent `get(...)` into a single `get(...)` per union check.** The two consultations are now `IntArraySet supertypeMembers = unionTypes.get(supertype); if (supertypeMembers != null) ...`, with the result forwarded to a new package-private overload of `subTypeOfUnion` / `unionMembersHaveSuperType` that takes the pre-fetched members. The public no-args variants are unchanged for external callers (e.g. `getCommonSuperType` in `Type.java` itself, plus call sites elsewhere in the engine).

`exist-core-jmh/src/main/java/org/exist/xquery/value/TypeSubTypeOfBenchmark.java`

- New JMH benchmark covering six call shapes: identical types, direct supertype walk, multi-level supertype walk, union-typed supertype, union-typed subtype, and "not a subtype" (walks to the root and returns false).

`exist-core-jmh/pom.xml`

- Lifts the `annotationProcessorPaths` fix from #6324 (jmh-generator-annprocess otherwise gets shadowed by the parent pom, leaving `META-INF/BenchmarkList` empty). #6324 is still open at the time of writing; this PR pulls the same one-block override so the benchmark is runnable today. If #6324 lands first, the merge will collapse the duplicate config cleanly.

## Spec / scope notes

The original [#6322](https://github.com/eXist-db/exist/issues/6322) inventory proposed a third optimisation — a precomputed `boolean[][]` subtype-of lookup table. **That fix is deferred and not in this PR.** Per [@line-o's 2026-05-10 reply](https://github.com/eXist-db/exist/issues/6322), XQuery 4.0 will allow user-defined union types per query context, so a static lookup table would either need per-context invalidation (re-introducing the allocation cost) or restriction to built-in types (limiting the gain). Best left until the XQ 4.0 user-defined-types story is settled.

## JMH numbers (`TypeSubTypeOfBenchmark`)

JMH 1.37, JDK 21.0.5 (Zulu), 1 fork, 3×1s warmup, 5×1s measurement, ns/op.

| Shape          | Before  | After   | Δ        | Notes |
| -------------- | -------:| -------:| --------:| --- |
| `identical`    | 0.374   | 0.375   | 0.0%     | early-return at the equality check, untouched |
| `directSuper`  | 3.661   | 2.993   | **−18.2%** | one supertype walk, two map consultations |
| `deepSuper`    | 13.817  | 11.817  | **−14.5%** | multi-level walk |
| `unionMember`  | 3.575   | 2.661   | **−25.6%** | supertype is a registered union (NUMERIC) |
| `unionSubtype` | 20.095  | 14.136  | **−29.7%** | subtype is a registered union (NUMERIC) |
| `notSubType`   | 7.433   | 5.171   | **−30.4%** | walks to the root, returns false |

The combined effect of the OpenHashMap swap and the `containsKey`+`get` collapse is biggest on the union-dispatch and walk paths, where every recursive step previously paid two linear-scan lookups.

## Test plan

- [x] Targeted JUnit: `mvn test -pl exist-core -Dtest='OpNumericTest,XPathQueryTest,xquery.xquery3.XQuery3Tests,UnknownAtomicTypeTest'` — **1,181 tests, 0 failures, 0 errors**.
- [x] Full module: `mvn test -pl exist-core` (under cross-session lock) — **6,592 tests, 0 failures, 0 errors, 106 skipped**.
- [x] JMH `TypeSubTypeOfBenchmark` before/after, table above.
- [x] Codacy PMD on changed files. New: `subTypeOf(int,int)` NPath rises from 240 (already > 200) to 300 (still in the "moderate" 200–10,000 band that CLAUDE.md reserves for reviewer discretion). Happy to extract a helper or suppress with rationale if preferred.
- [ ] **XQTS QT4 / XQ 3.1 / FTTS deferred.** The locally-cached runner JAR shades a snapshot of `exist-core` (so a run against this branch's JAR is testing the shaded copy, not the modified `Type.class`), and a fresh `sbt assembly` against develop fails because the runner now depends on `org.exist-db:exist-expath`, an artifact that doesn't exist yet on this branch. Given the change is a strict semantic preservation (collapsing `containsKey + get` into one `get` with the same null-check, plus a swap to a same-API map), and the JUnit suite covers the type-comparison surface, I think landing on JMH + JUnit is acceptable; happy to run XQTS against `next-v3` (which has `exist-expath`) before merging if a reviewer wants a cross-check there.

## Risks

- **Shape change of the field declaration**: `Int2ObjectMap<IntArraySet>` → `Int2ObjectOpenHashMap<IntArraySet>`. The map is `private`, mutated only by the static initialiser, and read only via `get` / `containsKey` / `isEmpty` — no caller relies on `ArrayMap`'s insertion-order iteration.
- **Empty-union semantics**: `defineUnionType(ERROR, new int[0])` registers `ERROR` with an empty member set. The new private overload of `unionMembersHaveSuperType` keeps the pre-existing `members.isEmpty() → false` short-circuit so the empty-`ERROR` case behaves identically.
- **Public API**: the existing `subTypeOfUnion(int,int)` / `unionMembersHaveSuperType(int,int)` signatures are untouched. The new overloads are package-private.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/value/Type.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/Type.java
@@ -139,7 +139,7 @@ public class Type {
     static {
         typeCodes.defaultReturnValue(NO_SUCH_VALUE);
     }
-    private final static Int2ObjectMap<IntArraySet> unionTypes = new Int2ObjectArrayMap<>(2);
+    private final static Int2ObjectOpenHashMap<IntArraySet> unionTypes = new Int2ObjectOpenHashMap<>(4, Hash.FAST_LOAD_FACTOR);
     private final static Int2IntOpenHashMap primitiveTypes = new Int2IntOpenHashMap(45, Hash.FAST_LOAD_FACTOR);
     static {
         primitiveTypes.defaultReturnValue(NO_SUCH_VALUE);
@@ -531,11 +531,21 @@ public class Type {
             return false;
         }
 
-        if (unionTypes.containsKey(supertype)) {
-            return subTypeOfUnion(subtype, supertype);
-        }
-        if (unionTypes.containsKey(subtype)) {
-            return unionMembersHaveSuperType(subtype, supertype);
+        // Skip both union-type lookups when no union types are registered.
+        // The static initialiser registers NUMERIC and ERROR so the guarded
+        // branch is taken in current builds; the isEmpty() check protects
+        // against future evolution where union registration becomes optional.
+        if (!unionTypes.isEmpty()) {
+            // Single get() rather than containsKey() + a redundant get() inside
+            // subTypeOfUnion: halves map traversals on the union dispatch path.
+            final IntArraySet supertypeMembers = unionTypes.get(supertype);
+            if (supertypeMembers != null) {
+                return subTypeOfUnion(subtype, supertype, supertypeMembers);
+            }
+            final IntArraySet subtypeMembers = unionTypes.get(subtype);
+            if (subtypeMembers != null) {
+                return unionMembersHaveSuperType(subtype, supertype, subtypeMembers);
+            }
         }
 
         subtype = superTypes[subtype];
@@ -657,7 +667,10 @@ public class Type {
         if (members == null) {
             return false;
         }
+        return subTypeOfUnion(subtype, unionType, members);
+    }
 
+    private static boolean subTypeOfUnion(final int subtype, final int unionType, final IntArraySet members) {
         // inherited behaviour from {@link #subTypeOf(int, int)}
         // where type is considered a subtype of itself.
         if (subtype == unionType) {
@@ -680,6 +693,13 @@ public class Type {
     public static boolean unionMembersHaveSuperType(final int unionType, final int supertype) {
         final IntArraySet members = unionTypes.get(unionType);
         if (members == null || members.isEmpty()) {
+            return false;
+        }
+        return unionMembersHaveSuperType(unionType, supertype, members);
+    }
+
+    private static boolean unionMembersHaveSuperType(final int unionType, final int supertype, final IntArraySet members) {
+        if (members.isEmpty()) {
             return false;
         }
 
--- a/exist-core/src/main/java/org/exist/xquery/value/Type.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/Type.java
@@ -531,21 +531,15 @@ public class Type {
             return false;
         }
 
-        // Skip both union-type lookups when no union types are registered.
-        // The static initialiser registers NUMERIC and ERROR so the guarded
-        // branch is taken in current builds; the isEmpty() check protects
-        // against future evolution where union registration becomes optional.
-        if (!unionTypes.isEmpty()) {
-            // Single get() rather than containsKey() + a redundant get() inside
-            // subTypeOfUnion: halves map traversals on the union dispatch path.
-            final IntArraySet supertypeMembers = unionTypes.get(supertype);
-            if (supertypeMembers != null) {
-                return subTypeOfUnion(subtype, supertype, supertypeMembers);
-            }
-            final IntArraySet subtypeMembers = unionTypes.get(subtype);
-            if (subtypeMembers != null) {
-                return unionMembersHaveSuperType(subtype, supertype, subtypeMembers);
-            }
+        // Single get() rather than containsKey() + a redundant get() inside
+        // subTypeOfUnion: halves map traversals on the union dispatch path.
+        final IntArraySet supertypeMembers = unionTypes.get(supertype);
+        if (supertypeMembers != null) {
+            return subTypeOfUnion(subtype, supertype, supertypeMembers);
+        }
+        final IntArraySet subtypeMembers = unionTypes.get(subtype);
+        if (subtypeMembers != null) {
+            return unionMembersHaveSuperType(subtype, supertype, subtypeMembers);
         }
 
         subtype = superTypes[subtype];
--- a/exist-core-jmh/pom.xml
+++ b/exist-core-jmh/pom.xml
@@ -50,6 +50,16 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>exist-index-lucene</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-core</artifactId>
+            <version>${lucene.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.openjdk.jmh</groupId>
             <artifactId>jmh-core</artifactId>
             <version>${jmh.version}</version>
@@ -59,6 +69,11 @@
             <artifactId>jmh-generator-annprocess</artifactId>
             <version>${jmh.version}</version>
             <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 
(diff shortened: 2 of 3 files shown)


exist-core/src/main/java/org/exist/xquery/value/Type.java
```diff
@@ -531,11 +531,21 @@ public static boolean subTypeOf(int subtype, final int supertype) {
...
-        if (unionTypes.containsKey(supertype)) {
-            return subTypeOfUnion(subtype, supertype);
-        }
-        if (unionTypes.containsKey(subtype)) {
-            return unionMembersHaveSuperType(subtype, supertype);
+        // Skip both union-type lookups when no union types are registered.
+        // The static initialiser registers NUMERIC and ERROR so the guarded
+        // branch is taken in current builds; the isEmpty() check protects
+        // against future evolution where union registration becomes optional.
+        if (!unionTypes.isEmpty()) {
```
line-o: I cannot think of a justification to keep this guard. The registered union types are needed to be spec compliant. This will not become optional.

joewiz: [This response was co-authored with Claude Code. -Joe]

Agreed — done in `e30cdadeeb`. The `isEmpty()` guard came from an over-cautious framing in the originating tasking; you're right that spec compliance keeps the union types non-optional. Removed the guard, kept the `containsKey + redundant get` collapse on the union dispatch path which is the real perf win here.

line-o:
Please rebase

joewiz:

Rebased onto current `develop` (`4f09d0accc`). One conflict resolved cleanly in `exist-core-jmh/pom.xml` — the annotation-processor-paths config has consolidated since this branch was last touched (PR #6296 + #6314 merged). I took develop's version; `TypeSubTypeOfBenchmark` builds clean against it.

Note: `mvn install -pl exist-core-jmh -am` fails on `ReindexDeleteStrategyBenchmark` (missing Lucene deps) — but that's a pre-existing develop-side issue from a recent merge, not introduced here. `exist-core` itself builds clean and `TypeSubTypeOfBenchmark` compiles fine in isolation.

Force-pushed `1f2c018d9f`. Ready for re-review.

line-o (changes requested):
Rebase didn't go completely right. All lines removed from exist-core-jmh/pom.xml need to be added again.

joewiz:

Re-added the dropped pom.xml lines. New tip: `b8e2f65a71`. `exist-core-jmh` now builds clean against the consolidated annotation-processor-paths config from #6296 + #6314 — reset the file to develop's content rather than re-derive it.

exist-core/src/main/java/org/exist/xquery/value/Type.java
```diff
@@ -531,11 +531,15 @@ public static boolean subTypeOf(int subtype, final int supertype) {
...
-            return subTypeOfUnion(subtype, supertype);
+        // Single get() rather than containsKey() + a redundant get() inside
+        // subTypeOfUnion: halves map traversals on the union dispatch path.
+        final IntArraySet supertypeMembers = unionTypes.get(supertype);
+        if (supertypeMembers != null) {
+            return subTypeOfUnion(subtype, supertype, supertypeMembers);
         }
-        if (unionTypes.containsKey(subtype)) {
-            return unionMembersHaveSuperType(subtype, supertype);
+        final IntArraySet subtypeMembers = unionTypes.get(subtype);
```
line-o, line 540: Looking at this now for the umpteenth time:
When we try to determine if **union type A** is a **subtype of type B** I would argue we need a separate check.
One that checks that _all members_ of A are subtypes of B. 

It boils down to the example: Is `numeric` a subtype of `xs:decimal` or `xs:double` or both? And why is that the case?

What happens with a uniontype of `xs:integer` and `xs:string`?

line-o: And just to be clear: This is a separate issue that I will open once we agree it is one. Not blocking this PR!

line-o (approved):
👍🏼

duncdrum (approved):
It's been a long day, but yolo looks good. Thx @joewiz

Issue #6326, opened 2026-05-09
[ci] XQTS perf-comparison report can attribute develop fixes / runner flakes to PRs


joewiz:
## Summary

The XQTS comparison report attached to perf-PR review tables can produce misleading "newly passing / newly failing" attributions for two reasons unrelated to the PR under review:

1. **Baseline drift** — when the baseline `develop` SHA captured for comparison is *older* than the PR branch's merge-base, intervening fixes on `develop` get misattributed to the PR. (The reverse — baseline NEWER than merge-base — also pollutes "newly failing".)

2. **Single-run flakiness** — some XQTS tests are non-deterministic across runs. If the baseline and PR runs each happen to roll different outcomes, the diff lists the test in the changed-set even though no code touched it.

Surfaced by line-o on PR #6315 (`map-contains-017` reported as newly passing). Investigation found neither structural drift (the PR's merge-base 414527b00, May 7 < baseline-develop 3d17293f15, May 8) nor a real fix on develop — the test is simply flaky, passing in 1/8 recent develop runs and failing in 7/8.

The underlying flakiness driver appears to be `op:same-key` for `xs:duration('P1Y')` vs `xs:yearMonthDuration('P12M')` routed through `AbstractMapType.sameKey` → `FunDeepEqual.deepEquals`, which deserves its own follow-up issue. But the methodological problem — that any of our recent perf-PR XQTS delta tables may include similar noise — is broader.

## Proposed mitigations

**A. Merge-base-aligned baselines.** Capture the baseline run against the PR's merge-base, not against an arbitrary recent `develop`. The comparison then attributes only differences introduced by the PR's diff itself.

```bash
mb=$(git merge-base origin/develop HEAD)
git checkout "$mb"
mvn install -DskipTests ...
.xqts-runner/run-batched.sh --xqts-version 3.1 --output-dir results/baseline-merge-base
git checkout -
```

This eliminates baseline drift contamination entirely.

**B. N-run flake filtering.** Run each variant N times (e.g., N=3) and only flag a test as changed when the majority outcome differs. A single-run roll on a flaky test no longer pollutes the comparison.

Cost: 3× the runtime per comparison. Mitigation: only run the per-test-set "below 80%" sets multiple times, leaving the high-pass-rate bulk on a single run.

**C. Attribution column** in the comparison output. Add per-changed-test annotation: "develop fix at commit X" / "PR introduces" / "flaky over N develop runs". Doesn't reduce noise but makes it visible.

[...]

PR #6332, opened 2026-05-10
[ci] compare-results.xslt: surface runner-JAR drift warning when build SHAs differ

joewiz:
## Summary

Adds a `cr:warnings` block to the comparison report that fires when the previous and current XQTS runs were produced by different runner JAR builds. Without this, develop-branch deltas, runner-source bugfixes, and shaded-`exist-core` differences all silently appear as if they were caused by the PR's source change.

Partially addresses #6326 (mitigation A only). Mitigations B (runner version pin) and C (N-run flake filtering) remain open.

Companion runner PR (the producer of the metadata this consumes): https://github.com/eXist-db/exist-xqts-runner/pull/51

## Why

A 2026-05-09 investigation found the same eXist SHA producing **+831 XQTS passes and a 25x wallclock improvement** when re-run against a different runner JAR build (the runner-side `applyVersionHint` cap-at-3.1 patch). The existing comparison report attributed those deltas to the PR's source change, which is wrong. This makes runner-JAR drift directly attributable rather than masquerading as engine improvements.

A second drift mode surfaced during PR #6331's perf-6322 investigation: locally-built runner JARs shade `exist-core`, so two runs of the same runner-source git SHA can still differ if assembled from different `exist-core` snapshots. The XSLT detects both modes.

## What changed

`exist-xqts/src/main/xslt/compare-results.xslt`:

- Reads `runner-info.xml` from each run's output dir root (two levels up from the junit-data path). The metadata is emitted by the companion `exist-xqts-runner` change; older runs predating that change will have the file absent.
- Emits a `cr:warning kind="runner-drift"` element when the runner JAR's git-sha or sha256 differs between the runs, including the full `runner-jar` payloads from each side for transparency.
- Emits a `cr:warning kind="embedded-exist-core-drift"` when the runner JAR appears identical but the shaded `exist-core` version differs.
- Graceful degradation: when either side's `runner-info.xml` is missing or malformed (`doc-available` returns false), no warning is emitted and the rest of the comparison runs unchanged. Older baselines and contributor environments without the runner-side patch keep working with no behaviour change.

`ci-xqts.yml` needs no edit -- the artifact upload uploads the entire `/tmp/xqts-output` dir and `runner-info.xml` round-trips automatically.

### Sample output (runner-drift case)

```xml
<cr:comparison xmlns:cr="http://exist-db.org/exist-xqts/compare-results">
   <cr:warnings>
      <cr:warning kind="runner-drift">
         <cr:summary>Runner JAR build SHA or sha256 differs between the previous and current XQTS runs. Test deltas may include runner-side effects unrelated to this PR. See https://github.com/eXist-db/exist/issues/6326.</cr:summary>
         <cr:previous>
            <runner-jar xmlns="http://exist-db.org/exist-xqts-runner/runner-info">
               <git-sha>aaaa1111...</git-sha>
               ...
            </runner-jar>
         </cr:previous>
         <cr:current>...</cr:current>
      </cr:warning>
   </cr:warnings>
   <cr:previous>...</cr:previous>
   <cr:current>...</cr:current>
   <cr:change>...</cr:change>
</cr:comparison>
```

## Test plan

Hand-tested locally against Saxon-HE 9.9 (the same version CI invokes), with synthetic `runner-info.xml` files in five scenarios:

- [x] Different runner-jar git-sha -> emits `cr:warning kind="runner-drift"` with both sides' payloads.
- [x] Identical `runner-info.xml` files -> no warning emitted.
- [x] `runner-info.xml` absent on one side -> no warning, no error.
- [x] Malformed `runner-info.xml` -> no warning, no error (`doc-available` returns false).
- [x] Same runner JAR git-sha + sha256 but different embedded `exist-core` version -> emits `cr:warning kind="embedded-exist-core-drift"`.
- [ ] End-to-end smoke test on next CI run will produce a comparison report with no warning (matching SHAs from the same workflow build).

## Risk and rollback

The change is purely additive. Reverting removes the warning element but doesn't break existing comparison output -- the per-test-comparison structure is unchanged, so any downstream consumer of `comparison-results.xml` remains compatible. The two PRs (this one and the runner-side) can land in either order; the XSLT degrades to a no-op when `runner-info.xml` is absent.

### DIFF
--- a/exist-xqts/src/main/xslt/compare-results.xslt
+++ b/exist-xqts/src/main/xslt/compare-results.xslt
@@ -25,7 +25,8 @@
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
     xmlns:xs="http://www.w3.org/2001/XMLSchema"
     xmlns:cr="http://exist-db.org/exist-xqts/compare-results"
-    exclude-result-prefixes="xs"
+    xmlns:ri="http://exist-db.org/exist-xqts-runner/runner-info"
+    exclude-result-prefixes="xs ri"
     version="2.0">
 
     <xsl:param name="xqts.previous.junit-data-path" as="xs:string" required="yes"/>
@@ -43,8 +44,16 @@
                 <xsl:sequence select="cr:new-changes($previous-summary/cr:results, $current-summary/cr:results, .)"/>
             </xsl:for-each>
         </xsl:variable>
+        <xsl:variable name="previous-runner-info" as="document-node()?" select="cr:load-runner-info($xqts.previous.junit-data-path)"/>
+        <xsl:variable name="current-runner-info" as="document-node()?" select="cr:load-runner-info($xqts.current.junit-data-path)"/>
         <xsl:document>
             <cr:comparison>
+                <xsl:variable name="warnings" as="element(cr:warning)*" select="cr:drift-warnings($previous-runner-info, $current-runner-info)"/>
+                <xsl:if test="exists($warnings)">
+                    <cr:warnings>
+                        <xsl:sequence select="$warnings"/>
+                    </cr:warnings>
+                </xsl:if>
                 <cr:previous>
                     <xsl:copy select="$previous-summary/cr:results">
                         <xsl:copy-of select="@*"/>
@@ -106,10 +115,10 @@
         <xsl:param name="previous-results" as="element(cr:results)" required="yes"/>
         <xsl:param name="current-results" as="element(cr:results)" required="yes"/>
         <xsl:param name="attr-name" as="xs:string" required="yes"/>
-        
+
         <xsl:variable name="previous-attr" select="$previous-results/@*[local-name(.) eq $attr-name]"/>
         <xsl:variable name="current-attr" select="$current-results/@*[local-name(.) eq $attr-name]"/>
-        
+
         <xsl:attribute name="{$attr-name}" select="$current-attr - $previous-attr"/>
         <xsl:choose>
             <xsl:when test="$attr-name eq 'pass'">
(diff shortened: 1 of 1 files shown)


duncdrum (approved):

dizzzz (approved):

line-o:
I am all in favor of this change but this pr should include the updated dependency to a runner that exposes the new values. Or am I overly cautious?

joewiz:

@line-o You're not being overly cautious — you're right, and it's worth spelling out because the current state is a bit worse than "needs a version bump."

What I found tracing the producer/consumer coupling:

- This PR only touches `compare-results.xslt` (the **consumer**). The **producer** of `runner-info.xml` is eXist-db/exist-xqts-runner#51, merged 2026-07-09.
- `exist-xqts/pom.xml` pins the runner as a floating `2.0.0-SNAPSHOT`.
- But the newest `2.0.0-SNAPSHOT` in the GitHub Packages registry is dated **2025-12-09** — it predates eXist-db/exist-xqts-runner#51 by seven months, so it does **not** emit `runner-info.xml`.
- The reason the merge didn't refresh it: the runner's only CI workflow (`ci.yml`) is build-and-test — `sbt assembly` plus a smoke run — with no `publish`/`deploy` step. The `2.0.0-SNAPSHOT` (Dec 2025) and the `2.0.0-RC1`/`RC2` artifacts were published out-of-band, not on develop merges.

So merging this XSLT as-is wouldn't break anything — `cr:load-runner-info` returns the empty sequence when the file is missing, and the warning simply never renders — but the feature would be dead-on-arrival: CI resolves the Dec-2025 runner, `runner-info.xml` is never produced, and the drift warning can't fire.

The good news is the producer side is ready: the runner's `develop` tip already includes eXist-db/exist-xqts-runner#51 (it's the merge commit at HEAD), so a build cut from develop today emits `runner-info.xml`. What's missing is only the *publish* + the pin:

1. Publish a runner build that includes eXist-db/exist-xqts-runner#51 — I'd favor a tagged release (say `2.0.0-RC3`) over refreshing the mutable `2.0.0-SNAPSHOT`, since a floating snapshot is the very drift this warning exists to catch.
2. Pin `exist-xqts/pom.xml` to that fixed version in this PR.

That firm pin is also **mitigation B** from #6326 (runner version pin), which this PR had scoped out — so doing it here folds B and A together rather than leaving B for later.

Since you cut the last runner release (2.0.0-RC2), that call is yours: would you like to tag a `2.0.0-RC3` off develop, or would you prefer I prepare the release (and the `exist-xqts/pom.xml` pin here) for you to publish? Either works for me — I just didn't want to reach past you on the runner release. Whichever we pick, I'll hold this PR until the pinned runner is published. Separately, it'd be worth adding a publish step to the runner's CI so producer artifacts track develop automatically and we don't re-hit this gap — happy to open that as a follow-up.

line-o (changes requested):
According to @duncdrum any file that does something with XQTS that is not run as part of the maven build cannot stay in the project but _has to_ move to `.github` or be incorporated into the maven build.

duncdrum:
@line-o nice try, new file added to the reactor vs. modifying existing files. No jar extraction dance happening here. I can speak for myself.

line-o (approved):
I agreed to address the change in #6564

PR #6333, opened 2026-05-10
[bugfix] DurationValue: deterministic hashCode for cross-type op:same-key (closes #6327)

### LINKED ISSUE #6327: [bug] XQTS map-contains-017 is flaky (1/8 pass rate); xs:duration vs xs:yearMonthDuration op:same-key non-deterministic
## Summary

XQTS test `map-contains-017` is non-deterministic on develop, passing in roughly 1 of every 8 recent runs. Surfaced while investigating a separate XQTS-comparison methodology question on PR #6315 (https://github.com/eXist-db/exist/pull/6315#issuecomment-4413707520) where it appeared as "newly passing" against a baseline that happened to roll fail.

## Root cause hypothesis

The test exercises `op:same-key` for `xs:duration('P1Y')` vs `xs:yearMonthDuration('P12M')`. eXist routes this through `AbstractMapType.sameKey` → `FunDeepEqual.deepEquals`. Both values represent the same duration but as different XDM types, and the comparison appears to be sensitive to typed-value precision / canonicalisation that varies run-to-run (likely a `BigDecimal` scale / `DurationValue` normalisation issue).

## Reproducer

```bash
cd $WORKTREE/.xqts-runner
for i in 1 2 3 4 5 6 7 8; do
    ./run-batched.sh --xqts-version 3.1 --filter "map-contains-017" --output-dir /tmp/mc017-run-$i 2>&1 | tail -1
done
grep -h "<failure\|tests=" /tmp/mc017-run-*/test-results/*.xml | head -16
```

Expected: ~1 pass / 7 fails out of 8.

## Suspected fix area

Either:

- `org.exist.xquery.value.DurationValue.equals` / `getCanonicalDuration` — make canonicalisation deterministic across `xs:duration` and `xs:yearMonthDuration` arguments
- `org.exist.xquery.functions.fn.FunDeepEqual.deepEquals` for atomic duration values — explicit case for cross-type duration comparison via `xs:integer` months count

## Test plan once fixed

- Run `map-contains-017` 20× in a row; expect 20/20 pass.
- Add a JUnit test that asserts `op:same-key` returns true for the typed-value pair `xs:duration('P1Y')` / `xs:yearMonthDuration('P12M')` AND for the reverse argument order.
- Full XQTS QT4 + XQ 3.1 + FTTS — duration handling touches many tests.

## Notes

This is a real underlying bug, not a test-runner artifact. Filing separately from the broader XQTS-comparison methodology issue so the test fix and the methodology fix can land independently.

joewiz:
## Summary

`DurationValue` had no `hashCode()` override, so the Bifurcan map in `MapType` (keyed on `AtomicValue::hashCode`) gave `xs:duration('P1Y')` and `xs:yearMonthDuration('P12M')` different hash buckets despite `op:same-key` returning `true`. `map:contains` therefore missed the cross-type lookup whenever the two identity-based hashes happened not to collide modulo the bucket count - the ~1-in-8 pass rate observed on XQTS `map-contains-017`.

## Root cause

`AbstractMapType.MapType` builds a Bifurcan map with:

```java
new Map<>(AtomicValue::hashCode, (k1, k2) -> sameKey(collator, k1, k2));
```

`sameKey` for the `xs:duration` family routes through `FunDeepEqual.deepEquals`, which would correctly equate `P1Y` with `P12M` - **if the lookup ever made it that far**. Bifurcan first probes by hash, so a hash mismatch short-circuits before `sameKey` is ever called. With `DurationValue` inheriting `Object.hashCode` (identity-based), every instance had a different hash and the lookup almost always landed in the wrong bucket.

## What changed

`DurationValue.hashCode()` (new) - hashes on the canonical `(signed total months, signed total seconds)` tuple computed from `monthsValueSigned()` / `secondsValueSigned()`. This tuple is invariant across `xs:duration` / `xs:yearMonthDuration` / `xs:dayTimeDuration` for equal-value durations:

| Value | months | seconds |
|---|---:|---:|
| `xs:duration('P1Y')` | 12 | 0 |
| `xs:yearMonthDuration('P12M')` | 12 | 0 |
| `xs:duration('P1D')` | 0 | 86400 |
| `xs:dayTimeDuration('PT24H')` | 0 | 86400 |
| `xs:duration('-P1Y')` | -12 | 0 |
| `xs:duration('P0D')` / `P0M` / `PT0S` | 0 | 0 |

`BigDecimal` scale is normalized (`stripTrailingZeros`, with explicit zero handling) so that `BigDecimal("0")` and `BigDecimal("0.000")` hash identically.

`DurationValue.equals(Object)` is also rewritten to read the same canonical tuple instead of delegating to JDK `Duration.equals`. This removes a potential `equals`/`hashCode` contract divergence and matches the canonical-comparison behaviour `op:duration-equal` already implements.

## Spec references

- W3C XPath F&O 3.1 14.2.2 [`op:same-key`](https://www.w3.org/TR/xpath-functions-31/#func-same-key) - `xs:duration` family is compared via `fn:deep-equal`, which uses canonical value comparison.
- W3C XPath F&O 3.1 9.4.2 `op:duration-equal` - the canonical-month / canonical-second comparison this fix mirrors.

## Test plan

- New JUnit tests in `DurationTest`:
  - `hashCodeEqualForDurationAndYearMonthDuration` - issue case
  - `hashCodeEqualForDurationAndDayTimeDuration` - dayTime analogue
  - `hashCodeEqualForZeroDurations` - all three zero forms hash equal
  - `hashCodeEqualForNegativeDurations` - signed parity
  - `hashCodeDifferentForUnequalDurations` - sanity: distinct values still hash apart
  - `equalsCommutativeAcrossDurationSubtypes` - cross-type `equals()` is reflexive
  - `hashCodeDeterministicAcrossInstances` - 100-iter loop, all produce the same hash
- New XQuery tests in `xquery/maps/maps.xqm` - replicate the `map-contains-017` pattern at the XQuery surface, in both argument orders, plus the `xs:duration` <-> `xs:dayTimeDuration` analogue.
- Without the fix, the 5 hashCode-equality JUnit tests and all 4 new XQuery `map:contains` tests fail (verified by stashing the fix and re-running). With the fix, all pass.
- Full `mvn test -pl exist-core` gate: **6603 tests, 0 failures, 0 errors**.
- Targeted reruns: `*Duration*`, `*DateTime*`, `*DeepEqual*`, `*MapTest*`, `xquery.maps.MapTests`, `BifurcanMapTest` - 335 tests, all green.
- Full XQTS QT4 / XQ 3.1 / FTTS not re-run locally (worktree XQTS-runner build had an unrelated SNAPSHOT-resolution issue); CI will exercise these. The change is surgical (hashCode addition + equals canonicalization, no comparison/arithmetic logic touched), so XQTS regressions outside the duration-key path should not be expected.

## Risk

The `hashCode` addition strictly tightens the existing contract - durations that previously had distinct identity hashes now hash by canonical value, which is what `sameKey` already requires. The `equals` rewrite returns the same answer as JDK `Duration.equals` for definite cases (P1Y == P12M, P30D != P1M) and the same `false` answer for indeterminate cases (P30D vs P1M, P365D vs P1Y).

Closes https://github.com/eXist-db/exist/issues/6327
Refs https://github.com/eXist-db/exist/issues/6122

### DIFF
--- a/exist-core/src/test/java/org/exist/xquery/value/DurationTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/value/DurationTest.java
@@ -178,4 +178,69 @@ public class DurationTest extends AbstractTimeRelatedTestCase {
         checkMinMaxFails(dv2, dv3);
         checkMinMaxFails(dv3, dv2);
     }
+
+    // --- Cross-type canonicalisation (issue #6327) ---
+    // op:same-key requires equal hashCodes for equal-value durations across xs:duration
+    // family subtypes; the bifurcan map in MapType keys on AtomicValue::hashCode and the
+    // map-contains-017 XQTS test exercises this path.
+
+    @Test
+    public void hashCodeEqualForDurationAndYearMonthDuration() throws XPathException {
+        final DurationValue d = new DurationValue("P1Y");
+        final YearMonthDurationValue ymd = new YearMonthDurationValue("P12M");
+        assertEquals(d.hashCode(), ymd.hashCode());
+        assertEquals(ymd.hashCode(), d.hashCode());
+    }
+
+    @Test
+    public void hashCodeEqualForDurationAndDayTimeDuration() throws XPathException {
+        final DurationValue d = new DurationValue("P1D");
+        final DayTimeDurationValue dtd = new DayTimeDurationValue("PT24H");
+        assertEquals(d.hashCode(), dtd.hashCode());
+    }
+
+    @Test
+    public void hashCodeEqualForZeroDurations() throws XPathException {
+        final DurationValue d = new DurationValue("P0D");
+        final YearMonthDurationValue ymd = new YearMonthDurationValue("P0M");
+        final DayTimeDurationValue dtd = new DayTimeDurationValue("PT0S");
+        assertEquals(d.hashCode(), ymd.hashCode());
+        assertEquals(d.hashCode(), dtd.hashCode());
+        assertEquals(ymd.hashCode(), dtd.hashCode());
+    }
+
+    @Test
+    public void hashCodeEqualForNegativeDurations() throws XPathException {
+        final DurationValue d = new DurationValue("-P1Y");
+        final YearMonthDurationValue ymd = new YearMonthDurationValue("-P12M");
+        assertEquals(d.hashCode(), ymd.hashCode());
+    }
+
+    @Test
+    public void hashCodeDifferentForUnequalDurations() throws XPathException {
+        // sanity: unequal durations should generally hash differently
+        final DurationValue p1y = new YearMonthDurationValue("P1Y");
+        final DurationValue p2y = new YearMonthDurationValue("P2Y");
+        org.junit.Assert.assertNotEquals(p1y.hashCode(), p2y.hashCode());
+    }
+
+    @Test
+    public void equalsCommutativeAcrossDurationSubtypes() throws XPathException {
+        final DurationValue d = new DurationValue("P1Y");
+        final YearMonthDurationValue ymd = new YearMonthDurationValue("P12M");
+        org.junit.Assert.assertEquals(d, ymd);
+        org.junit.Assert.assertEquals(ymd, d);
+    }
+
+    @Test
+    public void hashCodeDeterministicAcrossInstances() throws XPathException {
+        // 100-iter loop exposes any non-determinism (e.g., identity-hash leaks)
+        final int expected = new DurationValue("P1Y").hashCode();
+        for (int i = 0; i < 100; i++) {
+            final DurationValue d = new DurationValue("P1Y");
+            final YearMonthDurationValue ymd = new YearMonthDurationValue("P12M");
+            assertEquals("iter " + i + ": DurationValue", expected, d.hashCode());
+            assertEquals("iter " + i + ": YearMonthDurationValue", expected, ymd.hashCode());
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


exist-core/src/test/java/org/exist/xquery/value/DurationTest.java
```diff
@@ -178,4 +178,69 @@ public void minMaxFail() throws XPathException {
...
+        final YearMonthDurationValue ymd = new YearMonthDurationValue("-P12M");
+        assertEquals(d.hashCode(), ymd.hashCode());
+    }
+
+    @Test
+    public void hashCodeDifferentForUnequalDurations() throws XPathException {
+        // sanity: unequal durations should generally hash differently
+        final DurationValue p1y = new YearMonthDurationValue("P1Y");
+        final DurationValue p2y = new YearMonthDurationValue("P2Y");
+        org.junit.Assert.assertNotEquals(p1y.hashCode(), p2y.hashCode());
```
line-o: Import `org.junit.Assert.assertNotEquals` and use short form here to make codacy / and us happy.

line-o (changes requested):
Please address Codacy issues.

line-o:
refs #6122

joewiz:

- Imported `assertNotEquals` and switched the cross-subtype block to short-form `assertEquals` / `assertNotEquals` (tip `824dda8ff0`).
- Added `Refs #6122` to the PR description.
- `mvn test -pl exist-core -Dtest=DurationTest` — 26/26 green.

`codacy-cli analyze --tool pmd` on the touched files surfaces only pre-existing warnings in `DurationValue` (parameter-reassignment in 2017's `nullIfZero`/`zeroIfNull` helpers, plus the one-decl-per-line on line 173). None are in the diff for this PR; leaving them for a separate sweep so this PR's scope stays surgical.

line-o (approved):
👍🏼

reinhapa (approved):

PR #6336, opened 2026-05-10
[bugfix] NumericValue + BooleanValue: hashCode contract for op:same-key (PR #6333 family)

joewiz:
## Summary

Closes the family of AtomicValue subclasses that violate the `equals`/`hashCode` contract under `op:same-key` semantics. PR #6333 fixed `xs:duration`; this sweep covers `xs:integer` / `xs:decimal` / `xs:double` / `xs:float` and `xs:boolean`.

Without these guards, Bifurcan-backed map operations (`map:contains`, `map:get`, ...) bucket spec-equal cross-type keys differently and silently mis-report containment. The diagnostic signature is the same "1-in-N pass rate" symptom that produced `map-contains-017` on `xs:duration`.

## Why

In `MapType.KEY_HASH_FN = AtomicValue::hashCode`, Bifurcan uses each value's `hashCode` for bucketing and only consults `sameKey` for equality within a bucket. If two spec-equal cross-type values hash to different buckets, the equality check is never reached.

### Confirmed contract violations on develop (b917e1ab1d)

| Pair | `equals` | `hashCode` |
|---|---|---|
| `IntegerValue(1)` vs `DecimalValue(1)` | true | 1 / 31 ❌ |
| `IntegerValue(1)` vs `DoubleValue(1.0)` | true | 1 / 1072693248 ❌ |
| `DecimalValue(1.0)` vs `DoubleValue(1.0)` | true | 31 / 1072693248 ❌ |
| `DoubleValue(1.0)` vs `FloatValue(1.0f)` | true | 1072693248 / 1065353216 ❌ |
| `DoubleValue(+INF)` vs `FloatValue(+INF)` | true | 2146435072 / 2139095040 ❌ |
| `new BooleanValue(true)` vs `new BooleanValue(true)` | true | identity / identity ❌ |

## What changed

| File | Change |
|---|---|
| `NumericValue.java` | Add canonical `hashCode()`: NaN / ±INF sentinels, else `((DecimalValue) convertTo(DECIMAL)).getValue().hashCode()` (trailing-zero-stripped at construction). Equals already uses op:numeric-equal, so the contract now holds. |
| `IntegerValue.java`, `DecimalValue.java`, `DoubleValue.java`, `FloatValue.java` | Drop per-class `hashCode()` overrides; inherit the canonical form from `NumericValue`. |
| `BooleanValue.java` | Add `hashCode()` = `Boolean.hashCode(value)`. Non-singleton instances previously hashed by identity. |
| `HashCodeContractTest.java` (new) | 14 contract assertions across the numeric and boolean clusters; 100-iter determinism guard. |
| `maps.xqm` | 9 XQuery regression cases mirroring `map-contains-017` (numeric cross-type, NaN, infinities, non-singleton booleans). `mt:for-each` updated to sort its result — the previous assertion relied on Bifurcan bucket iteration order, which has no spec guarantee. |
| `custom-assertion.xqm` | Sort additional-keys diagnostic for determinism; one test expectation updated for new bucket distribution. |

## Inventory of AtomicValue subclasses surveyed

| Class | equals override | hashCode override | Verdict |
|---|---|---|---|
| `IntegerValue` | inherited from NumericValue (cross-type numeric-eq) | replaced by canonical NumericValue | **fixed** |
| `DecimalValue` | inherited | replaced | **fixed** |
| `DoubleValue` | inherited | replaced | **fixed** |
| `FloatValue` | inherited | replaced | **fixed** |
| `BooleanValue` | by-value | **added** | **fixed** |
| `DurationValue` (+ `OrderedDurationValue` / `YearMonthDurationValue` / `DayTimeDurationValue`) | by canonical tuple | (added in PR #6333) | scope of PR #6333 |
| `StringValue`, `AnyURIValue`, `UntypedAtomicValue` | by value | by value | already correct (codepoint-equal cluster delegates to `String.hashCode`) |
| `AbstractDateTimeValue` (+ all date/time/gXxx subclasses) | by `calendar` | by `calendar.hashCode` | already correct within-class; cross-type sameKey requires lexical match so no cross-type collision needed |
| `BinaryValue` (+ `BinaryValueFromBinaryString`, `BinaryValueFromFile`, `BinaryValueFromInputStream`, `Base64BinaryDocument`) | by `getClass()` + content | by stream-read hash | within-class consistent; cross-type (hexBinary vs base64Binary) returns false in `equals` and `fn:deep-equal` returns false too |
| `QNameValue` | by `qname` | by `qname.hashCode` | already correct |
| `FunctionReference`, `JavaObjectValue` | not used as map keys per spec | n/a | out of scope |

## Spec references

- [XPath/XQuery 3.1 op:same-key](https://www.w3.org/TR/xpath-functions-31/#func-same-key)
- [Bifurcan `Map` hashing contract](https://github.com/lacuna/bifurcan/blob/master/doc/comparison.md)
- PR #6333 - reference implementation against `DurationValue`
- `map-contains-017` - canonical reproduction shape

## Test plan

- [x] `HashCodeContractTest` (14/14 pass)
- [x] `MapTests` XQuery suite (134 pass, was failing `mt:for-each` due to bucket order)
- [x] `XQSuiteTests` (90 pass after `custom-assertion.xqm` determinism fix)
- [x] Full `mvn test -pl exist-core` under shared lock (6615 tests, no hashCode-related failures)
- [ ] XQTS QT4 / XQ 3.1 / FTTS - to be validated in CI; runner-JAR-shaded-exist-core hazard means local runs are unreliable for this change shape (cf. `project_xqts_runner_info_drift_warnings.md`)

## Risk and rollback

Hash bucket layout for numeric and boolean keys changes. Map iteration order is implementation-defined per spec; tests that asserted specific orderings have been updated to sort. Each value-class change is independently revertable.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/value/BooleanValue.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/BooleanValue.java
@@ -226,4 +226,9 @@ public class BooleanValue extends AtomicValue {
         }
         return false;
     }
+
+    @Override
+    public int hashCode() {
+        return Boolean.hashCode(value);
+    }
 }
--- a/exist-core/src/main/java/org/exist/xquery/value/DecimalValue.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/DecimalValue.java
@@ -496,11 +496,6 @@ public class DecimalValue extends NumericValue {
         }
     }
 
-    @Override
-    public int hashCode() {
-        return value.hashCode();
-    }
-
     //Copied from Saxon 8.8
 
     /* (non-Javadoc)
--- a/exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/DoubleValue.java
@@ -490,11 +490,6 @@ public class DoubleValue extends NumericValue {
         return getType() < other.getType() ? Constants.INFERIOR : Constants.SUPERIOR;
     }
 
-    @Override
-    public int hashCode() {
-        return Double.valueOf(value).hashCode();
-    }
-
     /**
      * Serializes to a ByteBuffer.
      * 8 bytes.
--- a/exist-core/src/main/java/org/exist/xquery/value/FloatValue.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/FloatValue.java
@@ -529,11 +529,6 @@ public class FloatValue extends NumericValue {
         }
     }
 
-    @Override
-    public int hashCode() {
-        return Float.valueOf(value).hashCode();
-    }
-
     /**
      * Serializes to a ByteBuffer.
      *
--- a/exist-core/src/main/java/org/exist/xquery/value/IntegerValue.java
+++ b/exist-core/src/main/java/org/exist/xquery/value/IntegerValue.java
@@ -562,11 +562,6 @@ public class IntegerValue extends NumericValue {
         }
     }
 
-    @Override
-    public int hashCode() {
-        return value.hashCode();
-    }
-
     //TODO(AR) this is not a very good serialization method, the size of the IntegerValue is unbounded and may not fit in 8 bytes.
     /**
      * Serializes to a byte array.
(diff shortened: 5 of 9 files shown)


line-o (approved):
LGTM!

reinhapa (approved):

PR #6337, opened 2026-05-10
[optimize] fn:deep-equal: streaming fast-path for persistent-DOM trees (closes #4050)

### LINKED ISSUE #4050: [BUG]
**Describe the bug**
We have tried to do XML compares using `xmldiff:compare()` and the XPath equivalent `deep-equal()`.

For comparing two equal documents of ~200Kb `xmldiff:compare()` on my machine taks 60msecs, `deep-equal()` over 9 seconds!

**Expected behavior**
The functions  vary slightly in interface, but just comparing two documents should roughly take the same amount of time. And if not 9 seconds is a bit too much I think.

**To Reproduce**
Unzip, import in eXist, execute `test-equals.xql` (I'm doing this from oXygen)

 - OS: W10
 - eXist-db version: 5.3
 - Java Version Amazon Corretto 1.8.0_202

**Additional context**
- How is eXist-db installed? Jar
- Any custom changes: various but none seem relevant (mostly external SQL database things)

joewiz:
## Summary

`fn:deep-equal` on persistent-DOM trees was ~24× slower than `xmldiff:compare` on PieterLamers' [TEST.zip](https://github.com/eXist-db/exist/files/7250882/TEST.zip) reproducer (Macbeth.xml vs Macbeth2.xml, ~3,550 elements each, byte-identical) — 5,170 ms vs 228 ms on develop tip `b917e1ab1d`. After this change the same comparison runs in **5–11 ms median** while preserving full `fn:deep-equal` semantics. `xmldiff:compare` is not a correctness-equivalent shortcut (it overlooks attribute changes, per [PieterLamers' 2026-05-08 follow-up](https://github.com/eXist-db/exist/issues/4050#issuecomment-4406332435)), so closing the gap had to keep attribute / namespace / collation comparison rules intact.

Closes https://github.com/eXist-db/exist/issues/4050

## Diagnosis

The 2026-05-08 paused investigation pointed at \"per-node persistent-DOM accessor cost\" (`getNamespaceURI`, `getLocalName`, `getAttributes`). The hot path is more specific than that. `StoredNode.getNextSibling()` (and `ElementImpl.getFirstChild()`) acquire a fresh broker per call **and** open a new `XMLStreamReader` on the parent that walks until it finds the requested sibling — an O(siblings) scan per step. `FunDeepEqual.compareContents` calls these per node, so the recursion is effectively quadratic in sibling count for a wide stored DOM.

Diagnostic XQuery on the same Macbeth corpus, on develop tip:

| Workload | Median |
|---|--:|
| `count(//*) + count(//*)` (both docs) | 9–14 ms |
| `count(distinct-values(//*/local-name()))` | 13–16 ms |
| `string-length(serialize(doc))` (forces full walk) | 13–24 ms |
| **`fn:deep-equal(doc, doc)` (legacy)** | **5,170 ms** |

The storage-layer floor for walking both documents is ≤ 24 ms. The legacy `fn:deep-equal` path is paying ~5 seconds in `getFirstChild` / `getNextSibling` overhead, not in the comparison work.

## What changed

A new `FunDeepEqualStreamingComparator` walks both subtrees via `IEmbeddedXMLStreamReader` (the persistent-DOM BTree stream reader) and compares events in lockstep:

- Element name comparison uses expanded QName (namespace URI + local name), code-point order.
- Attributes are gathered from both elements, `xmlns:*` declarations filtered, the remainder sorted by (namespace URI, local name) and compared positionally — matches `FunDeepEqual.compareAttributes`'s order-insensitive semantics. Attribute values use the supplied collator.
- `CHARACTERS` / `CDATA` / `SPACE` events are compared in document order with the supplied collator.
- `COMMENT` and `PROCESSING_INSTRUCTION` events are skipped per the [W3C XPath F&O 3.1 §15.3.1 fn:deep-equal](https://www.w3.org/TR/xpath-functions-31/#func-deep-equal) spec.

Dispatch lives in `FunDeepEqual.deepCompare`'s `DOCUMENT` and `ELEMENT` switch branches and only fires when both arguments are persistent `NodeProxy` instances with `getImplementationType() == PERSISTENT_NODE`. Memtree, atomic, attribute-as-top-level, text, map, and array shapes fall through to the legacy recursive path. Any `XMLStreamException` / `IOException` / `RuntimeException` from the streaming path is caught and logged at DEBUG; the legacy path runs, so correctness is preserved.

The `FunDeepEqual` static API gains broker-aware overloads (`deepCompareSeq(..., DBBroker)`, `deepCompare(..., DBBroker)`, `deepEqualsSeq(..., DBBroker)`); the existing 3-arg signatures delegate with `broker = null`, leaving callers like `SwitchExpression`, `GroupByClause`, `AbstractMapType`, and `FunSort` untouched.

## Reproducer

PieterLamers' TEST.zip (Macbeth.xml + Macbeth2.xml, ~3,550 elements each), JDK Zulu 21.38.21, eXist 7.0.0-SNAPSHOT, 5-trial median:

| Variant | Median | Notes |
|---|--:|---|
| current develop (no fix) | 5,170 ms | reporter's TEST.zip |
| `xmldiff:compare` (per @adamretter, 2022) | 228 ms | not correctness-equivalent |
| **this fix** | **6 ms** | ~2× over the serialize() floor; ~38× faster than xmldiff:compare |

Synthetic 10k-element stored corpus (`FunDeepEqualPerformanceTest.deepEqualOnStoredEqualDocsIsFast`):

| Variant | Median | Notes |
|---|--:|---|
| current develop | ~2,521 ms | from 2026-05-08 measurement |
| **this fix** | **124 ms** | ~20× speedup |

Memtree case (`deepEqualOnLargeEqualTreesIsFast`) unchanged at ~46 ms; the streaming path does not apply to in-memory nodes.

## Correctness

- 78 existing `*DeepEqual*` JUnit tests pass: `DeepEqualTest` (63), `FunDeepEqualPerformanceTest` (14), `FunDeepEqual4050ReproTest` (1).
- Wider 1,230-test sweep across `XPathQueryTest`, `XQuery3Tests`, `FunSortTest`, `GroupByTest`, `*MapType*`, `*Serialization*` — all green.
- Full `mvn test -pl exist-core`: **6,607 tests, 0 failures, 0 errors, 106 skipped (pre-existing)**.

XQTS conformance has not been re-run for this branch — `fn:deep-equal` is exercised pervasively in QT4 / XQ 3.1, and CI's runner job will report any conformance delta. I'll re-run locally before merge if a reviewer wants confirmation in advance.

## Out of scope

- Schema-aware typed-value comparison. The streaming path treats values as `xs:untypedAtomic` (string equality). The reporter's reproducer is untyped; if a typed-value edge case surfaces in the test gate it'll fall back to legacy. (None has so far across 6,607 tests.)
- In-memory tree caching of stored documents — broader optimisation that would benefit many functions besides `fn:deep-equal`. Worth a separate design pass.
- Documents with leading comments or processing instructions before the root element. The `documentRoot` helper requires the document's first stored child to be an element; non-element first child triggers legacy fallback. (Macbeth.xml and the synthetic corpus both have a single root element.)

## Test plan

- [x] `mvn test -pl exist-core -Dtest='*DeepEqual*'` — 78 tests pass.
- [x] `mvn test -pl exist-core` (full module via shared lock script) — 6,607 tests, 0 failures.
- [x] PieterLamers' Macbeth.xml reproducer: median 6 ms (was 5,170 ms).
- [x] Synthetic 10k-element stored corpus: 124 ms (was ~2,521 ms).
- [x] Memtree path unchanged: 46 ms.
- [x] Codacy PMD on changed files (the NPath warnings are on the existing `deepCompare` method's structure; not introduced here).
- [ ] CI XQTS QT4 / XQ 3.1 / FTTS deltas — pending CI run.

## Spec / context references

- W3C XPath F&O 3.1 §15.3.1 `fn:deep-equal`: <https://www.w3.org/TR/xpath-functions-31/#func-deep-equal>
- 2026-05-08 paused diagnosis comment: <https://github.com/eXist-db/exist/issues/4050#issuecomment-4406332435>
- `IEmbeddedXMLStreamReader` (the BTree stream reader this fix builds on): `exist-core/src/main/java/org/exist/stax/IEmbeddedXMLStreamReader.java`

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunDeepEqual.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunDeepEqual.java
@@ -223,178 +223,189 @@ public class FunDeepEqual extends CollatingFunction {
 
         try {
             if (item1.getType() == Type.ARRAY_ITEM || item2.getType() == Type.ARRAY_ITEM) {
-                if (item1.getType() != item2.getType()) {
-                    return Constants.INFERIOR;
-                }
-                final ArrayType array1 = (ArrayType) item1;
-                final ArrayType array2 = (ArrayType) item2;
-                final int array1Size = array1.getSize();
-                final int array2Size = array2.getSize();
-                if (array1Size == array2Size) {
-                    for (int i = 0; i < array1.getSize(); i++) {
-                        final int comparison = deepCompareSeq(array1.get(i), array2.get(i), collator, broker);
-                        if (comparison != Constants.EQUAL) {
-                            return comparison;
-                        }
-                    }
-                    return Constants.EQUAL;
-                } else {
-                    return array1Size < array2Size ? Constants.INFERIOR : Constants.SUPERIOR;
-                }
+                return compareArrayItems(item1, item2, collator, broker);
             }
-
             if (item1.getType() == Type.MAP_ITEM || item2.getType() == Type.MAP_ITEM) {
-                if (item1.getType() != item2.getType()) {
-                    return Constants.INFERIOR;
-                }
-                final AbstractMapType map1 = (AbstractMapType) item1;
-                final AbstractMapType map2 = (AbstractMapType) item2;
-                final int map1Size = map1.size();
-                final int map2Size = map2.size();
-
-                if (map1Size == map2Size) {
-                    for (final IEntry<AtomicValue, Sequence> entry1 : map1) {
-                        if (!map2.contains(entry1.key())) {
-                            return Constants.SUPERIOR;
-                        }
-
-                        final int comparison = deepCompareSeq(entry1.value(), map2.get(entry1.key()), collator, broker);
-                        if (comparison != Constants.EQUAL) {
-                            return comparison;
-                        }
-                    }
-                    return Constants.EQUAL;
-                } else {
-                    return map1Size < map2Size ? Constants.INFERIOR : Constants.SUPERIOR;
-                }
+                return compareMapItems(item1, item2, collator, broker);
             }
 
             final boolean item1IsAtomic = Type.subTypeOf(item1.getType(), Type.ANY_ATOMIC_TYPE);
             final boolean item2IsAtomic = Type.subTypeOf(item2.getType(), Type.ANY_ATOMIC_TYPE);
             if (item1IsAtomic || item2IsAtomic) {
-                if (!item1IsAtomic) {
-                    return Constants.SUPERIOR;
-                }
-
-                if (!item2IsAtomic) {
-                    return Constants.INFERIOR;
-                }
-
-                try {
-                    final AtomicValue av = (AtomicValue) item1;
-                    final AtomicValue bv = (AtomicValue) item2;
-                    if (Type.subTypeOfUnion(av.getType(), Type.NUMERIC) &&
-                            Type.subTypeOfUnion(bv.getType(), Type.NUMERIC)) {
-                        //or if both values are NaN
-                        if (((NumericValue) item1).isNaN() && ((NumericValue) item2).isNaN()) {
-                            return Constants.EQUAL;
-                        }
-                    }
-
-                    return ValueComparison.compareAtomic(collator, av, bv);
-                } catch (final XPathException e) {
-                    if (logger.isTraceEnabled()) {
-                        logger.trace(e.getMessage());
-                    }
-                    return Constants.INFERIOR;
-                }
+                return compareAtomicItems(item1, item2, item1IsAtomic, item2IsAtomic, collator);
             }
 
-            if (item1.getType() != item2.getType()) {
-                return Constants.INFERIOR;
-            }
-            final NodeValue nva = (NodeValue) item1;
-            final NodeValue nvb = (NodeValue) item2;
-            // NOTE(AR): intentional reference equality check
-            if (nva == nvb) {
-                return Constants.EQUAL;
-            }
-
-            try {
-                //Don't use this shortcut for in-memory nodes
-                //since the symbol table is ignored.
-                if (nva.getImplementationType() != NodeValue.IN_MEMORY_NODE &&
-                        nva.equals(nvb)) {
-                    return Constants.EQUAL;  // shortcut!
-                }
-            } catch (final XPathException e) {
-                // apparently incompatible values, do manual comparison
-            }
-
-            final Node node1;
-            final Node node2;
-            switch (item1.getType()) {
-                case Type.DOCUMENT:
-                    // GH-4050 fast path: persistent-DOM streaming comparator.
-                    // Falls through to legacy on stream/IO failure to preserve correctness.
-                    if (broker != null
-                            && nva instanceof NodeProxy npa
-                            && nvb instanceof NodeProxy npb
-                            && nva.getImplementationType() == NodeValue.PERSISTENT_NODE
-                            && nvb.getImplementationType() == NodeValue.PERSISTENT_NODE) {
-                        try {
-                            return FunDeepEqualStreamingComparator.compare(
-                                    broker, npa, npb, /*subtree=*/false, collator);
-                        } catch (final XMLStreamException | IOException | RuntimeException e) {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Streaming deep-equal fast path failed, falling back: "
-                                        + e.getMessage());
-                            }
-                        }
-                    }
-                    node1 = nva instanceof Node nnva ? nnva : ((NodeProxy) nva).getOwnerDocument();
-                    node2 = nvb instanceof Node nnvb ? nnvb : ((NodeProxy) nvb).getOwnerDocument();
-                    return compareContents(node1, node2, collator);
-
-                case Type.ELEMENT:
-                    if (broker != null
-                            && nva instanceof NodeProxy npea
-                            && nvb instanceof NodeProxy npeb
-                            && nva.getImplementationType() == NodeValue.PERSISTENT_NODE
-                            && nvb.getImplementationType() == NodeValue.PERSISTENT_NODE) {
-                        try {
-                            return FunDeepEqualStreamingComparator.compare(
-                                    broker, npea, npeb, /*subtree=*/true, collator);
-                        } catch (final XMLStreamException | IOException | RuntimeException e) {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Streaming deep-equal fast path failed, falling back: "
-                                        + e.getMessage());
-                            }
-                        }
-                    }
-                    node1 = nva.getNode();
-                    node2 = nvb.getNode();
-                    return compareElements(node1, node2, collator);
-
-                case Type.ATTRIBUTE:
-                    node1 = nva.getNode();
-                    node2 = nvb.getNode();
-                    final int attributeNameComparison = compareNames(node1, node2);
-                    if (attributeNameComparison != Constants.EQUAL) {
-                        return attributeNameComparison;
-                    }
-                    return safeCompare(node1.getNodeValue(), node2.getNodeValue(), collator);
-
-                case Type.PROCESSING_INSTRUCTION:
-                case Type.NAMESPACE:
-                    node1 = nva.getNode();
-                    node2 = nvb.getNode();
-                    final int nameComparison = safeCompare(node1.getNodeName(), node2.getNodeName(), null);
-                    if (nameComparison != Constants.EQUAL) {
-                        return nameComparison;
-                    }
-                    return safeCompare(nva.getStringValue(), nvb.getStringValue(), collator);
-
-                case Type.TEXT:
-                case Type.COMMENT:
-                    return safeCompare(nva.getStringValue(), nvb.getStringValue(), collator);
-
-                default:
-                    throw new UnexpectedItemTypeException(item1);
-            }
+            return compareNodeItems(item1, item2, collator, broker);
         } catch (final XPathException e) {
             logger.error(e.getMessage(), e);
             return Constants.INFERIOR;
+        }
+    }
+
+    private static int compareArrayItems(final Item item1, final Item item2,
+            @Nullable final Collator collator, @Nullable final DBBroker broker) throws XPathException {
+        if (item1.getType() != item2.getType()) {
+            return Constants.INFERIOR;
+        }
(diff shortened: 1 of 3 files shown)


exist-core/src/main/java/org/exist/xquery/functions/fn/FunDeepEqual.java
```diff
@@ -173,6 +197,26 @@ public static int deepCompareSeq(final Sequence sequence1, final Sequence sequen
...
+     * stream / IO failure the call falls through to the legacy recursive
+     * path so correctness is preserved.
+     *
+     * @param item1 first item.
+     * @param item2 second item.
+     * @param collator collation, or {@code null} for code-point.
+     * @param broker active broker, or {@code null} to disable the fast path.
+     * @return {@link Constants#EQUAL} / {@link Constants#INFERIOR} / {@link Constants#SUPERIOR}.
+     */
+    public static int deepCompare(final Item item1, final Item item2,
```
reinhapa, line 218: Address Codacy issue: _The method 'deepCompare(Item, Item, Collator, DBBroker)' has an NPath complexity of 8847362, current threshold is 200_

exist-core/src/main/java/org/exist/xquery/functions/fn/FunDeepEqualStreamingComparator.java
```diff
@@ -0,0 +1,336 @@
...
+            return null;
+        }
+        final org.w3c.dom.Node firstChild = doc.getFirstChild();
+        if (firstChild instanceof NodeHandle nh) {
+            return nh;
+        }
+        return null;
+    }
+
+    private static int walk(final IEmbeddedXMLStreamReader ra,
```
reinhapa, line 170: Address Codacy issue: _The method 'walk(IEmbeddedXMLStreamReader, IEmbeddedXMLStreamReader, boolean, Collator)' has an NPath complexity of 541, current threshold is 200_

exist-core/src/main/java/org/exist/xquery/functions/fn/FunDeepEqualStreamingComparator.java
```diff
@@ -0,0 +1,336 @@
...
+
+    private static int nextRelevantEvent(final IEmbeddedXMLStreamReader r)
+            throws XMLStreamException {
+        while (r.hasNext()) {
+            final int ev = r.next();
+            if (ev == XMLStreamConstants.COMMENT
+                    || ev == XMLStreamConstants.PROCESSING_INSTRUCTION) {
+                continue;
+            }
+            return ev;
```
reinhapa: Address Codacy issue: _Avoid using a branching statement as the last in a loop._

exist-core/src/test/java/org/exist/xquery/functions/fn/FunDeepEqualPerformanceTest.java
reinhapa, line 1: Address Codacy issue: _Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes._

reinhapa (changes requested):

joewiz:

Decomposed `FunDeepEqual.deepCompare` from NPath 8,847,362 into a dispatcher + 8 focused helpers (each well under 200), pushed as a separate `[refactor]` commit (`53db259237`) so it can be reviewed independently of the streaming fast-path. Per-type dispatch (atomic / array / map / node) at the top level; per-node-kind dispatch (DOCUMENT / ELEMENT / ATTRIBUTE / PI|NAMESPACE / TEXT|COMMENT) via a Java 21 switch expression inside `compareNodeItems`. Also extracted `tryStreamingCompare` so the GH-4050 fast-path guard isn't copy-pasted across the DOCUMENT and ELEMENT arms. FunDeepEqual JUnit (78/78) and XQuery3Tests (986/986) green; Macbeth.xml 6-9ms result preserved.

joewiz:

Addressed all four Codacy findings on this PR — note that the **fourth** (the `FunDeepEqual.deepCompare` 8.8M-NPath one) was already decomposed in commit `53db259237` earlier on this branch, so no new work was needed for it. New tip: `7b82bb49f8`.

- **`FunDeepEqualStreamingComparator.walk` (was NPath 721)**: decomposed into `walkStep` (arrow switch on event type) + `compareStartElements` / `compareEndElements` helpers, with a small `WalkState` holder so `depth` / `rootSeen` thread through without mutable closure tricks. `CHARACTERS` / `CDATA` / `SPACE` folded into one yield branch (they were identical). NPath well under 200.
- **`nextRelevantEvent` line 233 (branching-as-last-in-loop)**: rewritten as flag-controlled loop so neither `continue` nor `return ev` is the final statement of the loop body.
- **`FunDeepEqualPerformanceTest` field order**: moved the four `private static final String` constants to the field-declaration block at the top of the class.

`mvn test -pl exist-core -Dtest='*DeepEqual*'` — 78/78 green (`FunDeepEqual4050ReproTest` 1, `DeepEqualTest` 63, `FunDeepEqualPerformanceTest` 14). `codacy-cli analyze --tool pmd` on the touched files is clean for the three flagged sites. The two remaining warnings (`CompareObjectsWithEquals` / `UseEqualsToCompareStrings` on lines 326 + 341) are the intentional `a == b` reference-equality short-circuits in `compareNullable` / `safeCompare` — pre-existing, commented as such, mirrors `FunDeepEqual.safeCompare`; leaving them.

reinhapa (approved):

duncdrum:
@joewiz can you rebase please. There is no conflict, but I want to be safe, having just spent hours on  NodeProxy

joewiz:

Rebased onto current develop tip — clean, no conflicts across 309 commits. New tip: `c9f12fb1b6`.

Local smoke check post-rebase: FunDeepEqual JUnit + streaming comparator + performance test all green (14/14), GH-4050 fast-path numbers preserved (96ms / 55ms / 154ms on the equal/mismatched 10k-element trees).

Force-pushed; CI should run on the new tip.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunDeepEqualStreamingComparator.java
```diff
@@ -0,0 +1,357 @@
...
+        }
+        final AttrSnapshot[] out = new AttrSnapshot[kept];
+        System.arraycopy(tmp, 0, out, 0, kept);
+        Arrays.sort(out, ATTR_ORDER);
+        return out;
+    }
+
+    private static int compareNullable(@Nullable final String a, @Nullable final String b) {
+        // NOTE: intentional reference equality short-circuit (mirrors safeCompare).
+        if (a == b) {
```
duncdrum: I believe using caps would prevent the codacy warning `A == B` and keep the intention intact

line-o: Capitalizing only helps if codacy's check is dumb enough to take those for constants

line-o: But there should be a way to signal the intent to the linter in the comment

joewiz:

Done in \`badc60ba91\`. Switched \`compareNullable\`'s three return values to \`Constants.EQUAL\` / \`INFERIOR\` / \`SUPERIOR\` so it matches the sibling \`safeCompare\` method byte-for-byte on the compare-result idiom. The intentional reference-equality short-circuit (\`a == b\`) is preserved, with the existing comment explaining it.

Full-module gate green: \`Tests run: 6747, Failures: 0, Errors: 0, Skipped: 97\`.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunDeepEqualStreamingComparator.java
reinhapa: There is a `null` capable `Comparator.nullsFirst()` / `Comparator.nullsLast()` JDK alternative to handle null cases "correctly" also should identity comparisons be avoided going forward as this will be hidden traps for future value types.

reinhapa (changes requested):

joewiz:

Addressed in `23df77cdcd`. Both `compareNullable` and `safeCompare` now delegate to `Comparator.nullsFirst(...)` — the cached `NULLS_FIRST_NATURAL` instance for the no-collator path, and `Comparator.nullsFirst(collator::compare)` for the collator path. The `a == b` identity short-circuits are gone, so future value types subbed in here won't fall into the trap you described.

Equivalence verified for all four cases (both-null, a-null, b-null, both-non-null). 63 `DeepEqualTest` + 14 `FunDeepEqualPerformanceTest` pass unchanged.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunDeepEqualStreamingComparator.java
```diff
@@ -0,0 +1,345 @@
...
+        Arrays.sort(out, ATTR_ORDER);
+        return out;
+    }
+
+    /**
+     * Cached {@link Comparator} that places nulls first and otherwise uses
+     * String's natural order. Used by {@link #compareNullable(String, String)}
+     * and the no-collator branch of {@link #safeCompare(String, String, Collator)}.
+     */
+    private static final Comparator<String> NULLS_FIRST_NATURAL =
```
reinhapa: Address Codacy issue: _Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes._

reinhapa (changes requested):

joewiz:

Addressed in `4f6a7ccd6c` — the `NULLS_FIRST_NATURAL` field is hoisted up next to the existing `EOF` / `EMPTY_ATTRS` / `ATTR_ORDER` constants at the top of the class. No other changes. Codacy clean on the file. 77 `DeepEqualTest` + `FunDeepEqualPerformanceTest` cases pass unchanged.

This one I should have caught in my own pre-push Codacy run on the prior commit — I'd run Codacy on net-new files but not on the existing-file edit that introduced this field. Tightened the rule on my side (mandatory pre-push Codacy on every changed Java file, not just new ones); apologies for the extra round.

reinhapa (approved):

duncdrum (approved):
Lgtm. But we should rebase again

joewiz:

Rebased onto current develop — the rebase was clean (no conflicts), and all 77 deep-equal tests pass locally. CI is green. Thanks, Duncan!

PR #6338, opened 2026-05-10
[bugfix] repair JMX client

dizzzz:
repair JMX client. 

**Actions**
- [ ] Check each CLI parameter for NPEs and other issues.
- [ ] Check if we can think of implementation with less null checks.
- [ ] If data is not available, do not display 0 but N/A or alike.
- [ ] Discuss: additional improvements in this PR or additional PR. @line-o 

**arguments**
- [x] -a, --address <string>                     RMI address of the server
- [x] -c, --cache                                displays server statistics on cache 
- [x] -d, --db                                   display general info about the db 
- [x] -h, --help <argument-to-print-help-for>    <argument-to-print-help-for>: an argument to print help for
- [x] -i, --instance <string>                    The ID of the database instance to 
- [x] -j, --jobs                                 retrieve sanity check report from 
- [ ] -l, --locks                                lock manager: display locking 
- [x] -m, --memory                               display info on free and total 
- [x] -p, --port <integer>                       RMI port of the server
- [x] -s, --report                               retrieve sanity check report from 
- [x] -w, --wait <integer>                       while displaying server statistics: 

**ToDo:**
- [ ]  Need help for proving the locks are reported.
- [x] next PR: reformat etc for java21 like style
- [x] check jmx servlet - monex does not start loading data immediately

### DIFF
--- a/exist-core/src/main/java/org/exist/management/Agent.java
+++ b/exist-core/src/main/java/org/exist/management/Agent.java
@@ -31,13 +31,41 @@ import org.exist.util.DatabaseConfigurationException;
  */
 public interface Agent {
 
+    /**
+     * Initialise JMX MBeans for a new database instance.
+     *
+     * @param instance the broker pool representing the database instance
+     */
     void initDBInstance(BrokerPool instance);
 
+    /**
+     * Deregister all JMX MBeans associated with a database instance that is being closed.
+     *
+     * @param instance the broker pool representing the database instance
+     */
     void closeDBInstance(BrokerPool instance);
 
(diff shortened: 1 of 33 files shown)


duncdrum (approved):
Looks good. We could consider testing the shell script via bats, on CI that could happen against the `:debug` image. 

Not a blocker, if you like the idea we can create a new issue for it.

dizzzz:
I think a separate ticket/PR would be best, as I need to study how I could implement this.

reinhapa (approved):

PR #6343, opened 2026-05-11
[bugfix] map:entry: missing-key lookup NPE (closes #6340)

### LINKED ISSUE #6340: [bug] map:entry: NPE on missing-key lookup + NaN keys silently accepted (4 XQTS map-entry failures)
## Summary

The map-module XPST0003 triage on 2026-05-10 surfaced 4 real engine bugs in `map:entry` that were buried under the broader cluster of pre-CR-draft test failures. The triage's broader narrative (118 failures = pre-CR-draft tests eXist correctly rejects per the final 3.1 spec) is separate; these 4 are real bugs against the final 3.1 spec.

Two failure shapes:

- **NPE on missing-key map lookup** (2 tests) — looking up a key that doesn't exist in a map returns Java `null` instead of XPath empty sequence, which then NPEs in downstream `fn:empty` / `count` etc.
- **NaN keys accepted into the map** (2 tests) — XPath F&O 3.1 §17.1.7 specifies that `map:put` with a NaN key returns the map unchanged; by analogy `map:entry(NaN, v)` should return `map{}`. eXist constructs `map{NaN: v}` and reports `map:size eq 1` instead of `0`.

Discovered via XQTS test set `map-entry` (4 of 7 tests fail). Triage report: `joe-vault/Claude/exist/2026-05-10 map-module-xpst0003-triage-report.md`.

## Bug 1 — NPE on missing-key lookup (map-entry-003)

**Reproducer**:

```xquery
xquery version "3.1";
let $result := map:entry("foo", ("x", "y", "z"))
return empty($result("bar"))
```

**Expected**: `true` (looking up the absent key `"bar"` should return `()`, which is empty).

**Actual**: 

```
java.lang.NullPointerException: Cannot invoke "org.exist.xquery.value.Sequence.isEmpty()" because "seq" is null
    at org.exist.xquery.DynamicCardinalityCheck.eval(DynamicCardinalityCheck.java:75)
    at org.exist.xquery.functions.fn.FunEmpty.eval(FunEmpty.java:77)
    ...
```

Map lookup for a missing key is returning Java `null`, which then propagates into `DynamicCardinalityCheck.eval` and NPEs. The fix is to ensure map lookup returns `Sequence.EMPTY_SEQUENCE` (or equivalent) for missing keys.

## Bug 2 — NPE on missing-key lookup with untyped-atomic key (map-entry-004)

**Reproducer**:

```xquery
xquery version "3.1";
let $result := map:entry(xs:untypedAtomic("foo"), map{})
return empty($result("bar"))
```

**Expected**: `true`.

**Actual**: same NPE as Bug 1. Same root cause; included separately because the key type differs (`xs:untypedAtomic` rather than `xs:string`) and the value is a map rather than a sequence — useful for the regression test to cover both shapes.

## Bug 3 — NaN key accepted (map-entry-005, xs:double NaN)

**Reproducer**:

```xquery
xquery version "3.1";
let $result := map:entry(number('NaN'), 'NaN')
return map:size($result) eq 0
```

**Expected**: `true`. Per XPath F&O 3.1 §17.1.7 (`map:put`): "If the supplied key is `xs:double('NaN')` or `xs:float('NaN')`, the function returns a map that is identical to `$map`." `map:entry(NaN, v)` is semantically `map:put(map{}, NaN, v)`, which should therefore return `map{}` of size 0.

**Actual**: returns `map{NaN: "NaN"}` with `map:size eq 1`. The NaN key is being entered into the map. Fix: reject NaN keys in `map:entry` (and in `map:put`, and in any other constructor path) — silently no-op the entry.

## Bug 4 — NaN key accepted (map-entry-006, xs:float NaN)

**Reproducer**:

```xquery
xquery version "3.1";
let $result := map:entry(xs:float('NaN'), 'NaN')
return map:size($result) eq 0
```

**Expected**: `true` (same rule as bug 3).

**Actual**: returns `map{xs:float("NaN"): "NaN"}` with `map:size eq 1`. Same root cause as bug 3 but for the `xs:float` NaN; verify the fix covers both numeric types (and ideally any future NaN-bearing numeric: `xs:numeric`).

## Spec references

- XPath F&O 3.1 §17.1.4 (`map:entry`): https://www.w3.org/TR/xpath-functions-31/#func-map-entry
- XPath F&O 3.1 §17.1.7 (`map:put`): https://www.w3.org/TR/xpath-functions-31/#func-map-put — contains the explicit NaN-returns-original-map rule.
- XDM 3.1 §3.10 (Maps): https://www.w3.org/TR/xpath-datamodel-31/#dt-map — base semantics for missing-key lookup returning empty.

## Code surface (likely)

`exist-core/src/main/java/org/exist/xquery/functions/map/AbstractMapType.java` and subclasses — the `get()` method's return path for missing keys. And `MapEntry.java` / wherever `map:entry` is implemented — the key validation step.

## Affected XQTS test cases

| Test set | Test case | Bug | Failure mode |
|---|---|---|---|
| `map-entry` | `map-entry-003` | 1 | NPE in `DynamicCardinalityCheck.eval` (missing-key lookup returns null) |
| `map-entry` | `map-entry-004` | 2 | NPE, untyped-atomic key + map value |
| `map-entry` | `map-entry-005` | 3 | NaN key accepted (`xs:double`) |
| `map-entry` | `map-entry-006` | 4 | NaN key accepted (`xs:float`) |

## Suggested fix scope

Two surgical edits:

1. **`AbstractMapType.get` (or equivalent)** — when the key is absent, return `Sequence.EMPTY_SEQUENCE`, never `null`.
2. **`map:entry` / `map:put` / map-constructor key-validation** — silently reject NaN keys (`xs:double('NaN')` and `xs:float('NaN')`). The "silently reject" is per spec: don't error, just return the map unchanged (or for `map:entry`, return an empty map).

JUnit regression: 1 test class with 4 cases mirroring the 4 XQTS shapes above (plus `xs:numeric` for forward-compat).

XQTS verification: `--xqts-version HEAD --filter map-entry` should land 7/7 after the fix.

## Provenance

Discovered during the 2026-05-10 map-module XPST0003 triage (`feedback_xqts_3_1_vs_head_label_inversion.md` for context on why the 118-test cluster around this was spec-evolution noise, and these 4 were the real bugs hiding underneath). Filing separately because the fix scope is narrow and the bugs are well-isolated.

joewiz:
## Summary

Closes #6340. Surgical one-line fix: `SingleKeyMapType.get` returned Java `null` for keys that did not match the single entry's key. Callers using the function-call lookup syntax (`\$result(\"missing\")`) propagated that null through `DynamicCardinalityCheck` into `fn:empty` / `count`, raising NPE. The `?` lookup operator was already null-guarded in `Lookup.eval`, which masked the bug for the more idiomatic `?` syntax.

`MapType.get` already returns `Sequence.EMPTY_SEQUENCE` for absent keys; this aligns `SingleKeyMapType` with the same contract.

## Triage correction

Issue #6340 originally listed 4 bugs (2 NPE on missing-key, 2 NaN-key acceptance). Investigation showed:

- **The 2 NPE bugs are real** (XQTS `map-entry-003` and `map-entry-004` against final 3.1 Rec + corrections) — fixed here.
- **The 2 \"NaN-key\" failures are not bugs.** Per the final XPath F&O 3.1 Rec, NaN keys can be stored in maps and `op:same-key` treats NaN as equal to NaN. Saxon's 2014 XQTS revision (\"NaN can now be stored in the map\") updated the tests accordingly; eXist's current behaviour already matches that revision. The triage was looking at the pre-CR 2013 draft archive (the runner's `--xqts-version 3.1` archive), where `map-entry-005/006` still expect rejection. Per `feedback_xqts_3_1_vs_head_label_inversion.md`, HEAD is the canonical baseline for 7.0 conformance claims.

## What changed

- `exist-core/src/main/java/org/exist/xquery/functions/map/SingleKeyMapType.java` — `get()` returns `Sequence.EMPTY_SEQUENCE` instead of `null` on key miss.
- `exist-core/src/test/java/org/exist/xquery/functions/map/MapEntryRegressionTest.java` — new, 6 cases: the 2 NPE shapes from XQTS `map-entry-003/004`, plus 4 sanity cases covering NaN-key store + lookup (xs:double, xs:float, cross-type) and \`map:get\` on `SingleKeyMapType`.

## XQTS deltas (`--xqts-version HEAD`, the final 3.1 Rec + corrections)

| Test set | Before | After |
|---|---|---|
| `map-entry` | 7 pass / 2 NPE | **9 / 9** |

Against the older `--xqts-version 3.1` (2013 pre-CR draft) archive: 2 pass + 3 fail + 2 NPE → 4 pass + 3 fail, where the remaining 3 are pre-CR-draft expectations the final spec abandoned (NaN rejection in `map-entry-005/006` and `map:collation()` in `map-entry-007`) — out of scope.

## Spec references

- XPath F&O 3.1 §17.1.4 [`map:entry`](https://www.w3.org/TR/xpath-functions-31/#func-map-entry)
- XDM 3.1 §3.10 [Maps](https://www.w3.org/TR/xpath-datamodel-31/#dt-map) — missing-key lookup returns the empty sequence

## Test plan

- [x] `MapEntryRegressionTest` 6/6 pass on patched build
- [x] `mvn test -pl exist-core -Dtest='*Map*,*map*'` — 143/143 map-related JUnit tests pass (0 failures, 0 errors)
- [x] XQTS `map-entry` against `--xqts-version HEAD`: 9/9 pass
- [x] XQTS broader `^map-` (11 sets, 220 tests) on HEAD: 200 pass / 18 fail / 2 errors — no regression vs develop baseline; remaining failures are pre-existing (`map-call`, `map-find`, `map-merge`, `map-put`, `map-remove`, `map-size` and the 2 known `map-contains-905` / `map-get-905` class-cast errors)
- [x] Codacy PMD on changed files: 0 violations
- [ ] Full `mvn test -pl exist-core` (lock-coordinated) — deferred to CI; the local test lock has been held continuously by parallel sessions

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/SingleKeyMapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/SingleKeyMapType.java
@@ -70,7 +70,7 @@ public class SingleKeyMapType extends AbstractMapType {
         if (sameKey(collator, this.key, key)) {
             return this.value;
         }
-        return null;
+        return Sequence.EMPTY_SEQUENCE;
     }
 
     @Override
--- a/exist-core/src/test/java/org/exist/xquery/functions/map/MapEntryRegressionTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/functions/map/MapEntryRegressionTest.java
@@ -1,0 +1,109 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.functions.map;
+
+import org.exist.test.ExistXmldbEmbeddedServer;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.xmldb.api.base.ResourceSet;
+import org.xmldb.api.modules.XQueryService;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Regression tests for the map:entry bugs documented in
+ * <a href="https://github.com/eXist-db/exist/issues/6340">issue #6340</a>:
+ * NPE on missing-key lookup against a single-entry map produced by map:entry.
+ *
+ * <p>The NaN-key cases (XQTS map-entry-005 / map-entry-006) are also covered,
+ * mirroring the XQTS expectations: per XPath F&amp;O 3.1 maps can hold NaN keys,
+ * and looking up NaN finds the entry (op:same-key treats NaN as equal to NaN).</p>
+ */
+public class MapEntryRegressionTest {
+
+    @ClassRule
+    public static final ExistXmldbEmbeddedServer server =
+            new ExistXmldbEmbeddedServer(true, true, true);
+
+    /** XQTS map-entry-003: looking up an absent key returned Java null and NPE'd in fn:empty. */
+    @Test
+    public void missingKeyLookupSequenceValueReturnsEmpty() throws Exception {
+        assertEquals("true", run("""
+                let $result := map:entry("foo", ("x", "y", "z"))
+                return empty($result("bar"))
+                """));
+    }
+
+    /** XQTS map-entry-004: same NPE shape, untyped-atomic key + map value. */
+    @Test
+    public void missingKeyLookupUntypedAtomicKeyMapValueReturnsEmpty() throws Exception {
+        assertEquals("true", run("""
+                let $result := map:entry(xs:untypedAtomic("foo"), map{})
+                return empty($result("bar"))
+                """));
+    }
+
+    /**
+     * XQTS map-entry-005: NaN can be stored in a map, and looking up by NaN
+     * (op:same-key treats NaN as equal to NaN) retrieves the entry.
+     */
+    @Test
+    public void nanKeyXsDoubleStoredAndLookedUp() throws Exception {
+        assertEquals("true", run("""
+                let $result := map:entry(number('NaN'), 'NaN')
+                return map:size($result) eq 1 and exists($result(number('NaN')))
+                """));
+    }
+
+    /** XQTS map-entry-006: same as 005 but with xs:float NaN. */
+    @Test
+    public void nanKeyXsFloatStoredAndLookedUp() throws Exception {
+        assertEquals("true", run("""
+                let $result := map:entry(xs:float('NaN'), 'NaN')
+                return map:size($result) eq 1 and exists($result(number('NaN')))
+                """));
+    }
+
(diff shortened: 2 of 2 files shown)


duncdrum (approved):

line-o (approved):
I would prefer to have test tests in XQSuite tests, but this does not block the merge.

PR #6218, opened 2026-04-06
Implement XQuery 4.0 core functions (fn:, array:, map:, math:)

joewiz:
## Summary

82 new or updated functions across fn:, array:, map:, and math: namespaces for XQuery 4.0. Includes fn:replace empty-match version gating (XQ4 behavior only in 4.0 mode).

Additionally includes XQTS QT4 compliance fixes from 6 focused sessions targeting specific failure categories, plus a comprehensive fn: signature audit against W3C XQuery 4.0 PR197.

### Core XQ4 functions (original)
- 58 fn: functions (compare, characters, contains-subsequence, identity, trunk, foot, slice, etc.)
- 13 array: functions (build, index-of, index-where, of-members, slice, sort-by, sort-with, split, etc.)
- 7 map: functions (build, filter, of-pairs, etc.)
- 4 math: functions (cosh, sinh, tanh, cbrt)
- fn:replace/fn:tokenize empty-match: gated on xquery version "4.0"
- fn:doc multi-arity fix

### XQTS compliance fixes
- **Regex compliance**: Validate XPath regex patterns before Java regex fallback; support XQ4 lookaround translation
- **Keyword arguments + collation**: Rename function parameter names to match XQ4 F&O spec; implement `fn:collation(map)` overload
- **fn:element-to-map**: Fix XQ4 conformance for `fn:element-to-map` and `fn:element-to-map-plan`
- **fn:unparsed-text**: Improve conformance and add function type checking
- **Function call/ref**: Fix reserved function name handling, context item availability, and type promotions
- **Variable declaration error codes**: Fix XPST0008/XQST0049 alignment with XQuery 3.1 spec

### fn: Signature Audit (W3C XQuery 4.0 PR197)

Audit and fix `FunctionSignature` definitions across `org.exist.xquery.functions.fn` and `.array` against the W3C XQuery 4.0 spec. The QT4 keyword test set (`misc-BuiltInKeywords`) checks that `fn:foo(arg := ?) instance of function(...) as ...` matches the function-item signature W3C declares.

**Signature changes:**
- **collation parameters** widened from `xs:string` (exactly-one) to `xs:string?`: `fn:contains`, `fn:ends-with`, `fn:starts-with`, `fn:compare`, `fn:distinct-values`, `fn:index-of`, `fn:substring-before`, `fn:substring-after`, `fn:collation-key`, `fn:string-join`'s `separator`, `array:index-of`.
[...]

PR #6344, opened 2026-05-11
[bugfix] XQuery 3.1 mandatory fixes from v2/xq4-core-functions (audit extract #3 subset)

joewiz:
## Summary

Extracts the XQuery 3.1-mandatory commits from PR #6218 (`v2/xq4-core-functions`) for the eXist 7.0 conformance push, per the 2026-05-10 v2/* extraction audit. The remaining 4.0-only commits stay in #6218 for a post-7.0 cycle.

This is a **subset extraction** — 10 of 67 commits from the source branch were classified as 3.1-mandatory; the rest are 4.0-only or already-on-develop. See `2026-05-10 xq31-extraction-audit-report.md` for the audit's framing.

## Commits

In chronological source order, plus one develop-adaptation commit:

1. `00e55a8103` [bugfix] Improve XPath regex compliance: validate patterns and support XQ4 lookaround
2. `0d96ad8a19` [feature] Improve fn:unparsed-text conformance and add function type checking
3. `0aadff27a7` [bugfix] Fix function call/ref XQTS failures: reserved names, context, promotions
4. `4ab512073d` [bugfix] Fix variable declaration error codes for XQuery 3.1 compliance
5. `8dd6e9e765` [feature] Version-aware XPath regex validation with XQ4 extensions
6. `7a205f980c` [bugfix] Fix fn:contains-token collation parameter to accept empty sequence
7. `7966280a21` [bugfix] fn:load-xquery-module: check loaded module's own version
8. `e7ccdf97af` [bugfix] Tighten XPath regex validation to reject more invalid constructs
9. `0047944fb1` [bugfix] XPath regex: validate back-references and tighten char class grammar
10. `fe6a348703` [test] LoadXQueryModuleContentTest: align with backward-compat semantics
11. `a729f1fa7c` [bugfix] Adapt v2/xq4-core-functions extraction for develop's XQ 3.1-only parser

## Per-cluster scope

- **`fn-unparsed-text` family** (XPath F&O 3.1 §17.5) — encoding-agnostic dynamic resource lookup, FOUT1170 mapping, XML char validation, function-type checking
- **`prod-VarDecl` parser tightening** — XQST0054 → XQDY0054 (3.1 dynamic), XPTY0004 in VariableImpl
- **Regex validation chain** — XPath F&O 3.1 §5.6.1: reject Java/Perl extensions, validate back-references, tighten char-class grammar
- **`fn:contains-token`** — collation parameter accepts empty sequence per spec
- **`fn:load-xquery-module`** — check loaded module's own version (F&O 3.1 §C.1)
- **Function call/ref fixes** — reserved-name XPST0003 checks, context preservation for wrapped internal functions, base64Binary↔hexBinary promotion

## Excluded (4.0-only — stay in v2/xq4-core-functions)

54 commits covering: 50+ new XQ 4.0 fn: functions, array/map/math 4.0 extensions, record types, numeric literal extensions (`0x...`, `0b...`, `_`), keyword arguments, lambdas (`fn{...}`), parse-json XQ 4.0 compliance, misc-Subtyping parser, record coercion, element-to-map, from-dateTime widening, XQ 4.0 deep-equal options, hot map operations, collation(map) UCA, plus refactor/codacy commits piggybacking on the above.

## Excluded (already on develop OR no-op for develop)

- `935ea37cb3` xs:duration ordering version gate — fixes a 4.0-only gate that doesn't exist on develop
- `927895cb47` Restore fn:deep-equal attribute comparison — develop already has the call (the 'regression' only existed inside v2's DeepEqualOptions refactor)
- `5ce1a1a365` XQ4 try/catch err:map + cast errors — `\$err:map`/`\$err:stack-trace` are 4.0-only; develop already handles untypedAtomic→QName cast
- `b36833ffaf` fn:reverse lazy O(1) — structural conflict (v2 refactored RangeSequence to primitive longs; develop has accumulated its own RangeSequence work). Performance optimization with no XQTS yield, dropped from this extraction.

Plus already-shipped on develop: PR #6328 (fn:min/max), PR #6337 (fn:deep-equal SAX comparator), PR #6207 (prod-CastExpr broad fixes), PR #6331 (Type.subTypeOf), PR #6333 (DurationValue hashCode), PR #6336 (numeric/Boolean hashCode sweep).

## XQTS deltas (against W3C XQTS HEAD baseline at `4f09d0accc`)

Spot-check on the 16 affected test sets — tests F+E (failures + errors):

| Test set | baseline F+E | extract F+E | Newly passing |
|---|---:|---:|---:|
| `fn-unparsed-text` | 21 | 8 | **+13** |
| `fn-unparsed-text-available` | 9 | 8 | +1 |
| `fn-unparsed-text-lines` | 23 | 9 | **+14** |
| `prod-VarDecl` | 35 | 25 | **+10** |
| `prod-VarDecl.external` | 27 | 11 | **+16** |
| `fn-matches` | 19 | 2 | **+17** |
| `fn-matches.re` | 7 | 0 | **+7** |
| `fn-replace` | 10 | 0 | **+10** |
| `fn-tokenize` | 21 | 6 | **+15** |
| `prod-CastExpr` | 37 | 17 | **+20** |
| `prod-CastExpr.derived` | 22 | 19 | +3 |
| `fn-contains-token` | 45 | 3 | **+42** |
| `fn-load-xquery-module` | 0 | 0 | 0 |
| `prod-NamedFunctionRef` | 53 | 9 | **+44** |
| `prod-FunctionCall` | 63 | 28 | **+35** |
| **Total** | **392** | **145** | **+247** |

This far exceeds the audit's 50-80 estimate. The contains-token, named-function-ref, and function-call clusters dominate the gain.

## Develop-adaptation commit

`a729f1fa7c` makes two adjustments that are necessary because develop only supports XQuery 1.0/3.0/3.1 (the new XQ 4.0 parser is on `v2/new-parser`):

1. `FunUnparsedText.readLines`: catch `RuntimeException` from the dynamic text-resource lambda alongside `IOException`. The new dynamic-resource lookup path triggers any registered `ResourceFactory`; if a factory throws an unchecked exception, wrap it as `FOUT1170` instead of letting it escape. Restores `FunUnparsedTextTest#unparsedTextLines_noDataStream` to passing.
2. Remove `LoadXQueryModuleContentTest`: all three test cases use `xquery version "4.0"` syntax which develop's parser rejects with XQST0031. The production fix in `fn:load-xquery-module` ships unchanged; the test cases will return alongside the XQ 4.0 parser landing.

## Test plan

- [x] `mvn install -pl exist-core -am -DskipTests` green
- [x] Targeted JUnit on touched classes (ContainsTokenEmptyCollationTest, RegexUtilTest, FunMatchesTest, FunReplaceTest, FunUnparsedTextTest) — 8/8 passing
- [x] XQTS HEAD spot-check on affected sets (16 sets, +247 newly passing)
- [ ] CI gate (local full `mvn test` was contaminated by disk-full + concurrent BrokerPool contention from a parallel session — environmental, not code; CI will provide authoritative signal)

## Source

Subset cherry-picked from `joewiz:v2/xq4-core-functions` (PR #6218). The remaining 4.0-only commits stay in #6218 for the post-7.0 cycle. Per the paused rebase tasking, this extraction reduces the eventual rebase conflict surface.


exist-core/src/main/java/org/exist/xquery/functions/fn/FunReplace.java
reinhapa, line 1: Addess Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.regex.RegexUtil.hasXPath4Lookaround' due to existing static import 'org.exist.xquery.regex.RegexUtil.*'_

exist-core/src/main/java/org/exist/xquery/functions/fn/FunTokenize.java
reinhapa, line 1: Address Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.regex.RegexUtil.hasXPath4Lookaround' due to existing static import 'org.exist.xquery.regex.RegexUtil.*'_

exist-core/src/main/java/org/exist/xquery/value/SequenceType.java
```diff
@@ -41,6 +42,8 @@ public class SequenceType {
     private int primaryType = Type.ITEM;
     private Cardinality cardinality = Cardinality.EXACTLY_ONE;
     private QName nodeName = null;
+    private SequenceType[] functionParamTypes = null;
+    private SequenceType functionReturnType = null;
```
reinhapa: Remove unnecessary `null` assignment as this is the default

exist-core/src/main/java/org/exist/xquery/value/SequenceType.java
```diff
@@ -143,6 +176,14 @@ public boolean checkType(final Item item) {
         if (!Type.subTypeOf(type, primaryType)) {
             return false;
         }
+
+        // For function types, check parameter and return type compatibility
+        if (Type.subTypeOf(primaryType, Type.FUNCTION) && item instanceof FunctionReference) {
+            if (!checkFunctionType((FunctionReference) item)) {
```
reinhapa: Address Codacy isse: _These nested if statements could be combined_

exist-core/src/main/java/org/exist/xquery/value/SequenceType.java
```diff
@@ -162,6 +203,42 @@ public boolean checkType(final Item item) {
...
+     *
+     * @param funcRef the function reference to check
+     * @return true if the function matches the required function type
+     */
+    private boolean checkFunctionType(final FunctionReference funcRef) {
+        final FunctionSignature sig = funcRef.getSignature();
+
+        // Check arity: if we have typed parameter info, check against it
+        if (functionParamTypes != null) {
+            if (sig.getArgumentCount() != functionParamTypes.length) {
```
reinhapa: Address Codacy isse: _These nested if statements could be combined_

exist-core/src/test/java/org/exist/xquery/functions/fn/ContainsTokenEmptyCollationTest.java
reinhapa, line 1: Address Codacy issues: _The JUnit 4 test method name 'xxx' doesn't match '[a-z][a-zA-Z0-9]*'_

reinhapa (changes requested):

joewiz:

Addressed all 6 Codacy findings. New tip: `390016f998`.

- `FunReplace` / `FunTokenize`: dropped FQN on `hasXPath4Lookaround` / `translateXPath4Lookaround` (also unqualified `validateXPathRegex` in `FunTokenize` for consistency — same static-import scope).
- `SequenceType`: removed `= null` initializers on the two new function-test fields; combined the two flagged pairs of nested `if`s into single conjunctions.
- `ContainsTokenEmptyCollationTest`: renamed `emptyCollation_emptySequenceLiteral` → `emptyCollationEmptySequenceLiteral` and the other two methods similarly.

`codacy-cli analyze --tool pmd` on the touched files is clean for the six review findings. `mvn test -pl exist-core -Dtest=ContainsTokenEmptyCollationTest` — 3/3 green. The two remaining PMD warnings (`FunReplace.eval` NPath 313, `SequenceType.checkType` NPath 384) are moderate-tier and not in this review's scope.

exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
```diff
@@ -126,8 +126,19 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
     }
 
     private void analyzeString(final MemTreeBuilder builder, final String input, String pattern, final String flags) throws XPathException {
+        final boolean isXQuery40 = context.getXQueryVersion() >= 40;
         final Configuration config = context.getBroker().getBrokerPool().getSaxonConfiguration();
 
+        // XQ4: translate (*positive_lookahead:...) etc. to Java regex
+        if (isXQuery40 && org.exist.xquery.regex.RegexUtil.hasXPath4Lookaround(pattern)) {
```
line-o: Is this XQuery 4.0 handling left in here intentionally?

exist-core/src/main/java/org/exist/xquery/functions/fn/FunMatches.java
```diff
@@ -525,7 +525,36 @@ private Sequence evalGeneric(final Sequence contextSequence, final Item contextI
     }
 
 
-    private boolean matchXmlRegex(final String string, final String pattern, final String flags) throws XPathException {
+    private boolean matchXmlRegex(String string, String pattern, final String flags) throws XPathException {
+        final boolean isXQuery40 = context.getXQueryVersion() >= 40;
```
line-o: see above

joewiz:

Good catch — let me explain what's there and let you steer.

Both sites gate on `isXQuery40 = context.getXQueryVersion() >= 40`. On develop today:

- `XQueryContext.xqueryVersion` defaults to `31`
- `setXQueryVersion()` is only called from the parser, which rejects `xquery version "4.0"` with XPST0031 (no XQ 4.0 grammar on develop yet)
- So `isXQuery40` is never `true` → both XQ 4.0 lookaround-translation branches are dead code on develop

The `validateXPathRegex(this, pattern, isXQuery40)` call still ships and does the real 3.1 work: rejecting Java/Perl regex extensions that Saxon's XP30 mode would otherwise accept (this is the +17 / +15 / +10 / +7 lift on `fn-matches`, `fn-matches.re`, `fn-replace`, `fn-tokenize` in the PR's XQTS table).

So it's intentional in the sense that I preserved the version-gating shape rather than ripping the branches and re-adding them in #6218 — but you're right that it's mis-framed for "3.1-mandatory subset" if you read the diff strictly.

Two options, your pick:

1. **Leave as-is**: dead code on develop, zero behavior change, cleaner eventual #6218 merge (no re-adding).
2. **Strip the four `isXQuery40` lines + the `needsXQuery40JavaRegex` branch**: PR diff becomes purely 3.1, and #6218 carries the XQ 4.0 reactivation later.

I'm fine with either. (2) is ~5 minutes of edits. Want me to do (2)?

duncdrum:
I m leaning towards 'it's not a bug it's a feature' 

Xq4 is coming in general and exist in particular. Having targeted error messages when users demand it is nice in a away.

line-o:
@joewiz `xmlts.xquery3 recursion function calls.recursion-function-calls-002` fails

line-o:
I would like to us to _keep_ the `isXQuery40` check, _if_ we can raise an error with a "not implemented yet" description.

joewiz:

Both items addressed in `327393c41c` + `154ce4d648`.

**1. `isXQuery40` → "not implemented yet" error** (`327393c41c`)

Per your preferred middle ground: both call sites in `FunMatches.matchXmlRegex` and `FunAnalyzeString.analyzeString` now raise `XPST0017` with

> "XPath 4.0 lookaround syntax in regex patterns (e.g. `(*positive_lookahead:...)`) is not yet implemented in this XQuery 3.1 build. Rewrite the regex without lookaround."

instead of silently dead-coding the XQ4 branch. When XQuery 4.0 lands on develop, replacing the `throw` with the existing `translateXPath4Lookaround()` dispatch is a one-spot change per file. `validateXPathRegex` is now called with `isXQuery40 = false` explicitly. Added `RegexXPath4NotImplementedTest` (3 tests: both functions raise XPST0017 for lookaround patterns; plain patterns still work).

**2. `recursion-function-calls-002`** (`154ce4d648`)

Real regression — the failure was introduced by `0d96ad8a19` (function type checking in `SequenceType`). That commit added an arity check under `Type.subTypeOf(primaryType, Type.FUNCTION)`, but `MAP_ITEM` and `ARRAY_ITEM` are declared subtypes of `FUNCTION`, so the arity check also applied to typed `map(K, V)` tests. The map accessor signature has 1 argument (the key) while `map(xs:string, item()?)` carries 2 type parameters — they never match, so `instance of map(xs:string, item()?)` returned `false` on every map. In this test the recursive `local:join` branch was therefore bypassed, a raw map flowed into `array:append` / `array:flatten`, and `string-join` finally tripped `FOTY0013` ("A function item other than an array cannot be atomized").

Fix: narrow the function-type check to `primaryType == Type.FUNCTION` so the plain `function()` typed test still validates (maps/arrays continue to satisfy it via FUNCTION subtyping), while `map(K, V)` and `array(T)` tests revert to their pre-`0d96ad8a19` behaviour pending proper typed-test support.

While running the gate I also noticed the `xqts.org.exist-db.test.securitymanager.id.from-load-module` test was failing for an independent reason: the test fixture stores a module declared `xquery version "3.0"` and loads it without an explicit `xquery-version` option, which trips the stricter loaded-module version check added in `7966280a21`. Bumped the fixture to `"3.1"` in the same commit so the gate stays green.

Full-module gate after both fixes:

```
Tests run: 6706, Failures: 0, Errors: 0, Skipped: 97
BUILD SUCCESS  (Total time: 03:38 min)
```

Codacy/PMD on the changed files is clean (the pre-existing `FunMatches.eval` / `evalWithIndex` NPathComplexity findings are untouched).

exist-core/src/main/java/org/exist/xquery/functions/fn/FunMatches.java
```diff
@@ -525,34 +525,20 @@ private Sequence evalGeneric(final Sequence contextSequence, final Item contextI
...
-                final String javaPattern = org.exist.xquery.regex.RegexUtil.translateRegexp(
-                        this, pattern, flags.contains("x"), flags.contains("i"));
-                final int javaFlags = org.exist.xquery.regex.RegexUtil.parseFlags(this, flags);
-                return Pattern.compile(javaPattern, javaFlags).matcher(string).find();
-            } catch (final PatternSyntaxException e) {
-                throw new XPathException(this, ErrorCodes.FORX0002,
-                        "Invalid regular expression: " + e.getMessage(),
-                        new StringValue(this, pattern), e);
-            }
+            org.exist.xquery.regex.RegexUtil.validateXPathRegex(this, pattern, false);
```
reinhapa: Address Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.regex.RegexUtil.validateXPathRegex' due to existing static import 'org.exist.xquery.regex.RegexUtil.*'_

exist-core/src/main/java/org/exist/xquery/functions/fn/FunMatches.java
```diff
@@ -525,34 +525,20 @@ private Sequence evalGeneric(final Sequence contextSequence, final Item contextI
...
-
-
-        // XQ4: translate (*positive_lookahead:...) etc. to Java regex (?=...) syntax
-        if (isXQuery40 && org.exist.xquery.regex.RegexUtil.hasXPath4Lookaround(pattern)) {
-            pattern = org.exist.xquery.regex.RegexUtil.translateXPath4Lookaround(pattern);
+    private boolean matchXmlRegex(String string, final String pattern, final String flags) throws XPathException {
+        // XPath 4.0 lookaround syntax is not yet implemented in eXist's XQuery 3.1 runtime.
+        // When XQuery 4.0 lands (v2/xq4-core-functions), replace this guard with the
+        // translateXPath4Lookaround / Java-regex dispatch path.
+        if (org.exist.xquery.regex.RegexUtil.hasXPath4Lookaround(pattern)) {
```
reinhapa: Address Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.regex.RegexUtil.validateXPathRegex' due to existing static import 'org.exist.xquery.regex.RegexUtil.*'_

exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
```diff
@@ -125,18 +125,21 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
...
+    private void analyzeString(final MemTreeBuilder builder, final String input, final String pattern, final String flags) throws XPathException {
         final Configuration config = context.getBroker().getBrokerPool().getSaxonConfiguration();
 
-        // XQ4: translate (*positive_lookahead:...) etc. to Java regex
-        if (isXQuery40 && org.exist.xquery.regex.RegexUtil.hasXPath4Lookaround(pattern)) {
-            pattern = org.exist.xquery.regex.RegexUtil.translateXPath4Lookaround(pattern);
+        // XPath 4.0 lookaround syntax is not yet implemented in eXist's XQuery 3.1 runtime.
+        // When XQuery 4.0 lands (v2/xq4-core-functions), replace this guard with the
+        // translateXPath4Lookaround() dispatch path.
+        if (org.exist.xquery.regex.RegexUtil.hasXPath4Lookaround(pattern)) {
```
reinhapa: Address Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.regex.RegexUtil.validateXPathRegex' due to existing static import 'org.exist.xquery.regex.RegexUtil.*'_

exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
```diff
@@ -125,18 +125,21 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
...
+            throw new XPathException(this, ErrorCodes.XPST0017,
+                    "XPath 4.0 lookaround syntax in regex patterns (e.g. (*positive_lookahead:...)) "
+                            + "is not yet implemented in this XQuery 3.1 build. Rewrite the regex without lookaround.");
         }
 
-        // Pre-validate: reject constructs not valid in XPath regex
+        // Pre-validate: reject constructs not valid in XPath 3.1 regex
```
reinhapa, line 146: Address Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.regex.RegexUtil.validateXPathRegex' due to existing static import 'org.exist.xquery.regex.RegexUtil.*'_

exist-core/src/main/java/org/exist/xquery/functions/fn/FunAnalyzeString.java
```diff
@@ -125,18 +125,21 @@ public Sequence eval(final Sequence[] args, final Sequence contextSequence) thro
...
         if (!org.exist.xquery.regex.RegexUtil.hasLiteral(flags)) {
-            org.exist.xquery.regex.RegexUtil.validateXPathRegex(this, pattern, isXQuery40);
+            org.exist.xquery.regex.RegexUtil.validateXPathRegex(this, pattern, false);
```
reinhapa: Address Codacy issue: _Unnecessary use of fully qualified name 'org.exist.xquery.regex.RegexUtil.validateXPathRegex' due to existing static import 'org.exist.xquery.regex.RegexUtil.*'_

reinhapa (changes requested):

line-o:
@joewiz Is it possible to address the remaining codacy issues so that this PR can land? As it is addressing so many issue in the XQuery runtime it is of high priority.

line-o:
@joewiz we have conflicts now. Can you resolve them?

line-o (approved):
In light of the timeline and given this will let us pass almost 250 additional XQTS tests I would be in favor of pulling this in _despite_ the fact that Codacy flagged 14 issues.

It will be hard enough for us to incorporate this PR as is and addressing the _pre-existing_ NPath complexity issues will only raise the likelyhood of conflicting changes with other PRs waiting in the pipeline.

joewiz:

Thank you for the approval @line-o — really appreciate the framing about pre-existing NPath complexity carrying conflict risk with other PRs in the pipeline.

Pushed a small follow-up (469d60c5da) addressing 4 of the lower-risk Codacy findings that don't touch any of the NPath territory you flagged:

- `FunUnparsedText.java` — replaced `StandardCharsets.UTF_8` with the already-static-imported `UTF_8`
- `FunMatches.java` — collapsed a nested `if` into a single condition
- `XQueryContext.ModuleVertex.equals` — combined the two-step compare into a single `&&` return
- `RegexUtil.java` — removed unused `convertUnicodeBlockNames(String)` private method (leftover from an early Java-regex attempt)

The other 10 findings (mostly NPath complexity on `importModule`, `compileModule`, `validateXPathRegex`, `evalWithIndex`, etc., plus a handful of `AvoidReassigningParameters` and `FieldDeclarationsShouldBeAtStartOfClass`) are left untouched — they're pre-existing and refactoring them mid-flight would risk exactly the conflicts you described.

Ready when you are.

line-o:
@joewiz this PR needs to be rebased

joewiz:

Rebased onto current develop in ad70f181f7. One small conflict on `RegexUtil.java` — develop's HEAD has no extra imports between `javax.annotation.Nullable` and `java.util.regex.Pattern`; the PR's `[bugfix] Improve XPath regex compliance` commit added three. Resolved by keeping only `java.util.regex.Matcher` (the one actually used by the lookaround code at lines 303 and 318) and dropping the unused `ArrayList` / `List` imports. `mvn install -pl exist-core -am -DskipTests` clean.

line-o:
I would _love_ to be get this in (I think I repeat myself). There is a number of Codacy issues @joewiz could you address them?

joewiz:

Pushed 12340dd917 addressing the 13 Codacy findings:

**6 FQN nits**: dropped `org.exist.xquery.regex.RegexUtil.` qualifier on six call sites in FunAnalyzeString / FunMatches / FunReplace (each file already had `import static ...RegexUtil.*;`).

**1 unused field**: removed `passContextToBody` field + `setPassContextToBody` setter from UserDefinedFunction — the role was renamed to `propagateContextToBody` during the PR's iterations and the old names got left behind, no callers anywhere.

**1 field-at-top**: moved the static `XPATH4_LOOKAROUND` pattern to the top of `RegexUtil` with the rest of the constants.

**5 NPath complexity** on the new regex validator methods:
- `validateLookaroundConstraints` (was 68,401): extracted lookaround detection / body scanning / post-')' quantifier checks into 4 helpers + 2 small value-type holders
- `scanCharClass` (was 53,949): extracted `scanEscapeInCharClass`, `scanCategoryEscape`, `scanNestedCharClass`; replaced the multi-character case arm with a `Set` lookup
- `validateXPathRegex` (was 388 → 216 after the obvious decomposition): added a `classifyBackslashEscape` + `BackslashKind` enum so the dispatch is a single `switch` expression rather than an if-chain
- `countCapturingGroups` (was 361): extracted three small helpers (`skipBackslashForGroupCount`, `skipCharClassForGroupCount`, `isCapturingGroupOpen`)

Local `codacy-cli analyze --tool pmd` on the five touched files: clean. `XQuery3Tests` 1030/1030 pass. `RegexXPath4NotImplementedTest` 3/3 pass.

exist-core/src/main/java/org/exist/xquery/DynamicTypeCheck.java
```diff
@@ -149,6 +149,15 @@ private void check(Sequence result, Item item) throws XPathException {
             } else if (type == Type.ANY_URI && requiredType == Type.STRING) {
                     item = item.convertTo(Type.STRING);
                     type = Type.STRING;
+            //Binary type promotion (XQuery 4.0): xs:base64Binary ↔ xs:hexBinary
```
duncdrum: XQ 4.0 ? It's a small thing, mainly flagging for making sure we want it in during 3.1 conformance push.

exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
```diff
@@ -1082,7 +1082,17 @@ private AtomicValue convertForValueComparison(final AtomicValue value, final int
             }
 
             /*
-             *  d. Otherwise, a type error is raised [err:XPTY0004].
+             *  d. (XQuery 4.0) If each operand is an instance of one of the types
```
duncdrum: Goes with the other XQ 4 point.

duncdrum (commented):
Just a question about the type promotion.

exist-core/src/main/java/org/exist/xquery/GeneralComparison.java
line-o: Looks like a rule for comparing binary types coming with XQ4 is already added here.
We could drop it or keep it.

exist-core/src/main/java/org/exist/xquery/DynamicTypeCheck.java
line-o: Same XQ4 binary type promotion.

joewiz:

Good catch, @duncdrum and @line-o — both blocks are explicitly XQ 4.0 in the comments and don't belong in this XQ 3.1 conformance subset. Dropped them in ee64a1568f:

- `DynamicTypeCheck.java`: the xs:base64Binary ↔ xs:hexBinary type-promotion branch in the static-type-check ladder
- `GeneralComparison.java`: case (d) of `resolveNumericType` — cast both operands of a general comparison to xs:base64Binary when both are binary. Renumbered the now-final XPTY0004 case from (e) to (d).

They'll come back together with the rest of XQ 4.0 type coercion via #6260. XQuery3Tests still 1030/1030 locally.

duncdrum (approved):
Thanks @joewiz  from my side it's thunderbirds go 🚀

line-o:
Same here.

reinhapa (approved):

PR #6339, opened 2026-05-10
[feature] Implement `declare decimal-format` for XQuery 3.1

joewiz:
## Summary

Adds the XQuery 3.1 `declare decimal-format` prolog declaration. Per W3C XPath/XQuery 3.1 §4.16 (https://www.w3.org/TR/xquery-31/#prod-xquery31-DecimalFormatDecl), this prolog form lets an XQuery module declare named (or default) decimal-format properties that subsequent `fn:format-number()` calls reference.

This PR is an extraction from PR #6217 (`v2/declare-decimal-format`) per the 2026-05-10 v2/* extraction audit. The whole branch is XQ 3.1-mandatory by spec — extracting it directly to `develop` removes a 3.1 conformance gap from the 7.0 release path without waiting for the broader v2/xq4-* feature set.

## What changed

Three commits cherry-picked from `joewiz:v2/declare-decimal-format`:

1. `[feature] Implement declare decimal-format for XQuery 3.1`
2. `[refactor] Address review: add df prefix to decimal-format helpers, add Javadoc`
3. `[refactor] Use switch expression for decimal-format property dispatch`

Touched files:
- `exist-core/src/main/antlr/org/exist/xquery/parser/XQuery.g` — adds `DecimalFormatDecl` production
- `exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g` — tree-walker integration
- `exist-core/src/main/java/org/exist/xquery/DecimalFormat.java` — prolog handler with property semantics
- `exist-core/src/main/java/org/exist/xquery/ErrorCodes.java` — adds `XQST0097`, `XQST0098`
- `exist-core/src/main/java/org/exist/xquery/XQueryContext.java` — wiring
- `exist-core/src/test/xquery/numbers/format-numbers.xql` — test additions

One conflict resolved during cherry-pick: `ErrorCodes.java` had additive divergence between v2's older base (string-concat `XQDY0101`, no `XQST0118`/`XQST0125`) and current `develop` (text-block `XQDY0101`, `XQST0118`/`XQST0125` present). Resolution kept `develop`'s modern forms and appended the two new decimal-format codes (`XQST0097`, `XQST0098`) using the same text-block style.

## Spec references

- W3C XPath/XQuery 3.1 §4.16 — DecimalFormatDecl: https://www.w3.org/TR/xquery-31/#prod-xquery31-DecimalFormatDecl
[...]

PR #6349, opened 2026-05-11
[bugfix] fn:format-number: preserve full xs:decimal precision

joewiz:
## Summary

`fn:format-number` was truncating high-precision `xs:decimal` / `xs:integer` inputs to 16 significant digits before formatting, because the final BigDecimal coercion ran `multiply(BigDecimal.ONE, MathContext.DECIMAL64)`. Multiplying by ONE is otherwise a no-op — the `MathContext` was the only effect, and it was strictly destructive.

Removing the `MathContext.DECIMAL64` cap preserves full precision. The subsequent `.round(maxFractionalPartSize)` still handles the user-requested fractional-part size.

`xs:double` inputs are unaffected: `DoubleValue.toDecimalValue()` already uses `BigDecimal.valueOf(value)`, which yields the canonical short representation (e.g. `0.1` rather than the 50-digit exact bit-pattern decimal). So no regression there.

## What changed

- `FnFormatNumbers.format()` line 605: drop the `multiply(BigDecimal.ONE, MathContext.DECIMAL64)` from the final adjusted-number coercion.
- One-line diff (plus a single-sentence comment explaining the rationale).

## Spec references

- [XPath/XQuery 3.1 §4.7.3 fn:format-number](https://www.w3.org/TR/xpath-functions-31/#func-format-number)
- The relevant XQTS test cases (`numberformat63`/`numberformat64`) accept *either* full-precision output *or* `FOAR0002` per their `<any-of>` assertion. We now produce the full-precision output, matching what Saxon and BaseX return.

## XQTS 3.1 fn-format-number — before / after

Measured locally with the runner JAR built from this branch's exist-core, against `qt3tests-master`:

| Suite | Pass / Tests | % |
|-------|-------------:|--:|
| before (`65a1d3ded3`, gap-analysis run 2026-05-09) | 74 / 77 | 96.1% |
| after (this PR @ `bc327814fe`) | 78 / 78 | 100% |

Net delta from this commit alone: **+2** (numberformat63, numberformat64). The third previous failure, `numberformat901err`, was the `declare default decimal-format` parser case — that one was fixed independently by #6339 (merged ~2026-05-10) and is included in the post-numbers above. The +1 in test count is unrelated drift in the upstream qt3tests catalog.

## Test plan

- [x] `mvn install -pl exist-core -am -DskipTests` — clean build
- [x] `mvn test -pl exist-core -Dtest='NumberTests,XQuery3Tests,XPathQueryTest'` — 1,322 tests, 0 failures, 6 skipped (pre-existing)
- [x] XQTS 3.1 fn-format-number: 78 / 78
- [x] Codacy PMD on changed file: no new warnings introduced (4 pre-existing warnings unchanged)

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FnFormatNumbers.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FnFormatNumbers.java
@@ -844,7 +844,8 @@ public class FnFormatNumbers extends BasicFunction {
             }
         }
 
-        adjustedNumber = new DecimalValue(this, adjustedNumber.convertTo(Type.DECIMAL).toJavaObject(BigDecimal.class).multiply(BigDecimal.ONE, MathContext.DECIMAL64)).round(new IntegerValue(this, subPicture.getMaximumFractionalPartSize())).abs();
+        // Preserve full BigDecimal precision (XQTS numberformat63/64 require >16 sig digits for xs:decimal/xs:integer inputs); a previous MathContext.DECIMAL64 cap truncated to 16. xs:double inputs already arrive via BigDecimal.valueOf, so the canonical representation is preserved.
+        adjustedNumber = new DecimalValue(this, adjustedNumber.convertTo(Type.DECIMAL).toJavaObject(BigDecimal.class)).round(new IntegerValue(this, subPicture.getMaximumFractionalPartSize())).abs();
 
         /* we can now start formatting for display */
 


reinhapa (approved):

duncdrum (approved):

duncdrum:
So multiplying a number times 1 is actually changing the number. … computers are weird

PR #6375, opened 2026-05-14
Bugfix: address timeout newest jetty server

dizzzz:
attempt to fix #6373 and #6367


dizzzz:
the problem is internally with how jetty uses resources.

PR #6377, opened 2026-05-16
[bugfix] AbstractInternalModule: enforce sorted-array invariant when ordered=true (#6376)

### LINKED ISSUE #6376: [7.0-SNAPSHOT regression] Calling pkg:close() fails with err:XPST0017 even when signature is correctly registered
## Summary

In current eXist 7.0-SNAPSHOT (verified via Docker image `existdb/existdb:latest`, pulled 2026-05-16), calling **any** XQuery internal-module function named `close` with arguments fails to resolve, with the parser claiming no signature matches — even though `inspect:inspect-module-uri` and `util:registered-functions` both confirm the signature is correctly registered. This appears to affect all `pkg:close()` functions in the system; the bare local name `close` looks like it's being treated as a context-dependent keyword.

## Reproduction

```xquery
import module namespace sql="http://exist-db.org/xquery/sql";
sql:close("anything")
```

Result:
```
err:XPST0017 Unexpectedly received 1 parameter(s) in call to function 'sql:close()'.
Defined function signatures are:
sql:close($connection-handle as xs:long) as empty-sequence()
```

The parser shows the correctly-registered 1-arg signature **and** simultaneously says no signature matches a 1-arg call.

Same symptom reproduces with a fresh `cursor:close` function I added in [existdb-openapi#18](https://github.com/eXist-db/existdb-openapi/pull/18) using the standard `FunctionDSL` pattern. The function inspects correctly via `inspect:inspect-module-uri` but every call fails — `cursor:close()`, `cursor:close($x)`, `cursor:close#1` (function reference), and even `function-lookup(xs:QName("cursor:close"), 1)` (which returns the empty sequence).

Renaming the function from `close` to `release` (no other code changes) made it resolve and call correctly. That is the workaround used in existdb-openapi#18.

## Diagnostic data

- `util:registered-functions("http://exist-db.org/xquery/cursor")` correctly returns `cursor:close` alongside `cursor:eval` and `cursor:fetch`.
- `inspect:inspect-module-uri(xs:anyURI("http://exist-db.org/xquery/cursor"))` correctly shows the `cursor:close` signature with `var="cursor"` cardinality `exactly one` type `xs:string`.
- `function-lookup(xs:QName("cursor:close"), N)` returns empty for every `N` 0..5.
- Other functions in the same module (`cursor:eval`, `cursor:fetch`) work normally with the same registration pattern.

## Hypothesis

Looks like an XQuery 4.0 grammar update introduced a context-dependent keyword treatment of `close` (e.g., for window-clause `sliding window ... close ...` syntax) that's leaking into the function-call grammar where it shouldn't.

## Workaround

Rename `:close` functions to alternative names (`release`, `destroy`, `remove`, etc.). This is what existdb-openapi#18 did.

## Severity

Medium-high — this is a silent regression for downstream code that relies on `sql:close`, and likely affects other `pkg:close` functions across the ecosystem (jndi, mail, lucene, etc., none of which I confirmed working in this snapshot). The error message is misleading (claims arity mismatch when signature actually matches), making it hard to diagnose.

joewiz:
## Summary

- Closes #6376
- An `InternalModule` that declares `functionsOrdered=true` was previously trusted to pass a `FunctionDef[]` sorted by `FunctionId` (qname, arity). Pre-fix, `getFunctionDef()` would binary-search that array; an unsorted array silently made some functions unreachable via direct call, `fn:function-lookup`, and function references, while linear-scan APIs (`util:registered-functions`, `inspect:inspect-module-uri`) continued to see them.
- The constructor now defensive-copies and sorts when the invariant is violated, and logs a warning so the module author notices the latent bug in their own code.

## Root cause

Module declares `ordered=true` → `getFunctionDef()` uses `binarySearch(FunctionId)`. If the array isn't actually sorted by `FunctionId.compareTo` order (qname primary, arity secondary), binary search misses entries whose position violates the sort. Concrete example from the issue: a module with three functions declared in source order `eval`, `fetch`, `close` — alphabetically `close < eval < fetch`, so binary search at the midpoint `fetch` recurses left into `[eval]` only, never finding `close`. Renaming to `release` (which sorts after `fetch`) made the bug disappear by coincidence.

The mismatch between linear-scan visibility and binary-search reachability is what made this hard to diagnose — `util:registered-functions` and `inspect:inspect-module-uri` walked `mFunctions` linearly and showed the function as present.

## What changed

`exist-core/src/main/java/org/exist/xquery/AbstractInternalModule.java`:
- Constructor checks whether the array is sorted when `functionsOrdered=true`. If not, defensive-copies via `.clone()`, sorts with `FunctionComparator`, and stores the sorted copy. The caller's static array is not mutated.
- Logs a `WARN` with the offending module's class name and a pointer to #6376 so the module author knows to fix their declaration order.
- Already-sorted arrays are reused as-is (no copy, no log).
- `functionsOrdered=false` is unchanged — linear search doesn't care about order.
- Field declarations moved above the `FunctionComparator` inner class to satisfy PMD `FieldDeclarationsShouldBeAtStartOfClass` (pre-existing finding surfaced by Codacy).

`exist-core/src/test/java/org/exist/xquery/AbstractInternalModuleSortTest.java` (new):
- `unsortedArrayWithOrderedTrueStillFindsAllFunctions` — regression test for the bug.
- `callerArrayIsNotMutated` — verifies the caller's `static final FunctionDef[]` is left intact.
- `alreadySortedArrayIsReusedAsIs` — verifies the no-op path.
- `unorderedModuleDoesNotSort` — verifies `functionsOrdered=false` is unchanged.

## Test plan

- [x] `mvn test -pl exist-core -Dtest=AbstractInternalModuleSortTest` — 4/4 pass
- [x] Codacy clean (`pmd` on both files)
- [ ] CI green

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/AbstractInternalModule.java
+++ b/exist-core/src/main/java/org/exist/xquery/AbstractInternalModule.java
@@ -23,6 +23,8 @@ package org.exist.xquery;
 
 import java.util.*;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.exist.dom.QName;
 import org.exist.xquery.value.Sequence;
 
@@ -41,6 +43,13 @@ import javax.annotation.Nullable;
  */
 public abstract class AbstractInternalModule implements InternalModule {
 
+    private static final Logger LOG = LogManager.getLogger(AbstractInternalModule.class);
+
+    protected final FunctionDef[] mFunctions;
+    protected final boolean ordered;
+    protected final Map<QName, Variable> mGlobalVariables = new HashMap<>();
+    private final Map<String, List<?>> parameters;
+
     public static class FunctionComparator implements Comparator<FunctionDef> {
         @Override
         public int compare(final FunctionDef o1, final FunctionDef o2) {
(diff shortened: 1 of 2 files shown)


joewiz:

Closing in favor of #6378, which presents the bug as an issue with three fix options rather than pre-deciding the choice. The choice between defensive runtime sort, dropping the `functionsOrdered` flag, and a build-time test alone has API / performance trade-offs that are better made by the core devs who own the module-registration contract.

Branch `bugfix/6376-ordered-internal-module-sort` remains on the fork for reference if approach (1) is the chosen direction.

PR #6380, opened 2026-05-17
[bugfix] Restore getInstanceId() JavaBean naming on PerInstanceMBean (#6379)

### LINKED ISSUE #6379: [regression] JMX `InstanceId` attribute (and others) dropped after `getXxx()` → `xxx()` rename in 9af8663a (2026-05-11)

## Symptom

JMX clients no longer see `InstanceId` as an attribute on any of eXist's MXBeans. This was caught while testing monex's monitoring pane against current `develop` — the `<jmx:InstanceId>` element no longer appears in the JMX XML output, and `MBeanServerConnection.getAttribute(..., "InstanceId")` raises `AttributeNotFoundException`.

## Diagnosis

Bisected to commit `9af8663a` ("[refactor] Reformatting and improving code", 2026-05-11, @dizzzz). That commit renamed `getInstanceId()` → `instanceId()` in:

- The `PerInstanceMBean` interface (defines the contract)
- 8 concrete impl classes: `Database`, `CollectionCache`, `DiskUsage`, `LockTable`, `ProcessReport`, `SanityReport`, `Cache`, `CacheManager`

JMX uses the JavaBean naming convention to discover attributes — a method `getX()` declared on an MBean / MXBean interface is exposed as attribute `X`. A bare-named `x()` is not an attribute; it's invokable as an operation instead. Removing the `get` prefix silently drops the attribute from MBeanInfo.

## Scope of `InstanceId` loss

Nine MXBean interfaces extend `PerInstanceMBean` and inherit its (renamed) contract, so all nine lose the `InstanceId` attribute:

- `DatabaseMXBean`, `BinaryValuesMXBean`, `CollectionCacheMXBean`, `SanityReportMXBean`, `LockTableMXBean`, `DiskUsageMXBean`, `ProcessReportMXBean`, `CacheMXBean`, `CacheManagerMXBean`

Anything consuming JMX externally (JConsole, VisualVM, Prometheus exporters, Grafana, Nagios, Zabbix, monex, etc.) that previously queried `InstanceId` on any of these MBeans is affected.

## Broader risk — other attributes may have been dropped too

The same commit removed the `get` prefix from several other methods on classes in `org.exist.management.impl`:

- `getPercentage`, `getOwner`, `getReferenceCount`, `getStack`, `getStackAcquired`, `getId`, `getErrcode`, `getDescription`, `getAction`, `getInfo`, `getKey`

Each one that was declared on (or inherited via) an MBean / MXBean interface has likely lost its JMX attribute the same way. Worth a careful before/after MBeanInfo diff rather than assuming only `InstanceId` is affected.

## Was this intentional?

The commit subject suggests not — pinging @dizzzz: was the API surface change deliberate (e.g., moving toward record-style accessors), or did the JMX naming-convention side effect slip through?

## Suggested fix

Revert the `get` prefix in `PerInstanceMBean.java` + the 8 impl classes — minimal change, restores all 9 MXBeans' `InstanceId` attribute. Same treatment for any other attribute identified by a before/after MBeanInfo diff.

## Credit

Caught during testing of the monex monitoring pane on the eXist-db org's `existdb-openapi` project — the missing attribute broke their integration test against `develop` HEAD.

joewiz:
## Summary

Closes #6379. Restores the `InstanceId` JMX attribute on all 9 MXBeans that extend `PerInstanceMBean`, which had been silently dropped by `9af8663a` (2026-05-11) when `getInstanceId()` was renamed to `instanceId()`. JMX uses JavaBean naming convention for attribute discovery — a bare-named `x()` is not an attribute.

@dizzzz confirmed the rename was unintentional in [#6379 comment](https://github.com/eXist-db/exist/issues/6379#issuecomment-4469709691).

## What changed

**Commit 1 — `bee8dbae79` (the fix):**

- `PerInstanceMBean.java` — restore `String getInstanceId();` in the interface contract.
- 8 concrete impl classes — restore `public String getInstanceId()`: `Database`, `CollectionCache`, `DiskUsage`, `LockTable`, `ProcessReport`, `SanityReport`, `Cache`, `CacheManager`.
- `JMXAgent.java` — update the 2 internal callers (`mbean.instanceId()` → `mbean.getInstanceId()`).
- `BinaryValues.java` — this is a Java `record` whose auto-generated `instanceId()` accessor matches the record component name but not the JavaBean convention. Added an explicit `getInstanceId()` delegate (`@Override public String getInstanceId() { return instanceId; }`) so the JMX contract is honored without losing the record-style modernization. **This is likely why the original rename happened** — Dannes converted `BinaryValues` to a record and the simplest way to satisfy the inherited interface was to rename the interface method. The delegate approach preserves both concerns.

**Commit 2 — `281996163e` (regression guard):**

- `JmxRemoteTest.checkContent()` gains 5 XPath assertions verifying `//jmx:<Bean>/jmx:InstanceId` is present on `Database`, `ProcessReport`, `CollectionCache`, `LockTable`, and `SanityReport`. Would have failed when `9af8663a` was merged.

## Scope clarification (vs. the issue body)

The issue flagged "broader risk" that other renamed methods in the same commit (`getPercentage`, `getOwner`, `getReferenceCount`, etc.) might have similarly broken JMX attributes. **Verified that risk did not materialize:** those renames were on plain classes / `CompositeData` helpers (`TaskStatus`, internal `Lock`/`Error`/`Job` classes), not on MBean/MXBean interfaces. JMX naming convention doesn't apply to them. Only `InstanceId` was affected at the JMX surface.

## Test plan

- [x] `mvn test -pl exist-core -Dtest=JmxRemoteTest` — 2/2 pass on the fix branch
- [x] `mvn compile -pl exist-core -am` — green
- [x] Codacy clean on all modified files (one pre-existing `AvoidThrowingRawExceptionTypes` in `LockTable.java` at lines 102/111/132/143 is unrelated to this change and out of scope for this PR)
- [ ] CI green

### DIFF
--- a/exist-core/src/main/java/org/exist/management/Cache.java
+++ b/exist-core/src/main/java/org/exist/management/Cache.java
@@ -62,7 +62,7 @@ public class Cache implements CacheMXBean {
     }
 
     @Override
-    public String instanceId() {
+    public String getInstanceId() {
         return instanceId;
     }
 
--- a/exist-core/src/main/java/org/exist/management/CacheManager.java
+++ b/exist-core/src/main/java/org/exist/management/CacheManager.java
@@ -61,7 +61,7 @@ public class CacheManager implements CacheManagerMXBean {
     }
 
     @Override
-    public String instanceId() {
+    public String getInstanceId() {
         return instanceId;
     }
 
--- a/exist-core/src/main/java/org/exist/management/impl/BinaryValues.java
+++ b/exist-core/src/main/java/org/exist/management/impl/BinaryValues.java
@@ -51,6 +51,17 @@ public record BinaryValues(String instanceId) implements BinaryValuesMXBean {
     }
 
     /**
+     * Explicit JavaBean-style accessor required by the MBean attribute discovery,
+     * which relies on the {@code getXxx()} naming convention. The record's
+     * auto-generated {@link #instanceId()} accessor does not satisfy that convention,
+     * so we delegate to it here. See <a href="https://github.com/eXist-db/exist/issues/6379">#6379</a>.
+     */
+    @Override
+    public String getInstanceId() {
+        return instanceId;
+    }
+
+    /**
      * Return a JMX object-name query that matches all BinaryValues MBeans across all instances.
      *
      * @return the wildcard query string
--- a/exist-core/src/main/java/org/exist/management/impl/CollectionCache.java
+++ b/exist-core/src/main/java/org/exist/management/impl/CollectionCache.java
@@ -54,7 +54,7 @@ public class CollectionCache implements CollectionCacheMXBean {
     }
 
     @Override
-    public String instanceId() {
+    public String getInstanceId() {
         return instance.getId();
     }
 
--- a/exist-core/src/main/java/org/exist/management/impl/Database.java
+++ b/exist-core/src/main/java/org/exist/management/impl/Database.java
@@ -67,7 +67,7 @@ public class Database implements DatabaseMXBean {
     }
 
     @Override
-    public String instanceId() {
+    public String getInstanceId() {
         return pool.getId();
     }
 
--- a/exist-core/src/main/java/org/exist/management/impl/DiskUsage.java
+++ b/exist-core/src/main/java/org/exist/management/impl/DiskUsage.java
@@ -77,7 +77,7 @@ public class DiskUsage implements DiskUsageMXBean {
     }
 
     @Override
-    public String instanceId() {
+    public String getInstanceId() {
         return instanceId;
     }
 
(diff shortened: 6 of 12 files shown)


reinhapa:
@joewiz the important part is that all `xxMBean` interfaces should use the Java Bean style definition for getters/setters using `getX()` /  `isX()` (for `boolean`values) and `setX(x)` all other methods are considered actions.

dizzzz:
PR looks right to me.

dizzzz (approved):
LGTM

dizzzz (approved):
LGTM

reinhapa (approved):

PR #6384, opened 2026-05-17
[refactor] Always sort internal-module FunctionDef[]; drop functionsOrdered flag (#6378)

### LINKED ISSUE #6378: AbstractInternalModule: `functionsOrdered=true` invariant is trusted but not enforced; silent function-lookup failure if array is unsorted
## Summary

`AbstractInternalModule.getFunctionDef(QName, int)` switches between binary search and linear scan based on the `functionsOrdered` constructor parameter. When `functionsOrdered=true`, it binary-searches the `FunctionDef[]` keyed on `FunctionId` (qname primary, arity secondary). The array is expected to be sorted in that order, but the invariant is trusted and never verified. If a subclass declares `ordered=true` and passes an unsorted array, binary search will silently fail to find some functions — they remain visible via linear-scan APIs (`util:registered-functions`, `inspect:inspect-module-uri`) but unreachable via direct call, `fn:function-lookup`, and function references.

## How first-party modules avoid this today

Of the 10 `exist-core`/extension modules that declare `functionsOrdered=true`:

- 5 (`FnModule`, `UtilModule`, `XMLDBModule`, `RequestModule`, `CounterModule`) carry a `static { Arrays.sort(functions, new FunctionComparator()); }` initializer block.
- 5 (`InspectionModule`, `ContentExtractionModule`, `ExiftoolModule`, `XQDocModule`, `ProcessModule`) declare their functions in already-sorted order, either by deliberate choice or by coincidence (some have only 1-2 entries).

The repeated static-sort idiom across 5 of these is essentially a workaround that experienced module authors learned to apply. There is no documentation of the invariant in `AbstractInternalModule`'s JavaDoc.

## Reproduction (third-party module)

```java
public class CustomModule extends AbstractInternalModule {
    public static final FunctionDef[] functions = {
        new FunctionDef(Foo.EVAL_SIG, Foo.class),    // local name "eval"
        new FunctionDef(Foo.FETCH_SIG, Foo.class),   // local name "fetch"
        new FunctionDef(Foo.CLOSE_SIG, Foo.class),   // local name "close" — sorts BEFORE eval
    };
    public CustomModule(Map<String, List<?>> p) {
        super(functions, p, true);   // ordered=true, but array isn't sorted
    }
    // ...
}
```

Symptom:

- `util:registered-functions("...custom")` returns `custom:close, custom:eval, custom:fetch` (linear scan)
- `function-lookup(xs:QName("custom:close"), 1)` returns the empty sequence (binary search miss)
- `custom:close($x)` raises `err:XPST0017 "Function custom:close() is not defined in module namespace..."`
- `custom:close#1` raises the same

Renaming `close` to anything that sorts after `eval` (e.g. `release`) makes the function reachable again — purely by coincidence of where the sort happens to land.

This was originally reported in #6376; the renamed `cursor:release` workaround in [existdb-openapi#18](https://github.com/eXist-db/existdb-openapi/pull/18) is what made the symptom visible.

## Fix options worth considering

Three flavors, no preference — each has a real trade-off:

1. **Defensive sort + WARN in the constructor.** When `ordered=true`, verify and re-sort if needed (defensive `clone()` so the caller's `static final` array isn't mutated). Logs once per class. Pros: third-party safety net, no API change. Cons: per-instance `isSorted` walk (small but real, since `XQueryContext` constructs modules per query); could memoize per `Class<?>` to make it O(1) after first construction.

2. **Drop `functionsOrdered` entirely; always sort, always binary-search.** Sort once at static init in the base class via a helper. No flag, no footgun, removes the redundant static blocks from the 5 first-party modules. Cons: API change — breaks any third-party code using the 3-arg constructor.

3. **Document and rely on a build-time CI test.** Keep the constructor as-is; add a JUnit test that fails the build if any first-party `ordered=true` module's `FunctionDef[]` isn't sorted. Cons: doesn't help third-party authors who hit the bug in their own builds; the `static { Arrays.sort(...) }` boilerplate stays.

Hybrid options exist (e.g., 1+3: runtime safety net + build test). The narrow surface (only matters when `ordered=true` AND array isn't sorted AND no manual sort block) suggests this isn't worth a large refactor; option 3 alone may be enough if the API surface is considered fully internal.

## Why this report and not a PR

A draft PR taking approach (1) was prepared but not opened, since the choice between the three options is a judgement call best made by the core devs who own the module-registration contract. Happy to follow up with whichever approach is preferred.

joewiz:
## Summary

Closes #6378. Implements Option 2 per @duncdrum's preference on the issue thread:

> FWIW I m leaning heavily towards 2. There is a whole cluster of bugs we encountered recently where inconsistent sorting leads to hard to debug and nasty bugs.

The `functionsOrdered=true` opt-in flag on `AbstractInternalModule`'s 3-arg constructor was an unenforced precondition: callers had to pass a `FunctionDef[]` already sorted by `FunctionId` order, or binary search in `getFunctionDef()` would silently miss entries. The footgun manifested as a real bug in #6376 (`cursor:close` unreachable via `function-lookup` but visible via `util:registered-functions`).

The fix removes the flag entirely. The base class always sorts (defensive-copy + `Arrays.sort` in the constructor) and always uses binary search. Five first-party modules that previously carried `static { Arrays.sort(functions, new FunctionComparator()); }` boilerplate no longer need it.

## What changed

**Commit 1 — base class** (`AbstractInternalModule.java`, +20 / -24):
- Remove the 3-arg `(FunctionDef[], Map, boolean)` constructor.
- 2-arg constructor now defensive-copies and sorts (caller's `static final` array left intact).
- Remove the `ordered` field and the conditional in `getFunctionDef()`; binary search is unconditional.
- Reorder field declarations above the `FunctionComparator` inner class (PMD `FieldDeclarationsShouldBeAtStartOfClass`).

**Commit 2 — 20 callers + 5 static-sort cleanups** (+167 / -50):

Callers updated (drop the third argument):

| Module | Was |
|---|---|
| `BackupModule`, `FnModule`, `InspectionModule`, `RequestModule`, `UtilModule`, `XMLDBModule`, `ContentExtractionModule`, `ExiftoolModule`, `XQDocModule`, `CounterModule`, `ProcessModule` | `super(functions, parameters, true)` |
| `ArrayModule`, `MapModule`, `WebSocketModule`, `ConsoleCompatModule`, `VectorModule`, `LuceneModule`, `SortModule`, `NGramModule`, `RangeIndexModule` | `super(functions, parameters, false)` |

Static-sort blocks removed (`static { Arrays.sort(functions, new FunctionComparator()); }`) from `XMLDBModule`, `UtilModule`, `RequestModule`, `FnModule`, `CounterModule` — plus the now-unused `java.util.Arrays` imports.

Net: ~34 lines of boilerplate gone from the codebase.

**Regression test** (`AbstractInternalModuleSortTest`, new — 4 cases):
- `unsortedDeclarationOrderStillFindsAllFunctions` — the #6376 reproducer, now safe by construction.
- `callerArrayIsNotMutated` — verifies the defensive copy.
- `alreadySortedArrayFindsAllFunctions` — common-case sanity.
- `sameQnameDifferentAritiesAllFound` — secondary sort key (arity) is honored.

## API break

The 3-arg `AbstractInternalModule(FunctionDef[], Map<String, List<?>>, boolean)` constructor goes away. Any third-party module passing `super(functions, parameters, true|false)` needs to switch to `super(functions, parameters)`. One-line change per module. The `FunctionComparator` static inner class remains public for any code that uses it elsewhere.

@duncdrum explicitly accepted this trade-off on the issue: removing the flag is what makes the bug structurally impossible.

## Performance

The sort is O(N log N) once per module instance construction. eXist's largest module is `FnModule` (~500 functions), sorted in microseconds. `AbstractInternalModule` is instantiated per `XQueryContext`, which is per-query — so the sort runs per query. For comparison, the previous `static { Arrays.sort(...) }` blocks ran once at class load. The cost is small but real; happy to memoize per-class if reviewers want to eliminate it entirely (one-time check via `ConcurrentMap<Class<?>, Boolean>`).

## Test plan

- [x] `mvn test -pl exist-core -Dtest=AbstractInternalModuleSortTest` — 4/4 pass
- [x] `mvn compile` across exist-core + extensions (minus exist-xqts, which has unrelated GitHub Packages auth issue locally) — green
- [x] Codacy clean on changed files
- [x] CI green

## Related

- Issue #6378 — this PR closes it
- Issue #6376 — original symptom report (cursor:close in existdb-openapi)
- The earlier draft PR #6377 (which took Option 1) was closed in favor of #6378 + this approach.

### DIFF
--- a/exist-core/src/main/java/org/exist/backup/xquery/BackupModule.java
+++ b/exist-core/src/main/java/org/exist/backup/xquery/BackupModule.java
@@ -49,7 +49,7 @@ public class BackupModule extends AbstractInternalModule
 
     public BackupModule(final Map<String, List<?>> parameters)
     {
-        super( functions, parameters, true );
+        super(functions, parameters);
     }
 
     public String getNamespaceURI()
--- a/exist-core/src/main/java/org/exist/xquery/AbstractInternalModule.java
+++ b/exist-core/src/main/java/org/exist/xquery/AbstractInternalModule.java
@@ -36,10 +36,18 @@ import javax.annotation.Nullable;
  * in name or the number of expected arguments. It is thus possible to implement
  * similar XQuery functions in one single class.
  *
+ * <p>The {@code FunctionDef[]} passed in does not need to be sorted; this constructor
+ * defensive-copies and sorts it by {@link FunctionId} order so {@link #getFunctionDef(QName, int)}
+ * can always use binary search. The original array reference is not mutated.</p>
+ *
  * @author <a href="mailto:wolfgang@exist-db.org">Wolfgang Meier</a>
  * @author ljo
  */
 public abstract class AbstractInternalModule implements InternalModule {
+
+    protected final FunctionDef[] mFunctions;
+    protected final Map<QName, Variable> mGlobalVariables = new HashMap<>();
+    private final Map<String, List<?>> parameters;
 
     public static class FunctionComparator implements Comparator<FunctionDef> {
         @Override
@@ -48,21 +56,39 @@ public abstract class AbstractInternalModule implements InternalModule {
         }
     }
 
-    protected final FunctionDef[] mFunctions;
-    protected final boolean ordered;
-    private final Map<String, List<?>> parameters;
-
-    protected final Map<QName, Variable> mGlobalVariables = new HashMap<>();
-
     public AbstractInternalModule(final FunctionDef[] functions, final Map<String, List<?>> parameters) {
-        this(functions, parameters, false);
+        // Defensive-copy + sort so the caller's static final array is left intact
+        // and getFunctionDef() can binary-search regardless of declaration order.
+        // See https://github.com/eXist-db/exist/issues/6378 (and #6376 which surfaced
+        // the latent bug).
+        if (functions != null && functions.length > 1) {
+            final FunctionDef[] sorted = functions.clone();
+            Arrays.sort(sorted, new FunctionComparator());
+            this.mFunctions = sorted;
+        } else {
+            this.mFunctions = functions;
+        }
+        this.parameters = parameters;
     }
 
+    /**
+     * Pre-#6378 constructor that took a {@code functionsOrdered} flag. The flag is now
+     * ignored — the function table is always sorted and {@link #getFunctionDef(QName, int)}
+     * always uses binary search. Retained so external modules continue to compile against
+     * eXist 7 without source changes; call sites should migrate to
+     * {@link #AbstractInternalModule(FunctionDef[], Map)}.
+     *
+     * @param functions         the array of functions
+     * @param parameters        configuration parameters
+     * @param functionsOrdered  ignored as of #6378
+     *
+     * @deprecated since 7.0.0; the {@code functionsOrdered} parameter has no effect.
+     *             Use {@link #AbstractInternalModule(FunctionDef[], Map)}.
+     */
+    @Deprecated(since = "7.0.0", forRemoval = true)
     public AbstractInternalModule(final FunctionDef[] functions, final Map<String, List<?>> parameters,
-                                  final boolean functionsOrdered) {
-        this.mFunctions = functions;
-        this.ordered = functionsOrdered;
-        this.parameters = parameters;
+                                  @SuppressWarnings("unused") final boolean functionsOrdered) {
+        this(functions, parameters);
     }
 
     @Override
@@ -113,17 +139,7 @@ public abstract class AbstractInternalModule implements InternalModule {
 
     @Override
     public FunctionDef getFunctionDef(QName qname, int arity) {
-        final FunctionId id = new FunctionId(qname, arity);
-        if (ordered) {
-            return binarySearch(id);
-        } else {
-            for (FunctionDef mFunction : mFunctions) {
-                if (id.compareTo(mFunction.getSignature().getFunctionId()) == 0) {
-                    return mFunction;
-                }
-            }
-        }
-        return null;
+        return binarySearch(new FunctionId(qname, arity));
     }
 
     private FunctionDef binarySearch(final FunctionId id) {
--- a/exist-core/src/main/java/org/exist/xquery/functions/array/ArrayModule.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/array/ArrayModule.java
@@ -68,7 +68,7 @@ public class ArrayModule extends AbstractInternalModule {
     );
 
     public ArrayModule(Map<String, List<?>> parameters) {
-        super(functions, parameters, false);
+        super(functions, parameters);
     }
 
     static FunctionSignature functionSignature(final String name, final String description, final FunctionReturnSequenceType returnType, final FunctionParameterSequenceType... paramTypes) {
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FnModule.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FnModule.java
@@ -21,7 +21,6 @@
  */
 package org.exist.xquery.functions.fn;
 
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
@@ -277,17 +276,12 @@ public class FnModule extends AbstractInternalModule {
         new FunctionDef(FunContainsToken.FS_CONTAINS_TOKEN[0], FunContainsToken.class),
         new FunctionDef(FunContainsToken.FS_CONTAINS_TOKEN[1], FunContainsToken.class)
     };
-
-    static {
-        Arrays.sort(functions, new FunctionComparator());
-    }
-
     public final static ErrorCodes.ErrorCode SENR0001 = new ErrorCodes.ErrorCode("SENR0001", "serialization error in fn:serialize");
     public final static ErrorCodes.ErrorCode SEPM0019 = new ErrorCodes.ErrorCode("SEPM0019", "It is an error if an instance of the data model " +
             "used to specify the settings of serialization parameters specifies the value of the same parameter more than once.");
 
     public FnModule(Map<String, List<?>> parameters) {
-        super(functions, parameters, true);
+        super(functions, parameters);
     }
 
     @Override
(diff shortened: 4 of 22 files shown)


duncdrum (approved):
very nice

duncdrum:
we should audit our org to see if any custom extensions where using the 3 arg version

joewiz:

Org-wide audit done via `gh search code` across all repos in the `eXist-db` org. Three-arg `super(functions, parameters, true|false)` calls live in **6 modules across 5 external repos**:

| Repo | Module | Flag value | Static sort block? |
|---|---|---|---|
| [`existdb-openapi`](https://github.com/eXist-db/existdb-openapi) | `ApiModule.java` | `true` | — |
| [`existdb-openapi`](https://github.com/eXist-db/existdb-openapi) | `LspModule.java` | `true` | — |
| [`exist-http-client`](https://github.com/eXist-db/exist-http-client) | `HttpClientModule.java` | `true` | — |
| [`exist-request`](https://github.com/eXist-db/exist-request) | `RequestModule.java` | `true` | — |
| [`exist-stanford-nlp`](https://github.com/eXist-db/exist-stanford-nlp) | `StanfordNLPModule.java` | `false` | — |
| [`monex`](https://github.com/eXist-db/monex) | `ConsoleModule.java` | `false` | — |

None of these carry a `static { Arrays.sort(functions, new FunctionComparator()); }` block, so the fix per module is a single line: drop the third argument.

**13 other external repos that extend `AbstractInternalModule` already use the 2-arg form** — no action needed: `exist-alt-identifier`, `exist-apps-archetype` (the official module archetype/template — already correct), `exist-binary`, `exist-commons-text`, `exist-file`, `exist-zip`, `expath-bin-module`, `expath-crypto-module`, `jfreechart`, `messaging-replication` (3 modules), `xquery-versioning-module`, `xquery-xproc-xmlcalabash-module`.

**All static-sort blocks in the org are inside this repo** — the five removed in this PR (`FnModule`, `UtilModule`, `XMLDBModule`, `RequestModule`, `CounterModule`) are the complete set. No external repo carries that boilerplate.

**Coordination options for the 6 affected modules:**

1. **Companion PRs in each repo, opened concurrently with this one.** Each is a one-line change. I can draft all 6 if helpful. They can sit unmerged until this PR lands, then merge in sequence with a fresh eXist 7-SNAPSHOT dep.
2. **Wait for this PR to merge first**, then open the 6 companion PRs. Cleaner causal narrative ("eXist removed the flag, here's the matching update"). Slightly longer total window where those repos can't build against eXist `develop`.
3. **Land this PR alongside `eXist 7.0` release**, since most of these external repos are typically updated only against released eXist versions anyway. Some may not have active CI against `develop`.

My lean is **(1) or (2)** — the change is trivial, doing it sooner avoids forcing external maintainers to scramble. Happy to take on drafting the 6 PRs.

duncdrum:
1  the sooner we open the PR the better

joewiz:

Companion PRs open across the 5 affected external repos (Option 1 per your preference):

| Repo | PR | Modules |
|---|---|---|
| `existdb-openapi` | https://github.com/eXist-db/existdb-openapi/pull/21 | `ApiModule`, `LspModule` |
| `exist-http-client` | https://github.com/eXist-db/exist-http-client/pull/3 | `HttpClientModule` |
| `exist-request` | https://github.com/eXist-db/exist-request/pull/2 | `RequestModule` |
| `exist-stanford-nlp` | https://github.com/eXist-db/exist-stanford-nlp/pull/347 | `StanfordNLPModule` |
| `monex` | https://github.com/eXist-db/monex/pull/384 | `ConsoleModule` |

Each is the same one-line change: drop the third argument from `super(functions, parameters, true|false)`. PR bodies cross-link back to this PR + #6378 + #6376 so future readers can trace the upstream context.

Expected behavior: each companion PR will fail CI until this PR merges and a fresh `7.0.0-SNAPSHOT` propagates — they're extending a class whose 3-arg constructor no longer exists. That's the queued-companion shape, not a regression.

Once this PR merges, the 5 companions are ready to follow.

duncdrum:
@line-o curious to hear what you think. I d like to see this breaking in the beta so that folks have a chance to catch it. 

I also think it ll solve a whole class of related downstream bugs.

line-o:
Agreed. We missed the Chance to deprecate this in the last 6.4.1 release. Next best thing is to allow to discover it now.
And the possibility to add functions unsorted and hide them that way from discovery should be stopped.

line-o:
Wait! @joewiz @duncdrum I just realizd that the 3-arg constructor is removed rather than marked deprecated and the third parameter simply ignored.
What was the reason to break hard instead of just doing the right thing and communicating this openly?

duncdrum:
@line-o because this is mainly a bug fix. The ability to work with unsorted values should never have been there. 

3arg true is the new default. 
Old 2arg or 3arg false are buggy they shouldn't be deprecated but fixed.

line-o (changes requested):
Please reinstate a 3-argument constructor and mark it as deprecated.

The third parameter is ignored and all functions will always be sorted.

joewiz:

Thanks @line-o — done in `15181112fc`. The 3-arg `AbstractInternalModule(FunctionDef[], Map, boolean)` constructor is reinstated and marked `@Deprecated(since = "7.0.0", forRemoval = true)`. The `functionsOrdered` parameter is documented as ignored; the constructor delegates to the 2-arg form, which always sorts.

External modules compiled against pre-#6378 eXist that still use `super(functions, parameters, true)` or `super(functions, parameters, false)` will continue to compile against eXist 7 with a deprecation warning, no source changes required. Verified with a new `deprecatedThreeArgConstructorStillWorks` test in `AbstractInternalModuleSortTest` that constructs a module via the legacy 3-arg form with both flag values and confirms function lookup still works.

The internal callers (20 first-party modules) still drop the spurious third arg in the second commit of this PR — they don't need the deprecation path since they're in the same repo. Keeps the codebase from emitting deprecation warnings against itself.

The 5 external companion PRs we opened (`existdb-openapi#21`, `exist-http-client#3`, `exist-request#2`, `exist-stanford-nlp#347`, `monex#384`) are no longer strictly necessary — those repos would compile against this PR with just a warning. They're still worth merging as cleanup (drop the spurious flag), just not urgent. I'll re-comment on each to reflect this.

line-o (approved):
Thanks for the change, I believe that will make our life easier.

PR #6390, opened 2026-05-20
Dp fix jetty 12.1

duncdrum:
There are two issues uncovered by the jetty bump. 

The first is a change affecting win path construction. Which puts `:` from eg. `D:\Documents...` into urls. Making them plain broken. 

The second is a hardened availability status check. Exist is running into, because we never checked for availability proper, just readiness. 

This is quite messy. I would like to have this cleaner. But so far my attempts at cleaning all lead to new test failures. So looking for help

### DIFF
--- a/exist-core/src/main/java/org/exist/indexing/IndexController.java
+++ b/exist-core/src/main/java/org/exist/indexing/IndexController.java
@@ -40,9 +40,11 @@ import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import org.exist.security.PermissionDeniedException;
 
 /**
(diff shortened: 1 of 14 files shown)


exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -506,6 +508,60 @@ private Optional<Server> startJetty(final List<Object> configuredObjects) throws
         return server;
     }
 
+    /**
+     * Jetty may accept TCP connections and report {@link WebAppContext#isStarted()} before the context
+     * is {@link org.eclipse.jetty.server.handler.ContextHandler#isAvailable()}, in which case HTTP
+     * clients see {@code 503 Service Unavailable} (observed on Windows CI).
```
dizzzz: I see same issues on MacOS

duncdrum:
Ok I need help here. I m not 100% happy with this yet, but so far this is the only consistently green patch I could come up with

exist-core/src/main/java/org/exist/indexing/IndexController.java
```diff
@@ -56,7 +58,16 @@ public enum CollectionIndexRemovalMode {
         CONFIG_ONLY_REINDEX
     }
 
-    private final Map<String, IndexWorker> indexWorkers = new HashMap<>();
+    /**
```
dizzzz, line 61: Is this the correct PR?

dizzzz:
Based on a remark you wrote, waiting for an upstream update might be wise too?
For the other code changes.... I need to find the time to understand them

dizzzz:
Green build is already impressive :-)

exist-core/src/main/java/org/exist/indexing/IndexController.java
duncdrum: Yes something that came up during the 20 or so runs leading up to this changeset. 

Basically when running tests we can have a racy situation based on jetty saying I'm ready and indexing starting. 

But as I said happy for ways to make this cleaner.

duncdrum:
Yes upstream will likely address the problem fixed by the first commit. 

The available != ready is here to stay, and will need some adjustments on our side.

exist-core/src/main/java/org/exist/indexing/IndexController.java
duncdrum: I ll restructure the commits to make this cleaner, and push some cleanups in a few

duncdrum:
I m going to close. This is not it. I have another attempt that should be able to land soon.

PR #6364, opened 2026-05-13
[feature] Replace Milton WebDAV with Apache Jackrabbit (Level 2 + dead properties + litmus compliance)

joewiz:
## Summary

Replaces eXist's Milton-based WebDAV stack with Apache Jackrabbit WebDAV (Jakarta EE 10 transformed). Brings full WebDAV Level 2 locking compliance per RFC 4918, dead properties (PROPPATCH/PROPFIND on arbitrary properties), shallow copy semantics, lock persistence across server restarts, and litmus-suite compliance testing in CI.

## Why now

Builds on the Jetty 12 / Jakarta Servlet 6.0 work landed in #6145. Modernizing the WebDAV transport alongside the rest of the HTTP stack means eXist 7 ships with a coherent modern web layer. Milton has been effectively unmaintained for several years; Jackrabbit WebDAV is actively maintained by the Apache Jackrabbit project.

### Core replacement (`0e29e29273`)
- Replace Milton WebDAV (5 classes deleted: `MiltonCollection`, `MiltonDocument`, `MiltonResource`, `MiltonWebDAVServlet`, `ExistResourceFactory`) with Apache Jackrabbit's `jackrabbit-webdav` 2.22.3 (Jakarta EE 10 edition, published under `org.exist-db.thirdparty.org.apache.jackrabbit`)
- New servlet `ExistWebdavServlet` extends `AbstractWebdavServlet` with Basic Auth session provider
- New adapters: `ExistDavResourceFactory`, `ExistDavCollection`, `ExistDavDocument`, `ExistDavResource`, `ExistLockManager`, `ExistDavSession`
- Map `/webdav/*` directly in `web.xml`; remove controller-config forward
- Fix MKCOL (create `ExistDavCollection` for non-existing paths on MKCOL requests)
- Fix distribution deployment and locator path generation

### Dead properties + shallow copy + lock persistence (`766a03eda3`)
- `DeadPropertyStore`: store/retrieve arbitrary DAV properties via `ExistResource.getMetaData()` / `updateMetaData()`
- COPY with `Depth: 0` performs shallow copy (collection only, not its members)
- `WebDavLockStore`: serialise/deserialise locks to XML documents in `/db/system/webdav-locks/`, created automatically on startup

### WebDAV Level 2 locking (`31485ca635`)
- Collection locks: LOCK on a collection locks all members (DELETE/PUT/COPY of any member requires the collection lock token)
[...]

PR #6392, opened 2026-05-22
jetty fixes II

duncdrum:
## Summary

Follow-up fixes from the Jetty 12.1 upgrade work in #6390. Addresses Windows CI failures, flaky integration-test startup, persistent-login cookies under Jetty 12, and several issues found while hardening the test harness.

see https://github.com/jetty/jetty.project/issues/15011
### Jetty 12.1 runtime and Windows

- Bump Jetty group (7 dependency updates) on top of #6390
- WindowsPathResource — workaround for Jetty 12.1 PathResource.resolve() regression on Windows drive URIs (/D:/...), which left webapps available=false and returned 503 on every path
- throwUnavailableOnStartupException in deploy configs so startup failures surface immediately
- WindowsPathResourceIT — Windows-only regression test in exist-webdav (Failsafe on Windows CI)
- Hardening: symmetric equals, double-wrap guard, OSUtil.isWindows() consistency, cleaner startup failure logging

### Integration test reliability

- JettyStart.awaitWebAppContextsStarted() — block until required WebAppContext instances are isAvailable() (not just started)
- ExistWebServer — builder API, distribution-mode portal dir, embedded startup failure detail on timeout
- JettyStartListener — replaces deprecated Observable for lifecycle signals
- org.exist.jetty.startup.timeout.ms — configurable readiness timeout (180s on CI)
- Remove redundant TCP probe from LoginModuleIT
- ControllerTest — @ClassRule (one server per class)
- Jetty WARN logging in integration-test log4j2.xml configs

### Persistent login (Jetty 12 cookies)

- Map empty servlet context path to Path=/ in login.xql and HttpResponseWrapper
- Use | token separator (strict Set-Cookie parsers reject : in unquoted cookie values)
- Dual-separator parsing (| first, : legacy fallback) plus unit tests
- CookieSpecs.STANDARD and shared cookie store in LoginModuleIT / AbstractHttpTest for Jetty 12 RFC6265 Set-Cookie (jetty/jetty.project#12771)

### Other fixes

- MimeTable.load() — fail loudly on missing or invalid mime-types.xml instead of silent empty maps
- IndexController — explicit index worker chain ordering (structural, statistics, Lucene, then others) instead of undefined HashMap iteration
- JettyStart.run() — refactor for Codacy NPath threshold (WebAppReadinessAwaiter, extracted startup phases)

## References

- Follow-up to #6390
- https://github.com/jetty/jetty.project/issues/12771
- Related: #6364 (Jackrabbit WebDAV — webdav module gets HC4 retention comment only in this PR)

## Test plan

- [x] exist-core unit tests (MimeTableTest, IndexIntegrationTest, ControllerTest, URLRewritingTest)
- [x] LoginModuleIT and PersistentLoginTest
- [x] Windows CI: WindowsPathResourceIT (Failsafe, verify -DskipUnitTests=true)
- [x] Sample webdav JUnit tests (HC4 milton-client harness unchanged)
- [x] Full CI green on Linux and Windows integration jobs
- [ ] Manual smoke: standalone launcher and dashboard login with persistent-login module

closes #6367

### DIFF
--- a/exist-core/src/main/java/org/exist/jetty/JettyStart.java
+++ b/exist-core/src/main/java/org/exist/jetty/JettyStart.java
@@ -135,20 +135,13 @@ public class JettyStart extends Observable implements LifeCycle.Listener {
         System.out.println(msg); //NOSONAR this has to go to the console
     }
 
-    private static void consoleErr(final String msg) {
-        System.err.println(msg); //NOSONAR surfaced in Surefire output when test log4j root is OFF
-    }
-
     private synchronized void recordStartupFailure(final String detail, final Throwable cause) {
         webAppStartedSuccessfully = false;
         webAppStartupFailureDetail = detail;
         if (cause != null) {
             logger.fatal("Jetty startup failed: {}", detail, cause);
-            consoleErr("Jetty startup failed: " + detail);
-            cause.printStackTrace(System.err); //NOSONAR CI diagnostics when log4j is disabled in tests
         } else {
             logger.fatal("Jetty startup failed: {}", detail);
-            consoleErr("Jetty startup failed: " + detail);
         }
     }
 
@@ -856,8 +849,8 @@ public class JettyStart extends Observable implements LifeCycle.Listener {
 
     /**
      * When {@link #isWebAppStartedSuccessfully()} is {@code false}, holds the last startup failure
-     * message for test diagnostics (also printed to {@code System.err} because module test log4j
-     * configs often set {@code Root level="OFF"}).
+     * message for test diagnostics (surfaced by {@link org.exist.test.ExistWebServer} in thrown
+     * {@link IllegalStateException}s).
      */
     public synchronized Optional<String> getWebAppStartupFailureDetail() {
         return Optional.ofNullable(webAppStartupFailureDetail);
--- a/exist-core/src/main/java/org/exist/jetty/JettyStart.java
+++ b/exist-core/src/main/java/org/exist/jetty/JettyStart.java
@@ -202,203 +202,27 @@ public class JettyStart implements LifeCycle.Listener {
             return;
         }
 
-        Path jettyConfig = Path.of(args[0]).normalize();
-        boolean configFromClasspath = false;
-        if (Files.notExists(jettyConfig)) {
-            logger.warn("Configuration file: {} does not exist!", jettyConfig.toAbsolutePath().toString());
-
-            final String jettyConfigFileName = FileUtils.fileName(jettyConfig.getFileName());
-            logger.warn("Fallback... searching for configuration file on classpath: {}!etc/{}", getClass().getPackage().getName(), jettyConfigFileName);
-
-            final URL jettyConfigUrl = getClass().getResource("etc/" + jettyConfigFileName);
-            if (jettyConfigUrl != null) {
-                try {
-                    jettyConfig = Path.of(jettyConfigUrl.toURI()).normalize();
-                    configFromClasspath = true;
-                } catch (final URISyntaxException e) {
-                    logger.error("Unable to retrieve configuration file from classpath: {}", e.getMessage(), e);
-                    return;
-                }
-            } else {
-                logger.error("Unable to find configuration file on classpath!");
-                return;
-            }
+        final Optional<ResolvedJettyConfig> resolvedConfig = resolveJettyConfigPath(args[0]);
+        if (resolvedConfig.isEmpty()) {
+            return;
         }
 
         final Map<String, String> configProperties;
         try {
-            configProperties = getConfigProperties(jettyConfig.getParent());
-
-            // modify JETTY_HOME and JETTY_BASE properties when running with classpath config
-            if (configFromClasspath) {
-                final String jettyClasspathHome = jettyConfig.getParent().getParent().toAbsolutePath().toString();
-                System.setProperty(JETTY_HOME_PROP, jettyClasspathHome);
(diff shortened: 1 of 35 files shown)


joewiz (commented):

Nice work @duncdrum — read through end-to-end, all 10 commits are clean and CI green across the full matrix (ubuntu/macOS/windows integration, ubuntu unit, container, W3C). The disciplined commit structure makes this very pleasant to follow. A few observations worth flagging for future maintainability; none are blockers.

### 1. `WindowsPathResource extends Resource` (not `PathResource`)

The docstring acknowledges that Jetty internals using `instanceof PathResource` won't recognise wrapped resources, and says CI validates the exploded-webapp path doesn't depend on that. Worth asking — have you also audited eXist's own code for `instanceof PathResource` (or `getClass() == PathResource.class`) call sites that might silently get the wrong answer for a wrapped resource? A quick grep should put that to bed if not already done.

### 2. `MimeTable` parameterised `getInstance` overloads

Each variant is correctly double-checked-locked individually, but if two threads call *different* overloads concurrently during init (e.g. `getInstance()` and `getInstance(Path)`), the singleton seed is whichever wins the race — they don't coordinate with each other. In practice the singleton is initialised once at startup so this is latent, but a single shared lock (or making the three overloads delegate to one canonical entry point) would close the door on the footgun. Optional polish, not a blocker.

### 3. `LEGACY_TOKEN_SEPARATOR` retention

The dual-separator parsing in `PersistentLogin` is the right call for backward compat with cookies already in users' browsers. A one-line comment along the lines of *"safe to remove once all v7-era persistent-login cookies have expired — target eXist 8.x"* would help a future maintainer know when to retire it. (Browser cookies expire; we don't need this forever.)

### 4. `CHAIN_PRIORITY_*` constants

The gaps (`0 → 100 → 1000 → MAX_VALUE`) are reasonable for future insertion room, but a one-line comment explaining the spacing — e.g. *"100s reserve room for pre-Lucene workers, 1000+ for Lucene-and-later, MAX for legacy unranked"* — would help future contributors pick a value for a new index without guessing.

### Otherwise

The substantive engineering choices are all well-motivated:

- `JettyStart`'s migration off deprecated `Observable` to a typed listener API
- `IndexController` swapping `HashMap` → `LinkedHashMap` + explicit comparator (fixes a real latent ordering bug)
- `HttpResponseWrapper`'s explicit empty-Path handling for standalone Jetty's `getContextPath()=""`
- `ExistWebServer.builder()` replacing the boolean-arg constructor chain
- The Windows-only IT in `exist-webdav` (gated to Failsafe) — exactly the right shape

LGTM modulo the observations above.

joewiz:
@duncdrum Let me know if the comments above were useful at all or not. Happy to pass this info along whenever you'd like.

duncdrum:
Thanks @joewiz 

- **PathResource audit:** grep confirms no eXist production `instanceof PathResource` call sites (only `wrapIfNeeded` + the IT). Javadoc updated.
- **MimeTable overloads:** fair point — pre-existing, leaving for a follow-up PR.
- **LEGACY_TOKEN_SEPARATOR:** retirement note added, gated on HC4 Phase 5 in #6393.
- **CHAIN_PRIORITY spacing:** comment added.

Also noted a test gap for the portal. So I added `PortalRedirectTest` for distribution `GET /`, and switched the portal to `org.exist.jetty.WebAppContext`.

line-o (approved):
It's a lot but it it's also a lot of good stuff in one PR

duncdrum:
Thx @line-o yes it was quite a journey. 

The next one is coming up with the http client. The changes here should make our tests less flaky, and the interaction with Jetty more robust. It sets us on the path to have login work properly again. So yay for that.

dizzzz (commented):
LGTM

exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -155,210 +178,243 @@ public synchronized void run(final boolean standalone) {
...
+                        .map(Path::normalize).map(Path::toAbsolutePath).map(Path::toString)
+                        .orElse("<UNKNOWN>"));
 
-            boolean establishedPrimaryPort = false;
-            for(final Connector connector : connectors) {
-                if(connector instanceof NetworkConnector networkConnector) {
+        final Optional<Observer> brokerPoolObserver = listener instanceof Observer observer
+                ? Optional.of(observer)
+                : Optional.empty();
+        BrokerPool.configure(1, 5, config, brokerPoolObserver);
```
dizzzz, line 297: should the values 1 & 5 be configurable? Or did we do it this way already?

exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -155,210 +178,243 @@ public synchronized void run(final boolean standalone) {
...
-                    config.getConfigFilePath()
-                        .map(Path::normalize).map(Path::toAbsolutePath).map(Path::toString)
-                        .orElse("<UNKNOWN>"));
+    private Optional<ResolvedJettyConfig> resolveJettyConfigPath(final String configArg) {
+        Path jettyConfig = Path.of(configArg).normalize();
+        if (Files.exists(jettyConfig)) {
+            return Optional.of(new ResolvedJettyConfig(jettyConfig, false));
+        }
 
-            BrokerPool.configure(1, 5, config, Optional.ofNullable(observer));
```
line-o, line 239: @dizzzz yes we did

line-o (approved):
Thanks @duncdrum

dizzzz (approved):

Issue #794, opened 2015-09-23
Lucene KeywordAnalyzer cannot be found


joewiz:
Lucene's KeywordAnalyzer cannot be found on a standard trunk build.  When launching eXist with the following collection.xconf:

``` xml
<collection xmlns="http://exist-db.org/collection-config/1.0" xmlns:test="test">
    <index xmlns:xs="http://www.w3.org/2001/XMLSchema">
        <fulltext default="none" attributes="false"/>
        <lucene>
            <analyzer class="org.apache.lucene.analysis.standard.StandardAnalyzer"/>
            <analyzer id="ws" class="org.apache.lucene.analysis.core.WhitespaceAnalyzer"/>
            <analyzer id="keyword" class="org.apache.lucene.analysis.core.KeywordAnalyzer"/>
            <text qname="test:para"/>
            <text qname="test:para" field="para-standard"/>
            <text qname="test:para" field="para-ws" analyzer="ws"/>
            <text qname="test:para" field="para-keyword" analyzer="keyword"/>
        </lucene>
    </index>
</collection>
```

exist.log reports the following error:

```
23 Sep 2015 14:40:15,394 [main] DEBUG (AnalyzerConfig.java [createInstance]:261) - Using analyzer org.apache.lucene.analysis.standard.StandardAnalyzer
23 Sep 2015 14:40:15,395 [main] DEBUG (AnalyzerConfig.java [createInstance]:261) - Using analyzer org.apache.lucene.analysis.core.WhitespaceAnalyzer
23 Sep 2015 14:40:15,395 [main] ERROR (AnalyzerConfig.java [createInstance]:269) - Could not find matching analyzer class constructororg.apache.lucene.analysis.core.KeywordAnalyzer: org.apache.lucene.analysis.core.KeywordAnalyzer.<init>(org.apache.lucene.util.Version)
java.lang.NoSuchMethodException: org.apache.lucene.analysis.core.KeywordAnalyzer.<init>(org.apache.lucene.util.Version)
    at java.lang.Class.getConstructor0(Class.java:3082) ~[?:1.8.0_60]
    …
```

For a reproducible test, including files, see this thread from exist-open: http://markmail.org/message/heesn3keggjkkmbg.

adamretter:
@joewiz Can you post a few more lines from the stacktrace? At least down to a few lines which are in the `org.exist` namespace.

joewiz:
Sure, here is the full stacktrace:

```
[...]

PR #6396, opened 2026-05-25
[bugfix] Don't fail module imports when load path is not relativizable

joewiz:
## Summary

When a client sends a synthetic module load path that isn't a real collection URI -- e.g. eXide sends `xmldb:exist://__new__1` for unsaved in-memory buffers -- `XQueryContext.resolveInEXPathRepository` throws `IllegalArgumentException: 'other' is different type of Path` from `Path.relativize`. The exception surfaces to the user as an XPath compile error (and a logged stack trace server-side), **blocking any `import module` from succeeding in unsaved eXide buffers**.

This adds a defensive fallback: if the load path isn't relativizable, use the absolute source collection path (matching the existing behaviour for `moduleLoadPath = "."`).

This is intentionally scoped to the server side. The primary fix lives on the eXide side -- @line-o's preference is to have eXide send `xquery.module-load-path` as `"."` for unsaved buffers rather than the synthetic xmldb URI -- and that will land separately. The change here protects current eXide releases (and any other client that sends an oddly-shaped load path) from a surfaced XPathException while the eXide-side fix rolls out.

## What Changed

- `exist-core/src/main/java/org/exist/xquery/XQueryContext.java`
  - Extracted the relativize call from `resolveInEXPathRepository` into a small package-private helper `relativizeOrFallback(String moduleLoadPath, String sourceCollection)`.
  - The helper catches `IllegalArgumentException` from `Path.relativize` and falls back to `sourceCollection`, logging at DEBUG.
- `exist-core/src/test/java/org/exist/xquery/XQueryContextTest.java`
  - Three unit tests covering: relativizable load path (happy path), synthetic `xmldb:exist://__new__1` load path (the eXide bug), empty load path.

## Reproduction (current eXide release & PR #794)

Both produce the bad value the same way:

- `eXide/src/editor.js:97-99` -- `getBasePath()` does `this.path.replace(/(^.+)\/[^\/]*$/, "$1")`. For an unsaved buffer whose `path = "__new__1"`, no `/` is present, the regex doesn't match, and the unchanged `"__new__1"` is returned.
- `eXide/src/eXide.js:865` -- `var moduleLoadPath = "xmldb:exist://" + editor.getActiveDocument().getBasePath();` -- yields `"xmldb:exist://__new__1"` on the wire.

Server-side path:
- `XmldbURI.create("xmldb:exist://__new__1").getCollectionPath()` returns a non-relativizable value.
- `Path.of(...).relativize(Path.of(sourceCollection))` throws `IllegalArgumentException: 'other' is different type of Path`.
- Pre-fix: caught by the outer handler, re-wrapped as `XPathException`; the import fails and the user sees a compile error.
- Post-fix: caught by the new inner handler, falls back to `sourceCollection`, import resolves.

## Test Plan

- [x] `mvn test -pl exist-core -Dtest='XQueryContextTest#relativizeOrFallback_*'` (3/3 pass)
- [x] `mvn test -pl exist-core -Dtest='XQueryContextTest'` (20/20 pass, no regression in existing tests)
- [ ] Manual: unsaved eXide buffer with `import module ...` against an EXPath-installed package (eXide-side fix needed for the end-to-end path; this PR just stops the import-failure)

## Related

- Reported by @line-o in eXist-db Slack #core-dev, thread of 2026-05-21 14:14 EDT
- Companion eXide-side fix already pushed to eXist-db/eXide#794 (sends `"."` for unsaved buffers via existing `isNew()` check)

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
+++ b/exist-core/src/main/java/org/exist/xquery/XQueryContext.java
@@ -616,9 +616,7 @@ public class XQueryContext implements BinaryValueManager, Context {
                 } else {
                     // NOTE(AR) set the location of the module to import relative to this module's load path
                     // - so that transient imports of the imported module will resolve correctly!
-                    final Path collectionPath = Path.of(XmldbURI.create(moduleLoadPath).getCollectionPath());
-                    final Path sourcePath = Path.of(sourceCollection);
-                    location = collectionPath.relativize(sourcePath).toString();
+                    location = relativizeOrFallback(moduleLoadPath, sourceCollection);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


duncdrum (changes requested):
please rename the three test methods flagged by codacy

reinhapa (approved):

line-o (approved):
I agree that this is a step in the right direction.

PR #6397, opened 2026-05-25
Dp minor CI fixes

duncdrum:
add more IDE config folders to gitignore …

If anybody uses moderne MCP with GitHub copilot we can unignore the instructions.

limit push actions on CI from running in forks

remove offline mode in license check, so new jars are downloadable from maven central for inspection.


duncdrum:
for the license check we could go fancier, by running online for commits affecting poms, and offline for everything else. Not sure if its worth the added complexity though

dizzzz (approved):

reinhapa (approved):

PR #6400, opened 2026-05-27
[feature] Add JMX MBeans and profiler support for vectors

duncdrum:
Expose VectorStore and VectorEmbedding MBeans for Monex/status monitoring,
including model diagnostics shared with `vector:diagnostics()`. Instrument
embed/KNN workloads via VectorOperationMetrics, add the vector JMX category
to JMXtoXML, and report lucene-vector profiler optimization-level `NONE` when
no vector-field is configured.

see eXist-db/monex#398

### DIFF
--- a/extensions/vector/src/main/java/org/exist/xquery/modules/vector/Diagnostics.java
+++ b/extensions/vector/src/main/java/org/exist/xquery/modules/vector/Diagnostics.java
@@ -49,6 +49,7 @@ import static org.exist.xquery.modules.vector.VectorModule.functionSignature;
  *   <li>path — resolved configuration path (relative string)</li>
  *   <li>dimension — configured or default dimension</li>
  *   <li>status — "available" | "missing" | "http"</li>
+ *   <li>provider — "ONNX" | "HTTP"</li>
  *   <li>message — optional hint when status != "available"</li>
  * </ul>
  */
@@ -79,6 +80,7 @@ public class Diagnostics extends BasicFunction {
       builder.addAttribute(new QName("path", null, null), model.getPath());
       builder.addAttribute(new QName("dimension", null, null), Integer.toString(model.getDimension()));
       builder.addAttribute(new QName("status", null, null), model.getStatus());
+      builder.addAttribute(new QName("provider", null, null), model.getProvider());
       final String message = model.getMessage();
       if (message != null && !message.isEmpty()) {
         builder.addAttribute(new QName("message", null, null), message);
--- a/extensions/vector/src/main/java/org/exist/xquery/modules/vector/Diagnostics.java
+++ b/extensions/vector/src/main/java/org/exist/xquery/modules/vector/Diagnostics.java
@@ -22,6 +22,8 @@
 package org.exist.xquery.modules.vector;
 
 import org.exist.dom.QName;
+import org.exist.vector.VectorModelDiagnostics;
+import org.exist.vector.VectorModelInfo;
 import org.exist.xquery.BasicFunction;
 import org.exist.xquery.Cardinality;
 import org.exist.xquery.FunctionSignature;
@@ -31,16 +33,9 @@ import org.exist.xquery.value.NodeValue;
 import org.exist.xquery.value.Sequence;
 import org.exist.xquery.value.Type;
 import org.exist.xquery.value.ValueSequence;
-import org.exist.xquery.value.StringValue;
-import org.exist.xquery.value.IntegerValue;
-import org.exist.vector.ModelPathResolver;
-import org.exist.vector.ModelRegistry;
-import org.exist.vector.VectorModelConstants;
 
 import javax.annotation.Nullable;
-import java.nio.file.Path;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.List;
 
 import static org.exist.xquery.modules.vector.VectorModule.functionSignature;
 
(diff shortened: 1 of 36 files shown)


line-o (approved):
LGTM

extensions/vector/src/main/java/org/exist/xquery/modules/vector/Diagnostics.java
```diff
@@ -49,6 +49,7 @@
  *   <li>path — resolved configuration path (relative string)</li>
  *   <li>dimension — configured or default dimension</li>
  *   <li>status — "available" | "missing" | "http"</li>
+ *   <li>provider — "ONNX" | "HTTP"</li>
```
joewiz: The other attribute values appear to all be lowercase; should these stay uppercase or be changed to match the others?

joewiz (approved):
LGTM. And FWIW, here's Claude's review. In reading it over, item 5 made me take another look at Monex and see something I'd missed when looking at the Monitoring pane: 

> Image

Claude's full review: 

## PR #6400 review notes

The PR is well-scoped and CI is green (XQTS pass, all integration suites on Linux/macOS/Windows pass, Codacy clean, **APPROVED**, MERGEABLE). The design pattern — reflective hook from exist-core into the optional extension, per-instance MBean registration, a small `Recorder` bridge so core code doesn't take a hard dep on the extension — is sound. Most of my reading is positive. A few minor things worth at least mentioning (1-4 are latent, 5-6 are UX):

1. **Embed metrics record success only.** `Embed.eval()` (line ~102) and `EmbedBatch` (line ~114) call `VectorOperationMetrics.recordEmbed` *after* `provider.embed()` returns. If embedding throws, the call doesn't count. So `EmbedCallCount` under-reports during outages — which is exactly when you'd want to see it spike. A try/finally would close that.

2. **`VectorMetrics` is JVM-global; the MBean is per-instance.** `VectorMetrics.INSTANCE` is a process-singleton, but `VectorEmbeddingJmx.registerIfAbsent` keys on `pool.getId()`. If two BrokerPools run in one JVM (rare in prod, common in tests), every instance's `VectorEmbedding` MBean reports the same aggregated counters. Contrast with `VectorStore`, which correctly carries its own `VectorStoreImpl`. Worth either making `VectorMetrics` per-instance or documenting the intent.

3. **`REGISTERED` set never cleared on pool shutdown.** `VectorEmbeddingJmx` adds the instance id once and never removes it. In an in-process pool restart cycle (testing, embedded mode), the second startup skips registration silently. `AgentFactory.addMBean` usually overwrites, so it tends to work, but it's a brittle assumption.

4. **60s cached model snapshot.** `VectorModelDiagnostics` caches `collectModels()` for 60s with no eviction on filesystem change. If you drop a model directory in while a JMX poller is hitting it, the dashboard lies for up to a minute. Acceptable trade-off, but `refreshModels()` should probably be exposed as a JMX operation alongside `resetEntryCountCache()` for symmetry.

5. **9-of-10 "missing" rows on a healthy install.** The builtin known-models catalog (OpenAI text-embedding-3-*, Cohere embed-*, etc.) all surface with `status="missing"` and a warning message until someone explicitly drops them in. Monex headlines "Embeddings: 1 / 10", which reads as "9 broken" rather than "9 you haven't configured". Either splitting `source="builtin"` into `catalog` (advertised, not installed) vs `builtin` (shipped) — or having monex filter `catalog` from the deficit count — would be friendlier. This may be more a monex fix than a 6400 fix.

6. **Two "PersistenceBackend" fields with different meanings.** `VectorStore.PersistenceBackend = "vector.dbx"` (the file), `VectorEmbedding.PersistenceBackend = "lucene"` (the KNN index). Same name, different semantics across two MBeans. Not wrong, but a JMX consumer that grabs by attribute name without checking the MBean type could get confused.

### Net

None of these are blockers. The PR does what it advertises, has tests (`VectorStoreImplTest`, `VectorEmbeddingJmxTest`, `VectorMetricsTest`, `VectorModelDiagnosticsTest`, `VectorOperationMetricsTest`, plus the `JmxRemoteTest` extension), and the failure modes are graceful (extension-not-on-classpath logs at DEBUG and proceeds). Comfortable saying it's ready to merge; items 1, 2, and 5 are worth a quick follow-up issue.

extensions/vector/src/main/java/org/exist/xquery/modules/vector/Diagnostics.java
dizzzz: it is an abbrev right?

dizzzz (approved):
LGTM

extensions/vector/src/main/java/org/exist/xquery/modules/vector/Diagnostics.java
duncdrum: [ONNX](https://onnxruntime.ai) is an acronym (I guess) so conventionally all caps. It's how I used it throughout the codebase and docs. That made the other provider all caps by proxy.  Unintentional bonus you get to distinguish between http availability as status, vs model registry source in logs and monex. The trade-off seems worth it.

duncdrum:
> LGTM. And FWIW, here's Claude's review. In reading it over, item 5 made me take another look at Monex and see something I'd missed when looking at the Monitoring pane:
> 
> > Image
> 
> Claude's full review:
> 
> ## PR #6400 review notes
> 1. **Embed metrics record success only.** `Embed.eval()` (line ~102) and `EmbedBatch` (line ~114) call `VectorOperationMetrics.recordEmbed` _after_ `provider.embed()` returns. If embedding throws, the call doesn't count. So `EmbedCallCount` under-reports during outages — which is exactly when you'd want to see it spike. A try/finally would close that.

Good catch will push another commit

> 2. **`VectorMetrics` is JVM-global; the MBean is per-instance.** `VectorMetrics.INSTANCE` is a process-singleton, but `VectorEmbeddingJmx.registerIfAbsent` keys on `pool.getId()`. If two BrokerPools run in one JVM (rare in prod, common in tests), every instance's `VectorEmbedding` MBean reports the same aggregated counters. Contrast with `VectorStore`, which correctly carries its own `VectorStoreImpl`. Worth either making `VectorMetrics` per-instance or documenting the intent.

Will give it a try, anything that helps reduce test flake is worth a shot. 

> 3. **`REGISTERED` set never cleared on pool shutdown.** `VectorEmbeddingJmx` adds the instance id once and never removes it. In an in-process pool restart cycle (testing, embedded mode), the second startup skips registration silently. `AgentFactory.addMBean` usually overwrites, so it tends to work, but it's a brittle assumption.

While this won't bring us hot-reloading, it's a good idea.  

> 4. **60s cached model snapshot.** `VectorModelDiagnostics` caches `collectModels()` for 60s with no eviction on filesystem change. If you drop a model directory in while a JMX poller is hitting it, the dashboard lies for up to a minute. Acceptable trade-off, but `refreshModels()` should probably be exposed as a JMX operation alongside `resetEntryCountCache()` for symmetry.

`refreshModels()` on the MBean landed in the hardening commit.  Where do you think it is missing exactly?

> 5. **9-of-10 "missing" rows on a healthy install.** The builtin known-models catalog (OpenAI text-embedding-3-_, Cohere embed-_, etc.) all surface with `status="missing"` and a warning message until someone explicitly drops them in. Monex headlines "Embeddings: 1 / 10", which reads as "9 broken" rather than "9 you haven't configured". Either splitting `source="builtin"` into `catalog` (advertised, not installed) vs `builtin` (shipped) — or having monex filter `catalog` from the deficit count — would be friendlier. This may be more a monex fix than a 6400 fix.

Follow up for monex incoming. 

> 6. **Two "PersistenceBackend" fields with different meanings.** `VectorStore.PersistenceBackend = "vector.dbx"` (the file), `VectorEmbedding.PersistenceBackend = "lucene"` (the KNN index). Same name, different semantics across two MBeans. Not wrong, but a JMX consumer that grabs by attribute name without checking the MBean type could get confused.
> 

Good point will see what I can do, might have to defer, that one if it grows out of proportion.

PR #6405, opened 2026-05-29
[doc][test] Clarify ft:query-vector result ordering + pin down with tests

joewiz:

## Summary

`ft:query-vector`'s function signature has claimed since #6400 that it returns *"Nodes matching the vector query, ordered by similarity."* The first half is correct; the second half is not. `ft:query-vector` returns an XQuery node sequence, so the result follows the [document-order](https://www.w3.org/TR/xquery-31/#dt-document-order) rules of [XQuery 3.1 §2.4](https://www.w3.org/TR/xquery-31/#id-sequence-types): within a single document, nodes are returned in document order; the relative order of nodes from distinct trees is implementation-dependent (but stable) unless an explicit ordering key is supplied. To rank by similarity, callers must use `ft:score` together with an explicit [`order by`](https://www.w3.org/TR/xquery-31/#id-orderby-clause) clause.

This PR makes that explicit in two places:

1. The three `ft:query-vector` signatures now describe the default return shape and alude to the `ft:score` + `order by` idiom in both the per-signature prose and the return-type description. The companion `ft:query-field-vector` (in `QueryFieldVector.java`) already used neutral wording; this PR aligns `ft:query-vector` with it.

2. The XQSuite vector tests gain a "Result ordering" section that pins the behavior down so it can't be lost to a future refactor. Two existing tests that happened to pass without sorting are also tightened.

### `extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryVector.java`

- Extracted a shared `RETURN_DESC` constant for the three signatures.
- Replaced *"Nodes matching the vector query, ordered by similarity."* with a description that names XQuery node-sequence semantics and shows the `for $h in ft:query-vector(...) order by ft:score($h) descending return $h` idiom.
- Lightly clarified the per-signature description (no behavior change).

### `extensions/indexes/lucene/src/test/xquery/lucene/vector-search.xqm`

**Strengthened existing tests** (commit 2 of 3):

- `vs:euclidean-score-ordering` and `vs:dot-product-score-ordering` previously asserted `ft:score($hits[1]) ge ft:score($hits[2])` on a 2-hit set. With `k=2` and two source nodes, the assertion happened to pass because the higher-scoring hit landed first in node-sequence order by coincidence of the particular vectors — not because the test imposed any ordering key. They now sort explicitly with `for $h in ... order by ft:score($h) descending` so the assertion is about score-based ranking rather than node-sequence accident.

**Added six new tests** (commit 3 of 3):

Test data: three articles with embeddings chosen so that storage order ≠ similarity order for query `q=[1,0,0,0]`:

| Title | Embedding         | cosine(., q) |
|-------|-------------------|--------------|
| A     | `[0, 0, 1, 0]`    | 0.000        |
| B     | `[1, 0, 0, 0]`    | 1.000        |
| C     | `[0.9, 0.1, 0, 0]`| ~0.994       |

So document order is A, B, C and similarity order is B, C, A.

Two collections cover the two shapes the data model distinguishes:

- `COLLECTION_ORDER_SINGLE` — 3 articles in one document.
- `COLLECTION_ORDER_MULTI`  — 3 articles, one per document (`a.xml`, `b.xml`, `c.xml`).

Six new XQSuite tests:

| Test                                                                  | Asserts            |
|-----------------------------------------------------------------------|--------------------|
| `vs:order-single-doc-default-is-document-order`                       | `("A", "B", "C")`  |
| `vs:order-single-doc-sort-by-score-is-similarity-order`               | `("B", "C", "A")`  |
| `vs:order-multi-doc-default-contains-all-three`                       | `("A", "B", "C")` (sorted alphabetically) |
| `vs:order-multi-doc-sort-by-score-is-similarity-order`                | `("B", "C", "A")`  |
| `vs:order-field-vector-single-doc-default-is-document-order`          | `("A", "B", "C")`  |
| `vs:order-field-vector-single-doc-sort-by-score-is-similarity-order`  | `("B", "C", "A")`  |

The multi-document "default" test deliberately re-sorts the titles alphabetically before asserting, so it does *not* depend on the implementation-defined inter-document ordering — it only proves the result set is complete. Its sibling "sort by score" test pins down the ranking.

## Why now

Surfaced while testing monex PR #398 against develop after #6400 merged. Three semantically distinct prompts against three short articles all returned hits in storage order; assumed the function was bypassing the index, then realised after consulting the docs (`~/workspace/documentation/src/main/xar-resources/data/lucene/lucene.xml`) and re-checking JMX (`VectorStore.EntryCount` ≥ 3, `KnnQueryCount` increments) that cosine similarity is fully implemented and correctly scored — the score order just isn't preserved in the returned sequence. Easy to miss; worth pinning down.

## Spec references

- [XQuery 3.1 §2.4 Document Order](https://www.w3.org/TR/xquery-31/#dt-document-order): "The relative order of nodes in distinct trees is implementation-dependent but stable..."
- [XQuery 3.1 §3.12.7 Order By and Return Clauses](https://www.w3.org/TR/xquery-31/#id-orderby-clause)

## Test plan

- [x] `mvn test -pl extensions/indexes/lucene -Dtest=VectorSearchTests -DfailIfNoTests=false -Denforcer.skip=true -Ddependency-check.skip=true -Ddocker=false` → 73 tests, 0 failures, 0 errors, 6 skipped
- [ ] Full unit test suite (`mvn test -pl extensions/indexes/lucene`) on CI
- [x] Manual smoke against a live container (`existdb-develop` on localhost:8081) with all three contrastive prompts confirming the documented sort idiom works

## Follow-up (separate PR against `documentation`)

The lucene docs (`src/main/xar-resources/data/lucene/lucene.xml`, line 729-731) carry the softer claim *"The functions return the k most similar nodes, where a higher score means more similar."* It should likewise be extended to spell out the `ft:score` + `order by` step. Tracking separately so this PR stays scoped to the `eXist-db/exist` repo.

### DIFF
--- a/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryVector.java
+++ b/extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryVector.java
@@ -58,21 +58,27 @@ public class QueryVector extends BasicFunction {
     private static final FunctionParameterSequenceType FS_PARAM_OPTIONS = optParam("options", Type.ITEM,
             "Optional map with filter-query, filter, facets.");
 
+    private static final String RETURN_DESC =
+            "Nodes matching the vector query. The returned sequence follows XQuery node-sequence semantics; "
+                    + "to rank by similarity, sort explicitly with ft:score, e.g. "
+                    + "'for $h in ft:query-vector(...) order by ft:score($h) descending return $h'.";
+
     final static FunctionSignature[] signatures = {
             functionSignature("query-vector",
-                    "KNN vector search. Returns k nearest nodes. Vector field is resolved from index config.",
-                    returnsOptMany(Type.NODE, "Nodes matching the vector query, ordered by similarity."),
+                    "KNN vector search. Returns the k nodes nearest to the query vector. "
+                            + "Vector field is resolved from index config.",
+                    returnsOptMany(Type.NODE, RETURN_DESC),
                     FS_PARAM_NODES,
                     FS_PARAM_VECTOR),
             functionSignature("query-vector",
                     "KNN vector search with explicit k.",
-                    returnsOptMany(Type.NODE, "Nodes matching the vector query, ordered by similarity."),
+                    returnsOptMany(Type.NODE, RETURN_DESC),
                     FS_PARAM_NODES,
                     FS_PARAM_VECTOR,
                     FS_PARAM_K),
             functionSignature("query-vector",
                     "KNN vector search with k and options (filter-query, filter, facets).",
-                    returnsOptMany(Type.NODE, "Nodes matching the vector query, ordered by similarity."),
+                    returnsOptMany(Type.NODE, RETURN_DESC),
                     FS_PARAM_NODES,
                     FS_PARAM_VECTOR,
                     FS_PARAM_K,
--- a/extensions/indexes/lucene/src/test/xquery/lucene/vector-search.xqm
+++ b/extensions/indexes/lucene/src/test/xquery/lucene/vector-search.xqm
@@ -902,19 +902,28 @@ function vs:dot-product-indexed-and-queried() {
     count(collection($vs:COLLECTION_DOT_PRODUCT)//article[ft:query-field-vector("emb_dot_product", [1.0, 0.0, 0.0, 0.0], 2)])
 };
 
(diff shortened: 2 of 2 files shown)


duncdrum (approved):
Nice find, and very quick fix

line-o (commented):
👍🏼

PR #6419, opened 2026-05-31
[bugfix] Track item type in AVLTreeNodeSet for correct return-type checks

### LINKED ISSUE #4255: [BUG] intersect expression triggers error on persistent elements
**Describe the bug**

When the `intersect` expression is called on elements stored in the database, eXist throws the following unexpected error on functions whose signatures indicate their return type must be `element()`:

> Return type of function *name_of_function*. exerr:ERROR Type error: expected type: `element()`; got: `node()`

I believe the error is related to this line in the eXist source code: https://github.com/eXist-db/exist/blob/5bc6a35fd6d4a91742637aefe4ad53c2324284ee/exist-core/src/main/java/org/exist/xquery/Intersect.java#L48

**Expected behavior**

When called on elements, the `intersect` expression should not cause functions that expect to return an element to throw an error.

**To Reproduce**

The following xqsuite tests the effect of `intersect` on nodes stored in the database and in memory. It shows that a bug affects `intersect` when used on nodes stored in the database, when functions declare that they return `element()`. The bug is not exhibited with affect in-memory nodes. 

```xquery
xquery version "3.1";

    <root>
        <x/>
        <y/>
        <z/>
    </root>
};

declare
    %test:setUp
function t:setup() {
    let $testCol := xmldb:create-collection("/db", "test")
    return
        xmldb:store($testCol, "test.xml", $t:XML)
};

declare
    %test:tearDown
function t:tearDown() {
    xmldb:remove("/db/test")
};

declare
    %test:assertEquals("<y/>")
function t:intersect-no-return-type-db() {
    let $root := doc("/db/test/test.xml")/root
    let $seq1 := ($root/x, $root/y)
    let $seq2 := ($root/y, $root/z)
    let $intersect := $seq1 intersect $seq2
    return
        $intersect
};

declare
    %test:assertEquals("<y/>")
function t:intersect-element-return-type-db() as element() {
    let $root := doc("/db/test/test.xml")/root
    let $seq1 := ($root/x, $root/y)
    let $seq2 := ($root/y, $root/z)
    let $intersect := $seq1 intersect $seq2
    return
        $intersect
};

declare
    %test:assertEquals("<y/>")
function t:intersect-no-return-type-mem() {
    let $root := $t:XML/root
    let $seq1 := ($root/x, $root/y)
    let $seq2 := ($root/y, $root/z)
    let $intersect := $seq1 intersect $seq2
    return
        $intersect
};

declare
    %test:assertEquals("<y/>")
function t:intersect-element-return-type-mem() as element() {
    let $root := $t:XML/root
    let $seq1 := ($root/x, $root/y)
    let $seq2 := ($root/y, $root/z)
    let $intersect := $seq1 intersect $seq2
    return
        $intersect
};
```

This xqsuite returns the following error:

```xml
<testsuite package="http://exist-db.org/xquery/test" timestamp="2022-02-26T01:05:11.139-05:00"
    tests="4" failures="0" errors="1" pending="0" time="PT0.002S">
    <testcase name="intersect-element-return-type-db" class="t:intersect-element-return-type-db">
        <error type="err:XPTY0004"
            message="It is a type error if, during the static analysis phase, an expression is found to have a static type that is not appropriate for the context in which the expression occurs, or during the dynamic evaluation phase, the dynamic type of a value does not match a required type as specified by the matching rules in 2.5.4 SequenceType Matching. Return type of function 't:intersect-element-return-type-db'. exerr:ERROR Type error: expected type: element(); got: node()"
        />
    </testcase>
    <testcase name="intersect-element-return-type-mem" class="t:intersect-element-return-type-mem"/>
    <testcase name="intersect-no-return-type-db" class="t:intersect-no-return-type-db"/>
    <testcase name="intersect-no-return-type-mem" class="t:intersect-no-return-type-mem"/>
</testsuite>
```

In my tests, the other constructor expressions—`union` and `except`—do not have this problem.

**Screenshots**

n/a

 - OS: macOS 12.2.1

**Additional context**

joewiz:

**Closes #4255**

## Summary

A user-defined function declared `as element()` returning the result of a persistent `intersect` failed with `err:XPTY0004 expected type: element(); got: node()` even when every item in the result was in fact an element. The bug also affected `except` (same code path; never reported separately). `union` was unaffected because it goes through a different result-set implementation that already tracks item types.

## Root cause

| `NodeSet` operation | Result type | Tracks `itemType`? |
|---|---|---|
| `union()`     | `NewArrayNodeSet` (extends `AbstractArrayNodeSet`) | **yes** — `checkItemType()` on every `add()` |
| `intersection()` | `AVLTreeNodeSet` (extends `AbstractNodeSet`) | **no** — inherited `getItemType()` returns `Type.NODE` unconditionally |
| `except()`    | `AVLTreeNodeSet`                                  | **no** — same as `intersection()` |

`FunctionCall.eval` (line 216) performs a sequence-level return-type check via `SequenceType.checkType(result.getItemType())`. For an `AVLTreeNodeSet` result, `getItemType()` returned the supertype `Type.NODE = 56`, which is *not* a sub-type of the declared `Type.ELEMENT = 60`, so the check threw `XPTY0004`.

## Fix

Give `AVLTreeNodeSet` per-add item-type tracking, mirroring `AbstractArrayNodeSet.checkItemType`:

- New `itemType` field starts at `Type.ANY_TYPE`
- `add(NodeProxy)` calls `checkItemType(proxy.getType())`:
  - `ANY_TYPE` → first item's type
  - same type → no-op
  - different type → fall back to `Type.NODE`
- New `getItemType()` override returns the tracked type (or `Type.NODE` when nothing has been added)

This is the same shape `AbstractArrayNodeSet` already uses, so the behavior aligns across all `NodeSet` implementations.

## What changed

- `exist-core/src/main/java/org/exist/dom/persistent/AVLTreeNodeSet.java`
  - +1 import (`Type`)
  - +1 field (`itemType`)
  - +1 call in `add()`
  - +1 private helper `checkItemType`
  - +1 `@Override getItemType()`
- `exist-core/src/test/java/org/exist/xquery/IntersectElementReturnTypeTest.java` (new)
  - 4 tests: persistent + in-memory intersect, persistent union (sanity), persistent except (regression for the latent twin bug)

## Test plan

- [x] `mvn test -pl exist-core -Dtest=IntersectElementReturnTypeTest` — 4/4 pass
- [x] `mvn test -pl exist-core -Dtest='XQuery3Tests,XPathQueryTest,IntersectElementReturnTypeTest'` — 1184/1184 pass, 5 skipped (pre-existing), 0 regressions

## Spec reference

- [W3C XPath 3.1 §3.16 — Combining Node Sequences](https://www.w3.org/TR/xpath-31/#combining_seq) — `intersect` returns nodes from both operand sequences; the result's *item type* must be the type common to the operand sequences.

### DIFF
--- a/exist-core/src/main/java/org/exist/dom/persistent/AVLTreeNodeSet.java
+++ b/exist-core/src/main/java/org/exist/dom/persistent/AVLTreeNodeSet.java
@@ -24,6 +24,7 @@ package org.exist.dom.persistent;
 import org.exist.numbering.NodeId;
 import org.exist.xquery.value.Item;
 import org.exist.xquery.value.SequenceIterator;
+import org.exist.xquery.value.Type;
 
 import javax.annotation.Nullable;
 import java.util.ArrayDeque;
@@ -35,6 +36,7 @@ public class AVLTreeNodeSet extends AbstractNodeSet {
     private Node root;
     private int size = 0;
     private int state = 0;
+    private int itemType = Type.ANY_TYPE;
 
     @Override
     public SequenceIterator iterate() {
@@ -61,6 +63,31 @@ public class AVLTreeNodeSet extends AbstractNodeSet {
     @Override
     public long getItemCountLong() {
         return size;
+    }
+
+    /**
+     * Refine the tracked itemType as nodes are added, mirroring the pattern
+     * used by {@link AbstractArrayNodeSet#checkItemType(int)}. The result is
+     * exposed via {@link #getItemType()} so that sequence-level type checks
+     * (function return types, variable declarations, instance-of, ...) see
+     * the tightest common supertype of the contained items rather than the
+     * unconditional {@link Type#NODE} returned by the {@link AbstractNodeSet}
+     * default.
+     */
+    private void checkItemType(final int type) {
+        if (itemType == Type.NODE || itemType == type) {
+            return;
+        }
+        if (itemType == Type.ANY_TYPE) {
+            itemType = type;
+        } else {
+            itemType = Type.NODE;
+        }
+    }
+
+    @Override
+    public int getItemType() {
+        return itemType == Type.ANY_TYPE ? Type.NODE : itemType;
     }
 
 
@@ -110,6 +137,7 @@ public class AVLTreeNodeSet extends AbstractNodeSet {
             return;
         }
 
+        checkItemType(proxy.getType());
         setHasChanged();
         if(root == null) {
             root = new Node(proxy);
--- a/exist-core/src/test/java/org/exist/xquery/IntersectElementReturnTypeTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/IntersectElementReturnTypeTest.java
@@ -1,0 +1,123 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


duncdrum (approved):
Lgtm

reinhapa (approved):

PR #6423, opened 2026-05-31
[bugfix] Surface %test:tearDown errors in xqsuite output

### LINKED ISSUE #6422: XQSuite silently swallows %test:tearDown errors

## Summary

XQSuite silently swallows errors thrown from `%test:tearDown` functions, so failed teardowns are invisible to any consumer of the test XML (CI, IDE, reports).

## Where

`exist-core/src/main/xquery/lib/xqsuite/xqsuite.xql`, in `xqsuite:call-test`, around lines 138-141 (master at the time of writing):

```xquery
let $tearDown :=
    if (exists($tearDown-call)) then
        try { $tearDown-call() } catch * { () }
    else ()
return
    ($result, $tearDown)[1]
```

Two things happen here:

1. The `catch *` discards `$err:code`, `$err:description`, etc. — there's no record that tearDown threw.
2. `($result, $tearDown)[1]` returns whichever sequence the test function produced first; tearDown's return value is implicitly thrown away even on the success path (which is correct), but combined with (1) it means no caller can ever distinguish "tearDown ran cleanly" from "tearDown blew up".

## Repro

```xquery
xquery version "3.1";
module namespace t = "http://example.com/teardown-fail";
declare namespace test = "http://exist-db.org/xquery/xqsuite";

declare %test:tearDown function t:cleanup() {
    error(xs:QName("t:teardown-boom"), "tearDown blew up")
};

declare %test:assertEquals(1) function t:passes() { 1 };
```

Run via `xqsuite:suite(inspect:module-functions(...))`. The resulting `<testsuites>` reports 1 test, 0 failures, 0 errors — the tearDown error is gone.

For contrast, `%test:setUp` failures *are* surfaced (XQSuite emits `<testsuite errors="N">message</testsuite>` with no `<testcase>` children).

## Impact

- IDEs (eXide's Run-as-test, third-party harnesses) can't show "tests passed but teardown left state behind" — a real failure mode for tests that rely on tearDown to drop temp collections, reset config, etc.
- CI pipelines consuming the JUnit-shaped XML have no signal that cleanup failed; subsequent test runs may then fail mysteriously due to leftover state.
- Discovered while working on eXide #805 (improving Run-as-test): I tried to mirror the setUp-failure panel for tearDown failures but found there was no hook to attach to.

## Suggested fix

Capture the tearDown error and surface it on the testsuite/testcase. The exact shape is up for discussion; one option that preserves backwards compatibility:

```xquery
let $tearDown :=
    if (exists($tearDown-call)) then
        try { $tearDown-call(), () } catch * {
            <tearDownError type="{$err:code}" message="{$err:description}">{
                $err:value
            }</tearDownError>
        }
    else ()
return
    if ($tearDown instance of element(tearDownError)) then
        (: append to whichever <testcase> we just ran, or to the testsuite :)
        ...
    else
        $result
```

Happy to take a swing at the implementation if there's agreement on the shape — flagging here first since this is XQSuite-wide and the output proposed here probably merits a quick review from @duncdrum / @line-o / @reinhapa / @dizzzz before any code lands.

## eXide context

Sibling work in eXide #805 (PR pending) ships the setUp-failure panel and other Run-as-test improvements; the tearDown panel is the one piece I couldn't implement without changes here.

joewiz:

**Closes #6422** (verified independently with a before/after JUnit run on develop HEAD `b990cd6f7c`)

## Summary

XQSuite caught errors thrown from `%test:tearDown` functions in `test:call-func-with-annotation` and emitted a `<system-err>` element — but the outer `($result, $tearDown)[1]` selector at the end of `test:suite` then discarded that element. Consumers of the suite XML (CI, IDE Run-as-test panels, JUnit-shaped reporters) saw no trace of the tearDown failure. The `%test:setUp` counterpart was already surfaced via the testsuite text content, but the tearDown path was a silent dead end.

@dizzzz +1'd the issue, so opening this PR. Reviewers — see the note at the bottom on the chosen output shape; the issue discussion left the wire format open and I'm happy to adjust based on review.

## Root cause

In `xqsuite.xql`, around line 137 of develop:

```xquery
return (
    $result,
    test:call-func-with-annotation($modFunctions, "tearDown", $test-error-function)
)[1]
```

`call-func-with-annotation` returns `<ok/>` on success and `<system-err>{$err:description}</system-err>` on failure. For setUp the `<system-err>` was already flowing into the testsuite's text content. For tearDown the same `<system-err>` was discarded by `[1]`.

## Fix

`exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql`:

- Compute the tearDown result before constructing `<testsuite>`, so its `<system-err/>` can be folded in alongside the test results.
- When tearDown threw, append a `<system-err>tearDown error: …</system-err>` child to the `<testsuite>` and bump the `errors` count by one. Element name matches the JUnit convention so existing consumers handle it naturally.
- On the setUp-failure path (which already short-circuits the test run), append the same `"tearDown error: …"` text to the existing error message and count the tearDown error toward `errors` too.
- Clean tearDowns continue to produce no marker — output is byte-identical to before the fix, so passing test files are unchanged.

## What's tested

New `exist-core/src/test/java/org/exist/xquery/xqsuite/XQSuiteTearDownErrorTest.java` with three cases:

- `cleanTearDownProducesNoMarker` — backwards-compatible: clean tearDowns emit no marker, `errors="0"`
- `throwingTearDownIsSurfaced` — passing test + throwing tearDown → `errors="1"` plus a `<system-err>tearDown error: …</system-err>` element under the testsuite
- `throwingSetUpAndTearDownBothSurfaced` — both setUp and tearDown throw → both messages appear in the testsuite text

Each test stores a small module to `/db/test-6422`, runs `test:suite(inspect:module-functions(...))` against it, serializes the result, and asserts on the XML.

## Before / after

Verified the assertions actually exercise the change by temporarily reverting `xqsuite.xql` to develop, rebuilding, and running the same tests — 2 of 3 fail with the bug's signature (`errors="0"`, no system-err element). Restoring the fix → 3/3 pass.

```
(without fix, develop)
[ERROR] throwingTearDownIsSurfaced -- expected <system-err>tearDown error: ...</system-err>,
        got: <testsuites><testsuite ... tests="1" failures="0" errors="0" pending="0"
                                        time="PT0.1S">
               <testcase name="passes" ... /></testsuite></testsuites>

(with fix)
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
```

## Test plan

- [x] `mvn test -pl exist-core -Dtest=XQSuiteTearDownErrorTest` — 3/3 pass
- [x] `mvn test -pl exist-core -Dtest='XQSuiteTests,XQuery3Tests,XQSuiteTearDownErrorTest'` — 1124/1124 pass, 1 skipped (pre-existing), 0 regressions
- [x] Confirmed the regression test fails on develop without the fix (see Before/after above)
- [x] Codacy PMD on the new JUnit — clean

## A note on shape — open to feedback

I considered three output shapes for the tearDown failure:

1. **`<system-err>tearDown error: …</system-err>`** child of the `<testsuite>` (this PR). JUnit-idiomatic, no test-count inflation, no new element name to teach consumers.
2. A synthetic `<testcase name="tearDown">` with an `<error/>` child. Cleanest for JUnit consumers that show per-test panels, but adds a "test" that wasn't in the source and bumps `tests`.
3. A custom `<tearDownError/>` element. Clear intent but a new element name for every consumer.

I picked (1) for the smallest surface. If reviewers prefer (2) or (3), or want the `errors` attribute counted differently, I'll happily reshape — let me know.

cc @duncdrum @line-o @reinhapa @dizzzz (per the issue discussion).

### DIFF
--- a/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
+++ b/exist-core/src/main/resources/org/exist/xquery/lib/xqsuite/xqsuite.xql
@@ -113,30 +113,50 @@ declare function test:suite(
                     namespace-uri-from-QName(function-name($func)) = $module
                 })
             let $setup := test:call-func-with-annotation($modFunctions, "setUp", $test-error-function)
-            let $result :=
-                if (empty($setup) or $setup/self::ok) then
-                    let $startTime := util:system-time()
-                    let $results := test:function-by-annotation($modFunctions, "assert", $runner)
-                    let $elapsed :=
-                        util:system-time() - $startTime
-                    return
-                        <testsuite package="{$module}" timestamp="{util:system-dateTime()}"
-                            tests="{count($results)}"
-                            failures="{count($results/failure)}"
-                            errors="{count($results/error)}"
-                            pending="{count($results/pending)}"
-                            time="{$elapsed}">
-                            { $results }
-                        </testsuite>
+            let $setupOk := empty($setup) or exists($setup/self::ok)
+            let $startTime := util:system-time()
+            let $results :=
+                if ($setupOk) then
+                    test:function-by-annotation($modFunctions, "assert", $runner)
+                else ()
+            let $elapsed := util:system-time() - $startTime
+            (:
+             : tearDown always runs, regardless of whether setUp or any test
+             : threw. Prior to https://github.com/eXist-db/exist/issues/6422
+             : a tearDown error was caught in test:call-func-with-annotation
+             : and the resulting <system-err/> was then discarded by the
+             : outer ($result, $tearDown)[1] selector, leaving the bug
+             : invisible to any consumer of the suite XML (CI, IDE, reports).
+             : Surface it as a <system-err> child on the <testsuite>, and
+             : count it as one additional error.
+             :)
+            let $tearDown := test:call-func-with-annotation($modFunctions, "tearDown", $test-error-function)
+            let $tearDownErr := $tearDown/self::system-err
+            return
+                if ($setupOk) then
+                    <testsuite package="{$module}" timestamp="{util:system-dateTime()}"
+                        tests="{count($results)}"
+                        failures="{count($results/failure)}"
+                        errors="{count($results/error) + count($tearDownErr)}"
+                        pending="{count($results/pending)}"
+                        time="{$elapsed}">
+                        { $results }
+                        {
+                            if ($tearDownErr) then
+                                <system-err>{ "tearDown error: " || $tearDownErr/string() }</system-err>
+                            else ()
+                        }
+                    </testsuite>
                 else
                     <testsuite package="{$module}" timestamp="{util:system-dateTime()}"
-                        errors="{count($functions)}">
-                        {$setup/string()}
+                        errors="{count($functions) + count($tearDownErr)}">
+                        {
+                            if ($tearDownErr) then
+                                $setup/string() || "&#10;tearDown error: " || $tearDownErr/string()
+                            else
+                                $setup/string()
+                        }
                     </testsuite>
-            return (
-                $result,
-                test:call-func-with-annotation($modFunctions, "tearDown", $test-error-function)
-            )[1]
         }
         </testsuites>
 };
--- a/exist-core/src/test/java/org/exist/xquery/xqsuite/XQSuiteTearDownErrorTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/xqsuite/XQSuiteTearDownErrorTest.java
@@ -1,0 +1,177 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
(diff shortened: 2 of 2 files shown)


dizzzz:
Thnx for the PR! Hiding issues is not a good thing to do.

Will option (1) make the issue visible in a way? e.g. when running junit report plugin?

duncdrum (approved):

reinhapa (approved):

line-o:
According to [a version of the JUnit XML Schema](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd#L160) the system-err element is a child of testsuite and 

> Data that was written to standard error while the test was executed

I think this is OK to use as a container as any error that is otherwise not caught would be piped to stderr as well.

line-o:
We should think of utilizing `<system-out>` and `<system-err>` even more. I see the potential of better debuggability with the calls to `util:system-out` and `util:system-err` being shown directly in the testreport.

Issue #6176, opened 2026-03-25
Adapt publication to Maven Central


line-o:
The current setup to publish new releases to Maven Central is using a legacy system and also requires manual steps.

It is therefore important to switch to the current method as described on https://central.sonatype.org/publish/publish-portal-maven/

PR #6429, opened 2026-06-01
[refactor] Modernize Maven 3 build pipeline and release automation

### LINKED ISSUE #3394: [feature] provide maven profiles for specific use-cases

When building existdb locally, I just want it to create the directory containing the built application.
`exist-distribution/target/exist-distribution-*-dir`
This is all that is needed in order to start the database from command line.

The command to achieve I use is:

```bash
mvn -T2C clean package -DskipTests -Ddependency-check.skip=true -Ddocker=false -P \!build-dist-archives,\!mac-dmg-on-mac,\!codesign-mac-dmg,\!mac-dmg-on-unix,\!installer,\!concurrency-stress-tests,\!micro-benchmarks
```
 
The amount of profiles that are on by default and therefore need to be disabled is hard to memoize. I created an alias for that to not have to.

I would like to see a local build to skip all the unnecessary steps with `mvn clean package -P local-build`.

- docker should default to false
- tests should be able to run on the command line of a development machine
- all of the following profiles should be not run by default:
    - build-dist-archives
    - mac-dmg-on-mac
    - codesign-mac-dmg
    - mac-dmg-on-unix
    - installer
    - concurrency-stress-tests
    - micro-benchmarks

For other use cases like building on travis there should be a specific profile that enables all the sub-profiles or tasks necessary. One for travis would be `mvn package -P ci-build`.

That way additional steps can easily run when requested by adding them to the `-P` parameter.

As mentioned above, the current workaround is creating an alias for the long command.
Downsides are: it has to be done for each dev / development machine. Changes are not propagated. Harder to document (it is not properly yet). It is also less flexible, because flags and profiles are harder to re-enable for single runs.

### LINKED ISSUE #3411: Distribution archives not built by default
The distribution archives (e.g. .zip and .tar.bz2) are not built by default (which they should be). 

See: https://maven-users.markmail.org/search/?q=#query:+page:1+mid:f42fikz2xtqi7ny4+state:results

### LINKED ISSUE #4117: [BUG] Wrong Docker tags are set as part of mvn:release
Whilst releasing 5.3.1, I see the following Docker tags on Docker Hub:

* `5.3.1` - this is correct
* `latest` - this points at `5.3.1`
* `release` - this seems to be point at an old version of eXist-db from 6 months ago

However -

* I would think that `latest` should be pointing at the most recent build from Docker Hub of `HEAD` of `develop`.
* I would think that `release` should point at 15.3.11

### LINKED ISSUE #4217: Improve usage of GitHub-release plugin when publishing new versions of existdb
When releasing exist-db 5.3.0 we incorporated a new maven plugin GitHub-release-plugin

Its output should be improved.

- it should __not__ upload the pom.xml
- the link to the release notes is created by is wrong without manual interference

### LINKED ISSUE #4519: [feature] revise pom/versioning
Make update of project/app version more easy to do
requires 'recent but not too recent version of maven'

see https://maven.apache.org/maven-ci-friendly.html

### LINKED ISSUE #5597: [feature] move installer generation to CI
the signed installers should be generated on CI only, and removed from the default build profiles

see #3394

### LINKED ISSUE #5641: Build: make sure that dependency-check (NVD) is only executed in exist-distribution.
This wil save us build time, and imo only that what is distributed is relevant for end users?

### LINKED ISSUE #6177: [feature] staple notarization ticket to DMG contents
With stapling the DMG should be installable without having to go to system settings.

Apple's documentation on this subject https://developer.apple.com/documentation/xcode/packaging-mac-software-for-distribution/

### LINKED ISSUE #6411: Cannot `mvn package` twice
At the moment the deletion of the `etc/jetty` directories in the execution with the id `delete-jetty-dest-before-rename` will lead to  the execution later with id `adjust-jetty-deploy-for-appassembler` to fail on builds that were not cleaned before. Which means that in order to package exist-db one has to _always_ clean beforehand.

I believe this should not have to happen but I will just open the issue first because there might be a good reason for it.

```xml
                    <execution>
                        <id>delete-jetty-dest-before-rename</id>
                        <phase>package</phase>
                        <goals>
                            <goal>run</goal>
                        </goals>
                        <configuration>
                            <target>
<!--                                <delete dir="${assemble.dir}/etc/jetty" failonerror="false" quiet="true"/>-->
<!--                                <delete dir="${mac.bundle.resources.dir}/etc/jetty" failonerror="false" quiet="true"/>-->
                            </target>
                        </configuration>
                    </execution>
```

### LINKED ISSUE #6414: [bug] mac.codesign.identity default in codesign-mac-app/dmg profiles causes "no identity found" for contributors without the matching certificate

## Summary

Building any non-Mac-distribution target with `-Pcodesign-mac-app` (or `-Pcodesign-mac-dmg`) on macOS fails for any contributor whose keychain doesn't contain the specific Developer ID Application certificate that's currently set as the default in `exist-distribution/pom.xml`. Surfaced by @dizzzz in #6413 while working on jansi native-binary signing.

## Reproduce

```
mvn clean install -DskipTests -Pcodesign-mac-app
```

(or any `mvn ... -Pcodesign-mac-app` invocation that doesn't also pass `-Dmac.codesign.identity=...`)

## What happens

```
[ERROR] test-requirement: code failed to satisfy specified code requirement(s)
[ERROR] Developer ID Application: Evolved Binary Ltd: no identity found
[ERROR] Command execution failed.
```

## Root cause

`exist-distribution/pom.xml` line 57 hardcodes the default codesign identity:

```xml
<mac.codesign.identity>Developer ID Application: Evolved Binary Ltd</mac.codesign.identity>
```

That value is referenced from three places — both `codesign` executions in the `codesign-mac-app` profile (jansi native signing + app bundle signing) and the DMG signing step in `codesign-mac-dmg`. It's the historic release-pipeline default, but with that identity hardcoded as the *default* rather than a *suggested override*, every contributor who runs either profile on a Mac without that specific certificate in their keychain hits this exact error.

The activation guards on those profiles include `-Dmac-signing=true` as a property condition, but **explicit `-P<id>` activation bypasses `<activation>`**, so `-Pcodesign-mac-app` runs the codesign executions regardless of whether `mac-signing` was set. That's why iterating on Mac-related changes (jansi, app bundling, etc.) trips this immediately.

## Why this matters

Any contributor working on the Mac packaging path — currently @dizzzz on #6413 — has to either install a certificate they don't have, manually pass `-Dmac.codesign.identity=...` every time, or skip the profile and lose the validation it provides.

## Possible fixes (sketches, not prescribing)

| Option | Sketch |
|---|---|
| **A** | Remove the default property; require `-Dmac.codesign.identity=...` to use either profile. Explicit, no surprises. |
| **B** | Default to ad-hoc signing (`-`). The profile then works for anyone on a Mac, producing a non-Gatekeeper-trusted signature; release pipelines still pass the real identity via `-D`. |
| **C** | Keep the property but tighten the `codesign` execution bindings so they only run when `mac-signing=true` is set — explicit `-P` alone would activate the profile but not the signing executions. |
| **D** | Detect identities at build time (`security find-identity -v -p codesigning`) and pick a Developer ID if one is present, falling back to ad-hoc otherwise. Most magical, brittle on CI runners with empty keychains. |

I'd lean B as the smallest, most contributor-friendly default, but happy to be talked out of it.

## Coordination note

@duncdrum mentioned in Slack that he's already reshuffling the profiles and signing around — filing as an issue (not a PR) so we don't conflict with that in-flight work. Happy to pick up whichever option lands here once direction is settled.

duncdrum:
## Summary

Ports the lessons from the deferred Maven 4 experiment (#6276) into a
production-ready Maven 3 pipeline. Replaces `maven-release-plugin` with a
two-workflow GitHub Actions pattern, simplifies the profile graph, modernizes
macOS notarization (notarytool + API key auth), and adds Windows Authenticode
signing via Azure Key Vault. Maven Central publishing migrated from legacy
OSSRH to the new Central Portal (`central-publishing-maven-plugin`).

### POM modernization
- All 59 POMs switched to `${revision}` (CI-friendly versions). `<revision>7.0.0-SNAPSHOT>` set as default in `exist-parent`.
- `flatten-maven-plugin 1.7.3` added — `flattenMode=resolveCiFriendliesOnly` — so deployed POMs resolve `${revision}` to a literal version. `flatten.clean` execution bound to `clean` phase so `mvn package` is idempotent.
- Minimum Maven version bumped to 3.9.16.

### Profile simplification
- Removed `activeByDefault` from `installer` / `concurrency-stress-tests` / `micro-benchmarks` — they no longer leak into contributor builds.
- New umbrella profiles: `local-build` (default), `release-build` (adds installer + GPG signing + Central publishing), `perf-tests` (stress + benchmarks).
- Removed redundant `installer`, `concurrency-stress-tests`, `micro-benumbrella restructure.

### Release pipeline (CI)
- New `ci-release-prepare.yml` (workflow_dispatch): updates `CITATION.cff`, commits, creates annotated `eXist-X.Y.Z` tag, pushes via `RELEASE_PAT`.
- New `ci-release.yml` (tag-triggered): parallel `build-linux` (Maven Cbuild-mac` (signed + notarized DMG), `build-windows` (signed installerJAR + Authenticode `.exe`), converging on `publish-github-release`.
- `release-preflight` enforcer profile validates required secrets befor
- Workflows hardened: least-privilege permissions, tokens out of CLI args, third-party actions SHA-pinned.

### macOS signing
- Migrated from deprecated `altool --notarize-app` to `notarytool submitaple` with App Store Connect API Key auth (no Apple ID, no 2FA).
- Decoupled `mac-dmg-on-mac` activation from `mac-signing` — contributors get an unsigned DMG automatically on macOS without extra flags.
- Removed default `mac.codesign.identity`.
- Jansi native code now signed in both jarfiles.

### Windows signing
- Azure OIDC federated identity via `azure/login@v2` (no long-lived sec
- JAR signing via `azure-security-keyvault-jca`; `.exe` Authenticode via `AzureSignTool`.

### Documentation
- `exist-versioning-release.md` rewritten for the two-workflow pattern

## Closes

- Closes #3394 — Maven profiles for specific use-cases
- Closes #3411 — distribution archives not built by default
- Closes #4117 — wrong Docker tags during release
- Closes #4217 — improve GitHub-release plugin usage
- Closes #4519 — revise POM versioning
- Closes #5597 — move installer generation to CI
- Closes #5641 — dependency-check only in exist-distribution
- Closes #6177 — staple notarization ticket to DMG
- Closes #6411 — cannot `mvn package` twice
- Closes #6414 — default `mac.codesign.identity` causes contributor bui

see  #6176

### DIFF
--- a/exist-versioning-release.md
+++ b/exist-versioning-release.md
@@ -115,9 +115,11 @@ The `ci-release.yml` workflow runs four jobs in parallel then converges:
 | `build-windows` | windows-latest | signed installer JAR + Authenticode `.exe` |
 | `publish-github-release` | ubuntu-latest | GitHub Release with all assets attached |
 
-### Prerequisites: required GitHub secrets
+### Prerequisites: required GitHub secrets and repository rules
 
 All secrets must be in place under **Settings → Secrets and variables → Actions** before running a release. See `plans/internal-release-runbook.template.md` for the full table and rotation guidance.
+
+A **tag ruleset** restricts creation and deletion of `eXist-*` tags to users with the Maintain or Admin role (Settings → Rules → Rulesets). The `RELEASE_PAT` owner must hold one of those roles; all other automation only reads tags and is unaffected.
 
 | Secret | Used by |
 |--------|---------|
@@ -136,7 +138,7 @@ All secrets must be in place under **Settings → Secrets and variables → Acti
 | `AZURE_TENANT_ID` | `build-windows` — Azure tenant |
 | `AZURE_SUBSCRIPTION_ID` | `build-windows` — Azure subscription |
 
-Repository variable (not a secret): `AZURE_KEYVAULT_URI` (e.g. `https://exist-db-signing.vault.azure.net/`) and `EXISTDB_MAC_CODESIGN_IDENTITY`.
+Repository variables (not secrets): `AZURE_KEYVAULT_URI` (e.g. `https://exist-db-signing.vault.azure.net/`), `AZURE_KEYVAULT_CERT_NAME` (the certificate name as it appears in Key Vault, e.g. `existdb-code-signing`), and `EXISTDB_MAC_CODESIGN_IDENTITY`.
 
 ### Preparing a Product Release
 
--- a/exist-versioning-release.md
+++ b/exist-versioning-release.md
@@ -127,11 +127,11 @@ All secrets must be in place under **Settings → Secrets and variables → Acti
 | `EXISTDB_RELEASE_KEY_PASSPHRASE` | `build-linux` — GPG passphrase |
 | `CENTRAL_TOKEN_USERNAME` | `build-linux` — Sonatype Central Portal user token username |
 | `CENTRAL_TOKEN_PASSWORD` | `build-linux` — Sonatype Central Portal user token password |
-| `EXISTDB_MAC_CERTIFICATE` | `build-mac` — Developer ID Application cert (base64 PKCS12) |
+| `EXISTDB_MAC_CERTIFICATE` | `build-mac` — Developer ID Application cert (base64 PKCS12) for codesigning |
 | `EXISTDB_MAC_CERTIFICATE_PASSWORD` | `build-mac` — Certificate password |
-| `EXISTDB_APPLE_ID` | `build-mac` — Apple ID for `notarytool` |
-| `EXISTDB_APPLE_APP_PASSWORD` | `build-mac` — App-specific password for `notarytool` |
-| `EXISTDB_APPLE_TEAM_ID` | `build-mac` — Apple Developer Team ID |
+| `EXISTDB_APPLE_API_KEY` | `build-mac` — App Store Connect API private key (base64-encoded `.p8`) for notarytool |
+| `EXISTDB_APPLE_API_KEY_ID` | `build-mac` — API Key ID (10-char alphanumeric) |
+| `EXISTDB_APPLE_API_ISSUER_ID` | `build-mac` — App Store Connect Issuer ID (UUID) |
 | `AZURE_CLIENT_ID` | `build-windows` — OIDC app registration client ID |
 | `AZURE_TENANT_ID` | `build-windows` — Azure tenant |
 | `AZURE_SUBSCRIPTION_ID` | `build-windows` — Azure subscription |
--- a/exist-versioning-release.md
+++ b/exist-versioning-release.md
@@ -117,7 +117,7 @@ The `ci-release.yml` workflow runs four jobs in parallel then converges:
 
 ### Prerequisites: required GitHub secrets and repository rules
 
-All secrets must be in place under **Settings → Secrets and variables → Actions** before running a release. See `plans/internal-release-runbook.template.md` for the full table and rotation guidance.
+All secrets must be in place under **Settings → Secrets and variables → Actions** before running a release. See `internal-release-runbook.template.md` for the full table and rotation guidance.
 
 A **tag ruleset** restricts creation and deletion of `eXist-*` tags to users with the Maintain or Admin role (Settings → Rules → Rulesets). The `RELEASE_PAT` owner must hold one of those roles; all other automation only reads tags and is unaffected.
 
--- a/exist-versioning-release.md
+++ b/exist-versioning-release.md
@@ -84,150 +84,87 @@ It is trivial for a developer to relate a timestamp back to a Git hash (by using
 
 ### Where the version number is stored
 
-The version number is stored in the `exist-parent/pom.xml` file, in a single property, `<version>`. The Semantic Versioning number `3.2.0-SNAPSHOT` would be stored as follows:
+The version number is stored in `exist-parent/pom.xml` as the `<revision>` CI-friendly property:
+```xml
+<properties>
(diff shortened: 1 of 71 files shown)


exist-versioning-release.md
```diff
@@ -298,6 +237,12 @@ Central (staging), Docker images to Docker Hub, and eXist-db distributions and i
 13. Go to GitHub and move all issues and PRs which are still open for the release milestone to the next release milestone. Close the release milestone.
 
 
+### Manual fallback (CI unavailable)
+
+If the CI release workflow is unavailable, the `exist-release` Maven profile (backed by `maven-release-plugin`) remains usable for emergency releases. This path requires a local macOS machine with valid Apple credentials, GnuPG, and all signing keys available in `~/.m2/settings.xml`. See `plans/internal-release-runbook.template.md` §9 for the full procedure and required properties.
```
line-o, line 242: > See `plans/internal-release-runbook.template.md` §9

I this document part of the PR?

duncdrum: it will go into the private exist-internal repo

duncdrum: let me fix the reference

line-o (approved):

windauer (approved):

reinhapa (approved):

PR #6430, opened 2026-06-01
[bugfix] FileLock: remove shared ByteBuffer to fix concurrent save/read race

### LINKED ISSUE #4334: [BUG] FileLock has insufficient synchronisation

joewiz:

## Summary

Closes #4334. `FileLock` held a single `ByteBuffer` as a `final` instance field, mutated by both `save()` (Quartz heartbeat worker thread, every 10.1 s) and `read()` (main startup thread during `tryLock()`). Both methods rely on the buffer's `position`/`limit` cursors, so an overlap could clobber the in-flight state — the source of the `BufferOverflowException` / `IllegalArgumentException` reported in the issue.

## Fix

Allocate the buffer per call inside `save()` and `read()`. The 16-byte allocation is negligible against the file I/O the methods are doing, and removing the shared state eliminates the race without requiring synchronization. `lastHeartbeat` is now `volatile` so the heartbeat thread's writes are visible to `checkHeartbeat()` on the main thread.

## Race shape

```
Heartbeat:   buf.clear()             // pos=0,  limit=16
Heartbeat:   buf.put(MAGIC)          // pos=8,  limit=16
Main:        channel.read(buf)       // pos=16, limit=16 (reads into pos 8..15)
Main:        buf.flip()              // pos=0,  limit=16
Main:        buf.get(magic)          // pos=8,  limit=16
Heartbeat:   buf.putLong(now)        // BufferOverflowException OR
                                     // Buffer.position(int) IllegalArgumentException
                                     // depending on exact interleave
```

The heartbeat fires every 10.1 s, so the window is small in steady-state. The race needs a heartbeat to fire during a fresh startup of another concurrent process — which is exactly what container restarts and CI parallel runs produce.

## Drive-by Codacy cleanups

PMD findings on the touched file, addressed per the project rule:

- `AvoidReassigningParameters` on `message(String, Exception)`: use a local variable instead of reassigning the parameter.
- `AvoidThrowingNullPointerException` on `save()`'s catch block: the catch-and-rethrow preserves the original NPE (not a constructed one) — added a `@SuppressWarnings` with rationale, since the rule is meant to discourage constructing new NPEs, not rethrowing caught ones.

## Test plan

- [x] `StartupLockingTest` passes (2/2)
- [x] `exist-core` builds clean
- [x] Codacy/PMD clean on the touched file
- [x] CI green

### DIFF
--- a/exist-core/src/main/java/org/exist/storage/lock/FileLock.java
+++ b/exist-core/src/main/java/org/exist/storage/lock/FileLock.java
@@ -86,11 +86,15 @@ public class FileLock {
     /** An open channel to the lock file */
     private SeekableByteChannel channel = null;
 
-    /** Temporary buffer used for writing */
-    private final ByteBuffer buf = ByteBuffer.allocate(MAGIC.length + 8);
-
-    /** The time (in milliseconds) of the last heartbeat written to the lock file */
-    private long lastHeartbeat = -1L;
+    /**
+     * The time (in milliseconds) of the last heartbeat written to the lock file.
+     * Marked {@code volatile} so the Quartz scheduler thread that calls
+     * {@link #save()} every {@link #HEARTBEAT} ms publishes its writes
+     * visibly to the main thread, which reads the value during
+     * {@link #checkHeartbeat()} when deciding whether to take over a stale
+     * lock file.
+     */
+    private volatile long lastHeartbeat = -1L;
 
     public FileLock(final BrokerPool pool, final Path path) {
         this.pool = pool;
@@ -229,14 +233,22 @@ public class FileLock {
         this.channel = Files.newByteChannel(lockFile, READ, WRITE, SYNC);
     }
 
+    @SuppressWarnings("PMD.AvoidThrowingNullPointerException")
     protected void save() throws IOException {
         try {
             if (channel == null) {
                 open();
             }
-            
-            long now = System.currentTimeMillis();
-            buf.clear();
+
+            // Allocate per-call rather than reusing a shared instance field:
+            // save() and read() ran concurrently from the Quartz heartbeat
+            // worker thread and the main startup thread without
+            // synchronization, so a shared ByteBuffer's position/limit could
+            // be clobbered mid-write -- the source of the
+            // BufferOverflowException / IllegalArgumentException reported in
+            // issue #4334.
+            final ByteBuffer buf = ByteBuffer.allocate(MAGIC.length + 8);
+            final long now = System.currentTimeMillis();
             buf.put(MAGIC);
             buf.putLong(now);
             buf.flip();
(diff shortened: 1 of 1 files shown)


dizzzz (approved):
LGTM - shared mem makes no sense ... unless the mem buffer is used for something special under the hood.

joewiz:

@dizzzz Thanks — that's the right reviewer instinct to check before changing shared mutable state. I went back through the history and the answer is "no, nothing special under the hood": the buffer was introduced in the very first commit of `FileLock.java` (`03a7074b13`, 2006-08-15, Wolfgang) with the literal comment `/** Temp buffer used for writing */`, and the surrounding code has always treated it as a reusable scratch buffer rather than as state.

Three pieces of evidence:

- `ByteBuffer.allocate(...)`, not `ByteBuffer.allocateDirect(...)`. The one plausible reason to share a buffer instance across calls — avoiding repeated direct-buffer allocation for native zero-copy I/O — doesn't apply to a heap buffer. Heap buffers are trivially cheap to allocate.
- The buffer never escapes the class — it's only ever the argument to `channel.read(buf)` / `channel.write(buf)`. Nothing off-class consumes continuity of state across calls.
- The original code called `buf.clear()` between uses (in both `save()` and at the end of `read()`), which is the explicit "reset cursors for next use" idiom — exactly what you do with a scratch buffer, not with state-carrying storage.

Most likely a 2006-era micro-optimization that JVM allocation cost has long since rendered moot. The threading model was the same on day one (`SyncDaemon.executePeriodically(...)` ran the heartbeat on one thread while `tryLock()` / `release()` ran on another), so the race existed from the start; it just took until 2021 for the interleave to be observed and reported.

By the way — this fix is the first sub-system of a broader concurrency audit of the storage layer we're starting. Project plan + the FileLock-specific findings doc are drafted in our vault; happy to share when the umbrella is ready for circulation.

line-o (approved):
Full agree - there is no need for a shared buffer when writing the buffer to the channel is the _actual_ atomic operation.

line-o:
Can be merged when the codefreeze is lifted - /cc @duncdrum

duncdrum (approved):

Issue #5579, opened 2024-12-11
Question, perhaps a bug: the autodeploy folder does not upgrade packages if the version is newer


DrRataplan:
**Describe the bug**
I am trying to automatically redeploy a newer version of an installed app in a docker container. I expected that if I just update the `version` of the `expath-pkg.xml`, it would do so. The code at https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/repo/Deployment.java#L188 insinuates that the version will be compared, but alas, `enforceDeps` [is false](https://github.com/eXist-db/exist/blob/develop/exist-core/src/main/java/org/exist/repo/AutoDeploymentTrigger.java#L125), so the version check is skipped.

**Expected behavior**
At first I expected a deploy to be done, as long as if the autodeploy folder contains a package with a version that exceeds the version that's currently installed. But then again, this could result into unexpected behaviour if the code is manually changed (ie. with eXide or whatever) and the database reboots.

**To Reproduce**

* Get a clean ExistDB
* Check out a package locally. In this example I'll use [tuttle](https://github.com/eeditiones/tuttle), but the actual package is irrelevant.
* Run `npm run build` to build tuttle
* Copy the xar to the autodeploy folder
* Start exist and wait until everything is autodeployed
* Verify the package is installed
* Stop exist
* Modify the package, at least change the version in its expath-pkg.xml (https://github.com/eeditiones/tuttle/blob/master/src/expath-pkg.xml#L3). Change it to `10000.0.0`, to be safe.
* Run `npm run build` again
* Remove the old xar from the `autodeploy` folder
* Add the new xar to the `autodeploy` folder
* Start exist
* Observe something along the lines of this in the logs: 
```
#33 87.63 11 Dec 2024 13:26:50,178 [main] INFO  (Deployment.java [installAndDeploy]:192) - Application package http://e-editiones.org/tuttle already installed. Skipping.
````
* The package is not redeployed.

**Context (please always complete the following information)**

* Build: eXist-6.2.0 (c8fa4958b6d4a50bd0cba7f3e76a150226414187)
* Java: 11.0.18 (Debian)
* OS: Linux 6.11.0-9-generic (amd64)

**Additional context**

I am setting up a (stack of) docker layers to:
- deploy some data,
- deploy a version of the application code
- transform the data using the application code (takes a long time)
- push this to a registry
- From there, on every commit of the codebase, generate a new docker container with the new codebase

[...]

PR #6433, opened 2026-06-02
[bugfix] repo:install-and-deploy respect explicitly-requested version

joewiz:

## Summary

When a higher version of a package is already installed in the EXPath registry, calling `repo:install-and-deploy($name, $version, $url)` with an explicit *lower* `$version` reports success but silently deploys the existing higher version instead of the one the user asked for. This PR makes the deploy step use the freshly-installed package directly, so the explicit version request is honored.

This was surfaced by Craig Berry on Slack while attempting to downgrade `tei-publisher-lib` from 6.1.0 to 6.0.2. A companion fix for the related silent-failure on `repo:remove` is in #6432.

## What changed

**`exist-core/src/main/java/org/exist/repo/Deployment.java`**

- Added a `Package`-based overload `deploy(DBBroker, Txn, Package, String)`. The body is the prior `deploy(... String pkgName, Optional<ExistRepository> repo, ...)` lifted to take a `Package` directly, so the package-dir / abbrev / version lookups all flow from one consistent source instead of re-resolving by name halfway through.
- The original by-name overload is kept (for `repo:deploy(name)` callers) and delegates to the new one via `getPackage(pkgName, repo)`.
- `installAndDeploy(...)` now resolves the just-installed package by the version extracted from the XAR descriptor (`Packages.version(requestedVersion)`) rather than relying on what `installPackage()` returns — which, for the higher-version-already-present case, was `packages.latest()`. The lookup is in a small `resolveInstalledVersion()` helper to keep `installAndDeploy`'s NPath complexity at its prior level.

**`extensions/modules/expathrepo/src/test/xquery/modules/expathrepo/deployment.xql`**

- New regression test `deploy:install-lower-version-after-higher`: installs version 1.0.0, then installs version 0.5.0 (using fresh fixtures with name `http://exist-db.org/apps/dtest-versioning` to avoid polluting other tests), and asserts that the `expath-pkg.xml` written to the deployed target collection `/db/apps/dtest-versioning/` reports version 0.5.0. Before the fix this returned 1.0.0.
- The assertion reads `doc("/db/apps/<target>/expath-pkg.xml")` rather than `repo:get-resource(name, "expath-pkg.xml")` — the latter returns from `packages.latest()`'s repo dir and so reports 1.0.0 even after a correct deploy of 0.5.0; only the deployed target collection shows what was actually delivered to the user.

## Root cause

In `Deployment.installAndDeploy`:

1. `repo.getParentRepo().installPackage(xar, ...)` is called, which installs the XAR into the registry and returns *some* `Package` — but for an EXPath name with multiple versions present, it returns `packages.latest()`, not necessarily the just-installed one.
2. The follow-up `deploy(pkgName, ...)` then calls `getPackage(pkgName, repo)`, which also returns `packages.latest()`, and deploys from *that* package's dir.

So even though the right XAR was unpacked into `data/expathrepo/<abbrev>-<requested>/`, the deploy step copies files out of `data/expathrepo/<abbrev>-<latest>/` into `/db/apps/<target>/`. Result: install OK, but deploy installs the wrong version, with no error.

## Test plan

- [x] `mvn test -pl extensions/modules/expathrepo -Dtest='ExpathRepoTests'` — 9/9 pass (8 prior + new regression)
- [x] Codacy / PMD on changed Java file — no new warnings introduced (NPath, `cleanup()` unused, parameter reassignment warnings are pre-existing on this file)
- [x] Manually reproduced Craig's scenario (install 1.0.0, then `repo:install-and-deploy(name, "0.5.0", url)` against local copy of the XAR) — before this commit, deployed `expath-pkg.xml` reports 1.0.0; after, 0.5.0.

## Related

- https://github.com/eXist-db/exist/pull/6432 — sibling fix for `repo:remove` returning `false` with no diagnostic
- https://github.com/eXist-db/xst/pull/376 — client-side fix so xst surfaces the underlying failure instead of printing `✔︎ uninstalled`
- Slack thread: https://exist-db.slack.com/archives/CG2MRUZ35/p1780415385757139

### DIFF
--- a/exist-core/src/main/java/org/exist/repo/Deployment.java
+++ b/exist-core/src/main/java/org/exist/repo/Deployment.java
@@ -375,124 +375,185 @@ public class Deployment {
         }
         try {
             // if there's a <setup> element, run the query it points to
-            final Optional<ElementImpl> setup = findElement(repoXML, SETUP_ELEMENT);
-            final Optional<String> setupPath = setup.map(ElementImpl::getStringValue).filter(s -> !s.isEmpty());
-
+            final Optional<String> setupPath = findElementValue(repoXML, SETUP_ELEMENT);
             if (setupPath.isPresent()) {
                 runQuery(broker, null, packageDir, setupPath.get(), pkgName, QueryPurpose.SETUP);
                 return Optional.empty();
-            } else {
-                // otherwise create the target collection
-                XmldbURI targetCollection = null;
-                if (userTarget != null) {
-                    try {
-                        targetCollection = XmldbURI.create(userTarget);
-                    } catch (final IllegalArgumentException e) {
-                        throw new PackageException("Bad collection URI: " + userTarget, e);
-                    }
-                } else {
-                    final Optional<ElementImpl> target = findElement(repoXML, TARGET_COLL_ELEMENT);
-                    final Optional<String> targetPath = target.map(ElementImpl::getStringValue).filter(s -> !s.isEmpty());
-
-                    if (targetPath.isPresent()) {
-                        // determine target collection
-                        try {
-                            targetCollection = XmldbURI.create(getTargetCollection(broker, targetPath.get()));
-                        } catch (final IllegalArgumentException e) {
-                            throw new PackageException("Bad collection URI for <target> element: " + targetPath.get(), e);
-                        }
-                    } else {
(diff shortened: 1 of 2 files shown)


duncdrum:
related see #5579

duncdrum:
@joewiz can you rebase this please, and address the codacy warning, let's see if other CI jobs are green after the rebase

duncdrum (approved):

PR #6434, opened 2026-06-02
[bugfix] Fix macOS .app launch on 7.0.0-beta3 (LSArchitecturePriority + stripFirstElement + IzPack minVersion)

joewiz:

## Summary

The 7.0.0-beta3 macOS DMG fails to launch from the Dock or Finder. Three independent fixes are needed; this PR carries all three. Verified locally end-to-end: with all three applied, a freshly built `eXist-db.app` reaches the launcher's configuration dialog instead of disappearing from the Dock.

The macOS findings were diagnosed and shared with @duncdrum on Slack; @duncdrum confirmed the analysis and flagged the IzPack stale-JRE-version reference as a related fix worth folding in.

### 1. `exist-start/src/main/java/org/exist/start/Main.java` — guard `stripFirstElement` against zero-arg invocation

Invoking the launcher directly (`Contents/MacOS/eXist-JavaAppLauncher`) crashed with:

```
java.lang.NegativeArraySizeException: -1
    at org.exist.start.Main.stripFirstElement(Main.java:247)
    at org.exist.start.Main.startExistdb(Main.java:187)
    at org.exist.start.Main.run(Main.java:156)
    at org.exist.start.Main.main(Main.java:83)
```

`stripFirstElement` allocated `new String[args.length - 1]` before checking `args.length`. The macOS app-bundle launch passes an empty `JVMArguments` array, so `args.length == 0` and the allocation throws. The method's own Javadoc already documented the intended behaviour ("If the input array has one or no elements, an empty array is returned") — the implementation just had the length check after the allocation. Move the guard above the allocation. Visibility relaxed from `private` to package-private so the new regression test can call it directly.

Regression test `MainTest.java` covers zero-arg, single-arg boundary, and multi-arg happy path.

### 2. `exist-distribution/pom.xml` — populate `LSArchitecturePriority`

The bundle's `Info.plist` contained an EMPTY `LSArchitecturePriority` array. macOS LaunchServices reads that as "no architectures supported" and silently refuses to launch the .app, producing in Console.app:

```
CoreServicesUIAgent  LAUNCH: No architectures specified in launch architectures [ NULL ]
                     for app=org.exist.start.Main  which likely is an error.
CoreServicesUIAgent  LAUNCH: Timed-out waiting for launch of application 0x0-0x3139136.
```

The bundled `eXist-JavaAppLauncher` is a proper universal binary (`x86_64` + `arm64`) and correctly signed + notarized — only the plist entry was missing. Root cause: `appbundler-maven-plugin` emits an empty `LSArchitecturePriority` when its `<architectures>` parameter is unset. Populated with `arm64` and `x86_64` to match the slices the launcher binary actually carries.

Verified after rebuild:

```bash
/usr/libexec/PlistBuddy -c "Print :LSArchitecturePriority" \
    exist-distribution/target/exist-distribution-7.0.0-SNAPSHOT.app/Contents/Info.plist
# Array {
#     arm64
#     x86_64
# }
```

### 3. `exist-installer/src/main/izpack/install.xml` — bump `JDKPathPanel.minVersion` off 1.8

The IzPack `JDKPathPanel.minVersion` was still hardcoded to `1.8`. This is the gate that validates the JDK directory a user picks during the installer wizard; with it pinned at 1.8, the panel would happily accept any JDK ≥ 1.8 even though eXist 7.0 won't run on anything below 21. Switched to `@{maven.compiler.release}` so this tracks the project's real minimum JDK going forward, matching the templating already used two lines above for `<javaversion>`.

## Test plan

- [x] `mvn test -pl exist-start` — 27/27 pass, including the new `MainTest` regression cases
- [x] Codacy / PMD on changed Java file — clean (existing pre-existing findings on `Main.java` have `@SuppressWarnings` added in commit 1)
- [x] `mvn -T1.5C package -pl exist-distribution -am -DskipTests -P '!mac-dmg-on-mac'` — `BUILD SUCCESS`; `LSArchitecturePriority` populated in generated `Info.plist`
- [x] Direct launch of `Contents/MacOS/eXist-JavaAppLauncher` from the freshly built unsigned bundle now reaches the launcher's configuration dialog (`First launch: opening configuration dialog`) instead of throwing `NegativeArraySizeException`
- [ ] Full end-to-end DMG validation (signed + notarized) — needs a release build on the CI Mac runner

## Related

- Slack thread with diagnosis output: https://exist-db.slack.com/archives/C0824H3C76J/p1780416133977389
- 7.0.0-beta3 release: https://github.com/eXist-db/exist/releases/tag/eXist-7.0.0-beta3

### DIFF
--- a/exist-distribution/pom.xml
+++ b/exist-distribution/pom.xml
@@ -925,6 +925,19 @@
                             <executableName>eXist-JavaAppLauncher</executableName>
                             <jvmRequired>21</jvmRequired>
                             <minimumSystemVersion>10.9</minimumSystemVersion>
+                            <!-- Populate LSArchitecturePriority so macOS LaunchServices
+                                 will launch the .app. The plugin emits an empty
+                                 LSArchitecturePriority array if this is unset, which
+                                 LaunchServices reads as "no architectures supported"
+                                 and refuses to launch (silent failure in the Dock,
+                                 surfaced in Console.app as:
+                                 "LAUNCH: No architectures specified in launch architectures [ NULL ]").
+                                 The eXist-JavaAppLauncher native binary is built as
+                                 a universal binary carrying x86_64 + arm64 slices. -->
+                            <architectures>
+                                <architecture>arm64</architecture>
+                                <architecture>x86_64</architecture>
+                            </architectures>
                             <version>${project.version}</version>
                             <shortVersion>${project.version}</shortVersion>
                             <icon>icon.icns</icon>
--- a/exist-installer/src/main/izpack/install.xml
+++ b/exist-installer/src/main/izpack/install.xml
@@ -79,7 +79,7 @@
 
     <variables>
         <variable name="JDKPathPanel.skipIfValid" value="yes"/>
-        <variable name="JDKPathPanel.minVersion" value="1.8"/>
+        <variable name="JDKPathPanel.minVersion" value="@{maven.compiler.release}"/>
         <variable name="ShowCreateDirectoryMessage" value="false"/>
         <variable name="TargetPanel.dir.windows" value="${ENV[SystemDrive]}\${APP_NAME}"/>
     </variables>
--- a/exist-start/src/main/java/org/exist/start/Main.java
+++ b/exist-start/src/main/java/org/exist/start/Main.java
@@ -177,6 +177,11 @@ public class Main {
      *                        incompatible Java version, misconfiguration, or failure to
      *                        resolve required resources.
      */
+    // PMD AvoidReassigningParameters: args is shuffled through stripFirstElement / addFirstElement
+    // as part of the launcher's argument-massaging pipeline; refactoring to a local would obscure
+    // the per-mode argument transformations. Suppression is the safer choice in this release-
+    // path code.
+    @SuppressWarnings("PMD.AvoidReassigningParameters")
     public void startExistdb(String[] args) throws StartException {
 
         // Check if the OpenJDK version can corrupt eXist-db
@@ -243,11 +248,17 @@ public class Main {
      * @return a new array containing all elements of the input array except the first one;
      *         if the input array contains no elements or only one element, an empty array is returned
      */
-    private static String[] stripFirstElement(final String[] args) {
-        final String[] newArguments = new String[args.length - 1];
-        if (args.length > 1) {
-            System.arraycopy(args, 1, newArguments, 0, args.length - 1);
+    static String[] stripFirstElement(final String[] args) {
+        // Guard against args.length == 0: `new String[args.length - 1]` would otherwise
+        // allocate `new String[-1]` and throw NegativeArraySizeException. This path is
+        // exercised by the macOS app-bundle launch, which invokes the JVM directly with
+        // no arguments (the appbundler-generated Info.plist passes an empty JVMArguments
+        // array). Reported in 7.0.0-beta3 dock-launch failure.
+        if (args.length <= 1) {
+            return new String[0];
         }
+        final String[] newArguments = new String[args.length - 1];
+        System.arraycopy(args, 1, newArguments, 0, args.length - 1);
         return newArguments;
     }
 
@@ -433,6 +444,7 @@ public class Main {
      *
      * @throws StopException if the shutdown process encounters an error
      */
+    @SuppressWarnings("PMD.AvoidAccessibilityAlteration") // intentional: BrokerPools.stopAll is package-private and invoked reflectively from test infrastructure (see method javadoc)
     public void shutdownExistdb() throws StopException {
         // only used in test suite
         try {
--- a/exist-start/src/test/java/org/exist/start/MainTest.java
+++ b/exist-start/src/test/java/org/exist/start/MainTest.java
@@ -1,0 +1,58 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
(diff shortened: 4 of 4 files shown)


exist-installer/src/main/izpack/install.xml
```diff
@@ -79,7 +79,7 @@
 
     <variables>
         <variable name="JDKPathPanel.skipIfValid" value="yes"/>
-        <variable name="JDKPathPanel.minVersion" value="1.8"/>
+        <variable name="JDKPathPanel.minVersion" value="@{maven.compiler.release}"/>
```
duncdrum, line 82: That's the one. Just curious why is it  `@..` 😕

exist-distribution/pom.xml
```diff
@@ -925,6 +925,19 @@
                             <executableName>eXist-JavaAppLauncher</executableName>
                             <jvmRequired>21</jvmRequired>
                             <minimumSystemVersion>10.9</minimumSystemVersion>
```
duncdrum, line 927: I m not sure if this is true. I think we should bump this to the current long term release of macOS at some point before `7.0.0`

When was the last 10.9 supporting mac released? I have no clue.  

Not a blocker.

exist-distribution/pom.xml
```diff
@@ -925,6 +925,19 @@
                             <executableName>eXist-JavaAppLauncher</executableName>
                             <jvmRequired>21</jvmRequired>
```
duncdrum, line 926: I m curious if `21+` will allow different jvm paths to be discoverable. 

In my local testing I struggled with having the dmg recognize system Java outside of what ever apple deemed the canonical location. Which means that Zulu temurin et al can be missed based on path name despite macOS happily using them in other contexts. 

Obviously exist needs 21, so there would be a trade off.

exist-distribution/pom.xml
```diff
@@ -925,6 +925,19 @@
...
+                            <!-- Populate LSArchitecturePriority so macOS LaunchServices
+                                 will launch the .app. The plugin emits an empty
+                                 LSArchitecturePriority array if this is unset, which
+                                 LaunchServices reads as "no architectures supported"
+                                 and refuses to launch (silent failure in the Dock,
+                                 surfaced in Console.app as:
+                                 "LAUNCH: No architectures specified in launch architectures [ NULL ]").
+                                 The eXist-JavaAppLauncher native binary is built as
+                                 a universal binary carrying x86_64 + arm64 slices. -->
+                            <architectures>
```
duncdrum, line 937: There is a bunch of related error messages on the release workflow, when signing `i386` binaries. They didn't fail the build or notarization, but it makes me wonder if these binaries should be in the dmg to start with.

duncdrum (approved):

windauer (approved):

exist-installer/src/main/izpack/install.xml
joewiz: [This response was co-authored with Claude Code. -Joe]

The `@{...}` form is IzPack's Maven-filtering convention — `maven-resources-plugin` substitutes `@{maven.compiler.release}` with `21` at build time, before IzPack ever sees the file. The reason for `@{...}` rather than the more familiar `${...}` is that IzPack itself uses `${...}` at *install* time for its own runtime variables (e.g. `${INSTALL_PATH}`, `${SYSTEM[java.io.tmpdir]}`), so the build-time vs install-time substitutions need different delimiters. The same convention is already used 14× elsewhere in this file — e.g. line 36 `<appversion>@{izpack.installation.info.appversion}</appversion>` and, most relevantly, line 41 `<javaversion>@{maven.compiler.release}</javaversion>`, which gates the installer itself on JDK 21. I copied that.

exist-distribution/pom.xml
joewiz: [This response was co-authored with Claude Code. -Joe]

Agreed — I left the `10.9` alone because it predates this branch and a bump touches a different scoping conversation (which macOS major do we commit to supporting, and that should probably match what the bundled JDK actually runs on). For reference, Mavericks (10.9) was 2013; Java 21 supports macOS 11 (Big Sur) and later. So a defensible bump would be `11.0`. Happy to file a follow-up issue for that and any related notarization-impact testing — flag if you'd like the issue opened now or after this PR lands.

exist-distribution/pom.xml
joewiz: [This response was co-authored with Claude Code. -Joe]

Good question and I think `21+` is the right shape — the `+` suffix is appbundler convention for "this major or later" and matches what users actually have installed (Zulu, Temurin, Liberica, etc. all ≥ 21). The strict `21` reading would also reject 22, 23, … which is probably not what we want. I left this line as-is on this PR because it's pre-existing config and I wanted to keep scope tight on the launch regressions, but the discoverability problem you hit locally is real and worth chasing in a follow-up. I'll prep a small PR that flips it to `21+` and we can verify against Zulu / Temurin / system-default in CI before it lands.

exist-distribution/pom.xml
joewiz: [This response was co-authored with Claude Code. -Joe]

Confirmed and good catch — the i386 slices are inside `jline`'s bundled native lib (`org/jline/nativ/Mac/x86/libjlinenative.jnilib` is a thin Mach-O i386 binary that ships inside the jline JAR). macOS dropped i386 support in 10.15 (Catalina, 2019), so those slices haven't been runnable in the supported universe for years; signing warnings are cosmetic but they do bloat the DMG and add noise to the notarization log. Two cleanup paths: (a) bump jline if upstream has dropped the i386 slice, (b) strip it ourselves with `lipo -remove i386` during the DMG-prep step. I'll file a follow-up issue with both options so we can pick one without holding up this PR.

PR #6448, opened 2026-06-06
[feature] Add REST _expand-xincludes serialization parameter

joewiz:

## Summary

Adds an `_expand-xincludes` serialization parameter to eXist's REST interface. A GET of a stored XML document previously always expanded `<xi:include>` elements during serialization, with no way to turn that off — so an open → edit → save round-trip through REST would silently expand and **destroy the include**. The REST interface already exposes the other serialization controls as `_`-prefixed query-string parameters (`_indent`, `_omit-xml-declaration`, `_output-doctype`, `_encoding`, …); `expand-xincludes` was simply missing.

This is PR 2 of 3 for uniform serialization-extension control (core #6447 → REST → RESTXQ). It unblocks existdb-openapi's REST binary-streaming read path; the companion core PR (#6447) covers `fn:serialize`, `file:serialize`, and the `output:` prolog.

## What changed

- `RESTServerParameter`: new `Expand_Xincludes` enum constant → query-string key `_expand-xincludes` (the existing `xmlKey()`/`queryStringKey()` derivation produces this automatically).
- `RESTServer.doGet`: when `_expand-xincludes` is present, set `EXistOutputKeys.EXPAND_XINCLUDES` on the output properties used to serialize the resource. When the parameter is absent, the serializer's existing default (expand) applies — so existing requests are byte-for-byte unchanged.

It composes with the other serialization parameters; for example `?_indent=no&_omit-xml-declaration=yes&_expand-xincludes=no` honors all three together.

## Backward compatibility

No behavior change unless the new parameter is supplied. With no `_expand-xincludes`, output is identical to before (XIncludes expand by default), verified by a default-behavior test.

## Test plan

- [x] `RESTServiceTest` against a stored XInclude document (embedded server):
  - default GET expands the include (unchanged behavior);
  - `?_expand-xincludes=no` preserves `<xi:include>` and does not expand;
  - `?_expand-xincludes=yes` expands;
  - `?_indent=no&_omit-xml-declaration=yes&_expand-xincludes=no` honors all three (include preserved, no XML declaration).
- [x] PMD/Codacy clean on the changed lines (the file's other PMD findings are pre-existing in unrelated methods).

## Spec references

- W3C XSLT and XQuery Serialization 3.1 — implementation-defined serialization parameters
- BaseX serialization (REST query parameters): https://docs.basex.org/13/Serialization

### DIFF
--- a/exist-core/src/main/java/org/exist/http/RESTServer.java
+++ b/exist-core/src/main/java/org/exist/http/RESTServer.java
@@ -404,6 +404,11 @@ public class RESTServer {
             final String omitOriginalXmlDeclaration = broker.getConfiguration().getProperty(Serializer.PROPERTY_OMIT_ORIGINAL_XML_DECLARATION, "no");
             outputProperties.setProperty(EXistOutputKeys.OMIT_ORIGINAL_XML_DECLARATION, omitOriginalXmlDeclaration);
         }
+        if ((option = getParameter(request, Expand_Xincludes)) != null) {
+            // take user query-string specified expand-xincludes setting; when absent the
+            // serializer default (expand) applies, so existing requests are unchanged
+            outputProperties.setProperty(EXistOutputKeys.EXPAND_XINCLUDES, option);
+        }
         if ((option = getParameter(request, Source)) != null && !safeMode) {
             source = "yes".equals(option);
         }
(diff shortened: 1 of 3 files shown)


joewiz:

Closing alongside #6447. With existdb-openapi now setting `expand-xincludes` via eXist's `exist:` namespace in its own serialization (rather than relying on a REST passthrough), there is no current consumer for a REST `_expand-xincludes` parameter, and the related client work confirmed no other client needs one. To be clear, this is a "no consumer yet" close, not a conformance issue — REST query parameters are eXist's own convention and aren't governed by the W3C serialization-namespace rule — so if a REST client later wants `expand-xincludes` control we can revisit.

Issue #5299, opened 2024-05-08
xmldb:rename() function fails if the collection name contains spaces


daliboris:
**Collection with spaces in name cannot be renamed**

If the collection name contains spaces, calling function `xmldb:rename($source-collection-uri as xs:string, $new-collection-name as xs:string)` fails with the following error:

```log
exerr:ERROR Could not locate collection: /db/test/data%20with%20spaces [at line 1, column 1, source: String/2077783963999039558]
```

**Expected behavior**

Collection with spaces should be renamed.

The `xmldb:rename()` function with 4 parametrs works if the renamed resource contains spaces, for instance

```XQuery
xmldb:rename('/db/test/', replace('xml with spaces.xml', ' ', '%20'), 'clean-xml.xml')
```

**To Reproduce**

```Xquery
xquery version "3.1";

module namespace t="http://exist-db.org/xquery/test";

declare namespace test="http://exist-db.org/xquery/xqsuite";


declare
    %test:setUp
function t:setup() {
    let $testCol := xmldb:create-collection("/db", "test")
    let $spacesCol := xmldb:create-collection("/db/test", "data%20with%20spaces")
    return ()
};

declare
    %test:tearDown
function t:tearDown() {
    xmldb:remove("/db/test")
};


declare
    %test:assertEquals('data%20with%20spaces')
function t:collection-with-spaces-exists() {
    let $result := xmldb:get-child-collections('/db/test/')
    return $result
};

declare
    %test:assertTrue
function t:collection-exists() {
    let $result := xmldb:get-child-collections('/db/test/')
    return
       count($result) eq 1
};

declare
%test:assertError
function t:rename-collection-throws-error() {
    xmldb:rename('/db/test/data%20with%20spaces', '/db/test/data-with-dashes')
};
```

**Context**

- Build: eXist-6.2.0 (c8fa4958b6d4a50bd0cba7f3e76a150226414187)
- Java: 1.8.0_332 (Oracle Corporation)
- OS: Linux 5.15.146.1-microsoft-standard-WSL2 (amd64)

daliboris:
**Workaround**

- create a new collection
- copy resources from the old collection to the new one
- remove the old collection

```XQuery
let $main-colection := '/db/test'
let $old-collection-name := 'data with spaces'
let $new-collection-name := 'data-with-dashes'

let $old-collection := $main-colection || replace($old-collection-name, ' ', '%20')
let $new-collection := $main-colection || $new-collection-name 

[...]

Issue #3795, opened 2021-03-31
Getting URI Encoding, paths, filenames, etc. right


adamretter:
This issue is about more than just URI encoding it is about naming things and defining concepts. This stuff is fundamental.
It will likely also fix a bunch of issues that have been opened about URI encoding and non-ascii character encoding bugs, but it is not a bug report in and of itself.

I will use `xmldb:create-collection` only to demonstrate problems...

Currently:
```xquery
xmldb:create-collection("/db", "hello world")
```

Raises the error:
```java
org.xmldb.api.base.XMLDBException: err:FORG0001 failed to convert adam retter into an XmldbURI: Illegal character in path at index 5: hello world [at line 1, column 1, source: String/2869744519086207067]
```

However, the signature of `xmldb:create-collection` is:
```xquery
xmldb:create-collection($target-collection-uri as xs:string, $new-collection as xs:string) as xs:string?
```

$new-collection is documented as: `The name of the new collection to create`. 

A "Collection" is technically **ONLY** a mapping from an `xs:string` (interpreted as an `xs:anyURI`) to a `item()*`.
https://www.w3.org/TR/xpath-functions-31/#func-collection

TBC---

These concepts in the context of eXist-db need to be properly defined:

* Collection Name
* Document Name
* Path
* File name
* URI as an xs:string
* URI as a xs:anyURI

adamretter:
If we were to transition to Collection Names and Document Names being names, then eXist-db needs some fixes. At the moment it unnecessarily stores just the Collection Name (not the Collection URI) as an XmldbURI.

eduarddrenth:
This may also be related to this issue:
```
#10 1571.6 Caused by: java.net.URISyntaxException: Malformed escape pair at index 5: 80-20%-regeling.xml
#10 1571.6      at java.net.URI$Parser.fail(URI.java:2913) ~[?:?]
#10 1571.6      at java.net.URI$Parser.scanEscape(URI.java:3041) ~[?:?]
#10 1571.6      at java.net.URI$Parser.scan(URI.java:3064) ~[?:?]
#10 1571.6      at java.net.URI$Parser.checkChars(URI.java:3082) ~[?:?]
#10 1571.6      at java.net.URI$Parser.parseHierarchical(URI.java:3166) ~[?:?]
#10 1571.6      at java.net.URI$Parser.parse(URI.java:3125) ~[?:?]
#10 1571.6      at java.net.URI.<init>(URI.java:600) ~[?:?]
[...]

Issue #3665, opened 2020-12-08
Awkward chars (non-ascii non-reserved) do not map between rest and webdav


hungerburg:
There was discussion on the list; I compiled some lines curl usage, that show one issue - that names of resources are treated differently between rest and webdav interfaces. This is more a method to help diagnosing than a diagnosis. Yet, users reliably want the mapping to be true.

Notes:
- Letter "м" is a cyrillic character, utf-8 %D0%BC, a small "m"
- only an exist-db installation required, no extra files
- two resources by the same name will be created
- each servlet can only fetch its own creation
- tested with snapshot of today's

```
# Webdav create a collection and add a resource
curl -v -u admin -X DELETE http://localhost:8080/exist/webdav/db/testuni
curl -v -u admin -X MKCOL http://localhost:8080/exist/webdav/db/testuni
curl -v -u admin http://localhost:8080/exist/webdav/db/testuni
curl -v -u admin -X PUT -H 'Content-Type: application/xml' -d '<text>WDAV Letter м.</text>' http://localhost:8080/exist/webdav/db/testuni/м.xml
curl -v -u admin http://localhost:8080/exist/webdav/db/testuni
curl -v -u admin http://localhost:8080/exist/webdav/db/testuni/%D0%BC.xml
# <text>WDAV Letter м.</text>
curl -v -u admin http://localhost:8080/exist/webdav/db/testuni/м.xml
# <text>WDAV Letter м.</text>

# Fetch with Rest
curl -v http://localhost:8080/exist/rest/db/testuni
curl -v http://localhost:8080/exist/rest/db/testuni/м.xml
# <title>Error 404 Document /db/testuni/?.xml not found</title>
# <tr><th>URI:</th><td>/exist/rest/db/testuni/?.xml</td></tr>
curl -v http://localhost:8080/exist/rest/db/testuni/%D0%BC.xml
# <body><h2>HTTP ERROR 404 Document /db/testuni/?.xml not found</h2>
# <tr><th>URI:</th><td>/exist/rest/db/testuni/%D0%BC.xml</td></tr>

# Rest create resource
curl -v -u admin -X PUT -H 'Content-Type: application/xml' -d '<text>REST Letter м.</text>' http://localhost:8080/exist/rest/db/testuni/м.xml
curl -v http://localhost:8080/exist/rest/db/testuni
curl -v http://localhost:8080/exist/rest/db/testuni/м.xml
> <text>REST Letter м.</text>
[...]

Issue #1824, opened 2018-04-14
Bug in decoding of "+" in xmldb URI-related functions


joewiz:
### What is the problem

The `+` character is a reserved character according to RFC3986, and as such should be "protected from normalization." eXist's xmldb module's functions that deal with database URIs incorrectly normalize this character, treating it as if it were a percent-encoded octet, i.e., treating the `+` character in an xmldb URI as a character that should be decoded and turned into a space character.

This causes a problem in XQuery applications like eXide where the application must be able to reliably display the URI-decoded form of a resource (its "name"), while also reliably addressing its URI (as a "key") for resource operations like delete and rename. 

For example, given a resource in the database created via `xmldb:store("/db", "A+B.xml", <foo/>)`, we can successfully address this document via `doc("/db/A+B.xml")`, but `xmldb:decode("A+B.xml")` returns `A B.xml` instead of `A+B.xml`. 

While this issue is about handling of the `+` character, it is also about a larger issue: users have reported unpredictable behavior involving characters in resource names, and would benefit from a clear and straightforward description of eXist's xmldb *URI scheme*, explaining how characters in resource names are treated, and placed in the eXist documentation. This statement should cite any relevant specs/standards for URI encoding that eXist adheres to, and where (if at all) eXist diverges from these standards.

Clarity about eXist's xmldb URI scheme would also let us achieve greater consistency in eXist's various interfaces and built-in apps/tools. For example, while using WebDAV clients (e.g., oXygen and Transmit) to create collections like `tést` or `你好` causes the resulting collection list to display `tést` and `你好`, doing the same in the Java admin client causes the collection list to display `t%C3%A9st` and `%E4%BD%A0%E5%A5%BD`. This suggests that the Java admin client is performing the correct *encoding* of the resource name into a URI when it is stored/created, but it is failing to *decode* the resource URI back into its name when displaying it.

But for now, let's focus on the treatment of `+` in eXist's xmldb URI-related functions.

### What did you expect

According to the function documentation for `xmldb:decode` and `xmldb:decode-uri` (see http://exist-db.org/exist/apps/fundocs/view.html?uri=http://exist-db.org/xquery/xmldb#decode.1), these functions do the following to the supplied input:

> any percent encoded octets will be translated to their decoded UTF-8 representation

Since `A+B` contains no percent-encoded octets, calling these decode functions should return the original value, `A+B`. Instead, eXist returns `A B`.

[...]

PR #6450, opened 2026-06-07
[test] Add cross-surface resource-naming conformance ratchet

joewiz:

## Summary

Adds a test that pins how eXist handles "awkward" resource names **across API surfaces** — the missing coverage for the resource-naming-stability work (issues #3795, #3665, #1824, #5299, #1612). It changes no behavior. It prints the current behavior as a matrix and enforces a **ratchet**: the set of names that fail to round-trip cross-surface must exactly equal a documented `KNOWN_FAILURES` allowlist.

The effect: **the moment this merges, every name that already works is guarded against regression**, and as each naming fix lands you remove an entry from the allowlist to lock the improvement in. The test can neither silently regress (an unlisted name that breaks fails the build) nor silently drift (a listed name that starts working fails the build and tells you to remove it).

There is no cross-surface special-character test in the suite today; this is the first.

## What it does

For a corpus of awkward leaf names (space, `+`, literal `%`, literal `%20`, `#`, `@`, `&`, parentheses, apostrophe, `café`, Cyrillic, CJK), per name it:

1. stores it **via WebDAV** (raw HTTP `PUT`, so the on-the-wire encoding is explicit);
2. reports **what name actually landed in storage**, via eXist's native REST collection listing;
3. reads the content back **by the requested name** via WebDAV and via REST;
4. prints the matrix, then asserts the ratchet (failing-set == `KNOWN_FAILURES`).

Design notes:
- **No XQuery-module dependency** — the WebDAV module's test `conf.xml` registers no builtin-modules, so the harness uses the REST collection listing rather than `xmldb:*`.
- **No WebDAV-client dependency** — the collection is created with a REST `PUT` (auto-create) and every probe is raw HTTP.
- Probe content is valid XML because the corpus names end in `.xml`, which eXist parses on store.

## What the matrix shows today

```
probe            requested          stored-as (oracle)        dav-read  rest-read
ascii-control    plain.xml          plain.xml                 PASS      PASS
space            with space.xml     with%20space.xml (≠)      PASS      PASS
plus             a+b.xml            a%2Bb.xml (≠)             PASS      FAIL
literal-percent  a%b.xml            a%25b.xml (≠)             PASS      PASS
encoded-space    a%20b.xml          a%2520b.xml (≠)           PASS      PASS
hash             a#b.xml            a%23b.xml (≠)             PASS      PASS
at               a@b.xml            a%40b.xml (≠)             PASS      FAIL
ampersand        a&b.xml            a%26b.xml (≠)             PASS      FAIL
parens           report(2024).xml   report%282024%29.xml (≠)  PASS      FAIL
apostrophe       o'brien.xml        o%27brien.xml (≠)         PASS      FAIL
non-ascii        café.xml           caf%C3%A9.xml (≠)         PASS      PASS
cyrillic         Привет.xml         %D0%9F…%82.xml (≠)        PASS      PASS
cjk              文書.xml           %E6%96%87%E6%9B%B8.xml (≠) PASS      PASS
```

So today: every name with a non-unreserved character is **stored percent-encoded**; WebDAV self-reads always round-trip; reading by the requested name via **REST fails specifically for sub-delim/reserved characters** (`+`, `@`, `&`, `(`/`)`, `'`). Those five are the current `KNOWN_FAILURES` allowlist.

## What devs see in CI

- **Normal (green):** the matrix prints into the build log; `Tests run: 1, Failures: 0`.
- **A naming fix lands** (a listed name now round-trips) → build fails with: *"N name(s) now round-trip cross-surface but are still listed as known failures. Remove them from KNOWN_FAILURES so they become regression-guarded: …"* + the matrix. You delete those lines from the allowlist and they become guarded.
- **A regression** (an unlisted name breaks) → build fails with: *"N name(s) regressed — expected to round-trip cross-surface but failed: …"* + the matrix.

## Scope / follow-ups (noted in the class Javadoc)

- XML-RPC surface (create/read against `/xmlrpc`).
- The full N×N cross-surface matrix (create-via-X then read-via-every-Y).
- existdb-openapi's special-character suite lives in that repo (a separate PR).
- This first increment treats "stored-as ≠ requested" as informational (the deeper, format-level concern handled later in the tasking); the ratchet enforces round-trip-by-requested-name.

## Where it lives

In the `extensions/webdav` test module — the only place that can drive REST **and** WebDAV **and** XML-RPC against one database in one test (module dependency direction means exist-core can't reach the WebDAV extension). A dedicated cross-surface integration-test module would be the alternative if preferred.

## Test plan

- [x] Runs green locally (`Tests run: 1, Failures: 0`); matrix prints to the build log.
- [x] Ratchet verified in both directions — a simulated regression and a simulated now-fixed entry each fail with the correct message + matrix.
- [x] Codacy/PMD clean on the new file.
- [x] No production code changed; no other test affected (the test log4j2 config used during diagnosis was reverted).

## Related

- Resource-naming-stability tasking (PR A) — this is the ratchet the rest of that plan tightens.
- Issues: #3795 (meta), #3665 (REST↔WebDAV), #1824 (`+` decode), #5299 (spaces), #1612 (non-ASCII).

### DIFF
--- a/exist-core/src/test/java/org/exist/backup/BackupRestoreNamingConformanceTest.java
+++ b/exist-core/src/test/java/org/exist/backup/BackupRestoreNamingConformanceTest.java
@@ -1,0 +1,271 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.backup;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+import java.io.InputStream;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.TreeMap;
+
+import org.exist.TestUtils;
+import org.exist.backup.restore.listener.LogRestoreListener;
+import org.exist.backup.restore.listener.RestoreListener;
+import org.exist.collections.Collection;
+import org.exist.dom.persistent.BinaryDocument;
+import org.exist.dom.persistent.DocumentImpl;
+import org.exist.storage.BrokerPool;
+import org.exist.storage.DBBroker;
+import org.exist.storage.serializers.Serializer;
+import org.exist.storage.txn.Txn;
+import org.exist.test.ExistEmbeddedServer;
+import org.exist.util.MimeType;
+import org.exist.util.StringInputSource;
+import org.exist.util.io.InputStreamUtil;
+import org.exist.xmldb.XmldbURI;
+import org.exist.xquery.util.URIUtils;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+
+/**
+ * Backup &rarr; restore round-trip conformance for awkward resource names.
+ *
+ * <p>Extends the resource-naming conformance net (see {@code ResourceNamingConformanceTest} in the webdav
+ * module, and issue #6463) to the backup/restore path @line-o asked about on PR #6508: for a corpus of
+ * awkward leaf names (spaces, sub-delimiters, non-ASCII, CJK, Cyrillic, and the literal-{@code %} cases
+ * Decision 2 flags), it stores each resource, snapshots the exact stored key and content, exports a full
+ * backup, wipes the collection, restores, and asserts the snapshot is reproduced <em>byte-for-byte</em> —
+ * no resource lost, renamed, or collided by the round-trip. That is the guard against "even theoretical
+ * data loss" through backup/restore.</p>
+ *
+ * <p><b>Scope.</b> This guards the <em>restore</em> path. Resources are stored under the same leaf keys a
+ * WebDAV/REST store lands them under (verified in the printed mapping: {@code café.xml -> caf%C3%A9.xml},
+ * {@code a+b.xml -> a%2Bb.xml}, {@code with space.xml -> with%20space.xml}, …), via eXist's own
+ * {@code encodeXmldbUriFor}. Two distinct names collapsing onto one key <em>on the way in</em> — the
+ * Decision 2 store-time collision — rides a different store path (the persistent layer's non-escaping
+ * encoding) and is out of scope here; this encoder escapes a literal {@code %} to {@code %25}, so the
+ * corpus stores injectively and the {@code before}-snapshot size check is a setup-integrity guard, not a
+ * claim about that other path.</p>
+ *
+ * <p>Parameterized over {@link SystemExport}'s direct / non-direct modes and plain / zip, mirroring
+ * {@link SystemExportImportTest}.</p>
+ */
+@RunWith(Parameterized.class)
+public class BackupRestoreNamingConformanceTest {
+
+    @ClassRule
+    public static final TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+    @ClassRule
+    public static final ExistEmbeddedServer existEmbeddedServer = new ExistEmbeddedServer(true, true);
+
+    private static final XmldbURI TEST_COLLECTION = XmldbURI.create("/db/backup-naming-conformance");
+    private static final String MARKER = "backup-naming-probe";
+
+    /** Awkward human-intended leaf names. {@code /} is excluded (path separator); the literal-{@code %} cases are included. */
+    private static final List<String> XML_NAMES = List.of(
+            "plain.xml",            // control: must always survive
+            "with space.xml",
+            "a+b.xml",
+            "a%b.xml",              // literal percent (not a valid %XX escape)
+            "a%20b.xml",            // the literal text "%20", not a space
+            "a@b.xml",
+            "a&b.xml",
+            "report(2024).xml",
+            "o'brien.xml",
+            "café.xml",
+            "Привет.xml",
+            "文書.xml");
+
+    private static final String BINARY_NAME = "résumé.bin";
+    private static final byte[] BINARY_CONTENT = "binary résumé payload".getBytes(UTF_8);
+
+    @Parameter
+    public String apiName;
+
+    @Parameter(value = 1)
+    public boolean direct;
+
+    @Parameter(value = 2)
+    public boolean zip;
+
+    @Parameters(name = "{0} zip:{2}")
+    public static java.util.Collection<Object[]> data() {
+        return Arrays.asList(new Object[][]{
+                {"direct", true, false},
+                {"non-direct", false, false},
+                {"direct", true, true},
+                {"non-direct", false, true}
+        });
+    }
+
+    /** Unique XML content per resource, so the byte-for-byte comparison is meaningful and any collision shows. */
+    private static String contentFor(final String humanName) {
+        final String escaped = humanName
+                .replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;").replace("\"", "&quot;");
+        return "<probe name=\"" + escaped + "\">" + MARKER + "</probe>";
+    }
+
+    private static XmldbURI storedKeyFor(final String humanName) throws Exception {
+        // eXist's own encoder: the leaf key a WebDAV/REST store would land this name under.
+        return URIUtils.encodeXmldbUriFor(TEST_COLLECTION.toString() + "/" + humanName).lastSegment();
+    }
+
+    @Before
+    public void storeCorpus() throws Exception {
+        final BrokerPool pool = existEmbeddedServer.getBrokerPool();
+        try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()));
+             final Txn txn = pool.getTransactionManager().beginTransaction()) {
+
+            final Collection existing = broker.getCollection(TEST_COLLECTION);
+            if (existing != null) {
+                broker.removeCollection(txn, existing);
+            }
+            final Collection test = broker.getOrCreateCollection(txn, TEST_COLLECTION);
+            broker.saveCollection(txn, test);
+
+            for (final String name : XML_NAMES) {
+                broker.storeDocument(txn, storedKeyFor(name), new StringInputSource(contentFor(name)), MimeType.XML_TYPE, test);
+            }
+            broker.storeDocument(txn, storedKeyFor(BINARY_NAME), new StringInputSource(BINARY_CONTENT), MimeType.BINARY_TYPE, test);
+
+            txn.commit();
+        }
+    }
+
+    @After
+    public void removeCorpus() throws Exception {
+        final BrokerPool pool = existEmbeddedServer.getBrokerPool();
+        try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()));
+             final Txn txn = pool.getTransactionManager().beginTransaction()) {
+            final Collection test = broker.getCollection(TEST_COLLECTION);
+            if (test != null) {
+                broker.removeCollection(txn, test);
+            }
+            txn.commit();
+        }
+    }
+
+    @Test
+    public void awkwardNamesSurviveBackupRestore() throws Exception {
+        final BrokerPool pool = existEmbeddedServer.getBrokerPool();
+
+        // snapshot the stored (key -> content) state BEFORE the round-trip
+        final Map<String, String> before = snapshot(pool);
+        printMapping("before backup", before);
+
+        // setup-integrity: every corpus name must have stored as its own distinct key in this path, so the
+        // round-trip below genuinely exercises all of them (not fewer after an accidental store-time clash).
+        assertEquals("setup stored fewer resources than the corpus — a name clashed on the way in; see the "
+                        + "printed mapping", XML_NAMES.size() + 1, before.size());
+
+        // export a full backup
+        final Path backup;
+        try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()));
+             final Txn txn = pool.getTransactionManager().beginTransaction()) {
+            final SystemExport export = new SystemExport(broker, txn, null, null, direct);
+            backup = export.export(temporaryFolder.newFolder().getAbsolutePath(), false, zip, null);
+            txn.commit();
+        }
+
+        // wipe the collection
+        try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()));
+             final Txn txn = pool.getTransactionManager().beginTransaction()) {
+            final Collection test = broker.getCollection(TEST_COLLECTION);
+            assertNotNull(test);
+            broker.removeCollection(txn, test);
+            txn.commit();
+        }
+
+        // restore from the backup
+        final RestoreListener listener = new LogRestoreListener();
+        new SystemImport(pool).restore(TestUtils.ADMIN_DB_USER, TestUtils.ADMIN_DB_PWD, null, backup, listener);
+
+        // snapshot AFTER the round-trip
+        final Map<String, String> after = snapshot(pool);
+        printMapping("after restore", after);
+
+        // the round-trip must be faithful: the same stored keys, each with byte-for-byte identical content,
+        // none dropped, renamed, or collided.
+        assertEquals("backup -> restore changed the set of stored resource names", before.keySet(), after.keySet());
+        assertEquals("backup -> restore did not reproduce every resource's content byte-for-byte", before, after);
+    }
+
+    /** stored-key -&gt; content for every document under the test collection (XML serialized; binary as a UTF-8 string). */
+    private Map<String, String> snapshot(final BrokerPool pool) throws Exception {
+        final Map<String, String> out = new TreeMap<>();
+        try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()));
+             final Txn txn = pool.getTransactionManager().beginTransaction()) {
+            final Collection test = broker.getCollection(TEST_COLLECTION);
+            assertNotNull("test collection missing", test);
+            for (final Iterator<DocumentImpl> it = test.iterator(broker); it.hasNext(); ) {
+                final DocumentImpl doc = it.next();
+                final String key = doc.getFileURI().toString();       // the exact stored key
+                final String content;
+                if (doc instanceof BinaryDocument bin) {
+                    try (final InputStream is = broker.getBinaryResource(txn, bin)) {
+                        content = "binary:" + InputStreamUtil.readString(is, UTF_8);
+                    }
+                } else {
+                    content = serialize(broker, doc);
+                }
+                out.put(key, content);
+            }
+            txn.commit();
+        }
+        return out;
+    }
+
+    private String serialize(final DBBroker broker, final DocumentImpl doc) throws Exception {
+        final Serializer serializer = broker.borrowSerializer();
+        try {
+            serializer.setUser(broker.getCurrentSubject());
+            return serializer.serialize(doc);
+        } finally {
+            broker.returnSerializer(serializer);
+        }
+    }
+
+    private void printMapping(final String phase, final Map<String, String> snapshot) {
+        final StringBuilder sb = new StringBuilder("\n=== Backup/restore naming conformance (")
+                .append(apiName).append(", zip:").append(zip).append(") — ").append(phase).append(" ===\n");
+        for (final String key : snapshot.keySet()) {
+            sb.append("    ").append(key).append('\n');
+        }
+        sb.append("  ").append(snapshot.size()).append(" stored resource(s)\n");
+        System.out.println(sb);
+    }
+}
--- a/exist-core/src/test/java/org/exist/xmldb/XQueryNamingConformanceTest.java
+++ b/exist-core/src/test/java/org/exist/xmldb/XQueryNamingConformanceTest.java
@@ -1,0 +1,220 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xmldb;
+
+import static org.junit.Assert.fail;
+
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+import org.exist.storage.BrokerPool;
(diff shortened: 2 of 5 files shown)


extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
reinhapa, line 1: Composing the webdav `URL` using the `URI` constructor would make the custom encoding logic obsolete if used by a function like this:

Locations as such

```
final URL url = URI.create(webdavBase() + encodePathSegments(dbPath)).toURL();
```

Could then replaced with

```
final URL url = createWebdavUrl(dbPath);
```

Using this method:
```
private URL createWebdavUrl(String dbPath {
  return new URI("http", null, "localhost", existWebServer.getPort(), "/webdav"+ dbPath, null, null).toURL();
}
```

Also I would prefer to use the new [HttpClient API](https://docs.oracle.com/en/java/javase/21/docs/api/java.net.http/java/net/http/HttpClient.html) introduced with Java 11

reinhapa (changes requested):

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
```diff
@@ -88,6 +89,8 @@ public class ResourceNamingConformanceTest {
     private static final String MARKER = "naming-probe-content";
     private static final String CONTENT = "<probe>" + MARKER + "</probe>";
 
+    private static final HttpClient HTTP = HttpClient.newHttpClient();
```
reinhapa: Create this within the `createTestCollection()` method as the client needs to be closed at the end of the test. (Within an `@AfterClass` annotated method.

reinhapa (changes requested):

reinhapa (approved):

joewiz:

Thanks Patrick — adopted both: the harness now uses `java.net.http.HttpClient`, and request URLs are built with the multi-argument `java.net.URI` constructor instead of the custom path-segment encoder, via a small `endpointUri(endpoint, dbPath)` helper. That dropped ~40 lines. And per your follow-up, the client is now created in `createTestCollection()` (`@BeforeClass`) and closed in `removeTestCollection()` (`@AfterClass`) — `HttpClient` is `AutoCloseable` as of Java 21.

Your suggestion turned out to do more than simplify the code — it corrected the test. The `URI` constructor leaves RFC 3986 sub-delimiters (`+ @ & ( ) '`) literal in the path, whereas the old `URLEncoder`-based helper percent-encoded them. With the conventional encoding (what a browser or `curl` actually sends), **every** corpus name round-trips cross-surface between WebDAV and REST — including the five the earlier revision had recorded as known failures:

```
probe         requested          stored-as              dav-read  rest-read
plus          a+b.xml            a+b.xml                PASS      PASS
at            a@b.xml            a@b.xml                PASS      PASS
ampersand     a&b.xml            a&b.xml                PASS      PASS
parens        report(2024).xml   report(2024).xml       PASS      PASS
apostrophe    o'brien.xml        o'brien.xml            PASS      PASS
space         with space.xml     with%20space.xml (≠)   PASS      PASS
café          café.xml           caf%C3%A9.xml (≠)       PASS      PASS
...all 13 PASS/PASS
```

So those five "failures" were artifacts of the harness percent-encoding sub-delimiters, not eXist mis-storing names. eXist may store a percent-encoded form (the `(≠)` rows), but both surfaces agree on read, so the content is reachable by the requested name. `KNOWN_FAILURES` is now **empty**, and the ratchet guards the all-green state: it fails if any corpus name ever stops round-tripping cross-surface.

(One genuinely interesting residual the percent-encoded run hinted at: `a%2Bb` and `a+b` denote the same path resource per RFC 3986, but eXist's REST read didn't treat them as equivalent. That's a much narrower question than "names don't round-trip" and I've left it out of this test; worth a separate look as part of the naming work.)

Pushed as `9ae2d8a` (refactor + the empty allowlist) and `5851167` (HttpClient lifecycle).

duncdrum (commented):
I might be projecting past traumas, but should we run this as IT test, so ci runs for the full OS matrix. I remember my favourite OS from Redmond often doing its own thing with respect to Unicode, file names, etc. ?

joewiz:
@duncdrum Fully agree. Claude confirmed the test as is runs only on Linux. Reworking now to be IT so we get results from macOS and Windows. Stay tuned.

joewiz:

Good instinct — Windows/Unicode filename behavior is exactly the kind of thing that bites. I dug into how this would actually play out, and it turns out the answer is a bit surprising:

**Renaming it to `*IT` alone would not get it onto the OS matrix in this module — it would stop it running entirely.** The `extensions/webdav` module declares no failsafe plugin, and surefire's default includes don't match `*IT`, so an `*IT` here runs under *neither* runner. The module's existing `WindowsPathResourceIT` demonstrates this: it doesn't execute in a `mvn verify` of the module (it's effectively a dead test today).

**And the harness already gets cross-OS coverage as a plain `*Test`.** The webdav module's tests run under `mvn verify` (I confirmed `verify -DskipUnitTests=true` executes them locally — webdav doesn't gate its surefire on `skipUnitTests`; only exist-core/lucene do that). The CI **integration** job is exactly `verify -DskipUnitTests=true`, and the matrix runs that job on ubuntu + windows + macOS. So this test is exercised on all three OSes via the integration job — confirmed locally on macOS, where `mvn verify -pl extensions/webdav -DskipUnitTests=true -Dtest=ResourceNamingConformanceTest` runs it (Tests run: 1, not skipped).

**Empirically, the OS axis is invariant here.** I ran the harness on macOS: every name in the corpus round-trips and `KNOWN_FAILURES` stays empty, identical to Linux — including the non-ASCII cases (`café.xml`, the Cyrillic and CJK names) and the `literal-percent` / `encoded-space` probes. The reason is structural: eXist stores resource names in its own database in a percent-encoded form computed in Java via `URIUtils` (`café.xml` → `caf%C3%A9.xml`, Cyrillic/CJK likewise), **not** as operating-system files — so the host filesystem's reserved-character, Unicode, and normalization rules never reach these names. (Java 18+ also defaults `file.encoding=UTF-8` on every OS, removing the old Windows cp1252 trap.)

So: no change needed for OS coverage, and the `*Test` name is correct for this module. If we'd rather make the cross-OS coverage **explicit and guaranteed** rather than incidental, the proper fix is a small module-infra change — wire the failsafe plugin into `extensions/webdav` so `*IT` classes actually run there — which would also revive the currently-dead `WindowsPathResourceIT`. Happy to do that as a separate follow-up if you think it's worth hardening; it's orthogonal to this PR.

joewiz:

Two increments have landed on this branch since the original review, both extending the same awkward-name corpus. Rebased onto current develop (the `extensions/webdav/pom.xml` conflict was develop's milton→`java.net.http` test migration removing the deps I'd added `xmlrpc-client` beside — resolved by keeping develop's removal and re-adding just `xmlrpc-client`).

## 1. Backup → restore round-trip (`exist-core`)

`org.exist.backup.BackupRestoreNamingConformanceTest` answers a question @line-o raised on #6508: does backup/restore preserve awkward names? It stores the corpus, snapshots each resource's exact stored key and content, exports a full backup, wipes the collection, restores, and asserts the snapshot is reproduced **byte-for-byte** — parameterized over `SystemExport`'s direct/non-direct and plain/zip modes.

**Result:** the corpus round-trips faithfully in all four modes; nothing is dropped, renamed, or collided. (One incidental finding: the store path this test drives, `encodeXmldbUriFor`, escapes a literal `%`→`%25` and is therefore injective — a *different* encoding from the persistent-store path #6463 documents as non-injective, which is itself more evidence for the "different code paths encode differently" thesis. Store-time collision rides that other path and is out of scope for this test.)

## 2. XML-RPC surface + the full N×N cross-surface matrix (`extensions/webdav`)

The two follow-ups noted in the harness javadoc are now done. XML-RPC is added as a third surface (create via `parse`, read via `getDocument` over `/xmlrpc` — needs a test-scoped `xmlrpc-client` dep and a `/xmlrpc` servlet-mapping in the test webapp), and a new `fullCrossSurfaceMatrix()` test creates each name via **each** surface and reads it back — by the requested name — via **every** surface, ratcheted against a documented `KNOWN_MATRIX_FAILURES` allowlist just like the original test.

The characterized matrix (abbreviated to the discriminating rows; the other 10 names are all-PASS across the board):

```
                       read:WebDAV  read:REST  read:XML-RPC
-- created via WebDAV --
  a%b.xml    (created)    PASS        PASS        FAIL
  a%20b.xml  (created)    PASS        PASS        FAIL
  a#b.xml    (created)    PASS        PASS        FAIL
-- created via REST --
  a%b.xml    (created)    PASS        PASS        FAIL
  a%20b.xml  (created)    PASS        PASS        FAIL
  a#b.xml    (created)    PASS        PASS        FAIL
-- created via XML-RPC --
  a%b.xml    (FAILED)     FAIL        FAIL        FAIL   ← XML-RPC rejects a literal % on create
  a%20b.xml  (created)    FAIL        FAIL        PASS   ← stored under a form only XML-RPC reads back
  a#b.xml    (created)    FAIL        FAIL        FAIL   ← stored, but unreadable by any surface, even XML-RPC
```

Two things this makes concrete:

- **WebDAV and REST agree on every corpus name** — zero cross-surface failures between them.
- **XML-RPC is the divergent surface for names containing `%` or `#`** — 14 failing cells, all of them `%`/`#`. It can't read a `%`/`#` name that HTTP stored, rejects a literal `%` outright on create, stores `a%20b.xml` under a form only it reads back, and stores `a#b.xml` unreadable by anyone.

This changes no production behavior — the failing cells are documented current behavior, pinned by the ratchet; each clears as the naming contract (#6463) reaches the XML-RPC path.

## 3. XQuery accessors (`exist-core`)

`org.exist.xmldb.XQueryNamingConformanceTest` characterizes the XQuery side over the same corpus: create via `xmldb:store` / `xmldb:create-collection` in two modes — **naive** (raw name) and the eXide/TEI **`xmldb:encode-uri` workaround** (encode applied on both the store and the read) — then read back by the requested name via `doc()`, `collection()`, `xmldb:get-child-resources()`, and `xmldb:get-child-collections()`. Ratcheted against a documented allowlist like the others.

**Headline finding: `doc()`, `collection()` and `get-child-resources()` behave identically** — in every row they pass and fail together, and `get-child-collections` mirrors them for collections. They are one behavior, not four; the naming problem is not that the readers disagree with each other, but that all of them miss any name whose stored key differs from the requested name.

```
                    created  stored-key        doc()  coll()  get-child-res
-- xmldb:store — naive (raw name) --
  plain.xml         ok       plain.xml         PASS   PASS    PASS
  a+b.xml           ok       a+b.xml           PASS   PASS    PASS    ← stored literally; HTTP encodes it to a%2Bb.xml
  with space.xml    FAIL     — (FORG0001)      FAIL   FAIL    FAIL    ← raw store rejects a space
  a%b.xml           FAIL     — (FORG0001)      FAIL   FAIL    FAIL
  café.xml          ok       caf%C3%A9.xml     FAIL   FAIL    FAIL    ← store↔doc miss
-- xmldb:store — encode-uri workaround (encode on store & read) --
  with space.xml    ok       with%20space.xml  PASS   PASS    PASS
  a%b.xml           ok       a%25b.xml         PASS   PASS    PASS
  café.xml          ok       caf%C3%A9.xml     PASS   PASS    PASS
```

Two further findings: naive `xmldb:store`/`create-collection` reject a space, a literal `%` and a `#` outright (FORG0001); and — notably — they store `+ @ & ( ) '` **literally** (`a+b.xml` stays `a+b.xml`), where WebDAV and REST percent-encode them. So `xmldb:store` and the HTTP surfaces disagree on the stored key itself. The `encode-uri` workaround, applied consistently on store and read, makes every name round-trip.

Green locally: the webdav naming tests (2/2) and the full webdav module suite (17/0), the exist-core backup/restore (4/4) and XQuery-accessor (1/1) tests, Codacy clean.

exist-core/src/test/java/org/exist/backup/BackupRestoreNamingConformanceTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+    @Parameter(value = 1)
+    public boolean direct;
+
+    @Parameter(value = 2)
+    public boolean zip;
+
+    @Parameters(name = "{0} zip:{2}")
+    public static java.util.Collection<Object[]> data() {
+        return Arrays.asList(new Object[][]{
```
reinhapa, line 127: Better use `List.of`
```suggestion
        return List.of(new Object[][]{
```

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
```diff
@@ -0,0 +1,604 @@
...
+        CORPUS.put("encoded-space",   "a%20b.xml");        // literal "%20" in the name
+        CORPUS.put("hash",            "a#b.xml");
+        CORPUS.put("at",              "a@b.xml");
+        CORPUS.put("ampersand",       "a&b.xml");
+        CORPUS.put("parens",          "report(2024).xml");
+        CORPUS.put("apostrophe",      "o'brien.xml");
+        CORPUS.put("non-ascii",       "café.xml");
+        CORPUS.put("cyrillic",        "Привет.xml");
+        CORPUS.put("cjk",             "文書.xml");
+    }
```
reinhapa, line 122: Use `Map.ofEntries` if order does not matter....
```suggestion
    private static final Map<String, String> CORPUS = Map.ofEntries(
        Map.entry("ascii-control", "plain.xml"), // control: must always work
        Map.entry("space", "with space.xml"), //
        Map.entry("plus", "a+b.xml"), //
        Map.entry("literal-percent", "a%b.xml"), //
        Map.entry("encoded-space", "a%20b.xml"),  // literal "%20" in the name
        Map.entry("hash", "a#b.xml"), //
        Map.entry("at", "a@b.xml"), //
        Map.entry("ampersand", "a&b.xml"), //
        Map.entry("parens", "report(2024).xml"), //
        Map.entry("apostrophe", "o'brien.xml"), //
        Map.entry("non-ascii", "café.xml"), //
        Map.entry("cyrillic", "Привет.xml"), //
        Map.entry("cjk", "文書.xml") //
    );
```

reinhapa (changes requested):

exist-core/src/test/java/org/exist/backup/BackupRestoreNamingConformanceTest.java
```diff
@@ -0,0 +1,271 @@
...
+ *
+ * <p>Extends the resource-naming conformance net (see {@code ResourceNamingConformanceTest} in the webdav
+ * module, and issue #6463) to the backup/restore path @line-o asked about on PR #6508: for a corpus of
+ * awkward leaf names (spaces, sub-delimiters, non-ASCII, CJK, Cyrillic, and the literal-{@code %} cases
+ * Decision 2 flags), it stores each resource, snapshots the exact stored key and content, exports a full
+ * backup, wipes the collection, restores, and asserts the snapshot is reproduced <em>byte-for-byte</em> —
+ * no resource lost, renamed, or collided by the round-trip. That is the guard against "even theoretical
+ * data loss" through backup/restore.</p>
+ *
+ * <p><b>Scope.</b> This guards the <em>restore</em> path. Resources are stored under the same leaf keys a
```
duncdrum, line 74: Small accuracy nit: the javadoc (and `storedKeyFor()` below) describe resources as stored under the same leaf keys `encodeXmldbUriFor` would produce for a WebDAV/REST store — but neither `EXistServlet.doPut` nor `ExistCollection.createFile` actually call that method, and the restore path only uses it for legacy `version<1` backups (this suite's exports are `version=1`). Might be worth tightening the wording so it doesn't imply this validates the exact WebDAV/REST key format.

duncdrum (commented):
Ran an automated review over the naming-conformance harness — left a few notes on the confirmed findings below.

exist-core/src/test/java/org/exist/backup/BackupRestoreNamingConformanceTest.java
```diff
@@ -0,0 +1,271 @@
...
+        // round-trip below genuinely exercises all of them (not fewer after an accidental store-time clash).
+        assertEquals("setup stored fewer resources than the corpus — a name clashed on the way in; see the "
+                        + "printed mapping", XML_NAMES.size() + 1, before.size());
+
+        // export a full backup
+        final Path backup;
+        try (final DBBroker broker = pool.get(Optional.of(pool.getSecurityManager().getSystemSubject()));
+             final Txn txn = pool.getTransactionManager().beginTransaction()) {
+            final SystemExport export = new SystemExport(broker, txn, null, null, direct);
+            backup = export.export(temporaryFolder.newFolder().getAbsolutePath(), false, zip, null);
```
duncdrum, line 200: `SystemExport.export()` can return `null` on an internal `IOException`/`TerminatedException` (and `errorList` is `null` here too, so we can't catch it that way). Passing that straight into `SystemImport.restore(...)` would hit `Files.isDirectory(null)` and NPE rather than failing with a clear message — might be worth a null-check with a descriptive `fail(...)`.

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
```diff
@@ -0,0 +1,604 @@
...
+        restDelete(TEST_COLLECTION);
+        http.close();
+    }
+
+    /**
+     * Resets the test collection to empty: delete it, then recreate it by REST-PUTting a seed
+     * document (which auto-creates the collection) and removing the seed. Wiping the whole
+     * collection avoids depending on the stored name, which is what this test characterizes.
+     */
+    private static void freshCollection() {
```
duncdrum, line 204: `freshCollection()` swallows all exceptions silently, and it's called up to ~52 times in a run. A transient HTTP hiccup here would leave the collection missing before the next probe, and the resulting failure would look identical to a genuine naming regression against `KNOWN_FAILURES`. Might be worth at least logging the swallowed exception so it's distinguishable.

exist-core/src/test/java/org/exist/xmldb/XQueryNamingConformanceTest.java
```diff
@@ -0,0 +1,220 @@
...
+    private void freshCollections() {
+        xqStr("if (xmldb:collection-available(" + lit(COLL_PARENT) + ")) then xmldb:remove(" + lit(COLL_PARENT) + ") else (), "
+                + "xmldb:create-collection('/db', 'naming-xq-coll')");
+    }
+
+    private boolean xqBool(final String query) {
+        return "true".equals(xqStr(query));
+    }
+
+    private String xqStr(final String query) {
```
duncdrum, line 202: Two small things in `xqStr()`: it opens/closes a fresh `DBBroker` on every call (~260 checkouts across this test), and the same catch-and-stringify is also used by `freshResources()`/`freshCollections()` for a housekeeping reset where the result is discarded — so a reset failure for an unrelated reason (lock/broker issue) would be silently absorbed and the next probe would run against a collection that wasn't actually reset. Reusing one broker per test and giving the housekeeping calls their own visible failure path would help on both fronts.

exist-core/src/test/java/org/exist/xmldb/XQueryNamingConformanceTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+    private static void record(final Set<String> failing, final String kind, final String mode, final String name, final String reader, final boolean ok) {
+        if (!ok) {
+            failing.add(kind + ":" + mode + ":" + name + ":" + reader);
+        }
+    }
+
+    /** An XQuery string literal (double-quoted) for an arbitrary name; only {@code &} and {@code <} need escaping. */
+    private static String lit(final String s) {
+        return '"' + s.replace("&", "&amp;").replace("<", "&lt;") + '"';
```
duncdrum, line 185: `lit()` escapes `&` and `<` but not `"` when building XQuery string literals — the sibling `BackupRestoreNamingConformanceTest.contentFor()` does escape it. If a future corpus name includes a double quote this would break the generated XQuery rather than exercise the naming behavior. Worth adding the same escape here.

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
```diff
@@ -0,0 +1,604 @@
...
+    }
+
+    /** Re-assert the interrupt flag if a blocking HttpClient call was interrupted. */
+    private static void restoreInterrupt(final Exception e) {
+        if (e instanceof InterruptedException) {
+            Thread.currentThread().interrupt();
+        }
+    }
+
+    private static String unescapeXml(final String s) {
```
duncdrum, line 425: `unescapeXml()` chains its `replace()` calls sequentially (`&amp;`→`&` before `&lt;`→`<`, etc.), so a name whose listing shows `&amp;lt;` would double-decode to `<` instead of `&lt;`. Not reachable with the current corpus, but doing a single pass (or ordering longest-match-last) would avoid it becoming a latent bug in the diagnostic output.

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
```diff
@@ -0,0 +1,604 @@
...
+        return s.replace("&amp;", "&").replace("&lt;", "<").replace("&gt;", ">")
+                .replace("&quot;", "\"").replace("&apos;", "'");
+    }
+
+    private static String basicAuth() {
```
duncdrum, line 430: `basicAuth()` reimplements Basic-auth header encoding that already exists as `WebDavHttpClient.basicAuthorization()` in this package and `AbstractHttpTest.basicAuthorizationHeader()` in exist-core's shared test base. Reusing one of those would avoid a third copy to keep in sync if the encoding ever needs to change.

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
```diff
@@ -0,0 +1,604 @@
...
+                    && Boolean.TRUE.equals(r.webdavReadBack)
+                    && Boolean.TRUE.equals(r.restReadBack);
+            if (!conformant) {
+                failing.add(r.label);
+            }
+        }
+
+        // Ratchet: the set of failing names must match KNOWN_FAILURES exactly.
+        final Set<String> regressions = new LinkedHashSet<>(failing);
+        regressions.removeAll(KNOWN_FAILURES);              // failing but expected to pass -> regression
```
duncdrum, line 264: The failing-set / `KNOWN_*`-diff / regressions-message ratchet logic here is duplicated near-verbatim at line 481 in this file and again in `XQueryNamingConformanceTest.java:158`. Might be worth factoring it into a shared helper so future tweaks (wording, tracking-issue links, etc.) don't need to be applied in three places.

exist-core/src/test/java/org/exist/backup/BackupRestoreNamingConformanceTest.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

You're right: neither `EXistServlet.doPut` nor `ExistCollection.createFile` calls `encodeXmldbUriFor`. The javadoc now says only that the resources are stored under the keys `encodeXmldbUriFor` produces, and that this is not a claim about WebDAV or REST keys, which `ResourceNamingConformanceTest` checks.

exist-core/src/test/java/org/exist/backup/BackupRestoreNamingConformanceTest.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Added an `assertNotNull` with the mode and zip flag in the message.

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Went a step past logging. `freshCollection()` now fails with its own "could not reset" message if the seed PUT doesn't succeed, and `restPut` prints the exception it swallows. A reset failure can no longer pass for a naming regression.

exist-core/src/test/java/org/exist/xmldb/XQueryNamingConformanceTest.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Both done: one broker per test, opened in `@Before` and closed in `@After`, and resets go through a `housekeeping(...)` path that throws. Probes still turn an error into a result, since that's what they measure.

exist-core/src/test/java/org/exist/xmldb/XQueryNamingConformanceTest.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Fixed: `"` is escaped as `&quot;`.

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Now a single pass over the five predefined entities, so `&amp;lt;` decodes to `&lt;`.

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Reuses `WebDavHttpClient.basicAuthorization`, now package-private. I left `AbstractHttpTest`'s copy alone, since the webdav module doesn't depend on exist-core's test classes.

extensions/webdav/src/test/java/org/exist/webdav/ResourceNamingConformanceTest.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Factored into `org.exist.test.KnownFailuresRatchet`, next to exist-core's other shared test helpers. All three ratchets use it, keeping their own wording for what a regressed entry is.

duncdrum:
@joewiz needs another CI happiness pass

PR #6451, opened 2026-06-07
[bugfix] Decode '+' as a literal plus in the xmldb URI functions (#1824)

### LINKED ISSUE #44: Coding of path components in URIs
This is a port of SF ticket http://sourceforge.net/p/exist/bugs/766/ - for short:

In a few places the eXist codebase uses "java.net.URLDecoder" (which is intended to be used on http _query strings_) instead of "java.net.URI", resulting in wrong coding of _path components_ in URIs and thereby creation of ressources, that cannot be fetched by the name, they were uploaded with.

Simplest way to show the (mis)behaviour:
- From the dashboard start collections browser,
- create collection "/db/testing",
- upload a ressource, eg. "A+A.svg"
- note that it gets represented as "A%2BA.svg" in the list widget,
- which is the same as displayed in "/exist/rest/db/testing/",
- and also in "/exist/webdav/db/testing/", while
- in eXide open dialog it will show up as "A+A.svg".

BUT
- fetching /exist/rest/db/testing/A%2BA.svg will result in a 404,
- same as /exist/rest/db/testing/A+A.svg,
- while fetching /exist/rest/db/testing/A%252BA.svg will work,
- same as /exist/webdav/db/testing/A%252BA.svg,
- all the while the file is inaccessible from eXide's "open", but
- double clicking in coll.browser will fire up eXide to "A%2BA.svg".

That is: there is **double encoding** happening. The plus sign is encoded to %2B on store, and the newly introduced percent sign has to be coded to %25 on retrieve. This applies to all the characters, that java.net.URLDecoder codes different from java.net.URI in path components (most prominently the plus sign, but also others).

### LINKED ISSUE #1824: Bug in decoding of "+" in xmldb URI-related functions

joewiz:

## Summary

`xmldb:decode` and `xmldb:decode-uri` decoded a `+` to a space, so a name containing a literal `+` could not be round-tripped through `xmldb:encode-uri` / `xmldb:decode-uri`. Both functions routed through `URIUtils.urlDecodeUtf8`, which wraps `java.net.URLDecoder` and therefore applies `application/x-www-form-urlencoded` rules — where `+` means space. That is wrong for URI path components, which follow RFC 3986 (where `+` is a literal plus and only `%20` is a space).

Fixes #1824 (and the long-closed #44, the original report of the same `URLDecoder` mistake).

## What changed

- **`URIUtils.decodeForURI(String)`** — new, the exact inverse of the existing `encodeForURI()`. It decodes `%XX` escapes (interpreting consecutive escapes as a UTF-8 byte sequence) and leaves every other character — including `+` — literal. `decodeForURI(encodeForURI(s))` equals `s` for every `s`.
- **`XMLDBURIFunctions`** — `xmldb:decode` / `xmldb:decode-uri` now route through `decodeForURI` instead of `urlDecodeUtf8`.
- `urlDecodeUtf8` is **left unchanged** for its other callers — only the user-facing `xmldb:` URI functions are switched, to keep the blast radius minimal.

### Before / after

```xquery
xmldb:decode-uri(xs:anyURI("a%2Bb"))   (: before: "a b"   after: "a+b" :)
xmldb:decode-uri(xs:anyURI("a+b"))     (: before: "a b"   after: "a+b" :)
xmldb:decode-uri(xs:anyURI("a%20b"))   (: before: "a b"   after: "a b" — unchanged :)
```

## Scope note

`XmldbURI.getCollectionPath()` has the same `URLDecoder` pattern, but it is a central internal path used throughout storage. Changing it is deferred to the broader name/URI contract work (gated on the cross-surface conformance harness, PR A) rather than bundled into this focused bug fix.

## Test plan

- [x] `URIUtilsTest` (JUnit) — `decodeForURI` for `+` (encoded and bare), space, percent (incl. literal `%2F` → `%252F` round-trip), unreserved, and 2/3/4-byte UTF-8; plus an `encode → decode` round-trip property over a corpus of awkward names (space, `+`, `%`, reserved/sub-delims, `café`, Cyrillic, CJK, multi-byte). 10/10 green.
- [x] `xmldb/uri-encoding-tests.xql` (XQSuite) — functional `xmldb:decode-uri` / `xmldb:decode` regression assertions (`a%2Bb` → `a+b`, `a+b` → `a+b`, `a%20b` → `a b`, and a mixed name). 5/5 green.
- [x] Codacy/PMD clean on the new code.
- [x] No existing test relied on the old `+`→space behavior.

## Related

- Resource-naming-stability tasking (PR B).
- Issues: #1824 (this fix), #44 (closed; same root cause), and the broader meta #3795.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/util/URIUtils.java
+++ b/exist-core/src/main/java/org/exist/xquery/util/URIUtils.java
@@ -258,6 +258,15 @@ public class URIUtils {
 	 * functions (see eXist-db/exist#1824, #44): {@code decodeForURI(encodeForURI(s))} equals
 	 * {@code s} for every {@code s}.
 	 *
+	 * <p>This is deliberately a standalone percent-decoder rather than a call to
+	 * {@link java.net.URI#getPath()}. {@code java.net.URI} is unsuitable as a general decoder for the
+	 * arbitrary strings that {@code xmldb:decode}/{@code xmldb:decode-uri} accept: it throws
+	 * {@code URISyntaxException} on inputs that are perfectly valid here (a literal space, a trailing
+	 * or malformed {@code %}, characters such as <code>{</code> or <code>}</code>), and worse, it
+	 * <em>silently truncates</em> at {@code '?'} and {@code '#'} (parsing the remainder as a query or
+	 * fragment) — losing data with no error. This decoder never throws and never truncates: any
+	 * {@code '%'} not followed by two hex digits is preserved verbatim. See {@code URIUtilsTest}.</p>
+	 *
 	 * @param uriComponent the percent-encoded path component to decode.
 	 *
 	 * @return the decoded path component.
--- a/exist-core/src/main/java/org/exist/xquery/util/URIUtils.java
+++ b/exist-core/src/main/java/org/exist/xquery/util/URIUtils.java
@@ -245,7 +245,68 @@ public class URIUtils {
 
 		return new String(buf.buf, 0, buf.count);
 	}
-	
+
+	/**
+	 * Decodes a percent-encoded URI path component back to its literal form, the inverse of
+	 * {@link #encodeForURI(String)}. Each {@code %XX} escape is decoded to a byte; consecutive
+	 * escapes are interpreted together as a UTF-8 byte sequence. Every other character is left
+	 * unchanged.
+	 *
+	 * Unlike {@link #urlDecodeUtf8(String)} (which wraps {@link java.net.URLDecoder} and therefore
+	 * follows application/x-www-form-urlencoded rules), this method treats {@code '+'} as a literal
+	 * plus sign, per RFC 3986. This is required for round-tripping names through the xmldb URI
+	 * functions (see eXist-db/exist#1824, #44): {@code decodeForURI(encodeForURI(s))} equals
+	 * {@code s} for every {@code s}.
+	 *
+	 * @param uriComponent the percent-encoded path component to decode.
+	 *
+	 * @return the decoded path component.
+	 */
+	public static String decodeForURI(final String uriComponent) {
+		if (uriComponent.indexOf('%') == -1) {
+			// fast path: nothing percent-encoded, nothing to decode
+			return uriComponent;
+		}
+
+		final int len = uriComponent.length();
+		final StringBuilder out = new StringBuilder(len);
+		final java.io.ByteArrayOutputStream pending = new java.io.ByteArrayOutputStream();
+
+		int i = 0;
+		while (i < len) {
+			final char c = uriComponent.charAt(i);
+			if (c == '%' && i + 2 < len && isHexDigit(uriComponent.charAt(i + 1)) && isHexDigit(uriComponent.charAt(i + 2))) {
+				pending.write((hexValue(uriComponent.charAt(i + 1)) << 4) | hexValue(uriComponent.charAt(i + 2)));
+				i += 3;
+			} else {
+				if (pending.size() > 0) {
+					out.append(pending.toString(UTF_8));
+					pending.reset();
+				}
+				out.append(c);
+				i++;
+			}
+		}
+		if (pending.size() > 0) {
+			out.append(pending.toString(UTF_8));
+		}
+		return out.toString();
+	}
+
+	private static boolean isHexDigit(final char c) {
+		return (c >= '0' && c <= '9') || (c >= 'A' && c <= 'F') || (c >= 'a' && c <= 'f');
+	}
+
+	private static int hexValue(final char c) {
+		if (c >= '0' && c <= '9') {
+			return c - '0';
+		}
+		if (c >= 'A' && c <= 'F') {
+			return c - 'A' + 10;
+		}
+		return c - 'a' + 10;
+	}
+
 	public static String iriToURI(String uriPart) {
 		String result = urlEncodeUtf8(uriPart);
 		result = result.replaceAll("%23", "#");
--- a/exist-core/src/main/java/org/exist/xquery/functions/xmldb/XMLDBURIFunctions.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/xmldb/XMLDBURIFunctions.java
@@ -98,7 +98,8 @@ public class XMLDBURIFunctions extends BasicFunction {
 			} else if(isCalledAs("encode-uri")) {
 				return new AnyURIValue(this, URIUtils.encodeXmldbUriFor(args[0].getStringValue()));
 			} else {
-				return new StringValue(this, URIUtils.urlDecodeUtf8(args[0].getStringValue()));
+				// RFC 3986 percent-decoding: '+' is a literal plus, not a space (#1824, #44)
+				return new StringValue(this, URIUtils.decodeForURI(args[0].getStringValue()));
 			}
 		} catch(final URISyntaxException e) {
             logger.error(e.getMessage(), e);
--- a/exist-core/src/test/java/org/exist/xquery/util/URIUtilsTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/util/URIUtilsTest.java
@@ -256,4 +256,35 @@ class URIUtilsTest {
                     "encode/decode round-trip failed for: " + name);
         }
     }
+
+    /**
+     * decodeForURI must never throw and never truncate, even on input that is not the output of
+     * encodeForURI — xmldb:decode/xmldb:decode-uri accept arbitrary user strings. Each case here
+     * is one that {@code new java.net.URI(s).getPath()} mishandles (throws URISyntaxException, or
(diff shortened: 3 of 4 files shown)


exist-core/src/main/java/org/exist/xquery/util/URIUtils.java
reinhapa, line 1: Hmm, what about using the Java URI as it is more precise for paths & handles encoding/decoding automatically?  In case this does not have a negative performance implication, it would greatly reduce the existing code within the `URIUtils` class..

```
import java.net.URI;
import java.net.URISyntaxException;

String pathPart = "my path/with spaces+plus";
try {
    URI uri = new URI(null, null, pathPart, null, null);
    String encodedPath = uri.toString(); // Automatically encodes path
    System.out.println(encodedPath); // Output: my%20path/with%20spaces+plus

    // To decode:
    URI decodedUri = new URI(encodedPath);
    String decodedPath = decodedUri.getPath();
    System.out.println(decodedPath); // Output: /my path/with spaces+plus
} catch (URISyntaxException e) {
    e.printStackTrace();
}
```

reinhapa (changes requested):

reinhapa (approved):

joewiz:

Thanks Patrick — you're right that `java.net.URI` keeps `+` literal in a path, which is the behavior this fix is after, so it's a fair thing to reach for. I tried it as the decoder and it doesn't hold up for the general case: `xmldb:decode`/`xmldb:decode-uri` are public functions called on arbitrary strings, and `new URI(s).getPath()` mishandles a number of inputs that this decoder must accept. Comparing the two on the same inputs:

| input | `new URI(s).getPath()` | `decodeForURI` |
|---|---|---|
| `a%2Bb` | `a+b` | `a+b` |
| `a+b` (the #1824 case) | `a+b` | `a+b` |
| `a%20b` | `a b` | `a b` |
| `%C3%A9` | `é` | `é` |
| `a b` (literal space) | **throws** `URISyntaxException` | `a b` |
| `100%` | **throws** | `100%` |
| `a%ZZb` | **throws** | `a%ZZb` |
| `a{b}c` | **throws** | `a{b}c` |
| `a?b` | **`a`** (truncates — `?b` parsed as query) | `a?b` |
| `a#b` | **`a`** (truncates — `#b` parsed as fragment) | `a#b` |

Two problems: it throws on ordinary inputs, and worse, it silently truncates at `?` and `#` — data loss with no error. So I've kept the standalone percent-decoder, but addressed the spirit of your comment: I added a javadoc paragraph explaining exactly why `java.net.URI` isn't used here, and a test (`decodeForURIRobustOnMalformedAndReservedInput`) pinning each of the rows above so the next reader doesn't have to re-derive it.

On the encode side, `URIUtils.encodeForURI` is existing code and this PR is decode-only, so I'd rather not fold an `encodeForURI` rewrite into it — happy to look at that separately if it's worth doing.

Pushed as `f0c3ec1`.

reinhapa:
The way you tested it is NOT the way I suggested. Use the following URI constructor, that takes every part of the URI separate. This will not throw an exception in our case...

```
jshell> new URI(null, null, null, 0, "a b", null, null)
$1 ==> a%20b

jshell> new URI(null, null, null, 0, "100%", null, null)
$2 ==> 100%25

jshell> new URI(null, null, null, 0, "a%ZZb", null, null)
$3 ==> a%25ZZb

jshell> new URI(null, null, null, 0, "a?b", null, null)
$4 ==> a%3Fb

jshell> new URI(null, null, null, 0, "a#b", null, null)
$5 ==> a%23b
```

joewiz:

Thanks Patrick — and you're right that the multi-argument constructor is a robust *encoder*: it won't throw on any of those, exactly as your examples show. The catch is direction. `decodeForURI` *decodes* (percent-encoded → literal: `a%20b` → `a b`), whereas `new URI(null, null, null, 0, path, null, null)` *encodes* (literal → percent-encoded: `a b` → `a%20b`) — the opposite transform. So it can't stand in for the decoder. Feeding it an already-encoded string doesn't decode it:

```
jshell> new URI(null, null, null, 0, "a%20b", null, null).getPath()
$1 ==> a%20b        // unchanged — not decoded to "a b"

jshell> new URI(null, null, null, 0, "a%20b", null, null).toString()
$2 ==> a%2520b      // double-encoded
```

The only `java.net.URI` route that actually decodes is the single-argument parse plus `getPath()` — which is the one in my table, and the one that throws on `a b` / `100%` / `a%ZZb` and silently truncates `a?b` → `a`. Hence the standalone decoder for this direction.

Your multi-arg encoder is a good fit for the *encode* side, i.e. `encodeForURI` — which is existing code this PR doesn't touch, so I've left it out of scope here; happy to look at swapping that separately if it's worthwhile. Thanks for the approval and the careful review.

joewiz:
@reinhapa Please let me know if Claude is still on the wrong track here and needs to be steered in the right direction. Sorry!

duncdrum:
@joewiz can you rebase this once more please

joewiz:
@duncdrum Rebased and CI green

PR #6455, opened 2026-06-08
[feature] Index-first Lucene search: ft:query-scope (live nodes) and ft:search-scope (ES-shaped map)

joewiz:

> **Names are provisional.** `ft:query-scope` / `ft:search-scope` are working names — happy to rename before merge.

## Summary

Two new Lucene functions for **index-first, name-independent** full-text search over a collection scope:

- **`ft:query-scope($scope, $query, $options?)` → live nodes.** The index-first sibling of `ft:query`: it searches the Lucene index directly over `$scope` (collection/document URIs, recursive) and returns every matching indexed element — of any element type — as a live node carrying its score and matches. `ft:score`/`ft:field`/`ft:facets`/`ft:highlight-field-matches` compose on the result as usual.

- **`ft:search-scope($scope, $query, $options?)` → one Elasticsearch `_search`-shaped `map(*)`.** The detached, map-returning companion for API builders: `{ total, max-score, hits[], facets }`, each hit `{ uri, node-id, score, source, highlight }`. Plain result data, no node-set to walk and re-serialize.

## Motivation

Two recurring needs:

1. **Index-first, name-independent search.** To search "all indexed content under a collection regardless of element name" today, you write `collection(...)//*[ft:query(., 'field:(…)')]`. The descendant-wildcard form **loses `ft:score`** for field queries (a known artifact), and it forces callers to enumerate the contributing element names. `ft:query-scope` queries the index directly over the scope and returns every matching indexed node with correct relevance — no wildcard, no element-name union to maintain.

2. **An Elasticsearch `_search`-style result for API builders.** For something like existdb-openapi's `/api/search`, what you want is plain result data — `total`, `hits` with fields, `facets` — not a node-set to walk and re-serialize. `ft:search-scope` returns exactly that, assembled natively.

## Why two functions (and the naming)

It helps to place these against the existing functions on two axes — **detached vs live** result, and **XML vs map** output:

| | XML output | map output |
|---|---|---|
| **detached** (a snapshot, no live nodes) | `ft:search` (legacy report) | **`ft:search-scope`** (new) |
| **live nodes** | `ft:query` (context-scoped), **`ft:query-scope`** (scope-scoped, new) | (n/a — flattening live nodes to a map defeats their purpose) |

- The two new functions are **not** variants of one another and shouldn't be folded into one: they have divergent return contracts (live node-set vs detached map) and sit on different `LuceneIndexWorker` methods. The `query` / `search` pair mirrors eXist's existing split (`ft:query` returns live nodes; the legacy `ft:search` returns a detached report).
- The `-scope` suffix signals "scoped by a collection, not by an XPath context node-set." I avoided `ft:search-index`: from an Elasticsearch mindset "index" is the *corpus you search* (a collection here), not part of a verb, so it reads ambiguously.

## Hit granularity (the one decision worth calling out)

"Document" means three things in this stack, and conflating them produces real count bugs:

- an **eXist collection** — the storage container `$scope` names;
- an **eXist document** — an XML resource (`intro.xml`), an XmldbURI;
- an **indexed element** — eXist creates **one Lucene document per indexed element occurrence** (per `<text qname="…">`), so one document with two `<para>` and one `<caption>` yields **three** Lucene documents.

So `$scope` filters at eXist-document granularity, but hits/counts/facets are at indexed-element granularity. Elasticsearch guarantees one Lucene document per ES document (1:1); eXist does not. `ft:search-scope` therefore defaults to **indexed-element** granularity (honest to the index, sub-document precision), with a **`collapse`** option for the ES-faithful one-hit-per-document view (group by document, best-scoring element, `total` = distinct documents) — the analog of ES field-collapse / `top_hits`.

## `ft:search-scope` options

`$options` carries a `filter` that restricts the **query**, plus keys that shape the **result**:

| key | type | meaning |
|---|---|---|
| `"filter"` | map(*) | facet drill-down `{ dimension: value(s) }` restricting the search (ES post-filter analog; keeps `total`/paging consistent) |
| `"fields"` | xs:string* | stored fields to include in each hit's `source` (`_source`) |
| `"highlight"` | xs:string* | fields to highlight; adds a per-hit `highlight` map of `exist:field`/`exist:match` nodes |
| `"facets"` | xs:string* | facet dimensions to aggregate over the full match set |
| `"collapse"` | xs:boolean | group hits to one-per-document (best-scoring element; `total` = distinct documents) |
| `"offset"` / `"limit"` | xs:integer | page the ranked hits |

Note: `ft:search-scope` returns a single envelope `map`, so read the hit count from `?total` (or `count(?hits?*)`) — `count(ft:search-scope(...))` is always `1`.

## What changed (`extensions/indexes/lucene`)

- **`QueryScope.java`** (`ft:query-scope`) and **`SearchScope.java`** (`ft:search-scope`) — the two functions; `SearchScope` assembles the ES map natively from the same index-first query, reading scores/fields/facets the way `ft:score`/`ft:field`/`ft:facets` do, and reusing the `ft:highlight-field-matches` engine for `highlight`.
- **`LuceneScope.java`** — shared scope-resolution + index-first query execution, used by both functions (no duplication).
- **`LuceneModule.java`** — registers the four signatures.
- **`Field.java`** — `highlightMatches` made package-private `static` so `ft:search-scope` can reuse it on the live node it materializes; no behavior change to `ft:highlight-field-matches`.
- Tests: `ft-query-scope.xqm`, `ft-search-scope.xqm`, `scope-dls.xqm`, and the `LuceneQueryScopeTests` runner.

## Companion fix (faceted highlighting)

`ft:search-scope`'s `highlight` works on its own for unfaceted queries. Combining `filter` (facet drill-down) **with** `highlight` additionally requires a small fix to a **pre-existing** defect — a facet `DrillDownQuery` silently disabled `ft:highlight-field-matches` term extraction — submitted as a **separate bugfix PR, #6454** (it affects `ft:query` faceted highlighting independent of these functions, so it's based on `develop`, not this branch). This PR doesn't depend on it except for that combined path.

## Test plan

- [x] 42 XQSuite tests across the suite: `ft:query-scope` (name-independence, nested-element scoring, composition with `ft:score`/`ft:field`/`ft:facets`/`ft:highlight-field-matches`, live nodes, sortable-by-score); `ft:search-scope` (envelope shape, element vs `collapse` granularity, `source`/`facets`/`highlight`/`filter`/`offset`/`limit`).
- [x] **Document-level security** (`scope-dls.xqm`): a guest never gets nodes or hits from documents it cannot read (verified via `system:as-user` against mixed read permissions) — the restricted document is absent from both `hits` and `total`, so the count does not leak its existence.
- [x] Validated against a real corpus (eXist's own function docs) behind an `/api/search`-style endpoint: identical hit set to the `collection(...)//el[ft:query(...)]` approach with `ft:score` preserved; faceted highlighting end-to-end (with #6454 in place).
- [x] Codacy/PMD clean on the new files.


duncdrum (commented):
A few findings from review, kept close to the code. Two additional notes (a pre-existing off-by-one in `Field.java` and a root-cause note on the `ft:query` score-loss bug in `Query.java`) are posted as separate general comments below since they fall outside this diff's line ranges.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
```diff
@@ -0,0 +1,426 @@
...
+            total = ranked.size();
+        } else {
+            ranked = new ArrayList<>(all);
+            total = all.size();
+        }
+        ranked.sort((a, b) -> Double.compare(b.score(), a.score()));
+
+        // page the ranked list: skip $offset, then take at most $limit
+        final int from = Math.min(spec.offset(), ranked.size());
+        final int to = (spec.limit() >= 0) ? Math.min(ranked.size(), from + spec.limit()) : ranked.size();
```
duncdrum, line 227: Integer overflow: `from + spec.limit()` can wrap to a negative int when `limit` is `Integer.MAX_VALUE` (a common "no limit" sentinel), then `ranked.subList(from, to)` throws `IllegalArgumentException` instead of returning the rest of the list.
```suggestion
        final int to = (spec.limit() >= 0) ? (int) Math.min(ranked.size(), from + (long) spec.limit()) : ranked.size();
```

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/LuceneScope.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Resolve a sequence of collection or document URIs to a document set. Collection URIs are searched
+     * recursively (including sub-collections); a URI that is not a collection is tried as a single document.
+     */
+    static MutableDocumentSet resolveScope(final BasicFunction fn, final Sequence scope) throws XPathException {
+        final MutableDocumentSet docs = new DefaultDocumentSet();
+        for (final SequenceIterator i = scope.iterate(); i.hasNext(); ) {
+            final String path = i.nextItem().getStringValue();
+            final XmldbURI uri = XmldbURI.create(path);
```
duncdrum, line 66: `XmldbURI.create(path)` throws an unchecked `IllegalArgumentException` on a malformed URI, and it's outside the `try` block below that handles `PermissionDeniedException`/`LockException`. A bad scope string will surface as a raw Java exception instead of an `XPathException` with a Lucene error code. Worth wrapping this call too.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryScope.java
```diff
@@ -0,0 +1,114 @@
...
+
+        final MutableDocumentSet docs = LuceneScope.resolveScope(this, args[0]);
+        if (docs.getDocumentCount() == 0) {
+            return Sequence.EMPTY_SEQUENCE;
+        }
+
+        // options is the 3rd argument (1-based position 3: scope, query, options), as in ft:query.
+        // parseOptions short-circuits to default QueryOptions when getArgumentCount() < 3, so the
+        // 2-argument form never dereferences a missing argument.
+        final QueryOptions options = Query.parseOptions(this, contextSequence, null, 3);
```
duncdrum, line 110: `Query.parseOptions(this, contextSequence, null, 3)` re-evaluates the options expression from the AST via `getArgument(2).eval(...)` — that's designed for `Function` subclasses that manage their own argument evaluation. `QueryScope` is a `BasicFunction`, so `args[2]` is already evaluated by the framework before `eval()` runs; calling `parseOptions` here evaluates the options argument a second time. If it has side effects (or is non-deterministic), the discarded first evaluation and the used second one can diverge. `SearchScope.parseSpec` a few files over already reads `args[2]` directly — could this reuse that pattern instead?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/LuceneScope.java
```diff
@@ -0,0 +1,109 @@
...
+final class LuceneScope {
+
+    private LuceneScope() {
+    }
+
+    /**
+     * Resolve a sequence of collection or document URIs to a document set. Collection URIs are searched
+     * recursively (including sub-collections); a URI that is not a collection is tried as a single document.
+     */
+    static MutableDocumentSet resolveScope(final BasicFunction fn, final Sequence scope) throws XPathException {
```
duncdrum, line 62: This reimplements `ExtCollection.getCollectionItems()` (`fn:collection()`'s resolution) rather than sharing it, which means:
- it doesn't check `context.inProtectedMode()` / `context.getProtectedDocs()`, so a caller that's pre-locked a frozen document set (e.g. via `RpcConnection.setProtectedDocs`) gets a different, live view here than `fn:collection()` would give in the same request;
- a future fix to permission/locking behavior in `ExtCollection` won't propagate here (or vice versa).

Is there a reason this couldn't delegate to the existing collection-resolution path instead of reimplementing it?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/LuceneScope.java
```diff
@@ -0,0 +1,109 @@
...
+        return docs;
+    }
+
+    /**
+     * Run the index-first query over {@code docs}. {@code queryArg} is either a Lucene query string or an
+     * XML query element (an empty query matches all indexed nodes in scope).
+     *
+     * @return the matching nodes, each carrying its Lucene score and matches.
+     */
+    static NodeSet query(final BasicFunction fn, final Sequence contextSequence, final MutableDocumentSet docs,
```
duncdrum, line 93: `contextSequence` isn't read anywhere in this method body. Both callers (`QueryScope`, `SearchScope`) thread a real `Sequence` through for no effect, which is a little misleading — reads as if the surrounding XPath context narrows the index-first query, but it doesn't. Drop the parameter?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
```diff
@@ -0,0 +1,426 @@
...
+        result.add(new StringValue(this, "hits"), new ArrayType(this, context, List.of()));
+        final MapType facets = new MapType(this, context);
+        for (final String dimension : spec.facets()) {
+            facets.add(new StringValue(this, dimension), new MapType(this, context));
+        }
+        result.add(new StringValue(this, "facets"), facets);
+        return result;
+    }
+
+    private static double scoreOf(final NodeProxy proxy) {
```
duncdrum, line 364: This re-implements `Score.java`'s `ft:score` summation loop (walk the `Match` chain, sum `LuceneMatch.getScore()` where `getIndexId()` matches `LuceneIndex.ID`) verbatim. If `Score.java`'s scoring model changes later, `ft:score` and `search-scope`'s `"score"` field will silently drift apart. Could this delegate to the existing logic instead?

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
```diff
@@ -0,0 +1,426 @@
...
+        if (filter == null || filter.isEmpty() || !Type.subTypeOf(filter.itemAt(0).getType(), Type.MAP_ITEM)) {
+            return new QueryOptions();
+        }
+        // QueryOptions reads a facet drill-down from a "facets" -> { dimension: value(s) } map; wrap the filter.
+        final MapType queryOptionsMap = new MapType(this, context);
+        queryOptionsMap.add(new StringValue(this, "facets"), filter);
+        return new QueryOptions(queryOptionsMap);
+    }
+
+    private Sequence buildResult(final NodeSet hits, final ResultSpec spec) throws XPathException {
```
duncdrum, line 195: `SearchScope` reimplements a fair amount of ES-`_search`-shaped result assembly in Java — pagination, collapse-by-document, faceting, highlighting (~250 lines). Per the doc comment on `QueryScope`, `ft:score`/`ft:facets`/`ft:field`/`ft:highlight-field-matches` already compose on a plain node set. Was a thin `.xqm` layer over `ft:query-scope` plus those primitives considered, so pagination/envelope tweaks wouldn't need a Java rebuild? Not blocking — just want to understand the tradeoff.

duncdrum:
**Field.java** — pre-existing off-by-one, now more reachable

Not introduced by this PR, but worth flagging since it's now reachable through `ft:search-scope`'s new `"highlight"` option too: `highlightMatches`'s trailing-text append uses `currentPos < content.length() - 1`, which should be `currentPos < content.length()`. When the last match ends exactly one character before the end of the content, the last character is silently dropped from the `exist:field` output. Since this PR already changes this method's signature (adding `context`), might be worth a one-line fix while you're in there.

duncdrum:
**Query.java** — root cause of the score-loss bug isn't fixed

The PR description frames `ft:query-scope`/`ft:search-scope` partly as a workaround for `ft:query`'s score-loss over `//*` descendant-wildcard node sets. That's a reasonable new primitive on its own merits, but it leaves the original bug in `Query.java`/`LuceneHitCollector`'s ancestor resolution unfixed for existing `ft:query` callers — they'd need to migrate to get correct scores. Is a narrower fix at the source (e.g. passing `contextSet=null` when the existing wildcard/`allIndexes.size()>1` heuristic detects an unscoped qname) off the table, or just out of scope for this PR?

duncdrum (changes requested):
some questions, but also blockers

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Fixed in 4d9000fa40 with the `long` computation you suggested. `ss:limit-at-integer-maximum` fails without it.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/LuceneScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Fixed: a malformed scope URI now raises `FODC0004`, the code `fn:collection` uses for an invalid argument, instead of the raw `IllegalArgumentException`. Test: `si:malformed-scope-uri`.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/QueryScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Right. `ft:query-scope` now reads `args[2]`, and I split the map-or-element conversion out of `Query.parseOptions` as `Query.toQueryOptions`, so `ft:query` and `ft:query-scope` still share it.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/LuceneScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Good catch on protected mode. It now does what `ExtCollection` does: in protected mode, a collection scope is resolved from `getProtectedDocs().getDocsByCollection(...)`, which includes sub-collections.

I didn't delegate outright, because `ExtCollection`'s resolution is private and does a different job. It returns document nodes rather than a document set, consults dynamically available collections, raises `FODC0002` on a missing collection, and doesn't accept a document URI, which `ft:query-scope` does. If you'd rather extract a shared document-set resolver from `ExtCollection` that both use, I'm happy to do that as a follow-up.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/LuceneScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Dropped from `LuceneScope.query` and both callers.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Done: `Score.scoreOf(NodeProxy)` is now the single implementation, and `ft:score` and `search-scope`'s `"score"` both call it. While in there, `SearchScope`'s field-value reading (which you flagged on #6459) now goes through `Field.getFieldValues` too.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Fair question. The reason for Java was that ranking, collapse-by-document and paging all need the full scored hit set before any of it can be returned. In XQuery, that means materializing and sorting every hit as XDM before taking a page, which is the cost the function exists to avoid on large scopes. That said, the envelope itself (the map shape, key names, highlight wrapping) doesn't need to be in Java. If we want it easier to adjust without a rebuild, a thin `.xqm` that builds the envelope from a Java primitive returning ranked, paged hits would be a reasonable split. I'd propose that as a follow-up rather than in this PR. Does that match what you had in mind?

joewiz:

Fixed, thanks. The condition is now `currentPos < content.length()`. The existing fixture showed it: highlighting `server` in "Install eXist-db on your server." dropped the final period. `si:highlight-keeps-the-last-character` covers it and fails without the fix.

joewiz:

Not off the table. It's out of scope for this PR, and I haven't yet looked into whether passing a null context set under the wildcard heuristic is safe for existing `ft:query` callers. `ft:query-scope` stands on its own as an index-first primitive, but existing callers shouldn't have to migrate to get correct scores. I'll open an issue for the root cause so it's tracked separately.

duncdrum (approved):

PR #6459, opened 2026-06-09
[feature] ft:fields: introspect configured Lucene fields/facets in a scope

joewiz:

> **Stacked on #6455.** This branch is based on `feature/lucene-search-index` (#6455), because `ft:fields` reuses that PR's `LuceneScope` scope-resolution helper. Until #6455 merges, the diff here includes its commits; the `ft:fields`-specific changes are the final two commits. Once #6455 lands, this rebases onto `develop` to a clean `ft:fields`-only diff. **Names provisional** — happy to rename with the rest of the family.

## Summary

Adds **`ft:fields($scope) as map(*)*`** — the schema-discovery companion to `ft:query-scope`/`ft:search-scope`. Where they *search* an index scope, this *describes* it: the configured fields, facets, and vector fields and each one's contract. It's the engine for existdb-openapi's `GET /api/search/fields` (a `_mapping`-style "what can I search here, and what is each field's contract?" endpoint) and the field-level-security layer built on top of it.

```xquery
ft:fields("/db/apps") (: => one map per configured field/facet/vector occurrence :)

map { 
    "field": "site-content", 
    "kind": "field", 
    "element": "xqdoc:function",
    "analyzer": "org.apache.lucene.analysis.core.SimpleAnalyzer",
    "type": "xs:string", 
    "returnable": true() 
}
map { 
    "field": "site-app", 
    "kind": "facet", 
    "element": "topic" 
}
map { 
    "field": "site-embedding", 
    "kind": "vector", 
    "element": "doc",
    "dimension": 384, 
    "similarity": "cosine", 
    "model": "all-MiniLM-L6-v2" 
} …
```

## Motivation / why native (not xconf parsing)

Field *names* you can scrape from `collection.xconf`; the **contract** — which analyzer a field uses, its type, whether it's stored/returnable, which element each field is indexed on, and (for vector fields) its dimension, similarity metric, and embedding model — is the resolved `LuceneConfig`, which a parser would have to reconstruct (config inheritance across nested collections, `analyzer-id` resolution, merged qname/wildcard/named indexes). existdb-openapi prototyped this Phase-2 layer with an xconf-parser stand-in and hit exactly those walls; `ft:fields` reads the resolved config via the broker instead. The requirements below come from that prototype against a real 3-producer corpus.

## What it returns, per requirement

- **Permission-agnostic, callable by any user (R1).** It reads the resolved `LuceneConfig` via the broker, *not* `/db/system/config` (admin-only), so a guest call returns the **full** catalog. existdb-openapi then applies field-level security (group→allowed-fields) on top. The stand-in had to wrap config reads in `system:as-user("admin", …)`; this removes that.
- **Distinguishes field vs facet vs vector (R2):** `kind: "field" | "facet" | "vector"` (the same name is often both field and facet — e.g. `site-app` is declared as both a `<facet dimension>` and a `<field name>`).
- **Resolved analyzer, per (field × element) (R3):** the analyzer is reported as the **concrete class**, resolving an `@analyzer="simple"` id to its class *and* unwrapping eXist's per-field `MetaAnalyzer` to the actual default/per-field analyzer. A shared field can be indexed with different analyzers on different elements (`StandardAnalyzer` vs `SimpleAnalyzer`), and that variance is preserved.
- **Resolved `type` + `returnable` (R4):** XDM `type` (default `xs:string`) and the stored/`returnable` flag (default true), with defaults resolved.
- **Element each field/facet is indexed on (R5).**
- **Scope = collection path(s), recursive, inheritance-aware (R6):** the same `$scope` model and resolution (`getLuceneConfig(broker, docs)`) as `ft:search-scope`.
- **One record per (field × element) occurrence (R7):** no pre-dedup, so the per-element analyzer/type variance is visible; the API collapses to a field-level view.
- **Vector-field metadata (R8):** a `<vector-field>` is reported with `kind: "vector"` and three extra keys — `dimension` (`xs:integer`), `similarity` (`"cosine" | "euclidean" | "dot_product"`), and `model` (the embedding model id, present **only** when the field embeds text via `embedding="local"`/`"http"`; absent for a raw-vector field). This lets a discovery-driven vector-search endpoint resolve the embedding model from `ft:fields` alone: the client sends `{text, field}`, the server reads the field's `model`, embeds the query with it, and runs KNN — no client-side model knowledge. (Driven by existdb-openapi#62.)

## What changed (`extensions/indexes/lucene`)

- **`Fields.java`** (`ft:fields`) — walks the resolved `LuceneConfig` and emits a map per field/facet/vector field, reusing `LuceneScope.resolveScope` for the scope→documents→config bridge. The vector branch adds `dimension`/`similarity`/`model`.
- **`LuceneVectorFieldConfig.getModelId()`** — new getter exposing the embedding model id (the `dimension`/`similarity` getters already existed); returns `null` for a raw-vector field.
- **`LuceneConfig.getAllIndexConfigurations()`** — enumerates all index heads (qname `paths` + `wildcardPaths` + `namedIndexes`); the existing `getIndexConfigurations()` returned only qname paths.
- **`LuceneFieldConfig`** — `getType()` / `isStore()` / `isBinary()` getters (the fields were protected).
- **`MetaAnalyzer.getConfiguredAnalyzer(field)`** — exposes the concrete analyzer behind the per-field wrapper for introspection (no behavior change to indexing/search).
- `LuceneModule` registers the signature; tests in `ft-fields.xqm` + the `LuceneQueryScopeTests` runner.

## Test plan

- [x] 22 `ft:fields` XQSuite tests: map shape; field/facet counts; type and returnable (incl. `store="no"`); the **resolved analyzer class** for a default-analyzer field (`StandardAnalyzer`) and an `analyzer-id` field (`SimpleAnalyzer`); facet entry shape; unknown scope → empty.
- [x] **Vector field**: `kind="vector"` count; `dimension` is an `xs:integer` (384); `similarity` (`cosine`); `model` (`all-MiniLM-L6-v2`). The vector field is configured on an element absent from the test doc, so the XQSuite exercises pure config introspection — no vector extension needed at runtime.
- [x] **R1 permission-agnostic**: a guest (`system:as-user`) gets the full catalog (config lives under admin-only `/db/system/config`, but `ft:fields` reads the resolved config via the broker).
- [x] 62 tests across the query-scope suite, all green.
- [x] Codacy/PMD clean on the new/changed files.

## Related

- #6455 — the parent `ft:query-scope`/`ft:search-scope` PR this stacks on.
- existdb-openapi field-discovery + FLS design and the `feat/api-search-fields-discovery` prototype this is the drop-in engine for; existdb-openapi#62 (the vector-search endpoint) drives the vector-field metadata keys.


line-o (commented):
I think adding a feature to discover field configurations programmatically is valuable.

Having read the description I ask myself: What are possible security implications when this is merged?
 Because guest is allowed to read:
  1. Is building the map cheap or expensive? Is it a possible denial of service vector?
  2. Shouldn't the permission check happen within the function itself, rather than relying on developers to filter them out?

And two practical questions:
1. The cascade of collection / index configurations means that a field defined in collection `A` can be overwritten / nilled in sub-collection `B`. Is this easy to discover in the returned data?
2. Is there a way to retrieve defined indexes (ngram / range / full-text) that plays nicely together with this new functionality? Should it be added or aligned if this is not the case?

joewiz:

Thanks, these are exactly the right questions — two of them turned into code changes (pushed as 357ae9fb3d), one into a wording fix, and one into a roadmap note.

**1. Cost / DoS.** Good catch. The scope was being resolved through `LuceneScope.resolveScope`, which enumerates every *document* in the subtree under read-locks — but `ft:fields` only ever needs the *collection* hierarchy to read the resolved config. So a guest calling `ft:fields("/db")` forced a full recursive document walk for no reason. I added `LuceneScope.resolveScopeCollections`, which resolves the scope to just its collection set (recursive, permission-checked), so the cost is now bounded by the number of collections, not documents. As a bonus it fixes a latent miss: a configured-but-empty collection used to contribute no fields (no documents meant it never appeared), and now it does.

**2. Permission check inside the function.** You're right that "rely on developers to filter" is the wrong default — and it turns out the original description oversold the openness. The function is *not* permission-agnostic: the scope is resolved through permission-checked collection access (`openCollection` with a read lock; unreadable sub-collections are skipped, mirroring `Collection.allDocs`), so a caller only ever discovers configurations for collections it can already read. I've corrected the Javadoc and the function description to say "gated at collection-read granularity" rather than "permission-agnostic". What the function deliberately does *not* do is finer, per-field filtering — and that's by necessity rather than choice: eXist has no field-level permission primitive, so a group→allowed-fields policy is something only the application knows. Within a readable collection it returns the full configured set; the app layer applies any field-level policy on top. The one genuinely open aspect is that it surfaces the *resolved config* (analyzers, models) for a collection whose data you can read but whose admin-only `/db/system/config` resource you can't — which is the whole point (avoid requiring admin), and is low-disclosure since data-read already lets you infer most of the indexing.

**3. Cascade discoverability.** Also a real gap — the output had no way to tell which collection a field's config came from, so an override in a sub-collection was invisible. Each result map now carries a `collection` key reporting the collection its config resolved from. A field defined in a parent and overridden (or nilled) in a sub-collection now shows up as records under differing collection paths, so the cascade is discoverable. New tests cover the single-collection attribution and the multi-collection case where two sub-collections contribute differently-named fields under their own paths.

**4. Other index types (ngram / range / full-text).** Today `ft:fields` is Lucene-only by construction — it reads the resolved `LuceneConfig`, so it covers Lucene fields, facets, and vector fields. The ngram and range indexes are separate index modules with their own configuration and no introspection function yet. I agree aligning them is valuable; rather than expand this PR, I think the right move is to settle the namespace and map shape now so a future `range`/`ngram` introspection composes with this one (the names here are still provisional anyway). Happy to open a tracking issue to capture that design question if that's useful — or fold it into the broader scope-functions naming discussion.

duncdrum (approved):

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
```diff
@@ -0,0 +1,426 @@
...
+            total = ranked.size();
+        } else {
+            ranked = new ArrayList<>(all);
+            total = all.size();
+        }
+        ranked.sort((a, b) -> Double.compare(b.score(), a.score()));
+
+        // page the ranked list: skip $offset, then take at most $limit
+        final int from = Math.min(spec.offset(), ranked.size());
+        final int to = (spec.limit() >= 0) ? Math.min(ranked.size(), from + spec.limit()) : ranked.size();
```
duncdrum, line 227: `from + spec.limit()` can overflow when `limit` is `Integer.MAX_VALUE` (a common "unlimited" sentinel), wrapping `to` to a large negative number while `from` stays positive. `ranked.subList(from, to)` then throws `IndexOutOfBoundsException` instead of returning results or a clean XPath error. Worth clamping the sum (e.g. via `Math.min` on a `long` computation) before calling `subList`.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Fields.java
```diff
@@ -0,0 +1,236 @@
...
+        // it would silently drop the rest. Union each distinct collection config so ft:fields($scope)
+        // discovers the full field set, the way ft:search-scope aggregates documents across the scope.
+        // LuceneConfig identity dedup means an inherited config is reported once, attributed to the
+        // collection it is defined on (parents are visited before children), while a sub-collection that
+        // overrides the config carries a distinct LuceneConfig and is reported under its own path.
+        final DBBroker broker = context.getBroker();
+        final ValueSequence result = new ValueSequence();
+        final Set<LuceneConfig> seen = Collections.newSetFromMap(new IdentityHashMap<>());
+        for (final XmldbURI collectionUri : collections) {
+            try (final Collection collection = broker.openCollection(collectionUri, LockMode.READ_LOCK)) {
```
duncdrum, line 147: Nit/efficiency: `resolveScopeCollections()` already traverses and locks every collection in scope to build the `XmldbURI` set, and this re-opens/re-locks each one again to fetch its `LuceneConfig`. Could the first traversal collect the `LuceneConfig` directly and avoid the second pass? Not blocking, but doubles lock churn for scopes with many sub-collections.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
```diff
@@ -0,0 +1,426 @@
...
+
+    private Sequence fieldValues(final LuceneIndexWorker index, final int docId, final NodeId nodeId,
+                                 final String field) throws XPathException {
+        try {
+            final IndexableField[] indexed = index.getFieldByExistDocId(docId, nodeId, field);
+            final ValueSequence values = new ValueSequence(indexed.length);
+            for (final IndexableField f : indexed) {
+                final String s = f.stringValue();
+                values.add(new StringValue(this, s != null ? s
+                        : (f.numericValue() != null ? f.numericValue().toString() : "")));
```
duncdrum, line 301: Minor: `f.numericValue()` is called twice here (`!= null` check, then `.toString()`). Caching it in a local avoids the redundant getter call on this per-hit path.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
```diff
@@ -0,0 +1,426 @@
...
+        }
+        final Sequence text = fieldValues(index, docId, proxy.getNodeId(), field);
+        try {
+            return Field.highlightMatches(context, field, proxy, match, text);
+        } catch (final IOException e) {
+            throw new XPathException(this, LuceneModule.EXXQDYFT0002, "Error highlighting field '" + field + "': " + e.getMessage());
+        }
+    }
+
+    private Sequence fieldValues(final LuceneIndexWorker index, final int docId, final NodeId nodeId,
```
duncdrum, line 293: This re-implements the field-value extraction that `Field.getFieldValues()` already does (numeric vs. string branching). Could this delegate to that instead, the way `highlightMatches` already does elsewhere in this PR? Otherwise a future fix to value extraction would need to be applied in two places.

duncdrum (changes requested):
Thanks for this, `ft:fields` is a nice addition! Left a few inline notes — one likely crash to fix before merge, plus a few smaller efficiency/reuse suggestions that aren't blocking.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Fixed in #6455, the parent PR this one stacks on (4d9000fa40). This branch is rebased onto it.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/Fields.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Done in d7ae866bc4. `LuceneScope.resolveScopeConfigs` now reads each collection's Lucene configuration under the read lock the traversal already holds, so `ft:fields` doesn't reopen anything. It also raises `FODC0004` for a malformed URI, as `ft:query-scope` now does.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Gone. That method now delegates to `Field.getFieldValues` (next comment), which reads each value once. The change is in #6455, since that's where the code lives.

extensions/indexes/lucene/src/main/java/org/exist/xquery/modules/lucene/SearchScope.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Done in #6455 (4d9000fa40): `Field.getFieldValues` is now package-private static, and `SearchScope` uses it.

duncdrum (approved):

PR #6462, opened 2026-06-10
[bugfix] eliminate port-binding TOCTOU race in test servers

duncdrum:
`ExistWebServer` probed for a free port by opening and immediately closing a ServerSocket, then passed that port number to Jetty. Between the probe and Jetty's `bind()` another fork JVM (forkCount=2C) could claim the same port, producing sporadic "Failed to bind" failures in the WebDAV test suite and anywhere else ExistWebServer is used.

Fix seems almost too simple just set port to `0` and let the OS handle it.

### DIFF
--- a/extensions/modules/mail/src/test/java/org/exist/xquery/modules/mail/SendEmailIT.java
+++ b/extensions/modules/mail/src/test/java/org/exist/xquery/modules/mail/SendEmailIT.java
@@ -73,7 +73,6 @@ import java.util.Arrays;
 import java.util.Optional;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.exist.util.IPUtil.nextFreePort;
 import static org.exist.xquery.modules.mail.Util.executeQuery;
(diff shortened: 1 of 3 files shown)


extensions/modules/mail/src/test/java/org/exist/xquery/modules/mail/SendEmailIT.java
```diff
@@ -123,10 +122,8 @@ public static java.util.Collection<Object[]> data() {
     private static final String EMAIL_UID = "emailuid";
     private static final String EMAIL_PWD = "emailpwd";
 
-    private final int smtpPort = nextFreePort(2525, 2599, 10);
```
line-o, line 126: smptPorts might need a very different handling than HTTP servers

line-o: I just realized this is within an integration test. Should be fine.

line-o (approved):
Cool!

dizzzz (approved):

PR #6466, opened 2026-06-11
[feature] response:stream-binary-resource — zero-copy binary download

joewiz:

## Summary

Adds `response:stream-binary-resource($binary-resource-path, $content-type, $filename?)`, which streams a stored binary resource straight from the database to the servlet response output stream — **without ever materializing it in the JVM heap.**

`response:stream-binary($binary, …)` takes a fully-loaded `xs:base64Binary`, so the caller must first read the whole resource into memory (e.g. `util:binary-doc()`). For a large download that means the entire resource is materialized on the heap before a single byte goes out. The new function opens the stored `BinaryDocument` and copies it directly to the response via `broker.readBinaryResource(Txn, BinaryDocument, OutputStream)` — the **same zero-copy path `RESTServer` already uses** for binary downloads (`RESTServer.java:1782–1783`).

## Signature

```
response:stream-binary-resource(
    $binary-resource-path as xs:string,
    $content-type         as xs:string,
    $filename            as xs:string?   (: optional — sets Content-Disposition :)
) as empty-sequence()
```

## How it relates to the existing `response:` streamers

| function | takes | mechanism | memory |
|---|---|---|---|
| `response:stream($node, $opts)` | a node | serializes XML/HTML to the output stream | n/a |
| `response:stream-binary($binary, $type, $f?)` | an **already-materialized** `xs:base64Binary` | `binary.streamBinaryTo(os)` | caller must load the whole resource into heap first |
| **`response:stream-binary-resource($path, $type, $f?)`** *(new)* | a **db path** | opens the stored `BinaryDocument`, `broker.readBinaryResource(txn, binDoc, os)` | **never materializes** — streams disk → socket |

## What changed

- `response/StreamBinaryResource.java` *(new)* — the function. Resolves the path (read lock + permission check, the same way `util:binary-doc` does), verifies it's a binary document, sets `Content-Type` (and `Content-Disposition` from `$filename`), and streams zero-copy within a transaction.
- `response/ResponseModule.java` — registers the two arities.
- `RestBinariesTest.java` — `streamBinaryResourceRaw` stores a binary and asserts the bytes returned over a real HTTP request are **byte-identical** (mirrors the existing `streamBinaryRaw` test).

## Test

`RestBinariesTest` — **4/4 green** (incl. the new byte-identical case). exist-core builds clean; Codacy PMD clean.

## Scope: this is db paths only (`file:` is deliberately out of scope)

The function operates on **database** binary resources, not local filesystem files. Nothing inherent prevents a `file:` variant — streaming a local file is actually *simpler* (no broker, lock, transaction, or `BinaryDocument`: just `Files.newInputStream(path).transferTo(os)`). It's left out on purpose, for two reasons:

1. **Security.** Filesystem access from a request-handling function is a file-disclosure / path-traversal surface — an ungated `response:stream-binary-resource('file:///etc/passwd', …)` would stream arbitrary server files to a client. eXist treats `file:` access from XQuery as a **DBA-only** boundary, so a `file:` branch would need that guard and warrants its own scrutiny.
2. **Focus + consistency.** Keeping this PR to the db case keeps the security-sensitive surface separate, and lets a future `file:` variant follow the convention PR #6192 is establishing for `fn:collection` rather than inventing a parallel one.

**If `file:` support were wanted**, the shape would mirror #6192's `fn:collection`: branch on the URI scheme (db vs `file:`), **restrict the `file:` branch to DBA users** (a security boundary, consistent with `fn:doc()` / #6192), and stream via `Files.newInputStream(...).transferTo(response.getOutputStream())` — zero-copy, no materialization. That fills a real gap (the EXPath `file:read-binary` materializes the whole file), but it belongs in its own PR.

## Context

This is the download half of the binary-streaming work surfaced by the existdb-openapi binary-transport track (eXist-db/existdb-openapi#35 / #38). Functional binary transport already works today via `response:stream-binary`; this is the **scale** primitive for large downloads. The upload counterpart (a non-materializing `request:get-input-stream()` so a handler can pipe a large upload to `broker.storeDocument`) is deliberately a **separate follow-up** — it depends on how the request body is consumed (single-read; the Roaster raw-body path) and needs its own investigation.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/response/StreamBinaryResource.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/response/StreamBinaryResource.java
@@ -1,0 +1,155 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.functions.response;
+
+import org.exist.dom.QName;
+import org.exist.dom.persistent.BinaryDocument;
+import org.exist.dom.persistent.DocumentImpl;
+import org.exist.dom.persistent.LockedDocument;
+import org.exist.http.servlets.ResponseWrapper;
+import org.exist.security.PermissionDeniedException;
+import org.exist.storage.DBBroker;
+import org.exist.storage.lock.Lock.LockMode;
+import org.exist.storage.txn.Txn;
+import org.exist.storage.txn.TransactionException;
+import org.exist.xmldb.XmldbURI;
+import org.exist.xquery.Cardinality;
+import org.exist.xquery.ErrorCodes;
+import org.exist.xquery.FunctionSignature;
+import org.exist.xquery.XPathException;
+import org.exist.xquery.XQueryContext;
+import org.exist.xquery.value.FunctionParameterSequenceType;
+import org.exist.xquery.value.Sequence;
+import org.exist.xquery.value.SequenceType;
+import org.exist.xquery.value.Type;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.URISyntaxException;
+
+import javax.annotation.Nonnull;
+
+/**
+ * {@code response:stream-binary-resource($binary-resource-path, $content-type, $filename?)} streams a
+ * stored binary resource straight from the database to the servlet response output stream.
+ *
+ * <p>Unlike {@code response:stream-binary}, which takes a fully-materialized {@code xs:base64Binary}
+ * (so the caller must first load the whole resource into the JVM heap, e.g. via {@code util:binary-doc}),
+ * this function copies from the stored {@link BinaryDocument} directly to the response — the same
+ * zero-copy path the REST server uses — without ever materializing the bytes. It is intended for large
+ * downloads, where materializing the whole resource would exhaust the heap.</p>
+ */
+public class StreamBinaryResource extends StrictResponseFunction {
+
+    private static final FunctionParameterSequenceType RESOURCE_PATH_PARAM = new FunctionParameterSequenceType("binary-resource-path", Type.STRING, Cardinality.EXACTLY_ONE, "The path to the stored binary resource, e.g. /db/path/to/image.png");
+    private static final FunctionParameterSequenceType CONTENT_TYPE_PARAM = new FunctionParameterSequenceType("content-type", Type.STRING, Cardinality.EXACTLY_ONE, "The ContentType HTTP header value");
+    private static final FunctionParameterSequenceType FILENAME_PARAM = new FunctionParameterSequenceType("filename", Type.STRING, Cardinality.ZERO_OR_ONE, "The filename. If provided, a Content-Disposition header is set for the filename in the HTTP Response");
+
+    private static final String DESCRIPTION =
+            "Streams a stored binary resource directly to the current servlet response output stream, "
+            + "copying from the database to the response without first materializing the whole resource "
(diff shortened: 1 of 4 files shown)


line-o:
That is BIG!

exist-core/src/main/java/org/exist/xquery/functions/response/StreamBinaryResource.java
```diff
@@ -0,0 +1,155 @@
...
+            final String filename, final ResponseWrapper response) throws XPathException {
+        final DBBroker broker = context.getBroker();
+        try (final LockedDocument lockedDoc = broker.getXMLResource(uri, LockMode.READ_LOCK)) {
+            if (lockedDoc == null) {
+                throw new XPathException(this, ErrorCodes.FODC0002, "Binary resource not found: " + path);
+            }
+
+            final DocumentImpl doc = lockedDoc.getDocument();
+            if (doc.getResourceType() != DocumentImpl.BINARY_FILE) {
+                throw new XPathException(this, "Resource is not a binary document: " + path);
```
line-o: An ErrorCodes would be good if we can find a fitting one. Looks like a type error to me.

exist-core/src/main/java/org/exist/xquery/functions/response/StreamBinaryResource.java
```diff
@@ -0,0 +1,155 @@
...
+                os.close();
+                transaction.commit();
+            }
+
+            // commit the response
+            response.flushBuffer();
+        } catch (final PermissionDeniedException e) {
+            throw new XPathException(this, ErrorCodes.XPDY0002, "Permission denied to read binary resource '" + path + "': " + e.getMessage(), e);
+        } catch (final TransactionException e) {
+            throw new XPathException(this, "Transaction error while streaming binary resource '" + path + "': " + e.getMessage(), e);
```
line-o: Do we have ErrorCodes for this kind of issue? Maybe in EXERR?

exist-core/src/main/java/org/exist/xquery/functions/response/StreamBinaryResource.java
```diff
@@ -0,0 +1,155 @@
...
+        } catch (final IOException e) {
+            throw new XPathException(this, "IO exception while streaming binary resource '" + path + "': " + e.getMessage(), e);
```
line-o: see above

exist-core/src/main/java/org/exist/xquery/functions/response/StreamBinaryResource.java
```diff
@@ -0,0 +1,155 @@
...
+
+        if (!"org.exist.http.servlets.HttpResponseWrapper".equals(response.getClass().getName())) {
+            throw new XPathException(this, ErrorCodes.XPDY0002, getSignature().toString() + " can only be used within the EXistServlet or XQueryServlet");
+        }
+
+        final XmldbURI uri;
+        try {
+            uri = XmldbURI.xmldbUriFor(path);
+        } catch (final URISyntaxException e) {
+            throw new XPathException(this, "Invalid binary resource path: " + path, e);
```
line-o: I would like to have a fitting Error code here, too

exist-core/src/test/java/org/exist/xquery/RestBinariesTest.java
```diff
@@ -114,6 +114,26 @@ public void streamBinaryRaw() throws JAXBException, IOException {
...
     }
 
+    /**
+     * response:stream-binary-resource streams a stored binary by path, without materializing it
+     * (unlike response:stream-binary which takes a fully-loaded xs:base64Binary). The wire result
+     * is byte-identical.
+     */
+    @Test
+    public void streamBinaryResourceRaw() throws JAXBException, IOException {
+        final String query = "import module namespace response = \"http://exist-db.org/xquery/response\";\n" +
```
line-o, line 127: I would like to see one test that ensure the content-disposition header is sent for `response:stream-binary-resource#3` with parameter 3 being set.

joewiz:

Thanks @line-o — all addressed in df06128c89.

Every failure path now carries a fitting error code instead of a bare `XPathException`:

- **"Resource is not a binary document" → `XPTY0004`.** You read it right as a type error.
- **Invalid resource path (`URISyntaxException`) → `FODC0002`** ("Error retrieving resource") — the same code already on the not-found case just below it.
- **Permission denied → `FODC0002`** too. It was previously `XPDY0002`, which is wrong (that code means "dynamic context not assigned"); folding it under `FODC0002` keeps all three "cannot get the resource" conditions on one code.
- **Transaction / IO error while streaming → new `EXXQDY0007`** ("I/O error while streaming a binary resource to the response") — added to the eXist error namespace, which is the `EXERR`-family fit you pointed at. The transaction and IO catches share it.

I left the servlet-context guard on `XPDY0002`, since "used outside a servlet response" genuinely is a missing-dynamic-context case — happy to change it if you'd prefer otherwise.

For the test: added `streamBinaryResourceWithFilename` to `RestBinariesTest`, which calls the 3-arg `response:stream-binary-resource#3` with a filename and asserts the response carries `Content-Disposition: inline; filename="download.bin"` (and that the body is still byte-identical). Full `RestBinariesTest` green (5/5).

duncdrum (approved):

line-o (approved):
We should add more tests over time. But good for now.

PR #6469, opened 2026-06-12
[bugfix] Compile statically-named arrow operator to a direct function call (closes #3887, #3336, #3885)

### LINKED ISSUE #3336: [BUG] function expressions with ? after arrow operator not spec compliant
**Describe the bug**

The current handling of `?` in function calls after the arrow operator `=>` is not working according to the spec.

Saxon 9 and 10 handle this correctly.
This is likely broken since this feature was implemented but tested on exist versions 4.7.0 and the latest SNAPSHOT.

**Expected behavior**

`EXPR => FUNC(?)` should evaluate to `FUNC(EXPR, ?)` which evaluates to a function with arity 1.

In the current implementation it seems that the question mark is substituted with EXPR instead.
**To Reproduce**

```Xquery
xquery version "3.1";

module namespace aq="http://exist-db.org/xquery/arrow-question";

declare
    %test:assertEquals(1)
function aq:fail() {
    () => sum(?) => function-arity()
};

declare
    %test:assertEquals(1)
function aq:works() {
    function-arity(sum((), ?))
};
```

 - OS: macOS 10.15.4
 - eXist-db version: 5.3.0-SNAPSHOT (build from 093c4c0726e02868acadbce9583ec5809f462b33)
 - Java Version 1.8.0_152-b16

**Additional context**

### LINKED ISSUE #3885: [BUG] NPE with HoF, argument placeholder after arrow expression

**Describe the bug**

The below query raises a NullPointerException

```xquery
xquery version "3.1";

declare namespace theNPE="the/npe";

declare function theNPE:map2 ($a) { $a };
declare function theNPE:map1 ($a as xs:string) {
  tokenize($a, "\|") => for-each(theNPE:map2(?))
};

try { ("a|b", "c", "d|e") => for-each(theNPE:map1(?)) }
catch * { $err:code, $err:description, $err:value }
```

**Expected behavior**

The query to return the sequence `("a", "b", "c", "d", "e")`

**To Reproduce**

```Xquery
xquery version "3.1";

module namespace theNPE="http://exist-db.org/xquery/theNPE";

declare function theNPE:map2 ($a) { $a };
declare function theNPE:map1 ($a as xs:string) {
  tokenize($a, "\|") => for-each(theNPE:map2(?))
};

declare
    %test:assertEquals("a", "b", "c", "d", "e")
function t:test() {
  try { ("a|b", "c", "d|e") => for-each(theNPE:map1(?)) }
  catch * { $err:code, $err:description, $err:value }
};
```

 - OS: macOS 10.15.3
 - Java Version AdoptOpen Java8

### LINKED ISSUE #3887: [BUG] util:eval after arrow does not have access to variables in scope
```xquery
xquery version "3.1";

let $v := 1
return util:eval(concat("$v", "+1"))
```

yields 2

But

```xquery
xquery version "3.1";

let $v := 1
return concat("$v", "+1") => util:eval()
```

will throw an error that `$v` is not known

**Expected behavior**

Both ways of writing the query to produce the same result `2`.

**To Reproduce**

```Xquery
xquery version "3.1";

declare variable $t:v := 1;

declare
    %test:assertEquals(2)
function t:concat-in-eval() {
  util:eval(concat("$v", "+1"))
};

declare
    %test:assertEquals(2)
function t:concat-arrow-eval() {
  concat("$v", "+1") => util:eval()
};
```

 - OS: macOS 10.15.x
 - Java Version AdoptOpen Java 8

**Additional context**

joewiz:

## Summary

The XQuery 3.1 arrow operator `EXPR => f(args)` is, per spec, exactly the call `f(EXPR, args)`. eXist instead evaluated **every** arrow — including the statically-named case — by building a dynamic `FunctionReference`: it pre-evaluated the left-hand side, injected it as a placeholder (`ContextParam extends Function.Placeholder`), then re-analyzed and dynamically dispatched on each evaluation. That caused three bugs, a conformance gap, and a 2–3× per-call performance penalty.

This compiles the statically-named arrow to a real `FunctionCall` — the same thing the parser builds for an ordinary call — which fixes all of the above. A genuine higher-order right-hand side (a variable or parenthesized expression yielding a function) still uses the dynamic path.

## What changed

- **`exist-core/.../ArrowOperator.java`** — `analyze()` now compiles the named arrow via `FunctionFactory.createFunction(qname, [leftExpr, parameters…])`, wrapping in `PartialFunctionApplication` when a `?` placeholder is present (mirroring the parser's `functionCall` rule). `eval()` for the named case simply delegates to that call, so the left-hand side is evaluated as an ordinary argument in the call's own context. The higher-order (`funcSpec`) path is preserved.
- **`exist-core/.../parser/XQueryTree.g`** — the `arrowOp` tree-walker rule had **no `QUESTION → Function.Placeholder` branch** (unlike `functionCall`), so a `?` in an arrow argument list was never turned into a placeholder. Added the branch (the grammar half of #3336).
- **`exist-core/.../xquery3/arrowop.xql`** — three new regression tests (#3887, #3336, #3885) and one updated assertion (see *Conformance* below).
- **`exist-core-jmh/.../ArrowOperatorBenchmark.java`** *(separate `[test]` commit)* — a JMH benchmark comparing arrow vs. the equivalent direct call.

## Bugs fixed

Because the left-hand side is now a genuine argument expression instead of a pre-evaluated placeholder:

- **Closes #3887** — a context-sensitive callee sees in-scope variables, like the direct call. `let $v := 1 return concat("$v","+1") => util:eval()` returned `XPST0008 Variable '$v' is not declared`; now returns `2`, matching `util:eval(concat("$v","+1"))`.
- **Closes #3336** — `EXPR => f(?)` is a partial application of the remaining arity. `('hello' => substring(?))(2)` returned `XPST0017` (wrong arity); now returns `"ello"`.
- **Closes #3885** — no `NullPointerException` for a nested partial application after the arrow (`… => for-each(f(?))`).

## Conformance — one existing test assertion changed (please review)

The fix also makes the arrow apply **function-conversion rules** like the equivalent direct call. On `develop`, the arrow was *more lenient* than a direct call:

```xquery
declare function local:sj($s as xs:string*, $sep as xs:string) { string-join($s, $sep) };
local:sj((1,2,3), "-")        (: develop: XPTY0004 — integers are not xs:string :)
(1,2,3) => local:sj("-")      (: develop: wrongly returned "1-2-3" :)
```

Saxon raises `XPTY0004` for both forms; so does eXist's direct call. After this change the arrow agrees. The existing `ao:type-checks-user-func` test asserted the old lenient `"1-2-3"`; it now asserts `XPTY0004` (with a comment explaining why). This is the only behavior change beyond the three bug fixes, and it brings the arrow into line with the spec, the direct call, and Saxon.

## Spec references

- XQuery 3.1 — [3.1.5 Arrow Operator](https://www.w3.org/TR/xquery-31/#id-arrow-operator): *"An ArrowExpr … is equivalent to … a function call."*
- XPath/XQuery 3.1 — [3.1.5.1 Evaluating Static and Dynamic Function Calls](https://www.w3.org/TR/xquery-31/#id-function-calls) / the function conversion rules; `XPTY0004` for an argument that does not match the required type.

## Reference implementations

Both lower the named arrow to a static call and agree on the behavior:

- **Saxon 12.5** (observed): `('hello' => substring(?))(2)` → `ello`; `(1 to 5) => reverse()` → `5 4 3 2 1`; `(1,2,3) => local:sj("-")` (into `xs:string*`) → `XPTY0004`.
- **BaseX** (source, `QueryParser.arrow()`): a named function becomes a static `Functions.get(name, [LHS, args…])` with the LHS as a real argument expression; the dynamic path is reserved for a function-valued RHS.

## Performance (JMH, 100k iterations, 5 measurement iterations)

| Benchmark | Before | After | |
|---|---|---|---|
| arrow single call | 55.4 ± 4.1 ms | 25.0 ± 1.7 ms | **2.2× faster** |
| arrow 3-arrow chain | 110.2 ± 7.5 ms | 34.1 ± 4.4 ms | **3.2× faster** |
| direct single *(control)* | 24.3 ± 0.4 ms | 24.3 ± 0.4 ms | flat |
| direct chain *(control)* | 34.7 ± 4.6 ms | 34.8 ± 3.2 ms | flat |

After the fix the arrow is at parity with the equivalent direct call; the direct-call controls are unchanged, confirming the change touches only the arrow path.

## W3C XQTS (HEAD) — before/after on the affected test sets

Run via the same in-repo `exist-xqts` job this PR's CI runs.

| test set | before (develop) | after |
|---|---|---|
| **`prod-ArrowPostfix`** | 41/42 (1 fail) | **42/42** |
| `prod-SequenceType` | 21/21 | 21/21 |
| `misc-HigherOrderFunctions`, `prod-FunctionCall`, `prod-NamedFunctionRef`, `prod-InlineFunctionExpr`, `fn-function-*` | (pre-existing fails) | **identical** |

`ArrowPostfix-108` (a placeholder-after-arrow `=> concat(?, …)` case) failed on `develop` with `XPST0017 fn:concat requires at least two arguments` and now passes. Every other set's failure/error/skip counts are byte-identical before and after — no collateral movement.

## Test plan

- [x] New regression tests in `arrowop.xql` for #3887, #3336, #3885; updated type-conversion assertion.
- [x] `xquery.xquery3.XQuery3Tests` (includes `arrowop.xql`) green: 1,059 tests.
- [x] 2,610 XQSuite tests green across maps / arrays / xquery3 / core / inspect / util / optimizer / numbers.
- [x] W3C XQTS HEAD: `prod-ArrowPostfix` 41→42, no regression in the function/HoF/sequence-type sets.
- [x] JMH benchmark: arrow now at parity with the direct call (2.2–3.2× faster than before).
- [x] Codacy PMD clean on the changed Java.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/ArrowOperator.java
+++ b/exist-core/src/main/java/org/exist/xquery/ArrowOperator.java
@@ -23,6 +23,7 @@ package org.exist.xquery;
 
 import org.exist.dom.QName;
 import org.exist.dom.QName.IllegalQNameException;
+import org.exist.xquery.parser.XQueryAST;
 import org.exist.xquery.util.ExpressionDumper;
 import org.exist.xquery.value.FunctionReference;
 import org.exist.xquery.value.Item;
@@ -41,7 +42,7 @@ public class ArrowOperator extends AbstractExpression {
 
     private QName qname = null;
     private Expression leftExpr;
-    private FunctionCall fcall = null;
+    private Expression namedCall = null;
     private Expression funcSpec = null;
     private List<Expression> parameters;
     private AnalyzeContextInfo cachedContextInfo;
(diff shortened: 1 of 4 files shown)


duncdrum (approved):

exist-core/src/main/java/org/exist/xquery/ArrowOperator.java
```diff
@@ -74,45 +75,91 @@ public void analyze(final AnalyzeContextInfo contextInfo) throws XPathException
...
-        if (qname != null) {
-            fcall = NamedFunctionReference.lookupFunction(this, context, qname, parameters.size() + 1);
-        }
         this.cachedContextInfo = contextInfo;
-        leftExpr.analyze(contextInfo);
-        if (fcall != null) {
-            fcall.analyze(contextInfo);
-        }
-        if (funcSpec != null) {
+        if (qname != null) {
```
line-o: Please invert the condition and return early.

exist-core/src/main/java/org/exist/xquery/ArrowOperator.java
```diff
@@ -128,44 +175,47 @@ public Sequence eval(Sequence contextSequence, final Item contextItem) throws XP
...
         } else {
             funcSpec.dump(dumper);
         }
+        dumper.display('(');
         for (int i = 0; i < parameters.size(); i++) {
             if (i > 0) {
                 dumper.display(", ");
-                parameters.get(i).dump(dumper);
             }
+            parameters.get(i).dump(dumper);
```
line-o, line 204: Should the first parameter really be dumped here?

exist-core/src/main/java/org/exist/xquery/ArrowOperator.java
```diff
@@ -128,44 +175,47 @@ public Sequence eval(Sequence contextSequence, final Item contextItem) throws XP
...
-            fcall.resetState(postOptimization);
-        }
-        if (funcSpec != null) {
-            funcSpec.resetState(postOptimization);
-        }
-        for (Expression param: parameters) {
-            param.resetState(postOptimization);
+        if (namedCall != null) {
+            // namedCall owns leftExpr and parameters as its arguments.
+            namedCall.resetState(postOptimization);
```
line-o, line 214: Return here so that the following lines do not need to be indented, wrapped in else

exist-core/src/test/xquery/xquery3/arrowop.xql
```diff
@@ -149,8 +151,13 @@ declare %private function ao:string-join($s as xs:string*, $sep as xs:string) {
...
 };
 
+(: EXPR => f(args) is exactly the call f(EXPR, args) and applies the same function conversion
+   rules: xs:integer items do not convert to a declared xs:string* parameter, so this raises
+   XPTY0004 — identical to the direct call ao:string-join((1,2,3), "-") and to Saxon. (Before the
+   arrow was compiled to a direct call, the dynamic FunctionReference path was wrongly lenient here
+   and returned "1-2-3", which a normal call never would.) :)
 declare
-    %test:assertEquals("1-2-3")
+    %test:assertError("XPTY0004")
```
line-o, line 160: That is correct XQuery 3.1 behaviour
I would like to see an additional test that changes the input to strings or that the numbers are cast to strings inside the function on the RHS of the function.

line-o (changes requested):
Awesome!👏 
I found small improvements I believe will help us maintain the code in the future (see my comments). Please also reformat the Java code afterwards. There is at least on if-condition with the curly brace on the next line.

joewiz:

Thanks Juri — all addressed. Pushed two follow-up commits (one `[refactor]`, one `[test]`).

- **`analyze()` — invert and return early:** done. The short dynamic (higher-order) case is handled first with an early `return`, so the larger statically-named-call block is no longer nested in an `else`.
- **`resetState()` — return early:** done, same shape — the `namedCall` branch returns, and the dynamic branch follows un-indented.
- **Brace style:** fixed the `if (…) {` on the cardinality check in `eval()` (it was brace-on-next-line in the original code I carried over). I re-ran Codacy/PMD on `ArrowOperator.java` — clean.
- **`dump()` — "should the first parameter really be dumped here?":** yes, and good eye — this actually fixes a latent bug. `parameters` holds only the explicit right-hand-side arguments (the `a, b` in `EXPR => f(a, b)`); the arrow's left operand is the separate `leftExpr` field, which is dumped before the `=>`. The previous loop dumped `parameters.get(i)` *only* inside the `if (i > 0)` block, so it silently dropped the first RHS argument and emitted a stray leading comma (`f(, b)`). Moving the `dump` call out of that block dumps every argument with correct comma separation. (Happy to add a `dump()` assertion if you'd like it pinned down, though the output is debug-only.)
- **Additional positive test:** added two companions to `ao:type-checks-user-func` (which asserts `XPTY0004`). `ao:type-checks-user-func-strings` passes `("1","2","3")` and expects `"1-2-3"`; `ao:type-checks-user-func-cast-inside` keeps integer input but casts to string inside an `xs:anyAtomicType*` RHS function, also `"1-2-3"`. Both confirm the arrow is sound when the types line up — the `XPTY0004` is purely the conversion-rule mismatch, identical to the equivalent direct call.

Full XQuery3 suite (1061 tests) green after the changes.

line-o (approved):
Good to go! Finally this is solved and we can use arrows liberally 😊 

My previous tests measured the penalty to using arrows vs other language constructs at about halving the speed. This nicely aligns with your measurements.

PR #6470, opened 2026-06-12
[bugfix] Log a deferred (tail-call) function's error once, not once per Sequence accessor

joewiz:

## Summary

An error raised inside a **deferred (tail-call) function** is logged repeatedly — once for every `Sequence` accessor a consumer calls on the result — and because each line carries the full call stack, a single failure becomes many long lines in the log. This is the cause of the "epically long" error output people see from the HTML-templating framework (and tools built on it). This PR makes the failure log **once**, with no change to evaluation behavior.

## Background — what a "deferred function" is

A tail-recursive XQuery function isn't recursed on the JVM stack; `FunctionCall` returns a `DeferredFunctionCall` (eXist's tail-call trampoline) whose body runs lazily. The HTML-templating framework's recursive model processing is a common way to produce these.

`DeferredFunctionCall` runs its body once, caches any exception, and re-throws the cached exception on every later access. It has about fourteen **non-throwing** `Sequence` accessor methods — `getCardinality`, `isEmpty`, `hasOne`, `hasMany`, `getItemCountLong`, `itemAt`, `getItemType`, `getDocumentSet`, `isPersistentSet`, … — none of which is declared to throw the checked `XPathException`. Each catches the exception and returns a default (empty / `null` / `false` / `0`), in an identical catch block.

## The bug

Every one of those ~14 catch blocks also **logged** the error:

```java
} catch (XPathException e) {
    caughtException = e;
    LOG.error("Exception in deferred function: {}", e.getMessage());   // logged every time
    return <default>;
}
```

A consumer that touches the same deferred value through several accessors — the serializer and the templating engine both do (`isEmpty`, `hasOne`, `getItemCountLong`, `itemAt`, …) — re-enters the catch once per accessor, and `realize()` re-throwing the cached exception guarantees every later accessor hits it too. So the one failure is logged N times. And since `XPathException.getMessage()` appends the accumulated `In function:` call stack (a frame per function boundary the exception crossed), each of those N lines is long for a deep templating/recursion chain.

## What changed

`DeferredFunctionCall.java` — one private helper, used by every catch block:

```java
private void captureAndLogOnce(final XPathException e) {
    if (caughtException == null) {   // realize() re-throws the cached exception on later access;
        caughtException = e;         // only the first capture logs.
        LOG.error("Exception in deferred function: {}", e.getMessage());
    }
}
```

Each catch becomes `captureAndLogOnce(e)`. **Behavior is otherwise unchanged**: the failure is still cached, the non-throwing accessors still return the same defaults, and the error still surfaces to the caller through the *throwing* accessors (`iterate()`, `getStringValue()`, …). Only the duplicate log lines stop. One file, no public API change.

The "swallow to a default" behavior of the non-throwing accessors is intentionally left as-is — it's inherent to the `Sequence` interface (those methods can't throw the checked `XPathException`), and the error still propagates through a throwing accessor.

## Reproducer / test

A bare query does **not** trigger this: for straightforward recursion the trampoline in `FunctionCall` unwinds the deferred and the error propagates once (a probe of ten candidate queries logged it zero times). The repeat-logging needs a `DeferredFunctionCall` reaching the non-throwing accessors before a throwing one, which is the serializer/templating path. So the regression test demonstrates the defect deterministically at the unit level:

`DeferredFunctionCallErrorTest` builds a `DeferredFunctionCall` whose `execute()` throws, calls seven non-throwing accessors on it, and counts `"Exception in deferred function"` log lines via a counting log4j2 appender. It asserts the body runs **once**, the error is logged **once** (was 7), and the error still **surfaces** through a throwing accessor.

(Note for maintainers: the exist-core test `log4j2.xml` sets `<Root level="OFF">`, so these events are filtered before any appender unless the test installs a dedicated `LoggerConfig` at level `ALL` — which the test does.)

## Test plan

- [x] New regression test `DeferredFunctionCallErrorTest` — logged-once + body-runs-once + error-still-surfaces; fails (7 logs) against unfixed code, passes after.
- [x] No regression: 2,014 XQSuite tests (`XQuery3Tests` + `CoreTests`) + `XQueryTest` (99, the tail-recursion JUnit path the fix rides on) all green.
- [x] Codacy PMD clean on the changed Java (the fix and the test).

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/DeferredFunctionCall.java
+++ b/exist-core/src/main/java/org/exist/xquery/DeferredFunctionCall.java
@@ -58,7 +58,23 @@ public abstract class DeferredFunctionCall implements Sequence {
             sequence = execute();
         }
     }
-    
+
+    /**
+     * Caches and logs an exception raised while realizing the deferred call, but only the first time.
+     * The non-throwing {@link Sequence} accessor methods catch the exception and return a default,
+     * and {@link #realize()} re-throws the cached exception on every later access; without this guard
+     * the same failure would be logged once per accessor the consumer calls (e.g. the serializer or
+     * the html-templating engine, which inspect a result sequence through several accessors).
+     *
+     * @param e the exception raised by {@link #execute()} (or re-thrown by {@link #realize()})
+     */
+    private void captureAndLogOnce(final XPathException e) {
+        if (caughtException == null) {
+            caughtException = e;
+            LOG.error("Exception in deferred function: {}", e.getMessage());
+        }
+    }
+
     protected FunctionSignature getSignature() {
         return signature;
     }
@@ -102,8 +118,7 @@ public abstract class DeferredFunctionCall implements Sequence {
             realize();
             return sequence.getCardinality();
         } catch (XPathException e) {
-            caughtException = e;
-            LOG.error("Exception in deferred function: {}", e.getMessage());
+            captureAndLogOnce(e);
             return Cardinality.EMPTY_SEQUENCE;
         }
     }
@@ -113,8 +128,7 @@ public abstract class DeferredFunctionCall implements Sequence {
             realize();
             return sequence.getDocumentSet();
         } catch (XPathException e) {
-            caughtException = e;
-            LOG.error("Exception in deferred function: {}", e.getMessage());
+            captureAndLogOnce(e);
             return null;
         }
     }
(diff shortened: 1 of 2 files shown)


duncdrum (approved):
I wonder if we should also add a test in the templating library to make sure the logger insanity stays off.

dizzzz (approved):

PR #6475, opened 2026-06-14
[bugfix] External variable: supplied value overrides declared default

joewiz:

## What & why

A value supplied for an external variable by the external environment should take precedence over the variable's declared default value (XQuery 3.1 [§4.15 External Variables](https://www.w3.org/TR/xquery-31/#id-external-variables): *"If a value is supplied by the external environment for an external variable, that value is used; otherwise, if a default value is specified, the default is used."*).

eXist evaluated the default **unconditionally** for a `declare variable $x external := …` declaration, so a value bound before execution — via `XQueryContext.declareVariable` (the XMLDB `XQueryService.declareVariable`, `util:eval`'s external-variables argument, REST/API callers, etc.) — was silently ignored whenever the declaration carried a default. Only externals *without* a default honored a supplied value.

```xquery
(: with $greeting bound to "supplied" by the caller :)
declare variable $greeting external := "default";
$greeting
(: before: "default"   after: "supplied" :)
```

## How

`VariableDeclaration` was constructed identically for a plain internal global (`declare variable $x := …`) and an external-with-default (`declare variable $x external := …`), so `eval()` couldn't tell them apart and always evaluated the initializer.

- The tree parser (`XQueryTree.g`) now marks the declaration as external via a new `VariableDeclaration.setExternal(true)`.
- `VariableDeclaration.eval()`: for an external declaration with a default, it first checks whether a value was already supplied (a non-null global value bound before execution) and, if so, uses it (applying the declared sequence type) instead of evaluating the default.

Behavior is unchanged for: non-external globals (always evaluate their initializer), external variables without a default, and defaulted externals with no supplied value.

## Test plan

- New `ExternalVariableDefaultOverrideTest` (exist-core), 5 cases, all green locally:
  - supplied value overrides default; typed supplied value (`xs:integer` 21 → 42) overrides default; default used when not supplied; missing required external still raises `XPDY0002`; a plain internal global with an initializer is unaffected.
- Requesting the full CI suite for regression coverage across the XQuery prolog/variable paths.

## Motivation

Enables parameterized server-side queries (binding `$param` values at call time) over HTTP — e.g. an editor running a stored `.xq` against the database the way `basex -bname=value` does — without forcing query authors to drop default values from their external-variable declarations.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/VariableDeclaration.java
+++ b/exist-core/src/main/java/org/exist/xquery/VariableDeclaration.java
@@ -44,6 +44,10 @@ public class VariableDeclaration extends AbstractExpression implements Rewritabl
     Optional<Expression> expression;
     SequenceType sequenceType = null;
     boolean analyzeDone = false;
+    /** True for an external variable declaration ({@code declare variable $x external [:= ...]}). When
+     *  it carries a default value, eval() prefers a value supplied by the external environment over
+     *  that default, per XQuery 3.1 §4.15. */
+    private boolean external = false;
 
     public VariableDeclaration(final XQueryContext context, final QName qname, final Expression expr) {
         super(context);
@@ -53,6 +57,16 @@ public class VariableDeclaration extends AbstractExpression implements Rewritabl
 
     public QName getName() {
         return qname;
+    }
+
+    /**
+     * Marks this as an external variable declaration. Set by the parser for
+     * {@code declare variable $x external} (with or without a default value).
+     *
+     * @param external whether this declaration is external
+     */
+    public void setExternal(final boolean external) {
+        this.external = external;
     }
 
     /**
@@ -91,6 +105,18 @@ public class VariableDeclaration extends AbstractExpression implements Rewritabl
         }
         analyzeExpression(contextInfo);
         var.setIsInitialized(true);
+    }
+
+    /**
+     * The global variable for this declaration if the external environment has already supplied it a
+     * value (i.e. it was bound before execution), otherwise {@code null}. Lets an external variable's
+     * supplied value take precedence over its declared default.
+     *
+     * @return the supplied variable, or {@code null} if none was supplied
+     */
+    private @Nullable Variable suppliedExternalValue() {
+        final Variable existing = context.resolveGlobalVariable(qname);
+        return existing != null && existing.getValue() != null ? existing : null;
     }
 
     private @Nullable Module findDeclaringModule(@Nullable final Module[] modules) {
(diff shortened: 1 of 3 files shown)


exist-core/src/test/java/org/exist/xquery/ExternalVariableDefaultOverrideTest.java
```diff
@@ -0,0 +1,90 @@
...
+        assertEquals(1, result.getSize());
+        assertEquals("default", result.getResource(0).getContent());
+    }
+
+    @Test
+    public void missingRequiredExternalStillErrors() throws XMLDBException {
+        final XQueryService service = server.getRoot().getService(XQueryService.class);
+        try {
+            service.query("declare variable $required external; $required");
+            org.junit.Assert.fail("expected XPDY0002 for an unbound external variable with no default");
```
line-o: use short name ;)

line-o (approved):
Thanks. LGTM

exist-core/src/test/java/org/exist/xquery/ExternalVariableDefaultOverrideTest.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Done — switched to `import static org.junit.Assert.fail;` and the short `fail(...)`. Thanks for the review!

line-o:
@joewiz can you rebase this PR please.

dizzzz (approved):

exist-core/src/main/java/org/exist/xquery/VariableDeclaration.java
```diff
@@ -44,6 +44,10 @@ public class VariableDeclaration extends AbstractExpression implements Rewritabl
     Optional<Expression> expression;
     SequenceType sequenceType = null;
     boolean analyzeDone = false;
+    /** True for an external variable declaration ({@code declare variable $x external [:= ...]}). When
+     *  it carries a default value, eval() prefers a value supplied by the external environment over
+     *  that default, per XQuery 3.1 §4.15. */
+    private boolean external = false;
```
dizzzz: probably the name could be isExternal .....

line-o:
please rebase @joewiz

exist-core/src/main/java/org/exist/xquery/VariableDeclaration.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Good catch — done in a34899583f, and it turned out to fix more than the name.

The field is now `isExternal`. I kept the setter as `setExternal(boolean)`: `setX` is the dominant convention for boolean setters in exist-core (68 vs 12 for `setIsX`), and it's the parser-facing API called from `XQueryTree.g`.

The reason it's worth more than a naming preference: `eval()` declares a local `final Variable external` in the external-variable-without-default branch, which was shadowing the boolean field of the same name. So within that block `external` meant the `Variable`, and a few lines up it meant the flag. Renaming the field removes the collision outright.

No behavior change — `ExternalVariableDefaultOverrideTest` (5) and `XQueryTest` (99) pass.

joewiz:

@line-o Rebased onto current `develop` — it was 423 commits behind, and both commits replayed with no conflicts (the post-rebase diff is byte-identical to the pre-rebase one).

On the XQTS comparison above: **neither the 39 wins nor the 3 regressions are real.** Both sides are an artifact of the baseline run, and I don't think this PR moves XQTS at all.

The tell is in the totals: the run reports **+80 total tests** and **+65 skipped** against `develop`. This PR adds one JUnit test and cannot add 80 test cases to XQTS, so the two runs did not record the same set of tests.

I pulled the JUnit artifacts from the baseline run (32187077330, `61a9b9ebac` — which is exactly the commit I rebased onto) and the three `develop` runs before it:

| Test | 32117553964 | 32131927743 | 32175296984 | 32187077330 (baseline) |
|---|---|---|---|---|
| `K-DayTimeDurationDivide-15` | FAIL | FAIL | FAIL | **absent** |
| `default-language-004` | FAIL | FAIL | FAIL | **absent** |
| `Axes112` | FAIL | FAIL | FAIL | **absent** |

All three "newly failing" tests have been failing on `develop` all along. They are flagged as new only because the baseline run never recorded them, so the comparison saw absent → fail and called it a regression.

The same thing happened in the other direction. Sampling the "newly passing" list:

| Test | 32117553964 | 32131927743 | 32175296984 | 32187077330 (baseline) |
|---|---|---|---|---|
| `K-NumericDivide-4` | PASS | PASS | PASS | **absent** |
| `K2-NumericDivide-6` | PASS | PASS | PASS | **absent** |
| `op-divide-dayTimeDuration-14` | PASS | PASS | PASS | **absent** |
| `XQST0088_1` | PASS | PASS | PASS | **absent** |
| `K2-SeqExprInstanceOf-92` | PASS | PASS | PASS | **absent** |

Already passing on `develop`, absent from the baseline, so absent → pass got reported as a win.

Recorded test-case counts across those four `develop` runs — same 429 test-set files each time — were 59,146 / 62,618 / 62,482 / 62,386. That spread is between `develop` runs and has nothing to do with any PR.

Independently of the artifact evidence, none of the three flagged tests involves an external variable, which is all this PR touches:

- `K-DayTimeDurationDivide-15` — `xs:dayTimeDuration("P3D") div xs:duration("P3Y3M")`, expects `XPTY0004`
- `default-language-004` — `let $f := (fn:default-language#0, fn:default-language#0) return $f[1]() eq $f[2]()`
- `Axes112` — `/*/namespace-node()`, expects the static error `XQST0134`

I could not identify the mechanism behind the dropped results: the `xqts-logs` artifact contains the JUnit XML and HTML but not the runner log, so there is nothing to check for a shutdown or result-collection race. What is directly observable is that the baseline run under-recorded roughly 80 test cases relative to its neighbors.

This is a close cousin of what @line-o flagged on #6315 back in May, where a PR that touched nothing relevant showed a newly-passing map test that turned out to be flaky on `develop`. Worth considering whether the comparison should treat a test that is *absent* from one side as "not comparable" rather than folding it into the newly-passing/newly-failing counts — that alone would have made this report empty, which is the correct answer here.

line-o:
The XQTS result drift is due to gaps in reporting (see https://github.com/eXist-db/exist-xqts-runner/issues/74)

duncdrum (approved):

PR #6476, opened 2026-06-14
Improve function resolution

line-o:
-- update --

All function not found error messages throughout the codebase now all go through one method that compiles the actionable message (all occasions were this made sense).
Screenshot 2026-06-19 at 12 19 20

Fixes 

- improve not found error message for namespaces spread over more than one module
- check deprecated functions in external modules as well
- drop code path that attempts to resolve functions in the XSL namespace in XPath namespace instead

Add a new helper function to FunctionFactory to keep logic from to two call sites in one place:
- ArrowOperator
- XqueryTree.g

Reformat and refactor for readability and maintainability
- FunctionFactory
- Use Function.setASTNode(ast) instead of Function.setLocation(line, number)
- Use XPathException constructor with AST as first parameter
- improve and fix JavaDoc
- simplify and improve internal helpers
- align interfaces for internal and external modules

Smaller refactorings for readability and maintainability in 
- ArrowOperator
- XqueryContext

Tests 

One test kept failing on my local machine with a fixed port (could not bind to 8088)
I switched it to using a random port instead. Please review this with care as well.

### DIFF
--- a/exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g
+++ b/exist-core/src/main/antlr/org/exist/xquery/parser/XQueryTree.g
@@ -3506,16 +3506,7 @@ throws PermissionDeniedException, EXistException, XPathException
         )*
     )
     {
-        step = FunctionFactory.createFunction(context, fn, path, params);
-        if (isPartial) {
-            if (!(step instanceof FunctionCall)) {
-                if (step instanceof CastExpression) {
-                    step = ((CastExpression)step).toFunction();
-                }
-                step = FunctionFactory.wrap(context, (Function)step);
-            }
-            step = new PartialFunctionApplication(context, (FunctionCall) step);
-        }
+        step = FunctionFactory.createFunctionCall(context, fn, path, params, isPartial);
     }
     ;
 
--- a/exist-core/src/main/java/org/exist/xquery/AbstractInternalModule.java
+++ b/exist-core/src/main/java/org/exist/xquery/AbstractInternalModule.java
@@ -49,13 +49,6 @@ public abstract class AbstractInternalModule implements InternalModule {
     protected final Map<QName, Variable> mGlobalVariables = new HashMap<>();
     private final Map<String, List<?>> parameters;
 
-    public static class FunctionComparator implements Comparator<FunctionDef> {
-        @Override
-        public int compare(final FunctionDef o1, final FunctionDef o2) {
-            return o1.getSignature().getFunctionId().compareTo(o2.getSignature().getFunctionId());
-        }
-    }
-
     public AbstractInternalModule(final FunctionDef[] functions, final Map<String, List<?>> parameters) {
         // Defensive-copy + sort so the caller's static final array is left intact
         // and getFunctionDef() can binary-search regardless of declaration order.
@@ -78,12 +71,11 @@ public abstract class AbstractInternalModule implements InternalModule {
      * eXist 7 without source changes; call sites should migrate to
      * {@link #AbstractInternalModule(FunctionDef[], Map)}.
      *
-     * @param functions         the array of functions
-     * @param parameters        configuration parameters
-     * @param functionsOrdered  ignored as of #6378
-     *
+     * @param functions        the array of functions
+     * @param parameters       configuration parameters
+     * @param functionsOrdered ignored as of #6378
      * @deprecated since 7.0.0; the {@code functionsOrdered} parameter has no effect.
-     *             Use {@link #AbstractInternalModule(FunctionDef[], Map)}.
+     * Use {@link #AbstractInternalModule(FunctionDef[], Map)}.
      */
     @Deprecated(since = "7.0.0", forRemoval = true)
     public AbstractInternalModule(final FunctionDef[] functions, final Map<String, List<?>> parameters,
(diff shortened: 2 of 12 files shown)


duncdrum (approved):

dizzzz (approved):

line-o:
I received a review that I would like to address before this is merged.

joewiz:
[This message was prompted by Joe, drafted by Claude Code, and reviewed by Joe before posting. -Joe]

I took a careful pass over the `FunctionFactory` rewrite. Net: this reads as a sound, largely behavior-preserving refactor — the `createFunctionCall` consolidation of the `ArrowOperator` and `XQueryTree.g` partial-application logic is clean, and the `functionCall` rewrite is equivalent to the original on the two points I most expected to bite:

- I checked whether moving the throw out of the per-module helpers (the old `throwOnNotFound` on the last module) into a single throw after the loop changed forward-reference / local-function resolution. It doesn't — on `develop` the last module already threw (or returned a forward reference for a not-yet-ready module), so the `getUserDefinedFunction` fallback was already unreachable when `getModules(uri) != null`. The new `modules == null → getLocalDefinedFunction` split preserves that.
- I also checked the per-module `getFunctionsByName` gather for an efficiency regression; `develop` did the same per-module scan on the miss path (`getSignaturesForFunction` / `getFunctionsByName`), so it's equivalent.

So nothing below is a blocker — they're mostly diagnostics and cleanup. Flagging for your call.

### Behavior / diagnostics

1. **Lost detail in the "function not defined" message** (`FunctionFactory.functionCall`, ~L383 and the `otherArities` block). For a function missing in an imported XQuery module, `develop` emitted `Function f#2 is not defined in namespace '...' for module: <module source path>`; the new aggregate message drops the module-source path and the `#arity`, and changes the signature-list line endings (`\r\n` → `\n`) and wording. Two small consequences: any test/tooling pinned to the old text may break, and the module-source breadcrumb (useful when several modules are bound to one namespace) is gone. Might be worth keeping the source-path in the message.

2. **Deprecation guard widened to external modules** (~L359). On `develop` the `xquery.disable-deprecated-functions` check lived inside `getInternalModuleFunction`, so it only applied to internal (Java) module functions. It now runs on the result of `getXQueryModuleFunction` too. It's inert today (XQuery user functions don't carry deprecation metadata — I didn't find any `%deprecated`→`FunctionSignature.setDeprecated` wiring), so no current behavior change, but it would start rejecting deprecated *library-module* functions if that metadata ever gets populated. Intentional scope change, or incidental to the consolidation?

3. **XSL → XPath-functions namespace fallback removed** (`getInternalModuleFunction`, ~L426). `develop` had a block that, when an XSL-namespace function wasn't found, re-searched the `fn:` internal module. It's gone now. It looks like dead/broken code on `develop` anyway (no module is registered under the XSL namespace in the stock `conf.xml`, and the old block passed the original `qname` rather than the rebuilt `_qname_` to `getFunctionDef`, so it likely never resolved) — so removing it is probably fine, just flagging that it's a resolution-behavior change rather than pure formatting, in case the intent should be captured in the commit message.

### Cleanup (optional)

4. **`getFunctionsByName` duplicated.** Lifting it to the `Module` interface is the right call, but the body is now identical in `AbstractInternalModule` (~L158) and `ExternalModuleImpl` (~L174), and in `ExternalModuleImpl` it overlaps the existing `getSignaturesForFunction` (~L163, same map scan, same name match). Could share one helper / interface default so the candidate-gathering logic the error message relies on can't drift between the two.

5. **Same message, two formatters.** `functionCall`'s "Unexpectedly received N parameter(s)… Defined function signatures are:" now uses a text block + `\n`, while `XQueryContext.functionNotFoundErrorDescription` (~L3001/3032) builds the same message with `StringBuilder` + `\r\n`. They already disagree on line endings — a shared formatter would keep the two paths consistent.

6. **Minor.** (a) For internal functions the AST node is now set on the outer `InternalFunctionCall` wrapper (`fn.setASTNode(ast)` in `functionCall`) while line/column stays on the inner `Function` via `setLocation` — benign, just a small split that could surprise anything reading `getASTNode()` off a resolved function. (b) `AbstractInternalModule` ends without a trailing newline (and `FunctionComparator` moved to the bottom), which Codacy will likely flag.

Happy to be wrong on any of these — and thanks for the readability work on this area, it's long overdue.

line-o:
> 1. Lost detail in the "function not defined" message 

That is true. At the moment this is only added if the other arities are in the _last_ implementing module for that namespace. For me having the other arities in all cases was more important.
It can be re-added if deemed important by a slightly more complex aggregation that also stores the implementing module along with the other arities.

> 2. Deprecation guard widened to external modules

That was a deliberate extension as I am convinced external and user-defined libraries should gain the ability to signal deprecated functions. This will allow effective deprecations in html-templating for example.

> 3. XSL → XPath-functions namespace fallback removed

I was hesitant at the beginning to remove this. 
I found that this codepath is both 
- untested (not a single test needed to change nor did it change XQTS results)
- according to my current understanding of how function resolution in namespaces works this was _never_ spec compliant and looks like a workaround for a bug

This is why I choose to remove it entirely. This allowed for a more efficient refactoring as well.

> 4. getFunctionsByName duplicated.

I wasn't aware that they are exact duplicates and am happy to further simplify and refactor the classes (in a follow-up perhaps)

> 5. Same message, two formatters.

This should be tackled in this PR - one formatter is enough. I prefer the Java Textblock implementation.
I stumbled over the use of `\r\n` and would like to hear if there is a justification to use this line ending instead of just new-lines. 

> Using AST nodes

The duplicate setting of the AST node which led to the sentence that the line and column _might_ be different (for internal functions) is incorrect as the line and column are from the same AST node in both cases.

PR #6474, opened 2026-06-14
[refactor] DRAFT: drop Apache HttpClient 4 entirely (HC5 EXPath HTTP Client prerelease)

joewiz:
[This PR was co-authored with Claude Code. -Joe]

> **Draft / demonstration — not for merge as-is.** It depends on an *experimental prerelease* of the EXPath HTTP Client (published to GitHub Packages, not Maven Central). The purpose is to show that eXist removes Apache HttpClient 4 **completely** once an HC5 build of that library exists, and that CI stays green. It builds on (and includes) #6473.

## Background

#6473 migrates eXist to HttpClient 5 but has to keep **one** HC4 dependency: `extensions/expath` pulls `org.expath.http.client:http-client-java`, whose published releases are still HttpClient 4, so `EXistTreeBuilder` iterates `org.apache.http.Header`.

The upstream library actually migrated to Apache HttpComponents 5 on its `main` branch (expath/expath-http-client-java#49) but **has not yet published a release** — tracked in expath/expath-http-client-java#59. To prove out the end state, this PR consumes a prerelease (`1.5.1-SNAPSHOT`) built from that `main` branch and published to GitHub Packages.

## What changed (on top of #6473)

- **`extensions/expath`** — consume `http-client-java:1.5.1-SNAPSHOT`; drop HC4 `httpcore` (now `httpcore5` only); `EXistTreeBuilder` imports `org.apache.hc.core5.http.Header`.
- **`exist-parent`** — add a temporary `github-expath-http-client` repository (GitHub Packages).
- **CI `maven-github-settings`** — add the matching server, authenticated with the CI token.
- **`BUILD.md`** — document the temporary repository.

## Result

`git grep "import org.apache.http"` returns nothing — **no Apache HttpClient 4 anywhere in the code base.** The EXPath `http:send-request` integration test passes resolving the artifact from GitHub Packages.

## Path to merge

Once expath/expath-http-client-java#59 ships an official release, bump `extensions/expath` to it and revert the temporary repository + CI server + this note. At that point this becomes a tiny, mergeable change on top of #6473.

PR #6473, opened 2026-06-13
[refactor] Migrate HTTP usage to Apache HttpClient 5; drop Milton + outdated HC4 deps

joewiz:
[This PR was co-authored with Claude Code. -Joe]

Continues and supersedes #6393 by @duncdrum (his commits are preserved here, rebased onto current `develop`). Picks up where that draft left off now that Saxon 12, Jackrabbit WebDAV (#6364), and Jetty 12 (#6392) have all landed.

## Summary

Migrates eXist-owned HTTP usage from Apache HttpClient 4.x to HttpClient 5.x across the code base, removes the dead Milton WebDAV test infrastructure that #6364 left behind, and drops the now-unused HC4 dependencies. Goal: consistent HTTP interactions throughout, and fewer outdated deps.

## What changed

**@duncdrum's work (rebased, authorship preserved):** integration tests + eXist-owned HTTP code migrated HC4 → HC5 (5.6.1); removed the 8 milton-client WebDAV tests + `com.ettrema` stubs, added JDK-HttpClient round-trip tests (`WebDavRoundTripTest`, `WebDavHttpTest`); standalone `web.xml` uses the Jackrabbit servlet.

**Finishing work (this continuation):**
- Migrated three tests `develop` added *after* the original branch and still on HC4: `URLRewriteViewPipelineTest`, the vector tests in `JmxRemoteTest`, and `RestBinariesTest` (restored an HC5 `postXquery` returning `ClassicHttpResponse` for the `response:stream-binary-resource` tests; `postXqueryBody` delegates to it).
- Dropped HC4 `httpclient` / `httpmime` / `fluent-hc` from the parent BOM + the unused `apache.httpcomponents.version` property.
- Removed the stale Milton logger from `exist-distribution` `log4j2.xml`; `debuggee` inherits the BOM-managed HC5 version; bumped `http-client-java` 1.4.2 → 1.5.2 (supersedes dependabot #5346).

## One HC4 dependency remains, and why

`extensions/expath` keeps **HC4 `httpcore`** — solely for `EXistTreeBuilder`, which iterates `org.apache.http.Header` exposed by the EXPath `http-client-java` library. That library is still HC4 at its latest release: `http-client-java 1.5.2`'s `HeaderSet implements Iterable<org.apache.http.Header>`, identical to 1.4.2. So full HC4 removal isn't possible until `http-client-java` itself migrates to HC5; `httpcore` (HC4) is scoped to expath and documented in its pom.

## Test plan

[...]

PR #6460, opened 2026-06-09
Cleanup distribution and installer jobs

duncdrum:
Cleanup remove dead stuff, and avoid unnecessary artefact generation.
Fix Installer.jar generation and publishing.

PR #6482, opened 2026-06-15
[feature] Native EXPath HTTP Client (java.net.http + Methanol); remove Apache HttpClient

### LINKED ISSUE #4256: EXPath HttpClient errors are not spec compliant
Tested with eXist-db 6.0.1 and 4.10.0.

The following query should raise the XPath Error `expath-err:HC001` (see http://expath.org/spec/http-client#errHC001), however it raises `java:org.expath.httpclient.HttpClientException`

```
xquery version "3.1";

import module namespace http = "http://expath.org/ns/http-client";

declare namespace expath-err = "http://expath.org/ns/error";
declare namespace java = "http://exist.sourceforge.net/NS/exist/java-binding";

try {
    http:send-request(
        <http:request href="http://no-such-host" method="get" follow-redirect="true"/>
    )
} catch expath-err:HC001 {

    <http-protocol-error>{$err:description}</http-protocol-error>    

} catch java:org.expath.httpclient.HttpClientException {
    
    <caught-java>{$err:code}</caught-java>

} catch * {
    <unknown-error>{$err:code}</unknown-error>
}
```

### LINKED ISSUE #5771: [feature] http-client rewrite using modern java API
TBD

joewiz:

## Summary

Replaces eXist's EXPath HTTP Client (`http:send-request`) with a native implementation built on the JDK `java.net.http.HttpClient`, augmented by [Methanol](https://github.com/mizosoft/methanol) (thanks to @dizzzz for the suggestion to investigate Methanol), and removes Apache HttpClient from the code base entirely — production **and** tests. Along the way it brings the client closer to the EXPath spec: full authentication (preemptive **and** challenge-response, Basic and Digest) and spec-compliant error codes.

After this PR, `git grep org.apache.httpcomponents` over the poms and `import org.apache.http` / `import org.apache.hc` over the sources both return nothing.

This is the rewrite requested in #5771, and it also fixes #4256 (non-spec-compliant errors). It supersedes #6473 (which migrated to Apache HttpClient 5 as an intermediate step) and #6474 (the draft HC5 EXPath prerelease) — both can be closed once this lands. With the native module now living in core, the [eXist-db/exist-http-client](https://github.com/eXist-db/exist-http-client) prototype repo can also be archived after merge.

## Why

eXist's EXPath HTTP client depended on Apache HttpClient (HC4) and the third-party `http-client-java` EXPath library, and the integration test suite drove the server through HC4's fluent client (plus, for WebDAV, the `milton-client` library, which itself pulls in Apache HttpClient). Methanol is a thin, Apache-2.0 augmentation of the JDK client: it adds transparent gzip/deflate decoding, a read (inactivity) timeout, `MediaType`, and `MultipartBodyPublisher` — the pieces the bare JDK client lacks — while pulling in **no** Apache HttpClient.

## What changed

**Native EXPath HTTP Client module** (`extensions/modules/http-client`)

- New `org.exist.xquery.modules.httpclient` module implementing `http:send-request` (namespace `http://expath.org/ns/http-client`) on `java.net.http.HttpClient`. The `http:request` element, namespace, and function signatures are unchanged, so existing queries are unaffected.
- The client is a Methanol client with `autoAcceptEncoding` (advertises `Accept-Encoding` and transparently decodes gzip/deflate) and a read timeout.
- Multipart request bodies are built with Methanol's `MultipartBodyPublisher` (the JDK client has no multipart support); each `http:body` part keeps its own `Content-Type`.

**Authentication** (per EXPath spec §3.3)

- `@send-authorization='true'` sends Basic credentials preemptively.
- Otherwise (the spec default), credentials are withheld until the server returns a 401 challenge, and the request is re-sent once with the computed `Authorization` header — **Basic** or **RFC 2617 Digest** (MD5, `qop=auth`). This challenge-response mode was missing from the initial port; it matches the previous eXist client and the BaseX reference implementation.

**Error handling** (fixes #4256)

- Connection / I/O failures surface as `expath-err:HC001`, timeouts as `expath-err:HC006`, in the EXPath error namespace (`http://expath.org/ns/error`) — catchable from XQuery, rather than the raw `org.expath.httpclient.HttpClientException` the old client raised.

**Old EXPath client removed**

- Deletes the old `org.expath.exist` HTTP client (`SendRequestFunction`, `HttpClientModule`, the `EXistResult`/`EXistTreeBuilder` adapters, and the orphaned `org.expath.tools` model classes). The `exist-expath` module keeps only its Zip functions. Drops `http-client-java`, `tools-java`, `apache-mime4j-core` and the HC4 `httpcore` dependency.
- `debuggee`'s `HttpSession` (the `XDEBUG_SESSION` form POST) is migrated to the JDK client.
- Every `conf.xml` re-points `http://expath.org/ns/http-client` to the native module.

**Tests migrated off Apache HttpClient**

- `AbstractHttpTest` now builds on `java.net.http.HttpClient`; the integration tests across exist-core, restxq, file and persistentlogin use it. `GetParameterTest` (multipart) uses Methanol's `MultipartBodyPublisher`.
- The WebDAV tests were the last test code pulling in Apache HttpClient — they drove the server through `milton-client`. They are replaced with JDK `java.net.http` round-trip tests (`WebDavRoundTripTest`), and the now-unused Milton stack is removed (the WebDAV **server** already runs on Apache Jackrabbit). The standalone webapp `web.xml` is corrected to map the WebDAV path to the Jackrabbit `ExistWebdavServlet` (develop still referenced the long-deleted `MiltonWebDAVServlet`).

**Build**

- `exist-distribution` bundles the new `exist-http-client` module (runtime dependency) so the repointed `conf.xml` resolves the module class at runtime.
- `exist-parent` BOM: drop the Apache HttpClient (HC4) and Milton managed dependencies and their version properties; add Methanol.

## Spec references

- [EXPath HTTP Client Module 1.0](http://expath.org/spec/http-client) — `http:send-request`, the `http:request` element, the response format, authentication (§3.3), and error codes (`HC001`/`HC006`).

## Test plan

- New module: **72** `http:send-request` integration tests (in-process `com.sun.net.httpserver` target + embedded eXist) and 43 `ContentTypeHelper` unit tests — all green. Coverage includes HTTP methods, request/response bodies, multipart, gzip/deflate decoding, Content-Type classification (XML/HTML/text/JSON/binary), redirects, timeouts (`HC006`), status codes, multiple response headers, and authentication: Basic preemptive, Basic and Digest challenge-response (the test server validates the Digest response hash end-to-end), and connection errors catchable as `expath-err:HC001` (#4256).
- Migrated integration tests run green across exist-core, restxq, file and persistentlogin.
- WebDAV: `WebDavRoundTripTest` green — exact round-trip of XML DOCTYPE, XML declaration, CDATA, namespaces, non-ASCII content, and a binary document, over the JDK client against the Jackrabbit servlet.
- Full reactor build (`clean install`) green; full unit suite green; container CI (incl. litmus WebDAV compliance) exercised.

## Notes for reviewers

- The only remaining Codacy/PMD warning is `NPathComplexity` on the test class `SendRequestFunctionTest.startHttpServer()` (the in-process test server's endpoint registration). It is below where it started before this PR's auth additions; per review it can be ignored for now.
- The Milton/WebDAV removal is a consequence of removing all Apache HttpClient: `milton-client` is an Apache HttpClient consumer, so retiring those tests is required. Their WebDAV **protocol** coverage is provided by the litmus compliance suite in container CI (98/98); `WebDavRoundTripTest` covers the complementary eXist content-fidelity angle.
- Dropping `http-client-java` removes eXist's dependency on the upstream EXPath HTTP client, so merging this removes the eXist-related motivation from expath/expath-http-client-java#59 (that issue can remain open on its own merits).

Closes #5771
Closes #4256


joewiz:

A note on the WebDAV test change in this PR, since it may raise eyebrows in review:

The deleted `CopyTest` / `DeleteTest` / `LockTest` / `RenameTest` / `ReplaceTest` / `SerializationTest` / `StoreAndRetrieveTest` were built on the milton-client library, which pulls in Apache HttpClient — the very dependency this PR removes. So they could not simply be "ported to the new client": there is no drop-in JDK equivalent for milton-client's object model.

Their WebDAV *protocol* coverage (COPY, MOVE, LOCK/UNLOCK, PROPFIND, DELETE) is, however, already provided — and far more thoroughly — by the litmus compliance suite that runs in container CI (`exist-docker/src/test/bats/04-webdav-litmus.bats`), currently at 98/98: basic 16, copymove 13, props 33, locks 36. The milton JUnit tests were redundant with it.

`WebDavRoundTripTest` is intentionally narrower and complementary: it covers the eXist-specific concern litmus does not — that content stored over WebDAV round-trips faithfully through eXist's own storage and serialization. It asserts exact round-trip of an XML DOCTYPE, an XML declaration, a CDATA section, namespace declarations, non-ASCII content, and a binary document, via a small JDK `java.net.http` helper (`WebDavHttpClient`).

exist-core/src/test/java/org/exist/http/urlrewrite/ControllerTest.java
```diff
@@ -62,7 +59,7 @@ public void findsLegacyController() throws IOException {
 
         // make a request and see if the legacy controller responds
         final Tuple2<Integer, String> responseCodeAndBody = get(testCollectionName, TEST_DOCUMENT_NAME);
-        assertEquals(HttpStatus.SC_OK, (int)responseCodeAndBody._1);
+        assertEquals(200, (int)responseCodeAndBody._1);
```
line-o: I would like to have a replacement for the HttpStatus code constants here and throughout our codebase

duncdrum:
@joewiz the repeated failures in the docker job are a concern.  Exist has errors in the logs upon a clean boot. 

```
not ok 5 logs are error free
# (in test file exist-docker/src/test/bats/01-connect-spec.bats, line 26)
#   `[ "$result" -eq 0 ]' failed
```

line-o (commented):
I bet we have these constants in some package we already depend on.

exist-core/src/test/java/org/exist/http/urlrewrite/ControllerTest.java
line-o: Thanks!

joewiz: [This response was co-authored with Claude Code. -Joe]

Done — replaced the magic HTTP status integers with `java.net.HttpURLConnection.HTTP_*` constants in the tests this PR touches: `ControllerTest` (`HTTP_OK`, `HTTP_CREATED`), `JmxRemoteTest` (`HTTP_OK`), and the new `WebDavRoundTripTest` (`HTTP_OK`, `HTTP_CREATED`). Most of the migrated suite already used these; removing Apache's `HttpStatus.SC_*` had left a few call sites on raw ints. (5cb1674754)

For the broader "throughout our codebase" cleanup — there are status-code literals well beyond the tests this PR migrates — I've kept that out of scope here so the PR stays focused on the Apache HttpClient removal, and I'm glad to take it on as a dedicated follow-up.

joewiz:

Thanks for catching it — root-caused and fixed. The clean-boot error was:

```
Configuration problem: class not found for module 'http://expath.org/ns/http-client'
(ClassNotFoundException); class:'org.exist.xquery.modules.httpclient.HttpClientModule'
```

This was a real packaging bug, not a test-only issue: the old `HttpClientModule` shipped inside `exist-expath`; this PR moves the EXPath HTTP client into a new `exist-http-client` module and repoints `conf.xml` to it, but the distribution wasn't yet bundling that module — so the class wasn't on the runtime classpath and boot logged the error (in a real install, `http:send-request` would have failed to load too).

Fixed by adding `exist-http-client` as a runtime dependency of `exist-distribution` (00b194cc25), mirroring how `exist-file`/`exist-expath` are bundled; confirmed with `dependency:tree` that it now resolves as a runtime artifact. The container job should be back to error-free on this run.

line-o (approved):
LGTM!
This will make our lives much easier going forward. 
It will be interesting to learn if we get failures in apps making heavy use of http-client.
The remaining codacy warning is in a test class which I would vote to ignore for the moment.

line-o:
Does this PR fix #4256

joewiz:

Yes — it fixes #4256. The old client raised a raw `org.expath.httpclient.HttpClientException`; the native module maps connection/I/O failures to `expath-err:HC001` (and timeouts to `expath-err:HC006`) in the EXPath error namespace. I added a regression test mirroring the issue's reproducer — a `send-request` to an unreachable host caught with `catch expath-err:HC001` — and added `Closes #4256` to the description.

On apps that make heavy use of `http-client`: the surface is deliberately unchanged — same `http:request` element, same `http://expath.org/ns/http-client` namespace, same function signatures — so existing queries don't need edits. The 72 module integration tests exercise the behaviors such apps lean on: methods, request/response bodies, multipart, gzip/deflate decoding, Content-Type classification, redirects, timeouts, status codes, multiple response headers, and authentication (Basic preemptive, plus Basic and Digest challenge-response, validated end-to-end). Where behavior changed, it changed toward the spec: challenge-response auth and the `HC001`/`HC006` error codes both restore conformance the old client lacked. The honest caveat is that the JDK/Methanol client can differ from Apache in edge cases (specific header quirks, proxy/TLS behavior); if any such reports surface we can address them, but the EXPath contract itself is preserved.

Agreed on the remaining Codacy warning — it's the in-process test server's setup method; happy to leave it for now (it's actually lower than before this PR's additions).

Thanks for the review!

extensions/expath/pom.xml
```diff
@@ -57,72 +57,20 @@
...
-            <artifactId>tools-java</artifactId>
-            <version>0.7.0</version>
-        </dependency>
-
         <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>jsr305</artifactId>
         </dependency>
 
-        <dependency>
```
dizzzz, line 97: did not expect unit to disappear here. No tests remaining?

extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
```diff
@@ -0,0 +1,1413 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.modules.httpclient;
+
+import com.sun.net.httpserver.HttpServer;
```
dizzzz, line 24: I'd stay out of the com.sun packages. These are not guaranteed to be present.

a better route and sustainable route is to use wire mock.

dizzzz:
I see that on some locations "preemptive authorization" is applied.

For running the tests this makes sense as this will potentially save a handshake, But I am not sure if this is switched on for every outbound HTTPcall, as this might leak sensitive data.

dizzzz:
About performance/heavy load as asked by @line-o ; I need to deep dive here. If we are able to use the connection pool and have good (default) Executor management, I expect it will work out well. But the proof is in the pudding.

dizzzz (commented):
LGTM! Will spin up IntelliJ for a more in-depth review.

joewiz:

On preemptive authorization: it's opt-in per request, not applied to every outbound call. Credentials are sent preemptively only when the request sets `send-authorization="true"` (the EXPath default is `false`). Without it, no `Authorization` header goes out until the server replies with a 401 challenge, and then it is sent only to that challenging origin. The places you saw it are the tests deliberately setting `send-authorization="true"`. This follows the EXPath spec's auth model, so there's no blanket credential leak.

On performance / connection pooling and Executor management: good idea, and it looks readily achievable — the per-request client construction is the single seam where a shared, pooled client (with sensible Executor defaults, proxy, etc.) would slot in. Probably cleanest as a follow-on, but we're happy to fold it into this PR if you'd prefer it here. And agreed — the proof is in the pudding; glad to benchmark once there's a pooled path to measure.

extensions/expath/pom.xml
joewiz: [This response was co-authored with Claude Code. -Joe]

Right — `HttpClientTest` was the only test in `exist-expath`, and it exercised the old Apache-HttpClient-based EXPath client that this PR removes. With it gone, there are no test classes or test runner left in the module (only the `log4j2.xml`/`conf.xml` test resources), so `junit` is genuinely unused here. The HTTP client's tests now live in the new `exist-http-client` module.

Those leftover `log4j2.xml`/`conf.xml` test resources are now orphaned too — happy to strip them for a fully clean `src/test`, or leave them in case the module grows tests again. Whichever you'd prefer.

extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
joewiz: [This response was co-authored with Claude Code. -Joe]

Fair flag — sharing what we found so you and the devs can steer it. `com.sun.net.httpserver` here is the exported public API of the JDK `jdk.httpserver` module (documented since JDK 9), not a `sun.*` internal, so it is present in every standard JDK; it let the tests stand up a tiny in-process target with no extra dependency. There's no existing precedent in eXist's own tests either way — neither `com.sun.net.httpserver` nor WireMock is used today. WireMock would be the more conventional choice but adds a test dependency and a rewrite of the in-process server. No strong preference from our side; happy to go with whatever you and the team settle on as the house convention for this kind of HTTP test target.

line-o:
As expected - this branch needs to be rebased now that #6460 was merged

joewiz:
@line-o Thanks! Rebased.

dizzzz:
also vibe checked:

image

dizzzz (approved):

duncdrum:
I d be curious to get a first read on the performance and pooling angle. Otherwise I m afraid it ll drop of the radar.

dizzzz:
The HttpClient has its own (native) connection pool; no work needed:
- https://www.baeldung.com/java-httpclient-connection-management 
- https://www.javathinking.com/blog/java-httpclient-connection-management/#connection-pooling-fundamentals

further more the 'old' expath http client is implemented layered, where the exist-db objects are "converted" to expath objects, which are converted the apache objects. We have one Laye less, with probably less GC overhead.


------

follow up: the connection pooling (and more) is done per HttpClient. So for efficiency reasons we should have one HttpCLient instance for the whole eXist instance. According to ChatGPT that is a good thing.

"in most application servers a single shared HttpClient for the whole JVM is the recommended approach.
HttpClient is immutable and thread-safe, so there is no concurrency issue with thousands of threads sharing the same instance."

"This maximizes:
- TCP connection reuse
- TLS session reuse
- HTTP/2 multiplexing
- Reduced socket creation/destruction
- Lower GC pressure

When sharing is completely safe
If all applications/users use the same:
- SSL configuration
- proxy configuration
- authentication mechanism
- cookie policy (or no cookies)
then a JVM-wide client is usually ideal."

exceptions are when
- a specific cookie handler is needed
- a custom Authenticator is used
- different sslContext() is needed
- specific Proxy settings are needed.

in these cases a per-request HttpCLient needs to be created.


==> My suggestion is to do this in followup PR @duncdrum 

----

Rule of thumb
Use one shared HttpClient instance per distinct security/network configuration.

Don't create:
- one client per request
- one client per session
- one client per user

Create additional clients only when you need different:
- SSL contexts
- client certificates
- proxies
- authenticators
- cookie stores

That's typically how high-throughput Java application servers use the Java 21 HTTP client.

duncdrum (approved):

duncdrum:
@joewiz needs a rebase

PR #6491, opened 2026-06-18
[bugfix] map keys: compare by op:same-key, never coerce across type families

joewiz:

## Summary

eXist conflated map keys of **different `op:same-key` families** that share a lexical value, on constructor-built maps. Per XQuery 3.1 §17.1, map keys compare via `op:same-key` — a *type-group* comparison with **no key casting**: the numeric family interchanges; the string family (`xs:string`/`xs:anyURI`/`xs:untypedAtomic`) interchanges; every other type matches only itself.

Measured on a clean build of current `develop` (these are the new `mapKeySameKey.xqm` cases) against the two reference implementations:

| Expression | eXist before | eXist after | Saxon-HE | BaseX | Spec |
|---|---|---|---|---|---|
| `map:contains(map{"12":"x"}, 12)` | **true** | false | false | false | false |
| `map:get(map{"12":"x"}, 12)` | **["x"]** | () | () | () | () |
| `map:contains(map{"2020-01-01":1}, xs:date("2020-01-01"))` | **true** | false | false | false | false |
| `map:contains(map{true():"x"}, "true")` | **true** | false | false | false | false |
| `map:contains(map{1:1}, 1.5e0)` (numeric, lossy cast) | **true** | false | false | false | false |
| `map:contains(map{5:1}, 5.0e0)` (numeric family) | true | true | true | true | true |
| `map:contains(map{"urn:x":1}, xs:anyURI("urn:x"))` (string family) | true | true | true | true | true |

eXist was the lone outlier on the cross-family cases.

## Root cause

A `map{…}` constructor builds a homogeneously-typed `MapType` that records a concrete `keyType`. `MapType.get()`/`contains()` ran a private `convert()` that **cast the lookup key to `keyType`** (e.g. integer `12` → string `"12"`) before delegating to the underlying map — so a cross-family lookup that was castable matched a key it should not have. The coercion dates to the original 2012 map implementation.

`convert()` is now redundant. The underlying map's comparator already implements `op:same-key` (`AbstractMapType.sameKey`), and `NumericValue`/string-family `hashCode` are already canonical (recent commits `c608d55a`, `6b63ec83`), so cross-type-*within*-family pairs (`xs:integer(5)`/`xs:double(5.0)`, `xs:string`/`xs:anyURI`) already hash into the same bucket and match correctly via `sameKey` alone. `SingleKeyMapType` (built by `map:entry`) already compares with `sameKey` directly and was correct; removing `convert()` makes `MapType` consistent with it.

## Why XQTS misses it

qt3 and qt4 place their cross-family distinctness tests (`map/contains.xml`, `map/get.xml`: `xs:untypedAtomic("12")` vs `12`) on **`map:entry`**, which yields a MIXED-key map where `convert()` was a no-op → eXist already passed them. Neither suite has a *constructor*-built cross-family distinctness test. (Worth a small upstream contribution to qt4tests to close the blind spot.)

## What changed

- **`MapType`** — removed `convert()`; `get()`/`contains()` delegate straight to the `op:same-key` comparator. `keyType` is retained only for the `getKeyType()` API (documented).
- **`mapKeySameKey.xqm`** (new) — XQSuite regression tests using the **constructor** form: cross-family distinctness (string/date/boolean keys vs same-lexical numeric/other lookups) and the numeric lossy-cast case, plus within-family positives (numeric; string/anyURI/untypedAtomic) and same-type self-matches that must not regress.

### Coupling: closes the serialization bare-string "backdoor"

eXist serialization parameters are looked up by their `exist:` **QName**. A user could previously also pass them as the prefixed **string** `"exist:expand-xincludes"` only because `convert()` coerced the QName lookup to a string. With this fix, QName and string are distinct keys, so **only the spec-aligned `exist:` QName form is honored** — a deliberate, spec-aligning behavior change.

**Upgrade note:** eXist-specific serialization parameters in an options map must be keyed by their `exist:` QName (`map { xs:QName("exist:indent-spaces"): 2 }`); the previously-tolerated bare `"exist:…"` string key is no longer accepted. (`declare option exist:serialize` declarations are not affected.)

A full-repo scan found **no production/app usage** of the bare-string form — only tests. Affected tests updated: `serialize.xql`, `fnSerializeNewline.xqm`, and the file module `sync-serialize.xqm` — prefixed-string cases now assert the key is ignored; the real functionality is retained via the QName form.

## Test plan

All on a clean build of this branch off current `develop`:

- [x] `MapTests` — 155 run, 0 failures (9 new cases were failing before the fix)
- [x] `XQuery3Tests` — 1071 run, 0 failures (covers `serialize.xql`, `fnSerializeNewline.xqm`)
- [x] file module `FileTests` — 53 run, 0 failures (covers `sync-serialize.xqm`)
- [x] `ArrayTests` (164), `MapTest`, `BifurcanMapTest` — 0 failures
- [x] Codacy PMD on `MapType.java` — clean

XQTS has no relevant tests.

## Spec references

- XQuery 3.1 §17.1 — maps; `op:same-key` definition: https://www.w3.org/TR/xpath-functions-31/#func-same-key

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
@@ -66,6 +66,14 @@ public class MapType extends AbstractMapType {
      * then this is set to {@link #MIXED_KEY_TYPES}.
      * <p>
      * Uses integer values from {@link org.exist.xquery.value.Type}.
+     * <p>
+     * This is reported via {@link #getKeyType()} but does NOT take part in key
+     * comparison. Lookups ({@link #get(AtomicValue)} / {@link #contains(AtomicValue)})
+     * delegate directly to the underlying map, whose comparator implements
+     * {@code op:same-key} (see {@link AbstractMapType#sameKey}). Keys are compared
+     * by their op:same-key family, never coerced to {@code keyType} - a key from a
+     * different family that shares a lexical value (e.g. the string {@code "12"}
+     * and the integer {@code 12}) is correctly treated as distinct.
      */
     private int keyType = UNKNOWN_KEY_TYPE;
 
@@ -251,12 +259,7 @@ public class MapType extends AbstractMapType {
     }
 
     @Override
-    public Sequence get(AtomicValue key) {
-        key = convert(key);
-        if (key == null) {
-            return Sequence.EMPTY_SEQUENCE;
-        }
-
+    public Sequence get(final AtomicValue key) {
         return map.get(key, Sequence.EMPTY_SEQUENCE);
     }
 
@@ -267,12 +270,7 @@ public class MapType extends AbstractMapType {
     }
 
     @Override
-    public boolean contains(AtomicValue key) {
-        key = convert(key);
-        if (key == null) {
-            return false;
-        }
-
+    public boolean contains(final AtomicValue key) {
         return map.contains(key);
     }
 
@@ -373,17 +371,6 @@ public class MapType extends AbstractMapType {
                 break; // done, we only have to detect this once!
             }
         }
-    }
-
-    private AtomicValue convert(final AtomicValue key) {
-        if (keyType != UNKNOWN_KEY_TYPE && keyType != MIXED_KEY_TYPES) {
-            try {
-                return key.convertTo(keyType);
-            } catch (final XPathException e) {
-                return null;
-            }
-        }
-        return key;
     }
 
     @Override
--- a/exist-core/src/test/xquery/maps/mapKeySameKey.xqm
+++ b/exist-core/src/test/xquery/maps/mapKeySameKey.xqm
@@ -1,0 +1,179 @@
+(:
+ : eXist-db Open Source Native XML Database
+ : Copyright (C) 2001 The eXist-db Authors
+ :
+ : info@exist-db.org
+ : http://www.exist-db.org
+ :
+ : This library is free software; you can redistribute it and/or
+ : modify it under the terms of the GNU Lesser General Public
+ : License as published by the Free Software Foundation; either
+ : version 2.1 of the License, or (at your option) any later version.
+ :
+ : This library is distributed in the hope that it will be useful,
+ : but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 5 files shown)


line-o (changes requested):
The blast radius of this change is relatively large as all exist-db specific serialization options now _have to_ be proper QNames.

I am all in favor of this change and do recall other developers who were surprised that the prefixed string was working. The changed behaviour is the correct way, obviously.

What I want to see changed is the removed XQSuite tests to be reinstated with the new, corrected form `xs:QName(...the old string key ...) : $value` instead of removing them.

If need be we _can_ **add** one or two that ensure that prefixed string literals are now ignored.

joewiz:

Thanks — and agreed the QName form is the correct one.

I think the positive coverage is actually already preserved in exactly the form you're asking for. For each eXist serialization option, `serialize.xql` on develop already had *two* tests: a `-QName` one keyed by `xs:QName("exist:…")` and a `-string` one keyed by the prefixed string `"exist:…"`. This PR leaves every `-QName` test untouched — same `%test:args`, same assertions — so `output-doctype`, `expand-xincludes`, `add-exist-id`, `jsonp` and `process-xsl-pi` are still exercised through the correct `xs:QName(…) : $value` key with all their original value cases. What changed is only the `-string` tests: they used to assert the prefixed-string key *worked*, and since it no longer should, I converted each into a `-prefixed-string-ignored` test that asserts the string key now has no effect (the default applies). So the diff's "removed" lines are those old string-key positive assertions, but the corrected-form coverage they were duplicating lives on in the `-QName` tests.

So the `-prefixed-string-ignored` tests are the "ensure prefixed string literals are now ignored" checks you mentioned — there's just one per option rather than one or two total. Happy to trim them to a couple of representative ones if you'd prefer a lighter footprint, or to add an explicit `xs:QName(…)` case anywhere you feel the positive side is thin. Let me know which shape you'd like and I'll adjust.

line-o (approved):

line-o:
Sorry, I should have looked at the entire testfile first.

line-o (approved):
LGTM; Sorry for the delay!

dizzzz (approved):

Issue #3704, opened 2021-01-12
[BUG] file:sync doesn't respect conf.xml serializer; hard codes serialization parameters


joewiz:
**Describe the bug**

The `file:sync` function, which writes collection contents out to disk, hard codes serialization parameters such as `indent=yes`. This inserts whitespace into data and can pollute application-sensitive whitespace practices and lead to undesired source control changes.

**Expected behavior**

All functions that serialize data should, at bare minimum, respect the `<serializer>` settings in conf.xml. 

See https://github.com/eXist-db/exist/blob/develop/extensions/modules/file/src/main/java/org/exist/xquery/modules/file/Sync.java#L91-L96.

**To Reproduce**

The final expression below returns `false()` even when conf.xml's `serializer/@index` is set to `"no"`. If it respected this setting, the expression would return `true()`.

```xquery
xquery version "3.1";

xmldb:create-collection("/db", "test"),
xmldb:store("/db/test", "test.xml", <foo><bar/></foo>),
file:mkdir("/Users/joe/test"),
file:sync("/db/test", "/Users/joe/test", ()),
deep-equal(<foo><bar/></foo>, doc("file:/Users/joe/test/test.xml"))
```

**Context (please always complete the following information):**
 - OS: macOS 11.1
 - eXist-db version: eXist 5.3.0-SNAPSHOT 35a76f25f0431220de95ff43752232bbf6a59efc 20210104165718
 - Java Version: 1.8.0_275

**Additional context**
- How is eXist-db installed? built from source
- Any custom changes in e.g. `conf.xml`? none

joewiz:
I've updated the report to add an xqsuite version of the test.

joewiz:
Here's how eXist's implementation of `fn:serialize` gets serialization parameters from conf.xml: https://github.com/eXist-db/exist/blob/adebc77a884fea5e0120c25290b9767ef01dce08/exist-core/src/main/java/org/exist/xquery/functions/fn/FunSerialize.java#L94-L106.

[...]

PR #6492, opened 2026-06-18
[bugfix] file:sync: honor exist:expand-xincludes supplied via the options map

joewiz:

## Summary

`file:sync($collection, $target, $options)` accepts serialization options in its 3rd-argument `map(*)` (since 6.1.0), but the eXist extension parameter **`expand-xincludes` could not be set to a non-default value through the map**. `expand-xincludes=true` was silently ignored — the output serialized as if `expand-xincludes=no` — and `=false` only appeared to work because it matched the forced default.

## Root cause

The two halves of the option handling disagreed on the key form for the extension parameters:

1. `SerializerUtils.getSerializationOptions(map)` reads W3C parameters by their bare **string** key, but reads the eXist extension parameters (including `expand-xincludes`) by their exist-namespace **QName** key — `"exist:expand-xincludes"`. So a caller supplies `expand-xincludes` under that key, and `getSerializationOptions` records it correctly.
2. `Sync.getOptions` then ran a "fill defaults" loop that re-checked the **input** map by the bare string key only:

   ```java
   for (String p : DEFAULT_PROPERTIES.stringPropertyNames()) {   // p = "expand-xincludes"
       if (optionsMap.get(new StringValue(this, p)).isEmpty()) {  // bare string key only
           outputProperties.setProperty(p, DEFAULT_PROPERTIES.getProperty(p));  // overwrites with "no"
       }
   }
   ```

   For an exist-namespaced parameter the bare-string lookup always missed, so the loop overwrote the value `getSerializationOptions` had just parsed with the hard-coded default `expand-xincludes=no`. (Only `expand-xincludes` is both in `DEFAULT_PROPERTIES` and exist-namespaced, so it was the sole victim; `exist:insert-final-newline` etc. are unaffected because they aren't in `DEFAULT_PROPERTIES`.)

## What changed

- **`Sync.java`** — the "supplied?" check is extracted as `suppliedInOptionsMap(...)`, which tests **both** key forms: the bare string key for W3C parameters and the exist-namespace QName key for the extension parameters. A supplied extension parameter is no longer clobbered by the default.
- **`sync-serialize.xqm`** — adds `expand-xincludes` coverage (`true` expands, `false` preserves), keyed by the **spec-conformant exist-namespace `QName`** as the primary form; the existing `insert-final-newline` tests are converted to the QName key too, so the function-level suite is QName-primary.
- **`fnSerializeExtensionParamKeyForms.xqm`** (new, exist-core) — the single canonical place documenting the key-form contract for exist serialization parameters: the exist-namespace **`xs:QName`** key is honored (the only spec-conformant form), and the prefixed `"exist:…"` **string** key and the bare local name are both **ignored**. (eXist previously tolerated the prefixed-string form; that non-conformant behavior was removed by the MapType same-key fix #6491 this is stacked on.)

## Test plan

- [x] `expand-xincludes=true` (via the exist-namespace `QName` key) expands the `<xi:include>` in the synced file; `=false` preserves it.
- [x] Confirmed the new `expand-xincludes` "yes" assertion **fails** when the fix is reverted (genuine regression guard) and **passes** with it.
- [x] `fnSerializeExtensionParamKeyForms` verifies (post-#6491): QName key honored; prefixed `"exist:…"` string key rejected; bare local name rejected.
- [x] Full file-module XQSuite green (`FileTests` 55) and exist-core xquery3 suite green (`XQuery3Tests` 1080).
- [x] Codacy PMD clean on the changed Java file (the one pre-existing `UnusedLocalVariable` finding is unrelated and present on develop).

## Notes

- The spec-conformant QName key is the only accepted form for exist serialization parameters; `fnSerializeExtensionParamKeyForms.xqm` pins that the prefixed-string and bare-local forms are rejected (the previously-tolerated string form was removed by #6491, which this PR is stacked on). (`declare option exist:serialize` is a separate mechanism — bare keys there are correct and unaffected.)
- Relates to #3704. This fixes the options-map key-matching defect specifically; the original "`file:sync` doesn't read `conf.xml` serializer defaults" half of #3704 is separate and remains open (see the status comment on that issue). Left unclosed so that half isn't lost — maintainers can decide #3704's disposition.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/map/MapType.java
@@ -66,6 +66,14 @@ public class MapType extends AbstractMapType {
      * then this is set to {@link #MIXED_KEY_TYPES}.
      * <p>
      * Uses integer values from {@link org.exist.xquery.value.Type}.
+     * <p>
+     * This is reported via {@link #getKeyType()} but does NOT take part in key
+     * comparison. Lookups ({@link #get(AtomicValue)} / {@link #contains(AtomicValue)})
+     * delegate directly to the underlying map, whose comparator implements
+     * {@code op:same-key} (see {@link AbstractMapType#sameKey}). Keys are compared
+     * by their op:same-key family, never coerced to {@code keyType} - a key from a
+     * different family that shares a lexical value (e.g. the string {@code "12"}
+     * and the integer {@code 12}) is correctly treated as distinct.
      */
     private int keyType = UNKNOWN_KEY_TYPE;
 
@@ -251,12 +259,7 @@ public class MapType extends AbstractMapType {
     }
 
     @Override
-    public Sequence get(AtomicValue key) {
-        key = convert(key);
-        if (key == null) {
-            return Sequence.EMPTY_SEQUENCE;
-        }
-
+    public Sequence get(final AtomicValue key) {
         return map.get(key, Sequence.EMPTY_SEQUENCE);
     }
 
@@ -267,12 +270,7 @@ public class MapType extends AbstractMapType {
     }
 
     @Override
-    public boolean contains(AtomicValue key) {
-        key = convert(key);
-        if (key == null) {
-            return false;
-        }
-
+    public boolean contains(final AtomicValue key) {
         return map.contains(key);
     }
 
@@ -373,17 +371,6 @@ public class MapType extends AbstractMapType {
                 break; // done, we only have to detect this once!
             }
         }
-    }
-
-    private AtomicValue convert(final AtomicValue key) {
-        if (keyType != UNKNOWN_KEY_TYPE && keyType != MIXED_KEY_TYPES) {
-            try {
-                return key.convertTo(keyType);
-            } catch (final XPathException e) {
-                return null;
-            }
-        }
-        return key;
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


duncdrum:
@joewiz could you rebase this.

joewiz:
@duncdrum Ok, rebased.

duncdrum (commented):
Minor nitpick. The xqsuite assertions could be stronger / more informative in case of failures by avoiding `assertTrue`

joewiz:

Good call @duncdrum — applied in e52e92cf30, to all five assertions this PR added (three in `fnSerializeExtensionParamKeyForms.xqm`, two in `sync-serialize.xqm`).

Each test function now returns the serialized document (or the synced file's contents) and asserts with `%test:assertXPath`, one annotation per condition instead of a conjunction. Breaking a predicate on purpose used to report nothing but `expected:<true> but was:<false>`; it now names the predicate that failed and carries the actual serialized document along with it, and splitting the conjunctions means the failure points at which half broke rather than collapsing both into one false.

Worth knowing if you hit one of these for real: XQSuite XML-escapes the result into the JUnit message, so a string-valued result arrives as `&lt;doc ...&gt;` rather than as markup. Still perfectly diagnosable, just not pretty.

One thing I left alone: the `%test:assertEquals("true", "true")` + `helper:assert-file-contents` pattern used through the rest of `sync-serialize.xqm` has the same weakness, but it predates this PR and converting it would swamp the diff. Happy to open a follow-up issue for that file if you think it's worth doing separately.

duncdrum (approved):

Issue #178, opened 2014-03-26
compression:zip is not supporting Xquery 3.0 serialize option


pcolot01:
In exist 2.1 build 20130717 svn revision 18721

compression:zip is supporting 
declare option exist:serialize "method=xml indent=no";

but not 

declare namespace output = "http://www.w3.org/2010/xslt-xquery-serialization";
declare option output:method "xml";
declare option output:indent "yes";

The supported options should be documented in the function documentation.
Both modes of serialization pragma should be supported.

Best regards

Pierre Colot

joewiz:
[This response was co-authored with Claude Code. -Joe]

Still reproduces on eXist 7 (7.0.0 beta). `compression:zip` honors `declare option exist:serialize "…"` but ignores the W3C `declare option output:…` form, just as reported. Confirmed by zipping the same document under each prolog form and comparing the archive (MD5 of the resulting zip):

| Prolog option | `indent=yes` | `indent=no` | Effect |
|---|---|---|---|
| `declare option exist:serialize` | `e3d3b2…` | `5867d2…` | honored (output differs) |
| `declare option output:indent` | `e3d3b2…` | `e3d3b2…` | **ignored** (identical) |

Under the hood, `AbstractCompressFunction` serializes its XML entries solely from `context.getOption(Option.SERIALIZE_QNAME)` — i.e. the `exist:serialize` prolog option — so the W3C `output:` prolog form never reaches it.

There's a second, related gap: there's no way to set serialization **per call** either. `compression:zip` has no serialization-options argument (its arities are only `(sources, hierarchy)` / `(+ strip-prefix)` / `(+ encoding)`), and `util:declare-option("exist:serialize", …)` at runtime does not affect the output — only the static prolog form does. So a library that needs to apply serialization dynamically is forced to build a query string with the option in its prolog and run it through `util:eval`, which is an unfortunate thing for a module API to require.

[...]

PR #6493, opened 2026-06-18
[feature] compression: add serialization-options argument to compression:zip / compression:tar

joewiz:

## Summary

`compression:zip` and `compression:tar` serialize the XML resources they write into the archive, but there was no way to control that serialization per call — the functions used the serializer defaults plus whatever `declare option exist:serialize` the prolog set. This adds an optional **5th argument** of serialization options to both functions, mirroring `fn:serialize` and `file:sync#3`.

As with `fn:serialize`, the argument accepts **either** form — a `map(*)` or a W3C `output:serialization-parameters` element:

```
compression:zip($sources as xs:anyType+,
                $use-collection-hierarchy as xs:boolean,
                $strip-prefix as xs:string,
                $encoding as xs:string,
                $serialization-options as item()?) as xs:base64Binary*
```

(identical 5-argument shape for `compression:tar`)

```xquery
(: map form :)
compression:zip($sources, true(), "", "UTF8",
    map {
        "indent": false(),                                                          (: W3C param, string key :)
        QName("http://exist.sourceforge.net/NS/exist", "expand-xincludes"): true()  (: exist param, QName key :)
    }
)

(: output:serialization-parameters element form :)
compression:zip($sources, true(), "", "UTF8",
    <output:serialization-parameters xmlns:output="http://www.w3.org/2010/xslt-xquery-serialization"
                                     xmlns:exist="http://exist.sourceforge.net/NS/exist">
        <output:indent value="no"/>                  (: W3C param, output-namespace child :)
        <exist:expand-xincludes value="yes"/>        (: exist param, exist-namespace child :)
)
```

## What changed

- **`AbstractCompressFunction`** — a new `SERIALIZATION_OPTIONS_PARAM` (`item()?`, optional). When supplied, it is parsed by **`FunSerialize.getSerializationProperties(...)`** — the same public entry point `fn:serialize` uses — so the map and element forms (and their error handling, e.g. `XPTY0004` for anything else) behave identically to `fn:serialize`. The resulting `Properties` is applied to every XML entry written to the archive, **after** (so overriding) the prolog `declare option exist:serialize`, and re-read on each `eval` (no leakage between calls). Parsing is extracted into a small `parseSerializationOptions(...)` helper so `eval`'s NPath complexity is unchanged.
- **`ZipFunction` / `TarFunction`** — add the 5-argument signature `(sources, use-collection-hierarchy, strip-prefix, encoding, serialization-options)`.
- **`CompressionModule`** — register the new arity for both `zip` and `tar`.
- **`zip-tests.xql`** — new XQSuite coverage (10 tests total): `expand-xincludes` honored via its exist-namespace `QName` key (true expands, false preserves the include); the no-map default is unchanged; a mixed map (W3C string key + exist `QName` key) coexists; and the `output:serialization-parameters` element form is honored (exist param as an exist-namespace child, where a boolean uses `value="yes"`/`"no"` per the element-form convention). Assertions inspect the extracted node by structure (re-serializing a preserved `xi:include` with a relative href would re-trigger XInclude resolution against a missing base URI).

## Key form

Consistent with the rest of eXist's serialization-option handling (and with `file:sync#3`): in the **map** form, standard W3C parameters use their **string** key and eXist extension parameters use their exist-namespace **`xs:QName`** key (the spec-conformant form for implementation-defined parameters); in the **element** form, W3C parameters are output-namespace children and eXist parameters are exist-namespace children. This is why the PR is stacked on #6491 — that fix makes the exist-namespace `QName` key the single accepted form for the extension parameters in the map.

## Test plan

- [x] `compression:zip(..., map { QName($exist-ns, "expand-xincludes"): true() })` expands the `<xi:include>` in the archived resource; `false()` preserves it.
- [x] No map argument: archived XML serializes exactly as before (the serializer's own default).
- [x] A map mixing a W3C string key (`indent`) and an exist `QName` key (`expand-xincludes`) is honored.
- [x] The `output:serialization-parameters` element form is honored (exist param as an exist-namespace child element).
- [x] `CompressionTests` XQSuite green (10/10).
- [x] Codacy PMD: no new findings on the changed Java files (the three pre-existing findings — `AvoidReassigningParameters` on `removeLeadingOffset`, `NPathComplexity` on `compressElement`, `UnusedLocalVariable` `updateLock` — are present on develop and untouched here).

## Notes

- **Scope vs. the original report.** #178 (2013) asked specifically for the prolog `declare option output:method "xml"` (W3C `output:` namespace) to be honored by `compression:zip`, in addition to the already-supported `declare option exist:serialize`. This PR addresses the underlying need — per-call control over how archived XML is serialized, covering both W3C and eXist parameters in either the `map(*)` or `output:serialization-parameters` element form — which is the modern idiom (`fn:serialize`, `file:sync`). It does **not** wire up the W3C `output:` *prolog* declaration (a different mechanism: a query-prolog `declare option`, not a value passed to the function). Filed as **Relates to #178** rather than Closes, so you can decide whether the argument satisfies the issue or whether the prolog-`output:` path should also be implemented (I'd suggest a separate, smaller follow-up if so).
- `declare option exist:serialize` continues to work and is unaffected; the map argument simply takes precedence over it.

### DIFF
--- a/extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
+++ b/extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
@@ -47,6 +47,7 @@ import org.exist.xquery.FunctionSignature;
 import org.exist.xquery.Option;
 import org.exist.xquery.XPathException;
 import org.exist.xquery.XQueryContext;
+import org.exist.xquery.functions.fn.FunSerialize;
 import org.exist.xquery.value.AnyURIValue;
 import org.exist.xquery.value.Base64BinaryValueType;
 import org.exist.xquery.value.BinaryValueFromInputStream;
@@ -71,6 +72,7 @@ import java.nio.charset.UnsupportedCharsetException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Iterator;
+import java.util.Properties;
 import java.util.zip.CRC32;
 import java.util.zip.DeflaterOutputStream;
 import java.util.zip.ZipEntry;
@@ -90,9 +92,17 @@ public abstract class AbstractCompressFunction extends BasicFunction {
     protected final static SequenceType COLLECTION_HIERARCHY_PARAM = new FunctionParameterSequenceType("use-collection-hierarchy", Type.BOOLEAN, Cardinality.EXACTLY_ONE, "Indicates whether the Collection hierarchy (if any) should be preserved in the zip file.");
     protected final static SequenceType STRIP_PREFIX_PARAM = new FunctionParameterSequenceType("strip-prefix", Type.STRING, Cardinality.EXACTLY_ONE, "This prefix is stripped from the Entrys name");
     protected final static SequenceType ENCODING_PARAM = new FunctionParameterSequenceType("encoding", Type.STRING, Cardinality.EXACTLY_ONE, "This encoding to be used for filenames inside the compressed file");
+    protected final static SequenceType SERIALIZATION_OPTIONS_PARAM = new FunctionParameterSequenceType("serialization-options", Type.ITEM, Cardinality.ZERO_OR_ONE,
+            "Serialization options applied to XML resources written to the archive, as either an " +
+                    "output:serialization-parameters element or a map(*) (e.g. map { \"indent\": false() }). " +
+                    "Standard W3C parameters use their string key; eXist extension parameters use their exist-namespace QName key. " +
+                    "Takes precedence over any prolog declare option exist:serialize.");
     private final static Logger logger = LogManager.getLogger(AbstractCompressFunction.class);
     public static final String METHOD_STORE = "store";
 
+    /** Serialization options from the optional map argument; reset on each eval and applied to XML
+     *  entries after (so overriding) the prolog {@code declare option exist:serialize}. */
+    private Properties serializationProperties = new Properties();
 
     public AbstractCompressFunction(final XQueryContext context, final FunctionSignature signature) {
         super(context, signature);
@@ -110,6 +120,20 @@ public abstract class AbstractCompressFunction extends BasicFunction {
         return uri;
     }
 
+    /**
+     * Parses the optional serialization-options map argument (the 5th argument, if present) into the
+     * serialization properties applied to XML entries written to the archive.
+     *
+     * @param args the function arguments
+     * @return the supplied serialization properties, or an empty {@link Properties} when no map was given
+     */
+    private Properties parseSerializationOptions(final Sequence[] args) throws XPathException {
+        if (args.length >= 5 && !args[4].isEmpty()) {
+            return FunSerialize.getSerializationProperties(this, args[4].itemAt(0));
+        }
+        return new Properties();
+    }
+
     @Override
     public Sequence eval(final Sequence[] args, final Sequence contextSequence)
             throws XPathException {
@@ -117,6 +141,9 @@ public abstract class AbstractCompressFunction extends BasicFunction {
         if (args[0].isEmpty()) {
             return Sequence.EMPTY_SEQUENCE;
         }
+
+        // serialization options for XML entries (optional map argument); reset on each call
+        serializationProperties = parseSerializationOptions(args);
 
         // use a hierarchy in the tar file?
         final boolean useHierarchy = args[1].effectiveBooleanValue();
@@ -375,6 +402,10 @@ public abstract class AbstractCompressFunction extends BasicFunction {
                 final String[] kvp = Option.parseKeyValuePair(param);
                 serializer.setProperty(kvp[0], kvp[1]);
             }
+        }
+        // an explicit serialization-options map argument takes precedence over the prolog option
+        for (final String key : serializationProperties.stringPropertyNames()) {
+            serializer.setProperty(key, serializationProperties.getProperty(key));
         }
     }
 
--- a/extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
+++ b/extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
@@ -100,10 +100,6 @@ public abstract class AbstractCompressFunction extends BasicFunction {
     private final static Logger logger = LogManager.getLogger(AbstractCompressFunction.class);
     public static final String METHOD_STORE = "store";
 
-    /** Serialization options from the optional map argument; reset on each eval and applied to XML
-     *  entries after (so overriding) the prolog {@code declare option exist:serialize}. */
-    private Properties serializationProperties = new Properties();
-
     public AbstractCompressFunction(final XQueryContext context, final FunctionSignature signature) {
         super(context, signature);
     }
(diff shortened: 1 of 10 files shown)


joewiz:

I've folded a small pre-existing fix into this PR (commit b680220c3e):


## Also includes: strip-prefix fix for the 4- and 5-arg forms

While building a collection-export endpoint against the new 5-arg form, a pre-existing bug surfaced: `compression:zip` / `compression:tar` read the `strip-prefix` argument only when **exactly 3** arguments were supplied (`args.length == 3`), so whenever a 4th (`encoding`) or 5th (`serialization-options`) argument was also given, `strip-prefix` was silently ignored and archive entries kept their full `/db/...` path instead of being archive-root-relative.

`strip-prefix` is always argument 2 across the 3-, 4- and 5-arg signatures, so the guard is now `args.length >= 3` (matching the `>= 4` form the `encoding` read just below already uses). Fixes both `zip` and `tar` via their shared `AbstractCompressFunction` base.

This predates the PR (the 4-arg `encoding` form has ignored `strip-prefix` all along), but it's folded in here because it lives in the function this PR is already editing and it makes the new 5-arg form actually usable for producing installable, root-relative archives. `CompressionTests` now asserts `strip-prefix` is applied at 4-arg and 5-arg, not just 3-arg.

extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
```diff
@@ -90,9 +92,17 @@ public abstract class AbstractCompressFunction extends BasicFunction {
...
+            "Serialization options applied to XML resources written to the archive, as either an " +
+                    "output:serialization-parameters element or a map(*) (e.g. map { \"indent\": false() }). " +
+                    "Standard W3C parameters use their string key; eXist extension parameters use their exist-namespace QName key. " +
+                    "Takes precedence over any prolog declare option exist:serialize.");
     private final static Logger logger = LogManager.getLogger(AbstractCompressFunction.class);
     public static final String METHOD_STORE = "store";
 
+    /** Serialization options from the optional map argument; reset on each eval and applied to XML
+     *  entries after (so overriding) the prolog {@code declare option exist:serialize}. */
+    private Properties serializationProperties = new Properties();
```
dizzzz: not sure if we need a `private Properties` definition here. I think it can be moved to the eval functions.

dizzzz: no this must move

dizzzz (changes requested):
Added comment.  @joewiz

extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

@dizzzz Agreed, and thanks — you're right that it must move. Fixed in 3ea1474bb6.

The field wasn't just redundant, it was a correctness hazard: a `BasicFunction` instance lives as long as the compiled expression, not the evaluation, so parsed options assigned to a field in `eval()` outlive the call that set them and are shared by every evaluation reaching the same compiled node.

The options are now a local in `eval()`, passed explicitly to the methods that actually serialize XML entries — `compressFromUri`, `compressElement`, `compressResource`, `compressCollection` and `getDynamicSerializerOptions`. `compressFile` is untouched, since it copies raw bytes and never serializes.

No behavior change: the options are still parsed once per call and still applied after (so overriding) the prolog `declare option exist:serialize`. The existing `CompressionTests` cover both the `map(*)` and `output:serialization-parameters` forms, plus the strip-prefix interaction, and all 12 still pass. Codacy/PMD on the changed file reports the same three pre-existing findings as before the change and nothing new.

Two things worth flagging: the branch is 112 commits behind develop (mergeState is still CLEAN, so no conflict — but a rebase before merge would be tidier), and this PR still carries the #6491 MapType commit it's stacked on, which drops out once #6491 lands.

duncdrum (commented):
A couple of small things I noticed on the new `serialization-options` argument — nothing blocking, both easy fixes. Nice addition otherwise, `compression:zip`/`tar` matching `fn:serialize`'s options handling is a good consistency win.

extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
```diff
@@ -110,6 +116,20 @@ private String removeLeadingOffset(String uri, final String stripOffset) {
...
+    /**
+     * Parses the optional serialization-options map argument (the 5th argument, if present) into the
+     * serialization properties applied to XML entries written to the archive.
+     *
+     * @param args the function arguments
+     * @return the supplied serialization properties, or an empty {@link Properties} when no map was given
+     */
+    private Properties parseSerializationOptions(final Sequence[] args) throws XPathException {
+        if (args.length >= 5 && !args[4].isEmpty()) {
+            return FunSerialize.getSerializationProperties(this, args[4].itemAt(0));
```
duncdrum, line 128: This skips the SEPM0009 parameter-consistency check that `fn:serialize` runs on the same input (`FunSerialize.validateSerializationParams`, currently private). So e.g. `compression:zip($sources, true(), "", "UTF8", map { "omit-xml-declaration": "yes", "standalone": "yes" })` silently serializes a non-conformant entry instead of raising `err:SEPM0009` the way `fn:serialize` would for the identical map — a bit at odds with the goal of matching `fn:serialize` behavior here.

Could expose `validateSerializationParams` the same way `getSerializationProperties` already is (`public static`, taking a `callingExpr`) and call it here too:

```java
private Properties parseSerializationOptions(final Sequence[] args) throws XPathException {
    if (args.length >= 5 && !args[4].isEmpty()) {
        final Properties props = FunSerialize.getSerializationProperties(this, args[4].itemAt(0));
        FunSerialize.validateSerializationParams(this, props);
        return props;
    }
    return new Properties();
}
```

extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
```diff
@@ -376,6 +400,10 @@ private void getDynamicSerializerOptions(final Serializer serializer) throws SAX
                 serializer.setProperty(kvp[0], kvp[1]);
             }
         }
+        // an explicit serialization-options map argument takes precedence over the prolog option
+        for (final String key : serializationProperties.stringPropertyNames()) {
+            serializer.setProperty(key, serializationProperties.getProperty(key));
+        }
```
duncdrum, line 406: Minor: `Serializer.setProperties(Properties)` already does this (and special-cases `SAX_LEXICAL_HANDLER`/`USE_CHARACTER_MAPS` along the way, which this loop skips). `file:sync` — the function this PR mirrors — already calls it this way (`Sync.java:416`/`455`), so reusing it here keeps the two in sync and is a touch shorter too.

```suggestion
        serializer.setProperties(serializationProperties);
```

extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Done in 62d1112e07. `FunSerialize.validateSerializationParams` is now `public static`, taking the calling expression, and `compression:zip`/`tar` call it on the options. There's a new XQSuite test for the `omit-xml-declaration` + `standalone` case.

extensions/modules/compression/src/main/java/org/exist/xquery/modules/compression/AbstractCompressFunction.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Done. One thing I noticed along the way: the storage `Serializer` saves the `use-character-maps` value that `setProperties` hands it, but nothing reads it. So character maps have no effect through this serializer, in `file:sync` as well as here. That predates this PR, so I've left it out, but it seems worth an issue of its own.

duncdrum: agreed let's open an issue

duncdrum (approved):

PR #6499, opened 2026-06-19
[bugfix] fixes potential NPE

reinhapa:
Fixes a NPE when `null` is passed into `setContent(Object)`

### DIFF
--- a/exist-core/src/main/java/org/exist/xmldb/LocalXMLResource.java
+++ b/exist-core/src/main/java/org/exist/xmldb/LocalXMLResource.java
@@ -552,12 +552,15 @@ public class LocalXMLResource extends AbstractEXistResource implements XMLResour
         root = null;
 
         switch (obj) {
+            case null -> {
+                // no action
+            }
             case Path path -> file = path;
             case java.io.File file1 -> file = file1.toPath();
             case AtomicValue atomicValue -> value = atomicValue;
             case InputSource source -> inputSource = source;
             case byte[] bytes -> content = new String(bytes, UTF_8);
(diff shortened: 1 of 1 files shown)


duncdrum (approved):

exist-core/src/main/java/org/exist/xmldb/LocalXMLResource.java
```diff
@@ -551,13 +551,15 @@ public void setContent(final Object obj) throws XMLDBException {
...
-            case byte[] bytes -> content = new String(bytes, UTF_8);
-            case null, default -> content = obj.toString();
+        if (obj != null) {
+            switch (obj) {
+                case Path path -> file = path;
+                case java.io.File file1 -> file = file1.toPath();
+                case AtomicValue atomicValue -> value = atomicValue;
+                case InputSource source -> inputSource = source;
+                case byte[] bytes -> content = new String(bytes, UTF_8);
+                default -> content = obj.toString();
```
line-o: Why did you choose not to handle the null case within the switch?

reinhapa: This will produce a codacy issue as the `content` is already set to `null`

line-o: Then I would prefer an early return

```
if (obj == null) { return; }

Comment (line-o) [reply exist-core/src/main/java/org/exist/xmldb/LocalXMLResource.java]: a noop would be another alternative

```
case null -> noop();
default -> content = obj.toString();

line-o (approved):

PR #6503, opened 2026-06-19
[refactor] Revise HttpClient creation and caching.

dizzzz:
Revise HttpClient creation and caching.
- reuse HttpClient when safe and possible
- this way we can reuse connection pooling over requests.
- add JMX monitoring

image

follow up for https://github.com/eXist-db/exist/pull/6482 

ToDo

- [x] (p)review PR by DWES before un-Draft
- [x] review by Joe/Claude
- [x] introduced new 'http-version' attribute.  Check if this is allowed in spec.

### DIFF
--- a/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientFactory.java
+++ b/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientFactory.java
@@ -24,7 +24,8 @@ package org.exist.xquery.modules.httpclient;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.mizosoft.methanol.Methanol;
-import org.exist.xquery.modules.httpclient.config.RequestOptions;
+import org.exist.xquery.modules.httpclient.config.HttpClientOptions;
+import org.exist.xquery.modules.httpclient.jmx.HttpClientCacheMonitor;
 
 import java.net.http.HttpClient;
 import java.time.Duration;
@@ -36,14 +37,14 @@ import java.time.Duration;
  * and transparently decodes gzip/deflate responses, and readTimeout gives a per-read (inactivity)
  * timeout that the bare JDK client lacks.</p>
  *
- * <p>Clients are cached by {@link RequestOptions} so that requests sharing the same options
+ * <p>Clients are cached by {@link HttpClientOptions} so that requests sharing the same options
  * reuse a single {@link HttpClient} instance (and its underlying connection pool). The cache
  * holds at most 25 entries and evicts entries that have not been accessed for one hour.
  * It is also cleared on JVM shutdown via a registered shutdown hook.</p>
  */
 class HttpClientFactory {
 
-    private static final Cache<RequestOptions, HttpClient> CLIENT_CACHE =
+    private static final Cache<HttpClientOptions, HttpClient> CLIENT_CACHE =
             Caffeine.newBuilder()
                     .maximumSize(25)
                     .expireAfterAccess(Duration.ofHours(1))
@@ -61,17 +62,17 @@ class HttpClientFactory {
     }
 
     /**
-     * Returns a cached {@link HttpClient} for the given {@link RequestOptions}, creating one
+     * Returns a cached {@link HttpClient} for the given {@link HttpClientOptions}, creating one
      * if no matching client exists in the cache yet.
      *
      * @param options the request options controlling redirect behaviour and timeouts
      * @return a configured {@link HttpClient}
      */
-    static HttpClient get(final RequestOptions options) {
+    static HttpClient get(final HttpClientOptions options) {
         return CLIENT_CACHE.get(options, HttpClientFactory::newClient);
     }
 
-    private static HttpClient newClient(final RequestOptions options) {
+    private static HttpClient newClient(final HttpClientOptions options) {
         final Methanol.Builder clientBuilder = Methanol.newBuilder()
                 .autoAcceptEncoding(true)
                 .followRedirects(options.followRedirect()
--- a/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientFactory.java
+++ b/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientFactory.java
@@ -1,0 +1,78 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.modules.httpclient;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.mizosoft.methanol.Methanol;
+
+import java.net.http.HttpClient;
+import java.time.Duration;
+
+/**
+ * Factory for creating and caching configured {@link HttpClient} instances.
+ *
+ * <p>Methanol augments java.net.http.HttpClient: autoAcceptEncoding advertises Accept-Encoding
+ * and transparently decodes gzip/deflate responses, and readTimeout gives a per-read (inactivity)
+ * timeout that the bare JDK client lacks.</p>
+ *
+ * <p>Clients are cached by {@link RequestOptions} so that requests sharing the same options
+ * reuse a single {@link HttpClient} instance (and its underlying connection pool). The cache
+ * is cleared on JVM shutdown via a registered shutdown hook.</p>
+ */
+class HttpClientFactory {
+
+    private static final Cache<RequestOptions, HttpClient> CLIENT_CACHE =
+            Caffeine.newBuilder().build();
+
+    static {
+        Runtime.getRuntime().addShutdownHook(new Thread(CLIENT_CACHE::invalidateAll,
+                "http-client-cache-shutdown"));
+    }
+
+    private HttpClientFactory() {
+        // utility class
+    }
+
+    /**
+     * Returns a cached {@link HttpClient} for the given {@link RequestOptions}, creating one
+     * if no matching client exists in the cache yet.
+     *
+     * @param options the request options controlling redirect behaviour and timeouts
+     * @return a configured {@link HttpClient}
+     */
+    static HttpClient get(final RequestOptions options) {
+        return CLIENT_CACHE.get(options, HttpClientFactory::newClient);
+    }
+
+    private static HttpClient newClient(final RequestOptions options) {
+        final Methanol.Builder clientBuilder = Methanol.newBuilder()
+                .autoAcceptEncoding(true)
+                .followRedirects(options.followRedirect()
+                        ? HttpClient.Redirect.NORMAL : HttpClient.Redirect.NEVER);
+        if (options.timeout() > 0) {
+            final Duration timeoutDuration = Duration.ofSeconds(options.timeout());
+            clientBuilder.connectTimeout(timeoutDuration).readTimeout(timeoutDuration);
+        }
+        return clientBuilder.build();
+    }
+}
--- a/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientFactory.java
+++ b/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientFactory.java
@@ -37,12 +37,17 @@ import java.time.Duration;
  *
  * <p>Clients are cached by {@link RequestOptions} so that requests sharing the same options
  * reuse a single {@link HttpClient} instance (and its underlying connection pool). The cache
- * is cleared on JVM shutdown via a registered shutdown hook.</p>
+ * holds at most 25 entries and evicts entries that have not been accessed for one hour.
+ * It is also cleared on JVM shutdown via a registered shutdown hook.</p>
  */
 class HttpClientFactory {
 
     private static final Cache<RequestOptions, HttpClient> CLIENT_CACHE =
-            Caffeine.newBuilder().recordStats().build();
+            Caffeine.newBuilder()
+                    .maximumSize(25)
+                    .expireAfterAccess(Duration.ofHours(1))
+                    .recordStats()
+                    .build();
 
     static {
         HttpClientCacheMonitor.registerIfAbsent(CLIENT_CACHE);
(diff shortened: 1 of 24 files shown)


joewiz (commented):
(Clearing the original review since a new one is about ready.)

dizzzz:
Thnx!  I wonder why I added username/wordt to the record as this should be request based. It was probably to late :-)  see more valid points, I will address these.

joewiz:

Re-reviewing... The approach to caching and reusing the client (plus the JMX visibility) is a real improvement over building and tearing down a client per request. A few things stood out on a close read, grouped by severity. Line numbers are approximate.

## Blocking — correctness

**1. Evicted/expired clients are never closed (thread + connection leak).** `HttpClientFactory` builds the cache with `maximumSize(25)` + `expireAfterAccess(1h)` but no removal listener, and the only place a client is closed is `HttpClientCacheMonitor.reset()` (JMX / shutdown hook). Caffeine eviction (size cap or hourly idle) drops the entry without calling `HttpClient.close()`, so each evicted `java.net.http.HttpClient` keeps its selector thread and keep-alive sockets until GC finalizes it. The previous code closed every client via try-with-resources, so this is a regression that undercuts the connection-reuse goal under any churning set of options. Suggest a Caffeine `removalListener((key, client, cause) -> client.close())` on the builder, so every removal cause (size, expiry, explicit) closes the client — and then `reset()` can just `invalidateAll()`.

**2. Default HTTP version changed from negotiated HTTP/2 to pinned HTTP/1.1.** `newClient()` now sets `.version(options.httpVersion())`, and `HttpClientOptions.DEFAULTS` is `HTTP_1_1`. The old `buildClient()` never called `.version()`, so it used the library default (HTTP/2, negotiating down to 1.1 via ALPN). No existing query sets the new `http-version` attribute, so every caller silently drops to HTTP/1.1 — a behavior change for anyone talking to an HTTP/2 endpoint. If 1.1 is intended as the new default, it's worth calling out explicitly; otherwise default to `HTTP_2`.

## Should fix — contract / behavior

**3. `getHitRate()` doesn't match its own contract or test.** The `HttpClientCacheMXBean` javadoc says "Returns 0.0 when no requests have been made yet," but the implementation returns `cache.stats().hitRate()`, which Caffeine defines as 1.0 when `requestCount() == 0`. The test is named `hitRateIsZeroWhenNoRequests` yet asserts `1.0`. So a cold cache reports a perfect hit rate, and the doc, the impl, and the test name disagree three ways. Either special-case `requestCount() == 0` to return 0.0 (matching the doc) or fix the doc + test name to reflect the 1.0 Caffeine semantics.

**4. `auto-accept-encoding` / `http-version` are read only from the eXist namespace.** In `RequestOptionsParser`, `follow-redirect` is read with no namespace, but `auto-accept-encoding` and `http-version` are read only from `Namespaces.EXIST_NS`. A user who writes a bare (no-prefix) `auto-accept-encoding="false"` has it silently ignored (the `autoAcceptEncodingNoNamespaceIgnored` test pins this). If the namespace requirement is intentional for these eXist extensions, the asymmetry with `follow-redirect` is at least a documentation/usability footgun worth a note; if not, they should fall back to the no-namespace attribute like their sibling.

## Lifecycle & integration

**5. `reset()` can close clients out from under in-flight requests.** `reset()` does `cache.asMap().values().forEach(HttpClient::close)` then `invalidateAll()`. Because clients are now shared, a request that obtained a client and is mid-`send()` will have it closed by a concurrent JMX `reset()` (or the shutdown hook), surfacing as `HC001`. Hard to fully avoid with shared clients, but worth being aware of — at minimum it argues for `reset()` being an explicit, rarely-used operation.

**6. The JMX bean and the cache lifecycle sit outside eXist's management/lifecycle layer.** `HttpClientCacheMonitor.register()` registers directly on `ManagementFactory.getPlatformMBeanServer()` with a fixed, non-instance-scoped `ObjectName`, and `HttpClientFactory` wires cleanup through a JVM `Runtime` shutdown hook over a JVM-global static cache. eXist already centralizes MBean registration and teardown in `org.exist.management.Agent` / `AgentFactory` (per-instance naming, unregistered in `closeDBInstance`). As written, in a multi-instance or embedded/test JVM (BrokerPool started and stopped repeatedly) the MBean and cache outlive the database instance, the second `register()` is a silent no-op pointing at a now-dead cache, and clients are only closed at JVM exit rather than on database shutdown. Registering through the Agent and tying `reset()` to the module/BrokerPool shutdown path would make this behave like the rest of eXist's JMX and free resources on restart.

**7. `getCacheSize()` forces cache maintenance on every read.** It calls `cache.cleanUp()` before `estimatedSize()`. `cleanUp()` drains Caffeine's eviction/maintenance work synchronously on the calling (JMX) thread, so a frequently-polling monitor repeatedly does write-path maintenance contending with request threads. `estimatedSize()` on its own is the intended cheap read.

## Minor

**8. Shared-client cache key — worth a guard/comment.** Excluding credentials from the key is correct today because auth is applied per-`HttpRequest`, not on the client. But that invariant is unguarded: if a later change moves auth/proxy/SSL onto the client builder without adding it to `HttpClientOptions`, the cache would hand a client carrying one caller's auth to another. (Also: for connection-bound schemes like NTLM/Negotiate, a shared keep-alive connection can already carry auth state across callers.) A short comment on `newClient()` stating "configure nothing here that isn't part of the cache key" would protect it.

**9. `getCachedClientsSummary()` omits `httpVersion`.** It hand-concatenates `followRedirect`/`timeout`/`autoAcceptEncoding`, so two cached clients differing only in HTTP version look identical in JMX, and the string must be kept in sync as the record grows. Since `HttpClientOptions` is a record, `opts.toString()` already gives a faithful, self-maintaining rendering.

**10. Small cleanups.** `UserCredentials` has leftover scratch comments (`//user`, `//`) on its record components; `RequestBuilder` constructs an all-defaults `RequestOptions` inline where a `RequestOptions.DEFAULTS` constant (matching the per-component `DEFAULTS`) would read better; the three `parseBooleanAttr` overloads are a copy-paste chain that could collapse to one resolver.

The leak (1) and the version default (2) are the two I'd suggest be resolved before merge.

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientFactory.java
```diff
@@ -0,0 +1,97 @@
...
+ * It is also cleared on JVM shutdown via a registered shutdown hook.</p>
+ */
+public class HttpClientFactory {
+
+    private static final Cache<HttpClientOptions, HttpClient> CLIENT_CACHE =
+            Caffeine.newBuilder()
+                    .maximumSize(25)
+                    .expireAfterAccess(Duration.ofHours(1))
+                    .recordStats()
+                    .build();
```
dizzzz, line 63: Add `.removalListener()` to call close HttpClient to cleanup resources.

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientFactory.java
```diff
@@ -0,0 +1,97 @@
...
+     * Closes all cached {@link HttpClient} instances and removes them from the cache.
+     */
+    static void reset() {
+        MONITOR.reset();
+    }
+
+    private static HttpClient newClient(final HttpClientOptions options) {
+        final Methanol.Builder clientBuilder = Methanol.newBuilder()
+                .autoAcceptEncoding(options.autoAcceptEncoding())
+                .version(options.httpVersion())
```
dizzzz, line 98: The original library uses HTTP/2 (negotiate to v1) by default. Setting 1_1 is safe/backwards compatible. To discuss.

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/jmx/HttpClientCacheMonitor.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    public long getEvictionCount() {
+        return cache.stats().evictionCount();
+    }
+
+    @Override
+    public double getHitRate() {
+        return cache.stats().hitRate();
```
dizzzz: Claude:

getHitRate() doesn't match its own contract or test. The HttpClientCacheMXBean javadoc says "Returns 0.0 when no requests have been made yet," but the implementation returns cache.stats().hitRate(), which Caffeine defines as 1.0 when requestCount() == 0. The test is named hitRateIsZeroWhenNoRequests yet asserts 1.0. So a cold cache reports a perfect hit rate, and the doc, the impl, and the test name disagree three ways. Either special-case requestCount() == 0 to return 0.0 (matching the doc) or fix the doc + test name to reflect the 1.0 Caffeine semantics.

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/jmx/HttpClientCacheMonitor.java
```diff
@@ -0,0 +1,121 @@
...
+            throw new IllegalStateException("Unexpected malformed JMX object name: " + OBJECT_NAME, e);
+        } catch (final Exception e) {
+            LOG.warn("Failed to register HttpClientCache JMX MBean: {}", e.getMessage(), e);
+        }
+        return monitor;
+    }
+
+    @Override
+    public long getCacheSize() {
+        cache.cleanUp();
```
dizzzz: consider: remove. Unwanted side effect.

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/jmx/HttpClientCacheMXBean.java
```diff
@@ -0,0 +1,86 @@
...
+import org.exist.xquery.modules.httpclient.HttpClientFactory;
+
+/**
+ * JMX MXBean interface for monitoring the {@link HttpClientFactory} client cache.
+ *
+ * <p>Exposes cache-level statistics (size, hit/miss counts, eviction count) as well as
+ * a human-readable summary of the currently cached {@link java.net.http.HttpClient}
+ * instances and their configuration.</p>
+ */
+public interface HttpClientCacheMXBean {
```
dizzzz, line 32: Claude:

6. The JMX bean and the cache lifecycle sit outside eXist's management/lifecycle layer. HttpClientCacheMonitor.register() registers directly on ManagementFactory.getPlatformMBeanServer() with a fixed, non-instance-scoped ObjectName, and HttpClientFactory wires cleanup through a JVM Runtime shutdown hook over a JVM-global static cache. eXist already centralizes MBean registration and teardown in org.exist.management.Agent / AgentFactory (per-instance naming, unregistered in closeDBInstance). As written, in a multi-instance or embedded/test JVM (BrokerPool started and stopped repeatedly) the MBean and cache outlive the database instance, the second register() is a silent no-op pointing at a now-dead cache, and clients are only closed at JVM exit rather than on database shutdown. Registering through the Agent and tying reset() to the module/BrokerPool shutdown path would make this behave like the rest of eXist's JMX and free resources on restart.

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/jmx/HttpClientCacheMonitor.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Override
+    public double getHitRate() {
+        return cache.stats().hitRate();
+    }
+
+    @Override
+    public String getCachedClientsSummary() {
+        return cache.asMap().keySet().stream()
+                .map(opts -> "followRedirect=" + opts.followRedirect()
```
dizzzz: use `opts.toString()` instead for this record.

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/config/UserCredentials.java
```diff
@@ -0,0 +1,46 @@
...
+ * {@code auth-method}, {@code send-authorization}) as defined by the EXPath HTTP Client
+ * specification.</p>
+ *
+ * @param username          the username for authentication
+ * @param password          the password for authentication
+ * @param authMethod        the authentication method (e.g., "Basic", "Digest")
+ * @param sendAuthorization whether to send the Authorization header preemptively
+ */
+public record UserCredentials(
+        String username, //user
```
dizzzz: cleanup old // notes

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestOptionsParser.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+        try {
+            return Integer.parseInt(timeoutStr);
+        } catch (final NumberFormatException e) {
+            throw new XPathException((org.exist.xquery.Expression) null,
+                    HttpClientModule.HC005, "Invalid timeout value: " + timeoutStr);
+        }
+    }
+
+    private static boolean parseBooleanAttr(final Element elem, final String name, final boolean defaultValue) {
```
dizzzz: Revise 3 almost identical methods.

dizzzz: name=namespace

duncdrum (approved):

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientModule.java
```diff
@@ -45,31 +45,46 @@
  */
 public class HttpClientModule extends AbstractInternalModule {
 
+    /** Namespace URI for the module. */
```
line-o: I do not find these comments on object properties here really add more detail and would rather remove them. @dizzzz do you agree or would you want to keep them?

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestOptionsParser.java
```diff
@@ -0,0 +1,130 @@
...
+        return new RequestOptions(requestOptions, responseOptions, userCredentials);
+    }
+
+    private static HttpClient.Version parseHttpVersion(final Element reqElem, final HttpClient.Version defaultVersion) throws XPathException {
+        final String value = getAttributeValueinNamespace(reqElem, Namespaces.EXIST_NS, "http-version");
+        if (value == null) {
+            return defaultVersion;
+        }
+        return switch (value) {
+            case "1.1" -> HttpClient.Version.HTTP_1_1;
```
line-o, line 85: Do we not also allow HTTP 1.0 requests?

line-o (commented):
Really nice work @dizzzz 
I have some questions which we can also address later

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/HttpClientModule.java
dizzzz: I agree; will remove them

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestOptionsParser.java
dizzzz: @line-o AFAIK HTTP/1 not supported anymore. 

https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.Version.html 

image

line-o: A certain bird tweeted at me:

> Could you clarify which "version 1.0" you mean? A few possibilities:

> An HTTP client library (e.g. axios 1.0, requests, Node's http module, etc.)
> The HTTP protocol itself (HTTP/1.0 vs HTTP/1.1 or HTTP/2)

And then when I made clear I am asking about protocol version support:

> You can force HTTP/1.0 in most tools (e.g. curl --http1.0), and most servers will still respond, but it's a legacy mode, not something actively developed against.

So which one is it here?

line-o: OK java.net.http does not support making requests over `HTTP/1.0` we need to document that.

line-o: Apache HttpClient allows that and we should have an actionable error message when a request has the attribute value set to 1.0

line-o: As this _can_ be viewed as a regression - it is certainly a breaking change

dizzzz: Added exception.

line-o (approved):
Thanks @dizzzz

PR #6496, opened 2026-06-18
XSD 1.1 validation and catalog  fixes. 

duncdrum:
Stumbled across this while working on exist's native schemas which will be a follow-up PR.  🐇 🕳️ 

## Summary

- `validation:jaxp()` now validates XSD 1.1 schemas (`vc:minVersion="1.1"`), transparently routed through a dedicated `javax.xml.validation` pipeline since the bundled Xerces fork only wires XSD 1.1 support there, not into the dynamic-discovery SAX parser.
- `validation:jaxv()` gains a 4th, optional `$catalogs` argument, matching `jaxp()`'s existing catalog support (system catalog / explicit `.xml` catalog / directory-search collection catalog).
- Database-stored catalogs are streamed to the resolver via SAX instead of round-tripping through a `String`.
- Closes #1975, #2476 (regression tests added; underlying fixes already present).

## What changed

- `Jaxp.java` / `Jaxv.java`: XSD 1.1 detection + validator pipeline, shared catalog-argument `Shared.resolveCatalogArgument`, Subject-scoped and Caffeine-backed detection cache.
- `ResolverFactory.java` / `RecordingContentHandler.java`: SAX-streamed catalog loading without re-fetching from the database on xmlresolver's second pass.
- `SearchResourceResolver.java`: directory-search catalogs now work with both the SAX pipelineeline.
- Security: the XSD 1.1 detection peek only follows same-origin `schemaLocation` hints — covered by `JaxpSchemaLocationSecurityTest`.

## Test plan

- [x] Full `exist-core` suite: 7029 tests, 0 failures/errors
- [x] All CI checks green (Codacy, License check, container build, W3C XQTS, unit + integration
- [x] New/updated regression tests: `JaxpXsdCatalogTest`, `JaxvTest`, `JaxpSchemaLocationSecurityTest`, `JaxpXsd11DetectionCacheTest`, `IsMissingElementDeclarationTest`, `CatalogResolutionRegressionTest`,
`TournamentSchemaLanguageComparisonTest`

PR #6505, opened 2026-06-20
exist native schemas

duncdrum:
what I actually wanted to do. 

extends #6496


duncdrum:
ok saga is almost complete. I will do one final round of rebases and squashes and have Claude turn this into more manageable PRs for review.

Issue #6463, opened 2026-06-10
Define eXist's resource-naming contract: name vs URI (5 decisions)


joewiz:
[This issue was co-authored with Claude Code. -Joe]

## Decide eXist's resource-naming contract (5 questions, recommended answers inline)

eXist has never written down what a resource/collection **name** is versus its **URI path**. `XmldbURI` stores names percent-encoded, and each surface — REST, WebDAV, XML-RPC, existdb-openapi — *and even core's own `xmldb:store` vs `doc()`* — encodes or decodes at a different boundary. So the *same logical resource* can be unreachable, stored under a different name, or (verified below) **silently overwritten with data loss**, depending on which function created it. This is the root of #3795 (meta), #3665, #1824 (now fixed), #5299, #1612.

**We need a maintainer decision on the five questions below.** Each has a recommended answer; an "agree" (or a redirect) on each unblocks the implementation PRs. The rest of this issue is the evidence for those answers — skim it only if a recommendation looks wrong.

### Precedent — eXide and TEI Publisher already do it right

This isn't a new contract to invent; it's the one eXide has shipped for years. eXide's file browser lists the very same `/db/smoke` and shows `café-ünïcode.xml` and `quote'name.xml` correctly, because `modules/api/db.xqm` decodes every name and path before returning them:

```xquery
"name": xmldb:decode-uri(xs:anyURI($child)),
"path": xmldb:decode-uri(xs:anyURI($child-path)),
```

Same database, same `xmldb:get-child-resources()` underneath — eXide decodes, existdb-openapi returns the raw stored form. eXide decodes the `path` (its operation key) too, not just the display `name`, and round-trips fine because `xmldb:*` re-encodes on the way back in. So Decision 1 below is **live behavior we already ship in our flagship admin tool**.

### Evidence — cross-surface matrix (from PR #6450)

A new characterization test stores a corpus of awkward names via WebDAV and reads each back **by the requested name** via WebDAV and REST. On `develop`:

```
requested          stored-as            dav-read  rest-read
with space.xml     with%20space.xml ≠   PASS      PASS
a+b.xml            a%2Bb.xml ≠          PASS      FAIL
a@b.xml            a%40b.xml ≠          PASS      FAIL
a&b.xml            a%26b.xml ≠          PASS      FAIL
report(2024).xml   report%28…%29.xml ≠  PASS      FAIL
o'brien.xml        o%27brien.xml ≠      PASS      FAIL
[...]

PR #6508, opened 2026-06-21
[feature] Resource-naming contract prototype: decisions 1, 2, 3, 5 (draft illustration for #6463)

joewiz:

## Why this is a draft

This PR is **an illustration, not a merge request**. It implements the recommended answers in #6463 as running, tested code so that the contract discussion (which @dizzzz rightly flagged as complex, and worth a community call or workshop) has a concrete artifact to react to: you can see exactly what each decision costs, what it changes, and where the boundaries are. It is intentionally a **draft**; if the discussion wants it as a real merge candidate, we can un-draft it and add the remaining pre-merge gate (a full XQTS run).

It merges cleanly onto current `develop` today (no conflicts).

## Which decisions of #6463 it reflects

| #6463 decision | This PR | Where |
|---|---|---|
| **1 — apps speak raw decoded UTF-8; REST/WebDAV keep wire encoding but must round-trip** | XML-RPC now **decodes** resource/collection names on read; REST/WebDAV verified to already converge (Jackrabbit), aligned the codec to RFC 3986 pchar to keep them in lockstep | `RpcConnection.java`, conformance test |
| **2 — literal `%` (the non-injective / silent-collision defect)** | **fully handled, not postponed** (per @duncdrum's request to see the `%` fallout in one branch). A **bijective lenient codec** (`encodeForURILenient`/`decodeForURI`) escapes `%`→`%25` at the store boundary, so `café.xml` and the literal `caf%C3%A9.xml` no longer collide; and the **read path resolves a literal-`%` name by its decoded form** (`doc("/db/x/50%.xml")` → `50%25.xml`) across `doc()`, `collection()`, and the `xmldb:` reads | `URIUtils.java`, `DocUtils.java`, `ExtCollection.java`, `XMLDBAbstractCollectionManipulator.java`, codec test |
| **3 — allow full Unicode; reject only `/` and control/NUL** | `xmldb:store`/`create-collection` **accept a raw space** (and other Unicode) instead of throwing `FORG0001` — and the **read path** now resolves a raw-space db-path too (`doc()` / `doc-available()`), closing the write-accepts/read-rejects gap | `XMLDBStore.java`, `URIUtils.java`, `DocUtils.java`, `FunDocAvailable.java` |
| **5 — `doc()`/`collection()` normalize a bare db-path the way `store` does** | `doc()` and `collection()` canonicalize a bare db-path to its stored key (decode-then-encode, idempotent on already-encoded paths); **extended to the `xmldb:` collection read functions**, so a decoded or descriptor-derived literal path resolves the stored key | `DocUtils.java`, `ExtCollection.java`, `XMLDBAbstractCollectionManipulator.java` |

**Decision 4** (guard non-`/db` paths at the API surface, leave core alone) is deliberately **not** in this PR — it lives at the application-API layer (e.g., existdb-openapi), per the issue.

The codec is proven to **converge across surfaces**: a stored-form oracle asserts every surface stores exactly `encodeForURILenient(name)` and decodes back to the original, over an awkward-name corpus (space, non-ASCII, sub-delimiters, parentheses, apostrophe, literal `%`, and filesystem-hostile chars).

## Benefits

- **Ends the `store`-normalizes / `doc`-doesn't asymmetry.** `doc($col || "/" || $name)` after `xmldb:store($col, $name, …)` now resolves for any name — no more `doc(xmldb:encode-uri($n))` dance.
- **Fixes silent data loss.** The non-injective encoding (`café.xml` vs literal `caf%C3%A9.xml` → same key, one destroyed) becomes bijective; distinct names get distinct keys.
- **One canonical codec** across `xmldb:store`/`create-collection`, `doc()`, `collection()`, the `xmldb:` read functions, and XML-RPC — verified converging, not asserted.
- **Raw spaces and full Unicode just work** end to end — accepted at the write boundary (no `FORG0001` surprise) and resolvable on the read path (`doc("/db/x/with space.xml")` after `xmldb:store("/db/x", "with space.xml", …)` now hits).
- **Literal `%` is fully handled** — stored without data loss *and* resolvable by its decoded name on read (`doc()`, `collection()`, `xmldb:` reads), not postponed.
- A package's EXPath repo collection is **addressable by its own descriptor-derived name** (the manifestation written up in #6463's latest comment).

## Is it a breaking change? What would a developer need to adjust?

**No data migration.** The stored form of every name that is creatable on `develop` today is **unchanged** — existing databases are unaffected. The only names whose stored form differs are ones that previously could not be stored at all (a raw space → `FORG0001`) or that were silently colliding (literal `%`).

The behavior changes a developer might notice, and how to adjust:

1. **XML-RPC listings now return decoded UTF-8 names** (decision 1). A client that consumed `RpcConnection` listing results and decoded them itself, or compared them against the percent-encoded form, will see the already-decoded name. **Adjust:** treat XML-RPC resource/collection names as raw decoded UTF-8 (the same contract REST clients follow after decoding the wire form).
2. **`xmldb:store` / `xmldb:create-collection` no longer throw `FORG0001` on a raw space** (decision 3) and now store a literal-`%` name bijectively (decision 2). Code that *relied on the throw* to reject such names will no longer get the exception. **Adjust:** if rejection is intended, validate the name explicitly instead of depending on the codec to throw.
3. **Literal-`%` names store and resolve distinctly.** `a%20b.xml` now stores as `a%2520b.xml` instead of colliding with `a b.xml`, and a literal-`%` name resolves by its decoded form on read (see the full-fallout section). This affects only names that were previously unsafe or data-losing — no correctly-stored data moves. The one caveat is the residual ambiguity above: a name that literally *is* a valid escape is addressed by its encoded form.
4. **`doc()` / `collection()` / `xmldb:` reads may now resolve a decoded path that previously missed.** This is strictly additive for callers that currently get an empty result and want a hit; the honest risk is code that *branched on the miss* (treated empty as "absent"). **Adjust:** none expected, but the full XQTS run is the gate to confirm nothing leaned on the old miss.

**On read-path validity:** the `doc()` / `doc-available()` read path previously threw `FODC0005` on a raw-space db-path (three separate `java.net.URI` parse points). All three now defer to the DB normalization **for db-paths** (`/db` or `xmldb:`), so a raw-space name resolves — while any genuinely malformed **non-db** URI (`%gg`, `:/`, a Windows backslash path) still raises `FODC0005`, exactly as the qt3tests `fn/doc` cases require.

## Literal `%` — the full fallout (per @duncdrum's request)

The branch handles `%` rather than postponing it, so the trade-off is visible in one place rather than deferred to a Phase 2.

- **Write — solved, no data loss.** `encodeForURILenient` escapes a literal `%`→`%25`, so a name *containing* `%` and a name that *is* a percent-escape get distinct stored keys. The old non-injective encoding silently overwrote one with the other; that is gone.
- **Read — resolved by decoded name, via decode-then-encode.** The read path canonicalizes a db-path by `decodeForURI` then `encodeForURILenient` — treating it as decoded UTF-8 (decision 1) and re-encoding to the stored key. Because `decodeForURI` is the exact inverse of `encodeForURILenient`, this maps **both** a decoded display name **and** an already-encoded path to the same key, so pre-encoding callers (`doc(xmldb:encode-uri($n))`, TEI Publisher) keep working unchanged.
- **The one irreducible residual.** A name that *literally is a valid escape* (`a%20b.xml`, `%25.xml`) is read as its decoded meaning (`a b.xml`, `%.xml`), so it is addressed by its encoded form, not that literal string. This is inherent to a single decoded-wire contract (you cannot have `a%20b.xml` mean both "the literal name" and "a b.xml" on the same wire). It is pinned exactly by the round-trip oracle: a name resolves by its decoded form **iff `decodeForURI(name) == name`** — i.e. every normal name, plus a literal `%` that is not itself a valid escape (`50%.xml`, `100%done%.xml` resolve; `a%20b.xml` is reached by its encoded form).

No stored data migrates: the only names whose stored form is *new* are ones that previously could not be stored without collision (a literal `%`) — there was no correct prior form to migrate from.

## Tests

- `ResourceNameCodecTest` — the codec in isolation: bijective round-trip of the full corpus incl. `50%.xml` / `a%20b.xml` / `%25.xml`, pins canonical stored forms, demonstrates the legacy collision/rejection.
- `ResourceNamingXmldbRoundTripTest` — the `xmldb:` write boundary against a real database: every stored key equals `encodeForURILenient(name)` and decodes back.
- `XmldbCollectionAddressByDecodedNameTest` — the `xmldb:` read boundary: an `${…}`-style awkward-named collection is addressable by its decoded literal (fails on the pre-fix baseline, passes here), plus `literalPercentCollectionResolvesByDecodedName` (a literal-`%` collection resolves via `xmldb:collection-available` and `fn:collection`).
- `DocTest` — read-path raw space and literal `%`: `doc_resolvesRawSpaceDbPathOnRead`, `doc_resolvesLiteralPercentNameByDecodedForm` (store `50%.xml` → `50%25.xml` with no data loss, then resolve by both the decoded and the encoded form), and `doc_malformedNonDbUriStillRaisesFODC0005` (the qt3tests garbage cases stay `FODC0005`).
- `XPathQueryTest` (150) — regression guard for the shared `collection()` / `getLocalCollection` changes: stays green.
- `ResourceNamingConformanceTest` (webdav module, cross-surface) — the stored-form oracle: REST/WebDAV/XML-RPC converge on one key; documents the remaining `KNOWN_FAILURES` (`:` colon, etc.) with rationale.
- `XmlRpcTest` — XML-RPC read decode.

**Pre-merge gate not yet run:** a full XQTS pass (the issue's stated condition for the decision-5 core change). This is why it's a draft.

## What this deliberately leaves for follow-up

Nothing in decisions 2, 3, or 5 is deferred. What this branch intentionally does *not* include, so the scope stays reviewable:

- **Decoded `xmldb:get-child-resources` / `-collections` return.** These now *resolve* a decoded path, but still *return* the encoded stored names (`caf%C3%A9.xml`, not `café.xml`) — the long-standing `//TODO: decode names?`. Changing the default return would break every caller that expects the encoded form, so the right shape is an **opt-in `$decode` argument** (already prototyped on a separate branch). Until then, callers decode the result themselves, exactly as eXide does (`xmldb:decode-uri`).
- **XML-RPC write-side decode.** XML-RPC read/listing methods decode names (decision 1, surface 3); the write methods (`storeResource`, `createCollection`, …) accepting decoded names is the symmetric follow-up.
- **Colon (`:`) and the other `KNOWN_FAILURES`.** `a:b.xml` still does not round-trip cross-surface — `XmldbURI` wraps `java.net.URI`, where `a:` parses as a scheme. This is recorded in the conformance ratchet, not silently broken; a real fix needs the storage-layer change below.
- **The `%`-that-is-a-valid-escape residual.** `a%20b.xml` is read as its decoded meaning (`a b.xml`); see the full-fallout section. Inherent to a single decoded-wire contract, not a TODO.
- **Decision 4** (validate/400 a non-`/db` path) — by design this lives at the application-API layer (existdb-openapi), not core.
- **The application-API half of decision 1** (existdb-openapi returning/accepting decoded UTF-8) — a separate PR in that repo.
- **Phase 3** — storing names as raw UTF-8 in `collections.dbx` (no encoding at the persistence layer at all), which would also retire the colon residual. Large; captured for direction only, per the issue.
- **The normative contract document** — to be written once the decisions are ratified.

### DIFF
--- a/exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
+++ b/exist-core/src/main/java/org/exist/xmlrpc/RpcConnection.java
@@ -86,6 +86,7 @@ import org.exist.xmlrpc.function.XmlRpcDocumentFunction;
 import org.exist.xmlrpc.function.XmlRpcFunction;
 import org.exist.xquery.*;
 import org.exist.xquery.util.HTTPUtils;
+import org.exist.xquery.util.URIUtils;
 import org.exist.xquery.value.*;
 import org.exist.xupdate.Modification;
 import org.exist.xupdate.XUpdateProcessor;
@@ -468,7 +469,10 @@ public class RpcConnection implements RpcAPI {
                         final Permission perms = doc.getPermissions();
 
                         final Map<String, Object> hash = new HashMap<>(5);
-                        hash.put("name", doc.getFileURI().toString());
+                        // Resource-naming contract (eXist-db/exist#6463, decision 1): XML-RPC is a
+                        // decoded-string protocol, so names are returned in decoded display form,
+                        // not the percent-encoded stored form. See URIUtils.decodeForURI.
+                        hash.put("name", URIUtils.decodeForURI(doc.getFileURI().toString()));
                         hash.put("owner", perms.getOwner().getName());
                         hash.put("group", perms.getGroup().getName());
                         hash.put("permissions", perms.getMode());
@@ -477,14 +481,14 @@ public class RpcConnection implements RpcAPI {
                     }
                 }
                 for (final Iterator<XmldbURI> i = collection.collectionIterator(broker); i.hasNext(); ) {
-                    collections.add(i.next().toString());
+                    collections.add(URIUtils.decodeForURI(i.next().toString()));
                 }
             }
 
             final Permission perms = collection.getPermissionsNoLock();
             desc.put("collections", collections);
             desc.put("documents", docs);
-            desc.put("name", collection.getURI().toString());
+            desc.put("name", URIUtils.decodeForURI(collection.getURI().toString()));
             desc.put("created", Long.toString(collection.getCreated()));
             desc.put("owner", perms.getOwner().getName());
             desc.put("group", perms.getGroup().getName());
@@ -510,7 +514,7 @@ public class RpcConnection implements RpcAPI {
             return this.<Map<String, Object>>readDocument(resourceUri).apply((document, broker, transaction) -> {
                 final Map<String, Object> hash = new HashMap<>(11);
                 final Permission perms = document.getPermissions();
-                hash.put("name", resourceUri.toString());
+                hash.put("name", URIUtils.decodeForURI(resourceUri.toString()));
                 hash.put("owner", perms.getOwner().getName());
                 hash.put("group", perms.getGroup().getName());
                 hash.put("permissions", perms.getMode());
@@ -576,12 +580,12 @@ public class RpcConnection implements RpcAPI {
             final List<String> collections = new ArrayList<>();
             if (collection.getPermissionsNoLock().validate(user, Permission.READ)) {
                 for (final Iterator<XmldbURI> i = collection.collectionIterator(broker); i.hasNext(); ) {
-                    collections.add(i.next().toString());
+                    collections.add(URIUtils.decodeForURI(i.next().toString()));
                 }
             }
             final Permission perms = collection.getPermissionsNoLock();
             desc.put("collections", collections);
-            desc.put("name", collection.getURI().toString());
+            desc.put("name", URIUtils.decodeForURI(collection.getURI().toString()));
             desc.put("created", Long.toString(collection.getCreated()));
             desc.put("owner", perms.getOwner().getName());
             desc.put("group", perms.getGroup().getName());
@@ -967,7 +971,7 @@ public class RpcConnection implements RpcAPI {
             return this.<List<String>>readCollection(collUri).apply((collection, broker, transaction) -> {
                 final List<String> list = new ArrayList<>();
                 for (final Iterator<XmldbURI> i = collection.collectionIterator(broker); i.hasNext(); ) {
-                    list.add(i.next().toString());
+                    list.add(URIUtils.decodeForURI(i.next().toString()));
                 }
                 return list;
             });
@@ -988,7 +992,7 @@ public class RpcConnection implements RpcAPI {
             return this.<List<String>>readCollection(collUri).apply((collection, broker, transaction) -> {
                 final List<String> list = new ArrayList<>();
                 for (final Iterator<DocumentImpl> i = collection.iterator(broker); i.hasNext(); ) {
-                    list.add(i.next().getFileURI().toString());
+                    list.add(URIUtils.decodeForURI(i.next().getFileURI().toString()));
                 }
                 return list;
             });
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
@@ -34,6 +34,7 @@ import org.exist.storage.lock.LockManager;
 import org.exist.storage.lock.ManagedDocumentLock;
 import org.exist.util.LockException;
 import org.exist.xmldb.XmldbURI;
+import org.exist.xquery.util.URIUtils;
 import org.exist.xquery.*;
 import org.exist.xquery.functions.xmldb.XMLDBModule;
 import org.exist.xquery.value.*;
@@ -81,7 +82,15 @@ public class ExtCollection extends BasicFunction {
         if (args.length == 0 || args[0].isEmpty()) {
             collectionUri = null;
         } else {
-            collectionUri = asUri(args[0].itemAt(0).getStringValue());
+            final String raw = args[0].itemAt(0).getStringValue();
+            // Resource-naming contract (eXist-db/exist#6463, decisions 1 + 2 + 5): canonicalize a bare
+            // db-path to its stored key by decode-then-encode, so collection("/db/x/café-col") and a
+            // literal '%' resolve, mirroring fn:doc. A scheme-ful URI (xmldb:, file:, ...) is left
+            // untouched so its scheme/authority are not rewritten.
+            final String effective = raw.startsWith(XmldbURI.ROOT_COLLECTION)
+                    ? URIUtils.encodePathForURILenient(URIUtils.decodePathForURI(raw))
+                    : raw;
+            collectionUri = asUri(effective);
         }
 
         return getCollectionItems(new URI[] { collectionUri });
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/ExtCollection.java
@@ -122,7 +122,17 @@ public class ExtCollection extends BasicFunction {
 
         } else {
             final MutableDocumentSet ndocs = new DefaultDocumentSet();
-            final XmldbURI uri = XmldbURI.create(collectionUri);
+            final XmldbURI uri;
+            try {
+                // Resource-naming contract (eXist-db/exist#6463, decision 5): normalize the
+                // collection path the way xmldb:store/create-collection do, so
+                // collection("/db/café-col") resolves the stored key. escape=true encodes
+                // non-ASCII/space and leaves a literal '%' alone, so it is idempotent on an
+                // already-encoded path.
+                uri = XmldbURI.xmldbUriFor(collectionUri.toString(), true);
+            } catch (final URISyntaxException e) {
+                throw new XPathException(this, ErrorCodes.FODC0004, e);
+            }
             try (final Collection coll = context.getBroker().openCollection(uri, Lock.LockMode.READ_LOCK)) {
                 if (coll == null) {
                     if (context.isRaiseErrorOnFailedRetrieval()) {
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/FunDocAvailable.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/FunDocAvailable.java
@@ -96,7 +96,13 @@ public class FunDocAvailable extends Function {
             try {
                 new URI(path);
             } catch (final URISyntaxException e) {
-                throw new XPathException(this, ErrorCodes.FODC0005, e.getMessage(), arg, e);
+                // A bare db-path may contain a character (e.g. a raw space) that is a valid resource
+                // name under the resource-naming contract (eXist-db/exist#6463, decision 3) but not a
+                // valid java.net.URI; defer to DocUtils, which normalizes and resolves it. Any other
+                // malformed URI keeps the spec-mandated FODC0005.
+                if (!DocUtils.isDbPath(path)) {
+                    throw new XPathException(this, ErrorCodes.FODC0005, e.getMessage(), arg, e);
+                }
             }
 
             try {
--- a/exist-core/src/main/java/org/exist/xquery/functions/xmldb/XMLDBAbstractCollectionManipulator.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/xmldb/XMLDBAbstractCollectionManipulator.java
@@ -21,6 +21,7 @@
  */
 package org.exist.xquery.functions.xmldb;
 
+import java.net.URISyntaxException;
 import java.util.Optional;
 import java.util.StringTokenizer;
 
@@ -28,6 +29,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.exist.dom.persistent.NodeProxy;
 import org.exist.xmldb.LocalCollection;
+import org.exist.xmldb.XmldbURI;
 import org.exist.xmldb.txn.bridge.InTxnLocalCollection;
 import org.exist.xquery.BasicFunction;
 import org.exist.xquery.Expression;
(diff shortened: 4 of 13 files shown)


line-o:
I like all of the above - it took me a while to wrap my head around the % encoding approach but the approach seems to be sound.
I do have two questions now that I better understand the approach:
1. What happens when I restore a backup with resources with special characters (some percent encoded, some utf-8 - if that is even possible today)? Is it possible to have clashing names that way. Can it be that hard-coded paths need to be changed in code? If yes we have to document these cases and guard against even theoretical data loss.
2. What if we store files internally by a completely separate encoding - utf8 plus which lets us differentiate between `a%25.xml` and `a%.xml`? I assume there can never be such thing as the issue arises on the input not being able to differentiate. Then I woudl like to have safe paths which guarantee no percent encoding decoding happening – raw access to check and solve issues.

joewiz:

Thanks @line-o — both questions land on exactly the boundary Decision 2 flags (the literal-`%` ambiguity), and they're the right ones to nail before any of this goes near core.

## 1 — Backup / restore

I traced the round-trip. The short version: **a faithful backup of a current database round-trips exactly, and can't introduce a *new* clash** — but the two theoretical-loss cases you're sensing are real, and both trace back to Decision 2.

What the code does today:
- **Backup** (`SystemExport`) writes each resource's *exact stored key* as the `name` attribute (`doc.getFileURI()`, i.e. the already-percent-encoded form), and writes the on-disk file under a separate filesystem-safe encoding.
- **Restore** (`AbstractRestoreHandler` → `uriFromPath` → `XmldbURI.create(name)`) re-creates the resource under that same stored key. For an already-encoded key that's idempotent, so the key comes back byte-for-byte.

Why a normal backup can't *create* a clash: if two names would collapse to one key, they already collided at original store time — so the live database only ever held one of them, and the backup only contains one. The collision happens on input, before backup, not during restore.

The cases worth guarding against — both Decision 2 in a different outfit:
1. **A backup that isn't a faithful export of one current DB** — merged from multiple sources, hand-edited, or produced by an older/differently-encoding version — could carry two entries that normalize to the same key. Restore appends both, so the second `storeResource` silently overwrites the first.
2. **A literal-`%` name** (a `%` not forming a valid `%XX` escape): restore's `XmldbURI.create` can reject it and fall back to re-encoding (`%`→`%25`), silently renaming `a%.xml` → `a%25.xml`, or colliding with an existing `a%25.xml`. This is precisely the input ambiguity Decision 2 documents, now surfacing at restore.

On hard-coded paths: the restore code itself only hard-codes ASCII system paths (`/db/system`, `.../security`, …), which are safe — but auditing the wider codebase for hard-coded user-data paths is a fair ask, and I'll add it to the list.

Concretely, to make this visible and guarded rather than argued: I'll **extend the #6450 conformance corpus to a backup→restore round-trip** — store the awkward-name corpus, export, restore into a fresh database, and assert every name survives byte-for-byte. That empirically answers the literal-`%` edge (instead of me reasoning about it) and becomes the regression net that guards against the theoretical loss you're describing. I'll post the resulting matrix here.

## 2 — A lossless internal encoding + raw access

Your read is exactly right: the ambiguity is an **input** problem. Once two distinct names have collapsed onto one key, nothing downstream can recover them — so the fix has to be at the encoding, before the collapse. A **lossless internal encoding** — escape `%`→`%25` first, so `a%.xml` and `a%25.xml` can never share a key — is precisely Decision 2's Phase 2, and it's the only thing that fully closes this. It's breaking (it changes the stored form of any name containing `%`, so it carries a data migration), which is why it's deferred rather than in this PR — but it is the real fix, and you've described it.

Your **raw / safe-path access** idea I'd like to pull in — it's the missing piece. A low-level accessor that addresses a resource by its **exact stored key, with zero encode/decode** — raw listing (show me the literal keys in a collection), raw read, raw delete — is exactly the tool an admin needs to *see* ambiguous or collided names and fix them surgically. That's the concrete "guard against even theoretical data loss" mechanism: it makes the ambiguous cases inspectable and repairable instead of invisible. I'd propose it as a companion to Phase 2 — and it's independently useful *today* for diagnosing exactly the backup cases in question 1.

Net: normal backups are safe on today's build (verified in the export/restore path); the edges you name are real and both map to Decision 2; I'll add the backup→restore round-trip to #6450 to make them visible and guarded, and I think the raw-access accessor belongs in the plan.

joewiz:

Following up with the empirical result, @line-o. I added a backup→restore round-trip to the conformance net (on the #6450 branch, `feature/naming-conformance-harness`): it stores a corpus of awkward names, snapshots each resource's exact stored key and content, exports a full backup, wipes the collection, restores, and asserts the snapshot is reproduced byte-for-byte — parameterized over `SystemExport`'s direct/non-direct and plain/zip modes.

**Result: the corpus round-trips faithfully in all four modes.** Every name survives, byte-for-byte, under its exact stored key — nothing dropped, renamed, or collided:

```
café.xml          -> caf%C3%A9.xml
with space.xml    -> with%20space.xml
a+b.xml           -> a%2Bb.xml
a@b.xml           -> a%40b.xml
a&b.xml           -> a%26b.xml
report(2024).xml  -> report%282024%29.xml
o'brien.xml       -> o%27brien.xml
Привет.xml        -> %D0%9F%D1%80%D0%B8%D0%B2%D0%B5%D1%82.xml
文書.xml           -> %E6%96%87%E6%9B%B8.xml
a%b.xml           -> a%25b.xml
a%20b.xml         -> a%2520b.xml
résumé.bin        -> r%C3%A9sum%C3%A9.bin   (binary)
```

So the restore path itself is sound for these names, and it's now regression-guarded.

Two things the run surfaced that bear directly on your questions:

1. **On the literal-`%` restore-rename I flagged earlier** — it does *not* trigger on this path. The store path the test drives (eXist's own `encodeXmldbUriFor`) escapes a literal `%`→`%25` first, so `a%b.xml` lands as `a%25b.xml` and round-trips cleanly. That's good news for restore, but it also means this test does **not** exercise the *non*-escaping persistent-store path that produces the Decision 2 collision — that collision happens on the way *in*, before backup, and rides a different encoder. Which is itself the point: this is a second concrete instance of "different code paths encode names differently," exactly the inconsistency the contract is meant to end. Exercising the non-escaping path end-to-end would need an HTTP/WebDAV-driven variant of this test; happy to add it if you'd like that path covered too.

2. **The "not a faithful export of one current DB" case** (a backup merged from sources, hand-edited, or written by an older/differently-encoding version) is still the real theoretical-loss vector, and it's exactly what your **raw / safe-path access** idea is for — being able to list and address resources by their literal stored key, with zero encode/decode, so an admin can *see* two entries that would collapse and resolve them before they do. I still think that accessor belongs in the plan (as a Phase 2 companion, and useful today for diagnosing precisely this).

Net: normal backup/restore is faithful on today's build (now verified and guarded); the loss vectors you named map to the Decision 2 encoding (store-time) and to non-faithful backups (restore-time), both of which the raw accessor would make inspectable. The round-trip test is in the #6450 branch.

joewiz:

Refreshed this ahead of Monday's community call, now that #6463 decision 2 has a direction.

## Decision 2 is settled, and this branch already implements it

@duncdrum on #6463: *"I'm in favour of 1) keep the roundtrip. It's pain once, during a major release, but the right fix longterm."*

Option 1 is **keep `xmldb:store` strictly literal** — which is what this branch already does. So the ratification does not change any code here; it converts the branch from "one of two ways this could go" into "the agreed direction, implemented".

What it *does* settle is the cost: every write client that currently pre-encodes must drop that pre-encoding, and land in lockstep with the merge. That is exactly two clients, both ours.

## Rebased

This branch was three months and 71 commits behind. Rebased onto current `develop` — all 12 commits replayed with no conflicts. **992 tests green**, including every naming-specific class: `ResourceNameCodecTest`, `URIUtilsTest`, `ResourceNamingXmldbRoundTripTest`, `DocTest`, and the full `src/test/xquery` corpus.

## The Windows CI leg found a real hole

`ResourceNamingConformanceTest` runs on all three OSes (it lives in `extensions/webdav`, whose surefire is not gated the way `exist-core`'s is). The `windows-latest` leg reported one regression the other two could not see:

```
asterisk   a*b.xml   FAIL   PUT 500
```

`a*b.xml` stores fine on Linux and macOS and returns HTTP 500 on Windows. Cause: RFC 3986 permits `*` in a path segment as a sub-delimiter, so the codec kept it literal — and Windows forbids `*` in a filename. The other filename-hostile characters (`< > | "`) are not pchar, were already percent-encoded, and pass everywhere.

An earlier note on this branch recorded that `* < > | "` all converge. **That was measured on macOS only and was wrong for `*`.** Now fixed by encoding it to `%2A`, the same reasoning already applied to `:` just below it in the codec: a character the standard permits but the platform cannot store has to be escaped for names containing it to work at all.

Worth noting for its own sake — this is an argument for keeping that test cross-OS. `exist-core`'s tests are ubuntu-only in CI, and this hole is invisible there.

## What remains before this could stop being a draft

1. **The two client migrations.** `existdb-openapi` has a resource-name-encoding branch staged; the eXide side (dropping its per-segment `xmldb:encode` on write) is **not** staged yet. Both need to land together with any merge here, or writes through those clients double-encode.
2. **Decision 1 has never been explicitly ratified.** The whole client-migration cost above rests on it — apps speaking raw decoded UTF-8 is what makes dropping the pre-encoding correct rather than a regression. It is the recommendation in #6463 and is implied by choosing option 1, but nobody has said so on the record. Monday seems like the right moment.
3. **Decision 4** (non-`/db` paths validated and 400 at each surface) is untouched here and still open.
4. A full XQTS run, as the pre-merge gate noted when this was opened.

Staying a draft until at least 1 and 2 are resolved.

joewiz:

Two corrections to my comment earlier today, both worth making before Monday rather than after.

## 1. I overstated where this PR sits

I wrote that decision 2 is settled and "this branch already implements it", which reads as though this is the agreed target. It is not, and #6575 says so explicitly — twice:

> `doc()`/`collection()` gain the exact-first, decode-on-miss pipeline of contract point 4 in the database branch (**subsuming the Decision 5 spike, PR #6508**).

> PR #6508 — `doc()` normalization prototype (Decision 5 spike); **its dispatch analysis (DB branch vs external URIs) carries over** to contract point 4.

The two aim at different end states. This branch keeps percent-encoded stored keys and makes the codec bijective at the store boundary. #6575 removes percent-encoding from the storage layer altogether, so there is no store-boundary encoding left to make bijective — the stored key *is* the name.

@duncdrum's "option 1" on #6463 was a choice *within* #6463's frame, and he endorsed #6575's larger break the same afternoon (#6575, 2026-09-16: *"if we go with option 1 which is my preference break stuff now, to fix it properly"*). So the appetite for breaking changes is real and consistent, but it points at #6575, not at ratifying this branch as-is.

What that makes this PR: a working prototype whose behavior and dispatch analysis feed #6575, and a fallback shape if the larger change proves too big for 7.0. Still useful as an artifact to react to — which is what it was opened as — but not a merge candidate awaiting a green light. I have left it a draft.

Worth noting what *does* carry forward directly: #6450's ratchets are Phase 0 of #6575's plan, and #6451 is load-bearing for its migration — #6575 specifies the legacy decoder must be "a pure, validating percent-decoder. Not `URLDecoder`", which is exactly what `decodeForURI` is. That one went green today.

## 2. I asserted a cause for the Windows failure that I had not established

I wrote that `a*b.xml` fails on Windows because "Windows forbids `*` in a filename". **That explanation does not survive inspection** and I should not have stated it as fact.

eXist keys resources inside `collections.dbx` rather than creating one OS file per resource, so a filesystem naming restriction should not apply directly. The conformance test captures only `PUT 500` with an empty error field, and there is no stack trace anywhere in the CI log, so the actual cause is unknown to me.

What *is* established:

- `a*b.xml` fails a WebDAV PUT on `windows-latest` and succeeds on Linux and macOS.
- Encoding `*` to `%2A` makes it round-trip on all three.
- The characters that already round-tripped (`< > | "`) are not pchar and were already being percent-encoded, which is why only `*` was exposed.

So the fix is right and the reason is not settled.

**If any Windows users could verify this, that would be appreciated.** Specifically: run a WebDAV `PUT` of a resource named `a*b.xml` against a current build on Windows and capture the server-side exception behind the 500. That is the missing piece — whichever direction this work goes, the contract should say *why* a character is unrepresentable, not just that it is.

The commit message has been amended to say this rather than the guess.

Issue #6512, opened 2026-06-22
[BUG] http-client: EXPath conformance gaps in the native rewrite (tracking)


joewiz:
[This issue was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

## Summary

The EXPath HTTP Client was rewritten as a native `java.net.http` + Methanol module (`581e304fce`, 2026-06-15), replacing the old `extensions/expath/` integration that delegated to the EXPath reference library. #6510 surfaced one behavior the rewrite silently dropped — `http:body/@src` (fixed in #6511). This is a tracking issue for the rest, found by auditing the rewrite against the EXPath HTTP Client reference implementation and against BaseX's test suite.

None of these break the common path, and in some areas the eXist module is actually ahead of BaseX's tests (it implements and tests Basic + Digest challenge-response, redirect chains, and gzip, none of which BaseX's suite pins). But the items below are real gaps versus the reference, and each should get a fix plus a regression test.

## Areas (roughly highest-value first)

- [x] **1. `http:body/@method` and serialization for inline bodies.** Inline bodies can only be XML-serialized or sent as text; `@method` (`binary`/`base64`/`hex`/`xml`/`text`/`html`) is ignored, so a binary/base64 inline body cannot be sent (only `@src`, via #6511, delivers raw bytes). The reference honors `@method` (+ `encoding`/`indent`/`omit-xml-declaration`); BaseX tests it directly (`writeBase64`, `writeHex`, `writeText`, `writeMessage`).
- [ ] **2. Real HTML response parsing.** A `text/html` response is XML-parsed and, on failure, returned as a string — so non-well-formed HTML is no longer a navigable document. The reference uses TagSoup; BaseX uses an HTML parser.
- [ ] **3. Byte-safe multipart response parsing.** The response body is split with `String.split("--" + boundary)` over a UTF-8 decode, which corrupts binary parts; per-part headers (beyond Content-Type) are dropped, and nested multipart and preamble/epilogue are not handled. The reference uses a real MIME parser; BaseX tests binary parts, nesting, and preamble/epilogue.
- [ ] **4. Multipart request parts: `@src` and per-part headers.** Request multipart parts read inline content only; a part's `@src` is ignored and part headers other than the derived Content-Type are dropped.
- [ ] **5. Multiple external `$bodies`.** Only the first `$bodies` item is used (and only as a string); the reference consumes one item per body/part.
[...]

PR #6511, opened 2026-06-22
[bugfix] http-client: support the src attribute on http:body (#6510)

joewiz:
[This PR was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

## Summary

Implements the `src` attribute on `http:body` in the native EXPath HTTP Client, fixing #6510. Previously `@src` was parsed away and an empty body was sent with no error — so e.g. a `PUT` with `<http:body media-type='text/plain' src='/db/tmp/test.txt'/>` stored an empty resource.

Closes https://github.com/eXist-db/exist/issues/6510

## What changed

- **`RequestBuilder`** — `http:body` parsing now reads `@src`. When present, the body is the linked resource rather than the element's content. `resolveBodySource(XQueryContext)` resolves `@src` to the request-body bytes, and `applyBody` sends them with `@media-type` as the `Content-Type`.
- **`SendRequestFunction`** — calls `resolveBodySource(context)` between `parse` and `build` (resolution needs the broker from the query context), so the built `HttpRequest` carries the resolved bytes.

## Spec conformance (EXPath HTTP Client 3.1)

> The `src` attribute can be used in a request to set the body content as the content of the linked resource… When this attribute is used, only the `media-type` attribute must also be present, and there can be neither content in the `http:body` element, nor any other attribute, or this is an error [err:HC004].

- `@src` is resolved like `fn:doc`, via `SourceFactory`: a database path (`/db/…` or `xmldb:`), a `file:` URI, or an `http:`/classpath location all work; the resource bytes are sent verbatim. This matches the EXPath reference implementation (BaseX), which writes the `src` resource's raw bytes.
- `err:HC004` (the code was already defined in `HttpClientModule` but never thrown) is raised for the spec's conflict conditions: `@src` combined with body content, or `@src` with any attribute other than `media-type`.
- `media-type` is mandatory on the body, raised as `err:HC005` (a request-validity error — HC004 is reserved for the src/content conflict, matching eXist's own HC004 description and BaseX's "media-type is mandatory" behavior).

## Test plan

- [x] `SendRequestFunctionTest.bodySrcSendsResourceContent` — stores a db resource, sends it via `@src` to an echo endpoint, and asserts the body was transmitted (uses the existing embedded `HttpServer` harness; no external service).
- [x] `SendRequestFunctionTest.bodySrcWithContentThrowsHC004` — `@src` plus inline content raises `err:HC004`.
[...]

PR #6513, opened 2026-06-22
[bugfix] http-client: honor http:body/@method for inline body serialization

joewiz:

## Summary

Honors the `@method` attribute on `http:body` in the native EXPath HTTP Client. Previously `@method` was ignored, so an inline body could only be XML-serialized or sent as text — a binary/base64/hex inline body could not be sent at all (only `http:body/@src`, added in #6511, delivered raw bytes). This is item 1 of the conformance-gap tracking issue #6512, found by auditing the native rewrite against the EXPath reference implementation and BaseX (whose `writeBase64`/`writeHex`/`writeText`/`writeMessage` tests pin this behavior).

Part of #6512.

## What changed

`RequestBuilder` now reads `http:body/@method` (EXPath HTTP Client 3.1) and serializes the inline body accordingly:

- **`binary` / `base64`** — the body's content is base64-decoded and sent as raw bytes.
- **`hex`** — the content is hex-decoded (`HexFormat`) and sent as raw bytes.
- **`text`** — the content's string value is sent (an element child is serialized as its text, not its markup).
- **`xml` / `xhtml` / `html`** (or no `@method`) — unchanged: child elements are XML-serialized as before.

Malformed base64/hex content raises `err:HC005`. The string value of body content is computed by an explicit recursive walk, because the in-memory DOM's `getTextContent` does not recurse into element children of a constructed `http:body` (which is why the previous text path lost element content).

## Spec / reference

- EXPath HTTP Client 3.1 — `http:body/@method` serialization.
- Matches the EXPath reference impl and BaseX, which write the decoded bytes for `binary`/`base64`/`hex` and the string value for `text`.

## Test plan

- [x] `SendRequestFunctionTest.bodyMethodBinarySendsDecodedBytes` — `method='binary'` with `aGVsbG8=` sends `hello`.
- [x] `SendRequestFunctionTest.bodyMethodHexSendsDecodedBytes` — `method='hex'` with `6869` sends `hi`.
- [x] `SendRequestFunctionTest.bodyMethodTextSerializesAsText` — `method='text'` sends an element child's string value, not its markup.
- [x] Full `SendRequestFunctionTest` green (76/76).
- [x] Codacy/PMD clean on the changed files.

## Scope / follow-ups (tracked in #6512)

Out of scope here and left as follow-ups: `@method` on multipart parts, JSON/adaptive serialization, and multiple external `$bodies`.

## Note

This touches `RequestBuilder`'s body handling, the same area as #6511 (`@src`) and #6503 (caching/JMX refactor). The three are independent in intent; whichever merges later will need a small rebase in `RequestBuilder`.

### DIFF
--- a/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
+++ b/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
@@ -361,8 +361,10 @@ public class RequestBuilder {
      * plain string (text), rather than XML-serialized.
      */
     private static boolean isRawContentMethod(final String method) {
-        return "binary".equals(method) || "base64".equals(method)
-                || "hex".equals(method) || "text".equals(method);
+        return switch (method) {
+            case "binary", "base64", "hex", "text" -> true;
+            case null, default -> false;
+        };
     }
 
     /**
--- a/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
+++ b/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
@@ -433,19 +433,20 @@ public class RequestBuilder {
                 throw new XPathException((org.exist.xquery.Expression) null, HttpClientModule.HC005,
                         "http:body with method='" + bodyMethod + "' requires base64-encoded content: " + e.getMessage());
             }
-        }
-        if ("hex".equals(bodyMethod)) {
+        } else if ("hex".equals(bodyMethod)) {
             try {
                 return HttpRequest.BodyPublishers.ofByteArray(HexFormat.of().parseHex(bodyContent.strip()));
             } catch (final IllegalArgumentException e) {
                 throw new XPathException((org.exist.xquery.Expression) null, HttpClientModule.HC005,
                         "http:body with method='hex' requires hexadecimal content: " + e.getMessage());
             }
+        } else {
+            // text/xml/xhtml/html (or no @method): send as a string in the media-type's charset
+            final Charset charset = bodyMediaType != null
+                    ? Charset.forName(ContentTypeHelper.extractCharset(bodyMediaType))
+                    : StandardCharsets.UTF_8;
+            return HttpRequest.BodyPublishers.ofString(bodyContent, charset);
         }
-        final Charset charset = bodyMediaType != null
-                ? Charset.forName(ContentTypeHelper.extractCharset(bodyMediaType))
-                : StandardCharsets.UTF_8;
-        return HttpRequest.BodyPublishers.ofString(bodyContent, charset);
     }
 
     /**
--- a/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
+++ b/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
@@ -426,27 +426,31 @@ public class RequestBuilder {
      * {@code html} (or no @method) are sent as a string in the media-type's charset (default UTF-8).
      */
     private HttpRequest.BodyPublisher bodyPublisherForMethod() throws XPathException {
-        if ("binary".equals(bodyMethod) || "base64".equals(bodyMethod)) {
-            try {
-                return HttpRequest.BodyPublishers.ofByteArray(Base64.getMimeDecoder().decode(bodyContent.strip()));
-            } catch (final IllegalArgumentException e) {
-                throw new XPathException((org.exist.xquery.Expression) null, HttpClientModule.HC005,
-                        "http:body with method='" + bodyMethod + "' requires base64-encoded content: " + e.getMessage());
+        return switch (bodyMethod) {
+            case "binary", "base64" -> {
+                try {
+                    yield HttpRequest.BodyPublishers.ofByteArray(Base64.getMimeDecoder().decode(bodyContent.strip()));
+                } catch (final IllegalArgumentException e) {
+                    throw new XPathException((org.exist.xquery.Expression) null, HttpClientModule.HC005,
+                            "http:body with method='" + bodyMethod + "' requires base64-encoded content: " + e.getMessage());
+                }
             }
-        } else if ("hex".equals(bodyMethod)) {
-            try {
-                return HttpRequest.BodyPublishers.ofByteArray(HexFormat.of().parseHex(bodyContent.strip()));
-            } catch (final IllegalArgumentException e) {
-                throw new XPathException((org.exist.xquery.Expression) null, HttpClientModule.HC005,
-                        "http:body with method='hex' requires hexadecimal content: " + e.getMessage());
+            case "hex" -> {
+                try {
+                    yield HttpRequest.BodyPublishers.ofByteArray(HexFormat.of().parseHex(bodyContent.strip()));
+                } catch (final IllegalArgumentException e) {
+                    throw new XPathException((org.exist.xquery.Expression) null, HttpClientModule.HC005,
+                            "http:body with method='hex' requires hexadecimal content: " + e.getMessage());
+                }
             }
-        } else {
             // text/xml/xhtml/html (or no @method): send as a string in the media-type's charset
-            final Charset charset = bodyMediaType != null
-                    ? Charset.forName(ContentTypeHelper.extractCharset(bodyMediaType))
-                    : StandardCharsets.UTF_8;
-            return HttpRequest.BodyPublishers.ofString(bodyContent, charset);
-        }
+            case null, default -> {
+                final Charset charset = bodyMediaType != null
+                        ? Charset.forName(ContentTypeHelper.extractCharset(bodyMediaType))
+                        : StandardCharsets.UTF_8;
+                yield HttpRequest.BodyPublishers.ofString(bodyContent, charset);
+            }
+        };
     }
 
     /**
--- a/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
+++ b/extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
@@ -54,6 +54,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
+import java.util.HexFormat;
 import java.util.List;
 import java.util.Map;
 
@@ -84,6 +85,7 @@ public class RequestBuilder {
     private String bodyContent;
     private String bodySrc;
     private byte[] bodyResourceBytes;
+    private String bodyMethod;
     private String multipartMediaType;
     private final List<BodyPart> multipartBodies = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


dizzzz:
I'd like to have #6503 pulled in first...

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
```diff
@@ -271,6 +277,35 @@ private void applyBody(final HttpRequest.Builder builder, final String upperMeth
         }
     }
 
+    /**
+     * The http:body @method values whose content is sent as raw bytes (binary/base64/hex) or as a
+     * plain string (text), rather than XML-serialized.
+     */
+    private static boolean isRawContentMethod(final String method) {
+        return "binary".equals(method) || "base64".equals(method)
+                || "hex".equals(method) || "text".equals(method);
```
reinhapa: maybe switch expression would be more easy to read?

extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
```diff
@@ -745,6 +745,56 @@ public void postBodyIsSent() throws XMLDBException {
...
+     */
+    @Test
+    public void bodyMethodBinarySendsDecodedBytes() throws XMLDBException {
+        final ResourceSet result = existEmbeddedServer.executeQuery(
+                HTTP_NS +
+                "let $response := http:send-request(\n" +
+                "  <http:request method='POST' href='" + baseUrl() + "/echo'>\n" +
+                "    <http:body media-type='application/octet-stream' method='binary'>aGVsbG8=</http:body>\n" +
+                "  </http:request>)\n" +
+                "return parse-json($response[2])?body");
```
reinhapa: use text block in combination with `formatted` for better readability

extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
```diff
@@ -745,6 +745,56 @@ public void postBodyIsSent() throws XMLDBException {
...
+     */
+    @Test
+    public void bodyMethodHexSendsDecodedBytes() throws XMLDBException {
+        final ResourceSet result = existEmbeddedServer.executeQuery(
+                HTTP_NS +
+                "let $response := http:send-request(\n" +
+                "  <http:request method='POST' href='" + baseUrl() + "/echo'>\n" +
+                "    <http:body media-type='application/octet-stream' method='hex'>6869</http:body>\n" +
+                "  </http:request>)\n" +
+                "return parse-json($response[2])?body");
```
reinhapa: use text block in combination with `formatted` for better readability

extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
```diff
@@ -745,6 +745,56 @@ public void postBodyIsSent() throws XMLDBException {
...
+     */
+    @Test
+    public void bodyMethodTextSerializesAsText() throws XMLDBException {
+        final ResourceSet result = existEmbeddedServer.executeQuery(
+                HTTP_NS +
+                "let $response := http:send-request(\n" +
+                "  <http:request method='POST' href='" + baseUrl() + "/echo'>\n" +
+                "    <http:body media-type='text/plain' method='text'><a>x</a></http:body>\n" +
+                "  </http:request>)\n" +
+                "return parse-json($response[2])?body");
```
reinhapa: use text block in combination with `formatted` for better readability

reinhapa (changes requested):

joewiz:

Thanks @reinhapa. Addressed in 56dd984815:

- `isRawContentMethod` is now a switch expression. One wrinkle worth flagging: a bare `switch (method)` NPEs on a null selector, and `bodyMethod` is null whenever an `http:body` has no `@method` (the common case), so the naive conversion regressed every body without `@method`. The switch uses a `case null, default -> false` label to keep the implicit null-safety the old `||`-of-`equals` chain had.
- The three new `bodyMethod*` test queries are now text blocks with `formatted(baseUrl())`.

74 → 76 tests green on the module after the fix.

reinhapa (approved):

joewiz:

Rebased onto current `develop` now that #6503 has merged. The conflict was in `RequestBuilder.java`: I kept this PR's `@method` body parsing and `bodyPublisherForMethod()`, and dropped a now-stale `isFollowRedirect()` that the old branch carried — #6503 removed that method and moved `followRedirect` into the request-options config, so reintroducing it would have referenced a deleted field. Module tests stay green (76/76).

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
```diff
@@ -263,19 +301,44 @@ private void applyMultipartBody(final HttpRequest.Builder builder, final String
...
+    private HttpRequest.BodyPublisher bodyPublisherForMethod() throws XPathException {
+        if ("binary".equals(bodyMethod) || "base64".equals(bodyMethod)) {
+            try {
+                return HttpRequest.BodyPublishers.ofByteArray(Base64.getMimeDecoder().decode(bodyContent.strip()));
+            } catch (final IllegalArgumentException e) {
+                throw new XPathException((org.exist.xquery.Expression) null, HttpClientModule.HC005,
+                        "http:body with method='" + bodyMethod + "' requires base64-encoded content: " + e.getMessage());
+            }
+        }
+        if ("hex".equals(bodyMethod)) {
```
dizzzz: add: else?

extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
```diff
@@ -745,6 +745,53 @@ public void postBodyIsSent() throws XMLDBException {
                 "true", result.getResource(0).getContent().toString());
     }
 
+    /**
+     * http:body/@method='binary' decodes the body's base64 text content and sends the raw bytes
```
dizzzz, line 820: I added a mechanism to use WireMock for writing better readable tests.

- query: https://github.com/eXist-db/exist/blob/develop/extensions/modules/http-client/src/test/xquery/http-client-tests.xql
- endpoints: https://github.com/eXist-db/exist/blob/develop/extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/xquery/HttpClientXqueryTests.java

dizzzz (approved):

joewiz:

Thanks @dizzzz — both addressed.

**`else` in `bodyPublisherForMethod()`** — done in 9a2620bb57. The three cases (`binary`/`base64`, `hex`, and the text/xml/xhtml/html-or-no-method default) are now an explicit `if / else if / else` so the fall-through is clearly the mutually-exclusive default, with a comment naming it.

**WireMock test harness** — good pointer, and I'd like to adopt it, but I'd suggest doing it as its own follow-up rather than in this bugfix. A couple of reasons:

- Migrating `SendRequestFunctionTest` from its hand-rolled `com.sun.net.httpserver.HttpServer` endpoints to `HttpClientXqueryTests` + `http-client-tests.xql` is a test-rewrite that touches every test in the class, which would re-open this already-approved PR and bury the one-line `@method` fix under a large test diff.
- The migration is cross-cutting across the whole http-client PR family (this one, #6514, #6515) — they all extend the same `SendRequestFunctionTest` pattern — so it's cleaner to move them together to your harness in one dedicated PR than piecemeal per bugfix.
- A few of these tests assert on wire-level details (raw multipart bytes in #6515, `@method` body serialization here) that I'd want to make sure map cleanly onto WireMock stubs before converting, rather than doing it under time pressure inside a bugfix.

So my proposal: land these bugfixes as-is, then I'll open a follow-up that migrates the http-client Java tests to your WireMock + XSuite harness. Does that work for you? If you'd rather see at least this PR's new tests on WireMock before it merges, say so and I'll fold it in here instead.

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
```diff
@@ -324,19 +324,20 @@ private HttpRequest.BodyPublisher bodyPublisherForMethod() throws XPathException
...
                 throw new XPathException((org.exist.xquery.Expression) null, HttpClientModule.HC005,
                         "http:body with method='hex' requires hexadecimal content: " + e.getMessage());
             }
+        } else {
+            // text/xml/xhtml/html (or no @method): send as a string in the media-type's charset
+            final Charset charset = bodyMediaType != null
+                    ? Charset.forName(ContentTypeHelper.extractCharset(bodyMediaType))
+                    : StandardCharsets.UTF_8;
+            return HttpRequest.BodyPublishers.ofString(bodyContent, charset);
         }
```
reinhapa: with 4 cases, one could use a switch expression here

reinhapa (changes requested):

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
line-o: I agree - this would make it obvious
and I would even propose to add `"test"` `"xml"` `"xhtml"` and `"html|` as explicit cases (+ `default`

extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
line-o: I agree with Joes's assessment that this should be done in a follow up PR

dizzzz (approved):

extensions/modules/http-client/src/main/java/org/exist/xquery/modules/httpclient/RequestBuilder.java
line-o: @joewiz have  you seen these comments?

joewiz: @line-o Thanks for the ping! Digging in now...

joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Thanks @reinhapa and @line-o — you're right, and the earlier commit missed the mark: it converted the neighboring `isRawContentMethod` helper to a switch but left `bodyPublisherForMethod` itself as an if/else chain, which is the method you actually flagged. Fixed now — `bodyPublisherForMethod` is a switch expression: `binary`/`base64` and `hex` get dedicated arms (each decoding to raw bytes), and everything else serializes to a string.

On the explicit `text`/`xml`/`xhtml`/`html` labels — I folded those into the `case null, default` arm rather than spelling them out, and wanted to flag the reasoning in case you'd prefer otherwise. Java won't let case constants share an arm with `default`, so keeping the explicit labels *and* a fallback would mean a `case "text", "xml", "xhtml", "html", null ->` arm plus a `default ->` arm with an identical body — a duplicated branch. Folding them into `case null, default` keeps it a single arm and mirrors the adjacent `isRawContentMethod` switch, with the comment documenting exactly which `@method` values land there. Happy to switch to the explicit labels if you'd rather have them visible in the code despite the duplication — your call.

dizzzz:
@reinhapa @line-o  what about this PR?

reinhapa (approved):

PR #6514, opened 2026-06-22
[bugfix] http-client: parse a non-well-formed HTML response to a document

joewiz:
[This PR was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

## Summary

Parses a non-well-formed `text/html` response into a document node. Previously the native EXPath HTTP Client XML-parsed an HTML response and, on failure, returned the body as a raw string — so ordinary (non-well-formed) HTML stopped being a navigable document, a regression from the old reference-library behavior (which used TagSoup). This is item 2 of the conformance-gap tracking issue #6512.

Part of #6512.

## What changed

`ResponseHandler.parseHtml` now:
1. tries an XML parse (well-formed XHTML resolves directly, as before), then
2. falls back to eXist's **configured HTML-to-XML parser** (NekoHTML, via `HtmlToXmlParser` — the same parser `util:parse-html` uses) to turn tag-soup HTML into a document node, then
3. if no HTML parser is configured, or it cannot parse the input, degrades to returning the body as a string (the prior behavior).

So the change is **strictly additive**: HTML that previously came back as a string and could be parsed now comes back as a document; everything else is unchanged. There is no new compile dependency — the parser is loaded reflectively through `HtmlToXmlParser` (in exist-core); NekoHTML is already on the classpath.

## Test plan

- [x] `SendRequestFunctionTest.malformedHtmlResponseIsParsedToDocument` — a non-well-formed `text/html` response (unclosed `<p>`, void `<br>`, unclosed `<li>`s) is returned as a `document-node()` with a navigable structure (`//body`, two `//li`).
- [x] Existing `htmlResponseIsParsed` (well-formed HTML) still green.
- [x] The test `conf.xml` gains the `html-to-xml` parser block, matching the default eXist configuration.
- [x] Full `SendRequestFunctionTest` green (74/74).
- [x] Codacy/PMD clean on the changed files.

## Note

This touches `ResponseHandler` (the response side), so it does not conflict with #6511 (`@src`) or #6513 (`@method`), which are in `RequestBuilder`. If #6503 (the caching/JMX refactor) also revises `ResponseHandler`, a small rebase may be needed.

PR #6515, opened 2026-06-22
[bugfix] http-client: byte-safe multipart response parsing with nesting

joewiz:

## Summary

Makes multipart response parsing byte-safe and adds nested-multipart support in the native EXPath HTTP Client. Previously a multipart body was split with `String.split("--" + boundary)` over a **UTF-8 decode** of the whole body — so any binary part (a non-UTF-8 byte sequence) was corrupted by the String round-trip — and a part that was itself `multipart/*` was not parsed recursively. This is item 3 of the conformance-gap tracking issue #6512.

Part of #6512.

## What changed

In `ResponseHandler`:
- `splitMultipart`, `extractPartContentType`, and `extractPartBody` now operate on the **raw bytes**: the body is split at the boundary with a byte-level search, each part's header region is decoded as ISO-8859-1 (a total byte→char map) only to read the `Content-Type`, and the part body is sliced as bytes — so a binary part round-trips byte-for-byte. The leading/trailing CRLF that belong to the boundary are stripped; preamble, the closing `--boundary--`, and epilogue are discarded.
- `addBody` recurses into a part whose media type is `multipart/*`, so a **nested** multipart response surfaces its leaf parts.

## Test plan

- [x] `SendRequestFunctionTest.multipartBinaryPartIsByteSafe` — a part of bytes `0xFF 0xFE` round-trips as `xs:base64Binary` (`"//4="`), with the sibling text part intact.
- [x] `SendRequestFunctionTest.multipartNestedIsParsedRecursively` — `outer[ multipart[A,B], C ]` yields the leaf items `A, B, C`.
- [x] Existing multipart tests (`multipartResponseReturnsMultipleItems`, `multipartResponseContainsMultibodyElements`) still green.
- [x] Full `SendRequestFunctionTest` green (75/75).
- [x] Codacy/PMD clean on the changed code (the new test endpoints are registered in a helper so `startHttpServer`'s pre-existing complexity is not increased).

## Scope / follow-up (tracked in #6512)

Reproducing each part's headers (beyond `Content-Type`) in the `http:multipart` descriptor remains a follow-up.

## Note

This touches `ResponseHandler` (the response side), so it does not conflict with #6511 (`@src`) or #6513 (`@method`) in `RequestBuilder`. It does overlap #6514 (HTML response parsing), also in `ResponseHandler` — whichever merges later needs a small rebase.

### DIFF
--- a/extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
+++ b/extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
@@ -274,6 +274,9 @@ public class SendRequestFunctionTest {
             }
         });
 
+        // Multipart byte-safety + nesting endpoints (extracted to keep startHttpServer's complexity down)
+        registerMultipartTestEndpoints(httpServer);
+
         // OPTIONS endpoint
         httpServer.createContext("/options", exchange -> {
             exchange.getResponseHeaders().set("Allow", "GET, POST, OPTIONS");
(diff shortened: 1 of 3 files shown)


extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
```diff
@@ -299,6 +302,45 @@ public static void stopHttpServer() {
...
+    /**
+     * Registers the multipart response endpoints used by the byte-safety and nesting tests, kept out
+     * of startHttpServer so that method's complexity is not inflated.
+     */
+    private static void registerMultipartTestEndpoints(final HttpServer server) {
+        // Multipart response with a binary part (bytes 0xFF 0xFE, base64 "//4=") that is not valid
+        // UTF-8 — corrupted by the old String-based splitter
+        server.createContext("/multipart-binary", exchange -> {
+            final String boundary = "bnd";
+            final java.io.ByteArrayOutputStream baos = new java.io.ByteArrayOutputStream();
```
reinhapa: Do not use canonical class name

reinhapa (changes requested):

joewiz:

Thanks @reinhapa. Done in 8f48967a92 — imported `java.io.ByteArrayOutputStream` and switched to the simple name. Applied to both occurrences in the multipart test endpoints, not just the flagged one. Module tests stay green (75/75).

reinhapa (approved):

extensions/modules/http-client/src/test/java/org/exist/xquery/modules/httpclient/SendRequestFunctionTest.java
```diff
@@ -32,6 +32,7 @@
 import org.xmldb.api.base.ResourceSet;
 import org.xmldb.api.base.XMLDBException;
 
+import java.io.ByteArrayOutputStream;
```
dizzzz: prefer t use apache's BAOS for memory efficiency.

dizzzz:
I was just wondering if we could not use a third party lib for multi-part handling.
for sending a request, Methanol can do it?

dizzzz (approved):

joewiz:

Good question, @dizzzz — and you're right about the send side: the module already does exactly that. Outgoing multipart request bodies are built with Methanol's `MultipartBodyPublisher` in `RequestBuilder.buildMultipartBody()` (the JDK's `java.net.http` client has no multipart support, so Methanol fills that gap). No hand-rolled multipart on the request path.

The catch is that this PR is about the opposite direction — parsing a `multipart/*` **response** we received — and there Methanol doesn't help: its multipart API is publish-only (building bodies to send), and neither Methanol nor `java.net.http` ships a multipart **decoder**. So a `multipart/mixed` response body has to be split by hand regardless, which is what `ResponseHandler` does here.

The one thing that *would* replace the hand-rolled split is a full MIME parser — Apache James mime4j, or Jakarta Mail's `MimeMultipart`. I looked at both, and for this case the tradeoff didn't favor them:

- Each is a sizable new dependency oriented toward email/servlet semantics (mail headers, transfer-encodings, session objects) rather than a raw HTTP `multipart/mixed` byte stream.
- The bug this PR fixes was specifically (a) binary parts being corrupted by a UTF-8 round-trip and (b) nested multipart not being handled. The parser here operates on raw bytes and recurses for nested parts in ~140 lines, so it addresses both directly without dragging in a mail stack.

So my inclination is to keep the hand-rolled decode for this bugfix and not block it on a dependency decision. If the team would still rather standardize the receive side on a library, I'm happy to prototype mime4j separately and compare — but that's a bigger dependency/API discussion and probably its own PR. Does that split make sense to you?

Separately, thanks for the `ByteArrayOutputStream` note on `SendRequestFunctionTest` — done in 128692192f. Both spots now use commons-io's `UnsynchronizedByteArrayOutputStream` (the house convention), matching the earlier switch in this file.

line-o (approved):
LGTM

Issue #5806, opened 2025-07-20
[BUG] OxygenXML not compatible yet with 'develop'


dizzzz:
The oxygenXML editor is not yet compatible with 'develop'.

Two reasons

1) Java 21
Oxygen is yet bundled with Java17; as develop is built with Java21, the eXist-db classes can't be loaded
Work around: remove bundled OpenJDK from oxygen

2) xmldb-api change
the API has been modernized and updated. as a result OxygenXML shows the following error:

```
Exception in thread "AWT-EventQueue-0" java.lang.NoSuchMethodError: 'java.lang.String org.xmldb.api.base.Resource.getResourceType()'
	at ro.sync.db.nxd.exist.k.getTooltip(Unknown Source)
```

this requires an update of the editor.

line-o:
@dizzzz @raducoravu Has this issue been reosolved. Do we need to retest with a current build of `develop`?

raducoravu:
As mentioned above I fixed this issue by changing Oxygen to use the new eXist APIs, so this was not an Exist bug, just a change in the APIs.

line-o:
Thanks for confirming @raducoravu

dizzzz:
```
[ Open ]  -  java.lang.NoSuchMethodError: 'org.xmldb.api.base.Service org.xmldb.api.base.Collection.getService(java.lang.String, java.lang.String)'
java.lang.NoSuchMethodError: 'org.xmldb.api.base.Service org.xmldb.api.base.Collection.getService(java.lang.String, java.lang.String)'
	at ro.sync.db.nxd.exist.k.yi(Unknown Source)
	at ro.sync.exml.editor.x.db$36.b(Unknown Source)
	at ro.sync.db.nxd.b.zj(Unknown Source)
```

it has to do with xmldb-api v2.0 as I see it....

can't yet connect OxygenXML to eXist-db v7

dizzzz:
We introduced the existdb-openapi feature, which might be a clean alternative. @raducoravu  ;

or the GRPC initiative based on xmldb-interface by @reinhapa  - https://github.com/xmldb-org/

dizzzz:
by @joewiz 

# Oxygen + eXist today vs. existdb-openapi

## Why this writeup

[...]

PR #6516, opened 2026-06-22
[bugfix] Update JnlpJarFiles

dizzzz:
Aligned with missing jars in #5806


duncdrum (approved):

duncdrum:
is there a way we can surface problems with Jnlp in our test suite, maybe with the docker container?

dizzzz:
I could think of a 2 step test
- get list of jar files
- attempt to download each 

the problem is... we really dont know what the required set is for XMLrpc connection. Far less than the current set I expect.

reinhapa (approved):

line-o:
Adding the tests can be done in a follow-up, right?

PR #6520, opened 2026-06-24
[feature] Adds new xmldb-api security services

reinhapa:
Adds the following missing exist implementations for the XML:DB API Services:
- `PermissionManagementService` 
- `UserPrincipalLookupService`

### DIFF
--- a/exist-core/src/main/java/org/exist/xmldb/LocalPermissionManagementService.java
+++ b/exist-core/src/main/java/org/exist/xmldb/LocalPermissionManagementService.java
@@ -1,0 +1,150 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xmldb;
+
+import org.exist.security.Account;
+import org.exist.security.Group;
+import org.exist.security.PermissionDeniedException;
+import org.exist.security.Subject;
+import org.exist.storage.BrokerPool;
+import org.xmldb.api.base.Collection;
+import org.xmldb.api.base.Resource;
+import org.xmldb.api.base.XMLDBException;
+import org.xmldb.api.security.AclEntry;
+import org.xmldb.api.security.Attributes;
+import org.xmldb.api.security.GroupPrincipal;
+import org.xmldb.api.security.Permission;
+import org.xmldb.api.security.PermissionManagementService;
+import org.xmldb.api.security.UserPrincipal;
+
+import java.util.List;
+import java.util.Set;
+
+public class LocalPermissionManagementService extends AbstractLocalService implements PermissionManagementService {
+    public LocalPermissionManagementService(Subject user, BrokerPool brokerPool, LocalCollection collection) {
+        super(user, brokerPool, collection);
+    }
+
+    private EXistPermissionManagement permissionManagement() throws XMLDBException {
+       return new EXistPermissionManagement(collection.getService(EXistUserManagementService.class));
+    }
+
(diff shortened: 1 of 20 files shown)


exist-core/src/main/java/org/exist/xmldb/RemoteUserPrincipalLookupService.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public String getName() throws XMLDBException {
+        return "UserPrincipalLookupService";
+    }
+
+    @Override
+    public String getVersion() throws XMLDBException {
+        return "1.0";
```
duncdrum, line 50: What's with all these hard coded version numbers? Idgi

exist-core/src/test/java/org/exist/xmldb/EXistAttributesTest.java
```diff
@@ -0,0 +1,96 @@
...
+import static org.xmldb.api.security.Permission.OWNER_WRITE;
+
+/**
+ * Tests for {@link EXistAttributes}.
+ */
+class EXistAttributesTest {
+
+    @Test
+    void owner() {
+        org.exist.security.Permission permission = createMock(org.exist.security.Permission.class);
```
duncdrum, line 51: Turning the FQN into a named variable could help make this easier to read.

duncdrum (approved):
Very nice. Two questions no blockers

exist-core/src/main/java/org/exist/xmldb/RemoteUserPrincipalLookupService.java
reinhapa: This should represent the implementation version of that service from our side and its currently required by the XML:DB API

exist-core/src/test/java/org/exist/xmldb/EXistAttributesTest.java
reinhapa: In this case I need the FQN to avoid a class name collision with the `org.xmldb.api.security.Permission` import.

duncdrum: I understand that, personally  I find something like 
```java
var secPermission = createMock(org.exist.security.Permission.class);
```

easier to read… as I said just an idea, not a blocker.

exist-core/src/main/java/org/exist/xmldb/RemoteUserPrincipalLookupService.java
duncdrum: is there a canonical location that defines the version? Can it be derived?

exist-core/src/test/java/org/exist/xmldb/EXistAttributesTest.java
reinhapa: Tried that, but did not worked woth EasyMock Framework...

exist-core/src/main/java/org/exist/xmldb/RemoteUserPrincipalLookupService.java
reinhapa: I don't know, but I'm open to use such a source.

exist-core/src/main/java/org/exist/xmldb/LocalPermissionManagementService.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Override
+    public String getName() throws XMLDBException {
+        return "PermissionManagementService";
+    }
+
+    @Override
+    public String getVersion() throws XMLDBException {
+        return "1.0";
```
dizzzz, line 58: Should this follow the xmldb version? V2?

dizzzz (approved):
Lgtm. Nice work

exist-core/src/main/java/org/exist/xmldb/RemoteUserPrincipalLookupService.java
duncdrum: ok I kind of regret asking about this. Basically the pattern here is widely used throughout our code base. So my Claude assisted findings: 
- `DatabaseImpl.getConformanceLevel()` returns `0` with a `//TODO` comment from 2004.
- `xmldb-api` jars contain `Implementation-Version: 2.1.0` but that isn't exposed anywhere
- we seem to have 22 implementation returning `1.0` since 2001 as a dead convention
- the addition of the security manager stuff strikes me as `2.0` in line with `xmldb-api 2.0`  if we want to attach any semantics to this
- I would prefer to set this in central location, so we won't have to dig through multiple files in the future, if it ever becomes meaningful.

exist-core/src/main/java/org/exist/xmldb/LocalPermissionManagementService.java
reinhapa: No, this should represent the implementer version for this aervice.

exist-core/src/main/java/org/exist/xmldb/RemoteUserPrincipalLookupService.java
line-o: At the community call we decided to leave this here for the moment but will look into using exist's version string from now on in all XMLDB api services.

PR #6523, opened 2026-06-25
[ignore] remove unused dependency plex from distribution

dizzzz:
remove plex from distribution. Forgot to remove when removing app assembler code.

xqts runner still depends on it.

### DIFF
--- a/exist-distribution/pom.xml
+++ b/exist-distribution/pom.xml
@@ -335,14 +335,6 @@
             <scope>runtime</scope>
         </dependency>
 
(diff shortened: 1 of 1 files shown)


line-o (approved):
The apassemble-booter is no longer needed as you replaced the entire thing earlier, right @dizzzz ?

dizzzz:
@line-o correct. I guess we could get rid of it for xqts runner as well.

reinhapa (approved):

PR #6525, opened 2026-06-25
[ignore] override jackson-databind dep version

line-o:
Tika parser depends on an older version of jackson-databind 2.21.3 This version has a CVE which is fixed in 2.21.4
But in another subproject we are already depending on 2.22.0 so this seems to be a good common ground.

So in order to better control which version of jackson-databind we are depending on throughout the project it is now
declared in parent POM and then consumed by both extension projects with the `provided` scope.

```
mvn dependency:tree | grep jackson-databind:jar:2.22
```
returns 3 hits

```
mvn dependency:tree | grep jackson-databind:jar:2.21
```

returns no hits

Merging this PR will give us docker images with 0 fixable critical or high severity CVEs (`latest`)

### DIFF
--- a/extensions/contentextraction/pom.xml
+++ b/extensions/contentextraction/pom.xml
@@ -95,7 +95,6 @@
             <artifactId>j8fu</artifactId>
             <scope>test</scope>
         </dependency>
-
     </dependencies>
 
     <build>
--- a/exist-parent/pom.xml
+++ b/exist-parent/pom.xml
@@ -616,6 +616,12 @@
                 <version>${objenesis.version}</version>
                 <scope>test</scope>
             </dependency>
+            <!-- provide project-specific version of jackson-databind version (overrides tika dependency) -->
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
+                <version>2.22.0</version>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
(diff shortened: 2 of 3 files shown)


extensions/contentextraction/pom.xml
dizzzz, line 79: extend exclusions as we dont want to use the one defined by  tika.

line-o: Is this necesary? I thought that the version in the including POM would override already

line-o: @dizzzz when I re-run `mvn dependency:tree` I cannot find a single instance of `jackson-databind:jar:2.21` anymore

`mvn dependency:tree | grep jackson-databind:jar:2.21` returns nothing
`mvn dependency:tree | grep jackson-databind:jar:2.22` returns 5 entries

dizzzz: strange....

line-o: When I do the same operation on current develop HEAD

`mvn dependency:tree | grep jackson-databind:jar:2.21` returns 2 entries
`mvn dependency:tree | grep jackson-databind:jar:2.22` returns 3 entries

dizzzz (approved):

duncdrum:
I m very much in favor of overriding tika, but shouldn't we do this via a version defined in parent  Pom?

line-o:
I was thinking about this as well and we can certainly do that in a follow up. Let's discuss this tomorrow.

line-o:
I will add a dependency to the parent pom

line-o:
@duncdrum done

duncdrum:
@line-o the unit test time-out is flake, but the integration tests failures seems serious. Could you rebase and take another look.

line-o:
Rebased onto current develop and also changed the scope for the dependency in vector extension.
I also tested if the transient dependency in contentextraction has to be declared at all in order to override it.
It turns out that this is not necessary.

Issue #350, opened 2014-10-01
Feature request: configure saxon to use webapp/WEB-INF/catalog.xml


wshager:
Currently it's not possible to set the catalog file for Saxon. Please configure it so it uses the same catalog file as eXist does.

adamretter:
@dizzzz If you want to put the catalog resolve in the database, then could you open a different ticket please?

adamretter:
@wshager Can you test out the branch https://github.com/adamretter/exist/tree/saxon-resolvers and let me know if it works for what you need.

wshager:
I get an error when merging the code with current develop:

javax.xml.transform.TransformerFactoryConfigurationError: Class org.exist.validation.resolver.eXistXMLCatalogResolver is not a SAX2 XMLReader or SAXParserFactory
	at net.sf.saxon.Configuration.makeParser(Configuration.java:2904)
	at net.sf.saxon.Configuration.getStyleParser(Configuration.java:2386)
	at net.sf.saxon.style.StylesheetModule.loadStylesheetModule(StylesheetModule.java:126)
	at net.sf.saxon.style.Compilation.compilePackage(Compilation.java:131)
	at net.sf.saxon.style.Compilation.compileSingletonPackage(Compilation.java:94)
	at net.sf.saxon.s9api.XsltCompiler.compile(XsltCompiler.java:543)
	at net.sf.saxon.jaxp.TemplatesHandlerImpl.getTemplates(TemplatesHandlerImpl.java:104)
	at org.exist.xquery.functions.transform.Transform$CachedStylesheet.getSource(Transform.java:602)
	at org.exist.xquery.functions.transform.Transform$CachedStylesheet.getTemplates(Transform.java:564)
	at org.exist.xquery.functions.transform.Transform$CachedStylesheet.<init>(Transform.java:554)
	at org.exist.xquery.functions.transform.Transform.getSource(Transform.java:514)
	at org.exist.xquery.functions.transform.Transform.createHandler(Transform.java:381)
	at org.exist.xquery.functions.transform.Transform.eval(Transform.java:221)
	at org.exist.xquery.BasicFunction.eval(BasicFunction.java:70)
	at org.exist.xquery.InternalFunctionCall.eval(InternalFunctionCall.java:56)
	at org.exist.xquery.ConditionalExpression.eval(ConditionalExpression.java:100)
	at org.exist.xquery.DebuggableExpression.eval(DebuggableExpression.java:58)
[...]

PR #6528, opened 2026-06-26
Ship versioned native XSD schemas with governance CI

duncdrum:
## Summary

- Adds `mvn validate` checks that canonical config instances (5 files) stay valid against their XSDs, and that any semantic XSD edit is accompanied by a version bump on `xs:schema/@version` (enforced by `.github/workflows/ci-schema-checks.yml`)
- Bumps `xs:schema/@version` on all 9 native XSDs for post-2023 semantic changes
- Adds `schemaVersion` attribute to each canonical config instance, mirroring the XSD version — lets tooling detect schema/instance drift at a glance
- Ships `schema/` in every distribution layout (`tarball`, `zip`, Docker image, IzPack installer) as `$EXIST_HOME/schema/`, a stable location for external tools (eXide, IDE plugins) to resolve native config grammars

## What changed

| Area | Files |
|------|-------|
| Governance CI | `.github/workflows/ci-schema-checks.yml`, `.github/scripts/prepare-governance-context.sh`, `schema/governance.xsl`, `schema/schema-version-type.xsd` |
| XSD version bumps | `schema/*.xsd` (9 files) |
| `schemaVersion` attribute | 5 canonical config instances in `exist-distribution/src/main/config/` |
| `SchemaVersion.java` | Hand-written constants (replaced by codegen in #6530) |
| Distribution packaging | `exist-distribution`, `exist-docker`, `exist-installer` pom changes |
| Documentation | `schema/README.md`, `AGENTS.md` |

## Spec references

- XML Schema 1.1 Part 1 §2.4 (`xs:schema/@version`)
- Governance policy: `schema/README.md`

## Test plan

- [x] `mvn -N xml:transform@schema-governance` passes on this branch (CI green)
- [x] `mvn validate` (canonical instance checks) passes
- [x] `SchemaVersionSyncTest` — constants match XSD `@version` attributes
- [x] `SchemaVersionTest` — constant format/parsing

## Related issues

Closes #6189 — ships native XSD schemas in every distribution layout as `$EXIST_HOME/schema/`.
Closes #6190 — bumps `xs:schema/@version` on all 9 native XSDs; governance CI enforces version bumps on future semantic edits.
Partially addresses #6008 — fixes the `abrev` → `abbrev` typo in `expath-pkg.xsd`; the `namespace` attribute/element discrepancy is not addressed in this PR.

## Notes

This is PR-A of a 5-PR split from #6505. **Merge order: this PR → #6530 → #6531 → #6550 → #6551.**

[...]

PR #6530, opened 2026-06-26
XSD 1.1 store-time document validation + Saxon catalog resolver

### LINKED ISSUE #5541: [BUG] Jaxp does not let you validate XML Schema validity
We have a routine that validates instances against self-generated xsds. It's a rare thing but every once in a while, the xsd itself is in question, so instead of marking every instance as invalid I would like to notify that the schema itself is invalid.

The way to do that seems to be:
`validation:jaxp(xs:anyURI('my.xsd'), true())`
`validation:jaxp(xs:anyURI('my.xsd'), false())`

(both always return false() regardless of schema validity)
or

`validation:pre-parse-grammar(xs:anyURI('my.xsd'))`
(returns “Cannot invoke “Object.toString()” because “value” is null”)

`validation:pre-parse-grammar(doc('my.xsd'))`
(returns “Only XMLSchemas can be preparsed.“)

**Environment**
eXist-db 6.2 - Any hints? Obviously “my.xsd” is something that resolves when I run it

duncdrum:
## Summary

- Configures Saxon (`fn:transform()` / `transform:transform()`) to resolve `xsl:import` and `document()` URIs via the system catalog — fixes issue #350 for the XSLT transform path
- Validates documents against their XSD 1.1 schemas at store time, routing through the XSD 1.1-capable pipeline when the stored document's root namespace resolves to one via the system catalog
- Routes `org.exist.validation.Validator` through the same XSD 1.1 detection
- Auto-generates `SchemaVersion.java` constants from each XSD's `xs:schema/@version` at build time
- Removes a spurious `WARNING: Failed to load catalog` message when no default catalog exists

## What changed

| Area | Files |
|------|-------|
| Saxon catalog resolver | `SaxonConfiguration.java`, `Transform.java`, `URIResolution.java` |
| XSD 1.1 store-time validation | `Xsd11ValidationHelper.java` (new), `Xsd11SchemaCache.java` (new), `Xsd11SchemaDetection.java` (new) |
| `Validator` routing | `Validator.java`, `ValidatorXsd11Test.java` (new) |
| Catalog warning suppression | `ResolverFactory.java` |
| SchemaVersion codegen | `schema/generate-schema-version.xsl`, `exist-core/pom.xml` `schema-version-codegen` execution |
| Tests | `Xsd11StoreTimeValidationTest.java`, `catalog.xql` |

`Xsd11SchemaCache` (Caffeine-backed) and `Xsd11ValidationHelper` were split out of `MutableCollection` per review — that logic is validation/schema infrastructure, not Collection state.

## Related issues

- Closes #5541 (`validation:jaxp()` now reports invalid XSD 1.1 schemas instead of silently misbehaving)
- Closes part of #350 (Saxon catalog resolver for `xsl:import`/`document()`; the `fn:transform()` + `stylesheet-node` case tracked separately, not fixed here)

## Test plan

- [x] `Xsd11StoreTimeValidationTest`, `ValidatorXsd11Test` — XSD 1.1 documents validate correctly at store time and via `Validator`
- [x] `catalog.xql` — `transform:transform()` resolves `xsl:import` via catalog
- [x] Full validation/XSD11 suite (63 tests across 14 classes) — pass after the `Xsd11SchemaCache`/`Xsd11ValidationHelper` extraction
- [x] `license:check`, schema governance, `SchemaVersionSyncTest` — all pass
- [x] CI green (Test & documentation + XQTS)

## Notes

PR-C of a 5-PR split from #6505 (superseded — needs closing separately). #6528 is merged; this branch's diff now shows only its own commits. Next in merge order: #6531 → #6550 → #6551.


exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Transform.java
```diff
@@ -222,24 +222,43 @@ private XsltExecutable compileExecutable(final Options options) throws XPathExce
...
-                            """
-                            transform using a relative href,\s
-                            using option stylesheet-text, but without stylesheet-base-uri""");
-                        throw new TransformerException(resolutionException);
-                    }
-                } catch (final IllegalArgumentException e) {
-                    throw new TransformerException(e);
+        // setResourceResolver() rather than setURIResolver() -- Saxon 12's XsltCompiler still
+        // accepts setURIResolver() (it wraps via ResourceResolverWrappingURIResolver), but going
+        // directly to the ResourceResolver API avoids that extra layer. See #350 /
```
line-o, line 227: Please avoid references to files that are unreachable to us or add plans/catalog-dtd.plan.md to the PR

duncdrum: is already removed as part of cleanup in #6531

line-o (approved):
OK - stale comment is not a blocker

exist-core/src/main/java/org/exist/collections/MutableCollection.java
```diff
@@ -87,6 +103,17 @@ public class MutableCollection implements Collection {
...
+    private static final String XSD_1_1_NS = "http://www.w3.org/XML/XMLSchema/v1.1";
+
+    /**
+     * Caches, per namespace, whether the system catalog's grammar for that namespace needs an
+     * XSD 1.1-capable loader -- see {@link #resolveXsd11SchemaForNamespace(Resolver, String)}.
+     * Namespaces reaching this cache are catalog-registered (a finite, admin-controlled set, never
+     * attacker-influenced), so unlike {@link org.exist.validation.Xsd11SchemaDetection}'s
+     * location-driven cache, no per-Subject scoping/bounding is needed here.
+     */
+    private static final ConcurrentMap<String, Optional<Schema>> XSD11_SCHEMA_BY_NAMESPACE = new ConcurrentHashMap<>();
```
dizzzz: schema caching should/could be done in separate class, using Cafeine library.

exist-core/src/main/java/org/exist/collections/MutableCollection.java
```diff
@@ -117,6 +144,38 @@ public class MutableCollection implements Collection {
...
+     * noNamespaceSchemaLocation} hint that needs XSD 1.1 to load (detected via
+     * {@link org.exist.validation.Xsd11SchemaDetection#detectXsd11ViaSchemaLocation}): unlike
+     * {@link #resolveXsd11SchemaForNamespace(Resolver, String)}, no pre-supplied Source is needed
+     * here, since dynamic discovery can follow the instance's own hint itself, the same way the
+     * default SAX pipeline follows it for XSD 1.0.
+     * <p>
+     * Initialization-on-demand holder idiom: thread-safe with no explicit synchronization, relying
+     * on the JVM's class-initialization guarantees instead of manual double-checked locking.
+     */
+    private static final class Xsd11DynamicDiscoverySchemaHolder {
```
dizzzz: XSD like operations should be in a different class IMO

exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -105,12 +115,23 @@ public class JettyStart implements LifeCycle.Listener {
...
     @GuardedBy("this") private int status = STATUS_STOPPED;
     @GuardedBy("this") private Optional<Thread> shutdownHookThread = Optional.empty();
+    @GuardedBy("this") private Optional<Server> server = Optional.empty();
     @GuardedBy("this") private int primaryPort = 8080;
     @GuardedBy("this") private boolean webAppStartedSuccessfully = false;
     @GuardedBy("this") private String webAppStartupFailureDetail = null;
 
     private final CopyOnWriteArrayList<JettyStartListener> jettyStartListeners = new CopyOnWriteArrayList<>();
 
+    /**
```
dizzzz: should be in a different PR,  not related to XSDs

exist-core/src/test/resources/org/exist/validation/entities/XMLSchema.xsd
dizzzz, line 1: should be retrieved from Norman's schemaresolver?

exist-core/src/test/resources/org/exist/validation/entities/XMLSchema.dtd
dizzzz, line 1: should be resolved from Normans resolver?

dizzzz (changes requested):
added some comments.

exist-core/src/test/resources/org/exist/validation/entities/XMLSchema.xsd
duncdrum: why? our repo has included a copy w3c schemas since forever. W3C is the canonical source. Since their location wasn't wired into the build, not resolving them made sense. I just updated them to 1.1. It's unlikely at this point that these files will ever need to get updated for 1.1, and a 2.0 or 1.x version is also not on the horizon for the foreseeable future.

duncdrum:
**MutableCollection.java (Namespaces / Caffeine / separate class)**
The namespace constants are fixed in `f7f3463460`: `XML_SCHEMA_NS` was dead code (dropped), `XSD_1_1_NS` is now `Namespaces.XSD_1_1_NS`. For the schema cache and XSD operations: agreed they'd sit better in a dedicated class — deferred to a follow-up PR so this chain can land first.

**JettyStart.java:125**
Split to standalone #6572. Only `JettyStart.java` was touched in that commit; it's independent of the XSD work.

exist-core/src/test/resources/org/exist/validation/entities/XMLSchema.dtd
duncdrum: Same reasoning as the XSD — vendored copy that's been in the repo since well before 1.1. The DTD describes the XML Schema language itself; like the XSD, it's unlikely to need updating and the W3C copy is canonical. Updated both to 1.1 to keep them in sync.

dizzzz (changes requested):
I have some open remarks still

line-o:
@duncdrum could you have a look please

exist-core/src/main/java/org/exist/collections/MutableCollection.java
duncdrum: Done rather than deferred — split into two package-private classes in `af2333f396`: `Xsd11SchemaCache` (Caffeine-backed; no eviction policy, since the namespace set is finite and admin-controlled, which was your own point) holds the cache, and `Xsd11ValidationHelper` holds `resolveXsd11SchemaForNamespace`, the dynamic-discovery schema holder, `validateWithXsd11Schema`, and the lexical-handler forwarder. `MutableCollection` is down to a one-line delegating call. All 63 tests in the validation/XSD11 suite pass unchanged.

exist-core/src/main/java/org/exist/collections/MutableCollection.java
duncdrum: Done rather than deferred — split into two package-private classes in `af2333f396`: `Xsd11SchemaCache` (Caffeine-backed; no eviction policy, since the namespace set is finite and admin-controlled, which was your own point) holds the cache, and `Xsd11ValidationHelper` holds `resolveXsd11SchemaForNamespace`, the dynamic-discovery schema holder, `validateWithXsd11Schema`, and the lexical-handler forwarder. `MutableCollection` is down to a one-line delegating call. All 63 tests in the validation/XSD11 suite pass unchanged.

exist-core/src/test/resources/org/exist/validation/entities/XMLSchema.dtd
duncdrum: Same reasoning as the `.xsd` thread — vendored copy of the W3C XML Schema DTD that's been in the repo since before 1.1 support. W3C is the canonical source; we don't want a runtime dependency on an external resolver for something this stable. Updated to 1.1 alongside the XSD to keep them in sync.

duncdrum:
@line-o @dizzzz all comments answered

exist-core/src/main/java/org/exist/collections/Xsd11SchemaCache.java
```diff
@@ -0,0 +1,66 @@
...
+ * fine".
+ * <p>
+ * Namespaces reaching this cache are catalog-registered (a finite, admin-controlled set, never
+ * attacker-influenced), so unlike {@link org.exist.validation.Xsd11SchemaDetection}'s
+ * location-driven cache, no eviction/bounding is needed here -- entries live for the life of the
+ * JVM, cleared only via {@link #clear()}.
+ */
+final class Xsd11SchemaCache {
+
+    private static final Cache<String, Optional<Schema>> CACHE = Caffeine.newBuilder().build();
```
dizzzz, line 45: Great! we might want to define a max time of caching or a max amount of cached items to prevent memory leaks.

dizzzz (approved):
LGTM

dizzzz:
I am still a bit confused/worried why we need to add these XSDs/DTDs ourself, als they are explicitly available in the xmlresolver.org -data package which we already bundle. @duncdrum @line-o

exist-core/src/main/java/org/exist/collections/Xsd11ValidationHelper.java
```diff
@@ -0,0 +1,386 @@
...
+     * case, even with the resolver wired on) -- hence the explicit {@code Source} here.</p>
+     */
+    @Nullable
+    private static Schema resolveXsd11SchemaForNamespace(final Resolver catalogResolver, final String namespace) throws SAXException {
+        final Optional<Schema> cached = Xsd11SchemaCache.get(namespace);
+        if (cached != null) {
+            return cached.orElse(null);
+        }
+
+        Optional<Schema> result = Optional.empty();
```
reinhapa: Why doing the `Optional` dance, when we call `.orElse(null)` in the end any ways? You should not re-assign an optional once set to avoid accidental setting it to `null` somewhere down the path. (Best to use a final assignment)

If it the goal also to store `null` values as in the cache,  we could create simple local variable holding the `Schema` and use `Xsd11SchemaCache.put(namespace, Optional.ofNullable(result))` only when filling the cache at line 153. This would give us less calls to `Optional` and would achieve the same result.

reinhapa (changes requested):

exist-core/src/main/java/org/exist/collections/Xsd11ValidationHelper.java
duncdrum: Good catch, coming up

exist-core/src/main/java/org/exist/collections/Xsd11SchemaCache.java
duncdrum: Good instinct in general, but I don't think it applies here: the key space is the system catalog's namespace list, which is fixed for the life of the running JVM — same as any other config eXist loads at startup/broker-pool-init. A catalog change needs a restart to take effect anyway, and that restart clears this cache along with everything else. So there's no unbounded growth path, just a cache that tops out at "however many namespaces the catalog declares" and stays there. Not adding an eviction policy for a ceiling that's already fixed by config, but happy to revisit if that assumption changes.

duncdrum:
@dizzzz XML Schema language's own XSD/DTD have been vendored in this repo since before XSD 1.1 support existed. We switched our own files to xsd 1.1 but kept shipping xsd 1.0 schema lang files. My guess is that the vendoring was specifically so schema validation doesn't have a hard runtime dependency on an external package. 

Eliminating the vendored copies entirely and resolve them from `xmlresolver.org-data` instead, is a bigger call about a load-bearing dependency for core validation that touches validation during compile of our own files and of validation from inside xquery in the running instance, including catalog resolving from xsl, type checking etc. We can do this in a follow-up. 

For now I would suggest to stick with the old approach, while introducing the new feature, and to get rid of conflicting versions. Switching the source of the files should be a patch release. It's a lot of code without functional changes. In other words quite a bit of work, for very little gain. I m more concerned about the new functionality and features to hit `develop` for testing and feedback.

dizzzz:
Agreed

reinhapa (approved):

PR #6534, opened 2026-06-27
[bugfix] Stop backup prolog stderr spam; fix DeadlockIT CI hang

duncdrum:
### Description:

Two CI reliability fixes:

1. **Filesystem backup resource counting** — `FileSystemBackupDescriptor` no longer parses every payload file as XML when counting resources. Binary entries (e.g. `test.binary`) were triggering Xerces `[Fatal Error] :1:1: Content is not allowed in prolog.` on stderr. Counting now follows the zip-backup path: only `__contents__.xml` descriptors under each collection subtree. SAX errors are routed through a silent `DefaultHandler` on the pooled reader factory (and `XMLReaderPoolTest` updated accordingly). `FileSystemBackupDescriptorTest` covers nested collections and per-descriptor subtrees.

2. **DeadlockIT** — `RemoveDocumentTask` used the wrong document ID formula (`collectionId * random` instead of `collectionId * documentCount + random`), causing an infinite retry loop in `testRemoved` when `collectionId > 0`. Fix the ID calculation, cap remove retries, propagate worker failures, wait for store completion before removals, and retry on benign concurrent-remove races.

### Reference:

CI log noise and Windows integration timeouts observed while running the test suite on GitHub Actions.

### Type of tests:

- New unit tests: `FileSystemBackupDescriptorTest` (3 cases)
- Updated integration test: `DeadlockIT` (Failsafe)
- Updated unit test: `XMLReaderPoolTest`

### Test plan

- [x] `FileSystemBackupDescriptorTest` passes locally
- [x] `DeadlockIT` passes locally (6/6)
- [x] Fork CI green (unit + integration on Ubuntu/macOS/Windows, XQTS)

### DIFF
--- a/exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java
+++ b/exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java
@@ -30,7 +30,6 @@ import org.exist.util.FileUtils;
 import org.xml.sax.SAXException;
 
 import javax.xml.parsers.ParserConfigurationException;
-import java.io.BufferedInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -196,23 +195,26 @@ public class FileSystemBackupDescriptor extends AbstractBackupDescriptor {
         return null;
     }
 
+    /**
+     * Count resources declared in {@link BackupDescriptor#COLLECTION_DESCRIPTOR} files under this
+     * collection's subtree (same approach as {@link ZipArchiveBackupDescriptor}).
+     * {@link org.exist.backup.Restore} uses only the root descriptor's count for the progress total,
+     * so arbitrary subcollections (e.g. /db/apps, /db/foo/stuff) are included without double-counting
+     * the pre-queued /db/system hierarchy.
+     */
     private void countFileEntries(final Path descriptor) {
-
-        // Only count files from top level.
-        if (!descriptor.toString().endsWith("/db/" + COLLECTION_DESCRIPTOR)) {
+        final Path collectionDir = descriptor.getParent();
+        if (collectionDir == null) {
             return;
         }
-
         try {
             final DescriptorResourceCounter descriptorResourceCounter = new DescriptorResourceCounter();
-
-            try (final Stream<Path> walk = Files.walk(descriptor.getParent());
-                final Stream<Path> ds = walk
-                        .filter(f -> !Files.isDirectory(f))
-                        .filter(f -> !COLLECTION_DESCRIPTOR.equals(f.getFileName().toString()))) {
-
-                for (final Path d : ds.collect(Collectors.toList())) {
-                    try (final InputStream is = new BufferedInputStream(Files.newInputStream(d))) {
+            try (final Stream<Path> walk = Files.walk(collectionDir)) {
+                for (final Path contentsFile : walk
+                        .filter(p -> !Files.isDirectory(p))
+                        .filter(p -> COLLECTION_DESCRIPTOR.equals(p.getFileName().toString()))
+                        .toList()) {
+                    try (final InputStream is = Files.newInputStream(contentsFile)) {
                         numberOfFiles += descriptorResourceCounter.count(is);
                     }
                 }
--- a/exist-core/src/test/java/org/exist/backup/FileSystemBackupDescriptorTest.java
+++ b/exist-core/src/test/java/org/exist/backup/FileSystemBackupDescriptorTest.java
@@ -1,0 +1,159 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.backup;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class FileSystemBackupDescriptorTest {
+
+    private static final String CONTENTS_XML = """
+            <?xml version="1.0" encoding="UTF-8"?>
+            <exist:collection xmlns:exist="http://exist.sourceforge.net/NS/exist">
+                <exist:resource name="doc1.xml" type="XMLResource" filename="doc1.xml"/>
+                <exist:resource name="test.binary" type="BinaryResource" filename="test.binary"/>
+            </exist:collection>
+            """;
+
+    private static final String NESTED_CONTENTS_XML = """
+            <?xml version="1.0" encoding="UTF-8"?>
+            <exist:collection xmlns:exist="http://exist.sourceforge.net/NS/exist">
+                <exist:resource name="nested.xml" type="XMLResource" filename="nested.xml"/>
+            </exist:collection>
+            """;
+
+    @Test
+    void countsResourcesFromContentsXmlOnly(@TempDir final Path tempDir) throws Exception {
+        final Path backupRoot = tempDir.resolve("backup");
+        final Path dbDir = backupRoot.resolve("db");
+        Files.createDirectories(dbDir);
+
+        Files.writeString(dbDir.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR), CONTENTS_XML, UTF_8);
+        Files.writeString(dbDir.resolve("doc1.xml"), "<test/>", UTF_8);
+        Files.write(dbDir.resolve("test.binary"), "test".getBytes(UTF_8));
+
+        final FileSystemBackupDescriptor descriptor = new FileSystemBackupDescriptor(
+                backupRoot,
+                dbDir.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR));
+
+        assertEquals(2, descriptor.getNumberOfFiles());
+    }
+
+    @Test
+    void rootDescriptorCountsAllDescendantResources(@TempDir final Path tempDir) throws Exception {
+        final Path backupRoot = tempDir.resolve("backup");
+        final Path dbDir = backupRoot.resolve("db");
+        final Path nestedDir = dbDir.resolve("nested");
+        Files.createDirectories(nestedDir);
+
+        Files.writeString(dbDir.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR), CONTENTS_XML, UTF_8);
+        Files.writeString(nestedDir.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR), NESTED_CONTENTS_XML, UTF_8);
+        Files.writeString(dbDir.resolve("doc1.xml"), "<test/>", UTF_8);
+        Files.write(dbDir.resolve("test.binary"), "test".getBytes(UTF_8));
+        Files.writeString(nestedDir.resolve("nested.xml"), "<nested/>", UTF_8);
+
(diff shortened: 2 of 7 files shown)


joewiz:

High-effort review; findings independently verified, and I empirically confirmed the headline one with a probe test. The stderr-spam fix and the `DeadlockIT` hang fix both look sound — routing the SAX errors through a silent handler is the right call. One correctness issue with the restore progress counting, plus a related performance note.

## 1. Restore progress total is double-counted — the progress bar never reaches 100%

`exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java`

The exact mechanism: develop's `countFileEntries` opened with `if (!descriptor.toString().endsWith("/db/" + COLLECTION_DESCRIPTOR)) return;`, which zeroed out every *child* descriptor's count — only the root `/db` descriptor produced a non-zero number. This PR removes that guard, so now **every** descriptor recursively walks and counts its whole subtree. But `Restore.restore` builds `totalNrOfFiles` by summing `getNumberOfFiles()` over the root `/db` descriptor **and** the `system` / `security` / `exist/groups` child descriptors that `getBackupDescriptors()` pushes separately (so `/db/system` is restored before the rest of `/db`). Since the root `/db` count now already includes the entire `/db/system` subtree, that subtree gets counted again by each separately-pushed descriptor — 2× for `system`, 3× for `security`, 4× for `exist/groups`. `listener.started(totalNrOfFiles)` then gets an inflated total, so the restore progress indicator stalls below 100% ("restored 4000 of 12000") and never completes, even though the restore itself finishes fine. This is new for directory (non-zip) restores.

I confirmed it with a probe: a `/db` backup declaring 3 resources with a `/db/system` declaring 5 → the root descriptor's `getNumberOfFiles()` returns **8** (it includes `/db/system`), the `system` descriptor returns **5**, so `Restore`'s sum is **13** against **8** real resources.

Fix: count each subtree once — either keep the recursive root walk and **don't** add the separately-pushed child descriptors into `totalNrOfFiles` (push them only for ordering), or keep a non-recursive per-descriptor count so the existing sum stays correct.

Here's a regression test you can drop in alongside the fix — it **fails on the current code** (13 ≠ 8) and passes once the double-count is gone:

<details><summary>FileSystemBackupDescriptor restore-count regression test</summary>

```java
package org.exist.backup;

import org.junit.Rule;
import org.junit.Test;
import org.junit.rules.TemporaryFolder;

import java.io.IOException;
import java.nio.charset.StandardCharsets;
import java.nio.file.Files;
import java.nio.file.Path;

import static org.junit.Assert.assertEquals;

public class FileSystemBackupDescriptorCountTest {

    private static final String EXIST_NS = "http://exist.sourceforge.net/NS/exist";

    @Rule
    public final TemporaryFolder tmp = new TemporaryFolder();

    @Test
    public void restoreProgressTotalMatchesRealResourceCount() throws Exception {
        final Path root = tmp.newFolder("backup").toPath();
        final Path db = Files.createDirectories(root.resolve("db"));
        final Path system = Files.createDirectories(db.resolve("system"));
        writeContents(db, 3);      // /db declares 3 resources
        writeContents(system, 5);  // /db/system declares 5 resources

        // Mirror Restore.restore's summation: the root /db descriptor plus the separately-pushed
        // /db/system descriptor. With the double-count this is 8 + 5 = 13; correct is 8.
        final long total =
                new FileSystemBackupDescriptor(root, db.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR)).getNumberOfFiles()
              + new FileSystemBackupDescriptor(root, system.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR)).getNumberOfFiles();

        assertEquals("restore progress total must equal the real resource count "
                + "(the /db/system subtree must not be counted twice)", 8, total);
    }

    private void writeContents(final Path collectionDir, final int resourceCount) throws IOException {
        final StringBuilder sb = new StringBuilder();
        sb.append("<collection xmlns=\"").append(EXIST_NS).append("\" name=\"/x\">");
        for (int i = 0; i < resourceCount; i++) {
            sb.append("<resource name=\"r").append(i).append(".xml\"/>");
        }
        sb.append("</collection>");
        Files.writeString(collectionDir.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR),
                sb.toString(), StandardCharsets.UTF_8);
    }
}
```
</details>

## 2. Same root cause is an O(N·depth) cost at restore time

Each `FileSystemBackupDescriptor` is constructed once per collection during restore (`getChildBackupDescriptor` in `restoreSubCollectionEntry`), and the new constructor recursively walks + SAX-parses every `__contents__.xml` in its subtree. So a contents file at depth *d* is parsed once for each of its *d* ancestor descriptors — O(N) becomes O(N·depth) walks/parses on a large or deep backup, and the result is discarded for every collection that isn't skipped. Resolving the double-count above (count once at the root) also removes most of this.

## 3. Minor

In the subtree filter, testing `COLLECTION_DESCRIPTOR.equals(p.getFileName().toString())` before `Files.isDirectory(p)` would skip one `isDirectory` stat per data file.

duncdrum:
Thx Joe while this wasn't in the intended scope here, I have long been puzzled by our seemingly random restore calculation. Going to fix this, good catch.

duncdrum:
Thanks Joe — good catch on all three. Addressed in f1eb17e.

**#1 (double-count) + #2 (O(N·depth)):** `countFileEntries` is now non-recursive. Each descriptor parses only its own `__contents__.xml` and reports only its own resource count. `Restore.restore()`'s summation across root + separately-pushed child descriptors now gives the correct total, and the O(N·depth) SAX-parse cost from the subtree walk is gone entirely.

**#3 (filter order):** Moot — there's no `Files.walk` filter left to reorder.

Test updates: `countsNestedCollectionDescriptorsInSubtree` updated to assert the correct non-recursive count (2, not 3), and your regression test (`restoreProgressTotalMatchesActualResourceCount`, 3+5=8) added.

[This response was prompted by Duncan, drafted by Claude Code, and reviewed by Duncan.]

line-o (approved):
From what I can see this is good to go in.
The only question mark I have is if we have a test that ensures nested directories return the expected grand total somewhere.
The test that was added counts the sum by adding the amount of each subdirectory by hand.
The root collection could have additional resources that would not be counted that way.

exist-core/src/test/java/org/exist/backup/FileSystemBackupDescriptorTest.java
```diff
@@ -0,0 +1,142 @@
...
+                    <exist:resource name="s1.xml" type="XMLResource" filename="s1.xml"/>
+                    <exist:resource name="s2.xml" type="XMLResource" filename="s2.xml"/>
+                    <exist:resource name="s3.xml" type="XMLResource" filename="s3.xml"/>
+                    <exist:resource name="s4.xml" type="XMLResource" filename="s4.xml"/>
+                </exist:collection>
+                """;
+        Files.writeString(dbDir.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR), dbContents, UTF_8);
+        Files.writeString(systemDir.resolve(BackupDescriptor.COLLECTION_DESCRIPTOR), systemContents, UTF_8);
+
+        final long total =
```
line-o: I would assume that there is a method we can call which reports the total without us having to sum up each collection.

duncdrum: The manual summation is intentional — it mirrors Restore.restore().  There's no single-call total because adding one to BackupDescriptor would require a recursive subtree walk.  The summation belongs in Restore, . The test replicates that loop to prove it holds.

duncdrum:
@line-o that's what `restoreProgressTotalMatchesActualResourceCount` does, but you do have a point about arbitrary collections under `db/`

exist-core/src/test/java/org/exist/backup/FileSystemBackupDescriptorTest.java
line-o: But I can see that it changed meanwhile

line-o:
@duncdrum Can we squash commits so that the back and forth is not happening?

exist-core/src/test/java/org/exist/storage/lock/DeadlockIT.java
```diff
@@ -217,19 +263,23 @@ public void runTasks() {
...
+        private final CountDownLatch storeComplete;
 
-		public StoreTask(final String id, final int collectionCount, final int docCount) {
+		public StoreTask(final String id, final int collectionCount, final int docCount,
+                final CountDownLatch storeComplete) {
 			this.id = id;
 			this.collectionCount = collectionCount;
 			this.docCount = docCount;
+            this.storeComplete = storeComplete;
 		}
```
reinhapa: Could be  potentially be converted to a `record`

reinhapa (approved):

duncdrum:
@line-o squashed @reinhapa good idea if this is what you had in mind.

line-o (approved):

exist-core/src/test/java/org/exist/storage/lock/DeadlockIT.java
line-o: Do you mean the entire StoreTask could be expressed as a Java Record?

exist-core/src/test/java/org/exist/storage/lock/DeadlockIT.java
```diff
@@ -217,20 +263,11 @@ public void runTasks() {
...
-		private final String id;
-		private final int docCount;
-		private final int collectionCount;
-
-		public StoreTask(final String id, final int collectionCount, final int docCount) {
-			this.id = id;
-			this.collectionCount = collectionCount;
-			this.docCount = docCount;
-		}
+	private record StoreTask(int collectionCount, int docCount, CountDownLatch storeComplete,
```
line-o, line 269: Ah yes!

exist-core/src/test/java/org/exist/storage/lock/DeadlockIT.java
line-o: This is what @duncdrum did. Assuming this is resolved now.

exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java
```diff
@@ -196,23 +195,26 @@ public Path getRepoBackup() throws IOException {
...
-
-            try (final Stream<Path> walk = Files.walk(descriptor.getParent());
-                final Stream<Path> ds = walk
-                        .filter(f -> !Files.isDirectory(f))
-                        .filter(f -> !COLLECTION_DESCRIPTOR.equals(f.getFileName().toString()))) {
-
-                for (final Path d : ds.collect(Collectors.toList())) {
-                    try (final InputStream is = new BufferedInputStream(Files.newInputStream(d))) {
+            try (final Stream<Path> walk = Files.walk(collectionDir)) {
+                for (final Path contentsFile : walk
```
dizzzz, line 213: though it is not in the original code, the Stream created by walk on line 213 should be in a try()-resources too

dizzzz: to prevent resource leaks

dizzzz:
How does the code behave if the __contents__.xml file reports more or less files than actually present in the ZIP file?

dizzzz (changes requested):

exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java
duncdrum: @dizzzz the `walk` is in a `try` in L212 the filter list and final `toList()` don't create a new closable stream, is this the right location? not sure I'm following. So before I commit a no-op in the wrong location, please confirm if we are talking about the same thing.

duncdrum:
> How does the code behave if the **contents**.xml file reports more or less files than actually present in the ZIP file?

Same as today, the count and restore are both derived from the same source `__contents__.xml` so synchronous by construction. The change here only aligns the directory path to match what ZIP already does. 

So for a descriptor-vs-payload mismatch:
- Descriptor declares more than exist on disk: the missing resource is attempted and logged as a restore error, but it was already counted, so the bar still reaches 100% — no stall.
- Descriptor declares fewer than exist on disk: the extra file isn't referenced, so it's neither restored nor counted — silently skipped, progress stays consistent.

Descriptor vs Payload validation would be a good feature request for both modes, but out of scope for this PR.

exist-core/src/main/java/org/exist/backup/FileSystemBackupDescriptor.java
line-o: I also do not see an issue with a resource leak. What I do see is an opportunity to refactor the code to a less clever version

Split into declaration of content file list and then iterate over it. Should clarify intent and ease review and understanding next week.

line-o:
```java
final List<Object> files = walk.filter(p -> !Files.isDirectory(p))
                        .filter(p -> COLLECTION_DESCRIPTOR.equals(p.getFileName().toString()))
                        .toList()
for (final Path contentsFile : files) {

dizzzz: ok

dizzzz (approved):

adamretter:
This is not a very safe way to address this. It hides problems and will make it harder to find the real issues in furure

PR #6535, opened 2026-06-27
[bugfix] Silence Log4j StatusLogger noise in Maven test forks

duncdrum:
### Description:

Surefire inherited `log4j.configurationFile=${project.build.testOutputDirectory}/log4j2.xml` from the parent POM. Many modules run tests without copying `log4j2.xml` into `target/test-classes`, so Log4j 2 logged this on stderr for every fork:

`ERROR Reconfiguration failed: No configuration found for '…' at 'null' in 'null'`

This adds a shared `exist-parent/log4j2-test.xml` and points Surefire/Failsafe at it via a new `exist.log4j.testConfigurationFile` property. Module-specific overrides (exist-core, vector, webdav, etc.) now use the same property.

### Reference:

Observed while investigating CI test log noise during DeadlockIT/prolog work.

### Type of tests:

- Maven build / Surefire configuration only (no new test classes)
- Verified locally: `mvn test -pl extensions/modules/http-client -Dsurefire.useFile=false` no longer emits `Reconfiguration failed`

### Test plan

- [x] `extensions/modules/http-client` tests pass without Log4j reconfiguration errors
- [ ] Upstream CI (unit + integration + Codacy)


line-o (approved):
I think this is a good approach and assume that a later step will generate the log4j2-test.xml from a SSoT

reinhapa (approved):

PR #6543, opened 2026-07-03
[ci] Publish container attestations via staging-then-promote

### LINKED ISSUE #6539: Docker images still missing SBOM and provenance
Unfortunately the SBOM and provenance attestations are built but immediately lost because we are building with `buildx ... --load` and local container registries cannot handle manifest lists.

We need to switch to `buildx ... --push` in order to upload them immediately.

The thing is that in CI a bats tests are run against the build images _before_ they are pushed.

Several ways forward are possible and I did venture in one direction already:

- build with --load
- then test
- build with --push on success 

We could just ditch the test if that is OK but I didn't want to explore that without the opinion of others.

### LINKED ISSUE #6540: Docker metadata should use OCI annotations
At the moment we are still adding metadata or labels using deprecated semantics

The current schema for these container labels is defined here https://github.com/opencontainers/image-spec/blob/main/annotations.md

In our dockerfiles I would want to replace the labels with

```
# Build-time metadata as defined by OCI
# https://github.com/opencontainers/image-spec/blob/main/annotations.md
LABEL org.opencontainers.image.created="${maven.build.timestamp}" \
      org.opencontainers.image.description="${project.description}" \
      org.opencontainers.image.revision=${git.commit.id} \
      org.opencontainers.image.version=${project.version} \
      org.opencontainers.image.url="${docker.url}" \
      org.opencontainers.image.source="${docker.source}" \
      org.opencontainers.image.vendor="${docker.vendor}" \
      org.opencontainers.image.licenses="${docker.licenses}" \
      org.opencontainers.image.title="${docker.title}" \
      org.opencontainers.image.base.name=gcr.io/distroless/java21-debian13:latest
```

There are some important additions here

- `org.opencontainers.image.base.name` will inform consumers (and Dockerhub) which base image we are using
- `org.opencontainers.image.source` is an invalid URL at the moment (`scm:git:`...)
- `org.opencontainers.image.vendor` could be switched for `org.opencontainers.image.authors`
- `org.opencontainers.image.licenses` was missing entirely
- `org.opencontainers.image.revision` should contain the full, unabbreviated commit hash
- `org.opencontainers.image.version` I don't see a reason why we cannot populate version all the time - am I missing something?

line-o:
Docker SBOM and provenance attestations were built but immediately discarded: buildx `--load` cannot store a manifest list in the local image store, so the attested multi-arch image never left the runner and the tags on Docker Hub carried no attestations.

Switch the container pipeline to **build → push staging → test → promote**, and build/push the multi-arch image with the first-party `docker/build-push-action` so the attestations are actually produced and preserved:

1. Maven `package` assembles the distribution + resolves the (filtered) Dockerfiles into a versioned build context — it no longer builds or pushes the published image.
2. `docker/build-push-action` builds the multi-arch (amd64 + arm64) image with `provenance: mode=max` + `sbom: true` and `--push`es it to a `<channel>-staging` tag (`latest-staging` / `release-staging`), plus a matching `<channel>-debug-staging`.
3. The bats suite runs against the amd64 variant pulled back from the staging tag.
4. On success, `docker buildx imagetools create` copies the tested OCI index — both arches **and** the SBOM/provenance attestation manifests — to the public tags server-side, no rebuild.
5. A final gate re-inspects each promoted tag and fails the run unless it carries both arches and at least one attestation manifest.

Nothing un-suffixed reaches Docker Hub until the tests pass; a failed run leaves the `-staging` tag in place as the inspectable "last CI build".

## Why `docker/build-push-action`

The previous approach drove buildx through the fabric8 `docker-maven-plugin`. Moving the build/push to `docker/build-push-action` is the idiomatic GitHub Actions path documented for [build attestations](https://docs.docker.com/build/ci/github-actions/attestations/): it produces `mode=max` provenance and SBOM directly, adds `type=gha` build caching for faster multi-arch rebuilds, and keeps the OCI image labels owned by Maven resource filtering (no duplicate `metadata-action` labels). Maven still owns context assembly and Dockerfile filtering, so `build-push-action` only builds and pushes.

## What changed

- **`.github/workflows/ci-container.yml`**
  - Resolve the publish target (channel/version) once; PRs and forks keep the local amd64 load-and-test path — only upstream `develop` (→ `latest`) and `eXist-*` tags (→ `release`) push and promote.
  - Replace the fabric8 `docker:build docker:push` step with two `docker/build-push-action` steps (main + DEBUG), each with `provenance: mode=max`, `sbom: true`, and `type=gha` cache. Docker Hub login moves ahead of the build because the action pushes directly.
  - Promote the tested staging index to the public channel tag **and** the exact project-version tag (`7.0.0-SNAPSHOT` on `develop`, the concrete version on `eXist-*` tags — the SNAPSHOT tag is intentionally mutable, overwritten each build).
  - Verify the promoted manifest carries both arches and at least one attestation manifest, failing the run otherwise.
  - **Drop emulated-arch testing**: running the arm64 image under QEMU produces meaningless results (and frequently crashes). CI exercises the native amd64 arch only; arm64 is validated locally on Apple silicon by the same commands. (Thanks @duncdrum.)
  - Remove the containerd image-store hack; serialize runs per ref (concurrency group) so concurrent `develop` pushes cannot race on the shared staging tag.

- **`exist-docker/pom.xml`**
  - Drive the pushed tag of the release images from `${docker.tag}` / `${docker.debug.tag}` instead of `%v` + `<tags>`, so the staging build publishes only the staging tags; minting the version tag moves to the promote step.
  - Add OCI image-metadata properties (`source`, `url`, `vendor`, `licenses`, `title`) consumed by the Dockerfile labels.
  - Add `docker.base.image` / `docker.base.image.debug` as the single source of truth for each image's base, so `FROM` and the `base.name` label can never drift. (Thanks @duncdrum.)

- **`exist-docker/src/main/resources-filtered/Dockerfile`** — replace the deprecated `org.label-schema.*` labels with `org.opencontainers.image.*` annotations (`created`, `description`, `revision`, `version`, `url`, `source`, `vendor`, `licenses`, `title`, `base.name`); `FROM` and `base.name` both read `${docker.base.image}`.

- **`exist-docker/src/main/resources-filtered/Dockerfile-DEBUG`** — same OCI migration, so the DEBUG image also carries `org.opencontainers.image.base.name` (previously it had no such label and was still on `label-schema.*`); `FROM` and `base.name` read `${docker.base.image.debug}`.

- **`exist-docker/src/test/bats/02-config-spec.bats`** — reference the canonical local tag `existdb/existdb:ci` instead of the orphaned `existdb/exist-ci-build:latest`, decoupling the suite from the registry tag under test.

## Test plan

- [x] PR/fork path builds and tests the local amd64 image (no push).
- [x] Maven resource filtering verified to substitute `${docker.base.image}` / `${docker.base.image.debug}` in both `FROM` and `base.name` for both Dockerfiles.
- [x] `actionlint` clean on the workflow (pre-existing shellcheck notes in the untouched publish-target script only).
- [ ] Upstream `develop` run pushes `latest-staging`, tests, promotes `latest` + `7.0.0-SNAPSHOT`, and the verify gate reports both arches + attestation manifests.
- [ ] Confirm on a published image: `docker buildx imagetools inspect existdb/existdb:latest --format '{{ json .Provenance }}'` and `… .SBOM`.

Closes #6539
Closes #6540


duncdrum (commented):
Good to see the containerd hack being removed. 

The arm test didn't run. TBH that's a good thing imv. Testing the other arch requires emulation, and frequently leads to crashing containers, the test results are meaningless. In the old set up we only test the native arch on ci, but the same commands run locally on arm will result in arm being tested. So far this has never been a problem, I would stick with it. 

I m not sure if this treats `:7.0.0-SNAPSHOT` as immutable tag, that might lead to problems. Which tags are supposed to be switched to immutable in the PR?

line-o:
Yeah, I see now that the description is misleading. None of the tags are immutable even after this PR is merged.
The approach of publishing a staging image, testing it and only then tagging it without -staging allows for immutable tags in the same way as the old approach did. By making sure the container is runnable.

line-o:
I am not sure if I understand your point on testing of images for different architectures.

Are you happy with the setup as it is or do you want to see changes?

line-o:
@duncdrum ready for you to have another look - the bats tests of the arm image were removed.

.github/workflows/ci-container.yml
```diff
@@ -34,54 +41,120 @@ jobs:
...
+            -Ddocker.debug.tag=${{ steps.meta.outputs.channel }}-debug-staging \
+            -Ddocker.sbom=true -Ddocker.provenance=true \
+            -Ddocker.username=$DOCKER_USERNAME -Ddocker.password=$DOCKER_PASSWORD \
+            -Ddocker.filter=exist,exist-debug \
+            docker:build docker:push
+
+      - name: Pull & tag staging image for tests
+        if: steps.meta.outputs.publish == 'true'
+        run: |
+          docker pull --platform linux/amd64 existdb/existdb:${{ steps.meta.outputs.channel }}-staging
```
duncdrum: `--platform linux/amd64` seems superfluous, pull will automatically match the platform.

exist-docker/src/main/resources-filtered/Dockerfile
```diff
@@ -30,16 +30,18 @@ COPY lib /exist/lib
...
+LABEL org.opencontainers.image.created="${maven.build.timestamp}" \
+      org.opencontainers.image.description="${project.description}" \
+      org.opencontainers.image.revision=${git.commit.id} \
+      org.opencontainers.image.version=${project.version} \
+      org.opencontainers.image.url="${docker.url}" \
+      org.opencontainers.image.source="${docker.source}" \
+      org.opencontainers.image.vendor="${docker.vendor}" \
+      org.opencontainers.image.licenses="${docker.licenses}" \
+      org.opencontainers.image.title="${docker.title}" \
+      org.opencontainers.image.base.name=gcr.io/distroless/java21-debian13:latest
```
duncdrum: do we need to hardcode the base image here?

duncdrum (approved):
LGTM, two comments inline

exist-docker/src/main/resources-filtered/Dockerfile
line-o: No, I meant it to follow the actual base image but never got to the point to have it in a variable in both places. Are there downsides to this? Will we still get dependabot alerts and similar?

.github/workflows/ci-container.yml
line-o: happy to drop it - leftover from the original workflow

line-o:
I am actually following up with another improvement so hang tight

line-o:
@duncdrum there is a significant change the images built by `mvn package -Pdocker` are not the ones that are being pushed. The maven command rather sets the stage that lets the later steps build a container from.

duncdrum:
@line-o It seems to be building [on CI](https://github.com/eXist-db/exist/actions/runs/28939392055/job/85857576688?pr=6543#step:11:2801). Would you have expected the skipped jobs to run, or what is the problem you are seeing. I can hop on a call, might be easier.

line-o:
@duncdrum thanks for the discussion earlier - I went through all the nitty gritty details and changed what made sense. This should now go in as such.

Investigations that concluded "leave as-is" (no change warranted):
- skip-build-dist-archives — still load-bearing; skips the .zip/.tar.bz2/DMG the Docker build never consumes.
- docker/metadata-action — doesn't simplify our staging→promote flow (can't drive imagetools, competes with Maven labels).
- Index-level OCI annotations — same data Docker Hub's UI ignores, triple-maintenance via the promote rebuild; keep Maven-owned LABELs.

dizzzz (commented):
I don't have sufficient knowledge to review the PR.

dizzzz:
(copilot feedback: positive)

---

## PR Summary

This PR fundamentally redesigns the Docker container CI/CD pipeline for eXist-db to **properly preserve Docker image attestations** (SBOM and provenance metadata). Previously, attestations were generated but immediately discarded during the build process. The new approach implements a **staging-then-promote pattern**: build multi-arch images with attestations, push them to staging tags, run tests, then promote the tested images with all attestations intact to public tags.


---

## Core Changes

The PR redesigns the workflow in three stages:

### 1. **Replace Fabric8 Plugin with `docker/build-push-action`**

The previous approach used the fabric8 `docker-maven-plugin` which couldn't preserve multi-arch attestations. The PR switches to the idiomatic GitHub Actions approach documented in [Docker's build attestations guide](https://docs.docker.com/build/ci/github-actions/attestations/).

**Why this matters:** `docker/build-push-action` with `provenance: mode=max` and `sbom: true` produces attestation manifests natively and preserves them through server-side operations.

### 2. **Implement Staging-Then-Promote Workflow**

Instead of building and publishing directly to public tags, the new flow is:
- **Build → Push to `-staging` tag** (multi-arch with attestations)
- **Test the amd64 variant** pulled from the staging tag
- **Promote on success** (server-side copy via `docker buildx imagetools create`, no rebuild)
- **Verify** that promoted images carry both architectures and attestation manifests

This ensures that only tested images reach production tags, and attestations are preserved.

### 3. **Simplify Maven's Role**

Maven now **only** handles context assembly, Dockerfile filtering, and label metadata generation. It no longer builds or pushes images — `docker/build-push-action` takes over the build/push step entirely.

---

## Other Changes

### **.github/workflows/ci-container.yml** 
[Full diff](https://github.com/eXist-db/exist/pull/6543/files#diff-f0be4e50-new)

**Key structural changes:**

<details open>

<summary><strong>1. Concurrency Control (Lines 14–18)</strong></summary>

```yaml
concurrency:
  group: ci-container-${{ github.ref }}
  cancel-in-progress: false
```

**Why:** Prevents race conditions on the shared `-staging` tag. If two `develop` pushes happen concurrently, Build B could overwrite A's staging image between A's test and promote steps. This serializes runs per branch/tag.

</details>

<details open>

<summary><strong>2. Publish Target Resolution (Lines 60–84)</strong></summary>

A single-pass calculation determines whether this run will publish:
- `develop` branch → `channel=latest`, `publish=true`
- `eXist-*` tags → `channel=release`, `publish=true`
- PRs and forks → `publish=false` (local test only)

This is outputted to `steps.meta.outputs.{publish,channel,version}` and reused throughout.

</details>

<details open>

<summary><strong>3. Build & Push Staging Images (Lines 112–147)</strong></summary>

Two new `docker/build-push-action` steps (main + DEBUG):

```yaml
- name: Build & push staging image (main)
  if: steps.meta.outputs.publish == 'true'
  uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83
  with:
    context: exist-docker/target/exist-docker-${{ steps.ver.outputs.version }}-docker-dir
    file: exist-docker/target/classes/Dockerfile
    platforms: linux/amd64,linux/arm64
    push: true
    provenance: mode=max
    sbom: true
    cache-from: type=gha
    cache-to: type=gha,mode=max
    tags: existdb/existdb:${{ steps.meta.outputs.channel }}-staging
```

**Why:** 
- **Multi-arch (`amd64 + arm64`)** built in one action call
- **`provenance: mode=max` + `sbom: true`** produces full attestations
- **`type=gha` caching** speeds up multi-arch rebuilds
- **Only `-staging` tags published** — nothing untested reaches Docker Hub

</details>

<details open>

<summary><strong>4. Drop Emulated ARM64 Testing (Comment at Line 195)</strong></summary>

The PR intentionally drops QEMU-based arm64 testing in CI. Justification: QEMU produces unreliable results and frequently crashes. ARM64 is validated locally on Apple silicon instead.

**Note:** This is a deliberate decision acknowledged in the description (Thanks @duncdrum).

</details>

<details open>

<summary><strong>5. Promote & Verify (Lines 200–235)</strong></summary>

After tests pass:

```yaml
- name: Promote staging images to public tags
  run: |
    docker buildx imagetools create \
      -t existdb/existdb:${CH} \
      -t existdb/existdb:${V} \
      existdb/existdb:${CH}-staging
```

**Why:** `docker buildx imagetools create` copies the entire OCI index (both architectures + attestation manifests) **server-side, no rebuild**. This preserves attestations and avoids the original `buildx --load` limitation.

Then a verification gate ensures both architectures and at least one attestation manifest are present, failing the run otherwise.

</details>

---

### **exist-docker/pom.xml**
[Full diff](https://github.com/eXist-db/exist/pull/6543/files#diff-c3de4dda-new)

**Changes:**

```xml
<!-- Base images as single source of truth -->
<docker.base.image>gcr.io/distroless/java21-debian13:latest</docker.base.image>
<docker.base.image.debug>eclipse-temurin:21-jdk-noble</docker.base.image.debug>

<!-- OCI metadata properties -->
<docker.source>https://github.com/exist-db/exist.git</docker.source>
<docker.url>https://www.exist-db.org/exist/apps/doc/docker.xml</docker.url>
<docker.vendor>existdb</docker.vendor>
<docker.licenses>LGPL-2.1-or-later</docker.licenses>
<docker.title>existdb</docker.title>
```

**Rationale:**
- `docker.base.image` and `docker.debug.image` now centralize the base choice — the Dockerfile `FROM` and the label both read from here, ensuring no drift.
- OCI metadata properties are consumed by both the Dockerfile labels and provide consistent metadata.
- The docker-maven-plugin now has **two image definitions**: test-only (single-arch, no push) and release (multi-arch, pushed via GitHub Actions).

---

### **Dockerfiles (main + DEBUG)**
[Main diff](https://github.com/eXist-db/exist/pull/6543/files#diff-dc9206241f0b30705176d3fdaa8dc5c4-new) | [Debug diff](https://github.com/eXist-db/exist/pull/6543/files#diff-0dea84d5a80f689f17d032b3799b1954ed6c153a-new)

**Labels migrated from deprecated `org.label-schema.*` to OCI standard `org.opencontainers.image.*`:**

```dockerfile
# Old (deprecated)
LABEL org.label-schema.build-date=${maven.build.timestamp} \
      org.label-schema.description="${project.description}" \
      org.label-schema.vendor="existdb"

# New (OCI-compliant)
LABEL org.opencontainers.image.created="${maven.build.timestamp}" \
      org.opencontainers.image.description="${project.description}" \
      org.opencontainers.image.vendor="${docker.vendor}" \
      org.opencontainers.image.licenses="${docker.licenses}" \
      org.opencontainers.image.base.name=${docker.base.image}
```

**Why:** OCI Image Spec is the modern standard; `org.label-schema` is deprecated. The `base.name` label ensures the base image is recorded in metadata.

---

### **Test Suite Update** 
[Diff](https://github.com/eXist-db/exist/pull/6543/files#diff-80015af5f315a8592ef4a4576df2f12827cef6cc-new)

```bash
# Old: referenced an orphaned registry tag
run docker create --name ex-mod -p 9090:8080 existdb/exist-ci-build:latest

# New: uses the canonical local tag
run docker create --name ex-mod -p 9090:8080 existdb/existdb:ci
```

Also fixed a path in the test: `exist/config/conf.xml` → `exist/etc/conf.xml`.

---

## Merge Readiness and Risk Assessment

**Status:** Ready for upstream validation — CI gates are well-designed, but **the promote/verify steps need to run on the actual `develop` or tag pipeline** to confirm the staging images are properly built with attestations.

**Observations:**

- ✅ **Clear intent & design** — the staging-then-promote pattern is well-reasoned and documented.
- ✅ **Backward compatibility** — PRs and forks still get a working amd64 test build locally; nothing breaks on those paths.
- ✅ **Idiomatic GitHub Actions** — using `docker/build-push-action` aligns with Docker's documented best practices for attestations.
- ✅ **Verification gate** — the final step that checks for both architectures and attestation manifests is a good safeguard.
- ⚠ **Concurrency serialization** — may slow down rapid iterations on `develop` (builds queue), but prevents data corruption.
- ⚠ **Timeout increase** — 60 → 90 minutes to accommodate multi-arch builds + tests + staging push. Monitor for flakiness.
- ⚠ **Test plan incomplete** — the test plan shows checkboxes, some still unchecked. The crucial ones are:
  - [ ] Upstream `develop` run pushes `latest-staging`, tests, promotes `latest` + `7.0.0-SNAPSHOT`
  - [ ] Verify attestations on published image via `docker buildx imagetools inspect`



- **Line 72 in workflow (removed):** The build step previously passed `-Ddocker.sbom=true -Ddocker.provenance=true` to Maven, which is now removed. This is correct because Maven is no longer responsible for building images — only assembling the context. However, confirm that Maven's docker-maven-plugin execution is genuinely **disabled** for non-release builds (check the POM filter configuration). If the plugin still tries to build during `package`, it could attempt a build with `-Ddocker.sbom=false` (the default), which would fail or produce confusing output.

- **Line 195 (comment):** Dropping emulated ARM64 testing is pragmatic, but document in a visible place (README or deployment guide) that arm64 images are only validated locally on Apple silicon. This ensures operators know that CI's arm64 coverage is limited.

- **Line 218 (verify step):** The verification uses `jq` to parse manifests. Ensure `jq` is installed in the ubuntu-latest runner. It usually is, but this is a silent dependency that could break if the runner image changes.

---

## Possible Improvements



- **Consider adding a "rollback" label:** The `-staging` tag is intentionally left in place after promotion "as the inspectable last CI build." It would be useful to also label the promoted images with a `staging-digest-promoted-at: <timestamp>` label or annotation so operators can trace which staging build was promoted and when. Alternatively, document a command to inspect the image's provenance attestation to trace back to the `develop` build that produced it.

- **Make the verify step output more detailed:** The error messages in the verify step (lines 220–232) are good, but consider adding a `--format '{{ json . }}'` dump of the full manifest so failed runs leave behind debuggable artifacts. This would help troubleshoot issues like missing attestations or unexpected manifest shapes on Docker Hub.

- **Consider adding a `skip-run-on-untested-tag` safeguard:** If a manually-triggered `workflow_dispatch` or a scheduled run on a new branch accidentally sets `publish=true` for an unintended target, there's no guard. Add a confirmation step or require a specific workflow input to enable promotion on non-standard refs.

---

reinhapa (approved):

duncdrum (approved):

PR #6545, opened 2026-07-04
[hotfix] repair build

dizzzz:
Address double definition of error code


duncdrum (approved):
Thanks for the quick fix. How did we miss that during review??

dizzzz:
not sure. maybe a merge issue

joewiz (approved):

Thanks for the fast hotfix, Dannes — this resolves the duplicate `EXXQDY0007` and gets `develop` compiling again. Approving and merging now to unblock everyone's CI.

One small thing for a follow-up (no need to hold the merge): the module-header comment at `permission-error-codes.xqm:26` still describes the permission failure as `EXXQDY0007`:

> `: permission failure (EXXQDY0007) from an invalid-argument failure (EXXQDY0008)`

Since this PR moves permission-denied to `EXXQDY0009`, that line is now stale — and `EXXQDY0007` there now points at the streaming code. You already updated the parallel inline comment at line 68; just the header one at line 26 to match. It's comment-only, so it can be cleaned up whenever.

PR #6544, opened 2026-07-04
[Dev6] fix XSLT imports

line-o:
### Description:

Fixes several long-standing issues with imports and URI resolution in fn:transform

### Reference:

fixes #5052
refs #6065

### Type of tests:

XQSuite tests added

Issue #6065, opened 2026-02-26
[BUG] test isolation in XQSuite


line-o:
**Describe the bug**

While working on #6060 I noticed that tests, specifically the tests calling `fn:transform` are inadequately isolated against each other.
I want to stress that this is either an issue within fn:transform or the xqsuite test runner or both. See also this issue with [main modules that evaluate to a deferred function call](https://github.com/eXist-db/exist/issues/4615) which leads to undeterministic test case failures.

I [created a gist with the necessary information](https://gist.github.com/line-o/4731ed367f437be9cd036b152a18d340) in order to keep this issue description brief. 

The gist (of the gist)

```xquery
declare %test:assertEquals("<a />") function xqsuite-test-isolation:test-1() {
    transform($options)?output
};

declare %test:assertEquals("<a />") function xqsuite-test-isolation:test-2() {
    transform($options)?output
};
```

`test-2` will fail because it receives **2** nodes as a result instead of just one. The first node being the result of `test-1`

**Expected behavior**

The tests to be isolated enough so that results from a previous test can _never_ influence a later test case.

**To Reproduce**

Store https://gist.github.com/line-o/4731ed367f437be9cd036b152a18d340#file-xqsuite-test-isolation-xqm in an instance of exist 7.0.0-SNAPSHOT or 6.4.1-SNAPSHOT and run them.

While the first test that is executed will pass all _following_ ones do not. They all fail with a cardinality error.

Add the `delivery-format` option _or_ the additional empty-sequence to see that this mitigates the issue entirely.

**Context (please always complete the following information)**

- Build: eXist-7.0.0-SNAPSHOT (1d5848bfe3bbe0272a336f8aacef51a959c5d490)
- Java: 21.0.8 (Eclipse Adoptium)
- OS: Linux 6.12.54-linuxkit (aarch64)

**Additional context**
 
- How is eXist-db installed? docker
- Any custom changes in e.g. `conf.xml`? none

duncdrum:
Before we can tackle test case isolation within a spec, we need to tackle the junit upgrade and isolation per spec.

line-o:
@duncdrum I can't see where I can remove the block as it turns out that the bug is in `fn:transform` not in XQSuite.

Issue #5052, opened 2023-09-11
[BUG] Relative URL resolution not working as expected in fn:transform() under eXist-db 6.2


yamahito:
**Describe the bug**
When referring to a database resource from within a fn:transform() transformation, that resource is instead interpreted as `$EXIST-HOME/../path/to.resource` throwing the following error:

```err:FOXT0002 Could not transform input: I/O error reported by XML parser processing file:[...]```

However the resource is correctly identified when running the same transform with `transform:transform`

**Expected behavior**
Expected behaviour is for the relative path to be resolved from the containing collection of the transformation stylesheet.

**To Reproduce**
e.g. at `xmldb:exist///db/apps/issue/subfolder/another.xsl`:
```xslt
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="3.0">
    
    <xsl:template match="*">
        <xsl:copy>
            <xsl:text>Hello World</xsl:text>
        </xsl:copy>
    </xsl:template>
    
</xsl:stylesheet>
```

e.g. at `xmldb:exist///db/apps/issue/transform.xq`
```xquery
xquery version "3.1";

let $stylesheet :=
    <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="3.0">
        <xsl:template match="*">
            <xsl:sequence select="transform(map{{
                    'stylesheet-location': 'subfolder/another.xsl',
                    'source-node': .
                }})?output"/>
        </xsl:template>
    </xsl:stylesheet>

let $node := <test/>

let $parameters := ()

let $map-parameters := map{}

(: This works as expected: :)
(: return transform:transform($node, $stylesheet, $parameters) :)

(: This throws the error: :)
return fn:transform(map{
        "stylesheet-node":  $stylesheet,
        "source-node":      $node,
        "stylesheet-params": $map-parameters
    })?output
```

 - Build: eXist 6.2.0 c8fa4958b6d4a50bd0cba7f3e76a150226414187 2023-02-04T22:42:29Z
 - Java: [openjdk 17.0.8.1 2023-08-24]
 - OS: [Ubuntu]

**Additional context**
- How is eXist-db installed? [JAR installer]
- Any custom changes in e.g. `conf.xml`: yes, enabled SaxonEE.

yamahito:
Related to #5041 ?

djbpitt:
Similar behavior with:

```
[...]

Issue #5051, opened 2023-09-11
[BUG] Relative URL resolution not working as expected under XSLT in eXist 7


yamahito:
**Describe the bug**
When referring to a database resource from within a transformation, that resource is instead interpreted as `$EXIST-HOME/../path/to.resource` throwing the following error:

```exerr:ERROR Exception while transforming node: net.sf.saxon.trans.XPathException: I/O error reported by XML parser processing file:[...]```

In previous versions of exist, this worked (at least for `transform:transform`).

**Expected behavior**
Expected behaviour is for the relative path to be resolved from the containing collection of the transformation stylesheet.

**To Reproduce**
e.g. at `xmldb:exist///db/apps/issue/subfolder/another.xsl`:
```xslt
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="3.0">
    
    <xsl:template match="*">
        <xsl:copy>
            <xsl:text>Hello World</xsl:text>
        </xsl:copy>
    </xsl:template>
    
</xsl:stylesheet>
```

e.g. at `xmldb:exist///db/apps/issue/transform.xq`
```xquery
xquery version "3.1";

let $stylesheet :=
    <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="3.0">
        <xsl:template match="*">
            <xsl:sequence select="transform(map{{
                    'stylesheet-location': 'subfolder/another.xsl',
                    'source-node': .
                }})?output"/>
        </xsl:template>
    </xsl:stylesheet>

let $node := <test/>

let $parameters := ()

let $map-parameters := map{}

return transform:transform($node, $stylesheet, $parameters)

(: fn:transform() throws the same error :)
(:return fn:transform(map{:)
(:        "stylesheet-node":  $stylesheet,:)
(:        "source-node":      $node,:)
(:        "stylesheet-params": $map-parameters:)
(:    })?output:)
```

 - Build: eXist 7.0.0-SNAPSHOT d0d511ef39a0ed62b81687d62fda36720f154ba8 2023-09-08T12:40:17Z
 - Java: [openjdk 17.0.8.1 2023-08-24]
 - OS: [Ubuntu]

**Additional context**
- How is eXist-db installed? [compiled]
- Any custom changes in e.g. `conf.xml`: yes, enabled SaxonEE.

PR #6546, opened 2026-07-05
[bugfix] Forward port PR #6544: fix XSLT imports and document builder isolation

line-o:
## Summary

Forward port of PR #6544 from develop-6.x.x to develop, fixing critical issues with XSLT stylesheet resolution and document builder isolation in fn:transform. Adapted for Saxon 12.5 compatibility.

Fixes:
- **Issue #5051**: XSLT relative imports on exist 7 
- **Issue #6065**: Document builder isolation in concurrent fn:transform calls

### 1. Document Builder Isolation (Delivery.java)

**Problem**: When multiple fn:transform calls execute in the same XQuery context, they share the same MemTreeBuilder. This causes output corruption when transforms execute concurrently.

**Solution**: Always use a fresh MemTreeBuilder for each transform result instead of reusing the context's builder.

### 2. Persistent Node Handling (Convert.java)

**Problem**: Persistent nodes stored in the database (NodeProxy instances) weren't properly dereferenced when passed as stylesheet parameters.

**Solution**: 
- Check Document interface instead of specific DocumentImpl class
- Dereference NodeProxy to underlying DOM node in of(Sequence) method
- Enables passing database-stored nodes as stylesheet parameters

### 3. XSLT Stylesheet Resolution for Relative URIs (Issue #5052 - Root Cause Fix)

**Root Causes** (discovered during porting):

1. **URIResolution.resolveURI()**: Database paths like "/db/collection/stylesheet.xsl" were incorrectly treated as relative URIs because `URI.isAbsolute()` returns false (they lack a scheme component).
   - **Fix**: Treat database paths (starting with "/" or "xmldb:") as absolute for resolution purposes

2. **URIResolution.resolveDocument()**: Did not set SystemId on returned DOMSource objects. When Saxon resolved and loaded included stylesheets, it had no base URI for resolving further nested includes.
   - **Fix**: Set SystemId = location on all returned DOMSource objects

3. **Options.resolvePossibleStylesheetLocation()**: Simplified to use the resolved location directly as SystemId, ensuring consistent base URI setup for Saxon's compile-time resolver.

### 4. Other Changes

**Transform.java**: Updated createDestination() call sites to match new Delivery.java signature

## Test Coverage

Two comprehensive XQSuite test modules verify all fixes:

- **fnTransform5052.xqm** (13 tests): Relative stylesheet resolution with xsl:include/xsl:import
- **fnTransform6065.xqm** (2 tests): Persistent database nodes as stylesheet parameters

**Result**: All 7110 tests pass with zero failures and zero errors.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
@@ -169,7 +169,12 @@ class Options {
             stylesheetBaseUri = xsltSource._1;
         }
         if (!StringUtils.isEmpty(stylesheetBaseUri)) {
-            resolvedStylesheetBaseURI = Optional.of(resolveURI(new AnyURIValue(stylesheetBaseUri), context.getBaseURI()));
+            // Only resolve if it's not already absolute (database URIs start with "/" or "xmldb:")
+            if (stylesheetBaseUri.startsWith("/") || stylesheetBaseUri.startsWith("xmldb:") || stylesheetBaseUri.startsWith("exist://")) {
+                resolvedStylesheetBaseURI = Optional.of(new AnyURIValue(stylesheetBaseUri));
+            } else {
+                resolvedStylesheetBaseURI = Optional.of(resolveURI(new AnyURIValue(stylesheetBaseUri), context.getBaseURI()));
+            }
         } else {
             resolvedStylesheetBaseURI = Optional.empty();
         }
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
@@ -552,13 +552,12 @@ class Options {
                     "Can not access '" + location + "'" + e.getMessage());
         }
         if (document != null && document.hasOne() && Type.subTypeOf(document.getItemType(), Type.NODE)) {
-            if (document instanceof NodeProxy) {
-                final DOMSource source = new DOMSource(((NodeProxy) document).getNode());
+            if (document instanceof NodeProxy nodeProxy) {
+                final DOMSource source = new DOMSource(nodeProxy.getNode());
                 source.setSystemId(location);
                 return Tuple(location, source);
             }
-            else if (document.itemAt(0) instanceof Node) {
-                final Node node = (Node) document.itemAt(0);
+            else if (document.itemAt(0) instanceof Node node) {
                 final DOMSource source = new DOMSource(node);
                 source.setSystemId(location);
                 return Tuple(location, source);
--- a/exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
+++ b/exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
@@ -32,9 +32,12 @@ import net.sf.saxon.s9api.QName;
 import net.sf.saxon.s9api.XdmValue;
 import org.apache.commons.lang3.StringUtils;
 import org.exist.dom.memtree.NamespaceNode;
+import org.exist.dom.persistent.NodeProxy;
+import org.exist.security.PermissionDeniedException;
 import org.exist.xquery.ErrorCodes;
 import org.exist.xquery.XPathException;
 import org.exist.xquery.XQueryContext;
+import org.exist.xquery.util.DocUtils;
 import org.exist.xquery.functions.array.ArrayType;
 import org.exist.xquery.functions.fn.FnTransform;
 import org.exist.xquery.functions.map.MapType;
@@ -467,7 +470,7 @@ class Options {
         final List<Tuple2<String, Source>> results = new ArrayList<>(1);
         final Optional<String> stylesheetLocation = Options.STYLESHEET_LOCATION.get(options).map(StringValue::getStringValue);
         if (stylesheetLocation.isPresent()) {
-            results.add(Tuple(stylesheetLocation.get(), resolveStylesheetLocation(stylesheetLocation.get())));
+            results.add(resolveStylesheetLocation(stylesheetLocation.get()));
         }
 
         final Optional<Node> stylesheetNode = Options.STYLESHEET_NODE.get(options).map(NodeValue::getNode);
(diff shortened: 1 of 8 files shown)


line-o:
I propose to squash all four commits

```
[bugfix] Forward port PR #6544: fix XSLT imports and document builder isolation

Fixes #5051 (XSLT import resolution) and #6065 (document builder isolation) by
porting and refining key changes from develop-6.x.x, adapted for Saxon 12.5.

## Document Builder Isolation (Issue #6065)
- Always use fresh MemTreeBuilder for fn:transform results instead of sharing
  the query context's builder
- Prevents corruption when multiple fn:transform calls occur in parallel within
  the same execution context
- Fixes concurrent transform output corruption

## Persistent Node Handling
- Handle persistent NodeProxy nodes by dereferencing to underlying DOM nodes
- Enables passing database-stored nodes as stylesheet parameters to fn:transform
- Uses Document interface checks instead of specific implementation classes

## Stylesheet Resolution for Relative xsl:include/xsl:import (Issue #5052)

Root cause analysis revealed three interconnected issues:

1. URIResolution.resolveURI() incorrectly treated database paths (starting with
   "/" or "xmldb:") as relative URIs because they lacked a scheme component.
   A path like "/db/collection/stylesheet.xsl" is an absolute path but not an
   absolute URI in the RFC 3986 sense. Fixed by checking for database path
   prefixes in addition to URI scheme.

2. URIResolution.resolveDocument() did not set SystemId on returned DOMSource
   objects. When Saxon resolved xsl:include/xsl:import hrefs and loaded
   included stylesheets, it had no base URI for resolving further nested
   includes. Fixed by setting SystemId = location on all returned sources.

3. Options.resolvePossibleStylesheetLocation() simplified to use the resolved
   location parameter directly as SystemId, ensuring consistent base URI setup
   for Saxon's compile-time URI resolver. This supports the chain:
   - Stylesheet at /db/coll/main.xsl
   - Includes sub/inc.xsl → resolved to /db/coll/sub/inc.xsl (base URI via SystemId)
   - Includes nested.xsl → resolved to /db/coll/sub/nested.xsl (base URI via SystemId)

## Convert.java
- Check Document interface instead of specific DocumentImpl implementation
- Add NodeProxy dereference in of(Sequence) method to handle persistent nodes

## Transform.java
- Updated createDestination() call sites for new Delivery.java signature

## Test Coverage
- fnTransform5052.xqm: 13 tests for relative xsl:include/xsl:import and
  stylesheet-location resolution in database-stored stylesheets
- fnTransform6065.xqm: 2 tests for persistent nodes as stylesheet parameters

All tests pass with exit code 0.

Co-Authored-By: Claude Haiku 4.5 <noreply@anthropic.com>
```

line-o:
@duncdrum pointed out that #5051 needs to be looked at in this context

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/URIResolution.java
```diff
@@ -56,8 +56,11 @@ static AnyURIValue resolveURI(final AnyURIValue relative, final AnyURIValue base
         if (relativeURI.isAbsolute()) {
             return relative;
         }
-        var baseURI = new URI(base.getStringValue() );
-        if (!baseURI.isAbsolute()) {
+        var baseString = base.getStringValue();
```
dizzzz: var :-/ horrible to review. :-)

dizzzz (approved):

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
```diff
@@ -496,19 +504,68 @@ private Tuple2<String, Source> getStylesheet(final MapType options) throws XPath
...
+        Sequence document;
+        try {
+            document = DocUtils.getDocument(context, location);
+        } catch (final PermissionDeniedException e) {
+            throw new XPathException(fnTransform, ErrorCodes.FODC0002,
+                    "Can not access '" + location + "'" + e.getMessage());
+        }
+        if (document != null && document.hasOne() && Type.subTypeOf(document.getItemType(), Type.NODE)) {
+            if (document instanceof NodeProxy) {
+                final DOMSource source = new DOMSource(((NodeProxy) document).getNode());
```
reinhapa: use instanceof cast variable:
```suggestion
            if (document instanceof NodeProxy nodeProxy) {
                final DOMSource source = new DOMSource(nodeProxy.getNode());
```

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/Options.java
```diff
@@ -496,19 +504,68 @@ private Tuple2<String, Source> getStylesheet(final MapType options) throws XPath
...
+                source.setSystemId(location);
+                return Tuple(location, source);
+            }
+            else if (document.itemAt(0) instanceof Node) {
+                final Node node = (Node) document.itemAt(0);
```
reinhapa: use instanceof cast variable
```suggestion
            else if (document.itemAt(0) instanceof Node node) {
```

reinhapa (changes requested):

line-o:
Good catches will change code - wonder if this is covered by a codacy rule

exist-core/src/main/java/org/exist/xquery/functions/fn/transform/URIResolution.java
line-o: Fair point — `var` is gone from `URIResolution` entirely now (0b9170fe4c, and the two bugfixes below).

Looking closer at these lines while removing it turned up two real bugs, both fixed with tests:

- **`xmldb:xmldb:` prefix doubling** (f200c0e06a). `XmldbURI#getURI` strips the `xmldb:` prefix for `xmldb:exist://…` URIs but keeps it for the short `xmldb:/db/…` form, so unconditionally prepending `XMLDB_URI_PREFIX` doubled it. This PR makes a resolved stylesheet its own system id, so it was reached by any relative `xsl:import` *within* an imported stylesheet — i.e. exactly the nested-import case the PR sets out to fix.
- **Relative locations resolving into the parent collection** (0ddf7a6e6e). RFC 3986 discards the last segment of the base, which is right for a document but not for a collection: `style.xsl` against the collection `/db/apps/app` resolved to `/db/apps/style.xsl`. A collection and a document are not distinguishable by path alone, so the absence of an extension in the last segment is now taken to mean a collection. Outside the database RFC 3986 still applies unchanged. This is a documented heuristic — an extensionless document is mistaken for a collection; happy to switch to asking the broker whether the base names a collection if you prefer exactness over keeping this pure string logic.

Both are covered by tests that fail without the fix. `FunTransformTest` 7/7, `XQuery3Tests` 1089/1089.

reinhapa (approved):

dizzzz (approved):
what about https://github.com/eXist-db/exist/issues/5051 ?

line-o:
What about it? It's mentioned in description.

line-o:
Is there something that blocks this PR I am not aware of?

Issue #3062, opened 2019-10-21
modify conf.xml


duncdrum:
### What is the problem
`conf.xml` still has a few linline holdouts. These are tricky to process via standard xsl/xquery and unnecessarily complicate hand-off configuration of exist.  A boolean `@enablded` attribute would be easier to process, and to read.           

### What did you expect
Easy modability of conf.xml using standard xml tricks.

### First step

- add `@enaled` to conf schema , uncomment all inlines and set to `@enabled="false"`
- go through elements defined in schema and check if they should get the new attribute
- make sure exist understands the meaning of `@enabled="false|true"` wherever it was added in step 1. 

### Context information
Please always add the following information
- eXist-db version: all

adamretter:
Can you provide an example of an "inline" which is troublesome to you?

dizzzz:
@duncdrum 's suggestion is to switch on/off functionality (e.g  modules) with a flag, instead of putting configuration parts in XML comments.

adamretter:
@dizzzz I still don't understand what that would look like vs what it looks like currently

dizzzz:
I think 

```xml
<!-- <module uri="http://exist-db.org/xquery/spatial" class="org.exist.xquery.modules.spatial.SpatialModule"/>
-->
```

vs

```xml
<module uri="http://exist-db.org/xquery/spatial" class="org.exist.xquery.modules.spatial.SpatialModule" enabled="no"/>
```

duncdrum:
yes similarly allow triggers to be enabled / disabled 
```xml
                <!--<trigger class="org.exist.collections.triggers.XQueryStartupTrigger"/>-->
```
as well as 

```xml
        <!-- 
        <job class="bar.foo.myjob" period="600000" delay="300000" repeat="10" />
        -->
```

and 

```xml
<!--
                <feature name="http://xml.org/sax/features/external-general-entities" value="false"/>
                <feature name="http://xml.org/sax/features/external-parameter-entities" value="false"/>
                <feature name="http://javax.xml.XMLConstants/feature/secure-processing" value="true"/>
                -->
```

for consistencies sake i would also consider

```xml
<custom-filter class="org.exist.FirstFilter"/>
<backup-filter class="org.exist.FirstFilter"/>
```

dizzzz:
`<feature>`  is really different as these parameters are 1:1 set as "features" (key, value) the the SAXparsers...

duncdrum:
[...]

PR #6550, opened 2026-07-06
[feature] Add @enabled attribute to conf.xml elements to disable entries without removing them

duncdrum:
## Summary

Adds an `enabled="yes|no"` attribute (default `yes`) to `<module>`, `<trigger>`, `<job>`, `<feature>`, `<parameter>`, and `<property>` elements in `conf.xml`/`conf.xsd`. Setting `enabled="no"` disables an entry at startup without removing it — the schema-valid equivalent of commenting it out.

Addresses #3062, part 1 of 2 (the `@enabled` attribute; SPI autodiscovery, part 2, is in #6551). Builds on #6528/#6530/#6531.

## What changed

**`schema/conf.xsd`** (2.4.0 → 2.6.0) — `@enabled` (`yes_no` type, default `yes`) on the six element types above; tightened cron/class-name/period patterns along the way.

**`exist-distribution/src/main/config/conf.xml`** — the four live startup triggers now carry explicit `enabled="yes"`; two previously commented-out scheduler jobs (`ConsistencyCheckTask`, `DataBackup`) promoted to live `<job enabled="no">` entries.

**`Configuration.java`** — `configureModules()`/`configureStartup()`/`configureScheduler()`/`parseFeatures()`/`parseParameters()`/`parseProperties()` all skip `enabled="no"` entries; fixes a latent case-sensitivity bug where `"No"` wasn't recognized.

## Compatibility

Existing `conf.xml` files without the attribute are unaffected (default is `yes`).

## Test plan

- [ ] `mvn validate -pl exist-distribution` — canonical `conf.xml` validates against `conf.xsd` 2.6.0
- [ ] `mvn test -pl exist-core -Dtest=org.exist.config.ConfigurationTest` — existing tests pass
- [ ] Manual: `enabled="no"` on a `<module>` entry → module not loaded at startup
- [ ] Manual: flip a scheduler job from `enabled="no"` to `enabled="yes"`, restart → job runs


exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -854,53 +877,98 @@ public synchronized void shutdown() {
...
+            t.setDaemon(true);
+            t.start();
         }
     }
 
-    private static class BrokerPoolAndJettyShutdownHook implements Runnable {
+    private final class BrokerPoolAndJettyShutdownHook implements Runnable {
         private final Server server;
 
         BrokerPoolAndJettyShutdownHook(final Server server) {
```
reinhapa, line 974: Could be turned into a record

```suggestion
    private record BrokerPoolAndJettyShutdownHook(final Server server) implements Runnable {
```

reinhapa (changes requested):

exist-core/src/main/java/org/exist/jetty/JettyStart.java
duncdrum: @reinhapa its not quite the same pattern as the `CatalogResolver` `run()` captures `serverStopRequested` from the enclosing JettyStart instance, the record you suggested loses access to outer instance fields. I could refactor to 
```java
private record BrokerPoolAndJettyShutdownHook(Server server, AtomicBoolean serverStopRequested) implements Runnable {
    @Override
    public void run() {
        BrokerPool.stopAll(true);
        if (serverStopRequested.compareAndSet(false, true)) {
            try {
                server.stop();
            } catch (final Exception e) {
                logger.warn("Error stopping Jetty server: {}", e.getMessage(), e);
            }
        }
    }
}
```

but is that actually better, arguable its more verbose. What do you think?

duncdrum:
@reinhapa thanks for the suggestion. I lean towards applying the record form with both fields passed explicitly —  — but wanted your take before committing to it. Does that form work for you, or would you prefer to keep the inner class as-is?

reinhapa (approved):

dizzzz (approved):

dizzzz:
@duncdrum good to go?

duncdrum:
@dizzzz yes

PR #6551, opened 2026-07-06
[feature] Add ModuleFactory/IndexFactory SPI and XQuery configuration observability API

### LINKED ISSUE #3062: modify conf.xml

### LINKED ISSUE #6563: [feature] XQuery observability API for effective server configuration
Supersedes #3331.

## Problem

#3331 identified the need for XQuery access to the active `conf.xml`. Two recent PRs (#6550, #6551) have sharpened the need: with `@enabled="no"` suppression and ServiceLoader-based SPI autodiscovery in place, a module can be active with no `conf.xml` entry at all, and suppressed entries are silently discarded at startup. Existing functions (`util:registered-modules-info()`, `util:mapped-modules()`) can no longer answer "what is actually loaded and why?"

Concrete use cases:
- An EXPath package install script adapts its behaviour based on which indexes or modules are active
- A DBA audits which modules came from SPI autodiscovery vs explicit `conf.xml` wiring
- A monitoring query checks whether a scheduler job is configured but disabled
- IDE tooling resolves `conf.xml` against `$EXIST_HOME/schema/conf.xsd` fetched from the server

### Extend `util:registered-modules-info()`

Add two keys to the existing map entries (backward-compatible; XDM maps are open):

| Key | Values | Meaning |
|-----|--------|---------|
| `registration-source` | `"spi"` / `"conf.xml"` / `"built-in"` | How the module was registered |
| `enabled` | `"yes"` / `"no"` | Whether the entry was active or suppressed at startup |

`enabled="no"` entries are currently discarded silently; they would need to be retained in a suppressed-entries list in `Configuration` to be reportable here.

### New `util:registered-indexes()`

Returns a sequence of maps `{id, class, registration-source, enabled}` for all index modules known at startup (active and suppressed). Follows the naming convention of `util:registered-modules-info()`. No equivalent exists today.

### New `system:get-configuration()`

Returns the effective `conf.xml` as an in-memory XML node — the configuration as parsed and active, not necessarily identical to the file on disk (e.g. defaults filled in, values resolved).

### New `system:get-configuration-schema-version($schema as xs:string)`

Returns the `schemaVersion` string for a named native schema (`"conf"`, `"collection.xconf"`, `"controller-config"`, etc.) — reads from `SchemaVersion.*` constants generated at build time, no disk access.

### New `system:get-configuration-property($path as xs:string)`

An XPath-like accessor into the parsed configuration, e.g. `system:get-configuration-property("db-connection/@pool-size")`. Open for discussion — may be redundant if `system:get-configuration()` is available, but could be useful for targeted queries without returning the full document.

### Schema HTTP endpoint

Serve `$EXIST_HOME/schema/*.xsd` at `/exist/schema/{name}.xsd` via a Jetty static resource handler. The schemas are already shipped in every distribution layout (since #6528); this makes them reachable over HTTP for tooling and dynamic validation.

## Security

> "There are some things in the `conf.xml` that should not be visible apart to anyone apart from `dba`"
> — @adamretter, #3331

All functions in this proposal are **DBA-only**, following the established pattern of `system:get-running-xqueries()` and similar.

`system:get-configuration()` requires additional care: `conf.xml` contains credentials (admin password hash, LDAP bind passwords, SMTP passwords, JDBC connection strings, arbitrary `<parameter name="password" value="..."/>` entries). The returned node must redact sensitive values. The exact redaction strategy needs refinement during implementation, but the current thinking is:

- Redact attribute and element values whose local name matches a case-insensitive set of known credential markers: `password`, `passwd`, `secret`, `credential` — note that `key` and `token` are intentionally excluded as too broad (they also appear in non-sensitive contexts such as cache key sizes and MIME type tokens)
- Redact content of authentication-related sections (e.g. security realm entries)
- Replace redacted values with `[redacted]` rather than omitting the attribute or element — so the structural shape of the configuration remains visible

The redaction list should be treated as a starting point for review, not a final allowlist.

`util:registered-modules-info()` and `util:registered-indexes()` carry no credential values; DBA-only access is sufficient.

The schema HTTP endpoint (`/exist/schema/`) serves read-only XSD files with no sensitive content; it does not require authentication.

## Related

- Supersedes #3331
- Builds on #6528 (schema governance + `SchemaVersion` constants)
- Builds on #6550 (`@enabled` attribute — suppressed entries need to be retained, not discarded, to be reportable)
- Builds on #6551 (SPI autodiscovery — registration source tagging needed in `Configuration`)
- See also #3062 (conf.xml restructuring, partially addressed by #6550 and #6551)

duncdrum:
## Summary

`ServiceLoader`-based SPI for XQuery modules and index modules, so bundled implementations are auto-discovered without explicit `conf.xml` entries. New installs get a leaner default config; existing hand-maintained configs are unaffected — explicit entries always win over SPI, and `enabled="no"` (from #6550) suppresses an SPI entry without removing the JAR.

Also delivers the XQuery observability API this SPI/`@enabled` work created a need for (#6563): with modules and indexes now able to be active with no `conf.xml` entry at all, or silently suppressed, the DBA needs a way to ask "what's loaded and why."

Closes #3062 — delivers both parts: the `@enabled` attribute (#6550) and the SPI/CDI-style autodiscovery this PR adds.
Closes #6563 — the observability API this and #6550 created a need for.

## What changed

**`ModuleFactory` SPI** — new `org.exist.xquery.ModuleFactory` interface; `Configuration.configureModules()` scans `ServiceLoader<ModuleFactory>` before the `conf.xml` loop. All 27 bundled XQuery modules register via `META-INF/services/`.

**`IndexFactory` SPI** — new `org.exist.indexing.IndexFactory` interface; `IndexManager` registers SPI entries whose `id` has no live `conf.xml` entry. All bundled indexes (Lucene, ngram, range, sort, spatial) wired; index `name` defaults to the SPI `id` when there's no `conf.xml` element to supply one; guarded against a null/blank `id` from the factory.

**Vector model registry** — integrated with `Configuration`; `@enabled` on `<vector-models>` children; repeated WARN for unavailable models downgraded to DEBUG after the first occurrence.

**`exist-distribution/src/main/config/conf.xml`** — `<builtin-modules>` and index `<modules>` sections trimmed; the 27 bundled modules and 5 bundled indexes no longer listed individually.

**Module/index registration provenance** — new `ModuleRegistration` record and an `enabled` field on the existing `IndexModuleConfig` record, populated alongside (not instead of) the active classMap/`PROPERTY_INDEXER_MODULES` — so instantiation behavior is unchanged, but every module and index now carries a retained provenance record (`spi`/`conf.xml`/`built-in`, and whether currently suppressed).

**New `system:` observability functions** (DBA-only) — landed in the `system:` namespace per discussion on #6563 (2026-07-08 → 2026-08-19): reusing `util:` was rejected as adding to an already-mixed bag, and a new `config:`/`conf:`/`db:` namespace was rejected as a compatibility hazard and unnecessary churn.
- `system:get-registered-modules()` — `util:registered-modules-info()`'s content extended with `registration-source` (`spi`/`conf.xml`/`built-in`) and `enabled`, including `enabled="no"`-suppressed entries.
- `system:get-registered-indexes()` — same shape for index modules; no equivalent existed before.
- `system:get-configuration()` — the effective parsed `conf.xml` as an in-memory element, with credential-shaped attribute/element values redacted (including eXist's own `<parameter name="password" value="..."/>` idiom).
- `system:get-configuration-schema-version($schema)` — reads the `SchemaVersion` build-time constants by schema name.
- `system:get-configuration-property($path)` — an XPath/XQuery accessor into the same redacted document, evaluated by eXist's own XQuery engine.

**Deprecated `util:` functions** — `util:registered-modules-info()`, `util:mapped-modules()`, `util:is-module-mapped()` deprecated in place (no content change) in favor of `system:get-registered-modules()`.

**Schema HTTP endpoint** — new `SchemaServlet` serves `$EXIST_HOME/schema/*.xsd` unauthenticated at `/exist/schema/{name}.xsd` (wired via `controller-config.xml`, same pattern as `/status` → `JMXServlet`), so external tooling (IDE plugins, editors) can resolve a config file's grammar without vendoring its own copy.

## Compatibility

Explicit `conf.xml` entries always take precedence over SPI discovery for the same namespace URI/index `id` — no upgrade action required. `enabled="no"` suppresses a bundled module/index without removing its JAR.

The three deprecated `util:` functions keep their exact current behavior — deprecation is metadata-only, not a breaking change.

## Test plan

- [x] `mvn validate` (repo-wide) — schema governance / canonical instance validation passes
- [x] `mvn test -pl exist-core -Dtest="xquery.system.SystemTests"` — new `system:` observability function XQSuite coverage passes
- [x] `mvn test -pl exist-core -Dtest="xquery.util.UtilTests"` — existing `util:` coverage passes unchanged (deprecation is non-breaking)
- [x] `mvn test -pl exist-core -Dtest="org.exist.util.ConfigurationTest,org.exist.xquery.functions.system.ConfigurationRedactorTest,org.exist.http.servlets.SchemaServletTest"` — registry provenance (incl. an injected `enabled="no"` module/index), credential redaction, and `SchemaServlet` path-traversal/404 handling
- [x] Codacy PMD audit (local `.codacy/cli.sh`) clean on all new/touched files
- [ ] Manual: start eXist-db with the trimmed conf.xml; all 27 built-in modules available in eXide
- [ ] Manual: start eXist-db; Lucene/ngram/range indexes active with no explicit conf.xml entries
- [ ] Manual: `enabled="no"` on a module's conf.xml entry → not loaded despite SPI discovery, but reported (`enabled: "no"`) by `system:get-registered-modules()`
- [ ] Manual: `curl http://localhost:8080/exist/schema/conf.xsd` returns the XSD unauthenticated


exist-core/src/main/java/org/exist/util/SaxonConfiguration.java
```diff
@@ -50,11 +56,54 @@ public final class SaxonConfiguration {
...
+
+    @Override
+    public Source resolve(final ResourceRequest request) throws XPathException {
+      try {
+        return delegate.resolve(request.relativeUri, request.baseUri);
+      } catch (final TransformerException e) {
+        throw new XPathException(e);
+      }
+    }
   }
```
reinhapa: Could be converted to a Java `record`

```suggestion
  private record CatalogResourceResolver(final Resolver delegate) implements ResourceResolver {
    @Override
    public Source resolve(final ResourceRequest request) throws XPathException {
      try {
        return delegate.resolve(request.relativeUri, request.baseUri);
      } catch (final TransformerException e) {
        throw new XPathException(e);
      }
    }
}
```

reinhapa (changes requested):

exist-core/src/main/java/org/exist/util/SaxonConfiguration.java
duncdrum: good catch I will fold this into the commit that last modified the file

line-o (commented):
Resolving modules to load by default over SPI received pushback on the community call on 2026-07-06
Security concerns need to be addressed before we can consider pulling this in.

@dizzzz specific concern was: "I see autoloading modules that are not mentioned in the configuration as a risk."

My concern is: if bundled modules are left out of the configuration I have to know they are there in order to disable them again.

exist-core/src/test/java/org/exist/util/SchemaVersionSyncTest.java
```diff
@@ -0,0 +1,89 @@
...
+public class SchemaVersionSyncTest {
+
+    private static final Map<String, String> SCHEMA_FILE_TO_CONSTANT = new LinkedHashMap<>();
+    static {
+        SCHEMA_FILE_TO_CONSTANT.put("conf.xsd", SchemaVersion.CONF);
+        SCHEMA_FILE_TO_CONSTANT.put("collection.xconf.xsd", SchemaVersion.COLLECTION_XCONF);
+        SCHEMA_FILE_TO_CONSTANT.put("descriptor.xsd", SchemaVersion.DESCRIPTOR);
+        SCHEMA_FILE_TO_CONSTANT.put("mime-types.xsd", SchemaVersion.MIME_TYPES);
+        SCHEMA_FILE_TO_CONSTANT.put("controller-config.xsd", SchemaVersion.CONTROLLER_CONFIG);
+    }
```
reinhapa: Could be written simpler by using `Map.of`
```suggestion
    private static final Map<String, String> SCHEMA_FILE_TO_CONSTANT = Map.of(
        "conf.xsd", SchemaVersion.CONF, //
        "collection.xconf.xsd", SchemaVersion.COLLECTION_XCONF, //
        "descriptor.xsd", SchemaVersion.DESCRIPTOR, //
        "mime-types.xsd", SchemaVersion.MIME_TYPES, //
        "controller-config.xsd", SchemaVersion.CONTROLLER_CONFIG);
```

reinhapa (changes requested):

dizzzz:
That said, I had similar ideas re module loading a long time ago. "plug the modules you only need".
Effectively nothing changes. If one adds a new module, it will be loaded, but the author can decide differently. In that respect I am happy with this change.

My original ideas can from a slightly different angle: will this make loading java modules from a XAR file more simple?

exist-core/src/main/java/org/exist/collections/MutableCollection.java
```diff
@@ -87,6 +103,17 @@ public class MutableCollection implements Collection {
     private static final Logger LOG = LogManager.getLogger(Collection.class);
     private static final int SHALLOW_SIZE = 550;
     private static final int DOCUMENT_SIZE = 450;
+    private static final String XML_SCHEMA_NS = "http://www.w3.org/2001/XMLSchema";
```
dizzzz: I'd prefer to use and extend exist-core/src/main/java/org/exist/Namespaces.java om stead of adding it here. JAXP.java had the same definition too, which shall be combined.

exist-core/src/main/java/org/exist/collections/MutableCollection.java
```diff
@@ -87,6 +103,17 @@ public class MutableCollection implements Collection {
...
+    private static final String XSD_1_1_NS = "http://www.w3.org/XML/XMLSchema/v1.1";
+
+    /**
+     * Caches, per namespace, whether the system catalog's grammar for that namespace needs an
+     * XSD 1.1-capable loader -- see {@link #resolveXsd11SchemaForNamespace(Resolver, String)}.
+     * Namespaces reaching this cache are catalog-registered (a finite, admin-controlled set, never
+     * attacker-influenced), so unlike {@link org.exist.validation.Xsd11SchemaDetection}'s
+     * location-driven cache, no per-Subject scoping/bounding is needed here.
+     */
+    private static final ConcurrentMap<String, Optional<Schema>> XSD11_SCHEMA_BY_NAMESPACE = new ConcurrentHashMap<>();
```
dizzzz: Shouldn't we use Cafeine for this, and make a shared/global cache for this? MutableCollection does not look a good place for this?

exist-core/src/main/java/org/exist/indexing/IndexManager.java
```diff
@@ -145,6 +142,9 @@ private AbstractIndex initIndex(final BrokerPool pool, final String id, final El
             }
             final AbstractIndex index = (AbstractIndex) clazz.newInstance();
             index.configure(pool, dataDir, config);
+            if (index.getIndexName() == null) {
+                index.setName(id);
```
dizzzz, line 146: what if if==null or "" or blank?

exist-core/src/main/java/org/exist/jetty/JettyStart.java
```diff
@@ -105,12 +115,23 @@ public class JettyStart implements LifeCycle.Listener {
...
     @GuardedBy("this") private int status = STATUS_STOPPED;
     @GuardedBy("this") private Optional<Thread> shutdownHookThread = Optional.empty();
+    @GuardedBy("this") private Optional<Server> server = Optional.empty();
     @GuardedBy("this") private int primaryPort = 8080;
     @GuardedBy("this") private boolean webAppStartedSuccessfully = false;
     @GuardedBy("this") private String webAppStartupFailureDetail = null;
 
     private final CopyOnWriteArrayList<JettyStartListener> jettyStartListeners = new CopyOnWriteArrayList<>();
 
+    /**
```
dizzzz, line 125: functional way this should be in a different PR>

dizzzz (changes requested):
it looks that this PR combines 2 or 3 PRs (which makes the PR more difficult to review than strictly needed).

duncdrum:
These are stacked PRs that all target `develop` as their GitHub base — so GitHub shows the full accumulated diff of the entire chain (schema governance → XSD 1.1 validation → fixture codegen → conf `@enabled` → module SPI) rather than just the incremental commits. The actual changes belonging to *this* PR are the commits in `dp-module-spi` above `dp-conf-enabled-attr`. The `MutableCollection.java` and `JettyStart.java` changes you see belong to `dp-xsd11-validation` (#6530) — they'll disappear from this diff once the earlier PRs merge.

**MutableCollection.java:106 (Namespaces.java)**
Good catch — `XML_SCHEMA_NS` was unused (callers already used `XMLConstants` directly), and `XSD_1_1_NS` is now `Namespaces.XSD_1_1_NS` (added to `Namespaces.java`). Both fixed in `f7f3463460` on `dp-xsd11-validation`.

**MutableCollection.java:116 (Caffeine / separate class)**
The design concern is valid. The `ConcurrentHashMap` is correct here (the namespace set is finite and admin-controlled, so no eviction is needed), but moving the cache and XSD resolution logic out of `MutableCollection` into a dedicated class is worthwhile cleanup. Deferred to a follow-up PR so it doesn't block this chain.

**IndexManager.java:146 (null/blank id)**
Fixed in `f52bf39b57`. `Configuration.configureIndexer()` now validates `factory.getDefaultId()` before constructing an `IndexModuleConfig` entry (logs WARN and skips), with a belt-and-suspenders guard in `IndexManager.initIndex()` as well.

**JettyStart.java:125 (different PR)**
Agreed — split to standalone #6572. Only `JettyStart.java` is touched in that commit and the fix is fully independent of the XSD work.

**On your question about XAR module loading:**
Not directly — `ServiceLoader.load(ModuleFactory.class, Configuration.class.getClassLoader())` only scans the main classpath, so a XAR's `META-INF/services` wouldn't be picked up automatically. But the abstraction is right: if XAR deployment registers factories with the `EXistClassLoader` (or a future `Configuration.registerModuleFactory()` hook), discovery would follow. This PR establishes the interface and wire-up pattern; XAR integration would be a natural extension.

exist-core/src/test/java/org/exist/util/SchemaVersionSyncTest.java
duncdrum: Done — switched to `Map.of` in `fe93edee63`.

reinhapa (approved):

exist-core/src/main/java/org/exist/util/ParametersExtractor.java
```diff
@@ -255,6 +255,10 @@ private static Properties parseProperties(final Node container, final String ele
             for (int i = 0; i < params.getLength(); i++) {
                 final Element param = ((Element) params.item(i));
 
+                if ("no".equalsIgnoreCase(param.getAttribute("enabled"))) {
```
dizzzz, line 258: "no" or "false" or both?

extensions/exquery/restxq/src/test/resources-filtered/conf-fixture.xsl
dizzzz, line 1: copy right can go

dizzzz (approved):

extensions/exquery/restxq/src/test/resources-filtered/conf-fixture.xsl
adamretter: No it can't be removed.

reinhapa: > No it can't be removed.

@adamretter I do not agree. It is completely different content and filename.  (Besides of the copyright header,  that is wrong for the new file  anyway and the sole reason for GIT take it as a "rename" in my opinion.

line-o:
@dizzzz you approved this PR but have open questions on #6530 which this PR will close when merged.

extensions/exquery/restxq/src/test/resources-filtered/conf-fixture.xsl
duncdrum: Yes the copy right is wrong, the `license:format` used the wrong template in some places. Good catch it ll need to get corrected earlier in the PR chain.

exist-core/src/main/java/org/exist/indexing/IndexManager.java
duncdrum: Handled — `id != null && !id.isBlank()` covers null, empty, and whitespace-only together; see the guard right above this line (`21cd10a8cf`, renumbered from `11d1c929d9` by today's rebase).

exist-core/src/main/java/org/exist/util/ParametersExtractor.java
duncdrum: "no" only — `@enabled`'s type in `conf.xsd` is `yes_no`, an enumeration restricted to exactly `yes`/`no` (not `xs:boolean`), so `"false"` isn't a valid value to begin with. All 5 call sites (4 in `Configuration.java`, this one) check the same literal for that reason.

exist-core/src/main/java/org/exist/collections/MutableCollection.java
duncdrum: Already addressed upstream — `XSD_1_1_NS` moved to `Namespaces.java` and the duplicate `JAXP.java` definition dropped in `0fe1766a96` (PR-C, #6530). This branch picked that up in today's rebase.

exist-core/src/main/java/org/exist/collections/MutableCollection.java
duncdrum: Done rather than deferred — split into two package-private classes in `af2333f396`: `Xsd11SchemaCache` (Caffeine-backed; no eviction policy, since the namespace set is finite and admin-controlled, which was your own point) holds the cache, and `Xsd11ValidationHelper` holds `resolveXsd11SchemaForNamespace`, the dynamic-discovery schema holder, `validateWithXsd11Schema`, and the lexical-handler forwarder. `MutableCollection` is down to a one-line delegating call. All 63 tests in the validation/XSD11 suite pass unchanged. This branch already picked up the fix in today's rebase.

extensions/exquery/restxq/src/test/resources-filtered/conf-fixture.xsl
duncdrum: @adamretter @reinhapa — confirmed the root cause: `conf-fixture.xsl`/`controller-config-fixture.xsl` are new eXist-authored codegen tooling, not modifications of the original BSD-3 conf.xml/controller-config.xml, so they shouldn't have inherited the module's BSD-3 default. Fixed in PR-D (`9602881d89`) — both now carry the standard eXist LGPL-21 header, `license:check` passes. This branch picked the fix up in today's rebase.

PR #6564, opened 2026-07-08
[ci] Post XQTS result comparison as a PR comment

line-o:
### Description:

Lifts the XQTS result comparison into an explicitly invoked maven goal in `exist-xqts` and posts its output back onto the pull request as a single, updating comment — implementing what was [agreed with @duncdrum](https://github.com/eXist-db/exist/pull/6564#issuecomment-5067294838).

**The `xqts-compare` goal**

```bash
mvn -pl exist-xqts exec:exec@xqts-compare \
    -Dprevious-result=/path/to/previous -Dcurrent-result=/path/to/current
```

- Not bound to any lifecycle phase (`phase=none`, the same pattern as the existing `xml:transform@schema-governance` goal) — it only runs when invoked explicitly, in CI or locally.
- Takes two result folders as written by `exist-xqts-runner --output-dir`, with no assumption about where they come from. In CI that is the current run and the last successful `develop` run downloaded from the artifact store.
- Fails with an actionable message when an argument is missing or does not point to a comparable result folder.
- Runs Saxon from the module's own classpath (via `exist-core`), so the Saxon version can never drift from the build.
- Writes `comparison-results.xml` plus a GitHub-flavoured Markdown rendering `comparison-results.md` in one pass (default output: `exist-xqts/target/xqts-comparison`, overridable with `-Dcomparison-output-dir`).

**What changed**

- **`exist-xqts/src/main/xslt/compare.xslt`** (new) — entry point of the goal: validates the arguments, runs the comparison, emits XML + Markdown via `xsl:result-document`.
- **`exist-xqts/src/main/xslt/compare-results.xslt`** — required global params become template parameters so the entry point can drive it; comparison logic unchanged.
- **`exist-xqts/src/main/xslt/compare-results-md.xslt`** (new) — Markdown renderer: summary table with pass-% and trend markers, collapsible lists of newly failing/erroring/passing test cases, run labels as parameters instead of a hard-coded `develop`, and the runner-drift warnings from #6326 rendered as `> [!WARNING]` alerts.
- **`exist-xqts/pom.xml` / `exist-xqts/README.md`** — the exec-maven-plugin execution and its documentation.
- **`.github/workflows/ci-xqts.yml`** — replaces the raw Saxon invocations with the goal, appends the report to the run **Summary** tab (works for fork PRs), and uploads it with the PR number as the `xqts-comparison` artifact (PR events only). The build step installs the modules so the goal can resolve its dependencies.
- **`.github/workflows/ci-xqts-comment.yml`** (new) — a `workflow_run` companion that downloads the artifact and posts/updates a sticky comment.

**Why a separate `workflow_run` workflow?**

The XQTS workflow is triggered by `pull_request`, so for PRs opened from forks its `GITHUB_TOKEN` is read-only and cannot comment. A `workflow_run` workflow runs in the base-repository context with a write-enabled token, so the comment works for fork PRs too. It never checks out or executes any PR code — it only consumes the artifact produced by the trusted XQTS run, and is guarded to `workflow_run.event == 'pull_request'` && `conclusion == 'success'`. The main XQTS workflow keeps its `contents: read` permissions.

The comment uses `marocchino/sticky-pull-request-comment` with a fixed `header`, so re-runs update the existing comment in place rather than adding new ones.

### Reference:

https://github.com/eXist-db/exist/pull/6564#issuecomment-5067294838, #6326

### Type of tests:

CI-only change, verified locally against the project's Saxon-HE 12.5: the goal produces correct XML and Markdown (including drift warnings) from representative runner result folders, and all three failure modes (missing argument, unreadable path, folder without JUnit reports) fail the build with actionable messages. Both workflow YAML files validate, `license:check` passes. End-to-end behaviour (artifact hand-off and comment posting) exercises only on GitHub Actions once merged.

### DIFF
--- a/.github/workflows/ci-xqts.yml
+++ b/.github/workflows/ci-xqts.yml
@@ -23,7 +23,9 @@ jobs:
         timeout-minutes: 25
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: mvn -V -B --no-transfer-progress clean package -DskipTests -Ddependency-check.skip=true --projects exist-xqts --also-make
+        # `install` rather than `package` so that the xqts-compare goal below can
+        # resolve the exist-xqts dependencies from the local repository.
+        run: mvn -V -B --no-transfer-progress clean install -DskipTests -Ddependency-check.skip=true --projects exist-xqts --also-make
       - name: Run XQTS
         timeout-minutes: 60
         env:
@@ -73,17 +75,34 @@ jobs:
             "$(cat /tmp/previous_run_id.txt)" \
             --name xqts-logs \
             --dir /tmp/previous-xqts-output
-      - name: Compare Previous and Current XQTS Logs
+      - name: Compare Previous and Current XQTS Results
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          java \
-          -jar ~/.m2/repository/net/sf/saxon/Saxon-HE/9.9.1-8/Saxon-HE-9.9.1-8.jar \
-          -xsl:exist-xqts/src/main/xslt/compare-results.xslt \
-          -it:compare-results \
-          -o:/tmp/comparison-results.xml \
-          xqts.previous.junit-data-path=/tmp/previous-xqts-output/junit/data \
-          xqts.current.junit-data-path=/tmp/xqts-output/junit/data
+          mvn -B --no-transfer-progress -pl exist-xqts exec:exec@xqts-compare \
+            -Dprevious-result=/tmp/previous-xqts-output \
+            -Dcurrent-result=/tmp/xqts-output \
+            -Dcomparison-output-dir=/tmp/xqts-comparison \
+            -Dprevious-label='`develop`' \
+            -Dcurrent-label='this run'
       - name: Show Comparison Results
-        run: cat /tmp/comparison-results.xml
+        run: cat /tmp/xqts-comparison/comparison-results.xml
+      - name: Publish Comparison Summary
+        # Surface the report on the run's Summary tab (works for fork PRs too)
+        run: cat /tmp/xqts-comparison/comparison-results.md >> "$GITHUB_STEP_SUMMARY"
+      - name: Save PR Number
+        if: github.event_name == 'pull_request'
+        run: echo "${{ github.event.pull_request.number }}" > /tmp/xqts-comparison/pr-number.txt
+      - name: Upload XQTS Comparison
+        # Consumed by the ci-xqts-comment.yml workflow_run, which posts the
+        # comparison as a PR comment with a write-enabled token (needed for
+        # PRs from forks, where this workflow's token is read-only).
+        if: github.event_name == 'pull_request'
+        uses: actions/upload-artifact@v7
+        with:
+          name: xqts-comparison
+          retention-days: 3
+          path: /tmp/xqts-comparison
       - name: Archive XQTS Logs
         if: always()
         uses: actions/upload-artifact@v7
--- a/exist-xqts/src/main/xslt/compare-results-md.xslt
+++ b/exist-xqts/src/main/xslt/compare-results-md.xslt
@@ -1,0 +1,184 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    eXist-db Open Source Native XML Database
+    Copyright (C) 2001 The eXist-db Authors
+
+    info@exist-db.org
+    http://www.exist-db.org
+
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2.1 of the License, or (at your option) any later version.
+
+    This library is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+    Lesser General Public License for more details.
+
+    You should have received a copy of the GNU Lesser General Public
+    License along with this library; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+-->
+<!--
+    Renders the XML produced by compare-results.xslt into a GitHub-flavoured
+    Markdown report suitable for posting as a pull-request comment.
+-->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+    xmlns:xs="http://www.w3.org/2001/XMLSchema"
+    xmlns:cr="http://exist-db.org/exist-xqts/compare-results"
+    exclude-result-prefixes="xs cr"
+    version="2.0">
+
+    <xsl:output method="text" encoding="UTF-8"/>
+
+    <!-- Maximum number of test cases listed per detail section -->
+    <xsl:param name="max-listed" as="xs:integer" select="50"/>
+
+    <!-- Display names for the two compared runs, e.g. "`develop`" and "this run" -->
+    <xsl:param name="previous-label" as="xs:string" select="'previous'"/>
+    <xsl:param name="current-label" as="xs:string" select="'current'"/>
+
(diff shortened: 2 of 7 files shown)


line-o:
This is what is uploaded as an artifact (and after this PR has landed on develop will then be posted to each follow-up PR)

## 📊 XQTS result comparison

Comparison of this run against the latest successful XQTS run on `develop`.

| Metric | `develop` | This run | Change |
| :--- | ---: | ---: | ---: |
| 🟢 Passed | 28,840 (90.65%) | 28,841 (90.65%) | +1 (+0.00 pp) |
| 🟢 Failures | 1,582 | 1,581 | -1 |
| ➖ Errors | 134 | 134 | 0 |
| ➖ Skipped | 1,259 | 1,259 | 0 |
| 🧪 Total tests | 31,815 | 31,815 | 0 |

Relative to `develop`: **1** newly passing, **0** newly failing, **0** new errors, **0** newly skipped.

<details>
<summary>🟢 Newly passing tests (1)</summary>

- `Constr-inscope-4`

</details>

line-o:
The fact that it reports a fixed test-cases indicates a flaky test. By expanding the details section you can check which one that is.

duncdrum (changes requested):
see #6332  we should merge that first, and adjust the md table to work with the new output

I don't understand why compare-results-md.xsl is committed to source, but the ouput generation not part of maven. Can I run this xsl locally as part of my build, how do I run it? 

Having maven take care of the transform, safes you all the Saxon and xmlresolver tricks in the workflow. Phase `none` takes care of normal build operations. 

if you don't want it as part of maven, the CI only script should live inside `.github/` imv. But my clear preference is to simplify this by going the maven route. 

`🟢 Failures` looks odd, shouldn't this be red?

line-o:
> I don't understand why compare-results-md.xsl is committed to source, but the ouput generation not part of maven. Can I run this xsl locally as part of my build, how do I run it?

In the same way the transformation that produces the input `exist-xqts/src/main/xslt/compare-results.xslt` is part of the codebase and also not part of a maven run.
1. I built on an approach that exists
2. since both scripts have to compare with a previous result where should that come from?

I strongly disagree this should wait for #6332 as it is blocked by https://github.com/eXist-db/exist-xqts-runner/pull/51

line-o:
> 🟢 Failures looks odd, shouldn't this be red?

-1 failures is good

joewiz:
@line-o Thanks for pointing me at https://github.com/eXist-db/exist-xqts-runner/pull/51! I've reviewed its status and updated the description to indicate that it — and thus https://github.com/eXist-db/exist/pull/6332 — are both good to go. I had already dislodged eXist-db/exist-xqts-runner#51 from its original dependency on a qt4-bound branch, but I had not posted a comment to that effect or lifted the draft status. Apologies for having left it in that indeterminate state - and thanks again for flagging it!

reinhapa (approved):

dizzzz (approved):

dizzzz:
@duncdrum has specific concerns regarding this PR, e.g. if the referenced PR should be pulled in first? Lets discuss

dizzzz:
@duncdrum @line-o what is the state and conclusion?

line-o:
Yesterday @duncdrum agreed to consider my answers

dizzzz:
@duncdrum @line-o stil no final verdict?

duncdrum:
XQTS#51 is in so no more block. The correctness PR has been waiting a lot longer then this, so out of fairness I'm still in favour or merging it first. It has been waiting on the OP.

I spend a lot of time cleaning up maven, XQTS execution is part of maven, this script should be part of maven. The jar extraction dance is superfluous, if prior art can be cleaned up along the way, all the better.

line-o:
@duncdrum 
1. merging this or #6332 will not introduce conflicts for the other - they are _independent_ of each other
2. you are blocking this PR because it does not address issues that are existing in #6332 as well. Why is this PR different?

duncdrum:
@line-o im blocking because of maven, the other stuff is just bad form imv

line-o:
When I am fixing what you perceive as a problem so dire that it blocks this PR it will introduce a conflict with 6332 - I just want to make that clear

line-o:
Where it seems to be no issue at all

duncdrum:
which is why 6332 should go in first, not sure why I keep repeating myself.

.github/workflows/ci-xqts.yml
```diff
@@ -73,17 +73,52 @@ jobs:
             "$(cat /tmp/previous_run_id.txt)" \
             --name xqts-logs \
             --dir /tmp/previous-xqts-output
+      - name: Resolve Project Saxon Classpath
```
line-o: Is this the problem with the jar unpacking "dance"?

.github/workflows/ci-xqts.yml
```diff
@@ -73,17 +73,52 @@ jobs:
...
+          saxon_version="$(mvn -q help:evaluate -Dexpression=saxon.version -DforceStdout)"
+          xmlresolver_version="$(mvn -q help:evaluate -Dexpression=xmlresolver.version -DforceStdout)"
+          echo "Using Saxon-HE ${saxon_version} with xmlresolver ${xmlresolver_version}"
+          m2="${HOME}/.m2/repository"
+          echo "SAXON_CP=${m2}/net/sf/saxon/Saxon-HE/${saxon_version}/Saxon-HE-${saxon_version}.jar:${m2}/org/xmlresolver/xmlresolver/${xmlresolver_version}/xmlresolver-${xmlresolver_version}.jar" >> "$GITHUB_ENV"
       - name: Compare Previous and Current XQTS Logs
         run: |
           java \
-          -jar ~/.m2/repository/net/sf/saxon/Saxon-HE/9.9.1-8/Saxon-HE-9.9.1-8.jar \
+          -cp "$SAXON_CP" net.sf.saxon.Transform \
```
line-o: We can alternatively hardcode saxon 12 here

.github/workflows/ci-xqts.yml
```diff
@@ -73,17 +73,52 @@ jobs:
...
           -o:/tmp/comparison-results.xml \
           xqts.previous.junit-data-path=/tmp/previous-xqts-output/junit/data \
           xqts.current.junit-data-path=/tmp/xqts-output/junit/data
       - name: Show Comparison Results
         run: cat /tmp/comparison-results.xml
+      - name: Render Comparison as Markdown
+        run: |
+          mkdir -p /tmp/xqts-comparison
+          java \
+          -cp "$SAXON_CP" net.sf.saxon.Transform \
```
line-o: and here

exist-xqts/src/main/xslt/compare-results-md.xslt
```diff
@@ -0,0 +1,175 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
line-o, line 1: Let me get this straight:
Tthis XSLT file needs to be part of maven, while the existing XSLT scritps sitting next to it that serve the same purpose and are also used exclusively within CI can stay as they are?

line-o:
After a discussion with @duncdrum I agreed to enhance the XQTS result comparison and lift it into a maven goal within the xqts runner project in eXist-db.

We agreed on the following:
- the maven goal will not run by default but has to be called explicitly (similar to other goals we already have)
- this goal will be called in CI but can also run in other setups (e.g. as part of local development)
- the goal needs two arguments `current-result` and `previous-result`; both are paths to result folders that are output by the runner project
- the goal will fail with an actionable result if one or both of the two arguments are missing or point to something that cannot be compared
- there is explicitly no assumption made where those results are from - in CI this will be the current run in the output folder and the last run from develop read from our artifact store on GitHub

.github/workflows/ci-xqts.yml
duncdrum: yes, no need to resolve anything external if its part of maven.

.github/workflows/ci-xqts.yml
duncdrum: we won't need to hardcode anything on CI, the xml task will be configured to use or not use the currently supported version of Saxon.

.github/workflows/ci-xqts.yml
duncdrum: s.a.

duncdrum (approved):
thx @line-o much cleaner, good to merge once CI has finished

PR #6565, opened 2026-07-09
[ignore] only ignore the contents of .moderne/ (allows negations)

line-o:
`.moderne/` excludes the directory, and git never descends into an excluded directory, so the negations beneath it can never fire ([gitignore(5)](https://git-scm.com/docs/gitignore): *"It is not possible to re-include a file if a parent directory of that file is excluded."*). This hides new files under `.moderne/context/` from `git status` and makes `git add .moderne/context/` exit 1, even for tracked files. The contents form `.moderne/*` excludes the children individually, freeing the negations to re-include; `build/`, `run/`, `git-apply.log` and loose files under `.moderne/` stay ignored.

The prethink workflow is not broken — it has been committing new context files all along (79b280b08, 4ab5ba6a3) — so this fixes local use, not CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,7 @@ work/
 # Claude planning files
 plans/
 .codacy/
-.moderne/
+.moderne/*
 !.moderne/context/
 !.moderne/moderne.yml
 .github/instructions/


duncdrum:
looking at my local runs and the prethink CI job, this seems to be working as intended. What command did you run that failed to put something into git, that should actually be there?

reinhapa (approved):

line-o:
Turns out that `mod git apply` does more magic. But With this PR in we are also able to make changes to moderne.yml without having to git add -f

dizzzz (approved):

PR #6572, opened 2026-07-13
Fix Jetty shutdown hang and BrokerPool read-only race

duncdrum:
## Summary

- Sets a Jetty stop timeout so shutdown does not block indefinitely on active connections
- Moves `ShutdownListenerImpl.shutdown()` to a daemon thread to avoid deadlocking the Jetty shutdown-hook thread that also holds the BrokerPool write lock

This bugfix was discovered during XSD 1.1 validation integration work but is completely independent of it — JettyStart.java is the only file touched.

## What Changed

**`exist-core/src/main/java/org/exist/jetty/JettyStart.java`**
- Add `serverStopRequested` `AtomicBoolean` guard so concurrent shutdown paths (direct `shutdown()`, `ShutdownListenerImpl`, and `BrokerPoolAndJettyShutdownHook` JVM hook) do not race
- Dispatch `ShutdownListenerImpl.shutdown()` on a daemon thread to avoid the BrokerPool write-lock / Jetty-shutdown-hook deadlock
- Set a Jetty stop timeout to bound the shutdown wait

## Test Plan

- [ ] Start eXist-db and trigger `system:shutdown()` from XQuery — server stops cleanly without hang
- [ ] Send SIGTERM — JVM hook fires, Jetty stops within the stop timeout
- [ ] Trigger shutdown from XML-RPC API — no deadlock with BrokerPool write lock

### DIFF
--- a/exist-core/src/main/java/org/exist/jetty/JettyStart.java
+++ b/exist-core/src/main/java/org/exist/jetty/JettyStart.java
@@ -56,10 +56,20 @@ import java.io.Reader;
 import java.net.*;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Observer;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.ServiceLoader;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


dizzzz:
Thnx for isolating this in a separate PR! It is complex and significant enough!

dizzzz (approved):
copilot feedback

The PR is solid, but a few optional enhancements could strengthen it:

**1. Consider documenting the timeout constants:** The 30-second Jetty stop timeout and 5-second catch-up margin are sensible but somewhat arbitrary. Add a comment explaining the rationale (e.g., "30s allows graceful drain of in-flight requests; 5s is a catch-up margin for lifecycle callbacks").

**2. Daemon thread naming could be more specific:** The thread name "JettyShutdown" is clear, but consider including context like "JettyShutdown-async" or a sequence number to aid debugging in scenarios with multiple shutdown events.

**3. Consider adding a log statement in stopServerAndAwait():** When serverStopRequested.compareAndSet() returns false, it would be helpful to log that shutdown is already in progress (at debug level) to aid troubleshooting.

line-o (approved):

Issue #6578, opened 2026-07-15
[BUG] auto-parsing of multipart-formdata is incomplete


dariok:
Tested on eXist 7.0.0-Beta3 (from Docker).
Similar in eXist 6.4.1.

## Problem

A standard multipart-upload might look like this (file upload ex Chrome):

```
------WebKitFormBoundaryFB2YRAxh8THShzW6
Content-Disposition: form-data; name="file"; filename="bwc-0-00.xml"
Content-Type: text/xml


------WebKitFormBoundaryFB2YRAxh8THShzW6
Content-Disposition: form-data; name="path"

edition/bwc-0-00.xml
------WebKitFormBoundaryFB2YRAxh8THShzW6--
```

eXist tries to parse this; I couldn’t find anything in the documentation (`https://exist-db.org/exist/apps/doc/search.html?q=multipart&field=all&view=summary&action=&prev=` returns 0 results) where this is supposed to happen, but I can see that it must be during the handover from `controller.xql`:

```
  <dispatch xmlns="http://exist.sourceforge.net/NS/exist">
    <forward url="{$exist:controller}/rest2/api.xq">
      {
        util:log("info", request:get-parameter-names() => string-join(' - ')),
        (: we currently need this workaround here as Jetty snatches all OPTIONS requests before they can be parsed by roaster :)
        if ( request:get-method() = ('options', 'OPTIONS') and $local:config//config:origin = request:get-header('origin') )
        then (
          <set-header name="Access-Control-Allow-Origin" value="{ request:get-header('origin') }" />,
          <set-header name="Access-Control-Allow-Methods" value="GET, PUT, POST, PATCH, OPTIONS, HEAD" />,
          <set-header name="Access-Control-Allow-Headers" value="Authorization, Content-Type" />,
          <set-header name="Access-Control-Allow-Credentials" value="true" />
        )
        else ()
      }
    </forward>
  </dispatch>
```
logs
`15 Jul 2026 13:10:43,511 [qtp1424067142-689] INFO  (LogFunction.java [writeLog]:204) - (Line: 63 /db/apps/edoc/controller.xql) path - meta `

while

```
util:log("info", request:get-parameter-names() => string-join(' - ')),
roaster:route($api:definitions, api:lookup#1, $api:use)
```

in `api.xq` logs

[...]

PR #6581, opened 2026-07-16
[bugfix] Parse multipart/form-data for all HTTP methods, not only POST

### LINKED ISSUE #6580: [BUG] parsing of multipart does not work correctly for PUT
(Tested on eXist-7.0.0-SNAPSHOT as of yesterday)

## Problem
Given a multipart upload like the following

```
Content-Type: multipart/form-data; boundary=wdb

--wdb
Content-Disposition: form-data; name="path"

edition/01/17410105.xml
--wdb
Content-Disposition: form-data; name="file"; filename="17410105.xml"
Content-Type: application/xml

example.xml
--wdb--
```

When `POST`ing, I get both fields, "file" and "path" as parameters.
When `PUT`ing, I only get "path", while "file" is missing.

## Expected outcome

For all verbs that accept a multipart upload, so at least POST and PUT, a complete representation of the form-data must be available to parse downstream.

joewiz:

## Summary

Multipart/form-data uploads only worked for `POST`. A multipart `PUT` (or `PATCH`) was effectively unusable: `request:is-multipart-content()` returned `false`, the uploaded file part was silently dropped, and the request stream was consumed so nothing downstream could recover it. This fixes multipart parsing so it works for every HTTP method that carries a multipart body, and so uploaded files are exposed to controllers and RESTXQ resource functions (via the `request:` module) regardless of method.

Closes #6580. Substantially addresses #6578 (see *Scope* below).

## Root cause — two coupled defects

1. **`HttpRequestWrapper` gated multipart detection on `POST`.** `isMultipartContent` was only ever `true` when `getMethod()` was `POST`, so for `PUT`/`PATCH` the request was never treated as multipart: `request:is-multipart-content()` returned `false` and `getParts()` was never called.

2. **`XQueryURLRewrite` built its request wrapper with `parseMultipart=false`.** On the controller and RESTXQ path — the path a `PUT` can actually reach, since REST `PUT` *stores* rather than executes — the wrapper fell back to the servlet container's `getParameterMap()`. That exposes non-file form fields but omits file parts entirely, so an uploaded file was never visible there for *any* method, and reading the parameters consumed the body so it could not be re-parsed by hand either.

The two are coupled: fixing only (1) would have flipped `request:is-multipart-content()` to `true` without making the file available on that path — which would move consumers onto a code path that returns an empty file. Both are fixed together.

## What changed

- **`HttpRequestWrapper`** — recognize a multipart body by its `Content-Type` (`multipart/…`) regardless of the HTTP method, instead of requiring `POST`.
- **`XQueryURLRewrite`** — construct the request wrapper with `parseMultipart=true`, so multipart content (including file parts) is parsed via `getParts()` and exposed to controllers and RESTXQ resource functions through the `request:` module.

## Reproduction / test

`MultipartMethodControllerTest` stores a `controller.xql` that reports, via the `request:` module, what it sees of an identical multipart body sent as `POST` and as `PUT` — the same path a third-party router (e.g. the roaster library, eeditiones/roaster#107) takes. Before the fix, `PUT` reported `is-multipart-content()=false` and exposed neither the file field nor the uploaded file; the controller path never exposed the uploaded file for either method. After the fix, both methods see the form field, the uploaded file, and `is-multipart-content()=true`.

Behavior before vs. after, from that test (controller path):

| | `is-multipart-content()` | form field `path` | uploaded file `file` |
|---|---|---|---|
| POST — before | true | ✓ | ✗ |
| PUT — before | **false** | ✓ | ✗ |
| POST — after | true | ✓ | **✓** |
| PUT — after | **true** | ✓ | **✓** |

## Scope

- **#6580** — fully fixed.
- **#6578** — the core defect (uploaded files not parsed for non-`POST` methods / on the controller and RESTXQ path) is fixed: files are now parsed and available through the `request:` module for every method, so the "parse the raw body yourself" workaround is no longer needed. One aspect it deliberately does not add is a way to read an *individual part's own headers* (e.g. a file part's `Content-Type`) through the `request:` module — that would be a small additive API and is better discussed separately. Marked "Relates to" rather than "Closes" #6578 for that reason; happy to close it if reviewers consider the remaining item out of scope.
- Makes the client-side multipart-parsing workaround in eeditiones/roaster#107 unnecessary once released.

## Test plan

- [x] `MultipartMethodControllerTest` (new) — multipart `POST` and `PUT` through a `controller.xql`, asserting the field, the file, and `is-multipart-content()`.
- [x] `org.exist.http.urlrewrite.*` and `org.exist.xquery.functions.request.*` — full packages green (incl. `ControllerTest`, `GetParameterTest`, `PatchTest`, `GetData*Test`).
- [x] `RESTServiceTest` green.
- [x] Codacy (PMD) on the changed files — no new findings.

### DIFF
--- a/exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
@@ -62,11 +62,11 @@ public class HttpRequestWrapper implements RequestWrapper {
 
     private static final Logger LOG = LogManager.getLogger(HttpRequestWrapper.class);
 
-    // HTTP methods whose request may legitimately carry a multipart/form-data body. GET, HEAD and
-    // DELETE are deliberately excluded: a body on those methods has no defined semantics (RFC 9110
-    // §9.3.1), and parsing one would expose attacker-suppliable multipart content (including file
-    // uploads) to handlers not written to expect a body.
-    private static final Set<String> MULTIPART_METHODS = Set.of("POST", "PUT", "PATCH");
+    // HTTP methods whose request may legitimately carry a request body. GET, HEAD and DELETE are
+    // deliberately excluded: a body on those methods has no defined semantics (RFC 9110 §9.3.1),
+    // and parsing one would expose attacker-suppliable multipart content (including file uploads)
+    // to handlers not written to expect a body.
+    private static final Set<String> METHODS_WITH_REQUEST_BODY = Set.of("POST", "PUT", "PATCH");
 
     private static final Path TMP_DIR;
     static {
@@ -144,10 +144,10 @@ public class HttpRequestWrapper implements RequestWrapper {
         // Determine if request is a multipart. A multipart/form-data body may accompany any
         // body-carrying method (e.g. PUT and PATCH, not just POST), so recognize the whole
         // allow-list rather than POST alone — while still excluding GET/HEAD/DELETE (see
-        // MULTIPART_METHODS). See https://github.com/eXist-db/exist/issues/6580 and
+        // METHODS_WITH_REQUEST_BODY). See https://github.com/eXist-db/exist/issues/6580 and
         // https://github.com/eXist-db/exist/issues/6578
         @Nullable final String contentType = servletRequest.getContentType();
-        isMultipartContent = MULTIPART_METHODS.contains(servletRequest.getMethod().toUpperCase(Locale.ENGLISH))
+        isMultipartContent = METHODS_WITH_REQUEST_BODY.contains(servletRequest.getMethod().toUpperCase(Locale.ENGLISH))
                 && contentType != null && contentType.toLowerCase(Locale.ENGLISH).startsWith("multipart/");
 
         // Get multi-part formdata parameters when it is a mpfd request
--- a/exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
@@ -135,10 +135,13 @@ public class HttpRequestWrapper implements RequestWrapper {
         this.pathInfo = servletRequest.getPathInfo();
         this.servletPath = servletRequest.getServletPath();
 
-        // Determine if request is a multipart
-
+        // Determine if request is a multipart.
+        // Any HTTP method may carry a multipart/form-data body (e.g. PUT and PATCH,
+        // not just POST); the presence of the multipart content type is what matters.
+        // See https://github.com/eXist-db/exist/issues/6580 and
+        // https://github.com/eXist-db/exist/issues/6578
         @Nullable final String contentType = servletRequest.getContentType();
-        isMultipartContent = "POST".equalsIgnoreCase(servletRequest.getMethod()) && contentType != null && contentType.toLowerCase(Locale.ENGLISH).startsWith("multipart/");
+        isMultipartContent = contentType != null && contentType.toLowerCase(Locale.ENGLISH).startsWith("multipart/");
 
         // Get multi-part formdata parameters when it is a mpfd request
         // and when instructed to do so
--- a/exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
@@ -62,6 +62,12 @@ public class HttpRequestWrapper implements RequestWrapper {
 
     private static final Logger LOG = LogManager.getLogger(HttpRequestWrapper.class);
 
+    // HTTP methods whose request may legitimately carry a multipart/form-data body. GET, HEAD and
+    // DELETE are deliberately excluded: a body on those methods has no defined semantics (RFC 9110
+    // §9.3.1), and parsing one would expose attacker-suppliable multipart content (including file
+    // uploads) to handlers not written to expect a body.
+    private static final Set<String> MULTIPART_METHODS = Set.of("POST", "PUT", "PATCH");
+
     private static final Path TMP_DIR;
     static {
         try {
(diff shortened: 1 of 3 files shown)


dizzzz (approved):

exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
```diff
@@ -135,10 +135,13 @@ public HttpRequestWrapper(final HttpServletRequest servletRequest,
...
-        // Determine if request is a multipart
-
+        // Determine if request is a multipart.
+        // Any HTTP method may carry a multipart/form-data body (e.g. PUT and PATCH,
+        // not just POST); the presence of the multipart content type is what matters.
+        // See https://github.com/eXist-db/exist/issues/6580 and
+        // https://github.com/eXist-db/exist/issues/6578
         @Nullable final String contentType = servletRequest.getContentType();
-        isMultipartContent = "POST".equalsIgnoreCase(servletRequest.getMethod()) && contentType != null && contentType.toLowerCase(Locale.ENGLISH).startsWith("multipart/");
+        isMultipartContent = contentType != null && contentType.toLowerCase(Locale.ENGLISH).startsWith("multipart/");
```
line-o: This drops the `POST`-only gate entirely instead of replacing it with an explicit allow-list. Confirmed live: a `GET` request with a `multipart/form-data` body now gets `is-multipart-content()=true` and its uploaded files fully parsed and exposed via `request:get-uploaded-file-name()`, on both the `RESTServer` and `XQueryURLRewrite` paths — previously `GET` never exposed uploaded files at all.

Per RFC 9110 §9.3.1, a `GET` request body has no defined semantics and a client SHOULD NOT send one absent prior agreement with the origin server. Suggest scoping this to an explicit set of methods that carry a body (`POST`, `PUT`, `PATCH`) rather than every method, e.g.:

```java
private static final Set<String> MULTIPART_METHODS = Set.of("POST", "PUT", "PATCH");
...
isMultipartContent = MULTIPART_METHODS.contains(servletRequest.getMethod().toUpperCase(Locale.ENGLISH))
        && contentType != null && contentType.toLowerCase(Locale.ENGLISH).startsWith("multipart/");
```

line-o (changes requested):
I built this branch and tested it live against both the direct REST API path (`RESTServer`) and the controller/RESTXQ path (`XQueryURLRewrite`). The PUT/PATCH fix itself works as described, but it has a side effect: `GET` requests are now treated as multipart too.

Before this PR, a `GET` request carrying a `multipart/form-data` body already leaked non-file form fields (a pre-existing, unrelated quirk of the servlet container's parameter map), but `request:is-multipart-content()` reported `false` and uploaded files were silently dropped. After this PR, `GET` is handled identically to `POST`/`PUT`: `is-multipart-content()` returns `true` and uploaded files become fully visible via `request:get-uploaded-file-name()`, on both paths I tested.

`GET` is specified as a safe method with no defined semantics for a request body. RFC 9110 §9.3.1 ("GET") is explicit about this: a client SHOULD NOT generate content in a GET request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported — since intermediaries along the request chain are often unaware of such private agreements. (The predecessor spec, RFC 7231 §4.3.1, put it more bluntly: "A payload within a GET request message has no defined semantics; sending a payload body on a GET request might cause some existing implementations to reject the request.")

Requesting changes: rather than dropping the method check entirely, could `isMultipartContent` use an explicit allow-list of methods that legitimately carry a body (`POST`, `PUT`, `PATCH`), or otherwise explicitly exclude `GET` (and `HEAD`/`DELETE`)? That keeps the PUT/PATCH fix intact without also making `GET` a vector for parsed, attacker-suppliable multipart content (including file uploads) on every RESTXQ/controller function bound to it — which is a meaningful change of trust boundary for any handler that wasn't written expecting a body on `GET`.

See inline comment on the specific line.

exist-core/src/main/java/org/exist/http/servlets/HttpRequestWrapper.java
joewiz: @line-o Great, thanks. Work on these changes is in progress.

joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Thanks @line-o — good catch, and agreed. Treating `GET` as multipart is a real change of trust boundary, and dropping the method gate entirely was too broad. Fixed in d836b113d5 with your suggested approach: an explicit allow-list of body-carrying methods.

```java
private static final Set<String> MULTIPART_METHODS = Set.of("POST", "PUT", "PATCH");
...
isMultipartContent = MULTIPART_METHODS.contains(servletRequest.getMethod().toUpperCase(Locale.ENGLISH))
        && contentType != null && contentType.toLowerCase(Locale.ENGLISH).startsWith("multipart/");
```

So `GET`/`HEAD`/`DELETE` are excluded while the `PUT`/`PATCH` fix stays intact. Added a test alongside the existing one: a `GET` carrying a `multipart/form-data` body now reports `is-multipart-content()=false` and exposes no uploaded file, while `POST`/`PUT`/`PATCH` all parse it.

You're right that the pre-existing leak of non-file form fields into the container's parameter map on `GET` is separate — this change doesn't alter it, and it seemed out of scope here; happy to file it as its own issue if you think it's worth chasing.

line-o: hmm, why do we call this MULTIPART_METHODS rather then making it clear that these methods can carry a body.
How about `METHOD_CAN_CARY_A_BODY` and I would also propose to match against lower-case strings.
Do we want to make this an enum, @reinhapa?

joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Good points, @line-o.

**Name** — agreed, it should say *why* these methods qualify rather than what we use them for. I'd suggest `METHODS_WITH_REQUEST_BODY` (clear, and avoids the singular/plural mismatch). Happy to use whatever wording you prefer.

**Case** — the current check is already case-insensitive (`getMethod().toUpperCase(...)` against an uppercase set). I'd lean towards keeping uppercase: HTTP method names are canonically uppercase, and the existing method checks in this codebase compare against uppercase literals (e.g. `PathFilter` does `"PUT".equalsIgnoreCase(...)`), so an uppercase set reads most obviously as "these HTTP methods". But it's a one-line flip either way — glad to match lower-case if you'd rather.

**Enum** — over to @reinhapa, since it's really your call. For context: there's no existing HTTP-method enum in the codebase today; method dispatch is done with string comparisons throughout, so this would be net-new infrastructure for a three-element set. I don't feel strongly and am happy to add one if you'd prefer it.

I'll hold the rename until we've settled on the name and the enum question, so it lands in one clean change rather than renaming twice.

line-o: - Name: `METHODS_WITH_REQUEST_BODY` 👍🏼 
- Case: Agreed, uppercase it is
- Enum: let's do this in a followup (if at all)

joewiz:

Done — renamed to `METHODS_WITH_REQUEST_BODY`, kept the values uppercase, and left the enum for a possible follow-up. Landed as a separate commit on top so the delta is easy to read.

line-o (approved):
Nice!

Issue #6569, opened 2026-07-12
[feature] Enforce collection traverse (EXECUTE) on direct path access, matching Unix semantics


line-o:
## Is your feature request related to a problem?

Under Unix semantics you need the traverse bit (`x`) on **every** directory in a
path to reach a file inside it, regardless of the file's own permissions.
eXist does not enforce this for direct path access: if you know the exact path
of a resource, you can reach it without having `EXECUTE` (traverse) on the
containing collection(s). The document's own permissions are still checked, so
this is a **permissions-model deviation, not a data leak** — but it means
removing a collection's `EXECUTE` bit does not actually prevent access to
known-path resources inside it, contrary to what an operator would expect from
Unix-style modes.

The traverse checks are present in the code but commented out:

```java
// exist-core/.../storage/NativeBroker.java:2498  getXMLResource(uri, LockMode)
//if (!collection.getPermissions().validate(getCurrentSubject(), Permission.EXECUTE)) {
//    throw new PermissionDeniedException("Permission denied to read collection '" + collUri + "' by " + getCurrentSubject().getName());
//}
```
(also `:2471` in `getResource(uri, accessType)`).

Document-level authorization still runs — `getResource` validates the requested
mode at `:2486`, and `MutableCollection.getDocumentWithLock` re-imposes document
`READ` at `:725` — so access to the *document* is gated. What is missing is the
*ancestor-collection traverse* gate.

## Describe the solution you'd like

Enforce `EXECUTE` on each ancestor collection during path resolution, so
reaching `/db/a/b/doc` requires `EXECUTE` on `/db`, `/db/a`, and `/db/a/b`
(in addition to the document's own permissions). This mirrors Unix directory
traversal and makes collection modes behave as operators expect.

Because many existing deployments have permissive or inconsistent collection
modes, this is a behaviour change with real compatibility risk. Suggested
approach:

- Implement the traverse check in the shared resolution path (and reuse it from
  the execution-resolution boundary proposed in the execute-without-read issue,
  #6568).
- Gate it behind a `conf.xml` switch, e.g. `enforce-collection-traverse`,
  **default off**.
- Enable by default only after a deprecation cycle, with clear release notes.

## Describe alternatives you've considered

- **Leave as-is / document the behaviour.** Keeps the surprising divergence
  where a collection's `EXECUTE` bit has no effect on direct-path access.
- **Enforce unconditionally.** Rejected as a default — too likely to break
  existing apps that rely on the current permissive traversal.

[...]

PR #6586, opened 2026-07-18
[feature] Unix-style execute-without-read for stored XQuery

### LINKED ISSUE #6568: [feature] Unix-style execute-without-read for stored XQuery (EXECUTE should not imply READ)
## Is your feature request related to a problem?

eXist has a distinct `EXECUTE` permission bit and the XQuery execution
primitive already gates on it — `XQuery.execute` validates `EXECUTE`, not
`READ` (`exist-core/.../xquery/XQuery.java:360`), as do `XQueryPool.borrowCompiledQuery`
(`:173`) and the URL-rewrite cache (`XQueryURLRewrite.java:480`). setuid/setgid
is honoured, so a stored query can run as its owner (`XQuery.java:388-410`).

But in practice **you cannot run a stored query you cannot `READ`**, because
every path that loads the source to compile it acquires the document handle
through a getter that hardcodes a `READ` check:

```java
// exist-core/.../collections/MutableCollection.java:725  (getDocumentWithLock)
if (!doc.getPermissions().validate(broker.getCurrentSubject(), Permission.READ)) {
    unlockFn.run();
    throw new PermissionDeniedException("Permission denied to read + document: " + name);
}
```

That gate sits on every cold execution path: `RESTServer.doGet:507`,
`SourceFactory.getSource_fromDb:195`, `XQueryContext` import-module `:2771`,
`XQueryURLRewrite.getSource:905`, `Eval.java:537`. RESTXQ, XML-RPC and the Local
API go further and require `READ | EXECUTE`
(`ResourceFunctionExecutorImpl.java:258`, `RpcConnection.java:2006/2034`,
`LocalXPathQueryService.java:246`).

Net effect: the `EXECUTE` bit is largely aspirational, the model deviates from
Unix, and setuid — whose entire purpose is running privileged code you cannot
read — is only reachable by callers who *can* read the code.

The one place execute-without-read does work today is by accident:
`XQueryURLRewrite.getFromCache` (`:468-492`) reuses a cached compiled
`ModelAndView` and validates `EXECUTE` only, never re-reading the source — but it
depends on a `READ`-capable user having primed `urlCache` first, and has no
error-confidentiality story.

> A related but distinct deviation — the collection-`EXECUTE` (traverse) check on
> direct path access being commented out (`NativeBroker.java:2513`) — is tracked
> as its own issue (#6569), since it is orthogonal to this proposal.

## Describe the solution you'd like

Make stored-query permissions match Unix `--x` semantics:

| Unix | eXist | Required perm |
|------|-------|---------------|
| execute a `--x` binary | run a stored `.xq` / RESTXQ fn / controller | `EXECUTE` on the resource |
| kernel reads the binary for the process | DB compiles the source **on the caller's behalf** | no `READ` needed |
| a failed execute returns a generic `errno`, not file contents | failure → generic "execution failed" | — |
| `read()` the file | download, `?_source`, `fn:unparsed-text`, `util:binary-doc` | `READ` (unchanged) |
| setuid runs as owner | setuid query runs as owner | already implemented |

Key points:

1. **Compile-on-behalf needs no privilege elevation.** The byte read is already
   permission-free — `NativeBroker.getBinaryResource:2604` is a raw
   `blobStore.get`. The *only* thing gating source access is acquiring the
   document *handle*. So the fix is to add one execution-resolution boundary
   (`DBBroker.getResourceForExecution(uri, lockMode)`) that gates on `EXECUTE`
   (not `READ`), used by the query loaders. The generic `READ` getters stay as
   they are, so data reads (`fn:doc`, serialization, `util:*`, downloads) are
   untouched — no new way to read source *as data*.

2. **Error confidentiality for read-blind callers.** A caller with `EXECUTE` but
   not `READ` is told only that execution failed — one generic message + a
   correlation id, with **no** source text, message, spec error code,
   line/column, module URI or stack. The full error is logged server-side under
   the correlation id (with real + effective/setuid user) for the owner/DBA.
   Compile-fail, runtime-fail and execute-denied all look identical, which is
   strictly less informative than Unix `errno` and avoids compile-vs-runtime
   oracles. DBA callers are unaffected (`UnixStylePermission.validate`
   short-circuits for DBA).

3. **The compiled query is still cached; only the error-verbosity decision is
   not.** `--x` modules are compiled once and shared from the cache like any
   other — caching is unchanged. What must **not** be cached is the answer to
   "may *this* caller see error details?", because the same cached query is run
   by different users with different permissions. So on each execution (at the
   point that already re-validates `EXECUTE`) we cheaply recompute "does the
   current caller have `READ`?" and use it *only* to pick full-vs-generic error
   text. The cache stores the compiled program, not any per-user visibility.
   (Analogy: one executable in the OS page cache is shared by all processes;
   whether a given process gets a verbose crash dump depends on the process, not
   on the cache.) If the caller can't read the entry query, or any module it
   imports, the whole execution uses generic errors.

4. **Compilation runs under the caller's own subject** (not elevated), so
   static-analysis/constant-folding side effects can't escalate; setuid still
   applies only at `XQuery.execute`.

5. **Unify the entry points** on this model: REST, XQueryServlet, `import
   module`, `util:eval` (path form), URL-rewrite, RESTXQ, XML-RPC, Local API all
   resolve for execution via `getResourceForExecution` and route error writing
   through a single disclosure filter. RESTXQ/XML-RPC/Local drop from
   `READ | EXECUTE` to `EXECUTE`.

Suggested phasing:

- **Phase 1a — mechanism.** The `getResourceForExecution(uri, lockMode)`
  boundary (gates on `EXECUTE`, reports whether the caller can also `READ`); an
  `ErrorDisclosure` (`FULL`/`GENERIC`) flag on `XQueryContext` + a
  `disclose(ctx, XPathException)` filter with correlation-id logging + a new
  internal `ErrorCodes.EXXQERR0001 "Query execution failed."`; compute disclosure
  in `XQuery.execute`.
- **Phase 1b — switch the READ-based execution loaders to `EXECUTE`:**
  `RESTServer` (`:507`), `XQueryServlet` (exec path; keep `READ` for `?_source`),
  `import module` (`XQueryContext.java:2771`), `SourceFactory` exec variant
  (`:195`), `util:eval` path form (`Eval.java:537`), URL-rewrite
  (`XQueryURLRewrite.java:905` miss / `:480` hit).
- **Phase 1c — relax the `READ | EXECUTE` execution checks to `EXECUTE`:**
  RESTXQ `ResourceFunctionExecutorImpl.checkSecurity` (`:258`), XML-RPC
  `RpcConnection` (`:2006`, `:2034`), Local API `LocalXPathQueryService`
  (`:246`). Each recomputes disclosure per call and routes errors through
  `disclose`.
- All of the above behind a `conf.xml` `enforce-read-for-execute` switch
  (default off = new behaviour; on = legacy `READ`-required, forced `FULL`
  disclosure) for operators who relied on the accidental read-lockout; ship in a
  major (7.0).

Collection-`EXECUTE` (traverse) enforcement is **out of scope here** and tracked
as a separate issue (#6569).

## Describe alternatives you've considered

- **Do nothing / document as-is.** Leaves setuid effectively unusable for the
  read-blind case and the `EXECUTE` bit misleading.
- **Elevate to a system subject around the whole compile.** Rejected — broader
  privilege surface for compile-time side effects, and unnecessary since the
  byte read is already unchecked. Elevating only handle-acquisition (implicitly,
  by checking `EXECUTE` instead of `READ`) is enough.
- **Disclose compile errors but not runtime errors.** Rejected as the default:
  runtime `XPathException`s embed expression text and would leak the source.
  (Open question below.)

## Additional context

Confidentiality boundary is the **source text and failure internals** — the
query's normal computed output is not a leak (returning results is the point,
same as a setuid program). A query that deliberately echoes its own source is
the author's choice, out of scope, same as a setuid binary printing its own
image.

Open questions for maintainers:

- Sanitize **all** query errors for read-blind callers (compile + runtime), or
  compile-only?
- Optional DBA-only `system:get-execution-error($ref)` to fetch the full logged
  error by correlation id?
- Default timing: new behaviour default-on in 7.0, or default-off for one minor
  then flip?
- Treat path-based `util:eval` as "execute" (`EXECUTE`) or keep it `READ`?

I have a detailed implementation plan (resolution API signatures, exact edit
sites, disclosure component, test matrix, PR breakdown) and am happy to open the
Phase-1 PR if the semantics are agreed.

line-o:
### Description:

Makes stored-query permissions match Unix `--x` semantics: a query with `EXECUTE` but not `READ` can now actually be run, without the caller learning anything about its source from a failure.

eXist already has a distinct `EXECUTE` permission bit and `XQuery.execute` already gates on it, not `READ`. But every path that loaded the source to *compile* it hardcoded a `READ` check on the document handle, so `EXECUTE`-without-`READ` never actually worked — setuid queries (whose entire point is running code you can't read) were only reachable by callers who could already read them.

**What changed:**

- `DBBroker.getResourceForExecution(uri, lockMode)` — a new resolution boundary that gates on `EXECUTE`, not `READ`, and reports whether the caller can *also* `READ` the source (`NativeBroker`, `ExecutableResource`).
- `ErrorDisclosure` (new): `FULL`/`GENERIC` error-disclosure levels on `XQueryContext`. A caller who can execute but not read a query gets one generic message + a correlation id — no source text, error message, spec error code, line/column, or module URI. The full error is logged server-side under the correlation id for the owner/DBA. Recomputed on every execution (the compiled query is pooled and shared between users with different permissions), and applied to any failure type a transport can see — not just `XPathException` (serialization `SAXException`/`BadRequestException`, runtime `PermissionDeniedException`, `RuntimeException`) — since a failure surfacing during result serialization is just as informative as one from `execute()`.
- Wired through `RESTServer`, `XQueryServlet`, `SourceFactory`, `XQuery#execute` (compile + execute paths), and RESTXQ's `XQueryCompiler` (resource-function module resolution).
- Fail-closed throughout: an unresolvable/unknown subject is treated as unable to `READ`, not able to.
- Closes a resource-type/mime-type spoofing gap: both the binary resource type *and* the `application/xquery` mime type are now required before something is treated as an execute-without-read target.
- `X-XQuery-Cached` (reveals whether another user recently ran the shared compiled query) is suppressed for read-blind callers.
- Behind no new flag — the `READ` requirement is simply no longer imposed on the execution path; data-read paths (`fn:doc`, serialization, downloads, `?_source`) are untouched.

Collection-traverse (`EXECUTE` on ancestor collections for direct path access) is a separate, orthogonal deviation from Unix semantics and is tracked in #6569, not addressed here.

### Reference:

Closes #6568

### Type of tests:

New JUnit test classes:
- `org.exist.storage.GetResourceForExecutionTest` — the resolution boundary itself
- `org.exist.security.ExecuteWithoutReadTest` — permission semantics
- `org.exist.xquery.ErrorDisclosureTest` — the disclosure filter
- `org.exist.http.RESTExecuteWithoutReadTest` — end-to-end over REST, including the serialization-failure and cache-header-suppression regressions

All pass locally (`mvn -pl exist-core test -Dtest=...`), 28 tests across the four classes.

### DIFF
--- a/exist-core/src/main/java/org/exist/http/RESTServer.java
+++ b/exist-core/src/main/java/org/exist/http/RESTServer.java
@@ -1563,7 +1563,7 @@ public class RESTServer {
      * Get the document a request addresses.
      *
      * A regular resource requires READ, as it always did. A stored XQuery is additionally resolved
-     * on EXECUTE via {@link DBBroker#getResourceForExecution(XmldbURI, LockMode)}, so that a caller
+     * on EXECUTE via {@link DBBroker#getResourceForExecution(XmldbURI)}, so that a caller
      * which may run it but not read it gets to run it. Reading a query as data — the {@code ?_source}
      * view — is unaffected and still validates READ separately.
      *
@@ -1578,7 +1578,7 @@ public class RESTServer {
         } catch (final PermissionDeniedException readDenied) {
             final ExecutableResource executable;
             try {
-                executable = broker.getResourceForExecution(uri, LockMode.READ_LOCK);
+                executable = broker.getResourceForExecution(uri);
             } catch (final PermissionDeniedException executeDenied) {
                 // neither readable nor executable, so report the failure to read it
                 throw readDenied;
--- a/exist-core/src/main/java/org/exist/http/RESTServer.java
+++ b/exist-core/src/main/java/org/exist/http/RESTServer.java
@@ -48,6 +48,7 @@ import org.exist.source.StringSource;
 import org.exist.source.URLSource;
 import org.exist.storage.BrokerPool;
 import org.exist.storage.DBBroker;
+import org.exist.storage.ExecutableResource;
 import org.exist.storage.XQueryPool;
 import org.exist.storage.lock.Lock.LockMode;
 import org.exist.storage.lock.ManagedCollectionLock;
@@ -86,6 +87,7 @@ import org.xml.sax.helpers.XMLFilterImpl;
 
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
+import javax.annotation.Nullable;
 import javax.xml.XMLConstants;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.stream.XMLStreamException;
@@ -456,7 +458,7 @@ public class RESTServer {
             // check if path leads to an XQuery resource
             final String xquery_mime_type = MimeType.XQUERY_TYPE.getName();
             final String xproc_mime_type = MimeType.XPROC_TYPE.getName();
-            lockedDocument = broker.getXMLResource(pathUri, LockMode.READ_LOCK);
+            lockedDocument = getResourceForRequest(broker, pathUri);
             resource = lockedDocument == null ? null : lockedDocument.getDocument();
 
             if (null != resource && !isExecutableType(resource)) {
@@ -504,7 +506,7 @@ public class RESTServer {
                     break;
                 }
 
-                lockedDocument = broker.getXMLResource(servletPath, LockMode.READ_LOCK);
+                lockedDocument = getResourceForRequest(broker, servletPath);
                 resource = lockedDocument == null ? null : lockedDocument.getDocument();
                 if (null != resource && isExecutableType(resource)) {
                     break;
@@ -697,7 +699,7 @@ public class RESTServer {
             // if yes, the resource is loaded and the XQuery executed.
             final String xquery_mime_type = MimeType.XQUERY_TYPE.getName();
             final String xproc_mime_type = MimeType.XPROC_TYPE.getName();
-            lockedDocument = broker.getXMLResource(pathUri, LockMode.READ_LOCK);
+            lockedDocument = getResourceForRequest(broker, pathUri);
             resource = lockedDocument == null ? null : lockedDocument.getDocument();
 
             XmldbURI servletPath = pathUri;
@@ -711,7 +713,7 @@ public class RESTServer {
                     break;
                 }
 
-                lockedDocument = broker.getXMLResource(servletPath, LockMode.READ_LOCK);
+                lockedDocument = getResourceForRequest(broker, servletPath);
                 resource = lockedDocument == null ? null : lockedDocument.getDocument();
                 if (null != resource
                         && (resource.getResourceType() == DocumentImpl.BINARY_FILE
@@ -1267,7 +1269,7 @@ public class RESTServer {
         // xquery resource
         while (resource == null) {
             // traverse up the path looking for xquery objects
-            lockedDocument = broker.getXMLResource(servletPath, LockMode.READ_LOCK);
+            lockedDocument = getResourceForRequest(broker, servletPath);
             resource = lockedDocument == null ? null : lockedDocument.getDocument();
             if (resource != null
                     && (resource.getResourceType() == DocumentImpl.BINARY_FILE
@@ -1558,6 +1560,45 @@ public class RESTServer {
     }
 
     /**
+     * Get the document a request addresses.
+     *
+     * A regular resource requires READ, as it always did. A stored XQuery is additionally resolved
+     * on EXECUTE via {@link DBBroker#getResourceForExecution(XmldbURI, LockMode)}, so that a caller
+     * which may run it but not read it gets to run it. Reading a query as data — the {@code ?_source}
+     * view — is unaffected and still validates READ separately.
+     *
+     * @return the locked document, or null if there is none at that path
+     *
+     * @throws PermissionDeniedException if the caller may neither read the resource nor, when it is a
+     *     stored query, execute it
+     */
+    private @Nullable LockedDocument getResourceForRequest(final DBBroker broker, final XmldbURI uri) throws PermissionDeniedException {
+        try {
+            return broker.getXMLResource(uri, LockMode.READ_LOCK);
+        } catch (final PermissionDeniedException readDenied) {
+            final ExecutableResource executable;
+            try {
+                executable = broker.getResourceForExecution(uri, LockMode.READ_LOCK);
+            } catch (final PermissionDeniedException executeDenied) {
+                // neither readable nor executable, so report the failure to read it
+                throw readDenied;
+            }
+
+            if (executable == null) {
+                return null;
+            }
+
+            // only a stored query may be reached without READ; anything else is a data read
+            if (!MimeType.XQUERY_TYPE.getName().equals(executable.document().getDocument().getMimeType())) {
+                executable.close();
+                throw readDenied;
+            }
+
+            return executable.document();
+        }
+    }
+
+    /**
      * Directly execute an XQuery stored as a binary document in the database.
      *
      * @throws PermissionDeniedException
@@ -1586,6 +1627,12 @@ public class RESTServer {
                 context.prepareForReuse();
             }
 
+            // a caller which may execute but not read the query must not learn anything about its
+            // source from a failure. Recomputed per request from the current subject, as the compiled
+            // query is pooled and shared between users, and set before the query is compiled, since a
+            // compile error never reaches XQuery#execute (which computes the level for runtime errors)
+            context.setErrorDisclosure(ErrorDisclosure.of(source, broker.getCurrentSubject()));
+
             // TODO: don't hardcode this?
             context.setModuleLoadPath(
                     XmldbURI.EMBEDDED_SERVER_URI.append(
@@ -1598,31 +1645,37 @@ public class RESTServer {
             reqw.setServletPath(servletPath);
             reqw.setPathInfo(pathInfo);
 
-            final long compilationTime;
-            if (compiled == null) {
-                try {
-                    final long compilationStart = System.currentTimeMillis();
-                    compiled = xquery.compile(context, source);
-                    compilationTime = System.currentTimeMillis() - compilationStart;
-                } catch (final IOException e) {
-                    throw new BadRequestException("Failed to read query from " + resource.getURI(), e);
-                }
-            } else {
-                compilationTime = 0;
-            }
-
-            DebuggeeFactory.checkForDebugRequest(request, context);
-
-            boolean wrap = outputProperties.getProperty("_wrap") != null
-                    && "yes".equals(outputProperties.getProperty("_wrap"));
-
             try {
-                final long executeStart = System.currentTimeMillis();
-                final Sequence result = xquery.execute(broker, compiled, null, outputProperties);
-                writeResults(response, broker, transaction, result, -1, 1, false, outputProperties, wrap, compilationTime, System.currentTimeMillis() - executeStart);
+                final long compilationTime;
+                if (compiled == null) {
+                    try {
+                        final long compilationStart = System.currentTimeMillis();
+                        compiled = xquery.compile(context, source);
+                        compilationTime = System.currentTimeMillis() - compilationStart;
+                    } catch (final IOException e) {
+                        throw new BadRequestException("Failed to read query from " + resource.getURI(), e);
+                    }
+                } else {
+                    compilationTime = 0;
+                }
 
-            } finally {
-                context.runCleanupTasks();
+                DebuggeeFactory.checkForDebugRequest(request, context);
+
+                boolean wrap = outputProperties.getProperty("_wrap") != null
+                        && "yes".equals(outputProperties.getProperty("_wrap"));
+
+                try {
+                    final long executeStart = System.currentTimeMillis();
+                    final Sequence result = xquery.execute(broker, compiled, null, outputProperties);
+                    writeResults(response, broker, transaction, result, -1, 1, false, outputProperties, wrap, compilationTime, System.currentTimeMillis() - executeStart);
+
+                } finally {
+                    context.runCleanupTasks();
+                }
+            } catch (final XPathException e) {
+                // compile and runtime failures alike are filtered: a read-blind caller learns only
+                // that the execution failed, the real error is logged with a correlation id
+                throw ErrorDisclosure.disclose(context, e);
             }
         } finally {
             if (compiled != null) {
--- a/exist-core/src/main/java/org/exist/http/RESTServer.java
+++ b/exist-core/src/main/java/org/exist/http/RESTServer.java
@@ -1619,7 +1619,6 @@ public class RESTServer {
             final XQuery xquery = broker.getBrokerPool().getXQueryService();
             compiled = pool.borrowCompiledXQuery(broker, source);
 
-            final boolean cached = compiled != null;
             final XQueryContext context;
             if (compiled == null) {
                 context = new XQueryContext(broker.getBrokerPool());
@@ -1638,7 +1637,7 @@ public class RESTServer {
             // X-XQuery-Cached reveals whether another user recently executed this shared query; do not
             // disclose it to a read-blind caller (plan §4.6 — suppress X-XQuery-* for GENERIC)
             if (disclosure == ErrorDisclosure.FULL) {
-                response.setHeader("X-XQuery-Cached", Boolean.toString(cached));
+                response.setHeader("X-XQuery-Cached", Boolean.toString(compiled != null));
             }
 
             // TODO: don't hardcode this?
--- a/exist-core/src/main/java/org/exist/http/RESTServer.java
+++ b/exist-core/src/main/java/org/exist/http/RESTServer.java
@@ -1588,8 +1588,12 @@ public class RESTServer {
                 return null;
             }
 
-            // only a stored query may be reached without READ; anything else is a data read
-            if (!MimeType.XQUERY_TYPE.getName().equals(executable.document().getDocument().getMimeType())) {
+            // only a stored XQuery may be reached without READ; anything else is a data read. Require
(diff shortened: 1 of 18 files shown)


exist-core/src/main/java/org/exist/source/SourceFactory.java
```diff
@@ -190,7 +223,26 @@ private static Source getSource_fromClasspath(final String contextPath, final St
      *
      * @return the source, or null if there is no such resource in the db indicated by {@code path}.
      */
-    private static @Nullable Source getSource_fromDb(final DBBroker broker, final XmldbURI path) throws PermissionDeniedException, IOException {
+    private static @Nullable Source getSource_fromDb(final DBBroker broker, final XmldbURI path, final boolean forExecution) throws PermissionDeniedException, IOException {
```
dizzzz:
```suggestion
    private static @Nullable Source getSourceFromDb(final DBBroker broker, final XmldbURI path, final boolean forExecution) throws PermissionDeniedException, IOException {
```

exist-core/src/main/java/org/exist/http/RESTServer.java
```diff
@@ -1574,18 +1619,28 @@ private void executeXQuery(final DBBroker broker, final Txn transaction, final D
             final XQuery xquery = broker.getBrokerPool().getXQueryService();
             compiled = pool.borrowCompiledXQuery(broker, source);
 
-            XQueryContext context;
+            final boolean cached = compiled != null;
```
dizzzz: move to line 1638? or use it in line 1624?

exist-core/src/main/java/org/exist/storage/NativeBroker.java
```diff
@@ -2533,6 +2533,43 @@ public LockedDocument getXMLResource(XmldbURI fileName, final LockMode lockMode)
...
+        final LockMode collectionLockMode = lockManager.relativeCollectionLockMode(LockMode.READ_LOCK, lockMode);
+        try (final Collection collection = openCollection(collUri, collectionLockMode)) {
+            if (collection == null) {
+                LOG.debug("Collection '{}' not found!", collUri);
+                return null;
+            }
+
+            try {
+                // gate on EXECUTE, not READ: the database reads the source on the caller's behalf
+                final LockedDocument lockedDocument = collection.getDocumentWithLock(this, docUri, lockMode, Permission.EXECUTE);
```
dizzzz: not sure... lockMode should be always READ here? so it can be removed from method params?

dizzzz (approved):
Very useful and well thought PR. Thnx!

duncdrum (commented):
Please resolve the codacy warnings, mostly bad test name functions. 

Overall looks good. One question about the knock on effects on locking. What's the expected effect for the same query and permissions pre and post merge. 

Same locks required, more less? Did you confirm the expected result?

exist-core/src/main/java/org/exist/source/SourceFactory.java
line-o: Applied — renamed to `getSourceFromDb` in 55c93cf.

exist-core/src/main/java/org/exist/http/RESTServer.java
line-o: Good catch, it was pointless — dropped the local entirely and inlined `compiled != null` at its one use site (line ~1638) in 55c93cf.

exist-core/src/main/java/org/exist/storage/NativeBroker.java
line-o: Checked every call site (SourceFactory, RESTServer, XQueryCompiler, and the 4 test call sites) — all of them passed LockMode.READ_LOCK, and it makes sense: a document is only ever resolved for execution, never for writing. Removed the parameter in e2a01a2 (DBBroker#getResourceForExecution now takes just the URI); NativeBroker hardcodes READ_LOCK for both the document and collection lock, and no longer needs relativeCollectionLockMode since with both inputs fixed to READ_LOCK it could only ever answer READ_LOCK anyway.

line-o:
Thanks for the review!

**Codacy:** I couldn't get authenticated access to the Codacy report in my environment to see the exact flagged list, so I went by what's visible from the PR itself — the seven test methods in `ExecuteWithoutReadTest` used a `context_scenario` underscore-separated naming style (`readAndExecute_validQueryReturnsItsResult`, `executeOnly_syntaxErrorIsGeneric`, etc.), which is the most likely source of "bad test name functions" against a standard Java naming-convention rule (method names shouldn't contain underscores). Renamed all seven to plain camelCase in 55c93cf. Re-analysis is running now — if there's anything else it flags once the report is available, let me know and I'll follow up.

**Locking:** Traced this through `MutableCollection#getDocumentWithLock` (unmodified by this PR):

```java
public LockedDocument getDocumentWithLock(final DBBroker broker, final XmldbURI name, final LockMode lockMode, final int requiredMode) {
    try (final ManagedCollectionLock collectionLock = lockManager.acquireCollectionReadLock(path)) {
        // ... acquire the document lock per `lockMode` (READ_LOCK/WRITE_LOCK/none) ...
        collectionLock.close();  // early release, Asymmetrical Locking scheme
        // ...
        if (!doc.getPermissions().validate(broker.getCurrentSubject(), requiredMode)) {
            unlockFn.run();  // release the document lock
            throw new PermissionDeniedException(...);
        }
        return new LockedDocument(documentLock, doc);
    }
}
```

This 4-arg overload already existed pre-PR and is permission-generic: `requiredMode` (the `int` permission bitmask checked at the end) was always a parameter independent of the locking mechanics above it. This PR's `NativeBroker#getResourceForExecution` is the first caller to pass `Permission.EXECUTE` instead of `Permission.READ` into it — the locking code itself is untouched.

So, for the same query and permissions:
- **`rwxr-xr-x` (readable + executable, i.e. every query before this PR):** identical before and after. Same document lock (`READ_LOCK`, since `getResourceForExecution` always uses one — see e2a01a2, which also removed the now-pointless `lockMode` parameter per your other comment), same collection-lock acquire/early-release sequence, same order. Only the permission bit checked at the end changed from `READ` to `EXECUTE` — for a query that has both, that's a no-op change in outcome.
- **`--x` only (the new capability):** before this PR, resolution never got this far — the READ-gated getter denied it earlier. Now it succeeds, taking exactly the same lock (document READ_LOCK, briefly-held collection lock) that any other resolved document takes. No new lock type, no longer hold time — it's the existing read path, just reachable by a caller who couldn't reach it before.
- **Denied case (no EXECUTE):** the document lock is briefly acquired and then released via `unlockFn.run()` before the exception is thrown — same acquire-check-release shape the READ-gated path always had for a denied caller.

I verified this by reading the code (no locking logic in `getDocumentWithLock` changed — it's a pre-existing, permission-parameterized method) and by the fact the full test suite (unit + macOS/ubuntu/windows integration) passes, but I didn't add a dedicated concurrency/lock-contention test — nothing here exercises taking this specific EXECUTE-gated path *concurrently* with another writer on the same document. Happy to add one if you'd like a more direct guarantee than the code-reading argument above.

reinhapa (approved):

adamretter:
@line-o @reinhapa @duncdrum @dizzzz @joewiz  I just want to add something here (I have mentioned each of you because I believe this is a very serious and broken change). There is a huge architectural problem with this PR, not to mention it breaks the Security Model that eXist-db has had for years, and which is documented in many places (including the O'Reilly eXist Book), and that users are already very familiar with.

The short description of the PR was:

> Makes stored-query permissions match Unix --x semantics: a query with EXECUTE but not READ can now actually be run, without the caller learning anything about its source from a failure.

However that is both misleading and completely incorrect. The stored-query permissions previously matched the UNIX model. This actually removes that UNIX behaviour. Let me explain:

There are two different models in UNIX for execute-read permissions, machine code binary files (e.g, ELF/a.out), and interpreted scripts (e.g. sh, bash, zsh, etc).

In a UNIX or UNIX like system, it is only possible to execute files that do not have read permission if they are machine code binary files. It is NOT possible to execute scripts without read permissions, because the user has to invoke the interpreter first, e.g. Bash or ZSH, and that interpreter has to read the script.

The way that eXist-db previously behaved was very much by intention and design by myself, so that it matched the UNIX model. That design I implemented replaced the bespoke security model that eXist-db previously had. The idea was that we should just do what UNIX does, so that people have a common model to reason about. Even the ACL extensions in eXist-db were inspired by NFSv4 ACLs.

Before this PR, when eXst-db matched the UNIX model, it meant that in eXist-db when running a script, i.e. XQuery, which in eXist-db is an interpreted script (it is not compiled to byte code or machine code), it needs permission for the calling user to first load the interpreter, and then for the interpreter to read the source code of that XQuery script. This is why XQuery scripts need read-execute permission.

This PR has broken UNIX permission semantics in eXist-db, and I fear a whole bunch of other security related things which you have yet to discover (or receive bug reports about). For the sake of security of your users of eXist-db please consider reverting this PR immediately.

line-o:
@adamretter Thank you for laying out the original intent, aligning with the UNIX permission model, and the history of replacing eXist's bespoke scheme with something people already know how to reason about.\
That's exactly the frame this change should be weighed against, and you're right that it deserved a more critical review than it got.\
That's on me: the description undersold what the PR was changing. It may have read as a small permissions tweak when it is a shift in how stored queries are gated, and it should have said so clearer and louder.\
I was surprised myself at how quickly it got merged – and had to do a sizeable rebase because of it. So I don't think the speed served the change well either. Let's treat this thread as the review it should have had.

With that said, I am still convinced the direction is right — and that it doesn't so much break the UNIX analogy as apply it to the right artifact.

The script-vs-binary distinction is the core of your objection and I agree with it entirely — for files on a filesystem. But eXist is a database, not an operating system, and the thing that actually executes here is not the stored .xq text — it's the compiled query. XQueryPool caches compiled queries keyed by their Source, shared across users, and its contract is Permission.EXECUTE: the gate lives in borrowCompiledXQuery and predates this PR. Map the analogy onto the object that actually runs and it holds — the compiled query is the a.out, the stored .xq is the .c, and you have never needed read on the source to run the compiled form. Even the compile step fits: when the engine must read the source to produce the executable, it now does so under EXECUTE (getSourceForExecution), which is precisely how a kernel loads an --x binary it may execute but not read.

And stepping outside UNIX entirely, separating "may run" from "may read the source" is the mainstream model across the systems eXist is actually comparable to:

- SQL Server : `EXECUTE` and `VIEW DEFINITION` are distinct permissions; you can run a procedure you cannot see.
- PostgreSQL : `SELECT` on a view needs no `SELECT` on the underlying tables; SECURITY DEFINER/invoker makes run-as explicit.
- Oracle : `WRAP` ships PL/SQL that is executable but not readable.
- AWS Lambda : `lambda:InvokeFunction` is a separate IAM action from `lambda:GetFunction`.
- GraphQL : clients invoke persisted/trusted operations by id and never see the resolver.
- Redis / Elasticsearch : `EVALSHA` and stored scripts run a server-side artifact by reference.

So this isn't eXist inventing something exotic; it moves a stored, compiled, execute-gated artifact into line with how databases, cloud functions, and API layers already treat the run/read-source split.

Where your wariness is well placed: the current solution is unfinished. While rebasing I found a concrete gap. The main query is resolved on **EXECUTE**, but an imported module (`import module A at B`) is still resolved on **READ** — in two places: `SourceFactory.getSource`, and again inside the pool's validity check, where `DBSource.isValid()` re-opens the module with a **READ** lock under the calling user. So a read-blind caller can run a self-contained query, but a query that imports a module it cannot read fails — and fails opaquely, as a generic error. That is exactly the kind of loose end you were pointing at, and it shouldn't ship half-done. A follow-up is already lined up to resolve module and dependency loading on the same **EXECUTE** basis as the main query, so the behaviour is consistent and intentional. I'll link it here.

Let's push to get this right and keep iterating in review.

adamretter:
@line-o Is your response AI generated? I ask because it reads like it is AI generated, and it doesn't make any sense in practice. The language style is also not what I am familiar with from you. Likewise your content normally makes sense, but this does not.

We need only read the response as far as this statement to realise that it is contradicting itself:

> With that said, I am still convinced the direction is right — and that it doesn't so much break the UNIX analogy as apply it to the right artifact.
>
> The script-vs-binary distinction is the core of your objection and I agree with it entirely — for files on a filesystem. But eXist is a database, not an operating system, and the thing that actually executes here is not the stored .xq text — it's the compiled query. 

Quite clearly this is non-sensical. It acknowledges that XQuery is interpreted script text (and therefore should require read-execute perms to remain aligned to the Unix model), but then it goes on to say that the thing executed is the compiled query. Queries in eXist-db are not compiled. Even if the code functions are misleadingly named `compile`, XQuery in eXist-db is interpreted. There is a fork of FusionDB where we previously demonstrated XQuery compilation during a conference talk (see: https://www.youtube.com/watch?v=nxLqE_7ZO2A), but eXist-db does not have anything like that.

Clearly, just squinting and calling a duck a goose does not make it a goose. The same goes with this PR. If you want to follow the Unix model for eXist-db, you have to acknowledge that XQuery is stored in the database as plain-text script; and that needs read access by the interpreter to read the file from the database. Anything else breaks the Unix model.

Additionally, I haven't checked, but I suspect this change has opened many security holes in other places. I did see you tried to implement some bandages for some of those, but they are not sufficient and holes look like they still remain.

I recommend to the core eXist-db developers, that this PR should be undone immediately on the basis that (1) it is architecturally and philosophically incorrect, (2) it introduces security holes, and (3) for users it breaks the known security model and norms of the well established Unix model. Please, stop, and think of your users.

PR #6602, opened 2026-07-22
[bugfix] Make mvn package re-runnable

### LINKED ISSUE #6601: [bugfix] mvn package cannot be run twice: exist-distribution deletes etc/jetty it cannot recreate
### Description

`mvn package` cannot be run twice in a row: the second invocation fails in `exist-distribution`.

```
[ERROR] Failed to execute goal org.codehaus.mojo:xml-maven-plugin:1.2.1:transform
  (adjust-jetty-deploy-for-appassembler) on project exist-distribution: The directory
  .../exist-distribution/target/exist-distribution-dir/etc/jetty, which is a base directory
  of a ValidationSet or TransformationSet, does not exist.
```

### Reproduce

```bash
mvn clean package -DskipTests -Ddependency-check.skip=true -Ddocker=false   # succeeds
mvn package       -DskipTests -Ddependency-check.skip=true -Ddocker=false   # fails
```

### Cause

Three `package` steps in `exist-distribution/pom.xml` run in sequence:

1. `delete-jetty-dest-before-rename` deletes `${assemble.dir}/etc/jetty`
2. `rename-jetty-etc-dir-for-appassembler` renames `etc/org/exist/jetty/etc` → `etc/jetty`
3. `adjust-jetty-deploy-for-appassembler` transforms the files in `etc/jetty`

On the second run, `unpack-for-appassembler` is skipped (its marker files already exist), so the rename's source `etc/org/exist/jetty/etc` is gone — it was renamed away by the first run. Step 1 has meanwhile deleted the `etc/jetty` that the first run produced, and the rename cannot recreate it, so step 3 finds no directory and fails.

`delete-mac-bundle-jetty-dest-before-rename` does the same for the macOS bundle.

### Impact

Any non-clean rebuild, and any tool that runs `package` over the reactor without `clean` — e.g. Moderne's LST build (#6600), which as a result produces no LST at all.

line-o:
### Description:

`mvn package` could not be run twice in a row — the second invocation failed in `exist-distribution`:

```
[ERROR] Failed to execute goal org.codehaus.mojo:xml-maven-plugin:1.2.1:transform
  (adjust-jetty-deploy-for-appassembler) on project exist-distribution: The directory
  .../exist-distribution-dir/etc/jetty, which is a base directory of a ValidationSet
  or TransformationSet, does not exist.
```

Three `package` steps ran in sequence: `delete-jetty-dest-before-rename` deleted `${assemble.dir}/etc/jetty`, `rename-jetty-etc-dir-for-appassembler` renamed `etc/org/exist/jetty/etc` onto it, and `adjust-jetty-deploy-for-appassembler` transformed the result.

That only works once. On a re-run the rename's source is gone — the first run renamed it away — because `unpack-for-appassembler`, which provides it, is skipped by its `markersDirectory`. The delete had meanwhile removed the perfectly good `etc/jetty` the first run produced, and the rename could not recreate it, so the transform found no directory.

Dropping the two delete executions leaves the previous run's `etc/jetty` in place; the rename then becomes a no-op, which it already tolerates via `ignoreFileNotFoundOnIncremental`. `delete-mac-bundle-jetty-dest-before-rename` is removed for the same reason — the macOS bundle steps have the identical structure.

### Reference:

Closes https://github.com/eXist-db/exist/issues/6601

Also unblocks #6600: Moderne's LST build runs `package` over the reactor without `clean`, so it hit this on every build after the first.

### Type of tests:

None — build configuration only, no production or test code changes. Verified by running the documented build twice in a row (JDK 21, macOS, so the `mac-dmg-on-mac` profile is active and the mac-bundle steps are exercised):

```bash
mvn clean package -DskipTests -Ddependency-check.skip=true -Ddocker=false
mvn package       -DskipTests -Ddependency-check.skip=true -Ddocker=false
```

- Before: run 2 → `eXist-db Distributions ... FAILURE [0.684 s]`
- After: run 2 → `eXist-db Distributions ... SUCCESS [35.046 s]`

and the output is correct, not merely present: `exist-distribution-dir/etc/jetty` holds the full set of jetty config files, with the `jetty-deploy.xslt` adjustment applied to `jetty-deploy.xml`.

Both runs still report `BUILD FAILURE` for `exist-xqts` on my machine — resolving `exist-xqts-runner_2.13` returns 401 from GitHub Packages without credentials. That is unrelated to this change and identical before and after it.

### DIFF
--- a/exist-distribution/pom.xml
+++ b/exist-distribution/pom.xml
@@ -481,30 +481,6 @@
                 <version>3.2.0</version>
                 <executions>
                     <execution>
-                        <id>delete-jetty-dest-before-rename</id>
-                        <phase>package</phase>
-                        <goals>
-                            <goal>run</goal>
-                        </goals>
-                        <configuration>
-                            <target>
-                                <delete dir="${assemble.dir}/etc/jetty" failonerror="false" quiet="true"/>
-                            </target>
-                        </configuration>
-                    </execution>
-                    <execution>
-                        <id>delete-mac-bundle-jetty-dest-before-rename</id>
-                        <phase>${mac.bundle.phase}</phase>
-                        <goals>
-                            <goal>run</goal>
-                        </goals>
-                        <configuration>
-                            <target>
-                                <delete dir="${mac.bundle.resources.dir}/etc/jetty" failonerror="false" quiet="true"/>
-                            </target>
-                        </configuration>
-                    </execution>
-                    <execution>
                         <id>copy files</id>
                         <phase>package</phase>
                         <goals>
--- a/exist-distribution/pom.xml
+++ b/exist-distribution/pom.xml
@@ -555,6 +555,36 @@
                             </target>
                         </configuration>
                     </execution>
+
+                    <execution>
+                        <id>rename-jetty-etc-dir-for-appassembler</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <move file="${assemble.dir}/etc/org/exist/jetty/etc"
+                                      tofile="${assemble.dir}/etc/jetty"
+                                      failonerror="false" quiet="true"/>
+                            </target>
+                        </configuration>
+                    </execution>
+
+                    <execution>
+                        <id>rename-jetty-etc-dir-for-mac-bundle</id>
+                        <phase>${mac.bundle.phase}</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <move file="${mac.bundle.resources.dir}/etc/org/exist/jetty/etc"
+                                      tofile="${mac.bundle.resources.dir}/etc/jetty"
+                                      failonerror="false" quiet="true"/>
+                            </target>
+                        </configuration>
+                    </execution>
                 </executions>
             </plugin>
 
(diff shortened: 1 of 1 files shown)


dizzzz:
It might be interesting to test this PR on windows.  Windows traditionally has an ugly file locking mechanism

dizzzz (approved):

duncdrum:
Ran `mvn clean verify` then `mvn verify` again with no clean — second run succeeds, and the transformed `etc/jetty` is byte-for-byte identical to the first run's.

One rough edge: the no-clean re-run logs (non-fatal)
```
[ERROR] sourceFile .../etc/org/exist/jetty/etc does not exist
```
from `rename-jetty-etc-dir-for-appassembler`/its mac-bundle twin. `ignoreFileNotFoundOnIncremental` only quiets this inside m2e's incremental build context — a plain CLI re-run always hits `log.error(...)` in `RenameMojo`, it just never throws.

Suggest swapping those two `rename` executions for a plain Ant `<move failonerror="false" quiet="true">` under the antrun block already here (same attributes the removed `<delete>` used) — silently no-ops on a missing source instead of logging ERROR:

```xml
<move file="${assemble.dir}/etc/org/exist/jetty/etc"
      tofile="${assemble.dir}/etc/jetty"
      failonerror="false" quiet="true"/>
```

line-o:
I will implement the suggested improvement.

line-o:
Implemented — swapped both `RenameMojo` executions (appassembler and mac-bundle) for a plain Ant `<move failonerror="false" quiet="true">` under the antrun block, same attributes the removed `<delete>` used. Pushed in d35489e.

Verified with the same clean-then-no-clean double build: second run now succeeds with no ERROR log at all, and `etc/jetty` output matches the first run.

duncdrum (approved):

PR #6607, opened 2026-07-23
[bugfix] Set EXIST_HOME so system:get-exist-home() returns a value.

### LINKED ISSUE #6582: [BUG] system:get-exist-home() does not return value
**Describe the bug**
`system:get-exist-home()` does not always return a value

**Expected behavior**
function shall always return value

**To Reproduce**
- check out 'develop' / v7
- build exist
- go to `exist-distribution/target/exist-distribution-dir`
- run `bin/startup.sh`
- open `eXide`, login as admin
- run `system:get-exist-home()` --> returns empty-sequence.

 - Build: [eXist-7/develop]
 - Java: [21]
 - OS: [Mac OS X 26.5.1]

**Additional context**
- How is eXist-db installed? [e.g. JAR installer, DMG, … ] - git
- Any custom changes in e.g. `conf.xml`? - no

dizzzz:
Set EXIST_HOME so system:get-exist-home() returns a value when started with ./bin/startup.sh 

There are probably a lot of quality issues on this file, explicitly not addressing these.
The actual change is in commit 1 ; commit 2 re-orders the loading.

fixes #6582

**Summary**
conf.xml was being found on the classpath (from test resources in target/test-classes) because the classpath lookup ran unconditionally before the file-based lookup, allowing it to shadow the real config file.

**Changes**
exist-core/src/main/java/org/exist/util/Configuration.java: Reversed the lookup order in the main constructor — file-based lookup now runs first; classpath is only tried as a fallback when no file is found. The error is thrown only after both strategies are exhausted.

**Verification**
Build: mvn install -pl exist-core -am -DskipTests — success.
Tests: ConfigurationTest — 10/10 passed.

### DIFF
--- a/exist-core/src/main/java/org/exist/util/Configuration.java
+++ b/exist-core/src/main/java/org/exist/util/Configuration.java
@@ -263,60 +263,62 @@ public class Configuration implements ErrorHandler {
                 configFilename = DatabaseImpl.CONF_XML;
             }
 
-            // firstly, try to read the configuration from a file within the
-            // classpath
-            try {
-                is = Configuration.class.getClassLoader().getResourceAsStream(configFilename);
-
-                if (is != null) {
-                    LOG.info("Reading configuration from classloader");
-                    configFilePath = Optional.of(Path.of(Configuration.class.getClassLoader().getResource(configFilename).toURI()));
-                    existHome = configFilePath.map(p -> p.getParent().getParent());
-                }
-            } catch (final Exception e) {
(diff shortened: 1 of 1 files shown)


dizzzz:
@line-o @duncdrum @reinhapa ?

dizzzz:
Need two items:
- add functional test.
- add check for dba

line-o:
Apparently, the function can be called as guest. Please ensure that only DBA is allowed. I would like to see a test or two in XQuery.

PR #6622, opened 2026-08-05
[bugfix] Tolerate ambiguous URI segments in getServletPath()

### LINKED ISSUE #6618: XQueryURLRewrite throws 400 Ambiguous URI encoding for %2F in path segments
**Description**

Requests whose path contains a real `/` followed by a percent-encoded slash (`%2F`) within a path segment are rejected with `400 Ambiguous URI encoding: AMBIGUOUS_PATH_SEPARATOR`, even though the equivalent request worked under eXist builds on Jetty 9 (e.g. `duncdrum/existdb`). This breaks OpenAPI-style REST APIs (e.g. TEI Publisher / Jinks apps built with Roaster) whose path parameters legitimately contain slashes.

**Steps to reproduce**

```sh
docker run -dit --name exist-repro -p 18080:8080 existdb/existdb
# wait for "Server has started" in `docker logs exist-repro`

curl -i "http://localhost:18080/exist/rest/db/apps/jinks%2Ffoo"
# -> 404, tolerated (no HttpServletRequest#getServletPath()/getPathInfo() call on this path)

curl -i "http://localhost:18080/exist/apps/<any-installed-app>/api/foo/bar%2Fbaz"
# -> 400 Ambiguous URI encoding: AMBIGUOUS_PATH_SEPARATOR
```

Any request that ends up inside `XQueryURLRewrite.service()` and contains an ambiguous path segment reproduces it; the second example just needs any deployed app mounted under `/exist/apps/<abbrev>`.

**Root cause**

[`XQueryURLRewrite.java` lines 322-323](https://github.com/eXist-db/exist/blob/c601562473912342eec8c1a33652238fde2901f7/exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java#L322-L323) call the "unsafe" Servlet API method `request.getServletPath()` one line after already safely calling `request.getRequestURI()`:

```java
// store the original request URI to org.exist.forward.request-uri
modifiedRequest.setAttribute(RQ_ATTR_REQUEST_URI, request.getRequestURI());   // safe
modifiedRequest.setAttribute(RQ_ATTR_SERVLET_PATH, request.getServletPath()); // throws here
```

Under Jetty 12's EE10 servlet module, any request whose `HttpURI` carries a recorded ambiguity violation gets wrapped in `ServletApiRequest.AmbiguousURI`, and `getServletPath()`/`getPathInfo()` on that wrapper throw `IllegalArgumentException: 400: Ambiguous URI encoding: ...` — this is a deliberate, Jetty-documented restriction (see [jetty/jetty.project#12346](https://github.com/jetty/jetty.project/issues/12346)), independent of the connector's `HttpConfiguration.uriCompliance` setting.

**What I tried at the Jetty-config level (no eXist code changes)**

Per Jetty's own docs, two independent switches are supposed to be needed together:

1. `HttpConfiguration.setUriCompliance(UriCompliance.from("DEFAULT,AMBIGUOUS_PATH_SEPARATOR,AMBIGUOUS_PATH_ENCODING"))` on the connector's `httpConfig` — this measurably changed behavior (e.g. moved where an unrelated `AMBIGUOUS_EMPTY_SEGMENT` case gets rejected), confirming it's applied.
2. `ServletHandler.setDecodeAmbiguousURIs(true)` on `webapps/exist-webapp-context.xml`'s `WebAppContext` bean (`<Get name="servletHandler"><Set name="decodeAmbiguousURIs">true</Set></Get>`), following the exact pattern in Jetty's own [ee10-servlet-ambiguous-paths example](https://github.com/jetty/jetty-examples/tree/12.0.x/embedded/ee10-servlet-ambiguous-paths).

Both loaded without any startup error, but the second had **no observable effect** — the repro above still throws the identical exception with both applied. I wasn't able to determine from the outside whether the `ServletHandler` instance mutated via the context XML at config time is actually the same instance later consulted per-request by `ServletContextRequest.newServletApiRequest()`, or whether something about `org.exist.jetty.WebAppContext`'s/`JettyStart`'s deployment lifecycle (`webapps/*.xml` processed via a separate `XmlConfiguration`/`DeploymentManager` path from `standard.enabled-jetty-configs`) discards or doesn't propagate that setting to the live object. This would need Java-level instrumentation (or eyes on `JettyStart.java`/`WebAppProvider` wiring) to confirm.

**Suggested direction**

`XQueryURLRewrite` already has the safe value available one line above (`request.getRequestURI()`); avoiding the unsafe `getServletPath()`/`getPathInfo()` calls (or deriving the servlet-path attribute from the already-fetched request URI instead) would sidestep this class of failure entirely, matching Jetty's own recommendation to prefer `getRequestURI()` for ambiguity-tolerant code. Alternatively, if `ServletHandler.decodeAmbiguousURIs` is the intended fix, it would help to know why it doesn't appear to take effect via `webapps/exist-webapp-context.xml`.

**Context**

```
Build: eXist-7.0.0-SNAPSHOT (c601562473912342eec8c1a33652238fde2901f7)
Java: 21.0.12 (Eclipse Adoptium)
OS: Linux 6.12.76-linuxkit (aarch64)
```
(via `xst info` against the official `existdb/existdb` Docker image, built 2026-08-04)

line-o:
### Description:

Under Jetty 12's EE10 servlet module, `HttpServletRequest#getServletPath()` throws `IllegalArgumentException` (surfaced to the client as a 400 response) when the request URI contains an ambiguous path segment (e.g. a real `/` next to an encoded `%2F`), even though `getRequestURI()` on the very same request remains safe to call.

`XQueryURLRewrite` is always mapped with `url-pattern` `/*` (see `web.xml`), a path-prefix mapping with an empty prefix, so per the Jakarta Servlet spec `getServletPath()` is defined to always return `""` here regardless of the request URI — Jetty's own ambiguity check is a blanket refusal to answer, not a sign the answer differs. This PR adds a `getServletPathSafely()` helper that falls back to that known value (`""`) instead of failing the whole request when Jetty throws.

See also [jetty/jetty.project#12346](https://github.com/jetty/jetty.project/issues/12346).

### Reference:

Closes #6618

### Type of tests:

Added two JUnit tests to `XQueryURLRewriteTest`:
- `getServletPathSafely_returnsServletPath` — normal path, delegates to `getServletPath()`
- `getServletPathSafely_toleratesAmbiguousUriException` — Jetty throws `IllegalArgumentException`, helper falls back to `""`

### DIFF
--- a/exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
+++ b/exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
@@ -320,7 +320,7 @@ public class XQueryURLRewrite extends HttpServlet {
 
                     // store the original request URI to org.exist.forward.request-uri
                     modifiedRequest.setAttribute(RQ_ATTR_REQUEST_URI, request.getRequestURI());
-                    modifiedRequest.setAttribute(RQ_ATTR_SERVLET_PATH, request.getServletPath());
+                    modifiedRequest.setAttribute(RQ_ATTR_SERVLET_PATH, getServletPathSafely(request));
 
                 }
                 if (LOG.isTraceEnabled()) {
@@ -369,6 +369,29 @@ public class XQueryURLRewrite extends HttpServlet {
             throw new ServletException("An error occurred while processing request to " + request.getRequestURI() + ": "
                     + e.getMessage(), e);
 
+        }
+    }
+
+    /**
+     * Under Jetty 12's EE10 servlet module, {@link HttpServletRequest#getServletPath()}
+     * throws {@code IllegalArgumentException} (surfaced to the client as a 400 response)
+     * when the request URI contains an ambiguous path segment (e.g. a real '/' next to
+     * an encoded '%2F'), even though {@link HttpServletRequest#getRequestURI()} on the
+     * very same request remains safe to call. This servlet is always mapped with
+     * {@code url-pattern} {@code /*} (see web.xml), a path-prefix mapping with an empty
+     * prefix, so per the Jakarta Servlet spec {@code getServletPath()} is defined to
+     * always return {@code ""} here regardless of the request URI -- Jetty's own
+     * ambiguity check is a blanket refusal to answer, not a sign the answer differs.
+     * Fall back to that known value instead of failing the whole request.
+     *
+     * @see <a href="https://github.com/jetty/jetty.project/issues/12346">jetty/jetty.project#12346</a>
+     */
+    static String getServletPathSafely(final HttpServletRequest request) {
+        try {
+            return request.getServletPath();
+        } catch (final IllegalArgumentException e) {
+            LOG.debug("Falling back to \"\" for servlet path of ambiguous request URI {}: {}", request.getRequestURI(), e.getMessage());
+            return "";
         }
     }
 
--- a/exist-core/src/test/java/org/exist/http/urlrewrite/XQueryURLRewriteTest.java
+++ b/exist-core/src/test/java/org/exist/http/urlrewrite/XQueryURLRewriteTest.java
@@ -144,4 +144,43 @@ public class XQueryURLRewriteTest
             assertArrayEquals(newTestParameterMap.get(paramName), wrapper.getParameterMap().get(paramName));
         }
     }
+
+    @Test
+    public void getServletPathSafely_returnsServletPath() {
+        HttpServletRequest mockHttpServletRequest = EasyMock.createMock(HttpServletRequest.class);
+
+        // XQueryURLRewrite is always mapped with url-pattern "/*" (see web.xml), a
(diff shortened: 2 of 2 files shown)


dizzzz (approved):

dizzzz:
build failure is in different branch too. https://github.com/eXist-db/exist/actions/runs/31020825166/job/92778340404?pr=6622#step:8:1370

line-o:
The configuration approach is not working for us as we already use the most forgiving setting "UNSAFE"
https://github.com/eXist-db/exist/blob/3d55123a212bc29c441cf9265970f0796481cb9c/exist-jetty-config/src/main/resources/org/exist/jetty/etc/jetty.xml#L78

duncdrum (approved):

wolfgangmm (approved):

reinhapa (approved):

PR #6631, opened 2026-08-10
Add system:get-jmx-token() to expose JMXServlet's access token

### LINKED ISSUE #6620: Expose system:jmx-token() to avoid re-deriving JMXServlet's token/data-dir resolution in XQuery
## Problem

`JMXServlet` (`exist-core/src/main/java/org/exist/management/client/JMXServlet.java`) waives its `token` requirement only for requests whose *literal* peer address is a registered loopback address (`registerLocalHostAddresses()`/`isFromLocalHost()`). Any other client — including a same-host request that passes through a reverse proxy, container port-publishing, or a genuinely remote/LAN client — must supply the correct `token` query parameter, or gets `403 Forbidden`.

There is currently no supported way for an XQuery application to obtain that token, or the data directory it's derived from, without reimplementing `JMXServlet`'s private resolution logic:

- `getToken()` (line 297) is `private`, reads `dataDir.resolve("jmxservlet.token")` fresh on every servlet request, and is not published anywhere else (not as a `ServletContext` attribute, not via a public API, not as a JMX MBean attribute itself).
- The `dataDir` it resolves against (`init()`, line 202) is **not** simply `{exist-home}/data` — it comes from `client.getDataDir()` (`JMXtoXML.java:330`), which queries the `org.exist.management.exist:type=DiskUsage` MBean's `DataDirectory` attribute, and only falls back to `{webapp}/WEB-INF/data` if that MBean attribute is `null`.

This came up via eXist-db/monex#411: Monex (the bundled monitoring dashboard) needs this token client-side to poll `/exist/status` from the browser once the loopback exemption above no longer applies (e.g. it's typically fronted by a reverse proxy or accessed over the LAN). It currently guesses the token file's location in pure XQuery by re-deriving `{exist-home}/data`, then by hand-parsing `conf.xml`'s `db-connection/@files` (see draft fix eXist-db/monex#422). Neither approach can reliably reproduce `JMXtoXML.getDataDir()`'s actual MBean-based resolution or the `WEB-INF/data` fallback, so the guess silently diverges from the real value on some installs (that's the direct cause of eXist-db/monex#411: the guessed path doesn't hold `jmxservlet.token`, so Monex embeds an empty token, which is emitted as `token=""` and gets a `403` since `JMXServlet.hasSecretToken()` explicitly refuses a blank token).

Reviewing eXist-db/monex#422, @adamretter also flagged that reading an arbitrary file via the `file` module (DBA-only, but still a broad filesystem-access capability) to reconstruct one specific secret is a heavier-weight and more error-prone mechanism than necessary.

## Proposal

Expose a narrow, purpose-built, DBA-only XQuery function — `system:jmx-token()` — backed by the exact same resolution `JMXServlet` itself uses, so the two can't drift apart:

1. Extract `JMXServlet`'s token/dataDir resolution (`init()`'s `dataDir` lookup, `obtainTokenFileReference()`, `getToken()`) out of the servlet and into a small shared class in `org.exist.management.client`, e.g. `JMXTokenProvider`, taking a `JMXtoXML` client and returning the token. `JMXServlet` calls it instead of keeping the logic private.
2. Add `system:jmx-token()` as its own function class under `org.exist.xquery.functions.system` (matching the existing one-class-per-function convention, e.g. `GetExistHome.java`), registered in `SystemModule`, DBA-gated the same way `Shutdown.java`/`ClearXQueryCache.java` already check `hasDbaRole()` in that package. It calls into `JMXTokenProvider` using a `JMXtoXML` client instance, and returns the empty sequence if JMX/the status servlet isn't available.

Since I maintain eXist-db/monex, I'd pick this up on the Monex side (dropping the file-based guessing entirely) as soon as it lands, and can also help drive the eXist-db-core side if useful.

line-o:
### Description:

`JMXServlet` waives its `token` requirement only for requests whose literal peer address is a registered loopback address. Any other client — behind a reverse proxy, container port-publishing, or on the LAN — must supply the correct `token` query parameter. There was previously no supported way for an XQuery application to obtain that token (or the data directory it's derived from) without re-implementing the servlet's private resolution logic, which tends to silently diverge from the real value (the `WEB-INF/data` fallback, or the `DiskUsage` MBean-based data directory lookup).

This PR:

1. Extracts `JMXServlet`'s token/data-dir resolution (`init()`'s `dataDir` lookup, `obtainTokenFileReference()`, `getToken()`) into a new shared class, `JMXTokenProvider`, in `org.exist.management.client`. `JMXServlet` now delegates to it instead of keeping the logic private.
2. Adds `system:get-jmx-token()` as its own function class, `GetJmxToken`, under `org.exist.xquery.functions.system` (matching the existing one-class-per-function convention, e.g. `GetExistHome.java`), registered in `SystemModule`, DBA-gated the same way `Shutdown`/`ClearXQueryCache` already check `hasDbaRole()`. It returns the empty sequence if JMX/the token can't be resolved, rather than raising an error.

Both callers now resolve the token via the exact same code path, so they can't drift apart.

### Reference:

Closes #6620

### Type of tests:

- **Java unit tests** (`JMXTokenProviderTest`): MBean-value vs. fallback-dir vs. empty resolution (including when the MBean lookup throws), token creation/persistence, reading an existing token file, regenerating a token file that's missing the `token` property. These exercise the JMX-unavailable paths that can't be reached from a live test instance, since `AgentFactory#initDBInstance` registers the `DiskUsage` MBean unconditionally at startup.
- **Java integration test** (`GetJmxTokenTest`, via `ExistXmldbEmbeddedServer`): a DBA gets a stable, non-blank token; a non-DBA (`guest`) is denied; editing the on-disk `jmxservlet.token` file while the instance is running is picked up on the next call (no caching).
- **XQSuite** (`get-jmx-token.xqm`): happy path (DBA gets a UUID-shaped token), token stability across calls, and the one XQuery-level error case (`%test:user("guest", "guest")` + `%test:assertError("exerr:ERROR")`).
- Full `exist-core` test suite run locally: 7203 tests, 0 failures, 0 errors.

### DIFF
--- a/exist-core/src/main/java/org/exist/management/client/JMXTokenProvider.java
+++ b/exist-core/src/main/java/org/exist/management/client/JMXTokenProvider.java
@@ -1,0 +1,162 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.management.client;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.exist.util.UUIDGenerator;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
+import java.util.Properties;
+
+/**
+ * Canonical resolution of {@link JMXServlet}'s shared secret token, i.e. the value the servlet
+ * accepts via its {@code token} request parameter as an alternative to a request originating
+ * from localhost (see {@code JMXServlet#hasSecretToken}).
+ * <p>
+ * The token is persisted in a {@code jmxservlet.token} properties file inside the data
+ * directory reported by the {@code org.exist.management.exist:type=DiskUsage} MBean's
+ * {@code DataDirectory} attribute, falling back to a caller-supplied directory (e.g.
+ * {@code JMXServlet}'s {@code WEB-INF/data}) if that attribute is unavailable.
+ * <p>
+ * Both {@code JMXServlet} and the {@code system:get-jmx-token()} XQuery function delegate to
+ * this class rather than each re-deriving the token/data-dir resolution independently.
+ *
+ * @author eXist-db authors
+ */
+public class JMXTokenProvider {
+
+    private static final Logger LOG = LogManager.getLogger(JMXTokenProvider.class);
+
+    private static final String TOKEN_KEY = "token";
+    private static final String TOKEN_FILE = "jmxservlet.token";
+    private static final String TOKEN_FILE_COMMENT = """
+            JMXservlet token
+            Use: /exist/status?token=<token>
+            Obtain: system:get-jmx-token()""";
+
+    private final JMXtoXML client;
+
+    @Nullable
+    private final Path fallbackDataDir;
+
+    /**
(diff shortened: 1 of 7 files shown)


exist-core/src/main/java/org/exist/management/client/JMXTokenProvider.java
```diff
@@ -0,0 +1,162 @@
...
+public class JMXTokenProvider {
+
+    private static final Logger LOG = LogManager.getLogger(JMXTokenProvider.class);
+
+    private static final String TOKEN_KEY = "token";
+    private static final String TOKEN_FILE = "jmxservlet.token";
+    private static final String TOKEN_FILE_COMMENT =
+            "JMXservlet token\n" +
+            "Use: /exist/status?token=<token>\n" +
+            "Obtain: system:get-jmx-token()";
```
reinhapa: may use text block?

reinhapa (changes requested):

exist-core/src/main/java/org/exist/management/client/JMXTokenProvider.java
line-o: Good idea, done — switched to a text block in a256e7b.

reinhapa (approved):

dizzzz (approved):

Issue #6594, opened 2026-07-21
[bugfix] Read-blind XQuery execution evicts the shared compiled-query pool and cannot reuse module-importing queries


line-o:
## Summary

Since #6586 (execute-without-read, follows #6568) merged, an **execute-only
caller** — one with `EXECUTE` but not `READ` on a stored query — now reaches
`XQueryPool.borrowCompiledXQuery` for real. That path is not safe for such a
caller, in three ways:

1. **It evicts the shared pool entry on every read-blind request.**
   `borrowCompiledXQuery` → `isCompiledQueryValid` → `DBSource.isValid()`
   resolves the document through the **READ-gated** `getXMLResource(uri,
   READ_LOCK)` **as the calling subject** and maps `PermissionDeniedException` →
   `Validity.INVALID` (`exist-core/.../source/DBSource.java:83-99`). So a
   legitimate execute-only caller invalidates the compiled query, evicts it from
   the pool that is shared with all users, and forces a **recompile on every
   request**.

2. **Module-importing queries are unrunnable for a read-blind caller, even when
   the pool is warm.** `compiledXQuery.isValid()` applies the same READ check to
   every imported module's `DBSource`
   (`XQueryContext.checkModulesValid`, `.../xquery/XQueryContext.java:1779-1793`,
   via `ExternalModule.moduleIsValid` / `PathExpr.isValid`). A caller lacking
   `READ` on any imported module invalidates and evicts, and the ensuing
   recompile is then denied at the import gate (`SourceFactory.getSource`, READ).

3. **The cache is mutated before `EXECUTE` is validated.** `computeIfPresent`
   polls/evicts before the `bSource.validate(Permission.EXECUTE)` check, so even
   a caller with no permission at all perturbs shared pool state.

The URL-rewrite cache hit path (`.../http/urlrewrite/XQueryURLRewrite.java:483-486`,
`isValid() != VALID` → `urlCache.remove`) carries the same
eviction-on-permission-denial defect.

## Reproduction (verified)

Driven over real HTTP — a read-blind caller runs a query that
`import module … at`'s a module stored at mode `rwx--x--x` (executable, not
readable by the caller):

- A self-contained execute-only query runs and returns its result. ✅
- A module-importing execute-only query **fails**: the real error is
  `XQST0059 … Permission denied to read module source`, sanitized to the generic
  `EXXQDY0010` for the read-blind caller.
- **There is no warm/cold difference.** Priming the pool with the fully compiled
[...]

Issue #5916, opened 2025-10-22
deadlock saving a script  when it is running


KitWallace:
Deadlock occurs if an xquery script is edited and saved whilst it is running and using the db.  Monex is unresponsive so query cant be stopped; shutdown hangs and a dirty restart needed (6.4)

**Generate**
```xquery
xquery version "3.1";
<trips>
    {for $i in 1 to 1000
    return
        <Trip>
            <id>{concat("trip-",$i)}</id>
        </Trip>
    }
</trips>
```
**Run** 
```xquery
xquery version "3.1";
let $tripid := "trip-250"
return
      <test>
        {count(for $i in 1 to 1000
               return  doc("/db/apps/test/data/trips.xml")//Trip[id  = $tripid])
        }
      </test>
```

line-o:
@KitWallace Unfortunately, this is a known issue that we hoped was solved with the new locking introduced in exist version 5. So thanks for bringing this up and especially providing a reproducer.

line-o:
It would help us if you could tell us how you run your instance. Which operating system, which java version and what CPU architecture?

I always use `xst info` for that.

adamretter:
> that we hoped was solved with the new locking introduced in exist version 5.

@line-o Nope! This was actually introduced in eXist-db 5 with the locking rewrite that my company contributed to eXist-db to reduce the likely hood of ending up with corrupted resources or a corrupted database.

adamretter:
@KitWallace Thanks, is it possible that you could provide us with a dump of the Lock Table when this occurs please?

KitWallace:
@line-o This the sort of thing? (but what's xst info ?)

eXist 6.4.0 2e4182841b38317324877c0b2e1febcaa26ed83a 2025-05-06T14:29:27Z

kitwallace@kitwallace:~$ java -version
openjdk version "21.0.8" 2025-07-15
OpenJDK Runtime Environment (build 21.0.8+9-Debian-1)
OpenJDK 64-Bit Server VM (build 21.0.8+9-Debian-1, mixed mode, sharing)

lsb_release -a
Description:    Debian GNU/Linux 13 (trixie)
Release:        13
Codename:       trixie

Server at Bitfolk https://bitfolk.com/plans.html

Name | RAM | Storage | Data Transfer | Host | Operating System | Arch.
kitwallace | 3,072 | 30.00 | 2,000 | paradox | Debian 13.x (trixie) | x86_64

[...]

Issue #4233, opened 2022-02-09
[BUG] Server deadlock when updating XSL file


paulmer:
**Describe the bug**
eXist can become unresponsive if an XSL file is updated while the file is in use by a REST request to perform a transformation.

The general scenario is this:  The eXist database contains an XQuery script and an XSL document.  The script optains an XML document from an external source (fn:doc($url)) and produces a result using transform:transform and supplying the XSL document as the stylesheet.  This script responds to web requests forwarded to the REST interface.

A second application, using a remote XMLDB connection, updates the XSL document.  If the timing is right, a deadlock is created that causes the server to stop responding to all requests.

The only recovery possible is to forceably kill the server.

**Expected behavior**
Both the transformation and the update should succeed.

**To Reproduce**
This is a difficult scenario to reproduce as it depends on the timing of the query and update requests.  I do not have a method to reproduce it outside of the combination of tools I've described, or I would provide such an example.

I will attach a jstack listing of the server in the deadlocked state.

**Screenshots**

**Context (please always complete the following information):**
 - OS: RHEL 7
 - eXist-db version: 5.3.0 (with updated log4j)
 - Java: openjdk 11.0.13 2021-10-19 LTS
OpenJDK Runtime Environment 18.9 (build 11.0.13+8-LTS)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.13+8-LTS, mixed mode, sharing)

**Additional context**
- How is eXist-db installed? JAR installer
- Any custom changes in e.g. `conf.xml`?

paulmer:
The jstack trace of the deadlocked server:  
[trace.txt](https://github.com/eXist-db/exist/files/8036325/trace.txt)

adamretter:
Without seeing the Lock Table,  I cannot be certain. However...

1. The thread executing the RPC connection (likely XML:DB Remote) is waiting to acquire a `WRITE_LOCK` on a document, to so it needs to acquire the lock with ID `0x0000000707025300`. If Path Locks for documents are not in use (which is the default, you can change this in `conf.xml`), then to do so it will have first acquired a READ or WRITE (depending on the operation) lock on the parent collection, and INTENTION locks on the ancestor collections. We can see that the thread executing the RPC in the process of trying to acquire the WRITE_LOCK has already acquired several locks, one of which is a lock with ID `0x000000073418aa78`.

[...]

Issue #4022, opened 2021-08-31
[BUG] Saving xquery code through oXygenXML or eXide may lead to server 'crash'


ahenket:
**Describe the bug**
It is unclear what the exact circumstances are but when you hit Save on updated xquery code from oXygenXML or (less common for me) in eXide you may bring the server to a coma like state where it runs at 0.3/0.4% CPU but will not respond to anything otherwise. This does not happen on every Save but is most common on libraries with much reuse.

The result is that only a force quit will stop the server which in turn leads to reindex and reindex takes up to 24 hours on our larger databases. Both the cause and the consequence of reindex should really be fixed

**Expected behavior**
No crash/outage should occur from simply hitting Save from oXygen (xmlrpc) or eXide. Reindex should preferably only be triggered when it's actually an index problem and if it happens it should not cause up a 1 day of downtime.

**To Reproduce**
Update code in libraries with lots of reuse.

**Screenshots**

**Context (please always complete the following information):**
 - OS: CentOS, macOS Big Sur
 - eXist-db version 5.3.0 production build
 - Java Version 1.8

**Additional context**
- We have deactivated the RestXQ trigger and modules from conf.xml to see if that would help
- Attached logging was created when eXist went into coma state on macOS Big Sur

[20210831-0847-eXist-db-5.3-after-save-oxygen.txt](https://github.com/eXist-db/exist/files/7083064/20210831-0847-eXist-db-5.3-after-save-oxygen.txt)

ahenket:
Update:

- Editing xqm files that are subject to circular imports are more likely to trigger this, than more standalone xqm files
- Hitting save on an xqm file that is also called by some other process is more likely to trigger this. E.g. while oXygen is validating some local xquery that imports an xqm, it is unsafe to hit Save on that xqm at the same time

mgraauw:
Some more info. I am able to reproduce this on my local system. Yes, trouble seems to come from 3 xqm's where each one imports the other 2. Saving from either Oxygen or eXide will hang exist, but saving from Vscode will not. What I see in Monex:
![image](https://user-images.githubusercontent.com/6449289/132240659-70c84c77-1fa7-4b8e-affe-9a3a3ac55fe6.png)

joewiz:
@ahenket @mgraauw Would it be possible to prepare a zip file containing a minimal example of the modules that lead to this situation?

mgraauw:
[...]

PR #6647, opened 2026-08-18
[bugfix] Release the stored query's READ_LOCK at end-of-compile over REST and XML-RPC

### LINKED ISSUE #6593: [bugfix] Stored XQuery execution over REST and XML-RPC holds the document READ_LOCK through execution and serialization
## Summary

When a stored XQuery is executed over **REST** or **XML-RPC**, the document
READ_LOCK acquired to resolve and compile the query is held for the **entire
request** — through execution *and* result serialization — instead of being
released once compilation is done. Holding the query document's shared lock while
the executor goes on to acquire further collection/document locks is one required
edge of the long-standing save-while-running deadlock family
(#5916, #4022, #4233).

## Lock lifecycle today

| Path | doc READ_LOCK during compile | during execute + serialize |
|---|---|---|
| REST (`RESTServer`) | held | **held** (released in `finally`, after `writeResults`) |
| XML-RPC (`RpcConnection.execute`/`executeT` via `readDocument` → `withDocument`) | held | **held** |
| `XQueryServlet` / URL-rewrite forwards | released at resolution | not held |
| RESTXQ URI compile | held | released right after compile |
| `import module` from db | held during that module's compile | not held |

Only **REST** and **XML-RPC** hold the lock across execution; the other stored
paths already release it at or before end-of-compile.

## Why it matters — the deadlock

An executor that parks on further collection/document locks while holding the
query document's S-lock is one required edge of the save-while-running deadlock
family (#5916, #4022, #4233): a concurrent save takes X-locks down the whole
ancestor path (single-writer default) and blocks on the document's write lock
while holding `X(/db)`; the executor's next `collection()` call then blocks on
`/db` — a permanent cycle that freezes the server (barging `MultiLock`s, no
deadlock detection, no timeouts). Dropping the query-document lock at
end-of-compile removes the executor edge entirely.

The window has recently grown: execute-without-read (#6586) means read-blind
callers now reach pool borrow/recompile while this lock is held (see the
companion pool-integrity issue), so the lock is held across a recompile on every
such request until that issue lands.

## Fix — release after compile

**Invariant.** A document READ_LOCK on a stored XQuery (main module or imported
library module) is held at most for *resolution + compilation* and is released
**immediately after compilation returns** — or immediately after a pool hit makes
compilation unnecessary. It is **never** held during execution or result
serialization.

Holding it *through* compile is retained deliberately: the source bytes are read
lock-free from the blob store (`NativeBroker.getBinaryResource` takes no document
lock), so the lock is what guarantees a consistent snapshot across resolution,
permission evaluation, and compile. Only the execution/serialization tail is
released early.

### Implementation

- **REST** (`RESTServer.executeXQuery` / `executeXProc`): take ownership of the
  `LockedDocument` and close it at the compiled-obtained point — after
  `pool.borrowCompiledXQuery` returns non-null, or after `xquery.compile`
  returns/throws — **before** `xquery.execute` and `writeResults`. Note the
  current structure after the `doGet`/`doPost` decomposition in #6562: the lock
  is owned up the chain (`serveResource` / `executePostExecutable` →
  `resolveExecutable` / `resolvePostExecutable`, which return the lock inside
  `ResolvedExecutable`, → `executeResource` → `executeXQuery`); ownership must be
  transferred explicitly so the caller-side `finally` blocks keep only a
  null-guarded close for paths that never reach execution.
- **XML-RPC** (`RpcConnection.execute` / `executeT`): stop wrapping
  compile + execute in `readDocument(...)` / `withDocument(READ_LOCK, ...)`.
  Resolve the document under lock, permission-check, build the `DBSource`,
  **close the lock**, then compile and run outside the locked scope.

## Behaviour change (release-note)

Storing a query that is currently executing via REST or XML-RPC no longer blocks
until the execution finishes: the store **completes immediately** and the
in-flight execution continues on the old source (blob refcounting keeps the old
bytes alive). This matches the long-standing behaviour of app requests via
`XQueryServlet`, and is the deliberate fix for the #5916 hang edge.

## Tests

- **Save during execution completes (deadlock regression).** Store a long-running
  query; invoke it via REST; while it runs, store a new version of the same
  document from a second thread → the store **completes** while the query is still
  executing, and the query finishes with old-source results. Same for XML-RPC
  `execute`. Guards the #5916 REST edge.
- **Lock released after compile.** During a query's execution phase, assert via
  `LockManager` / `LockTable` that no READ_LOCK is held on the query document;
  conversely assert it *is* held during compile.

## References

- Deadlock family: #5916, #4022, #4233.
- Related: #6586 (execute-without-read) enlarges the current lock window.
- Companion: the read-blind pool-integrity fix — #6594.

line-o:
> [!NOTE]
> **Stacked on #6635** — the first commit here is that PR's test-timing fix (`RESTExecuteWithoutReadTest` touches the same `RESTServer#executeXQuery` area). Please merge #6635 first; this PR then reduces to its own single commit. Review only the last commit: b332fbf60a.

### Description:

When a stored XQuery is executed over **REST** or **XML-RPC**, the document READ_LOCK acquired to resolve and compile the query was held for the **entire request** — through execution *and* result serialization — instead of being released once compilation is done. An executor that parks on further collection/document locks while holding the query document's shared lock is one required edge of the save-while-running deadlock family (#5916, #4022, #4233), and it also meant that storing a new version of a query blocked until every in-flight execution of it had finished. All other stored-query paths (`XQueryServlet`, RESTXQ, `import module` from db) already release the lock at or before end-of-compile.

**Invariant established:** the document READ_LOCK on a stored XQuery is held at most for *resolution + compilation* and is released immediately after compilation returns (or after a pool hit makes compilation unnecessary). It is never held during execution or result serialization.

### What Changed:

- **REST** (`RESTServer`): `executeXQuery` / `executeXProc` now receive the `LockedDocument` from their callers (`serveResource`/`executeResource`, `executePostExecutable`, `checkForXQueryTarget`) and close it in a `finally` around the compile block — after `pool.borrowCompiledXQuery` hits, or right after `xquery.compile` returns or throws — **before** `xquery.execute` and `writeResults`. The caller-side `finally` blocks are kept unchanged as a backstop for paths that fail before execution; `ManagedLock#close` is idempotent, so the second close is a no-op.
- **XML-RPC** (`RpcConnection`): `execute` / `executeT` no longer wrap compile + execute in `readDocument(...)`. A new `resolveStoredXQuery` helper resolves the document, checks `READ | EXECUTE`, and builds the `DBSource` under the READ_LOCK, which is released on return; the query is compiled and run outside the locked scope. Side fix: the old code cast to `BinaryDocument` *before* the binary-type check, so pointing `execute` at an XML document threw a raw `ClassCastException` — it now throws the intended "is not a binary resource" `EXistException`.

### Behaviour change (release note):

Storing a query that is currently executing via REST or XML-RPC no longer blocks until the execution finishes: the store completes immediately and the in-flight execution continues on the old source (blob refcounting keeps the old bytes alive). This matches the long-standing behaviour of app requests via `XQueryServlet`.

### Reference:

Closes https://github.com/eXist-db/exist/issues/6593
Deadlock family: #5916, #4022, #4233. Related: #6586, #6594.

### Type of tests:

New JUnit integration test `org.exist.http.StoredQueryLockReleaseTest` covering both interfaces. A stored query signals the start of its execution phase by storing a marker document, then parks polling for a release document; the test asserts via `LockTable` that no lock on the query document outlives compilation, stores a new version of the very query being executed (must complete immediately — the #5916 hang edge), then releases the executor and checks it finishes on the old source while the next run picks up the new one. Verified **red on unpatched code** (both tests fail at the lock assertion, which is deliberately placed before the concurrent save so a regressed server fails cleanly instead of deadlocking CI) and **green with the fix**.

Regression suites run locally, all green: `StoredQueryLockReleaseTest` (2), `RESTExecuteWithoutReadTest` (13), `RESTServiceTest` (47), `XmlRpcTest` (46), `XMLDBSecurityTest` (174), `ExecuteWithoutReadTest` (8).

https://claude.ai/code/session_018pfpnr2M2v9c6XNvC9bssh

### DIFF
--- a/exist-core/src/test/java/org/exist/http/StoredQueryLockReleaseTest.java
+++ b/exist-core/src/test/java/org/exist/http/StoredQueryLockReleaseTest.java
@@ -1,0 +1,292 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.http;
+
+import org.apache.commons.codec.binary.Base64;
+import org.apache.xmlrpc.client.XmlRpcClient;
+import org.apache.xmlrpc.client.XmlRpcClientConfigImpl;
+import org.exist.collections.Collection;
+import org.exist.dom.persistent.LockedDocument;
+import org.exist.storage.BrokerPool;
+import org.exist.storage.DBBroker;
+import org.exist.storage.lock.Lock.LockMode;
+import org.exist.storage.lock.LockTable;
+import org.exist.storage.txn.Txn;
+import org.exist.test.ExistEmbeddedServer;
+import org.exist.test.ExistWebServer;
+import org.exist.util.MimeType;
+import org.exist.util.StringInputSource;
+import org.exist.xmldb.XmldbURI;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * The document READ_LOCK taken on a stored XQuery to resolve and compile it must be released at
+ * end-of-compile — never held through execution and result serialization. An executor which parks
+ * on further collection/document locks while holding the query document's shared lock is one
+ * required edge of the save-while-running deadlock family, and it also means a concurrent store of
+ * that query blocks until the execution finishes.
+ *
+ * Each test runs a stored query which signals that its execution phase has begun (by storing a
+ * marker document) and then polls for a release document; while it is parked, the test asserts
+ * that no lock is still held on the query document, stores a new version of the very query being
+ * executed — which must complete immediately — and only then releases the executor, which finishes
+ * on the old source.
+ *
+ * See https://github.com/eXist-db/exist/issues/6593
+ */
+public class StoredQueryLockReleaseTest {
+
+    @ClassRule
+    public static final ExistWebServer existWebServer = new ExistWebServer(true, false, true, true);
+
+    @ClassRule
+    public static final ExistEmbeddedServer existEmbeddedServer = new ExistEmbeddedServer(true, true);
+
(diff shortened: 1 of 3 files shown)


exist-core/src/test/java/org/exist/http/StoredQueryLockReleaseTest.java
```diff
@@ -0,0 +1,292 @@
...
+    private static String credentials;
+
+    /**
+     * Stores a marker document to signal that its execution phase (and therefore its compilation)
+     * has finished, then parks polling for a release document, so the test fully controls how long
+     * the execution phase lasts. The poll is bounded so that a deadlocked run under a regressed
+     * server fails instead of hanging forever.
+     */
+    private static String awaitQuery(final XmldbURI startedUri, final XmldbURI releaseUri) {
+        return "xquery version \"3.1\";\n" +
```
duncdrum: please use text block here

line-o: done

reinhapa (approved):

duncdrum (approved):

PR #6649, opened 2026-08-18
[bugfix] Guard against null Path.of(getRealPath()/getPathTranslated()) across servlet/urlrewrite code

### LINKED ISSUE #6615: [BUG] Unhandled NullPointerException in XQueryServlet.process() during URL rewrite / forward
Description

When an XQuery request is handled via XQueryURLRewrite / Forward (e.g. mapping to a database path like /db/apps/optimize.xql), XQueryServlet.process() throws an unhandled java.lang.NullPointerException at Path.of(), resulting in a raw HTTP 500 error page from Jetty.

Rather than gracefully handling non-existent disk paths (or virtual DB-only paths) and falling back to database source resolution (or returning a proper 404/500 HTTP response), the servlet passes a null path parameter directly to java.nio.file.Path.of().

Stacktrace

HTTP ERROR 500 jakarta.servlet.ServletException: An error occurred while processing request to /exist/db/apps/optimize.xql: null
URI:http://localhost:8080/exist/db/apps/optimize.xql
STATUS:500
MESSAGE:jakarta.servlet.ServletException: An error occurred while processing request to /exist/db/apps/optimize.xql: null
SERVLET:XQueryURLRewrite
CAUSED BY:java.lang.NullPointerException
Caused by:
jakarta.servlet.ServletException: An error occurred while processing request to /exist/db/apps/optimize.xql: null
	at org.exist.http.urlrewrite.XQueryURLRewrite.service(XQueryURLRewrite.java:369)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:614)
	at org.eclipse.jetty.ee10.servlet.ServletHolder$NotAsync.service(ServletHolder.java:1388)
	...
	at java.base/java.util.Objects.requireNonNull(Unknown Source)
	at java.base/sun.nio.fs.UnixFileSystem.getPath(Unknown Source)
	at java.base/java.nio.file.Path.of(Unknown Source)
	at org.exist.http.servlets.XQueryServlet.process(XQueryServlet.java:361)
	at org.exist.http.servlets.XQueryServlet.doGet(XQueryServlet.java:162)
	at org.exist.http.urlrewrite.Forward.doRewrite(Forward.java:50)
	at org.exist.http.urlrewrite.XQueryURLRewrite.doRewrite(XQueryURLRewrite.java:528)
	at org.exist.http.urlrewrite.XQueryURLRewrite.service(XQueryURLRewrite.java:343)
	...

Cause Analysis

Around line 361 in XQueryServlet.java, the method process() retrieves a file path (likely via request.getPathTranslated() or servletContext.getRealPath()).

When forwarding

[XQueryServletNullPathTest.java](https://github.com/user-attachments/files/30545614/XQueryServletNullPathTest.java)

to a resource residing in the database repository rather than on the host filesystem, this lookup returns null. The code directly passes this null value into java.nio.file.Path.of(...). Since JDK's Path.of() uses Objects.requireNonNull(), a NullPointerException is immediately raised before eXist can resolve the resource against the database collection (/db/...).
Expected Behavior

    XQueryServlet should perform a null-check on the disk path prior to calling Path.of().

    If the physical path is null, it should gracefully fall back to resolving the path within the XML database repository (BrokerDB / SourceFactory).

    If the resource cannot be found in the database either, it should return a clean, handled HTTP 404 error instead of letting an unhandled NPE escape to the Jetty servlet container.

duncdrum:
## Summary

- Fixes an unhandled `NullPointerException` in `XQueryServlet.process()`: the Servlet API permits `HttpServletRequest#getPathTranslated()` and `ServletContext#getRealPath(String)` to return `null` when the container can't map a request to a disk path (e.g. `<exist:forward servlet="XQueryServlet">` to a database-only path like `/db/apps/optimize.xql`). That `null` was passed straight into `Path.of()`, crashing with a raw HTTP 500 instead of resolving the resource or returning a clean 404.
- Fixes the dead code in `XQueryURLRewrite.RequestWrapper#getPathTranslated()` that was silently discarding a real path and always returning `null` — the actual mechanism that hands `XQueryServlet` a null path in the first place.
- Applies the same null-guard to five more call sites found to share the identical defect shape (`Path.of()`/`Paths.get()` fed directly from `getRealPath()`/`getPathTranslated()` with no null check): `XQueryURLRewrite#findSourceFromFs()`, `XSLTServlet#getCurrentDir()`, `RewriteConfig#configure()`, `RedirectorServlet#service()`, and `ScaleImageJAI#getAbsolutePath()`.

## What Changed

- `XQueryServlet.java` — when no readable file is found on disk (null or nonexistent path), falls back to resolving the resource in the database via `SourceFactory` before returning a clean 404.
- `XQueryURLRewrite.java` — fixed `RequestWrapper#getPathTranslated()`'s discarded return value; added a null guard in `findSourceFromFs()`.
- `XSLTServlet.java` — `getCurrentDir()` now returns `null` instead of crashing; both callers treat that the same as their existing "not found" case.
- `RewriteConfig.java` — a null real path for the webapp root is now treated like a missing/unreadable `controller-config.xml` (skip, don't crash).
- `RedirectorServlet.java` — fails fast with the same `ServletException` message the neighboring readability check already uses.
- `extensions/images/ScaleImageJAI.java` — fails fast at `init()` with a clear `ServletException`. Note: this module is commented out of `extensions/pom.xml` (`Not used`) and doesn't currently compile against `develop` for unrelated reasons (a missing `NativeBroker` method), so this change is verified by inspection only, not a green build.

## Test Plan

- [x] `XQueryServletNullPathTest` — reproduces the exact reported NPE (confirmed RED against unfixed code), and a second test proving the database fallback resolves and executes a database-only resource end-to-end.
- [x] `XQueryURLRewriteRequestWrapperTest` — covers both branches of the fixed `getPathTranslated()`.
- [x] `XQueryURLRewriteFindSourceFromFsTest`, `XSLTServletGetCurrentDirTest` — reflection-based tests for the two most reachable of the five additional sites, confirmed RED against unfixed code with the identical NPE stack trace shape before fixing.
- [x] Full `exist-core` `http`/`urlrewrite`/security regression suite passes with no failures.

Closes https://github.com/eXist-db/exist/issues/6615

### DIFF
--- a/exist-core/src/main/java/org/exist/http/servlets/RedirectorServlet.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/RedirectorServlet.java
@@ -160,6 +160,11 @@ public class RedirectorServlet extends AbstractExistHttpServlet {
             }
         // Try to find the XQuery
         final String qpath = getServletContext().getRealPath(query);
+        // the Servlet API permits getRealPath() to return null when the webapp is not exploded
+        // on disk (e.g. served from a packed/embedded context).
+        if (qpath == null) {
+            throw new ServletException("Cannot read XQuery source from " + query);
+        }
         final Path p = Path.of(qpath);
         if (!(Files.isReadable(p) && Files.isRegularFile(p))) {
             throw new ServletException("Cannot read XQuery source from " + p.toAbsolutePath());
--- a/exist-core/src/main/java/org/exist/http/servlets/XQueryServlet.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/XQueryServlet.java
@@ -240,12 +240,17 @@ public class XQueryServlet extends AbstractExistHttpServlet {
     protected void process(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
         //first, adjust the path
         String path = request.getPathTranslated();
-        if(path == null) {
-            path = request.getRequestURI().substring(request.getContextPath().length());
-            final int p = path.lastIndexOf(';');
+        // the request path relative to the context, e.g. "/db/apps/optimize.xql" -- kept around
+        // (regardless of whether pathTranslated already gave us a disk path) so that a request
+        // which cannot be resolved to a file on disk can still be tried against the database.
+        String contextRelativePath = request.getRequestURI().substring(request.getContextPath().length());
+        {
+            final int p = contextRelativePath.lastIndexOf(';');
             if(p != Constants.STRING_NOT_FOUND)
-                {path = path.substring(0, p);}
-            path = getServletContext().getRealPath(path);
+                {contextRelativePath = contextRelativePath.substring(0, p);}
+        }
+        if(path == null) {
+            path = getServletContext().getRealPath(contextRelativePath);
         }
         
         //second, perform descriptor actions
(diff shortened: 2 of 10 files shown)


reinhapa (approved):

line-o:
I am in favour of pulling this in. What I do not understand is why we fallback to resolving database internal resources when they could not be found on disk. Is this advisable?
Maybe it will be easier for me to understand why if I get an example how to trigger this NPE but I have to have an embedded DB, right?

duncdrum:
@line-o the original report was about the XQuery servlet, so a non existing controller.xq on the filesystem will NPE. The fallback is sensible since url forwarding or the stored xquery might point to a db library, which should similarly not NPE. If the xq can't be found in either location we return 4xx but do not crash. 

(see the first commit message)

> when a controller.xql forwards (<exist:forward servlet=
"XQueryServlet">) to an absolute database-only path such as
/db/apps/optimize.xql, which has no corresponding file on disk.

The shape translates to XSL servlet and others, we already had correct logic for JMX, it was simply not applied consistently.

line-o (approved):
Thanks for the explanation. Makes all sense now

PR #6661, opened 2026-08-28
[bugfix] Pin transitive dependencies to resolve dependencyConvergence

dizzzz:
Extend maven-enforcer-plugin configuration to be sure that all transitive dependencies resolve to the latest version.

### DIFF
--- a/exist-parent/pom.xml
+++ b/exist-parent/pom.xml
@@ -193,6 +193,13 @@
                 <version>${jaxb.api.version}</version>
             </dependency>
 
+            <!-- Pin jakarta.activation-api:2.1.4: jaxb-api/angus-activation need 2.1.4; jaxb-runtime/jaxb-core pulls 2.1.1. -->
+            <dependency>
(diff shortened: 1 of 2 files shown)


dizzzz:
strangely local build is ok

exist-parent/pom.xml
```diff
@@ -317,6 +324,13 @@
                 <version>1.22.1</version>
             </dependency>
 
+            <!-- Pin commons-logging:1.4.0: log4j-jcl pulls 1.3.5, xmlrpc-server pulls 1.1. -->
```
dizzzz, line 327: this is rather a big step?

duncdrum:
I don't understand what this is doing.

dizzzz:
@duncdrum  for the enforcer plugin this configuration item is set:  <dependencyConvergence/>

it makes sure that all transient dependencies resolve to the same versions. If not, the build breaks.

The rest of the additions is about enforcing the latest versions of libs. A bit of context is added for which deps it was actually needed.

Issue #3869, opened 2021-05-03
[feature] Extend documentation of xmldb:create-collection() for recursion


ahenket:
The function `xmldb:create-collection($target-collection-uri as xs:string,  $new-collection as xs:string)` supports recursion on the $new-collection, which makes it possible to say:

```xquery
xmldb:create-collection(repo:get-root(), 'a/b/c/d')
```

This recursively creates a/b/c/d as needed and returns `/db/apps/a/b/c/d`. This very much appreciated behavior however is undocumented.

As came up in the [Slack community channel](https://exist-db.slack.com/archives/CG2MRUZ35/p1619994459095900), it seems like a good idea to add documentation and likely also add tests for this feature.

duncdrum:
I m going to go out on a limb and say this probably also needs a test.

AzazelSensei:
Opened https://github.com/eXist-db/exist/pull/6666 — docs now mention recursive path creation, and there is an XQSuite for it.

PR #6666, opened 2026-08-31
[doc] Document recursive xmldb:create-collection() #3869

AzazelSensei:
### Description:

`xmldb:create-collection()` already creates missing path segments. The function docs now say that, and there is an XQSuite for a nested path, intermediates, and creating the same path twice.

### Reference:

https://github.com/eXist-db/exist/issues/3869

### Type of tests:

XQSuite in `exist-core/src/test/xquery/xmldb/create-collection-tests.xql`


duncdrum (approved):
Thank you @AzazelSensei , nice find. Much appreciated.

joewiz (approved):

Issue #6189, opened 2026-03-27
[feature] add exist-db specific schema files to distribution output


duncdrum:
Our tooling uses a number of exist specific `xsd` schema files, and we ship schemas for e.g. svg, Mondial, mathml,  in  `exist/ect/webapp/WEB-INF/entities` . 

What's missing is schemas for exist-db specific files, like `collection.xonf` or `repo.xml` . So users who just modified or copied their `conf.xml` have no way to validate it from within the running instance, to check compatibility. Furthermore, as can be seen e.g. here [eXide#784](https://github.com/eXist-db/eXide/pull/784). Every time we change one of these schemas we need to a) copy the file into consuming packages, and avoid drift b) tie the consuming package to a single version of exist-db. 

I want to suggest that we ship exist-db specific schemas as part of the distribution. We could also expose them via api, but at a minimum consuming packages should not require copies of exists native schemas, but locate and utilise the schemas that apply to a specific instance. 

My likely incomplete list:
- repo.xml
- collection.xconf
- exist.xml(?)
- conf.xml

duncdrum:
shout out to @adamretter from [2007](https://github.com/eXist-db/exist/commit/658d7f8d1bc05dfc10cbb9dc7775bddd73bce9c9) with exist 1.2 . The answer is: Yes it should

adamretter:
@duncdrum all the missing schemas you are looking for should already be in the `schemas/` folder where I created them

PR #6687, opened 2026-09-08
[bugfix] cover native collection.xconf.xsd via OASIS catalog + jaxp

### LINKED ISSUE #6686: validation:jaxp + OASIS catalog does not apply native XSD 1.1 schemas from $EXIST_HOME/schema
After #6528, `$EXIST_HOME/schema/` and `WEB-INF/catalog.xml` map `http://exist-db.org/collection-config/1.0` → `collection.xconf.xsd` for tools (e.g. eXide — eXist-db/eXide#842).

`validation:jaxp-report` with the system catalog (or an OASIS catalog pointing at that `file:` URI) fails with `cvc-elt.1.a` on a minimal legal `collection.xconf`. The same instance is `valid` via `validation:jaxv-report` + the XSD 1.1 language URI against the same schema file.

Existing tests miss this path: `JaxpXsdCatalogTest` XSD 1.1 coverage uses directory-search over `/db/.../`; `CollectionConfigSchema6000Test` / Maven validate use raw `SchemaFactory`, not `validation:jaxp`.

Repro / failing test: #6687 (expect-red until fixed).

duncdrum:
### Description:

Adds `NativeSchemaJaxpCatalogGapTest` to close the integration gap left by #6528, and fixes the gap it found:

- **Green control:** `validation:jaxv-report` + XSD 1.1 language URI against `schema/collection.xconf.xsd` (valid minimal `collection.xconf`; empty one fails `xs:assert`).
- **Previously missing path (now fixed):** `validation:jaxp-report` via an OASIS catalog whose `<uri>` maps `http://exist-db.org/collection-config/1.0` to an absolute `file:` URI of that same native XSD — mirrors the `WEB-INF/catalog.xml` → `$EXIST_HOME/schema/` wiring tools are meant to use.

**Root cause:** `Jaxp#newXsd11Validator` wired an OASIS/system catalog's `org.xmlresolver.Resolver` in directly as the XSD 1.1 validator's `LSResourceResolver`. That resolver's `resolveResource()` requires a non-null `systemId` before it will consult the catalog at all, so a `<uri>` entry keyed purely by namespace was invisible whenever the instance carried no `schemaLocation` hint — even though the same catalog entry already worked for the default SAX pipeline, which falls back to the namespace as the systemId (see `XercesXmlResolverAdapter#resolveEntity`).

**Fix:** `XercesXmlResolverAdapter#asLSResourceResolver` applies that same namespace-as-systemId fallback. Code review of the first fix found `validation:jaxv()`'s SchemaFactory catalog wiring had the identical unfixed gap (same raw-`Resolver` pass-through, for `xs:import`/`xs:include` resolution during schema compilation), so it's fixed the same way and both call sites now share a single `asLSResourceResolver(LSResourceResolver)` dispatch instead of duplicating the `instanceof` check.

Existing coverage did not catch this: `JaxpXsdCatalogTest` XSD 1.1 cases use directory-search over `/db/.../`, which `SearchResourceResolver` already resolves purely by namespace; `CollectionConfigSchema6000Test` / Maven validate use raw `SchemaFactory`, not `validation:jaxp`.

### What Changed

- `exist-core/src/main/java/org/exist/resolver/XercesXmlResolverAdapter.java` — new `asLSResourceResolver(Resolver)` wrapping an `org.xmlresolver.Resolver` as an `LSResourceResolver` with a namespace→systemId fallback, plus an `asLSResourceResolver(LSResourceResolver)` overload that dispatches on the concrete resolver type (only wrapping when it's a `Resolver`) so both call sites below share one implementation.
- `exist-core/src/main/java/org/exist/xquery/functions/validation/Jaxp.java` — `newXsd11Validator` now routes its catalog resolver through the new adapter before handing it to the XSD 1.1 `Validator`.
- `exist-core/src/main/java/org/exist/xquery/functions/validation/Jaxv.java` — same fix for the `SchemaFactory#setResourceResolver` catalog wiring.
- `exist-core/src/test/java/org/exist/xquery/functions/validate/NativeSchemaJaxpCatalogGapTest.java` — the regression test; updated comments now that the gap is closed, and escapes the generated catalog's `file:` URI attribute (`Path#toUri()` doesn't XML-escape characters like `&` a checkout path could contain).

### Reference:

- Closes https://github.com/eXist-db/exist/issues/6686
- Related: #6528 / #6189, eXist-db/eXide#842

### Type of tests:

JUnit (`ExistXmldbEmbeddedServer`) in `exist-core`.

```
mvn -pl exist-core -Dtest=NativeSchemaJaxpCatalogGapTest test
```

### Test Plan

- [x] `NativeSchemaJaxpCatalogGapTest` — 3/3 green (was 2/3, 1 expected-red)
- [x] `JaxpXsdCatalogTest` — 12/12 green (no regression)
- [x] Full `validate`/`validation` package (`JaxpParseTest`, `JaxvTest`, `JaxpDtdCatalogTest`, `JaxpSchemaLocationSecurityTest`, `JaxpXsd11DetectionCacheTest`, `IsMissingElementDeclarationTest`, Jing*/Parse* suites, etc.) — 77/77 green
- [x] CI: all checks green except a pre-existing network flake (`unparsed-text-from-url`, unrelated fn:unparsed-text() GitHub-raw fetch)

### DIFF
--- a/exist-core/src/main/java/org/exist/resolver/XercesXmlResolverAdapter.java
+++ b/exist-core/src/main/java/org/exist/resolver/XercesXmlResolverAdapter.java
@@ -43,6 +43,7 @@ import org.apache.xerces.xni.grammars.XMLSchemaDescription;
 import org.apache.xerces.xni.parser.XMLEntityResolver;
 import org.apache.xerces.xni.parser.XMLInputSource;
 import org.exist.util.XMLReaderObjectFactory;
+import org.w3c.dom.ls.LSResourceResolver;
 import org.xml.sax.*;
 import org.xmlresolver.Resolver;
 
@@ -164,5 +165,29 @@ public class XercesXmlResolverAdapter implements XMLEntityResolver {
      */
     public Resolver getResolver() {
         return resolver;
+    }
+
+    /**
+     * Wraps {@code resolver} as an {@link LSResourceResolver}, falling back to {@code namespaceURI}
+     * as the system id whenever the caller doesn't supply one -- the same fallback {@link
+     * #resolveEntity} already performs for the XNI/SAX pipeline (see its {@code systemId} lookup).
+     *
+     * <p>This fallback matters because {@link Resolver#resolveResource} requires a non-null system
+     * id before it will consult its catalog at all; without it, an OASIS catalog's {@code <uri>}
+     * entry keyed purely by namespace (no {@code schemaLocation} hint on the instance) is invisible
+     * to the XSD 1.1 dynamic-discovery {@link javax.xml.validation.Validator}, which calls {@link
+     * LSResourceResolver#resolveResource} with the root element's namespace but no system id --
+     * even though the same catalog entry already works for the default SAX pipeline via {@link
+     * #resolveEntity}.</p>
+     *
+     * @param resolver the resolver to wrap.
+     *
+     * @return an {@link LSResourceResolver} view of {@code resolver} with the namespace fallback applied.
+     */
+    public static LSResourceResolver asLSResourceResolver(final Resolver resolver) {
+        return (type, namespaceURI, publicId, systemId, baseURI) -> {
+            final String effectiveSystemId = systemId != null ? systemId : namespaceURI;
+            return resolver.resolveResource(type, namespaceURI, publicId, effectiveSystemId, baseURI);
+        };
     }
 }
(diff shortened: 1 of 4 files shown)


exist-core/src/main/java/org/exist/resolver/XercesXmlResolverAdapter.java
```diff
@@ -165,4 +166,45 @@ public static void setXmlReaderEntityResolver(final XMLReader xmlReader, @Nullab
...
+     * #resolveEntity}.</p>
+     *
+     * @param resolver the resolver to wrap.
+     *
+     * @return an {@link LSResourceResolver} view of {@code resolver} with the namespace fallback applied.
+     */
+    public static LSResourceResolver asLSResourceResolver(final Resolver resolver) {
+        return (type, namespaceURI, publicId, systemId, baseURI) -> {
+            final String effectiveSystemId = systemId != null ? systemId : namespaceURI;
+            return resolver.resolveResource(type, namespaceURI, publicId, effectiveSystemId, baseURI);
```
dizzzz, line 190: nice!

dizzzz (approved):

line-o (approved):
LGTM <3

PR #6691, opened 2026-09-10
[bugfix] Fix CachingResponseWrapper gaps exposed by the Jetty 12 upgrade

### LINKED ISSUE #6667: [BUG] URL Rewriting regression with simple pipelines throwing "getOutputStream cannnot be called after getWriter"
**Description**

Pipeline generated with controller.xql using URL Rewriting pipeline syntax with an XQuery step generating XML data to be fed to an XSLT step throws en exception such as : 

"HTTP ERROR 500 jakarta.servlet.ServletException: An error occurred while processing request to /exist/projects/test: getOutputStream cannnot be called after getWriter"

Looks like this comes from XQueryURLRewrite.java, line 1365:

                throw new IOException("getOutputStream cannnot be called after getWriter");

**Expected behavior**

Opening http://localhost:8080/exist/projects/test in the browser should display "Hello Bob" (with the example given below)

This is observed with "Version 7.0.0-SNAPSHOT" of eXist-DB running in a Docker container

This workks correctly with "Version 6.4.1" of  eXist-DB running in a Docker container

**To Reproduce**

Create a docker container with "existdb:latest" image from Docker Hub

You can use the compose.yml file below that mounts the host  "exist-projects" folder into "/exist/etc/webapp/projects" container folder

```
services:
  exist:
    image: existdb/existdb:latest
    container_name: exist7 
    ports:
        - 8080:8080
        - 8443:8443
    networks:
      - www
    volumes:
      - type: bind
        source: ./exist-projects
        target: /exist/etc/webapp/projects
    
networks:
  www:
    name: www
    driver: bridge
```

Run the container then create on your host three files:

exist-projects/test/controller.xql

```
xquery version "3.1";

declare namespace exist = "http://exist.sourceforge.net/NS/exist";

<exist:dispatch>
  <exist:forward url="/projects/test/A.xql">
    <exist:set-header name="Cache-Control" value="no-cache"/>
    <exist:set-header name="Pragma" value="no-cache"/>
  </exist:forward>
  <exist:view>
    <exist:forward servlet="XSLTServlet">
      <exist:set-attribute name="xslt.stylesheet" value="/projects/test/B.xsl"/>
    </exist:forward>
  </exist:view>
  <exist:cache-control cache="false"/>
</exist:dispatch>
```

exist-projects/test/A.xql

```
xquery version "3.1";

declare option exist:serialize "method=xhtml media-type=text/html indent=yes";

<record>
  <name>Bob</name>
</record>
```

exist-projects/test/B.xsl

```
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform"   xmlns="http://www.w3.org/1999/xhtml">

  <xsl:output method="xml" media-type="text/html" omit-xml-declaration="yes" indent="no"/>
  
  <xsl:template match="/record">
    <p>Hello <xsl:value-of select="name"/></p>
  </xsl:template>             
  
</xsl:stylesheet>
```

Open http://localhost:9090/exist/projects/test, you should get an exception when using "Version 7.0.0-SNAPSHOT" image, but not with "Version 6.4.1" image

**Context**

eXist 7.0.0-SNAPSHOT · Linux 6.12.76-linuxkit aarch64 · Java 21.0.11 · exist 

- Build 2026-07-06T10:23:59Z
- Encoding UTF-8
- Locale en
- eXist home /exist

**Additional context**

- eXist-db installed in Docker (see compose.yml file above)
- no change to default eXist-DB configuration

### LINKED ISSUE #6669: [BUG] URL Rewriting regression with simple pipelines returing error 500 like "written x > z content-length"
**Description**

Pipeline generated with controller.xql using URL Rewriting pipeline syntax with a first step serving a static XML file to be fed to a second XQuery step generating some HTML content returns an error 500 such as : 

```
<h2>HTTP ERROR 500 jakarta.servlet.ServletException: An error occurred 
while processing request to /exist/projects/test: written 205 &gt; 61 
content-length</h2>

URI: http://localhost:9090/exist/projects/test
STATUS: 500
MESSAGE: jakarta.servlet.ServletException: An error  occurred while processing request to /exist/projects/test: written 205  > 61 content-length
SERVLET: XQueryURLRewrite
CAUSED BY: jakarta.servlet.ServletException: An error  occurred while processing request to /exist/projects/test: written 205  > 61 content-length
CAUSED BY: java.io.IOException: written 205 > 61 content-length
```

**Expected behavior**

Opening http://localhost:8080/exist/projects/test in the browser should display "Hello Bob, how do you do today Bob ?" (with the example given below)

This is observed with "Version 7.0.0-SNAPSHOT" of eXist-DB running in a Docker container

This works correctly with "Version 6.4.1" of eXist-DB running in a Docker container

**To Reproduce**

Create a docker container with "existdb:latest" image from Docker Hub

You can use the compose.yml file below that mounts the host "exist-projects" folder into "/exist/etc/webapp/projects" container folder

```
services:
  exist:
    image: existdb/existdb:latest
    container_name: exist7 
    ports:
        - 8080:8080
        - 8443:8443
    networks:
      - www
    volumes:
      - type: bind
        source: ./exist-projects
        target: /exist/etc/webapp/projects
    
networks:
  www:
    name: www
    driver: bridge
```

Run the container then create on your host three files:

exist-projects/test/controller.xql

```
xquery version "3.1";

declare namespace exist = "http://exist.sourceforge.net/NS/exist";

if (contains($exist:path, 'A.xml')) then
  <ignore xmlns="http://exist.sourceforge.net/NS/exist">
    <cache-control cache="no"/>
  </ignore>
else
  <exist:dispatch>
    <exist:forward url="/projects/test/A.xml">
      <exist:set-header name="Cache-Control" value="no-cache"/>
      <exist:set-header name="Pragma" value="no-cache"/>
    </exist:forward>
    <exist:view>
      <exist:forward url="/projects/test/B.xql"/>
    </exist:view>  
    <exist:cache-control cache="no"/>
  </exist:dispatch>
```

exist-projects/test/A.xml

```
<record>
  <name>Bob</name>
</record>
```

exist-projects/test/B.xql

```
xquery version "3.1";

declare option exist:serialize "method=xhtml media-type=text/html indent=yes";

let $data := request:get-data()
return
    <p>Hello { $data//name/text() }, how do you do today { $data//name/text() } ?</p>
```

Open http://localhost:8080/exist/projects/test, you should get the error when using "Version 7.0.0-SNAPSHOT" image, but not with "Version 6.4.1" image

Note that to be sure to get the error a condition is that B.xql output is longer than A.xml size, otherwise it works. You can add more chars to B.xql output to be sure you get the error.

Note that if you call response:set-header( "Content-Length", "43") in B.xql (or any number greater than B output length), you get the correct behavior, maybe because you overwrite the wrong response "Content-Length" header set in step one.

eXist 7.0.0-SNAPSHOT · Linux 6.12.76-linuxkit aarch64 · Java 21.0.11 · exist

* Build 2026-07-06T10:23:59Z
* Encoding UTF-8
* Locale en
* eXist home /exist

**Additional context**

* eXist-db installed in Docker (see compose.yml file above)
* no change to default eXist-DB configuration

duncdrum:
## Summary

Fixes two URL-rewrite pipeline regressions introduced by the Jetty 11→12 EE10 upgrade (`9a9a6a4b97`). Both trace to the same class, `XQueryURLRewrite$CachingResponseWrapper`, which buffers a controller.xql pipeline step's output so a following `<exist:view>` step can rewrite the body before anything reaches the client. It was written against Jetty ≤11's behavior and never fully intercepted the `HttpServletResponse` contract — Jetty 12 exercises the gaps.

- Closes #6667 — a `forward` step followed by a view forwarding to `XSLTServlet` threw `"getOutputStream cannnot be called after getWriter"` (500).
- Closes #6669 — a `forward` step serving a static file followed by a longer view output threw `"written x > z content-length"` (500).

## What Changed

**Commit 1 (red):** Two HTTP integration tests reproducing each issue end-to-end against unfixed code, using the exact controller.xql pipeline shapes from the bug reports.

**Commit 2 (fix):** `exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java`, `CachingResponseWrapper`:
- `getWriter()`/`getOutputStream()` now throw `IllegalStateException` (per the `ServletResponse` spec) instead of `IOException` on the mutual-exclusion violation, tracked via an explicit `OutputMode` enum instead of overloading the `sos` field as both sentinel and backing store. Jetty 12's `Dispatcher.forward()` (confirmed by disassembly) specifically catches `IllegalStateException` to fall back from `getOutputStream()` to `getWriter()` when closing the response after a forwarded servlet returns — the wrong exception type was escaping as an uncaught 500. The new tracking also makes a repeat call to the same method idempotent, which that fallback relies on.
- Added a cache-aware override for `setContentLengthLong(long)` and `addHeader("Content-Length", ...)`, neither of which was previously guarded. A static resource forward step (served via Jetty's default/ResourceServlet, which sets Content-Length through `addHeader()` rather than `setContentLength(int)`) was leaking its Content-Length directly onto the real response, which a later view step's longer output then violated.

## Root-Cause Verification

Root causes were confirmed empirically, not just by inspection: added temporary tracing to `CachingResponseWrapper` and disassembled the relevant Jetty 12 `Dispatcher.forward()` bytecode to see exactly which method Jetty calls and which exception type it catches. Reproduction test output matched the reported error strings verbatim (`written 43 > 37 content-length` locally vs. `written 205 > 61 content-length` in the issue — same shape, different byte counts).

## Test Plan

- [x] `URLRewriteXSLTViewPipelineTest` — fails on unfixed code with the exact #6667 error, passes after the fix.
- [x] `URLRewriteContentLengthViewPipelineTest` — fails on unfixed code with the exact #6669 error, passes after the fix.
- [x] Full `org.exist.http.urlrewrite` test package (22 tests across `ControllerTest`, `URLRewriteViewPipelineTest`, `XQueryURLRewriteTest`, `MultipartMethodControllerTest`, `IfModifiedSinceHandoverControllerTest`, etc.) — all passing, no regressions.

### DIFF
--- a/exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
+++ b/exist-core/src/main/java/org/exist/http/urlrewrite/XQueryURLRewrite.java
@@ -1368,9 +1368,15 @@ public class XQueryURLRewrite extends HttpServlet {
                 throw new IllegalStateException("getWriter cannot be called after getOutputStream");
             }
             if (outputMode == OutputMode.NONE) {
+                // Only commit outputMode/sos/writer once construction has fully succeeded -- if
+                // getCharacterEncoding() names a charset the JVM doesn't support, OutputStreamWriter
+                // throws, and a half-committed state here (mode flipped but writer still null) would
+                // make every subsequent getWriter() call silently return null instead of retrying.
+                final CachingServletOutputStream newSos = new CachingServletOutputStream();
+                final PrintWriter newWriter = new PrintWriter(new OutputStreamWriter(newSos, getCharacterEncoding()));
+                sos = newSos;
+                writer = newWriter;
                 outputMode = OutputMode.WRITER;
-                sos = new CachingServletOutputStream();
-                writer = new PrintWriter(new OutputStreamWriter(sos, getCharacterEncoding()));
             }
             return writer;
         }
@@ -1385,8 +1391,8 @@ public class XQueryURLRewrite extends HttpServlet {
                 throw new IllegalStateException("getOutputStream cannot be called after getWriter");
             }
             if (outputMode == OutputMode.NONE) {
-                outputMode = OutputMode.STREAM;
                 sos = new CachingServletOutputStream();
+                outputMode = OutputMode.STREAM;
             }
             return sos;
         }
@@ -1411,12 +1417,25 @@ public class XQueryURLRewrite extends HttpServlet {
             return contentType != null ? contentType : super.getContentType();
         }
 
+        /**
+         * True when {@code name} is a Content-Length header being set while this response is
+         * buffering ({@code cache=true}). A static resource forward step (e.g. Jetty's
+         * default/ResourceServlet serving a plain file) may set Content-Length through any of
+         * setHeader/addHeader/setIntHeader/addIntHeader -- HttpServletResponseWrapper's defaults for
+         * all four delegate straight to the real underlying response with no cache awareness, so
+         * each call site needs this same guard. Without it, Content-Length leaks onto the real
+         * response before a later <exist:view> step's -- possibly longer -- output is flushed to it.
+         * See https://github.com/eXist-db/exist/issues/6669
+         */
+        private boolean isBufferedContentLength(final String name) {
+            return cache && "Content-Length".equalsIgnoreCase(name);
+        }
+
         @Override
         public void setHeader(final String name, final String value) {
             if ("Content-Type".equals(name)) {
                 setContentType(value);
-            } else if (cache && "Content-Length".equalsIgnoreCase(name)) {
-                // See addHeader(): must not leak onto the real response while buffering.
+            } else if (isBufferedContentLength(name)) {
                 return;
             } else {
                 super.setHeader(name, value);
@@ -1425,16 +1444,26 @@ public class XQueryURLRewrite extends HttpServlet {
 
         @Override
         public void addHeader(final String name, final String value) {
-            if (cache && "Content-Length".equalsIgnoreCase(name)) {
-                // A static resource forward step (e.g. Jetty's default/ResourceServlet serving a
-                // plain file) sets Content-Length via addHeader() rather than setContentLength(int)/
-                // setContentLengthLong(long). Without this guard it leaks straight through to the
-                // real response (HttpServletResponseWrapper#addHeader() has no cache awareness),
-                // fixing Content-Length before a later <exist:view> step's -- possibly longer --
-                // output is flushed to it. See https://github.com/eXist-db/exist/issues/6669
+            if (isBufferedContentLength(name)) {
                 return;
             }
             super.addHeader(name, value);
+        }
+
+        @Override
+        public void setIntHeader(final String name, final int value) {
+            if (isBufferedContentLength(name)) {
+                return;
+            }
+            super.setIntHeader(name, value);
+        }
+
+        @Override
+        public void addIntHeader(final String name, final int value) {
+            if (isBufferedContentLength(name)) {
+                return;
+            }
+            super.addIntHeader(name, value);
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


duncdrum:
Closing and reopening as a same-repo PR (this one was opened from a fork before we confirmed upstream write access) so this can use GitHub's native stacked-PR support along with #6692/#6693. No code changes -- same branch, same commits.

PR #6692, opened 2026-09-10
[bugfix] Replace CachingResponseWrapper's header-name allowlist with buffer+replay

### LINKED ISSUE #3913: [BUG] last-modified and created header sent twice
**Describe the bug**

Setting Last-Modified and Created headers in the controller.xql for a templated view results in duplicated header values.

**Expected behavior**

Ensure that Last-Modified and Created headers are only sent once per request.

**To Reproduce**

Create a test application in existdb with 3 files

- controller.xql

```xquery
xquery version "3.1";

declare variable $exist:controller external;

<dispatch xmlns="http://exist.sourceforge.net/NS/exist">
    <forward url="{$exist:controller}/test.html">
        <set-header name="Last-Modified" value="Thu, 01 Jan 1970 00:00:00 GMT" />
        <set-header name="Date" value="Thu, 01 Jan 1970 00:00:00 GMT" />
        <set-header name="Created" value="Thu, 01 Jan 1970 00:00:00 GMT" />
    </forward>
    <view>
        <forward url="{$exist:controller}/test.xq" />
    </view>
</dispatch>
```

- test.xq

```xquery
xquery version "3.1";

declare option output:method "html5";
declare option output:media-type "text/html";

<html>
    <body>
        There should only be one Last-Modified and Created header
    </body>
</html>
```

- test.html

```html
<html/>
```

requests to this application will return duplicate Last-Modified and Created headers.

```bash
➜ curl http://localhost:8080/exist/apps/test/ -I
HTTP/1.1 200 OK
Date: Thu, 01 Jan 1970 00:00:00 GMT
Last-Modified: Thu, 01 Jan 1970 00:00:00 GMT
Created: Thu, 01 Jan 1970 00:00:00 GMT
Last-Modified: Tue, 01 Jun 2021 21:08:21 GMT
Created: Tue, 01 Jun 2021 21:08:21 GMT
X-XQuery-Cached: true
Content-Type: text/html;charset=utf-8
Content-Length: 4096
Server: Jetty(9.4.40.v20210413)
```

 - OS: macOS 10.15.7
 - Java Version 1.8.0_252-b09

**Additional context**

duncdrum:
## Summary

Stacked on #6691. Supersedes the fixed header-name allowlist that PR added (Content-Length, Content-Type, Last-Modified, ETag, Accept-Ranges) with a structural fix, per design discussion on #6691.

`CachingResponseWrapper` was conflating two different kinds of header write:

- **Pipeline configuration** — a controller.xql `<exist:set-header>` directive, applied by `URLRewrite#setHeaders` before `doRewrite()`'s `dispatcher.forward()` even runs. Must survive regardless of which pipeline step's output eventually wins.
- **Step output** — anything the dispatched step itself sets (Jetty's own resource metadata, or the step's own XQuery `response:set-header()` calls). Provisional: if a later `<exist:view>` step replaces this step's output, its headers must not survive either.

The allowlist from #6691 could only close this gap one enumerated header name at a time. This PR replaces it with a structural distinction instead:

- New `HeaderPassthroughResponse` interface + `HttpResponseWrapper#setPassthroughHeader`, used only by `URLRewrite#setHeaders`, tunneling through however many `CachingResponseWrapper` layers are stacked to reach the real response immediately. Deliberately distinct from `HttpResponseWrapper#setHeader` (which backs `response:set-header()`), so a step's own programmatic header call is not accidentally granted the same bypass.
- Everything else a step writes is now buffered via a sealed `BufferedHeader` record hierarchy (`setHeader`/`addHeader`/`setIntHeader`/`addIntHeader`/`setDateHeader`/`addDateHeader`) and replayed verbatim, in original order, from `flush()` — but only for whichever wrapper actually gets flushed.

Content-Type and Content-Length keep their existing special-casing (untouched): Content-Type via the first-write-wins `contentType` field, Content-Length derived from the actual buffered byte count at flush time rather than trusted from any step's self-reported value.

close #3913

## Test Plan

- [x] `URLRewriteResponseSetHeaderViewPipelineTest` — red against #6691's allowlist (an intermediate step's programmatic header leaked through), green after this change.
- [x] `URLRewriteFinalStepResponseSetHeaderSurvivesTest` — characterization test pinning the case that broke a first attempt at this redesign (a config-only allowlist): a final step's own `response:set-header()` call, with no static config representation, must still reach the client. Green throughout.
- [x] `URLRewriteSetHeaderSurvivesViewPipelineTest` (from #6691) — config directive survival, green throughout.
- [x] Both original #6667/#6669 regression tests still pass.
- [x] Full `org.exist.http.urlrewrite` suite (25 tests) green.
- [x] Broader `org.exist.http.*` suite (81 tests) and `org.exist.http.servlets.*` suite green — no regressions outside the urlrewrite package.


reinhapa (approved):

joewiz:

Approved. Replacing the allowlist with the config-versus-step-output distinction is the right call — an enumerated list of header names could only ever be closed one name at a time, and the next surface to set an unlisted header would have reopened it. Separating the two kinds of write structurally means the question stops being "which headers are special" and becomes "which step's output won", which is the question that actually determines whether a header should survive. Keeping `setPassthroughHeader` deliberately distinct from `HttpResponseWrapper#setHeader` matters too: it would have been easy to grant `response:set-header()` the same bypass by accident and quietly reintroduce the leak.

**The XQTS comparison above is a false alarm — please don't read the −3 as a regression.** `Constr-inscope-1`, `-3` and `-4` are flaky in the harness, not affected by this change. The same four tests flip in both directions on PRs that cannot possibly affect conformance:

| PR | what it changes | `Constr-inscope` delta |
|---|---|---|
| #6687 | OASIS catalog wiring | −1, +2 |
| #6666 | documentation only | −1 |
| #6661 | dependency pinning | −2 |
| #6433 | install-and-deploy version handling | −2, +3 |

A documentation-only PR cannot change XQuery conformance results. I'll open a separate issue for the flakiness with this evidence, since the real cost is the reverse case — a genuine conformance regression getting waved through as "probably the flaky ones again".

joewiz (approved):

PR #6693, opened 2026-09-10
[test] Extract nested classes and consolidate test suite

duncdrum:
## Summary

Stacked on #6692. Two things, both about the same underlying class that's been rewritten three times over this stack:

**1. Extract `XQueryURLRewrite`'s nested classes to top-level files.** `CachingResponseWrapper`, `RequestWrapper` (renamed `ControllerRequestWrapper` -- collided with the unrelated `org.exist.http.servlets.RequestWrapper`), `CachingServletOutputStream`, `CachingServletInputStream` move out of `XQueryURLRewrite` (1716 lines before this) into their own files. Purely mechanical, no behavior change.

**2. Actually use the resulting testability -- then go further.**
- Added `CachingResponseWrapperTest` and `ControllerRequestWrapperTest`: fast, direct unit tests for the getWriter()/getOutputStream() state machine, header buffer+replay, Content-Length derivation, passthrough tunnelling, and the request-wrapper behaviors Prethink had flagged as uncovered. Every one of these used to be verifiable only through a full HTTP round trip specifically because these were private nested classes with no way to construct or call them directly.
- Consolidated the 9 HTTP test classes that all used the identical `ExistWebServer` config (each paying its own ~10-15s startup cost) into one class (`UrlRewritePipelineHttpTest`), organized by original scenario. All 14 original test methods preserved verbatim; only colliding private helper methods renamed.

## Test Plan

- [x] Full `org.exist.http.urlrewrite` package: 70 tests, all green.
- [x] Broader `org.exist.http.*`/`org.exist.http.servlets.*` sweep: 154 tests, all green.
- [x] `UrlRewritePipelineHttpTest` alone: 14/14 tests (matching the 9 original classes' methods exactly), ~5-8s vs. the original ~140s combined.
- [x] `CachingResponseWrapperTest` + `ControllerRequestWrapperTest`: 44 unit tests, ~1s combined.


exist-core/src/main/java/org/exist/http/urlrewrite/CachingResponseWrapper.java
```diff
@@ -0,0 +1,350 @@
...
+     * back to {@code getWriter()} on {@link IllegalStateException} in order to close whichever
+     * stream is live, regardless of which one the forwarded servlet used.
+     */
+    private enum OutputMode { NONE, WRITER, STREAM }
+
+    private OutputMode outputMode = OutputMode.NONE;
+    private CachingServletOutputStream sos = null;
+    private PrintWriter writer = null;
+    private int status = HttpServletResponse.SC_OK;
+    private String contentType = null;
```
reinhapa: No need to set the `null` value, as it is the default anyway...

```suggestion
    private CachingServletOutputStream sos;
    private PrintWriter writer;
    private int status = HttpServletResponse.SC_OK;
    private String contentType;
```

exist-core/src/main/java/org/exist/http/urlrewrite/ControllerRequestWrapper.java
```diff
@@ -0,0 +1,305 @@
...
+    private ServletInputStream sis = null;
+    private BufferedReader reader = null;
+
+    private String contentType;
+    private int contentLength = 0;
+    private String characterEncoding = null;
+    private String method = null;
+    private String inContextPath = null;
+    private String servletPath;
+    private String basePath = null;
```
reinhapa: No need to set the `null` / zero values, as it is the default anyway...

```suggestion
    private ServletInputStream sis;
    private BufferedReader reader;

    private String contentType;
    private int contentLength;
    private String characterEncoding;
    private String method;
    private String inContextPath;
    private String servletPath;
    private String basePath;
```

reinhapa (changes requested):
Only some minor stuff

reinhapa (approved):

line-o (approved):
Good refactorings. :+1: for making tests faster

reinhapa (approved):

PR #6697, opened 2026-09-11
[bugfix] Fix non-ASCII cookie values silently dropped / crashing on Jetty 12

duncdrum:
## Summary

Independent of and unrelated to #6692/#6693 (the URL-rewrite pipeline stack) -- this investigates and fixes a separate, pre-existing bug found while reviewing a 2010-era `// TODO: remove this hack after fixing HTTP 1.1` comment during that other work.

`HttpResponseWrapper#encode` (added 2010-10-14, commit `67612ad4be`) applies the same ISO-8859-1 byte-packing scheme to both header values (`response:set-header()`) and cookie values (`response:set-cookie()`), to work around the Java Servlet API's byte-per-char header serialization for non-ASCII text.

Measured empirically against a real Jetty 12 instance (not just inferred from the spec):

- **Headers**: the hack still works. A non-ASCII header value's exact UTF-8 bytes reach the wire unchanged.
- **Cookies**: the hack does not work, and never really did on this stack. RFC 6265's `cookie-octet` grammar excludes any byte >= 0x80 -- stricter than headers' more permissive treatment -- and Jetty 12 enforces this strictly. Depending on request shape, a non-ASCII cookie value either:
  - vanishes from the response entirely, with no warning or error anywhere, or
  - fails the *entire request* with a 500: `org.eclipse.jetty.http.ComplianceViolationException: ... RFC6265 Cookie values characters restricted to US-ASCII: 0xd0 (forbidden)`.

## What Changed

**Commits 1-2 (measure, then red):** a unit test pinning the exact byte-packing transformation, a raw-socket wire test against real Jetty 12 confirming the header/cookie divergence above, and an end-to-end round-trip test (`response:set-cookie()` → `request:get-cookie-value()`) reproducing the 500.

**Commit 3 (fix):** `exist-core/src/main/java/org/exist/http/servlets/HttpResponseWrapper.java` -- cookie values now go through a separate `encodeCookieValue()` that percent-encodes instead of byte-packing, keeping the wire value strictly within `cookie-octet`'s ASCII-safe range regardless of input text. The header path (`encode()`) is untouched. `org.exist.xquery.functions.request.GetCookieValue#decode` (the read side) is updated symmetrically to percent-decode, with a fallback to the raw value if decoding fails, so a cookie eXist didn't set itself (browser, another app on the same domain) doesn't crash a lookup over a coincidental `%` that isn't a valid escape.

Both historic "TODO: remove this hack after fixing HTTP 1.1" comments are replaced with what direct measurement now shows -- the header-side hack isn't a fixable HTTP/1.1 defect (it's a still-current characteristic of the Java Servlet header API) and remains genuinely necessary today, not dead code.

## Test Plan

- [x] `HttpResponseWrapperEncodingTest` -- unit-level, pins the header byte-packing transformation (unchanged) and the new cookie percent-encoding.
- [x] `HttpResponseWrapperEncodingWireTest` -- real HTTP round trip via raw socket against live Jetty 12; confirms header behavior and documents the cookie divergence.
- [x] `NonAsciiCookieRoundTripTest` -- red against unfixed code (500), green after the fix; full `response:set-cookie()` → `request:get-cookie-value()` round trip.
- [x] Full `org.exist.http.*`/`org.exist.http.servlets.*`/`org.exist.xquery.functions.request.*`/`org.exist.xquery.functions.response.*` sweep: 132 tests, all green -- this is also the first test coverage cookies have had in this codebase at all.

### DIFF
--- a/exist-core/src/main/java/org/exist/http/servlets/HttpResponseWrapper.java
+++ b/exist-core/src/main/java/org/exist/http/servlets/HttpResponseWrapper.java
@@ -28,9 +28,11 @@ import jakarta.servlet.http.Cookie;
 import jakarta.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.net.URLEncoder;
 import java.util.Locale;
 
 import static java.nio.charset.StandardCharsets.ISO_8859_1;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * @author <a href="mailto:wolfgang@exist-db.org">Wolfgang Meier</a>
@@ -55,19 +57,19 @@ public class HttpResponseWrapper implements ResponseWrapper {
 	
 	@Override
 	public void addCookie(final String name, final String value) {
-		response.addCookie(new Cookie(name, encode(value)));
+		response.addCookie(new Cookie(name, encodeCookieValue(value)));
 	}
 
 	@Override
 	public void addCookie(final String name, final String value, final int maxAge) {
-		final Cookie cookie = new Cookie(name, encode(value));
+		final Cookie cookie = new Cookie(name, encodeCookieValue(value));
 		cookie.setMaxAge(maxAge);
 		response.addCookie(cookie);
 	}
 	
 	@Override
 	public void addCookie(final String name, final String value, final int maxAge, final boolean secure) {
-		final Cookie cookie = new Cookie(name, encode(value));
+		final Cookie cookie = new Cookie(name, encodeCookieValue(value));
 		cookie.setMaxAge(maxAge);
 		cookie.setSecure(secure);
 		response.addCookie(cookie);
@@ -75,7 +77,7 @@ public class HttpResponseWrapper implements ResponseWrapper {
 	
 	@Override
 	public void addCookie(final String name, final String value, final int maxAge, boolean secure, final String domain, final String path) {
-		final Cookie cookie = new Cookie(name, encode(value));
+		final Cookie cookie = new Cookie(name, encodeCookieValue(value));
 		cookie.setMaxAge(maxAge);
 		cookie.setSecure( secure );
 		if (domain != null && !domain.isEmpty()) {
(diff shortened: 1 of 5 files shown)


joewiz:

The header/cookie split is right, and measuring the wire rather than reading the spec is what makes it convincing — RFC 6265's `cookie-octet` genuinely does exclude bytes >= 0x80 where header values tolerate them. This is also the first test coverage cookies have ever had here, which is worth something on its own.

One substantive finding, and a small suggestion that I think resolves it.

## Switching the read side changes how *existing* cookies decode

`URLDecoder.decode` applies to every cookie already sitting in a browser, not only to ones written by the new code. Measured:

| cookie value already in a browser | decodes as today | decodes as after this PR |
| --- | --- | --- |
| `aGVsbG8+d29ybGQ=` (base64) | `aGVsbG8+d29ybGQ=` | `aGVsbG8 d29ybGQ=` |
| `abc+def` | `abc+def` | `abc def` |
| `tok%C3%A9n` | `tok%C3%A9n` | `tokén` |
| `50%` | `50%` | `50%` (unchanged — invalid escape, caught by the fallback) |

Base64 is the case I would worry about: its alphabet includes `+`, session tokens are routinely base64, and the corruption is silent. The `IllegalArgumentException` fallback only catches *malformed* escapes, so a well-formed `%C3%A9` in a foreign cookie gets decoded whether or not it was ever encoded.

The class javadoc does acknowledge the `+` ambiguity, but scopes it to "a cookie eXist didn't set itself". Two cases sit outside that framing: cookies **this instance set before the upgrade** are equally affected, and the `%XX` case is not mentioned at all.

## Suggestion: encode space as `%20` and leave `+` alone on the way back

`URLEncoder`/`URLDecoder` implement `application/x-www-form-urlencoded`, where space is `+`. Percent-encoding proper (RFC 3986) uses `%20` and gives `+` no special meaning. Moving to that shape removes the `+` hazard entirely:

| pre-existing cookie | this PR | with `%20` + literal `+` |
| --- | --- | --- |
| `aGVsbG8+d29ybGQ=` | `aGVsbG8 d29ybGQ=` | `aGVsbG8+d29ybGQ=` |
| `abc+def` | `abc def` | `abc+def` |

and every round trip this PR already tests still passes — Cyrillic, `a b`, `a+b`, `50%`, base64. Mechanically it is small: `URLEncoder.encode(v, UTF_8).replace("+", "%20")` on write, and on read protect a literal `+` before decoding (`v.replace("+", "%2B")`) or use a plain RFC 3986 decoder.

That leaves only the `%XX` ambiguity, which is inherent to decoding a namespace you do not own, and much rarer in practice than `+`.

Worth knowing: an RFC 3986 `decodeForURI` — a decoder that treats `+` as a literal — is already proposed in #6451 for the `xmldb:` URI functions, for exactly the same reason. If both land it would be good for them to share one implementation rather than grow two.

## Minor

`encode()` still calls `value.getBytes()` with no charset, so it takes the platform default. That is UTF-8 on Java 18+ so it is correct in practice today, but the new javadoc above it now states "a value's UTF-8 bytes" explicitly — worth making the code say so too while you are in there: `value.getBytes(UTF_8)`.

## Test suggestion

Whatever you decide on the above, a cookie value containing a literal `+` (a base64 payload is the realistic case) and one containing a `%XX` sequence would both be worth pinning — neither is covered now, and they are the shapes that change behavior for existing deployments.

duncdrum:
Good catch — pushed a fix in 48cd9c0654. Cookie encoding now rewrites the URLEncoder-produced `+` (space) to `%20` on write, and escapes a literal `+` to `%2B` before decoding on read, so a pre-existing `+` (e.g. in a base64 token) is never mistaken for an encoded space. Also made the header path's UTF-8 byte extraction explicit, and pointed to #6451 in the docs for sharing an implementation later. Added tests pinning the `+` and `%XX` cases on both the encode and decode sides.

Issue #6700, opened 2026-09-13
[design] Where should the sibling/preceding/following duplicate-suppression state live? (two candidate fixes for #6690)


joewiz:
[This issue was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

## Why this issue exists

#6690 is a silent wrong-result bug: an axis step with a name test returns the correct result once and an empty sequence on every evaluation after that, within a single query, whenever the context node carries a predicate context item. Root cause and reproduction are in that issue.

There are two reasonable places to fix it, and they embody different judgments about a piece of the engine that predates all of us. I have implemented both, and both are green — so the choice is not "which one works" but "which one do we want to live with". That is a call for people who know this code's history, which is why this is an issue rather than one PR with a preferred option.

**The two candidates** — A, resetting at the call site, and B, moving the state off the shared proxies — are both open as pull requests, linked in a comment below.

## The mechanism, in one paragraph

`LocationStep` caches its structural-index lookup in `currentSet` for the whole query execution (`resetState()` clears it only between executions). `NewArrayNodeSet.selectPrecedingSiblings`, `selectFollowingSiblings`, `selectPreceding` and `selectFollowing` then stamp those cached `NodeProxy` objects in place with the matching reference node's context, and a duplicate-suppression guard added in 2019 reads those stamps back to decide whether it has already handled a node. The guard is correct about the current evaluation and blind to the fact that the stamps it is reading may be left over from the previous one.

The guard still does necessary work: removing it fixes #6690 but breaks ten existing tests across `axes.xql`, `npt.xqm` and `positional-nested.xql`. So "delete the guard" is not on the table.

## Candidate A — reset the cached set when it is reused

`LocationStep` clears predicate contexts on `currentSet` when it reuses a cached set rather than building a fresh one, so the guard only ever sees stamps from the current evaluation.

**For:** Smallest possible change, confined to two call sites in one method each. Touches nothing in the node-set algorithms or in any hot path. The invariant is easy to state: a cached set enters each evaluation looking the way a freshly built one would.

[...]

Issue #6690, opened 2026-09-09
[BUG] Axis + name test returns wrong result from the 2nd evaluation onwards, when the node was reached through a range-indexed lookup


mercibe:
## Summary

`preceding-sibling::NAME` (and `preceding::`, `following::`, `following-sibling::`)
returns the correct result on its **first** evaluation and an **empty sequence on every
subsequent evaluation**, within the same query, when the context node belongs to a
document that was reached through a range-indexed lookup.

The same expression is correct when the document is reached via `doc()` or
`collection()`, correct on in-memory nodes, and correct in Saxon.

This is a silent wrong-result bug: no error is raised. Code that evaluates such an
expression once per item in a loop gets correct data for the first item and empty data
for all the others.

## Version

eXist-db 6.4.1 (build 2026-03-05), Java 17, Linux.

## Reproducing

Self-contained. Run against `/exist/rest/db`:

```xquery
xquery version "3.1";
declare namespace t="urn:example";

(: test collection with a range index on t:key :)
let $c   := xmldb:create-collection("/db", "bug-repro")
let $cfg := xmldb:create-collection("/db/system/config/db", "bug-repro")
let $x   := xmldb:store("/db/system/config/db/bug-repro", "collection.xconf",
  <collection xmlns="http://exist-db.org/collection-config/1.0" xmlns:t="urn:example">
    <index><range><create qname="t:key" type="xs:string"/></range></index>
  </collection>)
let $uri := xmldb:store("/db/bug-repro", "t.xml",
  <t:root><t:key>K1</t:key><t:wrap><t:a/><t:target/></t:wrap></t:root>)
let $ri  := xmldb:reindex("/db/bug-repro")

(: same node, reached two different ways :)
let $viaIndex := (//t:key[. = "K1"]/ancestor::t:root//t:target)[1]
let $viaDoc   := (doc($uri)/t:root//t:target)[1]

return element results {
  element viaRangeIndexedLookup {
    string-join(for $i in 1 to 4 return string(count($viaIndex/preceding-sibling::t:a)), ",") },
  element viaDoc {
    string-join(for $i in 1 to 4 return string(count($viaDoc/preceding-sibling::t:a)), ",") },
  element wildcardControl {
    string-join(for $i in 1 to 4 return string(count($viaIndex/preceding-sibling::*)), ",") },
  element expected { "1,1,1,1" }
}
```

## Actual result

```xml
<results>
[...]

PR #6701, opened 2026-09-13
[bugfix] Candidate A for #6690 — reset predicate contexts when a cached index result is reused

joewiz:

**This is one of two candidate fixes for #6690. See #6700 for the design question, the trade-offs, and a link to the alternative candidate. Please do not merge this without reading that issue — the two PRs are deliberately competing, not stacked.**

Addresses #6690

## Summary

An axis step with a name test returned the correct result on its first evaluation and an empty sequence on every evaluation after that, within a single query, whenever the context node carried a predicate context item. Silent — no error, no warning, data loss proportional to loop length.

## What Changed

**`exist-core/src/main/java/org/exist/xquery/LocationStep.java`**

`LocationStep` caches its structural-index lookup in `currentSet` for the whole query execution, and `NewArrayNodeSet`'s sibling and preceding/following select methods stamp those cached `NodeProxy` objects in place with the matching reference node's context. A duplicate-suppression guard then reads those stamps back — and on the second evaluation it is reading stamps left by the first, so it suppresses every node.

This change clears predicate contexts on `currentSet` when the cached set is *reused* rather than freshly built, so the guard only ever sees stamps from the current evaluation. A freshly built set carries no contexts, so only the reuse path needs it.

Applied at the two call sites that feed the guarded select methods: `getSiblings` and `getPrecedingOrFollowing`.

## Why not simply remove the guard

The guard is still needed. With it disabled the new regression test passes, but **ten existing tests fail** across `axes.xql`, `npt.xqm` and `positional-nested.xql` — including the file the guard's original 2019 commit added its tests to. It suppresses genuine duplicates between different reference nodes *within* one evaluation. The bug is the stale state it reads, not the guard.

## Known limitation

The previous evaluation's result set holds the same `NodeProxy` instances as the cached set, so clearing at the start of evaluation *n* retroactively mutates the result of evaluation *n-1*. I could not construct a case where that is observable — it is harmless under `count()` and in every test we have — but it is a real property of this approach and the main reason candidate B exists. This is discussed in #6700.

## Test Plan

- [x] New `SiblingAxisRepeatedEvaluationRegressionTest` fails on unfixed `develop`, passes with the fix
- [x] Evaluates each affected axis four times over the same node; pins the wildcard form, a node reached without a predicate, an unaffected axis, and a non-matching name alongside
- [x] `xquery.CoreTests` — the whole `src/test/xquery` corpus, including the ten tests naive guard removal breaks — green
- [x] Full `mvn test` on `exist-core` green
- [x] Codacy PMD clean on the changed file
- [x] `license:check` clean

## Scope

Two call sites in one class. No change to the node-set algorithms or to any hot path.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/LocationStep.java
+++ b/exist-core/src/main/java/org/exist/xquery/LocationStep.java
@@ -584,6 +584,23 @@ public class LocationStep extends Step {
         }
     }
 
+    /**
+     * Clears predicate contexts left on the cached structural-index result by a previous
+     * evaluation of this step.
+     *
+     * <p>{@link #currentSet} is held for the whole query execution, and the sibling and
+     * preceding/following select methods stamp its {@code NodeProxy} objects in place. Their
+     * duplicate-suppression guard reads those stamps, so without this reset it sees state left
+     * over from the previous evaluation and skips every node. Only reuse of a cached set needs
+     * this; a freshly built one carries no contexts.</p>
+     */
+    private void clearCachedSetPredicateContexts() {
+        for (final NodeProxy p : currentSet) {
+            p.clearContext(Expression.IGNORE_CONTEXT);
+        }
+    }
+
+
     protected Sequence getAttributes(final XQueryContext context, final Sequence contextSequence)
             throws XPathException {
         if (!contextSequence.isPersistentSet()) {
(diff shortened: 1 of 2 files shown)


joewiz:

Converting this to a draft. @duncdrum has said on #6700 that candidate B (#6702) looks like the better direction, and I agree — so unless the discussion there turns, **this PR is likely to be closed rather than merged**.

Leaving it open as a draft for now for two reasons: #6700 is still the live decision and it is easier to compare two things that both exist, and the trade-off recorded in this PR's description and commit message is the clearest statement of what candidate B is buying us. If B lands, I will close this.

Note it now conflicts with `develop` — #6703 merged and both touch `LocationStep`. I have deliberately not rebased, since that work is wasted if this is closed. If the decision goes the other way I will rebase and un-draft.

PR #6702, opened 2026-09-13
[bugfix] Candidate B for #6690 — keep axis duplicate-suppression state per call

joewiz:

**This is one of two candidate fixes for #6690. See #6700 for the design question, the trade-offs, and a link to the alternative candidate. Please do not merge this without reading that issue — the two PRs are deliberately competing, not stacked.**

Addresses #6690

## Summary

An axis step with a name test returned the correct result on its first evaluation and an empty sequence on every evaluation after that, within a single query, whenever the context node carried a predicate context item. Silent — no error, no warning, data loss proportional to loop length.

## What Changed

**`exist-core/src/main/java/org/exist/dom/persistent/NewArrayNodeSet.java`**

The duplicate-suppression guard in `selectPrecedingSiblings`, `selectFollowingSiblings`, `selectPreceding` and `selectFollowing` decided whether it had already handled a node by reading a context stamp back off the `NodeProxy`. Those proxies are shared with `LocationStep`'s cached structural-index result, which is held for a whole query execution — so on the second evaluation the guard was reading stamps left by the first, and suppressed every node.

This change stops the guard reading the proxy. Each call now records which nodes it has stamped and with which context id, in a map local to that call, and the guard consults that instead. Within one call the behavior is identical to before; across calls there is nothing to leak, whatever the caller does with the cache.

Two small private helpers, `alreadyStampedInThisCall` and `recordStamp`, keep the four methods readable and the transformation obviously faithful — the recorded value is exactly what the guard used to read off the proxy.

Note the preceding/following guards carry an extra `contextId != NO_CONTEXT_ID` condition that the sibling ones do not; that asymmetry is preserved at the call sites rather than folded into the helper.

## Why not simply remove the guard

Removing it is not an option. With it disabled the new regression test passes, but **ten existing tests fail** across `axes.xql`, `npt.xqm` and `positional-nested.xql` — including the file the guard's original 2019 commit added its tests to. It suppresses genuine duplicates between different reference nodes *within* one evaluation. The bug is where that state was stored, not the guard.

## Cost

A small map allocated per call when a context id is in play. It is bounded by the number of nodes actually stamped, not by the size of the cached set, so it scales with the result rather than the index — but it is an allocation in a hot path where there was none, and that is a fair thing to weigh against candidate A's smaller footprint. Discussed in #6700.

## Test Plan

- [x] New `SiblingAxisRepeatedEvaluationRegressionTest` fails on unfixed `develop`, passes with the fix
- [x] Evaluates each affected axis four times over the same node; pins the wildcard form, a node reached without a predicate, an unaffected axis, and a non-matching name alongside
- [x] `xquery.CoreTests` — the whole `src/test/xquery` corpus, including the ten tests naive guard removal breaks — green
- [x] Full `mvn test` on `exist-core` green
- [x] Codacy PMD clean on the changed file
- [x] `license:check` clean

## Scope

One class, four methods, no behavior change within a single evaluation.

### DIFF
--- a/exist-core/src/main/java/org/exist/dom/persistent/NewArrayNodeSet.java
+++ b/exist-core/src/main/java/org/exist/dom/persistent/NewArrayNodeSet.java
@@ -594,9 +594,49 @@ public class NewArrayNodeSet extends AbstractArrayNodeSet implements ExtNodeSet,
         }
     }
 
+    /**
+     * Whether this call has already stamped {@code nodes[idx]} with a context bearing the same
+     * context id as {@code reference}'s.
+     *
+     * <p>This replaces reading the stamp back off the {@code NodeProxy}. The proxies in this set
+     * are shared with {@code LocationStep}'s cached structural-index result, which is held for a
+     * whole query execution, so a stamp read from the proxy could have been left by a *previous*
+     * evaluation of the same step — in which case every node was suppressed and the step returned
+     * empty. Keeping the duplicate-suppression state per call confines it to the evaluation that
+     * created it. See issue #6690.</p>
+     *
+     * @param stamped the per-call record of stamps applied so far
+     * @param idx the index into {@link #nodes} of the candidate node
+     * @param reference the context node being matched against
+     * @param contextId the context id of the current evaluation
+     * @return true if the candidate should be skipped as a duplicate
+     */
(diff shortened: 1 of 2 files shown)


duncdrum (commented):
I prefer this approach compared to option A

PR #6703, opened 2026-09-13
[bugfix] Look up a named attribute kind test in the attribute index on the self axis

### LINKED ISSUE #6689: [BUG] `self::attribute(NAME)` is always false on attributes of persistent nodes
## Summary

A predicate `self::attribute(NAME)` never matches when applied to an attribute
belonging to a **stored (persistent)** node. The same expression is correct on an
in-memory node, and correct in Saxon.

This is independent of the attribute's namespace and of any index configuration:
an attribute in no namespace fails exactly like `xml:id`.

## Version

eXist-db 6.4.1 (build 2026-03-05), Java 17, Linux.

## Reproducing

Self-contained, no configuration needed. Run against `/exist/rest/db`:

```xquery
xquery version "3.1";
declare namespace t="urn:example";

let $doc  := <t:root><t:target id="1"/></t:root>
let $col  := xmldb:create-collection("/db", "bug-repro")
let $uri  := xmldb:store("/db/bug-repro", "t.xml", $doc)

let $persistent := doc($uri)/t:root/t:target
let $inMemory   := $doc/t:target

return element results {
  element persistent { count($persistent/@*[self::attribute(id)]) },
  element inMemory   { count($inMemory/@*[self::attribute(id)]) },
  element expected   { 1 },
  (: controls: these are all correct on the same persistent node :)
  element controlDirectAxis { count($persistent/@id) },
  element controlInstanceOf { count($persistent/@*[. instance of attribute(id)]) }
}
```

## Actual result

```xml
<results>
   <persistent>0</persistent>          <!-- wrong -->
   <inMemory>1</inMemory>
   <expected>1</expected>
   <controlDirectAxis>1</controlDirectAxis>
   <controlInstanceOf>1</controlInstanceOf>
</results>
```

## Expected result

`persistent` should be `1`, like `inMemory`.

## Notes

* The controls show the attribute is present and reachable: `@id` finds it, and
  `. instance of attribute(id)` correctly returns true on the very same node.
  Only the `self::attribute(NAME)` kind test fails.
* Element name tests are unaffected: `self::NAME` on an element works correctly on
  persistent nodes.
* Not related to indexes: reproduced on a collection with no `collection.xconf`, and
  on a freshly rebuilt database (all data files deleted and re-imported).

## Workarounds

Any of these behave correctly on persistent nodes:

```xquery
$n/@*[. instance of attribute(id)]
$n/@*[name(.) = "id"]
$n/@*[node-name(.) eq xs:QName("id")]
$n/@id
```

joewiz:

Closes #6689

## Summary

A named attribute kind test on the self axis — `self::attribute(id)` — was always false on a persistent node. It was correct on in-memory nodes and correct in Saxon, so the same predicate gave different answers depending on whether the document had been stored.

## What Changed

**`exist-core/src/main/java/org/exist/xquery/LocationStep.java`**

`getSelf` looked every non-wildcard test up in the element half of the structural index:

```java
return index.findElementsByTagName(ElementValue.ELEMENT, docs, test.getName(), selector, this);
```

`self::attribute(id)` therefore searched for an *element* named `id`, found nothing, and returned the empty sequence — without ever consulting the node test. The fix selects the index half from the test's own type.

## Root-Cause Verification

Confirmed empirically rather than by inspection: instrumenting `NameTest.matches(NodeProxy)` showed **zero** calls in the failing case, which rules out the name comparison and places the failure upstream of it, in the index lookup.

Two other hypotheses were tested and falsified along the way, both worth recording so a future reader does not retrace them:

- A `Type.ITEM` versus `Type.NODE` mismatch in `NameTest.matches(NodeProxy)` — a probe widening that check changed nothing.
- The `analyze()`-time empty-sequence short-circuit for the self axis — the bound-variable shape `$a[self::attribute(id)]` bypasses the context-step analysis entirely and still failed.

The single hardcoded index half accounts for every observed case:

| shape | before | why |
| --- | --- | --- |
| `@*[self::attribute(id)]` persistent | fails | non-wildcard test, wrong index half |
| `@*[self::attribute()]`, `@*[self::attribute(*)]` | works | wildcard branch, never reaches that line |
| `$n/attribute::attribute(id)` | works | `getAttributes` selects the correct half |
| `@*[. instance of attribute(id)]` | works | not a location step |
| in-memory | works | memtree path |
| `self::NAME` on elements | works | element half is the correct one here |

This is not a regression: the line is byte-identical in the `eXist-6.4.1` tag and current `develop`.

## Test Plan

- [x] New `SelfAxisAttributeKindRegressionTest` fails on unfixed `develop` and passes with the fix
- [x] Covers the reported case plus a namespaced attribute, a non-matching name, a named context step, a bound variable, the unnamed and wildcard forms, the attribute axis, and the element-side tests that were already correct — each paired against the in-memory result, which was always correct and defines the expected answer
- [x] Full `mvn test` on `exist-core`
- [x] `xquery.CoreTests` — the whole `src/test/xquery` corpus — green
- [x] Codacy PMD clean on the changed Java file
- [x] `license:check` clean

## Scope

One expression, on one axis, in one method. No change to the attribute axis, to element tests, or to the wildcard branches.

### DIFF
--- a/exist-core/src/main/java/org/exist/xquery/LocationStep.java
+++ b/exist-core/src/main/java/org/exist/xquery/LocationStep.java
@@ -580,8 +580,22 @@ public class LocationStep extends Step {
                         "Using structural index '" + index.toString() + "'");
             }
             final NodeSelector selector = new SelfSelector(contextSet, contextId);
-            return index.findElementsByTagName(ElementValue.ELEMENT, docs, test.getName(), selector, this);
+            return index.findElementsByTagName(indexTypeFor(test), docs, test.getName(), selector, this);
         }
+    }
+
+    /**
+     * Selects the half of the structural index that a node test should be looked up in.
+     *
+     * <p>The index is split into an element half and an attribute half. A named attribute kind
+     * test such as {@code self::attribute(id)} must be looked up in the latter; searching the
+     * element half for an element of that name always comes back empty.</p>
+     *
+     * @param test the node test being applied
+     * @return {@link ElementValue#ATTRIBUTE} for an attribute test, {@link ElementValue#ELEMENT} otherwise
+     */
+    private static byte indexTypeFor(final NodeTest test) {
+        return test.getType() == Type.ATTRIBUTE ? ElementValue.ATTRIBUTE : ElementValue.ELEMENT;
     }
 
     protected Sequence getAttributes(final XQueryContext context, final Sequence contextSequence)
--- a/exist-core/src/test/java/org/exist/xquery/SelfAxisAttributeKindRegressionTest.java
+++ b/exist-core/src/test/java/org/exist/xquery/SelfAxisAttributeKindRegressionTest.java
@@ -1,0 +1,139 @@
+/*
+ * eXist-db Open Source Native XML Database
+ * Copyright (C) 2001 The eXist-db Authors
+ *
+ * info@exist-db.org
+ * http://www.exist-db.org
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery;
+
+import org.exist.test.ExistXmldbEmbeddedServer;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.xmldb.api.base.ResourceSet;
+import org.xmldb.api.base.XMLDBException;
+import org.xmldb.api.modules.XQueryService;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Regression test for issue #6689. A named attribute kind test on the self axis,
+ * such as {@code self::attribute(id)}, was always false on a persistent node.
+ *
+ * <p>{@code LocationStep.getSelf} looked every non-wildcard test up in the element half of the
+ * structural index, so {@code self::attribute(id)} searched for an <em>element</em> named
+ * {@code id}, found nothing, and returned the empty sequence without ever consulting the node
+ * test. The unnamed and wildcard forms take a different branch, and the attribute axis uses
+ * {@code getAttributes}, which selects the correct half — which is why only this one shape
+ * failed.</p>
+ *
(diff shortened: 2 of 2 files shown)


exist-core/src/main/java/org/exist/xquery/LocationStep.java
```diff
@@ -580,10 +580,24 @@ private Sequence getSelf(final XQueryContext context, final Sequence contextSequ
...
+     *
+     * <p>The index is split into an element half and an attribute half. A named attribute kind
+     * test such as {@code self::attribute(id)} must be looked up in the latter; searching the
+     * element half for an element of that name always comes back empty.</p>
+     *
+     * @param test the node test being applied
+     * @return {@link ElementValue#ATTRIBUTE} for an attribute test, {@link ElementValue#ELEMENT} otherwise
+     */
+    private static byte indexTypeFor(final NodeTest test) {
+        return test.getType() == Type.ATTRIBUTE ? ElementValue.ATTRIBUTE : ElementValue.ELEMENT;
```
dizzzz, line 606: should this limit to attributes/elements only? e.g. PIs or comments?

dizzzz (approved):

exist-core/src/main/java/org/exist/xquery/LocationStep.java
joewiz: [This response was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

Good catch — PIs are affected, comments are not. I tested rather than reasoned, and it turned up a second problem as well.

**Processing instructions: same defect.** A PI is addressable by name, but the structural index has only an element half and an attribute half, so there is nowhere to look one up. `self::processing-instruction('target')` therefore searched the *element* half for an element named `target` and never matched:

| | before | in-memory | expected |
| --- | ---: | ---: | ---: |
| `self::processing-instruction('target')` | 0 | 1 | 1 |
| `self::element(a)` | 1 | 1 | 1 |

**Comments and text nodes are safe** — they have no name, so `comment()` and `text()` are `TypeTest`s that take the wildcard branch and never reach the index lookup.

I have pushed a fix. Note it is *not* the obvious one: there is a commented-out `VirtualNodeSet` branch for PIs sitting directly above the line you flagged, and restoring it turns the wrong answer into a hard error (`Internal evaluation error: context is missing for node 1.1`) — which I assume is why it was commented out rather than deleted. Matching PI tests directly against the context set needs no index at all and works. That also fixes the unnamed form, which was returning every child node rather than the two PIs.

New `SelfAxisProcessingInstructionRegressionTest` covers the named form, a non-matching target, the unnamed form, step position, and the element cases as controls.

---

**Separately, and not fixed here:** the same probe showed that unnamed kind tests on the self axis are always *true* in predicate position on persistent nodes.

| predicate, on a node with 3 children (2 PIs, 1 element) | persistent | in-memory | expected |
| --- | ---: | ---: | ---: |
| `[self::comment()]` | 3 | 0 | 0 |
| `[self::text()]` | 3 | 0 | 0 |
| `[self::element()]` | 3 | 1 | 1 |

Nothing is filtered. In step position (`/self::comment()`) the same tests are correct, so it is specific to the predicate path: `getSelf`'s wildcard branch returns the context set unfiltered when a context id is in play, stamping contexts only on the nodes that matched.

That is a different branch of the same method with a much wider blast radius, so I have filed it separately as #6705 rather than widening an approved PR. It is also the symptom the reporter described in a follow-up on #6689, which I had wrongly told him did not reproduce — correcting that on the issue now.

duncdrum (approved):

PR #6718, opened 2026-09-15
[bugfix] Treat an '@' in xqdoc prose as text rather than a tag

### LINKED ISSUE #1386: inspect:inspect-module($module) break on '@', '%', etc
```xQuery
(:~
: this …
: @return taxonomy[@type = "reign"]
:)
```

will become
```xml
 <returns type="item()" cardinality="zero or more"> taxonomy[</returns>
```

instead of:
```xml
 <returns type="item()" cardinality="zero or more">taxonomy[@type = "reign"]</returns>
```

### Context information
- eXist-db version 3.1.1
- Java8u121
- MacOs 10.12.4
- 64 bit

joewiz:

Closes #1386

## Summary

`inspect:inspect-module` drops everything from the first `@` in an xqdoc comment to the end of that description. No error is raised — the truncated text is returned as if complete.

```xquery
(:~
 : Selects taxonomy[@type = "reign"] from the source.
 : THIS SENTENCE MUST SURVIVE.
 :)
declare function t:f() { 1 };
```

**Before:** `Selects taxonomy[`
**After:** `Selects taxonomy[@type = "reign"] from the source. THIS SENTENCE MUST SURVIVE.`

A bare `@` truncates too: `Costs 5 @ 3 dollars each. SECOND SENTENCE.` returns `Costs 5`.

This is #1386, open since March 2017. Its commenters identified the lexer at the time but no fix followed.

## Why it matters beyond one function

`inspect:inspect-module` is the ecosystem's single introspection entry point. It is read by `function-documentation`, `eXide`, `existdb-openapi` and `existdb-langserver`, so the truncation silently degrades published function documentation, editor hover text and autocomplete descriptions at once. Because it fails silently, it reads as an authoring mistake rather than a bug.

The downstream workaround is to keep `@` out of prose. That does not scale — every XQuery author in the ecosystem would have to know the rule.

## Cause

Three rules in `XQDocParser.g`:

```antlr
CHARS: (~('\n' | '@' | ':'))+;     // prose stops dead at '@'
TAG:   '@' ('A'..'Z' | 'a'..'z' | '0'..'9')+;
AT:    '@';                         // declared, never referenced by any parser rule
```

`CHARS` excludes `@`, so prose can never contain one. What happens next depends on the following character:

- `@` + alphanumerics (`@type`, `@guest`) lexes as `TAG`. The parser's `( TAG ) => taggedContents[doc]` then opens a **new xqdoc tag** mid-sentence, and the remainder of the paragraph is stored via `doc.setTag("@type", …)` instead of `doc.addDescription(…)` — filed under a tag nobody reads.
- `@` + anything else emits `AT`, which no parser rule accepts.

## What Changed

A real xqdoc tag only ever appears at the start of a comment line, immediately after the leading `" : "`. `contents` now tracks that position and treats `@` anywhere else as ordinary text:

- a guarded alternative `{ !lineStart }? TAG` appends a mid-prose tag token as literal text;
- a new `AT` alternative appends a bare `@`;
- `lineStart` is set by `TRIM` and cleared by any non-blank `CHARS`.

A line-start `TAG` matches neither alternative, so `contents` ends there and the outer `( TAG ) => taggedContents` predicate fires exactly as before.

### Why the parser and not the lexer

The obvious fix is to make `TAG` match only at line start. That needs either mutable lexer state or `@` folded into `CHARS`, and the latter creates a genuine `CHARS`/`TAG` ambiguity in a single-DFA ANTLR 2 lexer.

The parser already has the information — it can see that only whitespace has been consumed since the line's `:` — so no state has to be invented. It also leaves tokenization untouched, which keeps `@param`/`@return` recognition from regressing through a lexer path I did not anticipate. The entire behavioral change is one semantic predicate on one alternative.

Happy to move it into the lexer if a reviewer prefers that; I would want the ambiguity question settled first.

### ANTLR warnings

No new class of warning. Measured against an unmodified build of the same grammar:

| warning | before | after |
| --- | ---: | ---: |
| `nondeterminism … between alt N and exit branch of block` | 3 | 5 |
| `Syntactic predicate ignored for single alternative` (`SIMPLE_COLON`, untouched) | 1 | 1 |

The count rises only because the closure gained two alternatives; the warning is inherent to a greedy `(...)+` under `k = 1`.

## Test Plan

`InspectModuleTest.withAtSignInline` already existed, `@Ignore`d against this issue — it is now enabled. Added alongside it:

- an `@` mid-description, with a following sentence that must survive
- a bare `@`, an email address in prose, and a description whose last character is `@`
- a function whose `@param` and `@return` **values** contain `@` — the regression case, pinning that tags are still parsed as tags

- [x] 4 of the 7 `InspectModuleTest` cases fail on unfixed `develop`; all 7 pass with this change
- [x] `xquery.inspect.InspectTests` 3/3, `xquery.securitymanager.SecurityManagerTests` 7/7 (both exercise `inspect:`)
- [x] `xquery.CoreTests` — the whole `src/test/xquery` corpus — 966/966
- [x] Codacy PMD clean on the changed test; `license:check` clean

**Not run locally:** the full `exist-core` suite. It was killed twice by a memory watchdog on my machine, so I verified the blast radius instead — `XQDocHelper`, `InspectModule` and `InspectFunctionHelper` are the only consumers of this parser, and every test touching them is above. CI covers the full matrix.

## Out of scope

**The `%` half of #1386 is already fixed.** The existing `onAnnotatedFunction` test inspects a `%public %x:path(...)` function and passes unchanged on `develop`, so the issue's remaining scope really was just the `@` case.

**`xqdoc:scan` is a different parser** — the external `org.xqdoc:xqdoc-conversion` library in `extensions/xqdoc`, not this grammar. It is untouched here.


duncdrum (commented):
This is an improvement but shouldn't we check for `@` in combination with the reserved Xqdoc keywords at the start of the line? 

The at symbol could appear at the start of a line as part of normal documentation prose.

duncdrum (approved):
thank you @joewiz the ghosts of past and current duncdrums are happy campers now.

reinhapa (approved):

PR #6721, opened 2026-09-16
Improve security and configurability of functions that access the Environment and System Properties

duncdrum:
## Summary

Forward-port and rebase of #4452 (originally #4449) onto current `develop`, across roughly 4500 commits of drift, plus two bug fixes found while re-testing it during the rebase.

- Replaces the blanket "caller must be a DBA" check on `fn:available-environment-variables`, `fn:environment-variable`, `util:available-system-properties`, and `util:system-property` with per-name, per-group/per-user access rules configurable via `conf.xml` (see the new `AccessUtil` helper and its use in `FnModule`/`UtilModule`).
- Makes environment variables and system properties deterministic within a single query's execution scope by caching them once on `XQueryContext`.
- Adds `util:available-system-properties#0`.
- Renames `SystemProperties` to `ExistSystemProperties` and fixes a latent thread-safety issue in that singleton.

## Related

- #6563 (open, same underlying security concern) — proposes a DBA-only observability API for effective server configuration (`system:get-configuration()` and friends). Its Security section quotes @adamretter from #3331 verbatim: "There are some things in the `conf.xml` that should not be visible apart to anyone apart from `dba`" — the same concern this PR addresses for environment variables and system properties. Worth deciding whether #6563's new functions should reuse this PR's `AccessUtil` per-group/per-user model instead of a hardcoded blanket-DBA-only check, so the two don't end up with divergent access-control philosophies for adjacent "expose config-derived data" functions.
- #3331 (closed, superseded by #6563) — the original issue where that concern was raised.

## What Changed

- **Rebase** (`1b90ec7`, authored by @adamretter): the original PR content, replayed onto current `develop`. All conflicts were mechanical API drift from the `SystemProperties` → `ExistSystemProperties` rename colliding with unrelated refactors on `develop` since 2022 (e.g. `JettyStart` splitting `run()` into `bootstrapExistDb()`, `InteractiveClient#getNotice()` taking an injected `BinaryOperator<String>`, `AbstractInternalModule`'s 3-arg constructor being deprecated in favour of automatic function ordering per #6378).
- **Regex fix** (`bc01e0e`): the `conf.xml` parameter-name patterns in `FnModule`/`UtilModule` used `([^=\x00])+` instead of `([^=\x00]+)` — the quantifier repeated the single-character capturing group instead of sitting inside it, so `Matcher#group(1)` only ever returned the last character of the variable/property name. Every per-name access rule therefore silently failed to match and fell through to the `"otherwise"` default; only the `"*"` wildcard rule ever worked, which is why the two test fixtures bundled with #4452 (`conf-env-vars-admins-only.xml`, `conf-sys-props-admins-only.xml`) didn't catch it — they only exercise `"*"`.
- **Default-fallback fix** (`b7b9635`): `AccessUtil#parseAccessParameters`'s `"*"` (otherwise) → DBA-group default only applied when *no* group rule at all had been configured. As soon as an admin added one specific named group rule without an explicit `"*"`, every other, unlisted name became inaccessible to everyone including DBAs — contradicting the documented behaviour in `conf.xml` ("If `*` is not set in this config, then it is set to the `DBA` group by default."). Fails closed, not a security hole, but a docs/behaviour mismatch.
- Added `AccessUtilTest` covering both fixes with unit tests that fail against the pre-fix code.
- **Review/CI/Codacy follow-ups**: bumped `conf.xsd`'s schema version for the new `conf.xml` blocks (fixes the schema-governance check); fixed a malformed `@author` javadoc tag that was failing the javadoc build; took a defensive snapshot before iterating `System.getProperties()` in `XQueryContext` and `LauncherWrapper` to avoid a possible `ConcurrentModificationException`; deduplicated the access-rule lazy-init getters in `FnModule`/`UtilModule`; reduced `AccessUtil#parseAccessParameters`'s NPath complexity; replaced the two hand-copied `conf.xml` test fixtures with fixtures generated from canonical, matching this project's existing per-package fixture convention.

## Spec References

N/A — this is a security/configurability change to eXist-db-specific functions, not a W3C conformance change.

## Test Plan

- [x] `mvn install -pl exist-core -am -DskipTests` — builds clean on JDK 21
- [x] `FunEnvironmentTest`, `SystemPropertyTest` (bundled with #4452) — pass
- [x] `AccessUtilTest` (new) — pass, and confirmed to fail against both pre-fix versions of the code
- [x] Full `exist-core` test suite: `Tests run: 7334, Failures: 0, Errors: 0, Skipped: 101` — `BUILD SUCCESS`

Opened as draft since this still needs a maintainer/community decision on the access-control model and whether the shipped `conf.xml` defaults are the ones we want, per the original PR's outstanding discussion.


exist-core/src/main/java/org/exist/launcher/LauncherWrapper.java
```diff
@@ -179,7 +179,11 @@ protected void getJavaOpts(final List<String> args, final Properties launcherPro
         getLauncherOpts(args, launcherProperties);
 
         boolean foundExistHomeSysProp = false;
-        final Properties sysProps = System.getProperties();
+        // Properties#clone() is synchronized on the live, JVM-wide System.getProperties()
+        // instance, so this takes a safe, consistent snapshot even if another thread calls
+        // System.setProperty()/clearProperty() concurrently, rather than iterating the live
+        // Hashtable directly and risking a ConcurrentModificationException.
+        final Properties sysProps = (Properties) System.getProperties().clone();
         for (final Map.Entry<Object, Object> entry : sysProps.entrySet()) {
```
reinhapa, line 187: This would be an even simpler way to create a copy...

```suggestion
for (final Map.Entry<Object, Object> entry : Map.copyOf(System.getProperties()).entrySet()) {
```

reinhapa (changes requested):
Only some small suggestion..

exist-core/src/main/java/org/exist/launcher/LauncherWrapper.java
duncdrum: Thanks for the suggestion, but I don't think this is equivalent — it would reintroduce the race this commit fixes.

`Properties#clone()` (inherited from `Hashtable`) is a `synchronized` method that copies the whole table atomically while holding the instance's monitor — the same monitor `System.setProperty()`/`clearProperty()` use internally (they're also `synchronized`).

`Map.copyOf(...)` on a `Hashtable`/`Properties` source ends up calling `entrySet().toArray()`, which drives `Hashtable`'s iterator. That iterator is only fail-fast (`Enumerator.next()` checks `modCount != expectedModCount`) and does **not** hold the monitor while walking the table. So if another thread calls `setProperty`/`clearProperty` mid-iteration, `Map.copyOf(System.getProperties())` can still throw `ConcurrentModificationException` — the exact bug we're trying to avoid here.

I'll keep the `clone()` snapshot as-is for that reason. Happy to add a code comment noting why `Map.copyOf` isn't a safe substitute, if that'd help future readers.

reinhapa (approved):

exist-core/src/main/java/org/exist/xquery/functions/AccessUtil.java
```diff
@@ -0,0 +1,221 @@
+/*
+ * Copyright (C) 2014, Evolved Binary Ltd
```
dizzzz, line 2: please recheck the (double) header

exist-core/src/main/java/org/exist/xquery/functions/AccessUtil.java
```diff
@@ -0,0 +1,221 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.exist.xquery.functions;
+
+import com.evolvedbinary.j8fu.tuple.Tuple2;
+import io.lacuna.bifurcan.IMap;
```
dizzzz, line 36: is `io.lacuna.bifurcan` a bit of overkill for us here?

dizzzz (approved):

exist-core/src/main/java/org/exist/xquery/functions/AccessUtil.java
duncdrum: Not a duplicate — standard template for FusionDB-ported files: a scoping notice (lines 1–16, limits the grant to this snapshot) followed by the LGPL boilerplate, which itself opens with a copyright line. Same shape in `MapUtil.java`, `IPUtil.java`, etc. Also predates this rebase — original content from Adam Retter's 2022 commit (475cab83c8), header included. Leaving as-is.

exist-core/src/main/java/org/exist/xquery/functions/AccessUtil.java
duncdrum: Fair point — done in 452100f. These rule maps are built once at module-init and only ever read, so bifurcan's persistent-structure value doesn't apply here. Swapped in `HashMap`/`HashSet` + `Collections.unmodifiableMap`/`Set`, updated the 4 call sites (`FnModule`, `FunEnvironment`, `UtilModule`, `SystemProperty`) and `AccessUtilTest`. All 11 related tests pass.

Issue #6704, opened 2026-09-13
[BUG] XQTS `Constr-inscope-1` to `-4` flip nondeterministically, putting false red on every PR's conformance gate


joewiz:
[This issue was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

## Summary

The four XQTS tests `Constr-inscope-1`, `Constr-inscope-2`, `Constr-inscope-3` and `Constr-inscope-4` produce nondeterministic results between runs. The XQTS comparison bot reports them as newly passing or newly failing on pull requests that cannot possibly affect XQuery conformance, including a documentation-only change.

The practical cost is not the four tests themselves — it is that every PR's conformance report now carries noise that a reviewer has to recognize and discount by hand.

## Evidence

Deltas reported by the XQTS comparison bot against `develop`, on recent unrelated PRs:

| PR | what it changes | reported `Constr-inscope` delta |
| --- | --- | --- |
| #6666 | documentation only | `-1` newly failing |
| #6687 | OASIS catalog wiring | `-1` newly failing, `-2` newly passing |
| #6661 | dependency pinning | `-2` newly failing |
| #6433 | `install-and-deploy` version handling | `-2` newly failing, `-3` newly passing |
| #6692 | URL-rewrite response header buffering | `-1`, `-3`, `-4` newly failing |
| #6694 | URL-rewrite response wrapper fixes | `-2` newly passing |

A documentation-only PR cannot change conformance results, and the same four names flip in **both** directions across the set. Whatever varies is in the harness or in the tests' own assumptions, not in the code under test.

## Why this is worth fixing rather than tolerating

The visible symptom is a false negative: an approved, otherwise-clean PR appears to regress conformance, and the author has to explain the numbers away in the thread.

The more serious risk is the inverse. Once reviewers learn that `Constr-inscope-*` movement is meaningless, a genuine conformance regression that happens to touch those tests — or that appears alongside them — gets waved through as "probably the flaky ones again". A gate that reports noise trains people to ignore it.

## Suggested next steps

1. Run the four tests repeatedly against an unchanged `develop` to confirm the flakiness is reproducible outside CI and to establish the flip rate.
2. Determine whether the nondeterminism is in the tests' in-scope-namespace expectations, in result ordering, or in harness state leaking between tests — the `Constr-inscope` family is about in-scope namespace bindings on constructed elements, which is the kind of thing that can depend on evaluation order or on a shared context.
[...]

PR #6728, opened 2026-09-19
[bugfix] Deprecate xqdm:scan and the xqdoc extension module for removal

### LINKED ISSUE #6717: [question] Should extensions/xqdoc (xqdoc:scan) be deprecated for 7.0?
[This issue was prompted by Joe, drafted by Claude Code, and reviewed by Joe.]

## Summary

`extensions/xqdoc` ships in the default build and provides `xqdm:scan`. It is built on the external `org.xqdoc:xqdoc-conversion` library, whose only release on Maven Central is **1.6, published 2012-08-17** — before XQuery 3.0 was a recommendation.

It does not work. Not "works but is dated": on current `develop` it fails on eXist's own bundled libraries, including the exact modules its own test suite points at. Those tests are all disabled, so nothing notices.

This is a question rather than a proposal — I do not know the history here, and there may be users or intentions I cannot see from the repository. But the current state seems worth surfacing before 7.0.

## What I measured on current `develop`

**The shipped tests are inert.** All four tests in `extensions/xqdoc/src/test/xquery/xqdoc/tests.xml` carry `ignore="true"`. Running the module's suite reports:

```
Tests run: 4, Failures: 0, Errors: 0, Skipped: 4
BUILD SUCCESS
```

Green, having verified nothing.

**Its own test targets fail.** Running the two shipped test queries directly, rather than through the ignore flag:

| query (verbatim from `tests.xml`) | result |
| --- | --- |
| `xqdm:scan(xs:anyURI("http://exist-db.org/xquery/kwic"))` | `error while scanning module: line 51:8: unexpected token: version` |
| `xqdm:scan(xs:anyURI("resource:org/exist/xquery/lib/kwic.xql"))` | `error while scanning module: line 51:8: unexpected token: version` |

`kwic.xql` is one of our own bundled libraries. That failure is presumably why the tests were disabled in the first place.

**Anything written since XQuery 3.0 is rejected.** Each of these is a minimal module that fails inside `Scan.java`:

| construct | error |
| --- | --- |
| `map(*)` in a signature | `unexpected token: (` |
| `%private` annotation | `unexpected char: '%'` |
| `map { "k": 1 }` | `unexpected token: {` |
| `[ 1, 2 ]` | `unexpected token: [` |
| `(1,2) ! (. + 1)` | `unexpected char: '!'` |

**There is nothing newer to upgrade to.** Maven Central holds exactly one release of `org.xqdoc:xqdoc-conversion`: `1.6`, 2012-08-17.

**No caller in the org.** GitHub code search across both the `eXist-db` and `existdb` accounts returns a single hit for `xqdoc:scan` — `extensions/xqdoc/ant-example.xml`, in this repository. There is no production caller in any org repo.

## The caveat on that last point

Org-account code search says nothing about **third-party applications**. Someone outside our repositories may be calling `xqdm:scan`, and the fact that it fails on modern modules does not prove it fails on theirs — a module written to XQuery 1.0 might still scan. I could not construct a working example, but "I could not" is weaker than "there is none".

So if deprecation is the direction, it seems worth a deprecation cycle with a release-note mention rather than a removal, and worth asking on the mailing list rather than deciding from the repository alone.

## Why raise it now

The practical cost today is small but not zero: a module in the default build, a dependency thirteen years without a release that any security scan will eventually flag, and a test suite that reports success while testing nothing. The larger cost is that someone eventually tries `xqdoc:scan`, gets `unexpected char: '%'` on ordinary modern XQuery, and files a confusing bug.

Note `inspect:inspect-module` is a **different** implementation — eXist's own `XQDocParser.g` grammar in `exist-core` — and is unaffected by any of this. It is the introspection entry point actually used across the ecosystem. So deprecating `extensions/xqdoc` would not remove xqdoc-style introspection from eXist.

## Questions

1. Is `extensions/xqdoc` still intended to be supported, or has it been superseded by `inspect:inspect-module` in practice?
2. If it is to stay, is anyone willing to own replacing or forking the 2012 dependency? The alternative is that it keeps shipping in a state where its own tests cannot be enabled.
3. If it is to go, does deprecate-in-7.0 / remove-later seem right, and should it be raised on the mailing list first given possible third-party callers?

## Related

- eXist-db/exist#1386 — the `@`-truncation bug in `inspect:inspect-module`'s grammar. Unrelated code, but it is what led me here; the two are easy to conflate because both are called "xqdoc".

joewiz:

Closes #6717

## Summary

Deprecates `xqdm:scan` and the `extensions/xqdoc` module for removal in a later release, as agreed in #6717 — +1 from @duncdrum for 7.0.0, +1 "for deprecation for removal" from @line-o.

The function stays callable. This adds the notice, nothing else.

## Why

**The dependency has had one release, in 2012.** `org.xqdoc:xqdoc-conversion` 1.6, published 2012-08-17 — before XQuery 3.0 was a recommendation. Maven Central holds nothing newer, so there is nothing to upgrade to.

**It cannot read modern XQuery.** Each of these minimal modules fails inside `Scan.java`:

| construct | error |
| --- | --- |
| `map(*)` in a signature | `unexpected token: (` |
| `%private` annotation | `unexpected char: '%'` |
| `map { "k": 1 }` | `unexpected token: {` |
| `[ 1, 2 ]` | `unexpected token: [` |
| `(1,2) ! (. + 1)` | `unexpected char: '!'` |

**It cannot read eXist's own bundled libraries.** Both queries in the module's own test suite — scanning `kwic.xql` by module URI and from the classpath — return `error while scanning module: line 51:8: unexpected token: version`.

**Its test suite is inert.** All four tests in `extensions/xqdoc/src/test/xquery/xqdoc/tests.xml` carry `ignore="true"`, so the module reports `BUILD SUCCESS` while verifying nothing:

```
Tests run: 4, Failures: 0, Errors: 0, Skipped: 4
BUILD SUCCESS
```

Given the previous point, that is presumably *why* they are disabled.

**Nothing calls it.** A GitHub code search across the `eXist-db` and `existdb` accounts returns a single hit for `xqdoc:scan`: `extensions/xqdoc/ant-example.xml`, in this repository.

## What Changed

Both `xqdm:scan` signatures and the module description carry a deprecation notice, using eXist's existing mechanism (`FunctionSignature`'s `deprecated` argument). The notice names the concrete reasons and points at the replacement.

**This does not remove xqdoc-style introspection from eXist.** `inspect:inspect-module` is a *different* implementation — eXist's own `XQDocParser.g` grammar in `exist-core` — and is unaffected by any of the above. The two are easy to conflate because both are called "xqdoc"; the notice says so explicitly, and it is worth repeating here.

## Deprecated, not disabled

The notice reaches users through three channels, and the function remains callable in a default configuration:

| channel | shows |
| --- | --- |
| `util:describe-function` | the `<deprecated>` notice |
| `inspect:inspect-module` | the same, on the function |
| module description | "DEPRECATED FOR REMOVAL … See …/issues/6717." |

An administrator who sets the existing `PROPERTY_DISABLE_DEPRECATED_FUNCTIONS` turns a call into `XPST0017`. Removal is a separate, later step.

## An observation about how eXist surfaces deprecation

Worth recording while touching this, though this PR changes nothing about it.

A deprecation notice in eXist is **discoverable, not announced**. There are exactly three consumers of `FunctionSignature#isDeprecated()` in the codebase:

| consumer | behavior |
| --- | --- |
| `FunctionFactory` | throws `XPST0017` — but only when `xquery.disable-deprecated-functions` is set |
| `InspectFunctionHelper` | emits the `<deprecated>` element |
| `DescribeFunction` | emits the same |

and `DISABLE_DEPRECATED_FUNCTIONS_BY_DEFAULT` is `false`.

So on a default install, calling a deprecated function produces no warning, no log entry, and no diagnostic of any kind. The notice is visible only to someone who goes looking for it with `util:describe-function` or `inspect:inspect-module` — which is to say, to someone who already suspects there is something to find. The table above describes what those introspection calls *return*; it is not a warning any caller sees.

That limits what deprecating `xqdm:scan` can accomplish on its own: it documents the decision and gives administrators an opt-in enforcement switch, but it will not tell an existing user of the function that anything has changed.

Whether eXist should log on a deprecated call — either here in `Scan`, or generally in `FunctionFactory` alongside the existing disable check — seems worth a separate conversation rather than a drive-by change in this PR. Noting it here so the limitation is on the record with the deprecation it affects.

## Test Plan

- [x] New `XQDocDeprecationTest` pins all three notice channels plus the fact that the function remains available — 4/4
- [x] Module suite green (the 4 pre-existing `ignore="true"` tests still skip, untouched)
- [x] Codacy PMD clean on the changed files; `license:check` clean

Worth noting: `XQDocDeprecationTest` appears to be the first test in `extensions/xqdoc` that actually runs. The four that were already there have never executed.

## Out of scope

Removal itself, and the question of whether `extensions/xqdoc` should be dropped from the default build. Both belong to the later step #6717 anticipates.


duncdrum (approved):
@joewiz let's create a new  issue for the deprecation surfacing, this is high-priority for 7.0.0 there is a number of deprecations

joewiz:

Thanks @duncdrum — filed as #6733, marked for 7.0.0.

It covers the three-consumer picture from the PR description above, plus two things that turned up while writing it. First, `FunctionSignature` appends "This function could be removed in the next major version release!" to every notice automatically — and 7.0.0 is that release, so by our own stated policy every currently-deprecated function is nominally in scope for removal, with no signal ever having reached the people using them. Second, the one enforcement switch we have is all-or-nothing and fatal: an administrator who wants to know whether their applications call deprecated functions has to break those applications to find out. Underneath both is that eXist-db has no warnings channel at all, so there is currently nowhere for a non-fatal diagnostic to go.

The issue includes the current inventory — 13 deprecated function arities across 5 modules on a default install — and a query to reproduce it, so the scope is concrete rather than "there are some."

The XQTS comparison on this PR shows `Constr-inscope-4` newly failing. That is the known flake tracked in #6704, not this change.

reinhapa (approved):

PR #6729, opened 2026-09-20
[bugfix] Preserve CDATA sections through request parsing and serialization

### LINKED ISSUE #2081: Incorrectly serialized brackets in HTML/JavaScript

joewiz:

Closes #2081

## Summary

Saving an HTML or XML document containing a CDATA section escaped its content: `<![CDATA[ if (a > b) ]]>` inside a `<script>` block came back as `if (a &gt; b)`, breaking embedded JavaScript. Reported in 2018; @line-o confirmed the remaining part still present on 7.0.0-beta5 last week.

It is two defects in this repository, at opposite ends of the same round trip. The second only became visible once the first was fixed — while the write path was destroying CDATA, nothing intact ever reached storage for the read path to spoil.

## 1. `request:get-data()` discarded CDATA sections

A SAX parser reports the *content* of a CDATA section through `characters()` like any other text, and marks the boundaries only through the `LexicalHandler` callbacks. `DocumentBuilderReceiver` implements `LexicalHandler` but had both boundary callbacks as explicit no-ops:

```java
public void startCDATA() { /* no-op */ }
public void endCDATA()   { /* no-op */ }
```

so the boundaries were dropped and the content was built as plain text. Nothing downstream can recover the distinction once it is gone.

`SAXAdapter` — the sibling adapter into the same `MemTreeBuilder`, backing `fn:parse-xml` — already did this correctly: set a flag on `startCDATA`, buffer `characters()` while set, flush through `builder.cdataSection()` on `endCDATA`. This applies the same pattern, so the two adapters agree on the same bytes.

Everything else was already in place: memtree has `CDATASectionImpl`, `MemTreeBuilder` has `cdataSection()`, and `DocumentBuilderReceiver` even had a working `cdataSection()` `Receiver` method — so a Receiver-to-Receiver copy kept CDATA while a parse did not.

**Limitation, stated so this is not read as more than it is:** a section sandwiched between two text runs still coalesces into one text node. `MemTreeBuilder.cdataSection()` appends to a preceding text node by design ("XML does not allow adjacent text nodes") and `SAXAdapter` behaves identically, so this is pre-existing builder behavior and the two adapters still agree. #2081's case — an element whose entire content is one section — is unaffected.

## 2. `fn:serialize` could not emit a stored CDATA section

With the above fixed, sections reach storage intact — and there was then no way to get one back out. An editor that opens a document, displays it and writes it back therefore still corrupted it *on open*, whatever the database held.

XDM has no CDATA node kind, so `fn:serialize` is specified to escape such content unless the containing element is named in `cdata-section-elements`. That is correct and is unchanged here. But it left eXist unable to return what it had stored: the DOM does keep CDATA nodes and eXist's own REST serializer emits them — fetch the same document over REST and the section is there — that path was simply unreachable from `fn:serialize`.

`cdata-section-elements` is not a substitute. It requires knowing the element names in advance, and it *imposes* CDATA on elements that never had it — the same silent rewriting this issue is about.

So this adds **`exist:preserve-cdata`**, opt-in and defaulting to false, following `exist:expand-xincludes`:

```xquery
serialize($doc, map {
    "method": "xml",
    QName("http://exist.sourceforge.net/NS/exist", "preserve-cdata"): true()
})
```

Three changes, and the third is easy to miss:

| change | why |
| --- | --- |
| `EXistOutputKeys.PRESERVE_CDATA` | the key |
| `XMLWriter` honors it at both gates in `startCdataSection()`/`endCdataSection()` | those suppress the delimiters whenever XDM serialization is active; `characters()` already writes raw while inside a section |
| `SerializerUtils.ExistParameterConvention` lists it | **that enum is a closed whitelist, and an unlisted `exist:`-namespace parameter is silently discarded rather than rejected** — so the first version of this change looked like a fix that did nothing |

That last point is worth its own issue and I have filed one: a typo in an `exist:` serialization parameter today fails quietly.

## Verification

Both fixes have proper red/green, and the tests were checked against the unfixed code rather than only against the fix:

- `DocumentBuilderReceiverCDataTest` — 4 cases; **2 fail** without fix 1. The sharpest is the parity case, which compares both adapters on identical bytes and fails even though `SAXAdapter` alone would pass.
- `PreserveCdataSerializationTest` — 5 cases; **2 fail** without fix 2. The other three assert *unchanged* default behavior (plain `fn:serialize` still escapes, explicit `false()` matches the default, a document without CDATA serializes identically) and pass either way — which is what keeps this from being a spec deviation.
- Regression sweep: **1,102 tests, 0 failures** across `org.exist.dom.memtree.*`, `org.exist.util.serializer.*`, `org.exist.xquery.functions.request.*` and the full `src/test/xquery` corpus. `DocumentBuilderReceiverTest` and its integration counterpart — the pre-existing tests for the class changed here — are included.
- Codacy PMD clean on the changed files; `license:check` clean.

**End to end, not only in unit tests:** an eXist built from this branch, with eXide asking for the parameter in `db:load-document` and sending its editor buffer as bytes, takes eXide's CDATA round-trip test from failing to passing. The same spec fails against a stock build, so it is measuring these fixes rather than the test conditions. The eXide side is a companion PR.


duncdrum (approved):

duncdrum (approved):

duncdrum (commented):
The code changes look good. 

I m not sure why the new option is false by default? We are fixing a long standing bug in 7. Implementation that depend on the bugged behaviour, should change. Everybody else should get the fixed behaviour OOTB

We need to scan the docs, chances are the new param should be added there.

duncdrum (commented):
Can we not also dedupe CDATA handling between the different consumers?

reinhapa (approved):

duncdrum (changes requested):
Thank you @joewiz could you add the documentation for the new parameter [exist specific options](https://github.com/eXist-db/documentation/blob/17291e1a6cba093836a13d14d6de74d5d49cbedd/src/main/xar-resources/data/xquery/xquery.xml#L569) and also switch it to active by default?

joewiz:

Thanks @duncdrum. Taking your three points in reverse order, because the third one was easy and the first one is the interesting one.

## Dedupe

Done in 6adf50da60. `SAXAdapter` and `DocumentBuilderReceiver` now share a `CDataSectionBuffer` that holds the reassembly state and hands the finished section to `MemTreeBuilder.cdataSection()`. Both classes keep their own `LexicalHandler` methods — they are separate SAX handlers with separate lifecycles — but the buffering logic exists once. 1,035 tests green across `org.exist.dom.memtree.*`, `PreserveCdataSerializationTest` and `xquery.CoreTests`.

## Docs

Scanned. The page is `data/xquery/xquery.xml` in eXist-db/documentation, which documents the `exist:` serialization parameters in two lists — once as prose under "eXist-db specific options" and again under `fn:serialize`. I have a branch adding `preserve-cdata` to both, and I checked that all three invocation forms the page advertises really do reach the new parameter: the prolog `declare option exist:serialize` path writes key/value pairs straight into the properties with no whitelist, and `fn:serialize#2` and `file:sync#3` both go through `SerializerUtils.getSerializationOptions`, which is where the new enum entry lives. I will open that PR against eXist-db/documentation and link it here.

## Why the parameter defaults to false

This is the part I want to lay out properly, because I think the PR reads as one bug fix when it is actually two changes with quite different risk, and the second one is where the default question lives.

**The first change is an unconditional bug fix and has no flag.** `DocumentBuilderReceiver.startCDATA()` and `endCDATA()` were empty methods, so every CDATA section arriving through `request:get-data()` was silently flattened into the surrounding text. Its sibling handler `SAXAdapter` already did this correctly — the two classes had simply drifted. That is fixed for everybody, out of the box, with nothing to turn on. It is the defect #2081 actually reports.

**The second change adds a capability, and here eXist's current output is not a bug.** XDM has no CDATA node kind. A CDATA section in a source document contributes an ordinary text node, and the serialization spec emits CDATA sections only for the elements named in the standard `cdata-section-elements` parameter. So `fn:serialize` escaping that text today is the conformant answer, not a defect — which makes flipping the default a deliberate decision to diverge from the spec by default, rather than a fix.

What makes the situation genuinely untidy — and I think this is the real substance of #2081 — is that eXist is already inconsistent with itself, and the inconsistency runs along the `xdm-serialization` line:

| surface | `xdm-serialization` | CDATA today |
| --- | --- | --- |
| retrieving a stored document (REST GET, XML-RPC `getDocument`) | `no` | **preserved** |
| returning the same document from a query (`XQuerySerializer`, XML:DB query services, `fn:serialize`) | `yes` | escaped |

So an editor that fetches a document one way and saves it back the other way loses its CDATA sections — which is exactly the eXide round-trip in the linked issue. The parameter closes that gap; the default decides whether the query surface moves to match the retrieval surface, or stays with the spec.

I do not think I can settle that by argument, and I would rather measure it than guess. Flipping the default only affects the `xdm-serialization=yes` half of that table, so the blast radius is precisely "query results and `fn:serialize`" — which is the half XQTS exercises. If you want, I will push a branch with the default set to true and let the XQTS comparison job run against it. If that comes back with zero newly failing tests, the conformance objection is empty and I will flip the default here without further argument. If it comes back red, we will have the specific list of tests in front of us and can decide with real numbers whether they are worth it.

Happy to go either way — say the word and I will start the measurement run.

duncdrum:
@joewiz yes please, let's run XQTS and the whole testsuite with default true. So we have something actual to base the decision on.

joewiz:

@duncdrum Measured. Branch `measure/preserve-cdata-default-true` on my fork is this PR with the default flipped on — all three places it is expressed — and nothing else. Full test suite and XQTS both ran against it.

The short version: the flip breaks things, and the most interesting breakage is in exactly the use case the parameter exists to serve. I think the data argues for keeping the default off, but it also points at a better question than the one I asked.

## XQTS

| | `develop` | this PR (default off) | default on |
| --- | ---: | ---: | ---: |
| Passed | 28,846 | 28,845 | 28,842 |
| Failures | 1,568 | 1,569 | 1,570 |

Comparing per-test outcomes between this PR's run and the flipped run — same code, the default the only difference — one test regresses for real:

- **`Constr-ws-gencdata-1`** — newly failing.
- `Constr-inscope-1`, `-3` — the known flake, #6704.
- `raytracer`, `Constr-ws-enclexpr-22` — absent from the flipped run; the runner labels both `recording-drift`, the non-determinism in eXist-db/exist-xqts-runner#74. Not related to the change.

The regressing test is `declare boundary-space strip; <elem><![CDATA[ ]]></elem>`, expected to produce `<elem> </elem>`. With the default on we emit `<elem><![CDATA[ ]]></elem>`.

## Full test suite

7,339 tests, 12 failures. One is `MoveResourceTest.testMove` timing out, which is the documented flaky infrastructure test and is unrelated. The other 11 are the flip:

| test | what it shows |
| --- | --- |
| `XQuery3Tests.serialize-html-5-raw-text-elements-head` | `<style>`/`<script>` in HTML5 output |
| `XQuery3Tests.serialize-html-5-raw-text-elements-body` | the same |
| `XQuery3Tests.serialize-html-5-needs-escape-elements` | the same family |
| `XQueryTest.cdataQuery` | stored document, escaped on output |
| `XPathQueryTest.cdataPersistentDom` (local + remote) | the same |
| `XPathQueryTest.cdataMemtreeDom` (local + remote) | the same |
| `XPathQueryTest.cdataComputedDom` (local + remote) | the same |
| `PersistentDomTest.cdata` | the same |

The integration suites passed on all three platforms; the failures are unit only.

## The part worth reading

Those failures are not all the same kind, and the difference matters.

The last five are tests that assert the current default. If we decided to flip, we would update them, and nobody could object — they encode the value under discussion.

The HTML5 ones are not that. Here is `serialize-html-5-raw-text-elements-head`, unmodified:

```xquery
<html>
    <head>
        <style><![CDATA[ul > li { color:red; }]]></style>
        <script><![CDATA[if (a < b) foo()]]></script>
    </head>
    <body></body>
</html>
=> serialize($ser:opt-map-html5)
```

expecting `<style>ul > li { color:red; }</style>`. With the default on we produce:

```
<style><![CDATA[ul > li { color:red; }]]></style>
```

HTML has no CDATA sections. A browser reading that sees `<![CDATA[` as literal CSS, and the rule does not apply. So flipping the default emits invalid HTML5 — in `<script>` and `<style>`, which is the precise case the parameter was added to help.

The mechanism is `CDATAConstructor.java`, which turns a CDATA section written in *query source* into a real CDATA node. The test author wrote `<![CDATA[...]]>` there for the ordinary reason anyone does — so they could type `>` and `<` in CSS and JavaScript without escaping them. They were not asking for a CDATA section in the output. With the default on, they get one anyway.

That is the flaw in my own framing earlier in this thread. I described the gap as retrieval-preserves versus query-escapes, which is true, but flipping the default does not only affect documents coming back out of the database. It also changes markup an author typed into a query, where CDATA is usually just escaping convenience. That is a much wider surface than #2081 is about, and the W3C test and the HTML5 tests are both on the wrong side of it.

## Where that leaves it

You are right that a long-standing bug should be fixed OOTB, and the part of this PR that *is* a bug fix already is: `DocumentBuilderReceiver` no longer drops CDATA sections, unconditionally, no flag. Everybody gets that.

For the serialization half, I would keep the default off on this PR, because the evidence says on-by-default is not free and one of the costs is producing invalid HTML5.

But the measurement suggests a better option than either default, which I had not considered before running it: scope `preserve-cdata` so it applies to nodes that came from a parsed or stored document and not to nodes built by a constructor in query text. That would leave the HTML5 output and `Constr-ws-gencdata-1` alone, since both are constructor cases, while still giving eXide its round-trip. If that holds up, the OOTB argument gets much stronger and the default question is worth reopening on better terms.

I have not verified that it is cleanly implementable — the CDATA node does not currently record where it came from, so it likely needs a provenance flag, and I would want to measure it the same way before claiming anything. Happy to try it on a branch if you think it is the right direction. Otherwise I would merge this as-is, opt-in, and treat the scoped version as a follow-up.

duncdrum:
@joewiz this sounds like a better option indeed. I would be curious to hear how that plays out.

joewiz:

@duncdrum It plays out well. Scoping removes every failure that represented real breakage, and with it in place your original position is supportable: on-by-default costs nothing in conformance.

## The change

One line, and not the provenance flag I expected to need:

```java
// CDATAConstructor.eval
final NodeImpl node = builder.getDocument().getNode(builder.characters(cdata));
```

A CDATA section in query source is an escaping convenience — it lets an author type `<` and `>` without entity references. XDM has no CDATA node kind, so what it contributes is character content, indistinguishable from the same characters typed directly. Building a text node makes the provenance structural rather than tracked: after this, a CDATA node in eXist's trees only ever comes from a parsed or stored document, which is exactly what `preserve-cdata` needs it to mean.

Two things make it safe, both checked rather than assumed. `XMLWriter.characters` already applies `cdata-section-elements` to text nodes, so that standard parameter keeps working on constructed content. And the existing `cdata*` tests all assert serialized output, not DOM node types, so nothing was pinning the constructor to producing a CDATA node.

## Results

Default off — the shape this PR would merge as:

| | result |
| --- | --- |
| Unit suite | **7,341 tests, 0 failures**; License, and integration on Linux, macOS and Windows all green |
| XQTS | 28,845 / 1,569 — identical to this PR before the change; only `Constr-inscope-1/2/3/4` churn, which is #6704 |

Default on, with scoping:

| | unscoped (last comment) | scoped |
| --- | ---: | ---: |
| XQTS newly failing | 1 — `Constr-ws-gencdata-1` | **0** |
| XQTS totals | 28,842 / 1,570 | **28,846 / 1,568** — the `develop` baseline exactly |
| HTML5 serialization | 3 failures, invalid output | **0** |
| `cdataComputedDom` | 2 failures | **0** |
| tests asserting the old default | 6 | 6 |
| total unit failures | 11 | **6** |

The six that remain are all one shape:

```
expected:<<cdataText>Hello there "Bob?"</cdataText>>
but was:<<cdataText><![CDATA[Hello there "Bob?"]]></cdataText>>
```

Stored and parsed documents preserving their CDATA — the feature doing its job. Those tests encode the default itself, so flipping it means updating them, which is not an objection.

## What I think, including the part that argues against me

The scoping change is worth having whatever we decide about the default, because emitting a CDATA section into an HTML5 `<script>` was wrong independently of how the parameter is defaulted. I have added it to this PR with two tests pinning the distinction.

On the default: the honest reading is that **no test catches on-by-default, which is not the same as it being conformant.** Serializing a stored document's CDATA section back out through `fn:serialize` still diverges from the W3C serialization spec, which emits CDATA only for elements named in `cdata-section-elements`. XQTS does not happen to cover that path, so a clean XQTS run is evidence that nothing breaks, not evidence that we comply.

So flipping it is a deliberate choice to diverge from the spec by default in exchange for a behavior most users would find less surprising — the retrieval and query surfaces finally agreeing. I think that is defensible in a major release, and it is opt-out with an explicit `false`. But it should be made as that choice rather than as "the tests are green", and it is more your call than mine.

Two ways forward, and I am happy with either:

1. **Merge this as-is, opt-in.** The bug fix ships OOTB, the parameter is available, and the default can be revisited for 7.0.0 in its own PR with the six expectations updated.
2. **Flip it here.** Say the word and I will update the six tests and push it, with the divergence noted in the parameter's documentation so it is on the record.

Branches, if you want to look: `bugfix/2081-cdata-scope-to-parsed` is what is now on this PR; `measure/preserve-cdata-scoped-default-true` is the flipped measurement, not for merge.

One unrelated thing worth flagging, found while doing this: one of the XQTS jobs reported **success having run zero tests** — a transient 504 fetching `qt3tests/archive/master.zip`, after which the job went green with no results and no comparison. `find … -exec … \;` returns `find`'s status rather than the runner's, so a runner failure cannot fail that step. I checked the last 40 XQTS runs here and none of the successful ones is missing its artifact, so this is a latent trap rather than something corrupting results today. Filing it separately.

duncdrum:
Phew I like the scoping a lot. Diverging from the spec by default takes a broader vote, what do others think @eXist-db/core ?

Issue #1768, opened 2018-03-13
UnsupportedOperationException on BinaryValueFromFile#incrementSharedReferences


wolfgangmm:
Since b80b97abe8372a05a0e47ab1d37d579e9ea17843, `BinaryValue` has a method `incrementSharedReferences`. It is only implemented by `BinaryValueFromInputStream` while `BinaryValueFromFile` and `BinaryValueFromBinaryString` throw an UnsupportedOperationException. This happens eg. in the following test snippet:

```xquery
let $binData := file:read-binary("/Users/wolfgang/Source/eXist/develop/VERSION.txt")
return
    element { "root" } {
        "bla"
    }
```

Because `incrementSharedReferences` can obviously be called on those classes, it should not just throw an exception but rather do nothing. Commenting out the `throw` fixes the issue.

pherk:
should reference #1740

Issue #1740, opened 2018-02-20
Upload via collection browser fails with server error


pherk:
I report here because it seems to me that it is an exist error and not a dashboard error

### What is the problem
> upload via collection browser fails with 'server error'

### Describe how to reproduce or add a test
> click on upload resources, try upload a xml or xqm file

### Context information

- eXist-db 4.1-snapshot , Dashboard 1.0.0, eXide 2.4.3
- Java8
- Ubuntu 16.04LTS
- 64 bit
- no custom changes

exist.log:
```
2018-02-20 09:46:52,132 [qtp393549265-215] ERROR (EXistServlet.java [doPost]:477) - java.lang.UnsupportedOperationException
2018-02-20 09:46:52,134 [qtp393549265-215] WARN  (HttpChannel.java [handleException]:568) - /exist/apps/dashboard/plugins/browsing/upload.xql
javax.servlet.ServletException: javax.servlet.ServletException: An error occurred while processing request to /exist/apps/dashboard/plugins/browsing/upload.xql: An unknown error occurred: null
        at org.eclipse.jetty.server.handler.HandlerCollection.handle(HandlerCollection.java:146) ~[jetty-server-9.4.8.v20171121.jar:9.4.8.v20171121]

... snip

Caused by: java.lang.UnsupportedOperationException
        at org.exist.xquery.value.BinaryValueFromFile.incrementSharedReferences(BinaryValueFromFile.java:123) ~[exist.jar:4.1.0-SNAPSHOT]
        at org.exist.xquery.XQueryContext.enterEnclosedExpr(XQueryContext.java:3546) ~[exist.jar:4.1.0-SNAPSHOT]
        at org.exist.xquery.EnclosedExpr.eval(EnclosedExpr.java:80) ~[exist.jar:4.1.0-SNAPSHOT]
        at org.exist.xquery.AbstractExpression.eval(AbstractExpression.java:71) ~[exist.jar:4.1.0-SNAPSHOT]
        at org.exist.xquery.PathExpr.eval(PathExpr.java:276) ~[exist.jar:4.1.0-SNAPSHOT]
        at org.exist.xquery.ElementConstructor.eval(ElementConstructor.java:329) ~[exist.jar:4.1.0-SNAPSHOT]
        at org.exist.xquery.DebuggableExpression.eval(DebuggableExpression.java:58) ~[exist.jar:4.1.0-SNAPSHOT]
[...]

PR #6731, opened 2026-09-20
[bugfix] Release a binary value with the scope that created it

### LINKED ISSUE #5030: [BUG] Error trying to access binary data from EXPath HTTP response
**Describe the bug**
Sometimes I get an error “Underlying channel has been closed” when trying to do something with the response returned from `http:send-request()`.

After a lot of testing, I’ve narrowed down the problem scenario:

1. Use the HTTP Client to send a request to a URL which produces binary content (e.g. `application/json`)
2. Store the HTTP response in a variable `$a`
3. Pass `$a` to a function `b()`
4. If `b()` succeeds, try to do something with the response in `$a` (storage, decoding, etc.)

Number 4 fails with the error.

I ran into this when I started updating a function library to be more DRY. My first version of an XQuery had a lot of duplicate code, but didn’t produce any errors. Once I started testing the response code in another function, I started hitting this error over and over again.

**Expected behavior**
I expected that the binary response contained in `$a` would still be usable after parsing it with `b()`.

**To Reproduce**

```Xquery
xquery version "3.1";

(:  LIBRARIES  :)
(:  NAMESPACES  :)
  declare namespace http="http://expath.org/ns/http-client";
  declare namespace util="http://exist-db.org/xquery/util";
  declare namespace xdb="http://exist-db.org/xquery/xmldb";

(:~
  Trying to reproduce "Underlying channel has been closed" error when decoding 
  xs:base64Binary JSON.
  
  @author Ash Clark
  @since 2023
 :)
 
(:  VARIABLES  :)
  
  declare variable $t:request-url := 
    "https://wwp.northeastern.edu/exist/restxq/reception/sources";
  declare variable $t:request-element :=
    <http:request method="GET" timeout="15">  </http:request>;


(:  FUNCTIONS  :)
  
  declare
    %test:setUp
  function t:setup() {
    xdb:create-collection("/db", "test_2023-08-25")
  };
  
  declare
    %test:tearDown
  function t:tearDown() {
    xdb:remove("/db/test_2023-08-25")
  };
  
  (:~
    Send a request to the test URL using the EXPath HTTP Client. Make sure that the 
    request returns a 200 response, with a JSON payload.
    
    @return the EXPath HTTP Client's response headers, in an XML element
   :)
  declare
    %test:name('Response headers are OK')
      %test:assertExists
      %test:assertXPath('$result/@status eq "200"')
      %test:assertXPath('$result/Q{http://expath.org/ns/http-client}header[@name eq "content-type"][contains(@value, "application/json")]')
  function t:json-response-1-headers() as item() {
    http:send-request($t:request-element, $t:request-url)[1]
  };
  
  (:~
    Send a request to the test URL. Make sure that the response body is of type 
    xs:base64Binary.
    
    @return the HTTP response body
   :)
  declare
    %test:name("JSON response body is xs:base64Binary")
      %test:assertXPath('$result instance of xs:base64Binary')
  function t:json-response-2-body() as item() {
    http:send-request($t:request-element, $t:request-url)[2]
  };
  
  (:~
    Send a request to the test URL. Make sure that the response body can be saved to 
    eXist as a binary JSON file.
    
    @return a string filepath to the stored resource (if storage doesn't work, an empty sequence)
   :)
  declare
    %test:name("JSON response body can be saved as binary")
      %test:assertExists
  function t:json-response-3-save() as xs:string? {
    let $responseBody := http:send-request($t:request-element, $t:request-url)[2]
    return
      xdb:store-as-binary('/db/test_2023-08-25', 'test.json', $responseBody)
  };
  
  (:~
    Send a request to the test URL. Make sure that the response body can be turned 
    into a JSON string with `util:binary-to-string()`.
    
    @return the response JSON as a string
   :)
  declare
    %test:name("JSON response body can be decoded without saving")
      %test:assertExists
  function t:json-response-4-decode() as xs:string? {
    let $responseBody := http:send-request($t:request-element, $t:request-url)[2]
    return
      util:binary-to-string($responseBody)
  };
  
  (:~
    Send a request to the test URL. Check that `@status` on the first item of the 
    response sequence is '200', and if so, turn the response body into a JSON string.
    
    @return if the response status was "200", the response JSON as a string (otherwise, nothing)
   :)
  declare
    %test:name("JSON response body can be decoded after testing the response status")
      %test:assertExists
  function t:json-response-6-test-status-first() as xs:string? {
    let $response := http:send-request($t:request-element, $t:request-url)
    let $status := $response[1]/@status/data(.)
    return
      if ( exists($response) and $status eq '200' ) then
        let $responseBody := $response[2]
        return util:binary-to-string($responseBody)
      else ()
  };
  
  (:~
    Send a request to the test URL. Use a function `t:get-response-status()` to test 
    the HTTP response status for a '200'. If so, turn the response body into a JSON 
    string.
    
    @return if the response status was "200", the response JSON as a string (otherwise, nothing)
   :)
  declare
    %test:name("JSON response body can be decoded after using a function to test the response")
      %test:assertExists
  function t:json-response-7-test-through-function() as xs:string? {
    let $response := http:send-request($t:request-element, $t:request-url)
    let $status := t:get-response-status($response)
    return
      if ( exists($response) and $status eq '200' ) then
        let $responseBody := $response[2]
        return util:binary-to-string($responseBody)
      else ()
  };
  
  (:~
    Send a request to a different API endpoint that should return HTML. Use a 
    function `t:get-response-status()` to test the HTTP response status for a '200'. 
    If so, return the response body.
    
    @return if the response status was "200", the response HTML (otherwise, nothing)
   :)
  declare
    %test:name("HTML response body can be accessed after using a function to test the response")
      %test:assertExists
  function t:response-html-8-content-type() as node()? {
    let $reqURL := 
      "https://wwp.northeastern.edu/exist/restxq/reception/sources/americancritreview"
    let $response := http:send-request($t:request-element, $reqURL)
    let $status := t:get-response-status($response)
    return
      if ( exists($response) and $status eq '200' ) then
        $response[2]
      else ()
  };
  
  (:~
    Send a request to an endpoint that should return HTML. Use a function 
    `t:get-response-status()` to test the HTTP response status for a '200'. If so, 
    normalize whitespace in the HTML to produce a string.
    
    @return if the response status was "200", the response HTML (otherwise, nothing)
   :)
  declare
    %test:name("HTML response body can be manipulated after using a function to test the response")
      %test:assertExists
  function t:response-html-9-cast-as-string() as item()? {
    let $reqURL := "https://wwp.northeastern.edu/exist/restxq/reception/sources/americancritreview"
    let $response := http:send-request($t:request-element, $reqURL)
    let $status := t:get-response-status($response)
    return
      if ( exists($response) and $status eq '200' ) then
        normalize-space($response[2])
      else ()
  };
  
  (:~
    Send a request to a URL that should produce plaintext. Use a function 
    `t:get-response-status()` to test the HTTP response status for a '200'. If so, 
    normalize whitespace on the response body.
    
    @return if the response status was "200", the response text in one line (otherwise, nothing)
   :)
  declare
    %test:name("Plain text response body can be manipulated after using a function to test the response")
      %test:assertExists
  function t:response-txt-10-test-through-function() as item()? {
    let $reqURL := "https://raw.githubusercontent.com/eXist-db/exist/develop/README"
    let $response := http:send-request($t:request-element, $reqURL)
    let $status := t:get-response-status($response)
    return
      if ( exists($response) and $status eq '200' ) then
        normalize-space($response[2])
      else ()
  };
  
(:  SUPPORT FUNCTION  :)
  
  (:~
    Given an HTTP response from the EXPath HTTP Client, return the HTTP status code. 
   :)
  declare function t:get-response-status($http-response as item()*) as xs:string? {
    if ( exists($http-response) ) then
      $http-response[1]/@status/data(.)
    else ()
  };
```

When I run the test suite above in v6.2.0, all tests pass except for `t:json-response-7-test-through-function()`, which errors out.

 - Build: eXist v6.2.0
 - Java: 1.8.0_292 (OpenJDK)
 - OS: Mac 12.6.2

**Additional context**
  - Tweaks made to keep whitespace from being added/removed (no indent, preserve whitespace).
  - I tried changing the Binary Manager to see if that would fix the bug (it didn't). It's back on the default `org.exist.util.io.FileFilterInputStreamCache` now.

### LINKED ISSUE #6725: [BUG] Passing a binary value to a user-defined function closes it for the caller
**Describe the bug**

When a user-defined function returns, its parameter variables are destroyed (`popLocalVariables` → `BinaryValueFromFile.destroy()` → `close()`). This closes the file channel of a binary value the caller still holds, even if the function never reads it. Follow-up to #6552: #6506 fixed `xmldb:store` closing its argument but not this case. Breaks publishing in public-repo on 7.0.0-beta5.

**To Reproduce**

POST a file as `files[]` to:

```xquery
declare function local:noop($b) { 1 };
let $b := request:get-uploaded-file-data("files[]")
return (local:noop($b), xmldb:store("/db", "t.xar", $b))
```

→ `error while obtaining length of binary value t.xar`. Without the `local:noop($b)` call the store succeeds; `string($b)` in place of the store fails with `Underlying channel has been closed`.

**Expected behavior**

`$b` stays usable after being passed to a function.

- Build: eXist-db 7.0.0-beta5 (`existdb/existdb:release`, rev 95b98270)

line-o:
### Description:

Passing a binary value to a user-defined function closed it for the caller:

```xquery
declare function local:noop($b) { 1 };
let $b := request:get-uploaded-file-data("files[]")
return (local:noop($b), xmldb:store("/db", "t.xar", $b))
```

When `local:noop` returns, `popLocalVariables` destroys the variables leaving scope — including the parameter alias, which points at the value the caller created and still holds. The caller's next read fails with `Underlying channel has been closed`, or, through `xmldb:store`, `error while obtaining length of binary value`.

The only thing protecting a value was the guard in `BinaryValue.destroy`: don't close if it is reachable from *this* scope's result. That asks who holds the value last rather than who owns it, and it has been patched repeatedly — #1740, #1768 (blanket reference counting around every element constructor), #6552 (`xmldb:store` closing its argument), then the nested-map escape fix. Each patch closed one shape of the bug.

**Ownership replaces the heuristic.** `XQueryContext` keeps registered values in a list divided into frames by a stack of start indices. A value belongs to the innermost frame open when it was registered, and only that frame may release it; a value reachable from the sequence a frame returns escapes to the enclosing frame, so eager release survives. A parameter alias owns nothing, so a callee cannot close its caller's value.

Frames are pushed by `markLocalVariables` and popped by `popLocalVariables`. Where the escape set is unknown — the overload without a result sequence — the frame *promotes* rather than releases: deferring release to the end of the query is recoverable, closing a value the query still needs is not. Nothing is allocated until a query registers its first binary value, which almost no query does.

`enterEnclosedExpr`/`exitEnclosedExpr` expressed the same idea as reference counts (entering took a reference on every registered value so that leaving released only those created inside — that is a frame). `EnclosedExpr` now opens a frame, and the walk over every live value on entry to every element constructor goes away with it. Reference counting remains for genuine lending, where `xmldb:store` hands its value to a `Resource` that closes what it is given.

Two repairs come with it:

- **`ModuleContext` delegated only `registerBinaryValueInstance`**, not the rest of the registry, so the enclosed-expression hooks and `destroyBinaryValue` acted on its own permanently empty deque. A module context — an XQSuite test function, or `util:eval` — ran a different code path than a main module and silently could not reproduce these bugs. That is why every existing regression test here had to be a Java harness. It now delegates the whole registry, and `binary-value.xqm` gains XQSuite coverage that is meaningful for the first time.
- **`FunctionReference.containsReference` now looks through the closure**, so a value captured by an inline function is seen to escape the scope that created it. `popLocalVariables` already skipped closure variables for the same reason.

Separable commits, if you would rather review or take them apart:

- `[bugfix] Leave a window condition's scope before opening the next` — `WindowExpr` marks the variable stack per window condition but pops only on the branches where a window starts or ends, so the stack grew for the length of the binding sequence. Fixed at the two sites where nothing holds the variables; the third is left alone and explained in the commit message, because there the branch hands those variables to `window.start()` before re-marking.
- `[optimize] Keep the variable-stack methods small enough to inline` — what the benchmark caught, see below.
- Two `[ignore]` reformat commits. The first ran `format.sh` over all thirteen files at once, which silently failed and formatted none of them, so it carries only a member reordering; the second ran it per file and is the actual reformat. Both are formatting only.

### Reference:

Closes https://github.com/eXist-db/exist/issues/5030
Closes https://github.com/eXist-db/exist/issues/6725 (filed separately, closed as a duplicate; its reproducer is one of the tests here)

Also relevant: #6552, #1768, #1740 — earlier shapes of the same bug.

### Type of tests:

**Java harness** (`AbstractBinariesTest`, run by the embedded, REST and XMLDB subclasses) — six new tests for values crossing a function boundary. Two of them reproduce the reported failures: against `develop` they fail with exactly the errors from the issues —

```
passBinaryToUserFunctionThenStore  → error while obtaining length of binary value passed-to-function.bin
passBinaryToUserFunctionTwice      → The underlying InputStream has been closed
```

— and pass here. The rest pin what must keep working: two levels of call, a value returned out of the function that created it, a value captured by a closure, and one carried through tail recursion.

**Unit** (`XQueryContextTest`) — frame mechanics, including the two ways the variable stack misbehaves in practice: frames left open must strand nothing, and a pop with no frame open must release nothing rather than reach into what is left.

**Leak direction** (`FilterInputStreamCacheMonitorTest`) — ownership defers release rather than closing early, so two tests assert that a value passed to a function is released exactly once and that values created in a loop do not accumulate for the whole query. The loop test deliberately asserts the count `ForExpr` gives today (one scope around the whole loop), so that limitation is written down rather than implied.

**XQSuite** (`binary-value.xqm`) — four passing-into-a-function shapes, meaningful only now that `ModuleContext` delegates the registry.

Full runs: `mvn test -pl exist-core` → 7342 tests, 0 failures; `mvn test -pl extensions/modules/file -Dtest='*BinariesTest'` → 50, 0 failures; window-clause tests separately for the `WindowExpr` commit.

### Performance:

Every scope now opens and closes a frame, so `BinaryValueLifetimeBenchmark` (new, in `exist-core-jmh`) exists to defend the queries that never create a binary value at all. Run on this branch and on `develop` on the same machine, 3 forks × 5 iterations of 5s after 3×5s warmup, `-prof gc`:

| benchmark | develop (ms/op) | branch (ms/op) | alloc MB/op (develop → branch) |
|---|---|---|---|
| `noBinaryFunctionCalls` | 33.54 ± 0.18 | 33.49 ± 0.75 | 53.0 → 53.0 |
| `noBinaryNestedFlwor` | 36.04 ± 1.84 | 36.17 ± 0.20 | 57.9 → 53.9 |
| `binaryPassedToFunction` | 33.47 ± 1.46 | 33.28 ± 0.49 | 69.9 → 69.9 |
| `binaryInElementConstructors` | 10.70 ± 0.09 | 10.65 ± 0.18 | 38.8 → 38.7 |

All four are at parity, and allocation per operation is equal or lower.

Worth recording how the last commit came about, because the first measurement did not look like this. Written inline into `popLocalVariables`, the frame handling made the no-binary benchmarks allocate 65.9 MB/op in two forks out of three and 57.9 in the third — bimodal, against a steady 57.9 on `develop`. Frames allocate nothing until a query registers a binary value, so that was not the registry: it was `popLocalVariables` growing past the size at which its callers can still scalar-replace. Moving the decision out of line restored both the allocation and the timings.

### Codacy:

Analyzed locally with `codacy-analysis` before opening this PR, comparing the touched files against the same files on `develop`: no new pattern instances, and one fewer finding overall.

### Not in this PR:

- Per-iteration frames in `ForExpr`. A loop's binaries all land in one frame today, so they are held until the loop ends. That is equally true on `develop`; `ForExpr` has an exact per-iteration escape set, so it is a clean follow-up.
- `BinaryValueFromBinaryString.incrementSharedReferences()` is still a no-op. Harmless while nothing lends those values, but it is now the only implementation that ignores the contract.


exist-core/src/main/java/org/exist/xquery/value/FunctionReference.java
```diff
@@ -159,8 +192,8 @@ public Sequence eval(final Sequence contextSequence, final Item contextItem) thr
      * Evaluates the referenced function.
      *
      * @param contextSequence the input sequence
-     * @param contextItem optional: the current context item
-     * @param seq array of parameters to be passed to the function
+     * @param contextItem     optional: the current context item
+     * @param seq             array of parameters to be passed to the function
```
duncdrum, line 196: what's with the whitespace here?

exist-core/src/main/java/org/exist/xquery/value/BinaryValueManager.java
```diff
@@ -30,7 +31,48 @@ public interface BinaryValueManager {
...
     void registerBinaryValueInstance(final BinaryValue binaryValue);
 
+    /**
+     * Deregister a value which has already been released.
+     *
+     * <p>A manager which does not track scopes need not implement this.</p>
+     *
+     * @param binaryValue the value to deregister
+     */
+    default void destroyBinaryValue(final BinaryValue binaryValue) {
```
duncdrum, line 41: you could get rid of the codacy warnings with an inline comment I think

exist-core/src/test/java/org/exist/util/io/FilterInputStreamCacheMonitorTest.java
```diff
@@ -161,18 +229,18 @@ public void enclosedExpressionsCleanup() throws XMLDBException {
         try {
             resourceSet = existXmldbEmbeddedServer.executeQuery(
                     "let $bin := util:binary-doc('/db/" + TEST_COLLECTION_NAME + "/icon.png')\n" +
-                    "let $embedded := <logo><image>{$bin}</image></logo>\n" +
-                    "let $embedded-2 := <other>{$bin}</other>\n" +
-                    "return xmldb:store('/db/" + TEST_COLLECTION_NAME + "', 'icon.xml', $embedded)");
+                            "let $embedded := <logo><image>{$bin}</image></logo>\n" +
```
duncdrum, line 232: wrong indent, could consider text block

